PR #5, opened 2022-04-25
Python v2 write path including cellxgene support

johnkerl:
Please see [apis/python/README.md](https://github.com/johnkerl/TileDB-SingleCell/blob/python-write-path-v2/apis/python/README.md) for information.


util/mSCGroup.py
```diff
@@ -0,0 +1,305 @@
...
+        # Here we do not use tiledb.from_numpy, so that we can have more control over the schema.
+
+        obs_dim, var_dim = np.meshgrid(anndata.obs.index, anndata.var.index)
+
+        dom = tiledb.Domain(
+            tiledb.Dim(name="obs_id", domain=(None, None), dtype="ascii"),
+            tiledb.Dim(name="var_id", domain=(None, None), dtype="ascii"),
+        )
+        att = tiledb.Attr("data")
+        sch = tiledb.ArraySchema(domain=dom, attrs=(att,), sparse=True)
```
nguyenv, line 218: I think you have to add `allows_duplicates=True` as an argument here [to match the schema that was shown above](https://tiledb.slack.com/archives/C03BY1K3GP9/p1650821218628999?thread_ts=1650813691.071909&cid=C03BY1K3GP9).

util/mSCGroup.py
```diff
@@ -0,0 +1,305 @@
...
+            print("START  SCGroup.from_h5ad %s -> %s" % (input_path, self.uri))
+
+        anndata = self.read_h5ad(input_path)
+
+        anndata = self.decategoricalize(anndata)
+
+        self.write_tiledb_group(anndata)
+
+        if self.verbose:
+            print("FINISH SCGroup.from_h5ad %s -> %s" % (input_path, self.uri))
```
nguyenv, line 80: This is a really minor comment and doesn't matter that much so don't spend time changing this now. But prefer using f-strings over %-formatting in Python 3.6+ since it is easier to read.

`print(f"FINISH SCGroup.from_h5ad {input_path} -> {self.uri}")`

util/ingestor.py
```diff
@@ -0,0 +1,41 @@
...
+#
+# Note this removes and recreates the destination TileDB group on each invocation.
+# ================================================================
+
+from mSCGroup import SCGroup
+import sys, os, shutil
+
+input_path  = 'anndata/pbmc3k_processed.h5ad'
+output_path = 'tiledb-data/pbmc3k_processed'
+if len(sys.argv) == 3:
```
nguyenv, line 27: Not a necessary change now but can replace with `argparse` module in the future.

util/ingestor.py
```diff
@@ -0,0 +1,41 @@
...
+    input_path  = sys.argv[1]
+    output_path = sys.argv[2]
+if len(sys.argv) == 2:
+    input_path  = sys.argv[1]
+    # Example 'anndata/pbmc3k_processed.h5ad' -> 'tiledb-data/pbmc3k_processed'
+    output_path = 'tiledb-data/' + os.path.splitext(os.path.basename(input_path))[0]
+
+if not os.path.exists('tiledb-data'):
```
nguyenv, line 35: Maybe someone else can chime in about this but might want to change to using `tiledb.VFS`'s `is_dir`, `create_dir`, and `remove_dir` in the future so that it can be used on multiple backends

util/mSCGroup.py
```diff
@@ -0,0 +1,305 @@
...
+        if self.verbose:
+            print("    START  WRITING %s" % (obs_or_var_uri))
+
+        tiledb.from_pandas(
+            uri=obs_or_var_uri,
+            dataframe=obs_or_var_data,
+            name=obs_or_var_name,
+            sparse=True,
+            allows_duplicates=False,
+            coords_filters=coords_filters,
```
nguyenv, line 257: hmm I believe setting `coords_filters` will be deprecated in core, and there is already a `DeprecationWarning` in TileDB-Py for this, but if this is to keep the schema/array consistent with what we already have then ignore this comment.

util/ingestor.py
johnkerl: ok -- for uploading to `tiledb://...` URIs I suspect we want either:

* Skip the upload, as "already done"
* Use a tiledb-native `overwrite` flag

I used this only for quick iteration on local-disk testing ... I can add a comment to that effect.

util/mSCGroup.py
johnkerl: @nguyenv thanks -- this was a copy from another file I encountered while assembling this; perhaps @aaronwolen has insight -- I'll test this both with & without `coords_filters` & if all is well, I'll remove this ...

util/mSCGroup.py
johnkerl: Super-easy to change, will do -- thanks @nguyenv ! :)

util/util_tiledb.py
```diff
@@ -0,0 +1,77 @@
...
+import sys, os
+import tiledb
+
+def show_single_cell_group(uri):
+    """
+    Show some summary information about an ingested TileDB Single-Cell Group.
+    """
+    print('================================================================')
+    print('X/data:')
+    with tiledb.open(uri+'/X/data') as A:
```
nguyenv, line 18: `os.path.join(uri, "X", "data")`

johnkerl: @nguyenv will that misbehave when this client is run on Windows?

johnkerl: i'll add a comment to this effect

nguyenv: OIC. Disregard this comment then.

johnkerl: no it's good @nguyenv !! others will surely see the same thing & think the same thing so it's definitely worth pointing out why the explicit `/` :)

util/mSCGroup.py
johnkerl: @nguyenv there was `os.path.join` in another file, and this `+` here ... i went with `os.path.join` for consistency along with a portability note in `apis/python/README.md`

util/util_tiledb.py
johnkerl: @nguyenv there was `os.path.join` in another file, and this `+` here ... i went with `os.path.join` for consistency along with a portability note in `apis/python/README.md`

util/mSCGroup.py
johnkerl: https://github.com/single-cell-data/TileDB-SingleCell/pull/5/commits/1acf4e657433fee86f38bdd8fd887aeaba185f0f

aaronwolen: That's news to me but if that's the case I'll remove the option as well.

nguyenv:
Oh one more thing I forgot to mention yesterday is adding type annotations, but it really isn't important and definitely don't focus on that right now.

.github/workflows/ci.yml
```diff
@@ -36,11 +36,11 @@ jobs:
 
     - name: Install Python packages
       run: |
-        pip install -r util/requirements.txt
+        python -m pip install apis/python[ci]
```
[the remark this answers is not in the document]
johnkerl: @nguyenv thanks!! :) 
https://github.com/single-cell-data/TileDB-SingleCell/pull/5/commits/9628d4fe449c5166f4534e171fb3b75dadbe7918

apis/python/tests/test_tiledbsc.py
```diff
@@ -0,0 +1,74 @@
...
+    # to make this easier to interact with in the Python interpreter.
+    A = tiledb.open(os.path.join(output_path, 'X', 'data'))
+    df = A[:]
+    keys = [k for k in df.keys()]
+    assert keys == ['data', 'obs_id', 'var_id']
+    assert A.ndim == 2
+    A.close()
+
+    # Check obs
+    A = tiledb.open(os.path.join(output_path, 'obs'))
```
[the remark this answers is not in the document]
johnkerl: thanks @nguyenv !

https://github.com/single-cell-data/TileDB-SingleCell/pull/5/commits/d09794f3e962a29e61241f4edb58e8a58d32b222

https://github.com/single-cell-data/TileDB-SingleCell/pull/5/commits/68e9ed22cabbde1d9072b6d92765a055607ebe6b

util/ingestor.py
johnkerl: https://github.com/single-cell-data/TileDB-SingleCell/pull/5/commits/9b426cf421a8d1c3c161ad346a92e6ff21810b34

aaronwolen (approved):

ihnorton (approved):

johnkerl:
Awesome, thanks @aaronwolen and @ihnorton !

Further updates as in the `Status` section in `apis/python/README.md`, and further testing -- on more cellxgene datasets, and on comparison with the `tiledbsc` R package -- will yield more opportunities for ongoing improvement.

PR #8, opened 2022-04-26
Add config/context to python scgroup class

Shelnutt2:
Add config/context to python scgroup class to let the user set a TileDB configuration or TileDB context for opening/reading/writing the arrays and groups.


johnkerl (approved):
🚢

PR #21, opened 2022-04-28
Refactor tests to compare against a reference object

aaronwolen:
A couple changes to our existing tests that will make it easier to swap reference objects and reuse code:

- use fixtures to generate file path for the test h5ad file and (separately) read it into memory as an `anndata` object
- use the [pathlib-based `HERE` convention](https://github.com/scverse/anndata/blob/a3af650d40fd804cd866d55a674a7eeaa20fdec0/anndata/tests/test_readwrite.py#L22) to conveniently reference test datasets
- update tests to compare directly against the test `anndata` object

### DIFF
--- a/apis/python/tests/test_tiledbsc.py
+++ b/apis/python/tests/test_tiledbsc.py
@@ -11,8 +11,6 @@ HERE = Path(__file__).parent
 
 @pytest.fixture
 def h5ad_file(request):
-    # Make sure this works regardless of from what directory level the `python -m pytest ...` is invoked
-    ourdir = request.fspath.dirname
     input_path = HERE.parent / "anndata/pbmc3k_processed.h5ad"
     return input_path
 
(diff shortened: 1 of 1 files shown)


apis/python/tests/test_tiledbsc.py
```diff
@@ -5,19 +5,32 @@
...
 import os
+from pathlib import Path
 
-def test_import_anndata(request):
+HERE = Path(__file__).parent
+
+@pytest.fixture
+def h5ad_file(request):
     # Make sure this works regardless of from what directory level the `python -m pytest ...` is invoked
     ourdir = request.fspath.dirname
```
johnkerl, line 15: looks like this `ourdir` isn't needed anymore -- looks like we can remove it

i took care to make sure people could do

* `python -m pytest tests` from `.../TileDB-SingleCell/apis/python`
* `python -m pytest apis/python/tests` from `.../TileDB-SingleCell`

i just verified this still works -- you're getting the same result from `Path(__file__)`

johnkerl (approved):
🚢 
with nit

p.s. noice :D

apis/python/tests/test_tiledbsc.py
aaronwolen: Good call! I just pushed another commit to remove it.

Issue #12, opened 2022-04-27
from_h5ad writes zero-valued data into the sparse X array


bkmartinjr:
Create a new SOMA group using `from_h5ad` and a source dataset with a sparse X array.

The resulting TileDB X/data array will be sparse, but will contain zero valued data. Every coordinate (even those with zero value) is present in the resulting array.

aaronwolen:
@bkmartinjr this should be resolved by #28 but I'll wait for your confirmation before closing.

bkmartinjr:
Confirmed on a couple of test datasets. Thanks!

Issue #17, opened 2022-04-28
Examine default filters and compression


johnkerl:
Initial release of `tiledbsc-py` did not adjust these. Opportunities here for reduced disk usage in tiledb-group storage as well as lower runtime memory footprint.

johnkerl:
This is complete; #27 is follow-on.

Note #12 is similar (also affects storage space) but separate.

PR #23, opened 2022-04-28
Use CZI default parameters, pending configurability

johnkerl:
Per conversation with @bkmartinjr on Slack; also related to #17.

* On an upcoming PR we'll make such params user-specifiable
* The (quick) goal for the moment is to improve the defaults

Pleasant side effect is some disk-space reduction:

Before:

```
du -hs tiledb-data/*
237M  tiledb-data/bruce-559
192K  tiledb-data/bruce-nan
640M  tiledb-data/bruce-spatial
296K  tiledb-data/pbmc-small
299M  tiledb-data/pbmc3k_processed
```

After:

```
$ du -hs tiledb-data/*
105M  tiledb-data/bruce-559
192K  tiledb-data/bruce-nan
640M  tiledb-data/bruce-spatial
296K  tiledb-data/pbmc-small
237M  tiledb-data/pbmc3k_processed
```

### DIFF
--- a/apis/python/src/tiledbsc/soma.py
+++ b/apis/python/src/tiledbsc/soma.py
@@ -185,10 +185,10 @@ class SOMA():
         base_group.add(uri=X_uri, relative=False, name="X")
 
         # ----------------------------------------------------------------
-        obs_uri = self.write_obs_or_var(anndata.obs, "obs")
+        obs_uri = self.write_obs_or_var(anndata.obs, "obs", 256)
         base_group.add(uri=obs_uri, relative=False, name="obs")
 
-        var_uri = self.write_obs_or_var(anndata.var, "var")
+        var_uri = self.write_obs_or_var(anndata.var, "var", 2048)
         base_group.add(uri=var_uri, relative=False, name="var")
 
         # ----------------------------------------------------------------
@@ -258,7 +258,7 @@ class SOMA():
 
         dom = tiledb.Domain(
             tiledb.Dim(name="obs_id", domain=(None, None), dtype="ascii", filters=[tiledb.RleFilter()]),
-            tiledb.Dim(name="var_id", domain=(None, None), dtype="ascii", filters=[tiledb.ZstdFilter()]),
+            tiledb.Dim(name="var_id", domain=(None, None), dtype="ascii", filters=[tiledb.ZstdFilter(level=22)]),
             ctx=self.ctx
         )
 
@@ -274,6 +274,9 @@ class SOMA():
             sparse=True,
             allows_duplicates=True,
             offsets_filters=[tiledb.DoubleDeltaFilter(), tiledb.BitWidthReductionFilter(), tiledb.ZstdFilter()],
+            capacity=100000,
+            cell_order='row-major',
+            tile_order='col-major',
             ctx=self.ctx
         )
         tiledb.Array.create(X_array_uri, sch, ctx=self.ctx)
@@ -296,7 +299,7 @@ class SOMA():
         return X_array_uri
 
     # ----------------------------------------------------------------
-    def write_obs_or_var(self, obs_or_var_data, obs_or_var_name: str):
+    def write_obs_or_var(self, obs_or_var_data, obs_or_var_name: str, extent: int):
         """
         Populates the obs/ or var/ subgroup for a SOMA object.
         First argument is anndata.obs or anndata.var; second is "obs" or "var".  In the reference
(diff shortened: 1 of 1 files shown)


apis/python/src/tiledbsc/scgroup.py
```diff
@@ -234,7 +234,7 @@ def write_X(self, anndata: ad.AnnData):
 
         dom = tiledb.Domain(
             tiledb.Dim(name="obs_id", domain=(None, None), dtype="ascii", filters=[tiledb.RleFilter()]),
-            tiledb.Dim(name="var_id", domain=(None, None), dtype="ascii", filters=[tiledb.ZstdFilter()]),
+            tiledb.Dim(name="var_id", domain=(None, None), dtype="ascii", filters=[tiledb.ZstdFilter(level=22)]),
```
Shelnutt2, line 237: `level=22` might incur significant overhead/slowness for decompression aka reads. I'd want us to check this before we merge this. We might over trade ondisk size compared to read performance.

apis/python/src/tiledbsc/scgroup.py
```diff
@@ -249,6 +249,9 @@ def write_X(self, anndata: ad.AnnData):
             sparse=True,
             allows_duplicates=True,
             offsets_filters=[tiledb.DoubleDeltaFilter(), tiledb.BitWidthReductionFilter(), tiledb.ZstdFilter()],
+            capacity=100000,
+            cell_order='row-major',
+            tile_order='col-major',
```
Shelnutt2, line 254: Whats the rational for col-major ordering of the tiles?

apis/python/src/tiledbsc/scgroup.py
```diff
@@ -355,6 +361,8 @@ def write_obs_or_var(self, obs_or_var_data, obs_or_var_name: str):
             offsets_filters=offsets_filters,
             attr_filters=attr_filters,
             dim_filters=dim_filters,
+            capacity=100000,
+            tile=extent,
```
Shelnutt2, line 365: Tile extent does not have the same effect for sparse as it does dense. For sparse this _only_ effects sorting. Capacity is the main things to be concerned with here.

apis/python/src/tiledbsc/scgroup.py
johnkerl: Info from @bkmartinjr on Slack ...

apis/python/src/tiledbsc/scgroup.py
johnkerl: ok thanks! this is a shameless copy-paste of CZI's 'current best' but it's not strictly necessary to use that here _as a default_ --_ soon (next day or two, 🤞) we'll support exposing tiledb schema parameters to the `tiledbsc-r` user via args in the soma-constructor -- this can wait.

apis/python/src/tiledbsc/scgroup.py
bkmartinjr: Our own access patterns require a balance between two access patterns and we empirically found this is the best config.  

In our use cases, the dominant slices are:
    X[lots_of_scattered_row, :]
and
    X[:, small_number_of_columns]

Happy to share details if helpful.

I suspect that this is yet another hint that the API needs to allow caller-specified schema hints

apis/python/src/tiledbsc/scgroup.py
bkmartinjr: we found it was a good trade where the data was on S3 (which is normal for us) - we just threw a lot of CPU at it and it was net postive

apis/python/src/tiledbsc/scgroup.py
johnkerl: > I suspect that this is yet another hint that the API needs to allow caller-specified schema hints

> it's not strictly necessary to use that here as a default --_ soon (next day or two, 🤞) we'll support exposing tiledb schema parameters to the tiledbsc-r user via args in the soma-constructor -- this can wait until then.

@bkmartinjr 100% we are on the same page :)

Shelnutt2 (approved):

apis/python/src/tiledbsc/scgroup.py
Shelnutt2: Thanks for the details @bkmartinjr , I agree making it user configurable will be best long term. For now, happy to accept this!

johnkerl:
OK thanks @Shelnutt2 and @bkmartinjr -- let's keep these as defaults for the immediate term & leave them until we get user-configurable parameters going (soon -- #27)

Currently we're wanting to resolve #12, #17, #26 as higher-priority.

Issue #16, opened 2022-04-28
Switch default unit-test data from pbmc3k_processed to pbmc_small


johnkerl:
This will make `python -m pytest apis/python/tests`, as well as CI, run faster.

bkmartinjr:
Suggestion: include a sparse and dense variant, as the code paths are distinct.

PR #25, opened 2022-04-28
Speed up unit tests

johnkerl:
Context: #16.

Use a smaller file as the default test-data file. This takes execution time of `python -m pytest tests` from over to minute down to under 5 seconds. This helps not only productivity, but moreover, increases confidence: the test no longer appears to be 'hanging'.

Building atop https://github.com/single-cell-data/TileDB-SingleCell/pull/21 this is now a one-line change.

Note: we will add more economy-sized data files (including sparse and dense variants) in the future. For the moment -- while there's a single test-data file being regressed against -- we are picking an efficient, yet representative, file.


bkmartinjr:
@johnkerl - I can also hand-create you test H5ADs with different configs & sizes as you need them.  Just LMK what you want.  For example, we internally use some small files that have a wider diversity of dtypes, etc.

bkmartinjr (approved):
LGTM.

johnkerl:
> @johnkerl - I can also hand-create you test H5ADs with different configs & sizes as you need them. Just LMK what you want. For example, we internally use some small files that have a wider diversity of dtypes, etc.

@bkmartinjr that would be fantastic! :) 

I defer to @aaronwolen, @Shelnutt2 , and others with more context in single-cell -- for myself I would say what I've been trying to accumulate so far is:

* Smaller files for unit-test/CI in GitHub; larger ones we can share in S3 & not need to keep here in GitHub
* Diversity of array types: `csc_matrix` / `csr_matrix` / `numpy.ndarray` / etc
* Diversity of dtypes
* Really looking to surface diversity-related issues sooner than later -- less concern for obscure corner-cases; more concern for "common variants" people are likely to encounter

aaronwolen (approved):

Issue #26, opened 2022-04-28
Compare/synchronize storage between tiledbsc-py and tiledbsc-r


johnkerl:
# Summary

Compare `pbmc_small` as written by

* `tiledbsc-r`: [https://github.com/TileDB-Inc/tiledbsc](https://github.com/TileDB-Inc/tiledbsc)
* `tiledbsc-py`: [https://github.com/single-cell-data/TileDB-SingleCell/tree/main/apis/python](https://github.com/single-cell-data/TileDB-SingleCell/tree/main/apis/python)

Full write-up on this PR's [pbmc-small.md](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/sc-17240/pbmc-small-compare/apis/python/pbmc-small.md).

# Things to fix

* `X` R dim names are `var_id,obs_id` vs Python `obs_id,var_id` _and_ the matrix itself is also transposed
  * This is deferred to the separate issue https://github.com/single-cell-data/TileDB-SingleCell/issues/19
* `X` R attr name is `value`, Python attr name is `data`
  * [DONE on #24]
* `obs` R dim name `obs_id`, Python attr name `__tiledb_rows`
  * [DONE on #11]
* `var` similarly
  * [DONE on #11]
* `var` R has attribute `highly_variable` which Python lacks
  * This is because there is no `highly_variable` in our `pbmc-small.h5ad`; not a SOMA problem but rather due an issue in the `seurat`/`anndata-py` datapath that [produced this file](https://github.com/single-cell-data/TileDB-SingleCell/blob/main/apis/python/anndata/README.md)
* `obsm/tsne` R dim `obs_id`, attrs `tSNE_1,tSNE_2`; Python dims `__dim_0,__dim_1` and attr `""`
  * [DONE on #32]
* Also, switch CI / `python -m pytest tests` from `pbmc3k_processed.h5ad` (big) to `pbmc_small.h5ad` (small) to make them run more quickly
  * This is deferred to the separate issue https://github.com/single-cell-data/TileDB-SingleCell/issues/16
  * [DONE on #25]

# Setup

## R setup

```
install.packages("Seurat")
remotes::install_github("mojaveazure/seurat-disk")
install.packages('anndata')
```

## Python setup

```
pip install anndata
```

# Write pbmc_small datasets

## Convert Seurat to tiledbsc-r

```
library(tiledbsc)

data("pbmc_small", package = "SeuratObject")
[...]

PR #31, opened 2022-04-29
Let desc-tiledb.py handle more datasets

johnkerl:
Sparse/dense is still not uniformized between `tiledbsc-r` and `tiledbsc-py` (see also #26). Meawhile, our `desc-tiledb.py` inside `tiledbsc-py` was failing to print `obsm` et al. for some datasets written by `tiledbsc-r`.


aaronwolen (approved):
LGTM

Issue #50, opened 2022-05-02
Unique values in adata.raw.var are dropped by from_anndata


bkmartinjr:
Using from_anndata will fail to save the contents of `.raw.var`.

ScanPy/AnnData typically assume that `.raw.var` is a superset of `.var`, so it needs to be retained alongside `.raw.X` (they should share the `var_id` dimension)

johnkerl:
This has been resolved for quite some time; sorry for not having closed the issue sooner.

```
tiledbsc.__version__ 0.0.7.dev11
tiledb.__version__   0.16.2.dev5
core version         2.11.0

>>> soma = tiledbsc.SOMA('tiledb-data/pbmc3k_processed')

>>> soma.raw.X.data.df()
                                 value
obs_id           var_id
AAACATACAACCAC-1 ABCD4        0.693147
                 ABRACL       0.693147
                 AC003075.4   1.098612
                 AC009133.14  0.693147
                 ACAA2        0.693147
...                                ...
TTTGCATGCCTCAC-1 ZNF529       0.693147
                 ZNF557       0.693147
                 ZNF721       0.693147
                 ZNF827       0.693147
                 ZNF91        0.693147

[2238732 rows x 1 columns]

>>> soma.raw.var.df()
              n_cells
var_id
7SK-2               9
A1BG              167
A1BG-AS1           32
A2M                 8
A2M-AS1            30
...               ...
ZYX               403
ZZEF1              73
ZZZ3               60
hsa-mir-1199        6
hsa-mir-8072       33

[13714 rows x 1 columns]
>>>
```

PR #52, opened 2022-05-02
initial test for saving raw X and var

bkmartinjr:
Two additional tests to check the behavior of from_anndata for `.raw.X` and `.raw.var`.

Note that the `.raw.var` test is disabled due to issue #50


johnkerl:
@bkmartinjr sorry about the force-push -- I rebased this atop #64 & modified it to match

PR #53, opened 2022-05-02
Use dtypes from anndata objects as much as possible

johnkerl:
For https://github.com/single-cell-data/TileDB-SingleCell/issues/42


bkmartinjr (approved):
LGTM.   I also confirmed on a few test datasets.

aaronwolen (approved):

Issue #37, opened 2022-04-29
Chunked writes


johnkerl:
As of when this issue was created, https://github.com/single-cell-data/TileDB-SingleCell/tree/main/apis/python ingestion uses sparse matrices 'all at once' using in-memory conversion to COO. As @bkmartinjr and @stavrospapadopoulos point out, almost all sparse single-cell datasets are in CSR or CSC -- we could do writes incrementally/streaming using a constrained budget.

An easier step along the way (@stavrospapadopoulos): not to convert the whole `anndata` object to COO, but instead do it incrementally.

Steps:

* `X` in CSR format -- ingest as chunks of COO
* `X` in CSC format -- ditto (note, these are less common in the wild)
* Get some rough cost/benefit numbers on fragment-consolidation

Status of 2022-05-04:

* This is working well given the caveats on #55 as detailed [here](https://github.com/single-cell-data/TileDB-SingleCell/blob/main/apis/python/README-csr-ingest.md). This requires array-consolidation (offline) for improved performance.
* Once the global-order-write feature is ported to Python we'll be able to dispense with that -- rather than creating multiple fragments needing offline consolidation, we'll be able to do chunked writes which will update a single fragment.
* Another issue: pending a TileDB core release which includes https://github.com/TileDB-Inc/TileDB/pull/3143 we are currently using `tile_order=row_major` to avoid a segv when consolidating with `tiledb_order=cell_major`. Note that https://github.com/TileDB-Inc/TileDB/pull/3143 was backported to the 2.9 branch in https://github.com/TileDB-Inc/TileDB/pull/3146 and is scheduled for the upcoming core 2.9 release.

johnkerl:
See #86 for a requested continuation

johnkerl:
This is already working

PR #55, opened 2022-05-03
Enable chunked-CSR writes

johnkerl:
Context: #37.

Caveats are as noted within `README-csr-ingest.md` which is more easily reviewable at
https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/sc-17395/chunked-csr-enable/apis/python/README-csr-ingest.md


bkmartinjr (approved):
LGTM

johnkerl:
thanks @bkmartinjr !!

PR #56, opened 2022-05-03
add uns unit test

bkmartinjr:
Added a starter unit test for uns, focused on the very simple needs of the initial work slice.

Test is currently disabled, pending implementation of `uns` support.

### DIFF
--- a/apis/python/tests/test_anndata_uns.py
+++ b/apis/python/tests/test_anndata_uns.py
@@ -1,0 +1,39 @@
+from anndata import AnnData
+import tiledb
+from tiledbsc import SOMA
+import pandas as pd
+import numpy as np
+from scipy import sparse
+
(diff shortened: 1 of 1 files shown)


apis/python/tests/test_anndata_uns.py
```diff
@@ -0,0 +1,39 @@
...
+            "number": 1,
+            "string": "a string",
+            "list of strings": list(str(i) for i in range(10)),
+            "simple dict": {"A": 0, "B": "one"},
+        },
+    )
+    SOMA(tmp_path.as_posix()).from_anndata(adata)
+
+    # TODO: persistance for UNS is not yet defined. Update when it is.
+    assert (tmp_path / "uns").exists()
```
johnkerl, line 37: :)

Issue #60, opened 2022-05-04
from_anndata: obs/var categorical columns containing "nan" (not in category) fail to persist


bkmartinjr:
If an obs/var column is a Pandas Categorical, and contains a nan (ie, value not in category), it will fail.  Root cause is incorrect type handling of categoricals in `decategoricalize`, and is related to issue #30 

Example error of a string categorical with a nan:
```
    START  WRITING ./temp/out/obs
Traceback (most recent call last):
  File "tiledb/libtiledb.pyx", line 344, in tiledb.libtiledb._write_array
tiledb.cc.TileDBError: Unexpected object type in string conversion (tiledb/npbuffer.cc:289)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/bruce/TileDB-SingleCell/apis/python/temp/ingest_sweep.py", line 106, in <module>
    sys.exit(main())
  File "/home/bruce/TileDB-SingleCell/apis/python/temp/ingest_sweep.py", line 38, in main
    sc.from_anndata(ad)
  File "/home/bruce/TileDB-SingleCell/apis/python/venv/lib/python3.9/site-packages/tiledbsc/soma.py", line 96, in from_anndata
    self.write_tiledb_group(anndata)
  File "/home/bruce/TileDB-SingleCell/apis/python/venv/lib/python3.9/site-packages/tiledbsc/soma.py", line 223, in write_tiledb_group
    obs_uri = self.write_obs_or_var(anndata.obs, "obs", 256)
  File "/home/bruce/TileDB-SingleCell/apis/python/venv/lib/python3.9/site-packages/tiledbsc/soma.py", line 349, in write_obs_or_var
    tiledb.from_pandas(
  File "/home/bruce/TileDB-SingleCell/apis/python/venv/lib/python3.9/site-packages/tiledb/dataframe_.py", line 448, in from_pandas
    _from_pandas(uri, dataframe, tiledb_args)
  File "/home/bruce/TileDB-SingleCell/apis/python/venv/lib/python3.9/site-packages/tiledb/dataframe_.py", line 520, in _from_pandas
    _write_array(
  File "/home/bruce/TileDB-SingleCell/apis/python/venv/lib/python3.9/site-packages/tiledb/dataframe_.py", line 604, in _write_array
    libtiledb._setitem_impl_sparse(A, tuple(coords), write_dict, nullmaps)
[...]

PR #61, opened 2022-05-04
add test case for categorical undefined/nan values

bkmartinjr:
Test case for issue #60 

Note: the test case assumes the Pandas `to_numpy()` semantics, rather than the `astype()` semantics when handling "NA"s (aka a categorical nan).

Specifically, an integer categorical with an NA, eg,
> pd.Series(pd.Categorical([0,1,2,3], categories=np.array([1,2,3], dtype=np.int32)))

Will raise when using astype() as there is no IEEE NaN support in int32 arrays.  `to_numpy` converts it to min value of that type rather than raising an error.

```
In [1]: pd.Series(pd.Categorical([0,1,2,3], categories=np.array([1,2,3], dtype=np.int32)))
Out[1]: 
0    NaN
1      1
2      2
3      3
dtype: category
Categories (3, int64): [1, 2, 3]

In [2]: pd.Series(pd.Categorical([0,1,2,3], categories=np.array([1,2,3], dtype=np.int32))).to_numpy(np.int32)
Out[2]: array([-2147483648,           1,           2,           3], dtype=int32)

In [3]: pd.Series(pd.Categorical([0,1,2,3], categories=np.array([1,2,3], dtype=np.int32))).astype(np.int32)
ValueError                                Traceback (most recent call last)
...
```

### DIFF
--- a/apis/python/tests/test_type_diversity.py
+++ b/apis/python/tests/test_type_diversity.py
@@ -306,3 +306,86 @@ def test_from_anndata_zero_length_str(tmp_path):
             adata.obs.keys()
         )
         assert adata.n_obs == len(obs.query(dims=[]).df[:])
+
+
+# TODO: re-enable when #60 is resolved
+@pytest.mark.skip(reason="Fails: filed as issue #60")
+def test_from_anndata_category_nans(tmp_path):
+    """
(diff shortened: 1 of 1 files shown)


bkmartinjr:
Closing - superceded

PR #65, opened 2022-05-06
Ingest uns

johnkerl:


apis/python/src/tiledbsc/soma.py
```diff
@@ -464,6 +471,104 @@ def write_raw_group(self, raw_group_uri: str, raw: ad.Raw):
...
+            uns_group.add(uri=component_uri, relative=False, name=key)
+
+        uns_group.close()
+
+        if self.verbose:
+            print(util.format_elapsed(s, f"    FINISH WRITING {uns_group_uri}"))
+
+
+    # ----------------------------------------------------------------
+    def write_uns_pandas_dataframe(self, array_uri, array: pd.DataFrame):
```
aaronwolen, line 543: Any reason this arg is named `array` rather than `dataframe`?

aaronwolen (approved):
One minor question but otherwise looks great

apis/python/src/tiledbsc/soma.py
johnkerl: no _good_ reason @aaronwolen ;)

PR #70, opened 2022-05-11
Corner-case with raw dense X having zero items

johnkerl:
Context: https://github.com/single-cell-data/TileDB-SingleCell/issues/69


Shelnutt2 (approved):

bkmartinjr (approved):
LGTM

Issue #72, opened 2022-05-11
Implement Python classes for SOMA arrays/groups


johnkerl:
As of the initial 'ingest-only' version we have at present, we have a single `SOMA` class with lots of methods. Those methods should be moved to individual classes:

* For parity with `tiledbsc-r`
* To facilitate comparisons `tiledbsc-py` <-> `tiledbsc-r`
* To make it easier to complete readback/outgest logic
* To faciliate development of the nascent SOMA-query API

johnkerl:
#78 is merged.

PR #78, opened 2022-05-12
Implement class structure for SOMA arrays and groups

johnkerl:
# Overview

As of the initial 'ingest-only' version we have at present, we have a single `SOMA` class with lots of methods. Those methods should be moved to individual classes:

* For parity with `tiledbsc-r`
* To facilitate comparisons `tiledbsc-py` <-> `tiledbsc-r`
* To make it easier to complete readback/outgest logic
* To faciliate development of the nascent SOMA-query API

Context: #72 

# Notes

* This is a big, cross-cutting PR -- which is a card which I try to rarely play in my career. Reviewer, I owe you a beer. If you thought _you_ owed _me_ one, then I now owe you two. ;)
* Names are largely drawn from [tiledbsc-r](https://github.com/TileDB-Inc/tiledbsc/tree/main/R).
* Rest assured all unit tests that were ever passing, are still passing.

# Review guide

* Base object [TileDBObject](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/sc-17395-class-structure/apis/python/src/tiledbsc/tiledb_object.py)
  * Propagates basic things like config options, `ctx`, verbosity, indent level for printing to the terminal, etc
  * For groups & arrays both
* Group object [TileDBGroup](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/sc-17395-class-structure/apis/python/src/tiledbsc/tiledb_group.py)
  * Simple wrapper around tiledb group. When we add a `soma.uns['neighbors']['params']['method']` via overloading the index operator, that'll go right here.
* Array object [TileDBArray](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/sc-17395-class-structure/apis/python/src/tiledbsc/tiledb_array.py)
* Various array/group classes like in `tiledbsc-py`
  * See [__init.py__](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/sc-17395-class-structure/apis/python/src/tiledbsc/__init__.py) for a quick view
  * See also [this README.md section](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/sc-17395-class-structure/apis/python/README.md#tiledb-group-structure) for my "rosetta stone" to help keep all this straight in my head
* [soma.py](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/sc-17395-class-structure/apis/python/src/tiledbsc/soma.py) is now quite a bit slimmed down

# What isn't done on this PR

* The ingestor still does `shutil.rmtree`. In-place updates are pending my finding the right overwrite-mode syntax for `tiledb.from_numpy` in `UnsArray`; other array-creation sites in the code appear to be working OK with in-place updates.
* TileDB-Cloud URIs aren't fully supported yet -- this will be done separately.
* Overloading the index operator so we can say things like `soma.uns['neighbors']['params']['method']` will be powerful and liberating, but, will be done on a separate PR.
* Likewise, currently-in-plan SOMA-API updates are not here. Yet rest assured that the current class-strucuting work will make that easier, quicker, and less error-prone.
* Readback via `to_anndata` is largely in place, as a pure-Python prototype. It still has a bug on readback of `obsp` and `varp` (same bug that was present before this PR); needs dealing with separately.


apis/python/src/tiledbsc/annotation_dataframe.py
```diff
@@ -0,0 +1,118 @@
...
+            ctx=self.ctx,
+            mode=mode,
+        )
+
+        if self.verbose:
+            print(util.format_elapsed(s, f"{self.indent}FINISH WRITING {self.uri}"))
+
+
+    # ----------------------------------------------------------------
+    def to_dataframe(self) -> pandas.DataFrame:
```
aaronwolen, line 100: I'm probably just confused about the intent here but why not return just the dataframe here? The labels are already contained in the dataframe's index if they need to be referenced later.

apis/python/src/tiledbsc/annotation_dataframe.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        Reads the TileDB obs or var array and returns a type of pandas dataframe
+        and dimension values.
+        """
+
+        if self.verbose:
+            s = util.get_start_stamp()
+            print(f"{self.indent}START  read {self.uri}")
+
```
aaronwolen, line 109: with tiledb.open(self.uri) as A:
    retval = A.df[:]

The `retval` DataFrame should already have index `obs_id`/`var_id`, so no need to set and name it separately.

aaronwolen (approved):
I had a couple of questions/comments but nothing preventing 🚢 . Looks great!

aaronwolen:
> Reviewer, I owe you a beer. If you thought you owed me one, then I now owe you two. 
Noted 🍻

apis/python/src/tiledbsc/annotation_dataframe.py
aaronwolen: Disregard. In the original version I reviewed this was still `to_dataframe_and_labels()`.

johnkerl: @aaronwolen great minds think alike! I had the same conclusion independently when trying to use this PR to build another on top of it. "Why not just return the dataframe? The labels are already present as df.index ..." :D

apis/python/src/tiledbsc/soma.py
```diff
@@ -23,8 +23,8 @@
 
 class SOMA(TileDBGroup):
     """ Single-cell group
-    Class for representing a group of TileDB arrays that consitute an SOMA ('stack of matrices,
-    annotated') which includes:
+    Class for representing a group of TileDB groups/arrays that consitute an SOMA ('stack of matrices, annotated')
```
aaronwolen, line 26: Typo: consitute

apis/python/src/tiledbsc/soma.py
```diff
@@ -141,721 +154,82 @@ def read_10x(self, input_path: str):
...
-    #         +-- vpbar: array
-
-    def write_tiledb_group(self, anndata: ad.AnnData):
+    def from_anndata(self, anndata: ad.AnnData):
         """
         Top-level writer method for creating a TileDB group for a SOMA object.
         """
 
+        # Without _at least_ an index, there is nothing to indicate the dimension indices.
+        if anndata.obs.index.empty or anndata.var.index.empty:
```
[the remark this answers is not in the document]
johnkerl: @aaronwolen I think we're talking about two different things.

* If `obs` and/or `var` have zero attributes then we do create an attribute-less array that serves to store the full set of tim labels.
  * https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/sc-17395-class-structure/apis/python/src/tiledbsc/util_ann.py#L236
  * https://github.com/single-cell-data/TileDB-SingleCell/pull/66
  * https://github.com/single-cell-data/TileDB-SingleCell/issues/45
* But in this check we aren't checking if there are zero _attributes_ but rather zero _labels_ which (IIUC) should remain an error condition ...

apis/python/src/tiledbsc/annotation_dataframe.py
johnkerl: @aaronwolen nice, thanks! :)

johnkerl: ... alas,

```
KeyError: b'AAATTCGAATCACG'
```

on outgest -- issue is with the 'before' code the df is like

```
                   orig.ident  nCount_RNA  nFeature_RNA  RNA_snn_res.0.8  letter.idents groups  RNA_snn_res.1
obs_id
b'AAATTCGAATCACG'           0       327.0            62                1              1     g2              1
b'AAGCAAGAGCTTAG'           0       126.0            48                0              0     g1              0
b'AAGCGACTTTGACG'           0       443.0            77                1              1     g1              1
b'AATGCGTGGACGGA'           0       389.0            73                1              1     g1              1
b'AATGTTGACAGTCA'           0       100.0            41                0              0     g1              0
...                       ...         ...           ...              ...            ...    ...            ...
b'TTACGTACGTTCAG'           0       228.0            39                0              0     g1              0
b'TTGAGGACTACGCA'           0       787.0            88                0              0     g1              2
b'TTGCATTGAGCTAC'           0       104.0            40                0              0     g2              2
b'TTGGTACTGAATCC'           0       135.0            45                0              0     g1              2
b'TTTAGCTGTACTCT'           0       462.0            86                1              1     g1              1
```

and after

```
                orig.ident  nCount_RNA  nFeature_RNA  RNA_snn_res.0.8  letter.idents groups  RNA_snn_res.1
obs_id
AAATTCGAATCACG           0       327.0            62                1              1     g2              1
AAGCAAGAGCTTAG           0       126.0            48                0              0     g1              0
AAGCGACTTTGACG           0       443.0            77                1              1     g1              1
AATGCGTGGACGGA           0       389.0            73                1              1     g1              1
AATGTTGACAGTCA           0       100.0            41                0              0     g1              0
...                    ...         ...           ...              ...            ...    ...            ...
TTACGTACGTTCAG           0       228.0            39                0              0     g1              0
TTGAGGACTACGCA           0       787.0            88                0              0     g1              2
TTGCATTGAGCTAC           0       104.0            40                0              0     g2              2
TTGGTACTGAATCC           0       135.0            45                0              0     g1              2
TTTAGCTGTACTCT           0       462.0            86                1              1     g1              1
```

namely a bytes-vs-strings difference. I'll add a clarifying comment.

Thank you!! :D

PR #79, opened 2022-05-13
Support basic group-indexing operations on SOMA members [WIP]

johnkerl:
These are a few basic operations we'll need to do on SOMA objects. This is prep for the upcoming SOMA-query-API work.


aaronwolen (approved):
Nice!

Issue #74, opened 2022-05-11
from_anndata ingestion failure


bkmartinjr:
```
In [10]: ad = anndata.read_h5ad("/home/bruce/corpora-data-prod/d4db74ad-a129-4b1a-b9da-1b30db86bbe4.h5ad")

In [11]: tiledbsc.SOMA("/tmp/test3").from_anndata(ad)
START  SOMA.from_ann
  START  DECATEGORICALIZING
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [11], in <cell line: 1>()
----> 1 tiledbsc.SOMA("/tmp/test3").from_anndata(ad)

File ~/TileDB-SingleCell/apis/python/src/tiledbsc/soma.py:85, in SOMA.from_anndata(self, anndata)
     82     s = util.get_start_stamp()
     83     print(f"START  SOMA.from_ann")
---> 85 anndata = self.decategoricalize(anndata)
     86 self.write_tiledb_group(anndata)
     88 if self.verbose:

File ~/TileDB-SingleCell/apis/python/src/tiledbsc/soma.py:175, in SOMA.decategoricalize(self, anndata)
    172     new_var = anndata.var
    174 for key in anndata.obsm.keys():
--> 175     anndata.obsm[key] = util._to_tiledb_supported_array_type(anndata.obsm[key])
    176 for key in anndata.varm.keys():
    177     anndata.varm[key] = util._to_tiledb_supported_array_type(anndata.varm[key])

File ~/TileDB-SingleCell/apis/python/src/tiledbsc/util.py:100, in _to_tiledb_supported_array_type(x)
     95     return dtype
     97 # If a Pandas categorical, use the type of the underlying category.
     98 # If the array contains NaN/NA, and the primitive is unable to represent
     99 # a reasonable facsimile, i.e. not string or float, raise.
--> 100 if pd.api.types.is_categorical_dtype(x.dtype):
    101     categories = x.cat.categories
    102     cat_dtype = categories.dtype

File ~/TileDB-SingleCell/apis/python/venv/lib/python3.9/site-packages/pandas/core/generic.py:5575, in NDFrame.__getattr__(self, name)
   5568 if (
   5569     name not in self._internal_names_set
   5570     and name not in self._metadata
   5571     and name not in self._accessors
   5572     and self._info_axis._can_hold_identifiers_and_holds_name(name)
   5573 ):
   5574     return self[name]
[...]

PR #75, opened 2022-05-12
Fix issue #74: pandas.DataFrame, not numpy.ndarray, in an obsm slot

johnkerl:
Context: #74

Problem:

* From #74 it seems clear enough an `obsm` slot has a `pandas.DataFrame` in it, not `numpy.ndarray`, and we crash while attempting to access its (non-existent) `.dtype`
* The fix seems straightforward enough
* However, attempting to write a unit-test case to cover this, I can't even _construct_ an `anndata` object with `pandas.DataFrame` in its `obsm` slot :( -- making me think there is something I don't understand.

```
tests/test_type_diversity.py:411:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/anndata/_core/anndata.py:291: in __init__
    self._init_as_actual(
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/anndata/_core/anndata.py:512: in _init_as_actual
    self._obsm = AxisArrays(self, 0, vals=convert_to_dict(obsm))
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/anndata/_core/aligned_mapping.py:235: in __init__
    self.update(vals)
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/_collections_abc.py:856: in update
    self[key] = other[key]
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/anndata/_core/aligned_mapping.py:151: in __setitem__
    value = self._validate_value(value, key)
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/anndata/_core/aligned_mapping.py:215: in _validate_value
    return super()._validate_value(val, key)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = AxisArrays with keys: is_a_numpy_ndarray
val = <bound method DataFrame.reset_index of    value ...>
[...]

PR #84, opened 2022-05-14
Fix issue #74: pandas.DataFrame, not numpy.ndarray, in an obsm slot

johnkerl:
Re-do of PR #75: see narrative there. Solves issue #74.


bkmartinjr (approved):
LGTM.  Tested on previous failures and all pass.

PR #98, opened 2022-05-19
Move tools to tools/ directory

johnkerl:
There are getting to be a few of these now -- organizing them in their own place helps neaten the visual presentation of the repository, and helps highlight their purpose.


apis/python/interactor.py
```diff
@@ -1,43 +0,0 @@
-#!/usr/bin/env python -i
```
johnkerl, line 1: I haven't used this script in ages (a week, if not two) -- if anyone else uses it lmk, else I plan to jettison it.

aaronwolen (approved):

PR #109, opened 2022-05-23
Flag some for-internal-use methods more clearly as such

johnkerl:
While Python of course does not have method-access categories such as `public`, `protected`, or `private`, nonetheless we can more clearly mark some for-internal-use methods as being such.

Also, when a user does something like `soma.obs.` and then tries to tab-complete in the Python interpreter, the methods/attributes having leading underscores are left out by default, allowing the user to focus on methods which truly belong to the SOMA API and which aren't simply tiledb-specific implementation details toward that API. (Implementation-oriented users can tab-complete with `soma.obs._` if they like, and/or use `dir(soma.obs)`.)

This becomes ever more important as we flesh out the SOMA API last week and this week.

### DIFF
--- a/apis/python/README.md
+++ b/apis/python/README.md
@@ -248,7 +248,7 @@ VDAC3            1.1250     30.971519               8.986513                   2
 ```
 
 ```
->>> soma.obsm.get_member_names()
+>>> soma.obsm._get_member_names()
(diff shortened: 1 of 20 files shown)


aaronwolen (approved):
This is great.

PR #115, opened 2022-05-24
Store uns scalars as group metadata, not 1D arrays

johnkerl:
Context: #88 

Example data:

```
>>> soma = tiledbsc.SOMA('tiledb-data/brown-adipose-tissue-mouse')

>>> soma.uns.show()

uns:
X_normalization: log1p, base e, scaled
publication_doi: https://doi.org/10.1038/s41586-020-2496-1
schema_version: 2.0.0
title: Brown adipose tissue — A single-cell transcriptomic atlas characterizes ageing tissues in the mouse

uns/layer_descriptions:
X: log(counts_per_thousand+1)
raw.X: raw

uns/neighbors:

uns/neighbors/params:
uns/neighbors/params/n_pcs/
[5]
uns/neighbors/params/metric/
['euclidean']
uns/neighbors/params/method/
['umap']
uns/neighbors/params/n_neighbors/
[15]

uns/louvain:

uns/louvain/params:
uns/louvain/params/random_state/
[0]

uns/leiden:

uns/leiden/params:
uns/leiden/params/random_state/
[0]
uns/leiden/params/n_iterations/
[-1]
uns/leiden/params/resolution/
[1]
uns/age_colors/
['#e1f3b2' '#97d6b9' '#1f80b8']

uns/pca:
uns/pca/variance/
[113.979164   97.67646    67.93497    56.686123   43.88914    23.913618
  23.484827   22.598204   20.3256     16.948996   14.354713   13.151188
  11.807495   10.862301   10.696552   10.350647   10.0984745   9.615314
   9.299154    8.849509    8.519048    8.121199    7.8701077   7.6040773
   7.582742    7.469643    7.2797194   7.218048    7.066573    6.9701815
   6.914394    6.772224    6.7097783   6.620041    6.520022    6.416898
   6.396799    6.333943    6.2921176   6.22901     6.15171     6.124492
   6.089446    6.0435967   6.022932    5.935481    5.910855    5.874158
   5.8458056   5.801149 ]
uns/pca/variance_ratio/
[0.03001015 0.02571773 0.01788694 0.01492518 0.01155579 0.00629634
 0.00618344 0.00595    0.00535163 0.00446259 0.00377953 0.00346264
 0.00310886 0.00285999 0.00281635 0.00272527 0.00265888 0.00253166
 0.00244842 0.00233003 0.00224302 0.00213827 0.00207216 0.00200212
 0.0019965  0.00196672 0.00191671 0.00190048 0.00186059 0.00183521
[...]

Issue #106, opened 2022-05-20
Reversible typecasts (parent task)


johnkerl:
When we ingest data into TileDB from, say, `AnnData`, we need to remap a few things to be storable:

* Categorical strings to strings
* `bool` to `uint8`
* Unicode array dimensions to ASCII dimensions -- necessary for TileDB at present
* Unicode array attributes to ASCII array attributes -- TileDB allows attributes to be Unicode, but, they are not queryable via the `QueryCondition` API. See also #99.

Source links:

* [util_ann.decategoricalize](https://github.com/single-cell-data/TileDB-SingleCell/blob/main/apis/python/src/tiledbsc/util_ann.py#L255)
* [util._to_tiledb_supported_array_type](https://github.com/single-cell-data/TileDB-SingleCell/blob/main/apis/python/src/tiledbsc/util.py#L81)

Related: #105

Item 2 will be addressed in TileDB Core 2.10, where we'll have native boolean support. The remaining ones, though, are ongoing concerns.

Proposal for item 1:

* Use TileDB group/array metadata tags
* Indicate the following at the SOMA-group level:
  * Source provenance (e.g. `anndata`, `seurat`)
  * Language and package version of the code that wrote the data
* Indicate the following at the SOMA-array level:
  * Either a dict from _all_ columns to their original types, or, just the 'exceptions' (e.g. categorical-string columns)

Proposal for items 3 and 4:

* See #99

johnkerl:
Split out remaining issues to #866 and #867. cc @maniarathi

PR #116, opened 2022-05-24
Initial group/array metadata access methods

johnkerl:
This introduces some metadata accessors but doesn't yet put them to use.

Places to use them include #106, where:

* For columns, we'll store original category-type names (for anndata categoricals which were downcast to TileDB strings)
* For arrays, we'll store information like "this was originally CSR" or "this was originally CSC"
* For `uns`, we'll store scalars no longer as length-1 arrays but rather as group metadata
* For SOMA objects, we'll store information like "this was written by `tiledbsc-r v0.1.2`

See also #115 which is stacked atop this.

Note: this is stacked atop #144, hence the higher line-count -- see https://github.com/single-cell-data/TileDB-SingleCell/pull/116/commits

### DIFF
--- a/apis/python/src/tiledbsc/tiledb_object.py
+++ b/apis/python/src/tiledbsc/tiledb_object.py
@@ -1,7 +1,7 @@
 import tiledb
 from .soma_options import SOMAOptions
 
-from typing import Optional
+from typing import Optional, List, Dict
 
 import os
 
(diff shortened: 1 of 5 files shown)


apis/python/src/tiledbsc/tiledb_object.py
```diff
@@ -80,3 +80,47 @@ def exists(self):
             raise Exception(
                 f"Internal error: expected _object_type {self._object_type()} but found {found} at {self.uri}."
             )
+
+    def metadata(self) -> Dict:
+        """
+        Returns metadata the group/array.
```
aaronwolen, line 86: Missing word: Returns metadata _from_ the group/array

PR #120, opened 2022-05-25
Implement chunked-X writes for CSC and dense

johnkerl:
Context: #37

This is validated with very thorough testing at
https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/more-chunking/apis/python/tests/test_chunked_writes.py


apis/python/src/tiledbsc/tiledb_group.py
```diff
@@ -29,7 +29,14 @@ def __init__(
         """
         See the TileDBObject constructor.
         """
-        super().__init__(uri=uri, name=name, parent=parent)
+        super().__init__(
```
johnkerl, line 32: This is an old bug -- this was failing to propagate non-null `soma_options` through the super-chain.

aaronwolen (approved):
New chunking functionality looks good to me!

We could avoid storing those large h5ad files in the repo by leveraging `scanpy.datasets.pbmc3k_processed()` to create the same anndata object and then modify as needed for each test.

PR #130, opened 2022-05-27
Add initial SOMA and SOMACollection C++ classes

gspowley:
Add SOMA and SOMACollection C++ classes with initial capabilities:
* Opening a SOMACollection and listing the SOMAs
* Opening a SOMA and listing the arrays


libtiledbsc/include/tiledbsc/soma.h
```diff
@@ -0,0 +1,46 @@
...
+    //= public static
+    //===================================================================
+    static SOMA open(std::string_view uri);
+
+    //===================================================================
+    //= public non-static
+    //===================================================================
+    SOMA(std::string_view uri);
+
+    std::map<std::string, std::string> list_arrays();
```
johnkerl, line 22: I believe this is a map from group-element name to group-element URI -- let's say that as a comment

libtiledbsc/include/tiledbsc/soma.h
```diff
@@ -0,0 +1,46 @@
...
+    //===================================================================
+    //= private static
+    //===================================================================
+
+    //===================================================================
+    //= private non-static
+    //===================================================================
+    Context ctx_;
+    std::string uri_;
+    bool group_uri_override_ = false;
```
johnkerl, line 34: Can we add a quick note (for those reading only the `.h` on a flyover) what this is for?

libtiledbsc/include/tiledbsc/soma_collection.h
```diff
@@ -0,0 +1,45 @@
...
+    //= public static
+    //===================================================================
+    static SOMACollection open(std::string_view uri);
+
+    //===================================================================
+    //= public non-static
+    //===================================================================
+    SOMACollection(std::string_view uri);
+
+    std::map<std::string, std::string> list_somas();
```
johnkerl, line 22: ditto the above -- if this is a map from name to URI, let's say that as a comment

libtiledbsc/src/cli/cli.cc
```diff
@@ -1,14 +1,38 @@
...
+        printf("Usage: %s uri\n", argv[0]);
+        return 1;
+    }
 
-    std::cout << "&sc: " << &sc << std::endl;
+    try {
+        walk_soco(argv[1]);
+    } catch (const std::exception& e) {
+        printf("ERROR: %s\n", e.what());
+    }
```
johnkerl, line 35: a nit, but i do like final newlines ... whenever someone does `cat` of a file, then their shell prompt won't appear after the last LF-less code line.

libtiledbsc/src/soma.cc
```diff
@@ -0,0 +1,60 @@
...
+//===================================================================
+SOMA SOMA::open(std::string_view uri) {
+    return SOMA(uri);
+}
+
+//===================================================================
+//= public non-static
+//===================================================================
+SOMA::SOMA(std::string_view uri) {
+    // Remove all trailing /
```
johnkerl, line 19: golden -- ❤️

libtiledbsc/src/soma.cc
```diff
@@ -0,0 +1,60 @@
...
+                        std::string(parent) + "/" + member.name().value();
+
+        if (member.type() == Object::Type::Group) {
+            // Member is a group, call recursively
+            auto subgroup = Group(ctx_, member.uri(), TILEDB_READ);
+            build_uri_map(subgroup, path);
+        } else {
+            auto uri = member.uri();
+            if (is_tiledb_uri(uri) && !is_tiledb_uri(uri_)) {
+                // Member uri is registered on tiledb://, but the root uri is
```
johnkerl, line 49: "Member URI" and "root URI" and "relative URI"

libtiledbsc/src/soma_collection.cc
```diff
@@ -0,0 +1,59 @@
...
+//= public static
+//===================================================================
+SOMACollection SOMACollection::open(std::string_view uri) {
+    return SOMACollection(uri);
+}
+
+//===================================================================
+//= public non-static
+//===================================================================
+SOMACollection::SOMACollection(std::string_view uri) {
```
johnkerl, line 18: In the Python impl I made a (perhaps overly lasagna-code)

```
SOMACollection extends TileDBGroup
SOMA extends TileDBGroup
AssayMatrix extends TileDBArray
...
TileDBGroup extends TileDBObject
TileDBArray extends TileDBObject
```

and I put things like trailing-slash strippers or what have you in the `TileDBObject` level -- food for thought ...

libtiledbsc/src/soma_collection.cc
```diff
@@ -0,0 +1,59 @@
...
+    for (uint64_t i = 0; i < group.member_count(); i++) {
+        auto member = group.member(i);
+        auto path = parent.empty() ?
+                        member.name().value() :
+                        std::string(parent) + "/" + member.name().value();
+
+        if (member.type() == Object::Type::Group) {
+            auto subgroup = Group(ctx_, member.uri(), TILEDB_READ);
+
+            // TODO: Check group metadata to see if subgroup is a SOMA or SOCO
```
johnkerl, line 45: for reference
https://github.com/single-cell-data/TileDB-SingleCell/pull/133

scripts/update-tiledb-version.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python3
```
johnkerl, line 1: just curious, why this file is here on this PR? seems like magic, but, old magic not new magic ...

johnkerl (approved):
🚢  with some nits :)

libtiledbsc/include/tiledbsc/soma.h
Shelnutt2: Further should this be an `unordered_map`?

libtiledbsc/include/tiledbsc/soma_collection.h
Shelnutt2: Further should this be an `unordered_map`?

libtiledbsc/src/cli/cli.cc
Shelnutt2: In the future..should we use spdlog here?

libtiledbsc/include/tiledbsc/soma.h
gspowley: I'll add more comments and update to `unordered_map`.

johnkerl: why not ordered map?

libtiledbsc/src/cli/cli.cc
gspowley: I noticed the missing LF too. I'll fix that. 

I worked on spdlog today, but the cmake for Windows failed, so saving it for a future PR.

libtiledbsc/include/tiledbsc/soma.h
gspowley: Unordered maps are more efficient because they don't sort the keys. It probably doesn't matter much in this case, since the maps should be small.

I made it ordered to begin with because I liked seeing sorted names when printing the map, but sorting can be done later.

libtiledbsc/src/soma_collection.cc
gspowley: I though about putting it in "utils", but saved that for another PR too.

scripts/update-tiledb-version.py
gspowley: I though this and `.clang-format` were useful, and I used them to generate the code in this PR, so I threw them in...

johnkerl: TIL 😁

PR #136, opened 2022-05-31
Webdocs experiment

johnkerl:
Please view this solely as a conversation-starter.

* Please first read the "Rationale" section: https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/web-docs-experiment/README-docs.md#rationale
* Link to rendered content: https://tiledb-singlecell-docs.s3.amazonaws.com/docs/overview.html 
* Setup notes: https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/web-docs-experiment/README-docs.md
  * Intentionally, no CI is set up yet for this experiment
* This PR is biggish as it introduces a framework for API+examples, while also introducing example/vignette content
  * The examples are overlong and could benefit from getting some bio-user eyes on them
* Please note all the Python API and examples content is provisional -- we can make this what we want it to be
* Note R API and examples are still hosted at https://github.com/TileDB-Inc/tiledbsc


aaronwolen (approved):
These look great! I can't speak to how/whether we can use this approach for the R docs but that's something we can discuss later. For now it's great to have a hosted website w/ docs for the python API.

johnkerl:
@aaronwolen thanks! Here's the deal about R docs:

* For now of course they're in the separate repo https://github.com/TileDB-Inc/tiledbsc and well-hosted at https://tiledb-inc.github.io/tiledbsc/
* R docs already do API+vignette marvelously
* GitHub Pages is fine with single-language stuff (R _or_ Python) but I haven't yet found a way to get GitHub Pages to handle multi-language content within a single repo -- which is why I went around that for this PR and chose S3 static hosting for the moment
* We of course (independent of R/Python) need to set up CI, and also decide if we want S3 static hosting or something else
* But to your question about R docs: I think we'll be best-served if we have CI render the R docs as now, and have this doctree point to that (and vice versa)

Basically letting the R-document framework do what it already does well right out of the box, and trying to get an almost-as-good setup for Python as well here, & having them each point to the other.

Issue #99, opened 2022-05-19
Queryability of dataframe attribute columns


johnkerl:
Parent context: #106.

Context:

* Users are going to want to have unicode values in their `obs`/`var` columns
* Unicode values are not currently queryable in the TileDB query-condition logic
  * Error message: `tiledb.cc.TileDBError: [TileDB::QueryCondition] Error: Clause non-empty attribute may only be var-sized for ASCII strings: cell_type`
* Workaround for the near term:
  * Force certain “key” dataframe columns (or maybe all columns) to be stored as ASCII
  * At write time, this works — "α,β,γ" stores as "\xce\xb1,\xce\xb2,\xce\xb3"
  * At read time: since SOMA is an API: utf8-decode those strings when a query is done & give the user back "α,β,γ"

johnkerl:
Nothing more to do here.

When we have true Unicode values on disk, that will be an internal change, and existing unit-test cases will ensure we don't have a regression.

PR #138, opened 2022-06-01
Extend ASCII-queryability workaround

johnkerl:
Context: issue #99. (See also: issue #106 and PRs #101 and #117.)

As pointed out by @bkmartinjr in a Slack conversation, the selection of string columns to store as ASCII on #101 was artificial, and likely to cause confusion. The better approach he suggested was to do this to _all_ string attributes.

This PR implements that better approach.

We will need this store-as-ASCII, decode-to-Unicode on read workaround for a couple months or so now pending some non-trivial work in TileDB core to fully support `QueryCondition` logic on Unicode attributes. As a reminder on the important of this workaround, string attributes are _non-queryable_ unless they are stored as ASCII, until full core-level support is in place; cc @Shelnutt2 and @ihnorton.

Note for @ebezzi and @bkmartinjr -- we will need a full re-ingest once this is merged and once https://github.com/TileDB-Inc/TileDB-Py/pull/1110 is merged and deployed in a `TileDB-Py` release which we include here in `tiledbsc-py`.

### DIFF
--- a/apis/python/src/tiledbsc/annotation_dataframe.py
+++ b/apis/python/src/tiledbsc/annotation_dataframe.py
@@ -85,6 +85,15 @@ class AnnotationDataFrame(TileDBArray):
         #   >>> A.meta.items()
         #   (('__pandas_index_dims', '{"obs_id": "<U0"}'),)
         # so the set_index is already done for us.
+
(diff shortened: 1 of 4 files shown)


aaronwolen (approved):
🚢

PR #140, opened 2022-06-01
Group-names readback

johnkerl:
At the moment we can't read back group-element names from the TileDB Group API. Previously I worked around this by taking basename of element URI and calling that "name" but this isn't a good heuristic in cases like this:

```
/Users/johnkerl/mini-corpus/tiledb-data-seurat/adipocytes-seurat/soma_RNA"
/Users/johnkerl/mini-corpus/tiledb-data-seurat/wilms-tumors-seurat/soma_RNA"
```

as these have the same basename.

On this PR is a better way to get the names these were added with.

On an upcoming TileDB-Py release we'll have direct TileDB-API support for this.

### DIFF
--- a/apis/python/src/tiledbsc/tiledb_group.py
+++ b/apis/python/src/tiledbsc/tiledb_group.py
@@ -152,4 +152,10 @@ class TileDBGroup(TileDBObject):
         member name to member URI.
         """
         with self._open("r") as G:
-            return {os.path.basename(e.uri): e.uri for e in G}
+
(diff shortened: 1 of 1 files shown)


aaronwolen (approved):
This works but as a heads-up: I am planning to stop adding the `soma_` prefix to group directory names. I previously used the prefix as a weak identifier for SOMAs/SCGroups. Now I just assume that all group members of a SOCO (other than `uns`) are SOMAs.

PR #143, opened 2022-06-01
Add spdlog to C++ API

gspowley:
Add `spdlog` support to the C++ API.

Example usage:
``` c++
#include <tiledbsc/logger_private.h>
...
    LOG_CONFIG("debug");
...
    LOG_INFO("walking soco URI = '{}'", uri);
...
```

### DIFF
--- a/libtiledbsc/cmake/Modules/FindSpdlog_EP.cmake
+++ b/libtiledbsc/cmake/Modules/FindSpdlog_EP.cmake
@@ -46,11 +46,17 @@ else()
   SET(SPDLOG_NO_DEFAULT_PATH)
 endif()
 
-find_package(spdlog
+# Fix issue on windows where spdlog::spdlog is already defined
+if (NOT TARGET spdlog::spdlog)
+  find_package(spdlog
         PATHS ${EP_INSTALL_PREFIX}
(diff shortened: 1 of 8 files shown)


libtiledbsc/cmake/Modules/FindSpdlog_EP.cmake
```diff
@@ -73,8 +79,11 @@ if (NOT SPDLOG_FOUND)
       PREFIX "externals"
       # Set download name to avoid collisions with only the version number in the filename
       DOWNLOAD_NAME ep_spdlog.zip
-      URL "https://github.com/gabime/spdlog/archive/v1.10.0.zip"
-      URL_HASH SHA1=aa2d4ff13b5393dea83d46caf545c6a303c889cd
+      URL "https://github.com/gabime/spdlog/archive/v1.9.0.zip"
```
ihnorton, line 82: Is the downgrade in order to match tiledb version?

ihnorton: Oh I see below - IIUC we need to change our own logger definition.

ihnorton (approved):

PR #144, opened 2022-06-02
Update to TileDB-Py 0.15.2

johnkerl:
Include some features from [TileDB-Py 0.15.2](https://github.com/TileDB-Inc/TileDB-Py/releases/tag/0.15.2):

* `"name" in G`
* with-open-as for groups

More to come on #115 and #116.


nguyenv (approved):
Awesome - looks good!

PR #152, opened 2022-06-07
Coalesce two dataframe methods into one

johnkerl:
We had two dataframe methods; there should be just one.

Moreover this makes for a nice API enhancement used by @ebezzi . Namely:

```
>>> df = soma.X['data'].dim_select(None,['AKR1C3','VDAC3'])

>>> csr = tiledbsc.util.X_and_ids_to_sparse_matrix(df, 'obs_id', 'var_id', 'value', soma.obs.ids(), soma.var.ids())

>>> csr.shape
(80, 20)

>>> csr.mean(axis=1).shape
(80, 1)
```


aaronwolen (approved):
:shipit:

PR #80, opened 2022-05-13
SOMA-collection sketching [WIP]

johnkerl:
I'm splitting this once-big, now-shrinking PR up into more easily reviewable pieces. Once all those pieces are created, reviewed, accepted, and merged, and as I rebase this PR atop them, eventually this PR will be zero-length and will be closed.

Context: #95

Issue #83, opened 2022-05-13
API: move to/from_ functions into "io" namespace


bkmartinjr:
Per May 11 meeting discussion: move the to/from functions (eg, to_anndata()) into a SOMA.io namespace.

PR #157, opened 2022-06-09
SOMA-collection-level slice/batch queries [needs revision]

johnkerl:
# Overview

* ~[cartographer](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/soma- slice/apis/python/examples/cartographer.py), due originally to @bkmartinjr , is a portable/reusable example of how to take a group of `.h5ad` files and create a `SOMACollection` having uniform `obs` schema~ split out to #161
* ~[splitter](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/soma-slice/apis/python/tools/splitter) is a little tool that was used to create the four small `.h5ad` files for unit-test/example-documentation purposes~ split out to #167
* [examples/soco-slice-query.md](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/soma-slice/apis/python/examples/soco-slice-query.md) is a walk-through example of using this code.
* [test_soco_slice_query.py](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/soma-slice/apis/python/tests/test_soco_slice_query.py) similar, as a unit-test case.
* [examples/soco-batch-query.md](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/soma-slice/apis/python/examples/soco-batch-query.md) is another walk-through example of using this code.
* The "meat" of the PR is in:
  * Mods to [soma.py](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/soma-slice/apis/python/src/tiledbsc/soma.py)
  * New [soma_slice.py](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/soma-slice/apis/python/src/tiledbsc/soma_slice.py)

# Walkthrough

* [soco-slice-query.md](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/soma-slice/apis/python/examples/soco-slice-query.md) shows you how to actually execute a slice query. This demonstrates the power of TileDB SOMAs to deliver targeted data for analysis at low latency. Notice that this examples writes the sliced output in both `SOMA` and `.h5ad` format, so you can go on to do all manner of analysis using for example [Scanpy](https://scanpy.readthedocs.io/en/stable/tutorials.html).
* [soco-batch-query.md](https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/soma-slice/apis/python/examples/soco-batch-query.md) shows you how to do a longer-running, cross-cutting batch query -- in this example, computing the mean of `X/data`, grouping by `obs['cell_type_ontology_term_id']`.

# Notes

* This replaces #80.
* The `SOMASlice` object currently has a (well-isolated) runtime dependency on `AnnData` for its `concat` -- namely, it converts a list of `SOMASlice` to a list of `AnnData`, leverages `AnnData`'s `concat`, then converts the concatenated `AnnData` object back to `SOMASlice`.


apis/python/examples/soma-slice-query.md
```diff
@@ -0,0 +1,200 @@
...
+import os
+import shutil
+
+soco_path = './soco-attribute-filter'
+if os.path.exists(soco_path):
+    shutil.rmtree(soco_path)
+
+soco = tiledbsc.SOMACollection(soco_path)
+if not soco.exists():
+    soco._create()
```
ebezzi, line 19: I would drop the underscore here, as this is not a private method (since it's being used externally). Also, a `create_if_not_exists()` method could be helpful.

apis/python/examples/soma-slice-query.md
```diff
@@ -0,0 +1,200 @@
...
+
+for name, h5ad in [
+    ('subset-soma-01', './anndata/subset-soma-01.h5ad'),
+    ('subset-soma-02', './anndata/subset-soma-02.h5ad'),
+    ('subset-soma-03', './anndata/subset-soma-03.h5ad'),
+    ('subset-soma-04', './anndata/subset-soma-04.h5ad'),
+]:
+    soma_path = os.path.join(soco_path, name)
+    soma = tiledbsc.SOMA(soma_path)
+    tiledbsc.io.from_h5ad(soma, h5ad)
```
ebezzi, line 29: I would argue that this is a non-pythonic way to call a function with such a name. Generally, I would expect that a `from_*` function returns the created object, instead of mutating an argument. I understand that the object is disk-based so it makes sense to pre-create it and pass it as an argument, but then I'd probably rename the function to something that makes the side effect more explicit (like `write_from_h5ad`). @bkmartinjr your opinion?

johnkerl: @ebezzi originally it was `soma.from_h5ad(h5ad)` but there was a desire to have a separate `tiledbsc.io` namespace (#83) in order to more clearly separate I/O (peripheral) from SOMA-API processing (central).

aaronwolen (commented):
Still working through it but I posted a couple of comments for now.

apis/python/src/tiledbsc/soma.py
```diff
@@ -174,3 +175,141 @@ def var_keys(self):
         An alias for `soma.var.ids()`.
         """
         return self.var.ids()
+
+    # ----------------------------------------------------------------
+    def dim_slice(self, slice_obs_ids, slice_var_ids) -> Dict:
+        """
+        Subselects the SOMA's obs, var, and X/data using the specified obs_ids and var_ids.
```
johnkerl, line 182: This needs to use
https://anndata.readthedocs.io/en/stable/generated/anndata.AnnData.layers.html
in order to preserve _all_ layers of `X`, e.g. `data` and `rankit`

johnkerl: done

apis/python/examples/soma-slice-query.md
bkmartinjr: I'm not sure if it is non-pythonic (I lack expertise to make that determination).  But I do find it awkward and likely error-prone as the two step process (create the soma, write the soma) allows for cases that will fail (eg, the soma is modified before `from_h5ad` is called).

I personally prefer a single function that both creates the soma and ingests the H5AD, eg,   `soma.io.from_h5ad(soma_path: str, h5ad_path: str, ...) -> soma`.   Ie, this function would call the SOMA constructor.

johnkerl: good to know @bkmartinjr -- thanks for the feedback! I'm fine with this; only wish I'd known sooner :)

johnkerl: @bkmartinjr @ebezzi I'll go ahead & make this change but on a separate PR -- this `soma.io` business is old & predates this PR, and this change can be neatly factored out

aaronwolen: Maybe it's worth posting this as a separate issue for discussion/feedback? The R API currently works this way too (ie, ingestion and SOMA instantiation are separate steps) and altering that behavior would require a lot of small updates to docs/tests/etc.

bkmartinjr:
> * `SOMASlice` class is at present a thin wrapper around `AnnData`. In fact we could do away with the former entirely and keep only the latter. Originally I wanted our own `SOMASlice`, but `AnnData`'s `concat` is non-trivial and very poweful -- we should be re-using this, not re-inventing it

This implies that SOMA is dependent on AnnData - is that something we actually want permanently included into the API?   (I understand that soma.io must be dependent, but I had always imagined that the core SOMA/SOCO did not require AnnData or Seurat).  Given the ongoing evolution of both external packages, I would be more comfortable if we wrap the dependency fully, and  review our public API capabilities closely with the view that (long-term) we may need to preserve the option for a non-AnnData implementation.   (to be clear, I'm fine using AnnData as the underlying engine, for now)

johnkerl:
> This implies that SOMA is dependent on AnnData - is that something we actually want permanently included into the API? (I understand that soma.io must be dependent, but I had always imagined that the core SOMA/SOCO did not require AnnData or Seurat). Given the ongoing evolution of both external packages, I would be more comfortable if we wrap the dependency fully, and review our public API capabilities closely with the view that (long-term) we may need to preserve the option for a non-AnnData implementation. (to be clear, I'm fine using AnnData as the underlying engine, for now)

@bkmartinjr agreed :^/

Totally agreed on 'for now' ...

... the main thing we would need to re-invent would be `concat` (which is non-trivial).

bkmartinjr:
> main thing we would need to re-invent would be `concat` (which is non-trivial)

I assume you only use a subset of the full AnnData.concat functionality, right?  Can that subset be formalized, to limit the SOMA API surface area?  (or perhaps, is it worth doing so?)

apis/python/tools/cartographer
```diff
@@ -0,0 +1,346 @@
+#!/usr/bin/env python
```
bkmartinjr, line 1: this file is very opinionated about the higher level schema (defined above the SOMA container API), and not really a SOMA tool.  Suggest that it be moved to the examples directory, and clearly documented for what it is.  There really is no such thing as a generic SOMA-level "atlas", as any atlas makes choices about how to integrate data.

if you would like a general purpose SOMA tool, perhaps a completely unopinionated SOCA creator that just add the entirety of a SOMA or H5AD to the SOCO?

johnkerl: > Suggest that it be moved to the examples directory, and clearly documented for what it is. 

awesome @bkmartinjr -- this is exactly the kind of feedback i was looking for.

> perhaps a completely unopinionated SOCA creator that just add the entirety of a SOMA or H5AD to the SOCO?

this will be a straightforward combine of the existing `tools/ingestor` and `tools/populate-soco` -- long-overdue for a merge :)

johnkerl:
> I assume you only use a subset of the full AnnData.concat functionality, right? Can that subset be formalized, to limit the SOMA API surface area? (or perhaps, is it worth doing so?)

Ooh yes @bkmartinjr I like this very much!!! :) I'll put up a commit to this effect ASAP. :)

apis/python/tools/cartographer
```diff
@@ -0,0 +1,346 @@
...
+    # ----------------------------------------------------------------
+    def _init_soco(self) -> SOMACollection:
+        """
+        Makes sure the destination SOMACollection exists for first write.
+        """
+        soco = SOMACollection(
+            self.atlas_uri, name="atlas", ctx=self.ctx, verbose=self.verbose
+        )
+        if not soco.exists():  # Must be done first, to create the parent directory
+            soco.create_unless_exists()
```
bkmartinjr, line 221: a more flexible API would be along the lines of os.mkdir, which allows:
* create if it doesn't exist, otherwise no error
* create if it does not exist, error out if it does
using the signature:    `os.mkdr(path, exists_ok: bool)`

So I'd suggest:   `soco.create(exists_ok: bool = False)`

johnkerl:
> > I assume you only use a subset of the full AnnData.concat functionality, right? Can that subset be formalized, to limit the SOMA API surface area? (or perhaps, is it worth doing so?)
> 
> Ooh yes @bkmartinjr I like this very much!!! :) I'll put up a commit to this effect ASAP. :)

https://github.com/single-cell-data/TileDB-SingleCell/pull/157/commits/f597661f884814efb8879b849c9a2763e25d61b2

@bkmartinjr I already like this much more :D

apis/python/tools/cartographer
johnkerl: thanks @bkmartinjr ! I'll do this as a separate PR

apis/python/src/tiledbsc/soma_collection.py
```diff
@@ -125,3 +134,124 @@ def cell_count(self) -> int:
         Returns sum of `soma.cell_count()` over SOMAs in the collection.
         """
         return sum(soma.cell_count() for soma in self)
+
+    # ----------------------------------------------------------------
+    def attribute_filter(
```
bkmartinjr, line 139: couple of issues:
* the requirement to redundantly provide the attribute names (ie, both in the query string and as a separate attr_name) seems unnecessary.  Other than the early error check, what is the purpose of `obs_attr_names` and `var_attr_names`
* this routine returns only obs, var and X - what about the other obs/var indexed sub-objects in the collection, such as raw, obsm, etc?

apis/python/src/tiledbsc/soma_collection.py
```diff
@@ -125,3 +134,124 @@ def cell_count(self) -> int:
...
+            # column in its var at all.
+            if not soma.var.has_attr_names(var_attr_names):
+                continue
+
+            soma_slice = soma.attribute_filter(obs_query_string, var_query_string)
+            if soma_slice != None:
+                # print("Slice SOMA from", soma.name, soma.X.data.shape(), "to", soma_slice.ann.X.shape)
+                soma_slices.append(soma_slice)
+
+        return SOMASlice.concat(soma_slices)
```
bkmartinjr, line 169: concat() makes assumptions about how the user wants to process the data.  I think it is preferred if you can simply return a list of slices.

that steps around issues such as duplicate IDs.

It might also make it easier to return partially evaluated queries which allow the user to later ask for sub-objects (eg, X, obsm, ..)

apis/python/src/tiledbsc/soma_collection.py
```diff
@@ -65,6 +67,13 @@ def remove(self, soma: SOMA) -> None:
...
 
+    # ----------------------------------------------------------------
+    def keys(self) -> None:
+        """
+        Returns the names of the SOMAs in the collection.
+        """
+        return self._get_member_names()
+
     # ----------------------------------------------------------------
     def __iter__(self) -> List[SOMA]:
```
bkmartinjr, line 78: __iter__ needs to pass the ctx down to the SOMA, or there is no way to use it with a custom TileDB Ctx object.

eg, line 84 should be:
> soma = SOMA(uri=uri, name=name, parent=self, ctx=self._ctx)

Without this, there is no way to open a SOCO with a custom Ctx, and have it ripple through the iterator.

apis/python/src/tiledbsc/soma_collection.py
```diff
@@ -1,11 +1,13 @@
...
 from .soma_options import SOMAOptions
 from .soma import SOMA
+from .soma_slice import SOMASlice
 from .tiledb_group import TileDBGroup
 
+import tiledb
+import pandas as pd
+
 
 class SOMACollection(TileDBGroup):
```
bkmartinjr, line 12: __init__ is failing to pass ctx to the super.__init__

bkmartinjr: example problem:  I have a SOCO with hundreds of large SOMAs in it.  I expected the following to work, but it does not:

```
soco = tiledbsc.SOMACollection("./atlas", ctx=tiledb.Ctx({"py.init_buffer_bytes": 4 * 1024 ** 3}))
sum(len(s.obs.ids()) for s in soco)
```

Because the buffer size is not passed down correctly, this iterator fails and there is no fix:
```
TileDBError                               Traceback (most recent call last)
<ipython-input-63-533fc342913f> in <module>
----> 1 sum(len(s.obs.ids()) for s in soco)

<ipython-input-63-533fc342913f> in <genexpr>(.0)
----> 1 sum(len(s.obs.ids()) for s in soco)

~/TileDB-SingleCell/apis/python/src/tiledbsc/annotation_dataframe.py in ids(self)
     74             # TileDB string dims are ASCII not UTF-8. Decode them so they readback
     75             # not like `b"AKR1C3"` but rather like `"AKR1C3"`.
---> 76             retval = A.query(attrs=[], dims=[self.dim_name])[:][self.dim_name].tolist()
     77             return [e.decode() for e in retval]
     78 

tiledb/libtiledb.pyx in tiledb.libtiledb.Query.__getitem__()

tiledb/libtiledb.pyx in tiledb.libtiledb.SparseArrayImpl.subarray()

tiledb/libtiledb.pyx in tiledb.libtiledb.SparseArrayImpl._read_sparse_subarray()

TileDBError: Exceeded maximum retries ('py.max_incomplete_retries': '100') (tiledb/core.cc:1221)

```

apis/python/src/tiledbsc/soma_slice.py
```diff
@@ -0,0 +1,124 @@
...
+                continue
+            # This works in Python -- not just a reference/pointer compare but a contents-compare :)
+            assert list(slicei.X.keys()) == list(slice0.X.keys())
+            assert list(slicei.obs.keys()) == list(slice0.obs.keys())
+            assert list(slicei.var.keys()) == list(slice0.var.keys())
+
+        # Use AnnData.concat.
+        # TODO: try to remove this dependency.
+        anns = [soma_slice.to_anndata() for soma_slice in soma_slices]
+        annc = ad.concat(anns)
```
bkmartinjr, line 107: this is an inner join on the second axis, which is not what you want....it will result in data being dropped.

It should read:  `ad.concat(anns, join="outter")`

apis/python/src/tiledbsc/soma_slice.py
```diff
@@ -0,0 +1,124 @@
...
+                continue
+            # This works in Python -- not just a reference/pointer compare but a contents-compare :)
+            assert list(slicei.X.keys()) == list(slice0.X.keys())
+            assert list(slicei.obs.keys()) == list(slice0.obs.keys())
+            assert list(slicei.var.keys()) == list(slice0.var.keys())
+
+        # Use AnnData.concat.
+        # TODO: try to remove this dependency.
+        anns = [soma_slice.to_anndata() for soma_slice in soma_slices]
+        annc = ad.concat(anns, join="outer")
```
bkmartinjr, line 107: this also fails to merge the secondary axis, so you end up with no `var` annotations.  Likely missing a parameter, such as `merge="first"`

apis/python/src/tiledbsc/__init__.py
```diff
@@ -1,6 +1,8 @@
 from .soma_collection import SOMACollection
 from .soma import SOMA
 from .soma_options import SOMAOptions
+from .soma_collection import SOMACollection
+from .soma_slice import SOMASlice
 
 from .tiledb_object import TileDBObject
```
bkmartinjr, line 7: should these be exported to the top-level namespace, or should they be internal-only?

bkmartinjr:
I have had a chance to write some code based upon this PR, which led to a couple of observations around attribute filters/queries:

1. There needs to be a means to indicate what data is materialized as a result of the query. Currently, obs, var and X['data'] are materialized, which is too much for some use cases, and too little for others. Much like the native tiledb `array.query()` allows the user to specify columns, the SOMA/SOCO query should let the user decouple the actual slicing from the result materialization, and control what is materialized in memory.
2. For large scale efforts, partial (cursor) queries are needed (eg, much like `return_incomplete=True`)
3. The result of a query is a SOMASlice, which can currently only be instantiated into an AnnData object.  More commonly I suspect people will want to get it back as piece parts, eg, only get the obs Pandas dataframe, or only get a scipy sparse version of the X slice. Have you considered making queries produce a view, and then having the view object provide individual accessors like SOMA provides (eg, `soco.attribute_filter(...).obs.df()` or some such)?

apis/python/src/tiledbsc/soma_slice.py
johnkerl: done on a commit by @bkmartinjr :)

apis/python/src/tiledbsc/soma_slice.py
johnkerl: done on a commit by @bkmartinjr :)

apis/python/src/tiledbsc/__init__.py
johnkerl: @bkmartinjr `SOMACollection` is 100% external -- I don't see how it couldn't be -- what am I missing?

`SOMASlice` -- kind of depends on what we suggest/allow people to do with slice results -- which is still a WIP here ...

johnkerl:
> I have had a chance to write some code based upon this PR, which led to a couple of observations around attribute filters/queries:
> 
> 1. There needs to be a means to indicate what data is materialized as a result of the query. Currently, obs, var and X['data'] are materialized, which is too much for some use cases, and too little for others. Much like the native tiledb `array.query()` allows the user to specify columns, the SOMA/SOCO query should let the user decouple the actual slicing from the result materialization, and control what is materialized in memory.
> 2. For large scale efforts, partial (cursor) queries are needed (eg, much like `return_incomplete=True`)
> 3. The result of a query is a SOMASlice, which can currently only be instantiated into an AnnData object.  More commonly I suspect people will want to get it back as piece parts, eg, only get the obs Pandas dataframe, or only get a scipy sparse version of the X slice. Have you considered making queries produce a view, and then having the view object provide individual accessors like SOMA provides (eg, `soco.attribute_filter(...).obs.df()` or some such)?

@bkmartinjr this is fantastic feedback & exactly the kind of thing I wanted to hear :D 

For #2, yes definitely -- and let's do it on a separate PR (this one is big enough)

johnkerl:
A series of rebases has gotten the commit-count on this PR out of control.

I'll revise this and re-submit.

johnkerl:
After rebases this PR is completely out of control (> 200 commits including mostly merge commits).

Closing, and will resubmit.

johnkerl:
re-opening long enough to squash

johnkerl:
https://github.com/single-cell-data/TileDB-SingleCell/pull/173 is the re-do

PR #160, opened 2022-06-14
Update to tiledb-py 0.15.4 [WIP]

johnkerl:
WIP

Note: `pip install --upgrade tiledb`


johnkerl:
Closing as there is a 0.15.5, but, that too doesn't do what we need to get `bool` support in `tiledb.from_pandas` context (`AnnotationDataFrame` as used by `obs` and `var`):

https://github.com/TileDB-Inc/TileDB-Py/blob/dev/tiledb/dataframe_.py#L125-L127

PR #186, opened 2022-06-23
Use logging.info for ingest/outgest prints

johnkerl:
Also manually verified `ingestor -q` and `outgestor -q` are silent.

### DIFF
--- a/apis/python/README-dev.md
+++ b/apis/python/README-dev.md
@@ -1,6 +1,8 @@
 * Most things are configured using GitHub Actions at `../../.github/workflows`
 * Pre-push suggestions:
   * `black .`
+  * `isort .`
+  * `flake8`
   * `python -m pytest tests`
 * PyPI:
   * https://pypi.org/project/tiledbsc/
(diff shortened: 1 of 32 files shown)


gsakkis (commented):
This is perfectly fine for scripts or more generally applications that have full control over logging. However it is not ideal for libraries that need to play well with others: the logging system if quite flexible so that the driver application can enable logging for library X1 at level L1 with handler(s) H1 and formatter F1, for library X2 at level L2 with handler(s) H2 and formatter F2, and so on.

To make it a well-behaved library, two things need to change:

1. Replace the `logging.info` calls with `logger.info`, where `logger` is a `logging.Logger` instance. A common practice is to create one logger for each module (that needs to log anything):
```
import logging

logger = logging.getLogger(__name__)
```

2. Remove the `logging.basicConfig` (or any kind of logging configuration for that matter) from the library. Configuration is the responsibility of the driver application, not an individual library. In this case this implies that the `verbose` flag from `SOMA` and `SOMACollection` should go.
Having said that, the library _may_ configure logging in its entry points (e.g. CLI scripts), in this case for `tools` and `examples`.

More about Python logging:
- https://docs.python-guide.org/writing/logging/#logging-in-a-library
- https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library

Shelnutt2 (approved):

PR #193, opened 2022-06-24
Pre-commit hooks

gsakkis:
Adds [pre-commit](https://pre-commit.com/) hooks. 

This concerns project devs and is completely optional. To enable it, install it by running in the project root:
```pip install pre-commit && pre-commit install``` 
Then on every git commit it it runs all configured linters locally on the changed files and prevents the commit if there is any error.

Also lints two Python scripts that were not caught by CI because they are not under `apis/python`.


johnkerl (approved):
Awesome, thanks @gsakkis !!! :D

PR #196, opened 2022-06-25
Neaten logger-init code

johnkerl:
Following great advice from @gsakkis 

To check:

* `tools/outgestor`
* `tools/outgestor -q`
* Make sure the first one prints progress updates and the latter does not


aaronwolen (approved):
🚀

PR #197, opened 2022-06-26
Initial type checking with mypy on CI and fixes

gsakkis:
Add [mypy](http://mypy-lang.org/) CI integration. Python type annotations are no more than (typically incomplete or sometimes just plain wrong) documentation without a type checker.

On https://github.com/single-cell-data/TileDB-SingleCell/commit/689b0e28964a28f97c7c4128cc93b4e92a8483b9:
```
$ mypy .
...
Found 55 errors in 15 files (checked 43 source files)
```

On https://github.com/single-cell-data/TileDB-SingleCell/commit/973847d8331750afed87cf5d8033ead9e95267f2:
```
$ mypy .
src/tiledbsc/raw_group.py:53: error: Argument "parent" to "AssayMatrixGroup" has incompatible type "AnnotationDataFrame"; expected "Optional[TileDBGroup]"  [arg-type]
Found 1 error in 1 file (checked 43 source files)
```
This might be a legtimate bug, I wasn't sure of the correct fix so I left it out.

**Note**: These are almost the bare minimum checks and fixes; mypy has a `--strict` flag that does several stricter checks. As of this PR, `mypy --strict` yields an additional 200 errors under `apis/python/src` alone, (ideally) to be addressed in future PR(s).


johnkerl:
> This might be a legtimate bug, I wasn't sure of the correct fix so I left it out.

It is, bug venial -- the 'back link' from nodes to parents is used for (a) propagating `soma_options` etc, which are all the same, and (b) for indentation levels in verbose ingest.

I'll fix this.

Also @gsakkis maybe I can put up some mypy-fixes as additional commits on this PR -- ?

johnkerl:
@gsakkis the force-push just now was my rebasing onto current `main`

johnkerl:
@gsakkis with my last commit this is clean in CI.

I think we should merge this as-is, then address `mypy --strict .` on one or more subsequent PRs ...

gsakkis:
@johnkerl I pushed one more [small fix](https://github.com/single-cell-data/TileDB-SingleCell/pull/197/commits/eecd889ae7da49d3791f7d791764694ea1432a54); the logger name used to be "tiledbsc.logging" instead of "tiledbsc".

> I think we should merge this as-is, then address mypy --strict . on one or more subsequent PRs ...

Agreed :+1:

PR #201, opened 2022-06-27
Work around issue uploading 1D uns to tiledb URIs

johnkerl:
To be revised following advice from @ihnorton


johnkerl:
out of date

PR #199, opened 2022-06-26
Disallow duplicates

johnkerl:
Following R impl at https://github.com/TileDB-Inc/tiledbsc/pull/69

PR #208, opened 2022-06-30
Optimize queries for allow_duplicates=False

johnkerl:
Following on #199.

# Summary

Setting `allow_duplicates=False` is the path of least surprise for users. This ensures that, if a matrix is ever re-written/updated, when people query, expecting a matrix of numbers, they'll get that -- vs a matrix of lists where each cell contains historical values.

However, this comes at some read-performance cost. The get-all-the-data timing queries done on #199 were insufficient; better queries are shown in the details below.

Using TileDB Core 2.10, which is released, we see about a 2x slowdown going to the 'safe' mode `allow_duplicates=False'. Using dev version of core with mods to be released in TileDB Core 2.11 (in a couple weeks) we see the slowdown more like 1.2x.

# Timing results

We took the same 59K cell (393M total) dataset, stored in four ways:

* `dupes-noconsol` -- `allow_duplicates=True`, `X/data` not consolidated (in a half-dozen fragments)
* `dupes-consol` -- a copy of that, but with `X/data` consolidated into a single fragment.
* `nodupes-noconsol` -- `allow_duplicates=False`, `X/data` not consolidated (in a half-dozen fragments)
* `nodupes-consol` -- a copy of that, but with `X/data` consolidated into a single fragment.

The cell-types available in that dataset are as follows:

```
>>> soma.obs.df().groupby('cell_type').size().sort_values()
cell_type
plasmacytoid dendritic cell          575
plasmablast                          586
platelet                            1007
dendritic cell                      1038
alpha-beta T cell                   1659
natural killer cell                 3248
B cell                              6131
CD4-positive, alpha-beta T cell     6726
CD8-positive, alpha-beta T cell     8658
monocyte                           29878
dtype: int64
```

Running the below query code on all four, we get the following timings using core 2.11:

| nobs | dupes-noconsol_seconds | dupes-consol_seconds | nodupes-noconsol_seconds | nodupes-consol_seconds |
| --- | --- | --- | --- | --- |
| 575 | 1.215 | 1.201 | 1.228 | 1.243 |
| 586 | 1.186 | 1.198 | 1.174 | 1.197 |
| 1007 | 1.678 | 1.763 | 1.708 | 1.786 |
| 1038 | 2.321 | 2.431 | 2.160 | 2.218 |
| 1659 | 2.252 | 2.823 | 2.558 | 2.598 |
| 3248 | 4.046 | 3.444 | 3.849 | 3.587 |
| 6131 | 4.982 | 4.561 | 5.101 | 4.720 |
| 6726 | 4.670 | 4.820 | 4.955 | 4.504 |
| 8658 | 5.284 | 5.743 | 6.253 | 6.049 |
| 29878 | 17.523 | 15.439 | 18.501 | 16.221 |

# Query code

<details>

```
#!/usr/bin/env python

import tiledb
import tiledbsc
import sys
import time

ctx = tiledb.Ctx({"py.init_buffer_bytes": 8 * 1024**3})

# Public bucket:
# s3://tiledb-singlecell-data/profile-data/dupes-consol
# s3://tiledb-singlecell-data/profile-data/dupes-noconsol
# s3://tiledb-singlecell-data/profile-data/nodupes-consol
# s3://tiledb-singlecell-data/profile-data/dupes-noconsol
# uri = 's3://tiledb-singlecell-data/profile-data/nodupes-consol'
# uri = 's3://tiledb-singlecell-data/profile-data/nodupes-consol'
uri = '/tmp/nodupes-consol'
if len(sys.argv) >= 2:
    uri = sys.argv[1]
soma = tiledbsc.SOMA(uri, ctx=ctx)

# Choices for cell_type:
#
# >>> soma.obs.df().groupby('cell_type').size().sort_values()
# cell_type
# plasmacytoid dendritic cell          575
# plasmablast                          586
# platelet                            1007
# dendritic cell                      1038
# alpha-beta T cell                   1659
# natural killer cell                 3248
# B cell                              6131
# CD4-positive, alpha-beta T cell     6726
# CD8-positive, alpha-beta T cell     8658
# monocyte                           29878
# dtype: int64

which_attr = 'cell_type'
which_value = 'B cell'
if len(sys.argv) == 3:
    which_value = sys.argv[2]
if len(sys.argv) == 4:
    which_attr  = sys.argv[2]
    which_value = sys.argv[3]
query_string = f'{which_attr} == "{which_value}"'

print('URI:', uri)
print('query_string:', query_string)

o1 = time.time()
slice = soma.query(
    obs_query_string=query_string,
)
o2 = time.time()
print(f"got shapes obs:{slice.obs.shape} var:{slice.var.shape} X/data:{slice.X['data'].shape}")
fsec = "%.3f" % (o2 - o1)
print(f"SECONDS_TOTAL={fsec} NOBS={slice.obs.shape[0]} QUERY='{query_string}' URI={uri}")
```

</details>

### DIFF
--- a/apis/python/src/tiledbsc/soma.py
+++ b/apis/python/src/tiledbsc/soma.py
@@ -64,6 +64,20 @@ class SOMA(TileDBGroup):
         :param uri: URI of the TileDB group
         """
 
+        # People can (and should) call by name. However, it's easy to forget. For example,
+        # if someone does 'tiledbsc.SOMA("myuri", ctx)' instead of 'tiledbsc.SOMA("myury", ctx)',
+        # behavior will not be what they expect, and we should let them know sooner than later.
+        if name is not None:
+            assert isinstance(name, str)
+        if soma_options is not None:
+            assert isinstance(soma_options, SOMAOptions)
+        if config is not None:
+            assert isinstance(config, tiledb.Config)
+        if ctx is not None:
+            assert isinstance(ctx, tiledb.Ctx)
+        if parent is not None:
+            assert isinstance(parent, TileDBGroup)
+
         if ctx is None and config is not None:
             ctx = tiledb.Ctx(config)
         if soma_options is None:
@@ -315,9 +329,18 @@ class SOMA(TileDBGroup):
         # obs, but no rows with cell_type == "blood".
         if slice_obs_df is None:
             return None
-        obs_ids = list(slice_obs_df.index)
-        if len(obs_ids) == 0:
+        if len(slice_obs_df.index) == 0:
             return None
+        # At the tiledb multi-index level, if we're say slicing on obs_ids but not var_ids,
+        # we'll do `A.df[obs_ids, :]`. We can't pass a `:` down the callstack to get there,
+        # but we pass `None` instead.
+        #
+        # It's important to do this. Say for example the X matrix is nobs=1000 by nvar=2000,
+        # and we have a query that has 158 obs_ids. At the tiledb multi-index level, doing
+        # `A.df[{158 obs ids}, {all 2000 var ids}]` is non-performant while
+        # `A.df[{158 obs ids}, :]` is performant.
+        if obs_ids is not None or obs_query_string is not None:
+            obs_ids = list(slice_obs_df.index)
 
         slice_var_df = self.var.query(
             query_string=var_query_string, ids=var_ids, attrs=var_attrs
@@ -326,9 +349,11 @@ class SOMA(TileDBGroup):
         # in its var, but no rows with feature_name == "MT-CO3".
         if slice_var_df is None:
             return None
-        var_ids = list(slice_var_df.index)
-        if len(var_ids) == 0:
+        if len(slice_var_df.index) == 0:
             return None
+        # See above comment re keeping obs_ids == None if that's what it came in as.
+        if var_ids is not None or var_query_string is not None:
+            var_ids = list(slice_var_df.index)
 
         # TODO:
         # do this here:
--- a/apis/python/src/tiledbsc/soma_collection.py
+++ b/apis/python/src/tiledbsc/soma_collection.py
@@ -30,6 +30,20 @@ class SOMACollection(TileDBGroup):
         :param uri: URI of the TileDB group
         """
 
+        # People can (and should) call by name. However, it's easy to forget. For example,
+        # if someone does 'tiledbsc.SOMACollection("myuri", ctx)' instead of 'tiledbsc.SOMA("myury", ctx)',
+        # behavior will not be what they expect, and we should let them know sooner than later.
+        if name is not None:
+            assert isinstance(name, str)
+        if soma_options is not None:
+            assert isinstance(soma_options, SOMAOptions)
+        if config is not None:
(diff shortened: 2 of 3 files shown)


apis/python/src/tiledbsc/soma.py
```diff
@@ -64,6 +64,20 @@ def __init__(
         :param uri: URI of the TileDB group
         """
 
+        # People can (and should) call by name. However, it's easy to forget. For example,
```
johnkerl, line 67: Found during debug. `soma = tiledbsc.SOMA(uri, ctx)` is incorrect while `soma = tiledbsc.SOMA(uri, ctx=ctx)` is correct. This needs to be surfaced proactively to the user.

apis/python/src/tiledbsc/soma.py
```diff
@@ -315,9 +329,18 @@ def query(
...
             return None
-        obs_ids = list(slice_obs_df.index)
-        if len(obs_ids) == 0:
+        if len(slice_obs_df.index) == 0:
             return None
+        # At the tiledb multi-index level, if we're say slicing on obs_ids but not var_ids,
+        # we'll do `A.df[obs_ids, :]`. We can't pass a `:` down the callstack to get there,
+        # but we pass `None` instead.
+        #
+        # It's important to do this. Say for example the X matrix is nobs=1000 by nvar=2000,
```
johnkerl, line 338: We were doing the following:

* For example a 1000x2000 matrix
* 80 obs IDs -> 80 obs IDs
* `None` (for `:`) -> 2000 var IDs

is not new. However, it performed OK with `allow_duplicates=True` but performed poorly with `allow_duplicates=False`.

apis/python/src/tiledbsc/soma_collection.py
```diff
@@ -30,6 +30,20 @@ def __init__(
         :param uri: URI of the TileDB group
         """
 
+        # People can (and should) call by name. However, it's easy to forget. For example,
+        # if someone does 'tiledbsc.SOMACollection("myuri", ctx)' instead of 'tiledbsc.SOMA("myury", ctx)',
+        # behavior will not be what they expect, and we should let them know sooner than later.
+        if name is not None:
```
johnkerl, line 36: Arg-checking as above, for `SOMACollection` as well as `SOMA`.

johnkerl:
@gsakkis with my last commit this is clean in CI.

I think we should merge this as-is, then address `mypy --strict .` on one or more subsequent PRs ...

ihnorton (approved):

apis/python/src/tiledbsc/soma.py
gsakkis: You may want to enforce [keyword-only arguments](https://peps.python.org/pep-3102/) to avoid such incorrect usage. 

Manual runtime type checking is considered unidiomatic in general and doesn't scale well with the complexity of annotations (think of `Dict[str, List[Optional[int]]`). For the cases that really need it, there are 3rd party libraries such as [Typeguard](https://typeguard.readthedocs.io/en/latest/userguide.html) or [Pydantic](https://pydantic-docs.helpmanual.io/) that use and enforce annotations at runtime transparently.

PR #210, opened 2022-07-01
Support group member deletion

johnkerl:
Python analog of https://github.com/TileDB-Inc/tiledbsc/pull/71

### DIFF
--- a/apis/python/src/tiledbsc/annotation_matrix_group.py
+++ b/apis/python/src/tiledbsc/annotation_matrix_group.py
@@ -105,6 +105,20 @@ class AnnotationMatrixGroup(TileDBGroup):
         self._add_object(annotation_matrix)
 
     # ----------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


aaronwolen (approved):

gsakkis:
Although there is a `__delete__` magic method for [descriptors](https://docs.python.org/3/howto/descriptor.html#descriptor-protocol), that's probably not what you mean here. Since there are no tests for the intended usage as there are for `remove`, I guess the intention is to override [`__delitem__`](https://docs.python.org/3/reference/datamodel.html#object.__delitem__) (and perhaps [`__delattr__`](https://docs.python.org/3/reference/datamodel.html#object.__delattr__) as well since you have overriden `__getattr__` to be essentially an alias to `__getitem__`).

johnkerl:
@gsakkis thanks for your kind and expert advice! I'll take a careful look at all these.

PR #211, opened 2022-07-01
Support quiet/info/debug levels for data ingest

johnkerl:
Logging ingests is crucial -- while some intro datasets take a few seconds to upload, gigascale data takes more time. During this time it's essential to give the user something to see, some indication that something is happen.

Before this PR, the log levels were silent, or verbose. During notebook/doc work this week I found the latter was taking up more than a full screen in notebook views -- which is overkill.

On this PR we strike a middle ground: `info` level gives a brief summary, with something every so often for the time-consuming bits (if any), but minimalistic in character-count. The old verbosity is still available using `tiledbsc.logging.debug()`.

Before-and-afters using `tiledbsc.logging.info()` in a notebook:

Screen Shot 2022-07-01 at 4 27 10 PM

### DIFF
--- a/apis/python/src/tiledbsc/annotation_dataframe.py
+++ b/apis/python/src/tiledbsc/annotation_dataframe.py
@@ -1,3 +1,4 @@
+import os
 from typing import List, Optional, Set, Tuple
 
 import numpy as np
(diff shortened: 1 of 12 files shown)


aaronwolen (approved):
Very nice!

PR #212, opened 2022-07-01
Allow skip ingest if exists

johnkerl:
During notebook/doc work this week (see also the [public soma-testing notebook](https://cloud.tiledb.com/notebooks/details/johnkerl-tiledb/d3d7ff44-dc65-4cd9-b574-98312c4cbdbd/preview), I found myself skipping ingest for alread-uploaded data.

I suspect our users will be doing a lot of iffing unless we help them out with a way to express this common desire in a single line.

### DIFF
--- a/apis/python/src/tiledbsc/io.py
+++ b/apis/python/src/tiledbsc/io.py
@@ -3,10 +3,24 @@ import scanpy
 import tiledb
 
 import tiledbsc
+import tiledbsc.logging
 import tiledbsc.util
 import tiledbsc.util_ann
 
 from .logging import logger
+
+
+# ----------------------------------------------------------------
+def from_h5ad_unless_exists(soma: tiledbsc.SOMA, input_path: str) -> None:
+    """
+    Skips the ingest if the SOMA is already there. A convenient keystroke-saver
+    so users don't need to replicate the if-test.
+    """
+    if tiledbsc.util.is_soma(soma.uri):
+        tiledbsc.logging.logger.info(f"Already exists, skipping ingest: {soma.uri}")
+        return
+    else:
+        from_h5ad(soma, input_path)
 
 
 # ----------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


aaronwolen (approved):
I think I would personally leave this out of the API and let the users add their own `if` tests, but that's just a personal preference to keep the number of methods smaller/focused.

gspowley:
What about an `overwrite=false` option to `from_h5ad` and `from_10x`and handling it in the existing functions?

johnkerl:
> I think I would personally leave this out of the API and let the users add their own if tests, but that's just a personal preference to keep the number of methods smaller/focused.

@aaronwolen I thought the same, and also looked at every single uploader cell in my notebook having if-tests ... and imagined the replication ... I think this one is worthwhile. :)

johnkerl:
> What about an overwrite=false option to from_h5ad and from_10xand handling it in the existing functions?

@gspowley just saw this, after I merged on @aaronwolen 's accept --

* There are actually a few ways for if-exists: do nothing; spend the time to overwrite the data (maybe with new/updated content); remove and replace.
* I want to distinguish between the first & second especially
* The second doesn't exist _yet_ due to open `TileDB-Py PR https://github.com/TileDB-Inc/TileDB-Py/pull/1185 not yet merged/released -- this affects `uns` data and is the only remaining reason users can't do option 2 yet

johnkerl:
@gspowley once https://github.com/TileDB-Inc/TileDB-Py/pull/1185 is in place (soon! -- and we're still on 0.0.x revs with room to iterate) I think the best option will be:

* `tiledbsc.io.from_h5ad(soma, path, overwrite=False)`
* If false: do what `from_h5ad_unless_exists()` is doing now
* If true: re-write the new data over top of the old

aaronwolen:
> I thought the same, and also looked at every single uploader cell in my notebook having if-tests ... and imagined the replication ... I think this one is worthwhile. :)

Another option would for the particular demo use case would be to have the notebook optionally cleanup after itself. I do something similar in the R vignettes for local testing: https://github.com/TileDB-Inc/tiledbsc/blob/bc1b25262765f14e074c75794cdf928563ab4961/vignettes/introduction.Rmd#L185-L187.

johnkerl:
@aaronwolen what about tiledb-cloud notebooks?

johnkerl:
@aaronwolen nice! I need to think about this for various backends ...

PR #216, opened 2022-07-06
Strict static typing

gsakkis:
Enable `mypy --strict` and fix remaining typing issues.


johnkerl (approved):
Thanks @gsakkis !!!! :D

PR #217, opened 2022-07-06
Add SOMAQuery to C++ API

gspowley:
This PR adds the `SOMAQuery` class to the C++ API. `SOMAQuery` implements an `X/data` query with the following features:
* Set query conditions on `obs` and `var`
* Select columns of `obs` and `var` to materialize
* Limit the query to specific `obs_id`s and `var_id`s
* Incomplete queries on `obs` and `var` are handled by `SOMAQuery` internally
* Incomplete queries on `X/data` are handled by calling `SOMAQuery::next_results()` until it returns no results

This PR also adds:
* `ThreadPool` class from the TileDB core library
* WIP support for pybind11

Example `SOMAQuery` usage:
```c++
// Open a SOMA and create a SOMAQuery
auto soma = SOMA::open(soma_uri);
auto sq = soma->query();

// Build an obs query condition: cell_type == "B cell"
std::string obs_attr = "cell_type";
std::string obs_val = "B cell";
auto ctx = soma->context();
auto obs_qc = QueryCondition::create(*ctx, obs_attr, obs_val, TILEDB_EQ);

// Set the obs query condition
sq->set_obs_condition(obs_qc);

// Select obs and var columns to materialize
std::vector<std::string> obs_cols = {obs_attr};
sq->select_obs_attrs(obs_cols);
std::vector<std::string> var_cols = {"var_id"};
sq->select_var_attrs(var_cols);

// Submit the query and process the results
while (auto results = sq->next_results()) {
    // process results
}
```

### DIFF
--- a/libtiledbsc/cmake/Modules/FindSpdlog_EP.cmake
+++ b/libtiledbsc/cmake/Modules/FindSpdlog_EP.cmake
@@ -39,7 +39,7 @@ include(TileDBCommon)
 set(CMAKE_PREFIX_PATH ${CMAKE_PREFIX_PATH} "${EP_INSTALL_PREFIX}")
 
 # First try the CMake find module.
-if (TILEDB_SPDLOG_EP_BUILT)
+if (TILEDBSC_SPDLOG_EP_BUILT)
   # If we built it from the source always force no default path
   SET(SPDLOG_NO_DEFAULT_PATH NO_DEFAULT_PATH)
 else()
@@ -101,7 +101,7 @@ if (NOT SPDLOG_FOUND)
       DEPENDS ${SPDLOG_DEPENDS})
     list(APPEND EXTERNAL_PROJECTS ep_spdlog)
     list(APPEND FORWARD_EP_CMAKE_ARGS
-            -DTILEDB_SPDLOG_EP_BUILT=TRUE
+            -DTILEDBSC_SPDLOG_EP_BUILT=TRUE
             )
   else()
     message(FATAL_ERROR "Unable to find spdlog")
@@ -128,10 +128,3 @@ elseif(TARGET spdlog::spdlog)
     endif()
   endif()
 endif()
-
-# If we built a static EP, install it if required.
-# TODO: Fix this block if needed, it breaks CMAKE_BUILD_TYPE=Coverage
-#if (TILEDB_SPDLOG_EP_BUILT AND TARGET spdlog::spdlog)
-#  include(TileDBCommon)
-#  install_target_libs(spdlog::spdlog)
-#endif()
--- a/libtiledbsc/src/cli.cc
+++ b/libtiledbsc/src/cli.cc
@@ -112,18 +112,7 @@ int main(int argc, char** argv) {
 
     try {
         // walk_soco(argv[1]);
-    } catch (const std::exception& e) {
-        LOG_FATAL(fmt::format(
-            "URI '{}' is not a SOMACollection. {}", argv[1], e.what()));
-    }
-
-    try {
         // slice_soma(argv[2]);
-    } catch (const std::exception& e) {
-        LOG_FATAL(e.what());
-    }
-
-    try {
         soma_query(argv[2]);
     } catch (const std::exception& e) {
         LOG_FATAL(e.what());
--- a/libtiledbsc/test/CMakeLists.txt
+++ b/libtiledbsc/test/CMakeLists.txt
@@ -24,7 +24,7 @@ if (TILEDBSC_TESTING)
         unit_column_buffer.cc
         unit_managed_query.cc
         unit_soma.cc
-        # unit_thread_pool.cc
+        unit_thread_pool.cc
     )
 
     target_link_libraries(unit_sc
--- a/libtiledbsc/CMakeLists.txt
+++ b/libtiledbsc/CMakeLists.txt
@@ -183,12 +183,6 @@ endif()
 # Definitions for all targets
 add_definitions(-D_FILE_OFFSET_BITS=64)
 
-# Disable deprecation warnings
-if (NOT MSVC)
-  message(WARNING "Building without deprecation warnings")
-  add_compile_options(-Wno-deprecated-declarations)
-endif()
-
 # AVX2 flag
 include(CheckAVX2Support)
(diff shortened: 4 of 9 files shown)


libtiledbsc/cmake/Modules/FindSpdlog_EP.cmake
```diff
@@ -130,7 +130,8 @@ elseif(TARGET spdlog::spdlog)
 endif()
 
 # If we built a static EP, install it if required.
-if (TILEDB_SPDLOG_EP_BUILT AND TARGET spdlog::spdlog)
-  include(TileDBCommon)
-  install_target_libs(spdlog::spdlog)
-endif()
+# TODO: Fix this block if needed, it breaks CMAKE_BUILD_TYPE=Coverage
+#if (TILEDB_SPDLOG_EP_BUILT AND TARGET spdlog::spdlog)
```
Shelnutt2, line 134: Do we want this commented out?

libtiledbsc/include/tiledbsc/column_buffer.h
```diff
@@ -3,9 +3,15 @@
 
 #include <stdexcept>  // for windows: error C2039: 'runtime_error': is not a member of 'std'
 
+// TODO: pybind11 used for intermediate testing, remove after adding arrow
```
Shelnutt2, line 6: Do we want pybind11 in these functions? In TileDB-VCF pybind11 is only used in the python API files?

libtiledbsc/src/cli.cc
```diff
@@ -23,24 +23,83 @@ void walk_soco(std::string_view uri) {
 
 void slice_soma(std::string_view soma_uri) {
     Config conf;
-    // conf["config.logging_level"] = "5";
+    conf["config.logging_level"] = "5";
```
Shelnutt2, line 26: Should this be a user set setting?

libtiledbsc/src/cli.cc
```diff
@@ -23,24 +23,83 @@ void walk_soco(std::string_view uri) {
...
+    auto obs_qc = QueryCondition::create(*ctx, obs_attr, obs_val, TILEDB_EQ);
+    std::vector<std::string> obs_cols = {obs_attr};
+    sq->set_obs_condition(obs_qc);
+    sq->select_obs_attrs(obs_cols);
+
+    uint64_t var_val = 50;
+    auto var_qc = QueryCondition::create<uint64_t>(
+        *ctx, "n_cells", var_val, TILEDB_LT);
+    std::vector<std::string> var_cols = {"var_id"};
+    //    sq->set_var_condition(var_qc);
```
Shelnutt2, line 78: Should this be commented out?

libtiledbsc/src/cli.cc
```diff
@@ -23,24 +23,83 @@ void walk_soco(std::string_view uri) {
...
+    sq->select_var_attrs(var_cols);
+
+    std::vector<std::string> obs_ids = {
```
Shelnutt2, line 81: Should this be hard coded?

libtiledbsc/src/cli.cc
```diff
@@ -52,14 +111,20 @@ int main(int argc, char** argv) {
     LOG_CONFIG("debug");
 
     try {
-        walk_soco(argv[1]);
+        // walk_soco(argv[1]);
```
Shelnutt2, line 114: Empty try?

libtiledbsc/src/cli.cc
```diff
@@ -52,14 +111,20 @@ int main(int argc, char** argv) {
...
     } catch (const std::exception& e) {
         LOG_FATAL(fmt::format(
             "URI '{}' is not a SOMACollection. {}", argv[1], e.what()));
     }
 
     try {
-        slice_soma(argv[2]);
+        // slice_soma(argv[2]);
```
Shelnutt2, line 121: Empty try?

libtiledbsc/test/CMakeLists.txt
```diff
@@ -23,6 +23,8 @@ if (TILEDBSC_TESTING)
         $<TARGET_OBJECTS:TILEDB_SC_OBJECTS>
         unit_column_buffer.cc
         unit_managed_query.cc
+        unit_soma.cc
+        # unit_thread_pool.cc
```
Shelnutt2, line 27: Should this be commented out?

libtiledbsc/cmake/Modules/FindSpdlog_EP.cmake
gspowley: Ideally, we don't want this commented out. It causes a failure when building a `Coverage` build though. I can look into fixing it now.

libtiledbsc/include/tiledbsc/column_buffer.h
gspowley: Right, we don't want pybind11 here. It was useful for initial experiments/tests and will be removed with the PR that adds arrow support.

libtiledbsc/src/cli.cc
gspowley: `cli.cc` is sandbox...

libtiledbsc/src/cli.cc
gspowley: `cli.cc` is sandbox...

libtiledbsc/src/cli.cc
gspowley: `cli.cc` is sandbox...

libtiledbsc/src/cli.cc
gspowley: `cli.cc` is sandbox...

libtiledbsc/test/CMakeLists.txt
gspowley: I'll add that test back.

libtiledbsc/cmake/Modules/FindSpdlog_EP.cmake
gspowley: This block of cmake code came from the TileDB core library. In the core, it's used to conditionally install a static version of spdlog when building a static version of tiledb. Since we don't have the option of building a static version of tiledbsc, I removed this block of code.

PR #227, opened 2022-07-23
SOMA API v1 experiments [WIP/RFC]

johnkerl:
This is some very, very rough initial work to get a first pass at looking for implementation-level gotchas with the [nascent v1 SOMA-API spec](https://github.com/single-cell-data/SOMA/blob/spec-revision/brainstorming.md).

* Basic walk-throughs:
  * See `tests/test_v1_*` at https://github.com/single-cell-data/TileDB-SingleCell/tree/kerl/v1-api-spec/apis/python/tests
    * Primarily https://github.com/single-cell-data/TileDB-SingleCell/blob/kerl/v1-api-spec/apis/python/tests/test_v1_soma_experiment_basic.py
  * See also the notebook https://cloud.tiledb.com/notebooks/details/johnkerl-tiledb/030375f5-2002-4d97-9b2a-0db0e810622b/preview
* In-package code status:
  * Basic `obs`, `var`, `X` have ingest-from-arrow working
  * Not all `read` methods exist yet; all of `ReadResult` is not implemented yet
  * `get_schema` (TileDB schema to Arrow schema) is not implemented yet
  * `delete` is not implemented yet
* Status with dependencies:
  * Indexing on string columns currently fails due to Unicode-vs-ASCII issues which were [dealt with in v0](https://github.com/single-cell-data/TileDB-SingleCell/issues/99) but are not yet dealt with in v1 
  * When `SOMADataFrame` is stored as dense (not the case at the moment, because of these things):
    * `exp.obs.read(ids=[0,2,4])` (discontiguous) SEGVs in core
    * This is done using the tiledb `df` accessor
    * We really want the `multi_index` accessor but we get the runtime error `raise TileDBError("`return_arrow=True` requires .df indexer`")` (@ihnorton FYI)
* Much more to do. Just a shared progress-point before I take a week of PTO.


johnkerl:
Superseded by #282

PR #239, opened 2022-08-12
Parallelize soma collection queries

Shelnutt2:
Parallelize soma collection queries via threading. This works as a short term item for performance improvements. We have a number of additional changes in flight for Soma collections that will yield better and longer term improvements.

### DIFF
--- a/apis/python/src/tiledbsc/soma_collection.py
+++ b/apis/python/src/tiledbsc/soma_collection.py
@@ -1,4 +1,5 @@
 from typing import Dict, Iterator, Optional, Sequence, Set, Union
+from concurrent.futures import ThreadPoolExecutor
 
 import tiledb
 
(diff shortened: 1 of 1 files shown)


apis/python/src/tiledbsc/soma_collection.py
```diff
@@ -210,33 +211,39 @@ def query(
...
-                var_query_string=var_query_string,
-                obs_ids=obs_ids,
-                var_ids=var_ids,
-            )
-            if soma_slice is not None:
-                # print("Slice SOMA from", soma.name, soma.X.data.shape(), "to", soma_slice.ann.X.shape)
-                soma_slices.append(soma_slice)
+        soma_slice_futures = []
+
+        with ThreadPoolExecutor() as executor:
```
johnkerl, line 216: on a PR probably tomorrow I'll add the kwarg `max_workers=self._soma_options.max_workers` which will take just minutes to code up, and will be a reasonable boundedness and keep this from getting out of control for larger collections. perhaps defaulting `soma_options.max_workers` to 5 or 10.

johnkerl: other than that this looks fine. i just pushed a commit with black/isort/flake8 mods to attempt CI pass.

johnkerl:
CI fails are in C++, separate issue to be handled separately

apis/python/src/tiledbsc/soma_collection.py
johnkerl: thanks @Shelnutt2 !

PR #245, opened 2022-08-16
Continuation of #244

johnkerl:
Let the `X` layer name be selectable in `from_anndata` as well as`to_anndata` (which was done on #244).


johnkerl:
CI fails are in C++, orthogonal to this PR

Shelnutt2 (approved):

PR #248, opened 2022-08-17
Fix NaN issue on SOMA-slice concat

johnkerl:
As at https://discourse.scverse.org/t/help-with-concat/676

At https://cloud.tiledb.com/notebooks/details/johnkerl-tiledb/292c8886-cb2b-4ba4-a939-828bcc5fbccb/preview verifed that the `del` workaround is no longer needed:

Screen Shot 2022-08-17 at 11 36 34 AM


aaronwolen (commented):
Looks good and solves the immediate issue. The explanatory comments were super informative and helped me think through this a little more. 

My concern is our method is making assumptions about how users want to concatenate these slices which could produce unintended consequences. For example, if `var2` (from your [post](https://discourse.scverse.org/t/help-with-concat/676)) had an extra column, `vd`...

```
       va     vb vc  vd
gene1  80   True  h   1
gene4  90  False  i   2
```

...then running...

```py
varc = pd.concat([var1, var2], join="outer")
varc = varc[~ varc.index.duplicated()]
```

...produces...

```py
       va     vb vc   vd
gene1  30   True  c  NaN
gene2  40  False  d  NaN
gene3  50   True  e  NaN
gene4  90  False  i  2.0
```

...losing `gene1`'s value.

Just a contrived but the point is I'm not sure we can accommodate all use cases without exposing all of `anndata.concat`'s args. At that point it probably makes more sense for users to use that directly. 

In summary, maybe we don't need a `SOMAslice.concat` method and just provide a list of `AnnData` objects for users to concatenate themselves.

What do you think, @johnkerl? Also curious to hear from @ambrosejcarr and @bkmartinjr.

johnkerl:
I do want something low-keystrokey — opt-in now — to keep example notebooks punchy & concise & not have people have to spell out messy concats in every notebook.

Also we can add a caveat (maybe even an assertion in the code) that this new method only works right if the `obs` & `var` have the same columns.

Which is what this does (in some of our notebook examples):
https://cloud.tiledb.com/notebooks/details/johnkerl-tiledb/292c8886-cb2b-4ba4-a939-828bcc5fbccb/preview

```
obs_attrs_set = None
var_attrs_set = None
for soma in soco:
    if obs_attrs_set is None:
        obs_attrs_set = set(soma.obs.keys())
        var_attrs_set = set(soma.var.keys())
    else:
        obs_attrs_set = set(soma.obs.keys()).intersection(obs_attrs_set)
        var_attrs_set = set(soma.var.keys()).intersection(var_attrs_set)
obs_attrs = sorted(list(obs_attrs_set))
var_attrs = sorted(list(var_attrs_set))
```

and i’d like that to be a ready-to-call function as well

So i propose:

* Make a new static/class method `SOMA.find_common_obs_and_var(somas: Sequence[SOMA]) -> Tuple[List[str], List[str]]` (or maybe a better name) which gets the common `obs` keys & the common `var` keys. Its body would be the thing I showed above.
* `soco.query` would be like now but would not concat -- it would just return `List[SOMASlice]`.
* Make a new static/class method `SOMASlice.concat(soma_slices: List[SOMASlice]) -> SOMASlice` which would:
  * assert obs/var keys the same
  * do the ‘new and improved’ concat that i’m trying to bugfix on this PR
  * people can call this if they want; they don't have to if they don't
  * i think there are valuable, hard-won learnings on this concat bug-fix which i don't want to throw away entirely & have others have to re-implement (or fail to re-implement)

aaronwolen:
The above plan makes sense to me!

johnkerl:
@aaronwolen ready for re-review

aaronwolen (approved):

johnkerl:
thanks @aaronwolen !!!

PR #249, opened 2022-08-18
Support update-write of SOMAs

johnkerl:
Heretofore, `from_anndata` / `from_hd5` would fail if the SOMA already existed. Now, those can be used to update a SOMA..

Note: this was the plan all along; it was blocked by a missing `TileDB-Py` feature which is now deployed, so it's OK to move forward now.

### DIFF
--- a/apis/python/src/tiledbsc/uns_array.py
+++ b/apis/python/src/tiledbsc/uns_array.py
@@ -5,6 +5,7 @@ import pandas as pd
 import scipy.sparse as sp
 import tiledb
 
+import tiledbsc.logging
 import tiledbsc.util as util
 
 from .logging import log_io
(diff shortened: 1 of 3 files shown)


apis/python/src/tiledbsc/uns_array.py
```diff
@@ -89,13 +90,15 @@ def from_numpy_ndarray(self, arr: np.ndarray) -> None:
...
-        # overwrite = False
-        # if self.exists:
-        #     overwrite = True
-        #     log_io(None, f"{self._indent}Re-using existing array {self.uri}")
-        # tiledb.from_numpy(uri=self.uri, array=arr, ctx=self._ctx, overwrite=overwrite)
-        # TODO: find the right syntax for update-in-place (tiledb.from_pandas uses `mode`)
-        tiledb.from_numpy(uri=self.uri, array=arr, ctx=self._ctx)
+        if self.exists():
+            tiledbsc.logging.logger.info(
+                f"{self._indent}Re-using existing array {self.uri}"
```
aaronwolen, line 95: Maybe _updating_ existing array to be [consistent with R](https://github.com/TileDB-Inc/tiledbsc/blob/6c3331611a4f06558c3c802dcf8813ff0b33bdae/R/AssayMatrix.R#L58-L62)?

aaronwolen (approved):
One minor suggestion but looks good

apis/python/src/tiledbsc/uns_array.py
johnkerl: This is better wording with or without the R precedent!

It was a copypasta from somewhere else, somewhen else -- thanks for the catch 😁

PR #253, opened 2022-08-22
Fix for MacOS localhost build

johnkerl:
Build failed with

```
make[3]: *** No targets specified and no makefile found.  Stop.
```

which was because the build does `cd` to `build/libtilesbsc` which is an empty directory, while `cmake` outputs had been written to `libtiledbsc` rather than `build/libtiledbsc`, so `build/libtilebsc` was an empty directory.

Also note I found that in spite of this PR,  I needed to clean out `cmake` artifacts from `libtiledbsc` -- just a caveat for anyone else running into this issue.

### DIFF
--- a/scripts/bld
+++ b/scripts/bld
@@ -16,6 +16,13 @@ export pybind11_DIR=$(python -m pybind11 --cmakedir)
 # remove existing build and install directories
 rm -rf build dist
 
+if [ $(uname) = "Darwin" ]; then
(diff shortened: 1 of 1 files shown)


gspowley (approved):

Shelnutt2:
@johnkerl just fyi you should always start from a clean build directory for cmake projects when making build related changes.

PR #254, opened 2022-08-23
Use dictionary encoding for X/data var_id dim

johnkerl:


apis/python/src/tiledbsc/util.py
```diff
@@ -18,7 +18,7 @@
 
 SOMA_OBJECT_TYPE_METADATA_KEY = "soma_object_type"
 SOMA_ENCODING_VERSION_METADATA_KEY = "soma_encoding_version"
-SOMA_ENCODING_VERSION = "0"
+SOMA_ENCODING_VERSION = "0.1"
```
johnkerl, line 21: note `SOMA_ENCODING_VERSION` is present in `tiledbsoma` (currently `"1"`) so my preference is we have this dotted syntax

if there were ints, we'd, i guess, start `tiledbsoma` at 1000 or 10000 or something -- ?

they need to increment independently for `tiledbsc` and `tiledbsoma`

johnkerl:
Current narrative is we should leave this a configuration option, not the default.

PR #256, opened 2022-08-24
Support return_arrow for various queries

johnkerl:
Note that ingest-from-arrow is a separate issue and is not covered on this PR.

### DIFF
--- a/apis/python/src/tiledbsc/annotation_dataframe.py
+++ b/apis/python/src/tiledbsc/annotation_dataframe.py
@@ -1,8 +1,9 @@
 from concurrent.futures import ThreadPoolExecutor
-from typing import Optional, Sequence, Set, Tuple
+from typing import Optional, Sequence, Set, Tuple, Union
 
 import numpy as np
 import pandas as pd
+import pyarrow as pa
 import tiledb
 
 import tiledbsc.util as util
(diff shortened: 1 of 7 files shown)


apis/python/src/tiledbsc/annotation_dataframe.py
```diff
@@ -220,6 +244,52 @@ def _ascii_to_unicode_dataframe_readback(self, df: pd.DataFrame) -> pd.DataFrame
...
+        # pyarrow's way of handling 'bytes'
+        if len(series) > 0 and (
+            type(series[0]) == pa.LargeBinaryArray
+            or type(series[0]) == pa.LargeStringScalar
+        ):
+            return (array_number, True, series.cast(pa.string()))
+        else:
+            return (array_number, False, None)
+
+    def _ascii_to_unicode_arrow_readback(self, df: pa.Table) -> pa.Table:
```
Shelnutt2, line 263: What is the purpose of this function? This makes a full copy of fields and then creates a new pyarrow table from python arrays. Why do we need to perform a copy?

johnkerl: @Shelnutt2 it's a separate function so as to be called from the `ThreadPoolExecutor`.

I'm open to feedback if you have a better way.

It does a copy _only_ when the need to do ASCII-to-Unicode. And that is _crucial_ -- not optional -- since without this user-level queries like `'cell_type == "myeloid cell"'` will return nothing since columns would have values like `b"myeloid cell"` which is `!=` to `"myeloid cell"`.

Please read https://github.com/single-cell-data/TileDB-SingleCell/pull/141 -- we had extensive discussions on this a couple months ago. We must (can't not) do ASCII-to-Unicode conversions until such time as we have core mods in place which:

* Permit string attrs to be Unicode _and be queryable using query conditions_
* Permit string dims to be Unicode at all

cc @ihnorton

Shelnutt2: This will be followed up separately.

Shelnutt2 (approved):

PR #259, opened 2022-08-25
tiledbsc.logging.info() by default for good uploader UX

johnkerl:
The first screenshot is currently the default, and is less-good UX since it gives the user no idea about the ETA or progress or even liveness.

The second screenshot is currently only opt-in via `tiledbsc.logging.info()`, and is good UX.

This PR makes the good UX the default UX.

Screen Shot 2022-08-25 at 12 43 31 PM

Screen Shot 2022-08-25 at 12 51 02 PM


gsakkis (approved):
LGTM :ok_hand:

PR #260, opened 2022-08-25
Improve write-retry logic

johnkerl:
Especially important for trivial transients like 'Connection reset' on an HTTP socket for cloud writes

### DIFF
--- a/apis/python/src/tiledbsc/annotation_matrix.py
+++ b/apis/python/src/tiledbsc/annotation_matrix.py
@@ -145,8 +145,16 @@ class AnnotationMatrix(TileDBArray):
             self._create_empty_array([matrix.dtype] * nattr, attr_names)
 
         df = pd.DataFrame(matrix, columns=attr_names)
-        with tiledb.open(self.uri, mode="w", ctx=self._ctx) as A:
-            A[dim_values] = df.to_dict(orient="list")
+        exc = None
+        for _ in range(self._soma_options.num_write_retries):
+            try:
+                with tiledb.open(self.uri, mode="w", ctx=self._ctx) as A:
+                    A[dim_values] = df.to_dict(orient="list")
+                break
+            except tiledb.TileDBError as e:
+                exc = e
+        if exc is not None:
+            raise exc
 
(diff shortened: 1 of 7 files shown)


apis/python/src/tiledbsc/annotation_matrix.py
```diff
@@ -131,8 +131,16 @@ def _numpy_ndarray_or_scipy_sparse_csr_matrix(
             self._create_empty_array([matrix.dtype] * nattr, attr_names)
 
         df = pd.DataFrame(matrix, columns=attr_names)
-        with tiledb.open(self.uri, mode="w", ctx=self._ctx) as A:
-            A[dim_values] = df.to_dict(orient="list")
+        exc = None
```
johnkerl, line 134: maybe https://github.com/litl/backoff would be better -- ?

gsakkis: Another (more popular?) option is [tenacity](https://github.com/jd/tenacity), I had used it in the past as well its precursor ([retrying](https://github.com/rholder/retrying)).

Using any of these libraries is definitely better than copy-pasting the same boilerplate 15 times.

johnkerl:
Setting back to draft for rework

Shelnutt2 (changes requested):
This type of retry logic should not be at the application level here. Further I think we should spent more time understanding the root cause for this and why the existing logic was not sufficient.

johnkerl:
> This type of retry logic should not be at the application level here. Further I think we should spent more time understanding the root cause for this and why the existing logic was not sufficient.

@Shelnutt2 the scenario is:

* User doing a time-consuming upload of a largish SOMA from `.h5ad` to a `tiledb://...` URI
* Half an hour in (say) there is an exception due to (some examples I've seen):
  * 502 (transient)
  * Connection reset (transient)

This happens not always but often enough I've seen it a few times, for myself or for a colleague.

The crucial UX is that this should not abort the upload -- it makes us look non-robust to ordinary server-side transients.

I'm happy to learn about what/where "the existing logic" is -- looking forward to it!

PR #261, opened 2022-08-25
Build and install libtiledbsc with pip install

gspowley:
Build and install the `libtiledbsc` native library as part of the pip install process.
* Local non-editable install: `pip install .`
* Local editable install: `pip install -e .`

The `./scripts/bld` script is reused for pip install, CI, and local builds.

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -1,7 +1,139 @@
-# This is only used for `python setup.py develop` for local editable installs.
-# Otherwise one should use `pip install .` for local non-editable installs,
-# or `pip install tiledbsc` for PyPI installs.
-import setuptools
+# This file is enables building the libtiledbsc external package as part of the
+# tiledbsc module install process.
+#
+# Local non-editable install:
+#   `pip install .`
+#
+# Local editable install:
+#   `pip install -e .`
+#
+# Install from PyPI
+#   `pip install tiledbsc`
+#
+# Based on ideas from https://github.com/pybind/cmake_example
+# The `bld` script here is reused for pip install, CI, and local builds.
+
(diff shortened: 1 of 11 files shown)


apis/python/setup.py
```diff
@@ -1,7 +1,123 @@
...
+MODULE_NAME = "tiledbsc"
+EXT_NAME = "tiledbsc.libtiledbsc"
+
+
+def find_or_build(setuptools_cmd):
+    # Setup paths
+    python_dir = os.path.abspath(os.path.dirname(__file__))
+    src_dir = f"{python_dir}/src/{MODULE_NAME}"
+    scripts_dir = f"{python_dir}/../../scripts"
+    lib_dir = f"{python_dir}/../../dist/lib"
```
Shelnutt2, line 33: I think this can be `lib`, `lib32` or `lib64` depending on the platform. I also believe it can be `bin` on Windows?

gspowley: The `lib` directory is set in [CMakeLists.txt](https://github.com/single-cell-data/TileDB-SingleCell/blob/53aad18894a5a3fb3632deee4ca55ab230a2f00a/libtiledbsc/src/CMakeLists.txt#L245), so it's consistent across platforms.

Windows support is a good point though. A windows build is working on my laptop, but failing in CI, so `libtiledbsc` is not supported in windows for now.

Shelnutt2: We should eventually switch to use the path provided by `GNUInstallDirs` so the installations will be validate on the various linux systems.

Shelnutt2 (approved):

johnkerl (approved):

PR #266, opened 2022-08-28
Continuing to try to fix PyPI CI with libtiledbsc

johnkerl:
https://github.com/single-cell-data/TileDB-SingleCell/runs/8059247274?check_suite_focus=true

```
Checking dist/bin: ERROR    InvalidDistribution: Unknown distribution format: 'bin'                
```

Before https://github.com/single-cell-data/TileDB-SingleCell/pull/261 the command

```
python -m pip wheel --no-deps --wheel-dir=dist apis/python
```

created only

```
dist/tiledbsc-0.1.x-cp310-cp310-linux_x86_64.whl
```

but now there is

```
$ ls -l dist
total 776
drwxrwxr-x 2 ubuntu ubuntu   4096 Aug 22 22:11 bin
drwxrwxr-x 3 ubuntu ubuntu   4096 Aug 22 22:11 include
drwxrwxr-x 2 ubuntu ubuntu   4096 Aug 22 22:11 lib
-rw-r--r-- 1 root   root   781539 Aug 28 17:24 tiledbsc-0.1.9-cp310-cp310-linux_x86_64.whl

$ ls -l dist/*
-rw-r--r-- 1 root   root   781539 Aug 28 17:24 dist/tiledbsc-0.1.9-cp310-cp310-linux_x86_64.whl

dist/bin:
total 932
-rwxr-xr-x 1 ubuntu ubuntu 952952 Aug 22 22:10 tdbsc

dist/include:
total 4
drwxrwxr-x 2 ubuntu ubuntu 4096 Aug 22 22:11 tiledbsc

dist/lib:
total 1932
-rw-r--r-- 1 ubuntu ubuntu 982176 Aug 22 22:11 libtiledbsc.cpython-310-x86_64-linux-gnu.so
-rw-r--r-- 1 ubuntu ubuntu 992000 Aug 22 22:10 libtiledbsc.so
```

This is a follow-on to https://github.com/single-cell-data/TileDB-SingleCell/pull/265 -- there, the fix was unambiguous and immediate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,8 +43,7 @@ jobs:
       run: python -m pytest apis/python/tests
 
     - name: Build wheel distribution
-      if: matrix.os == 'ubuntu-22.04' && github.event_name == 'release'
-      run: rm -rvf dist/bin && python -m pip wheel --no-deps --wheel-dir=dist apis/python
+      run: python -m pip wheel --no-deps --wheel-dir=dist-wheel apis/python
 
     - name: Publish package to TestPyPI
       if: matrix.os == 'ubuntu-22.04' && github.event_name == 'release'
@@ -54,6 +53,7 @@ jobs:
         repository_url: https://test.pypi.org/legacy/
         user: __token__
         password: ${{ secrets.TEST_PYPI_TOKEN }}
+        packages_dir: dist-wheel
         verbose: true
 
     - name: Publish package to PyPI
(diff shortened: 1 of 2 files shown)


gspowley:
This looks like an issue with `pip wheel` and `libtiledbsc` both using `dist` as the destination of their respective (and different) builds.

If `pip wheel` needs to use `dist` we can modify `libtiledbsc` to use a different directory. Otherwise, I recommend changing `--wheel-dir` to use something other than `dist`, so `libtiledbsc` can install to `dist` and remain consistent with other TileDB projects (like TileDB -VCF).

PR #272, opened 2022-08-30
Flexible collection-mapper

johnkerl:
As pointed out by @bkmartinjr , the [soma.query](https://github.com/single-cell-data/TileDB-SingleCell/blob/main/apis/python/src/tiledbsc/soma.py#L303-L313) / [SOMASlice](https://github.com/single-cell-data/TileDB-SingleCell/blob/main/apis/python/src/tiledbsc/soma_slice.py) API, while handy, gives the user (at present) all of `obs`, `var`, and `X`, no more, no less.

While we can (and should) take user feedback on how best to adapt to their needs, a simple alternative proposed by @bkmartinjr and echoed again today by @mdylan2 is a simple 'mapper' where users can simply transform a list of `SOMA` objects (or a `SOMACollection`) into a list of what they want.


apis/python/src/tiledbsc/soma.py
```diff
@@ -547,6 +547,99 @@ def from_soma_slice(
...
+
+        output = {}
+        for name, future in futures.items():
+            output[name] = future.result()
+        return output
+
+    @classmethod
+    def _map_sequentially(
+        cls,
+        somas: List[SOMA],  # type: ignore[no-untyped-def]
```
johnkerl, line 631: i've spent two hours of my life, which i'll never get back, struggling and failing to find a syntax that `mypy` will accept for these _incredibly simple_ callbacks -- struggling and failing to find an acceptable syntax using `Callable` as well as struggling and failing to find an acceptable ignore-no-untyped-def syntax.

I'm open to constructive, non-dismissive, supportive feedback here from anyone who has it -- thank you in advance -- 🙏

Shelnutt2:
We need a design doc for this before we work on the implementation. Let's put one together so we can agree upon how this should work.

apis/python/src/tiledbsc/soma.py
```diff
@@ -547,6 +547,98 @@ def from_soma_slice(
...
+                        None,
+                    )
+                else:
+                    futures[soma.name] = executor.submit(
+                        soma_callback,
+                        soma,
+                        data[soma.name],
+                    )
+
+        output = {}
```
thetorpedodog, line 622: like below, you can do `return {name: ft.result() for name, ft in futures.items()}`.

apis/python/src/tiledbsc/soma.py
```diff
@@ -547,6 +547,98 @@ def from_soma_slice(
 
         return soma
 
+    @classmethod
+    def map(
+        cls,
+        somas: List[SOMA],
+        soma_callback: Callable[[SOMA, Optional[Any]], Any],
```
thetorpedodog, line 554: If somewhere at the module level you add

```python
_T = TypeVar("_T")
```

then here you can change `soma_callback` to be a `Callable[[Soma, Optional[Any]], _T]`, and make the return type `Dict[str, _T]`. That will give the caller an even better type-checking experience, since now the return type from the callable will be automatically carried through to the result.

If you do this, you’ll also want to make the same change down in the `_map_...` functions.

(note: suggesting the use of `_T` rather than `T` as a name so that the TypeVar variable doesn’t look exported. using just `T` is semantically exactly the same and completely fine if you think it’s better)

apis/python/src/tiledbsc/soma.py
```diff
@@ -547,6 +547,98 @@ def from_soma_slice(
 
         return soma
 
+    @classmethod
+    def map(
+        cls,
+        somas: List[SOMA],
```
thetorpedodog, line 553: does this need to be a `List` in particular, or will any `Iterable` (or, if you need to be slightly more specific, `Sequence`) work here?

thetorpedodog (commented):
I’m glad you figured out your type annotation issues yesterday! I decided to give it a once-over and overall it looks very good. I have a couple small suggestions that can make the experience even better for callers.

PR #273, opened 2022-08-31
Update ASCII storage for dataframes

johnkerl:
On collaboration with @nguyenv on https://github.com/TileDB-Inc/TileDB-Py/pull/1304 -- new/upated feedback (thanks @nguyenv !) is that all along we can obtain reduce type-conversion at readback by a one-line change in this package.

Note this means that while before this PR we can store (encode/decode) user-supplied Unicode data in dataframes, as of this PR we no longer can. That ability will be restored in a future core release -- perhaps early 2023. Meanwhile unit tests on this PR that used to validate the ability to encode/decode user-supplied Unicode data in dataframes have been commented out, to accommodate the feature loss.

Summary:

* Write string dims/attrs as `"ascii"` per se, not `bytes` and not `str`
* Since old SOMAs written before this PR won't have all `"ascii"`, conditionally retain the decode-on-readback logic which is _essential_ since otherwise `b"foo" != "foo"`
* Please see also the clear presentation at https://github.com/single-cell-data/TileDB-SOMA/issues/99


johnkerl:
Status note: the CI failure is puzzling me as unit tests are passing for me locally 👀

johnkerl:
Sorry for the delay.

@Shelnutt2 @nguyenv this is ready for review.

johnkerl:
@nguyenv @ihnorton @Shelnutt2 ping 🙏

apis/python/src/tiledbsoma/annotation_dataframe.py
```diff
@@ -76,10 +75,19 @@ def ids(self) -> Sequence[str]:
...
-            # TileDB string dims are ASCII not UTF-8. Decode them so they readback
-            # not like `b"AKR1C3"` but rather like `"AKR1C3"`.
+
+            # TileDB string dims are ASCII not UTF-8. Decode them so they readback not like
+            # `b"AKR1C3"` but rather like `"AKR1C3"`. Update as of
+            # https://github.com/TileDB-Inc/TileDB-Py/pull/1304 these dims will read back OK.
             retval = A.query(attrs=[], dims=[self.dim_name])[:][self.dim_name].tolist()
-            return [e.decode() for e in retval]
+
+            retval = [e.decode() for e in retval]
```
nguyenv, line 84: Could something like this work?

```
retval = A.query(attrs=[], dims=[self.dim_name])[:][self.dim_name]
return np.frombuffer(retval, dtype="U").tolist()
```

apis/python/src/tiledbsoma/annotation_dataframe.py
```diff
@@ -76,10 +75,19 @@ def ids(self) -> Sequence[str]:
...
+
+            if len(retval) > 0 and isinstance(retval[0], bytes):
+                return [e.decode() for e in retval]
+            else:
+                # list(...) is there to appease the linter which thinks we're returning `Any`
+                return list(retval)
```
nguyenv, line 90: Could something like this work?

```
retval = A.query(attrs=[], dims=[self.dim_name])[:][self.dim_name]
return np.frombuffer(retval, dtype="U").tolist()
```

johnkerl: no sorry @nguyenv !

```
    return np.frombuffer(retval, dtype="U").tolist()
ValueError: itemsize cannot be zero in type
```

johnkerl: But regardless all I'm doing is putting an `if` around known-good code, invoking it when needed for older arrays which predate this PR!

johnkerl: @nguyenv what works remains to accept or reject this PR?

nguyenv (approved):

johnkerl:
Thanks @nguyenv ! :)

PR #281, opened 2022-09-13
Split tiledbsc and tiledbsoma into branches [WIP]

johnkerl:
* `main-v0` has `tiledbsc`
* `main` has `tiledbsoma`


johnkerl:
Miskeyed; will redo

PR #290, opened 2022-09-16
Experiment to fix Windows R-CMD-check errors

johnkerl:


johnkerl:
Will do a separate PR based on advice from @eddelbuettel

PR #292, opened 2022-09-16
Update R CI logic

johnkerl:
Follow-on from https://github.com/single-cell-data/TileDB-SOMA/pull/285:

* The R code is now in an `apis/R` subdirectory; some spellings need updating
* Particular handling for Windows with `apis\R` vs `apis/R` when using YAML syntax
* Following advice from @eddelbuettel regarding use of helper scripts -- basic idea is to rely less on YAML syntax and more on shell-script syntax, as these are clear and unambiguously expressive


.github/workflows/R-CMD-check.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
 
       - name: Install libgeos on Linux
         if: runner.os == 'Linux'
-        run: sudo apt-get update -y && sudo apt-get install -y libgeos-dev
+        run: sudo ./tools/ci/install-linux-dependencies.sh
```
eddelbuettel, line 37: (nean == Not Even A Nit (TM))  I usually draw at 'too tiny script so I would have kept most shell expressions here, specifically one-liners but it is six of one and half a dozen of another and if you like it better the other way :man_shrugging:

tools/ci/build-and-check-r-package.sh
```diff
@@ -0,0 +1,21 @@
...
+
+echo "::group::Check TileDB R package"
+## set flag to tolerate missing suggested packages (as eg vignette building tools)
+export _R_CHECK_FORCE_SUGGESTS_=FALSE
+## set an 'under valgrind' variable (see SC-19185)
+export _RUNNING_UNDER_VALGRIND_=TRUE
+## tell valgrind to use '-s'
+export VALGRIND_OPTS="-s"
+## check package
+R CMD check --use-valgrind --as-cran --no-manual --ignore-vignettes tiledb_*.tar.gz
```
eddelbuettel, line 18: Using valgrind is a nice touch but _expensive_. I doubt I would turn it on in each run.  You could make it an environment-variable toggle.  But again, if it works for you ...

tools/ci/install-pkgdown-dependencies.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+# Context: .github/workflows/*.yaml
+export R_REMOTES_NO_ERRORS_FROM_WARNINGS='true'
+Rscript -e 'install.packages("remotes")'
+Rscript -e 'install.packages("pkgdown")'
```
eddelbuettel, line 5: That is style wart:   `Rscript -e 'install.packages(c("remotes", "pkgdown"))'`  is more compact.

tools/ci/install-r-dependencies.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+# Context: .github/workflows/*.yaml
+export R_REMOTES_NO_ERRORS_FROM_WARNINGS='true'
+Rscript -e 'install.packages("remotes")'
+Rscript -e 'install.packages("rcmdcheck")'
```
eddelbuettel, line 5: See above.

tools/ci/build-and-check-r-package.sh
johnkerl: ooooops this is a copy-paste error on my part, i did _not_ intend to use `valgrind` ... 👀

tools/ci/install-r-dependencies.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+# Context: .github/workflows/*.yaml
+export R_REMOTES_NO_ERRORS_FROM_WARNINGS='true'
+Rscript -e 'install.packages(c("remotes", "rcmdcheck"))'
```
eddelbuettel, line 4: You don't actually _use_ the `rcmdcheck` package as you invoke `R CMD check --as-cran --no-manual --ignore-vignettes tiledb_*.tar.gz`.  So you don't need to install it.

eddelbuettel (commented):
Also:   R-CMD-check.yaml is buttugly if all other file in the directory are neatly lowercased.  Maybe r-pkg-check.yaml or something like that.  Another 'nean' ...

tools/ci/build-and-check-r-package.sh
```diff
@@ -0,0 +1,17 @@
...
+echo "::group::Build TileDB R package"
+## build package
+R CMD build --no-build-vignettes --no-manual .
+echo "::endgroup::"
+
+echo "::group::Check TileDB R package"
+## set flag to tolerate missing suggested packages (as e.g. vignette-building tools)
+export _R_CHECK_FORCE_SUGGESTS_=FALSE
+## check package
+R CMD check --as-cran --no-manual --ignore-vignettes tiledb_*.tar.gz
```
eddelbuettel, line 14: That is a likely bug.  The package you copied this from is called `tiledb_X.Y.Z.tar.gz` so the file-globbing is correct.  Your package here appears to be `tiledbsoma` per the `Package: tiledbsoma` in `DESCRIPTION` so you need to take out the underscore there, or inject 'soma' in front of it.

eddelbuettel (changes requested):

tools/ci/build-and-check-r-package.sh
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/296

thank you! :)

PR #307, opened 2022-09-19
debug [experimental / please do not review]

johnkerl:
Teammates: I know I don't "need" (in some sense) to open a PR to accomplish what this is doing; I could use a branch. Please note that it has always been (and continues to be) my strong preference to use draft PRs because they're easier to *discover* -- the most recent draft PR is always right there at the top of the PR list; my most recent branch, is a pop-up list and a scroll-down through perhaps dozens of other branches, & I've always preferred to let the machine make it easier for me to find things.

Hence my marking this as "please do not review".

Cheers!


johnkerl:
completed; see also for #313 for the keepables.

PR #313, opened 2022-09-21
cmake supportability findings

johnkerl:
This is in the context of SC 21583. After an extended debug session, these are mods worth keeping, as quality-of-life improvements.


johnkerl:
Per internal conversation

Issue #315, opened 2022-09-21
Python tiledbsoma: repr fails when accessing data on a private bucket


bkmartinjr:
This occurs when accessing data on a private bucket.  I have all of the various AWS environment variables set correctly (demonstrated by the successful `atlas.exists()` expression).

It appears that the repr may not be inheriting the context used to open the original SOMA object?

```
In [20]: config
Out[20]: {'py.init_buffer_bytes': 1073741824, 'vfs.s3.region': 'us-west-2'}

In [21]: atlas_uri
Out[21]: 's3://bruce-uswest-tmp/tmp/atlas-20220916/'

In [22]: atlas = soma.SOMACollection(atlas_uri, ctx=tiledb.Ctx(config))

In [23]: atlas.get_uri()
Out[23]: 's3://bruce-uswest-tmp/tmp/atlas-20220916/'

In [24]: atlas.exists()
Out[24]: True

In [25]: atlas
Out[25]: ---------------------------------------------------------------------------
TileDBError                               Traceback (most recent call last)
File ~/projects/TileDB-SingleCell/venv/lib/python3.9/site-packages/IPython/core/formatters.py:707, in PlainTextFormatter.__call__(self, obj)
    700 stream = StringIO()
    701 printer = pretty.RepresentationPrinter(stream, self.verbose,
    702     self.max_width, self.newline,
    703     max_seq_length=self.max_seq_length,
    704     singleton_pprinters=self.singleton_printers,
    705     type_pprinters=self.type_printers,
    706     deferred_pprinters=self.deferred_printers)
--> 707 printer.pretty(obj)
    708 printer.flush()
    709 return stream.getvalue()

File ~/projects/TileDB-SingleCell/venv/lib/python3.9/site-packages/IPython/lib/pretty.py:410, in RepresentationPrinter.pretty(self, obj)
    407                         return meth(obj, self, cycle)
    408                 if cls is not object \
    409                         and callable(cls.__dict__.get('__repr__')):
--> 410                     return _repr_pprint(obj, self, cycle)
    412     return _default_pprint(obj, self, cycle)
    413 finally:

File ~/projects/TileDB-SingleCell/venv/lib/python3.9/site-packages/IPython/lib/pretty.py:778, in _repr_pprint(obj, p, cycle)
    776 """A pprint that just redirects to the normal repr function."""
    777 # Find newlines and replace them with p.break_()
--> 778 output = repr(obj)
    779 lines = output.splitlines()
[...]

PR #318, opened 2022-09-21
Propagate ctx through reprs and member-constructs

johnkerl:
Addresses #315


bkmartinjr (approved):
LGTM. TY for the fix!

bkmartinjr:
FYI - this PR did not fix the bug.  I added the required change to the bug report.

PR #319, opened 2022-09-21
Enable option for enhancing debug-log visibility

johnkerl:
This is in the context of SC 21583. After an extended debug session, these are mods worth keeping, as quality-of-life improvements. This is opt-in via the two environment variables checked at the top of `scripts/bld`.

### DIFF
--- a/libtiledbsoma/src/CMakeLists.txt
+++ b/libtiledbsoma/src/CMakeLists.txt
@@ -1,4 +1,9 @@
-message(STATUS "Starting TileDB-SC build.")
+if(DEFINED ENV{LIBTILEDBSOMA_DEBUG_BUILD})
+  message(STATUS "")
+  message(STATUS "START libtiledbsoma/src/CMakeLists.txt")
+else()
+  message(STATUS "Starting TileDB-SC build.")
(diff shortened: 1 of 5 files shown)


libtiledbsoma/src/CMakeLists.txt
```diff
@@ -1,4 +1,9 @@
-message(STATUS "Starting TileDB-SC build.")
+if(DEFINED ENV{LIBTILEDBSOMA_DEBUG_BUILD})
+  message(STATUS "")
+  message(STATUS "START libtiledbsoma/src/CMakeLists.txt")
+else()
+  message(STATUS "Starting TileDB-SC build.")
```
gspowley, line 5: LGTM. 

Minor request: Could you change this to `TileDB-SOMA`? We missed this one in the conversion to SOMA.

gspowley (approved):

libtiledbsoma/src/CMakeLists.txt
johnkerl: dangit i was gonna do just that but sneezed & forgot :)

PR #333, opened 2022-09-23
revise version API to match spec

### LINKED ISSUE #310: Python tiledbsoma: version operations do not match latest spec
the separate operations "get SOMA version" (aka API version) and "get implementation version" (aka software package release version) need to be implemented.  Currently there is just a single "get version" which is obsolete

bkmartinjr:
Fixes #310 

Change version API to match spec:
* removed get_version()
* added get_SOMA_version() and get_implementation_version()


johnkerl (commented):
@bkmartinjr  LGTM but (unbeknownst to you probably, my apologies) @Shelnutt2 and @eric-hughes-tiledb had some thoughts on string vs int versions at

https://github.com/single-cell-data/TileDB-SOMA/pull/254#discussion_r952130372

That said, maybe we can accept this PR as-is -- to make code match current spec -- _and then_ have a follow-on regarding a spec mod to use ints here

bkmartinjr:
I agree that we should build to the spec.  If the spec changes, this is super easy to update.  Tying to live with spec & code being different sounds like mass confusion in my head!

johnkerl:
> I agree that we should build to the spec. If the spec changes, this is super easy to update. Tying to live with spec & code being different sounds like mass confusion in my head!

Seconded.

johnkerl (approved):

PR #350, opened 2022-09-28
Replace array->timestamp() with array->open_timestamp_end()

eddelbuettel:
The `timestamp()` accessor is deprecated, replacing it with `open_timestamp_end()` is recommended in the header.


gspowley (approved):
Perfect, thanks @eddelbuettel!

PR #351, opened 2022-09-29
Bump R version

johnkerl:
For first `tiledbsoma` release: with R and Python both

### DIFF
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -7,7 +7,7 @@ Description: Interface for working with 'TileDB'-based Stack of Matrices,
     from and export to in-memory formats used by popular toolchains like
     'Seurat', 'Bioconductor', and even 'AnnData' using the companion Python
     package.
-Version: 0.1.3.9006
+Version: 0.1.12
 Authors@R: c(
     person(given = "Aaron",
            family = "Wolen",
--- a/apis/r/NEWS.md
+++ b/apis/r/NEWS.md
@@ -83,6 +83,16 @@ Effectively, this stores the data on disk in column-major order. However, the `t
 - Added `with_allocation_size_preference()` helper to temporarily set the allocation size preference for testing.
 - Tests were added to verify the internal `dgtmatrix_to_dataframe()` will error out if an input list contains non-layerable matrices.
 
+# tiledbsoma 0.1.12
+
+## Package name
+
+The package is now `tiledbsoma`. This R code has been moved from [tiledbsc](https://github.com/TileDB-Inc/tiledbsc); now, Python and R `tiledbsoma` are both in this repo.
+
+## Version
+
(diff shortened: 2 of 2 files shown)


apis/r/DESCRIPTION
```diff
@@ -7,7 +7,7 @@ Description: Interface for working with 'TileDB'-based Stack of Matrices,
     from and export to in-memory formats used by popular toolchains like
     'Seurat', 'Bioconductor', and even 'AnnData' using the companion Python
     package.
-Version: 0.1.3.9006
+Version: 0.1.12
```
eddelbuettel, line 10: Can we make it 0.1.3.9007 and then integrate the first 'proof of light' C++ bindings from R?

eddelbuettel (dismissed):

apis/r/DESCRIPTION
johnkerl: @eddelbuettel thanks!

This is on the `main-old` branch not `main` -- completely independent of all `main` development involving C++ etc.

johnkerl: `main` is (for now) effectively a `dev` branch

aaronwolen: The motivation here is to sync the python/R api versions. You could bump the version to 0.1.13.9000 in your PR.

aaronwolen (approved):

apis/r/DESCRIPTION
johnkerl: See also
**https://github.com/single-cell-data/TileDB-SOMA/wiki/Developer-resources**

aaronwolen (changes requested):
Can you update `NEWS.md` with the new version and indicate the bump is to sync with python?

johnkerl:
Good call @aaronwolen ! :)

johnkerl:
@aaronwolen 
https://github.com/single-cell-data/TileDB-SOMA/pull/351/commits/e66026e96781c2ebf5b25789ef5f05060e7821d8

apis/r/NEWS.md
```diff
@@ -83,6 +83,16 @@ Effectively, this stores the data on disk in column-major order. However, the `t
 - Added `with_allocation_size_preference()` helper to temporarily set the allocation size preference for testing.
 - Tests were added to verify the internal `dgtmatrix_to_dataframe()` will error out if an input list contains non-layerable matrices.
```
aaronwolen, line 85: Thanks for adding! But this whole block should be moved to the top of `NEWS.md`, and `# tiledbsc (development version)` should be replaced with `# tiledbsoma 0.1.12` (like [so](https://gist.github.com/aaronwolen/4962f6cfc25a362cadd1175390042c4b)).

aaronwolen (dismissed):

apis/r/DESCRIPTION
eddelbuettel: >  This is on the main-old branch 

Gee. Likely not the last time I fell into _that_ pit.  Never mind then.

aaronwolen (approved):
Thanks!

PR #355, opened 2022-10-02
Conform to spec by reading as pyarrow.Table not pyarrow.RecordBatch

johnkerl:
This is the first in a group of three related PRs:

* This PR: conform to https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md by having `read` return`pyarrow.Table` not `pyarrow.RecordBatch` (as in an outdated version of that spec)
* #359: Forward-port https://github.com/single-cell-data/TileDB-SOMA/pull/273 from `main-old` which will truly have ASCII columns, obviating the need for our `util_arrow.ascii_to_unicode_pyarrow_readback`
* #360: Drop in C++ acceleration-library support for `read` methods, which will go in cleanly now
  * The C++ code returns `pyarrow.Table` and with the first PR our unit tests will be ready to go
  * When the C++ code reads Unicode cells it returns them as `pyarrow.LargeBinaryArray` (needing decode) but when we are properly writing ASCII cells via the Python write path then the C++ code will read ASCII cells and return them as strings (no longer needing decoding)

These three changesets could be done in a single PR, but that would be unmerciful to the reviewers.

### DIFF
--- a/apis/python/src/tiledbsoma/soma_dataframe.py
+++ b/apis/python/src/tiledbsoma/soma_dataframe.py
@@ -222,10 +222,6 @@ class SOMADataFrame(TileDBArray):
                 # This is the 'decode on read' part of our logic; in dim_select we have the
                 # 'encode on write' part.
                 # Context: https://github.com/single-cell-data/TileDB-SOMA/issues/99.
-                #
-                # Also: don't materialize these on read
-                # TODO: get the arrow syntax for drop
-                # df.drop(ROWID, axis=1)
                 yield util_arrow.ascii_to_unicode_pyarrow_readback(table)
 
     def read_all(
@@ -243,7 +239,7 @@ class SOMADataFrame(TileDBArray):
         # TODO: platform_config,
     ) -> pa.Table:
         """
-        This is a convenience method around ``read``. It iterates the return value from ``read`` and returns a concatenation of all the table-pieces found. Its nominal use is to simply unit-test cases.
+        This is a convenience method around ``read``. It concatenates all partial read results into a single Table. Its nominal use is to simplify unit-test cases.
         """
         return pa.concat_tables(
             self.read(
@@ -390,7 +386,7 @@ class SOMADataFrame(TileDBArray):
         id_column_name: Optional[str] = None,
     ) -> pd.DataFrame:
         """
-        Reads from SOMA storage into memory.  Iterates over batches from ``read_as_pandas``, concatenating the output into a single dataframe.  Convenient for unit-test use; also, handy whenever you're certain that the data being queried can be read entirely into memory.
+        This is a convenience method around ``read``. It concatenates all partial read results into a single DataFrame. Its nominal use is to simplify unit-test cases.
         """
         dataframes = []
         generator = self.read_as_pandas(
--- a/apis/python/src/tiledbsoma/soma_dataframe.py
+++ b/apis/python/src/tiledbsoma/soma_dataframe.py
@@ -245,7 +245,7 @@ class SOMADataFrame(TileDBArray):
         """
         This is a convenience method around ``read``. It iterates the return value from ``read`` and returns a concatenation of all the table-pieces found. Its nominal use is to simply unit-test cases.
         """
-        return util_arrow.concat_tables(
+        return pa.concat_tables(
             self.read(
                 ids=ids,
                 value_filter=value_filter,
@@ -271,13 +271,13 @@ class SOMADataFrame(TileDBArray):
 
         return self._cached_is_sparse
 
-    def write(self, values: pa.Table) -> None:
+    def write(self, values: pa.RecordBatch) -> None:
         """
-        Write an Arrow.Table to the persistent object.
+        Write an Arrow.RecordBatch to the persistent object.
 
-        :param values: An Arrow.Table containing all columns, including the index columns. The schema for the values must match the schema for the ``SOMADataFrame``.
+        :param values: An Arrow.RecordBatch containing all columns, including the index columns. The schema for the values must match the schema for the ``SOMADataFrame``.
 
-        The ``values`` Arrow Table must contain a ``soma_rowid`` (uint64) column, indicating which rows are being written.
+        The ``values`` Arrow RecordBatch must contain a ``soma_rowid`` (uint64) column, indicating which rows are being written.
         """
         self._shape = None  # cache-invalidate
 
--- a/apis/python/src/tiledbsoma/soma_dataframe.py
+++ b/apis/python/src/tiledbsoma/soma_dataframe.py
@@ -271,13 +271,13 @@ class SOMADataFrame(TileDBArray):
 
         return self._cached_is_sparse
 
-    def write(self, values: pa.RecordBatch) -> None:
+    def write(self, values: pa.Table) -> None:
         """
-        Write an Arrow.RecordBatch to the persistent object.
+        Write an Arrow.Table to the persistent object.
 
-        :param values: An Arrow.RecordBatch containing all columns, including the index columns. The schema for the values must match the schema for the ``SOMADataFrame``.
+        :param values: An Arrow.Table containing all columns, including the index columns. The schema for the values must match the schema for the ``SOMADataFrame``.
 
-        The ``values`` Arrow RecordBatch must contain a ``soma_rowid`` (uint64) column, indicating which rows are being written.
+        The ``values`` Arrow Table must contain a ``soma_rowid`` (uint64) column, indicating which rows are being written.
         """
         self._shape = None  # cache-invalidate
 
--- a/apis/python/src/tiledbsoma/soma_dataframe.py
+++ b/apis/python/src/tiledbsoma/soma_dataframe.py
@@ -169,9 +169,9 @@ class SOMADataFrame(TileDBArray):
         # TODO: batch_size
         # TODO: partition,
         # TODO: platform_config,
-    ) -> Iterator[pa.RecordBatch]:
+    ) -> Iterator[pa.Table]:
         """
-        Read a user-defined subset of data, addressed by the dataframe indexing column, optionally filtered, and return results as one or more ``Arrow.RecordBatch``.
+        Read a user-defined subset of data, addressed by the dataframe indexing column, optionally filtered, and return results as one or more ``Arrow.Table``.
 
         :param ids: Which rows to read. Defaults to ``None``, meaning no constraint -- all rows.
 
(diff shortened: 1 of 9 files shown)


apis/python/src/tiledbsoma/soma_dataframe.py
```diff
@@ -219,18 +219,16 @@ def read(
...
-                    # This is the 'decode on read' part of our logic; in dim_select we have the
-                    # 'encode on write' part.
-                    # Context: https://github.com/single-cell-data/TileDB-SOMA/issues/99.
-                    #
-                    # Also: don't materialize these on read
-                    # TODO: get the arrow syntax for drop
-                    # df.drop(ROWID, axis=1)
-                    yield util_arrow.ascii_to_unicode_pyarrow_readback(batch)
+            for table in iterator:
+                # XXX COMMENT MORE
```
johnkerl, line 223: This will go away entirely in another PR in this set of three, as noted in this PR's description field

apis/python/src/tiledbsoma/soma_indexed_dataframe.py
```diff
@@ -258,14 +258,12 @@ def read(
...
-            for df in iterator:
-                batches = df.to_batches()
-                for batch in batches:
-                    # XXX COMMENT MORE
-                    # This is the 'decode on read' part of our logic; in dim_select we have the
-                    # 'encode on write' part.
-                    # Context: # https://github.com/single-cell-data/TileDB-SOMA/issues/99.
-                    yield util_arrow.ascii_to_unicode_pyarrow_readback(batch)
+            for table in iterator:
+                # XXX COMMENT MORE
```
johnkerl, line 262: ditto

apis/python/src/tiledbsoma/util_arrow.py
```diff
@@ -121,36 +121,31 @@ def get_arrow_schema_from_tiledb_uri(
...
             )
             new_fields.append(nfield)
         else:
             new_fields.append(old_field)
-    return pa.RecordBatch.from_arrays(new_fields, names=names)
+    return pa.Table.from_arrays(new_fields, names=names)
 
 
-def concat_batches(batch_generator: Iterator[Any]) -> pa.RecordBatch:
+def concat_tables(table_generator: Iterator[Any]) -> pa.Table:
```
bkmartinjr, line 143: This routine is obsolete  and can be removed.  You can directly use pyarrow.concat_tables -- it accepts an iterator, so there is no benefit to creating a list to then hand to pyarrow.concat_tables.

In fact, there is a downside - this code requires that all tables be fully materialized prior to concatenation.

apis/python/src/tiledbsoma/util_arrow.py
```diff
@@ -121,36 +121,31 @@ def get_arrow_schema_from_tiledb_uri(
...
     Implements the 'decode on read' part of our ASCII/Unicode logic
     """
     # TODO: COMMENT/LINK HEAVILY
-    names = [ofield.name for ofield in record_batch.schema]
+    names = [ofield.name for ofield in table.schema]
     new_fields = []
     for name in names:
-        old_field = record_batch[name]
-        if isinstance(old_field, pa.LargeBinaryArray):
+        old_field = table[name]
```
bkmartinjr, line 132: Just confirming this is known to work with chunked arrays (which can happen for very large query results, ie, column size > 2GB)

apis/python/src/tiledbsoma/soma_sparse_nd_array.py
```diff
@@ -266,9 +264,9 @@ def read_as_pandas_all(
         row_ids: Optional[Sequence[int]] = None,
```
bkmartinjr, line 264: Just a heads up: this hard-wired 2D assumption is a bug (fixed in `main`).  Need to make sure this is merged correctly with main. 

I'd like to re-review post-merge.

bkmartinjr (dismissed):
Comments:
* SOMA*DataFrame - need unit tests for the full read/write API (eg, there are no Pandas unit tests for SOMADataFrame.read_as_pandas, etc).
* remove util_arrow.concat_tables and use pyarrow.concat_tables directly.  They are sematically equivalent, but the pyarrow implementation is potentially more efficient for very large queries, as it does not require the pre-alloc of all tables before concatenating.
Also, I'd like to re-review after main has been merged/reconcilied with this branch, as there are a lot of pending changes to the NdArray code.

apis/python/src/tiledbsoma/util_arrow.py
johnkerl: @bkmartinjr to be clear the _column_ is not the `pa.LargeBinaryArray` here but rather the _cell value_

apis/python/src/tiledbsoma/soma_sparse_nd_array.py
johnkerl: @bkmartinjr i indeed have recently merged PRs to rebase onto! :)

johnkerl:
Converting to draft while I rebase -- lots of great activity to merge in from the last day or two of commits :)

apis/python/src/tiledbsoma/util_arrow.py
johnkerl: also reminding this will go away entirely on #359

johnkerl: (that's how Unicode encoded as bytes reads back in Arrow unless we actively decode it -- hence #359 which will remove that need)

johnkerl:
rebased

johnkerl:
> SOMA*DataFrame - need unit tests for the full read/write API (eg, there are no Pandas unit tests for SOMADataFrame.read_as_pandas, etc).

@bkmartinjr 100% agreed but this PR (Arrow `RecordBatch` -> `Table`) isn't the right place for it.

johnkerl:
@bkmartinjr ready for round two :)

thetorpedodog (commented):
This is pretty straightforward, and everything looks good style-wise, so I leave it to others to judge the content of the change. LGTM but not going to hit the "approve" button so that it doesn't spuriously show up as submittable. Please bump me if you want any more feedback!

apis/python/src/tiledbsoma/soma_dataframe.py
```diff
@@ -169,9 +169,9 @@ def read(
         # TODO: batch_size
         # TODO: partition,
         # TODO: platform_config,
-    ) -> Iterator[pa.RecordBatch]:
+    ) -> Iterator[pa.Table]:
```
bkmartinjr, line 172: why not type the `ids` (aka coordinates) parameter to something more specific?  See for example SOMADenseNdArray.

I think you can literally use the `SOMADenseCoordinates` type (in types.py) as this is a dense dataframe, which only takes a scalar or slice.

bkmartinjr: Also, we need to decide on whether or not `ids` is optional.  The behavior is inconsistent between DataFrame and NdArray - in the former, None is "all", in the latter, None is not accepted (only the explicit `slice(None)` is accepted).

CC: @thetorpedodog - if you want to propose a standard convention, we can pull it through the entire package.

apis/python/src/tiledbsoma/soma_dataframe.py
```diff
@@ -217,18 +217,16 @@ def read(
...
-                    # df.drop(ROWID, axis=1)
-                    yield util_arrow.ascii_to_unicode_pyarrow_readback(batch)
+            for table in iterator:
+                # XXX COMMENT MORE
+                # This is the 'decode on read' part of our logic; in dim_select we have the
+                # 'encode on write' part.
+                # Context: https://github.com/single-cell-data/TileDB-SOMA/issues/99.
+                #
+                # Also: don't materialize these on read
+                # TODO: get the arrow syntax for drop
```
bkmartinjr, line 227: I'm not sure I understand this last `TODO` comment - we aren't dropping columns, but rather transcoding them to unicode.  Maybe clarify the comment?

apis/python/src/tiledbsoma/soma_dataframe.py
```diff
@@ -243,11 +241,11 @@ def read_all(
         # TODO: partition,
         # TODO: result_order,
         # TODO: platform_config,
-    ) -> pa.RecordBatch:
+    ) -> pa.Table:
         """
-        This is a convenience method around ``read``. It iterates the return value from ``read`` and returns a concatenation of all the record batches found. Its nominal use is to simply unit-test cases.
+        This is a convenience method around ``read``. It iterates the return value from ``read`` and returns a concatenation of all the table-pieces found. Its nominal use is to simply unit-test cases.
```
bkmartinjr, line 246: is this comment still valid?  The PyArrow documentation describes concat_tables, so I'm not sure we need to recapitulate it here.

Or perhaps reword to describe the intent, eg, "Concatenate all partial read results into a single Table"?

apis/python/src/tiledbsoma/soma_dataframe.py
```diff
@@ -243,11 +241,11 @@ def read_all(
         # TODO: partition,
         # TODO: result_order,
         # TODO: platform_config,
-    ) -> pa.RecordBatch:
+    ) -> pa.Table:
```
bkmartinjr, line 244: `ids` parameter needs a type.  See comment on `read`

apis/python/src/tiledbsoma/soma_dataframe.py
thetorpedodog: If we want to make it not an Optional value, we can easily preserve the "get everything if unspecified" behavior by doing

`ids: Union[Sequence[int], slice] = slice(None)`

I like this solution because it ends up with an explicit default with already-specified semantics.

apis/python/src/tiledbsoma/soma_dataframe.py
johnkerl: yeah it's old & i'll remove it

not sure how it's showing up as "new", must be a copypasta

apis/python/src/tiledbsoma/util_arrow.py
```diff
@@ -121,36 +121,20 @@ def get_arrow_schema_from_tiledb_uri(
...
+        if len(old_field) > 0 and isinstance(old_field[0], pa.LargeBinaryScalar):
```
bkmartinjr, line 133: I think the preferred pyarrow approach is to use the pyarrow.types helper functions rather than `isinstance`.  in this case, I believe that would be `pyarrow.types.is_large_binary`

apis/python/src/tiledbsoma/util_arrow.py
bkmartinjr: I believe you misunderstood my question.  A column in a table is ChunkedArray.  If there is more than one chunk in the table column, is it correctly handled?  

I believe I have convinced myself it is fine, as `pa.array` on line 134 will concatenate all of the chunks into a single `Array`.

bkmartinjr (commented):
comments:
* print's in the test_soma_indexed_dataframe.py that should be removed (eg, line 108-113)
* the two dataframe objects would benefit from having some tests that check if their write methods raise the expected error when the `value` param is unexpected (eg, not a Table, or does not match the DataFrame schema).

bkmartinjr (approved):
The conversion from RecordBatch -> Table looks fine.  Other refinement suggestions left inline which I believe would help the code.  I leave it to you if you want to fix in this PR, in another PR, or file bugs.

apis/python/src/tiledbsoma/soma_dataframe.py
johnkerl: @bkmartinjr @thetorpedodog thanks!

Optionality of `ids` is orthogonal to this PR which is solely `pa.RecordBatch` -> `pa.Table`, and has been open long enough, and rebased enough times -- I created https://github.com/single-cell-data/TileDB-SOMA/issues/374 to track that

apis/python/src/tiledbsoma/util_arrow.py
johnkerl: @bkmartinjr will do, but recall this is going away on the very next PR in this stack (#359)

apis/python/src/tiledbsoma/soma_dataframe.py
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/issues/374 also

aaronwolen (commented):
🚀

PR #356, opened 2022-10-02
add error check for zero length dimensions in ndarray

### LINKED ISSUE #297: Python tiledbsoma: assertion when creating zero-length SparseNdArray
Python tiledbsoma.SparseNdArray will not create an array with shape (0,0).  It is often convenient to create these as placeholders.  It currently asserts when requested to do so.

---

Digging into this more (see below), currently electing to raise a usable exception, and not support (0,) shaped dimensions.

bkmartinjr:
Fixes #297 

TileDB does not support zero-length domains for a dimension.  Add error check and raise a meaningful exception.


johnkerl (approved):
🚢

PR #359, opened 2022-10-03
Use true ASCII attributes in dataframes

johnkerl:
# What

This is a forward-port of #273 from the `main-old` branch to the `main` branch.

# Why

* Write string dims/attrs as "ascii" per se, not bytes and not str per @Shelnutt2 
* Please see also the clear presentation at https://github.com/single-cell-data/TileDB-SOMA/issues/99

# Lack of backward compatibility

* Since old SOMAs written before #273 didn't have all "ascii", on `main-old` (see #273) we conditionally retained the decode-on-readback logic which was essential since otherwise b"foo" != "foo"
* But on this `main` (greenfield) branch we do not keep backward compatibility -- any pre-existing datasets with non-ASCII attributes will need to be regenerated

# Prerequisites

* This depends on https://github.com/TileDB-Inc/TileDB-Py/pull/1337 -- now merged and in `TileDB-Py` 0.17.5 which is here as of #394
* This PR is stacked atop #355 -- now merged

# PR context

This is the second in a group of three related PRs:

* #355: Conform to https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md by having `read` return`pyarrow.Table` not `pyarrow.RecordBatch` (as in an outdated version of that spec)
* This PR: Foreward-port https://github.com/single-cell-data/TileDB-SOMA/pull/273 from `main-old` which will truly have ASCII columns, obviating the need for our `util_arrow.ascii_to_unicode_pyarrow_readback`
* #360: Drop in C++ acceleration-library support for `read` methods, which will go in cleanly now
  * The C++ code returns `pyarrow.Table` and with the first PR our unit tests will be ready to go
  * When the C++ code reads Unicode cells it returns them as `pyarrow.LargeBinaryArray` (needing decode) but when we are properly writing ASCII cells via the Python write path then the C++ code will read ASCII cells and return them as strings (no longer needing decoding)

These three changesets could be done in a single PR, but that would be unmerciful to the reviewers.

# Problem

Issue I'm seeing on this PR which confuses me:

Script:

```
#!/usr/bin/env python

import os, shutil
import pyarrow as pa
import tiledbsoma as t
import tiledb

uri = "foo"
if os.path.exists(uri):
    shutil.rmtree(uri)

schema = pa.schema(
    [
        ("soma_rowid", pa.uint64()),
        ("A", pa.int64()),
        ("B", pa.float64()),
        ("C", pa.string()),
    ]
)

sdf = t.SOMADataFrame(uri=uri)
sdf.create(schema=schema.remove(schema.get_field_index("soma_rowid")))

data = {
    "soma_rowid": [0, 1, 2, 3],
    "A": [10, 11, 12, 13],
    "B": [100.1, 200.2, 300.3, 400.4],
    "C": ["this", "is", "a", "test"],
}
n_data = len(data["soma_rowid"])
rb = pa.Table.from_pydict(data)
sdf.write(rb)
```

TileDB schema:

```
ArraySchema(
  domain=Domain(*[
    Dim(name='soma_rowid', domain=(0, 18446744073709551614), tile=2048, dtype='uint64', filters=FilterList([ZstdFilter(level=3), ])),
  ]),
  attrs=[
    Attr(name='A', dtype='int64', var=False, nullable=False, filters=FilterList([ZstdFilter(level=-1), ])),
    Attr(name='B', dtype='float64', var=False, nullable=False, filters=FilterList([ZstdFilter(level=-1), ])),
    Attr(name='C', dtype='ascii', var=True, nullable=False, filters=FilterList([ZstdFilter(level=-1), ])), <----- AS DESIRED
  ],
  cell_order='row-major',
  tile_order='row-major',
  capacity=100000,
  sparse=True,
  allows_duplicates=False,
)
```

Reading back from C++ or from Python:

```
>>> import tiledbsoma.libtiledbsoma as clib
>>> clib.SOMAReader('foo')
<tiledbsoma.libtiledbsoma.SOMAReader object at 0x7fe69b851070>
>>> sr=clib.SOMAReader('foo')
>>> sr.
sr.read_next(        sr.results_complete( sr.set_dim_points(   sr.set_dim_ranges(   sr.submit(
>>> sr.submit()
>>> o=sr.read_next()
>>> o
pyarrow.Table
soma_rowid: uint64
A: int64
B: double
C: large_string <------- this is why I am seeing unit-test failures -- same reason
----
soma_rowid: [[0,1,2,3]]
A: [[10,11,12,13]]
B: [[100.1,200.2,300.3,400.4]]
C: [["this","is","a","test"]]
```

In summary:

* TileDB arrays are created with `"ascii"` attributes
* When I read them back as Arrow using C++ or Python, I now get `large_string` not `string` in the Arrow readback schema

cc @gspowley and @nguyenv 

@nguyenv I'll get you a simpler reprex tomorrow

Update 2022-10-04: simpler repro https://gist.github.com/johnkerl/278bb246c3aa05cc4dfb1694503744c6


gspowley:
The C++ reader is always returning arrow type `large_string` for tiledb type `TILEDB_STRING_ASCII`.

https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/src/pyapi/arrow_adapter.h#L114-L116

johnkerl:
> The C++ reader is always returning arrow type `large_string` for tiledb type `TILEDB_STRING_ASCII`.
> 
> https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/src/pyapi/arrow_adapter.h#L114-L116

cc @bkmartinjr

apis/python/src/tiledbsoma/soma_dataframe.py
```diff
@@ -441,39 +432,13 @@ def write_from_pandas(
...
-        # IMPLEMENTATION:
-        # Python types -- float, string, what have you -- appear as dtype('O') which is not useful.
-        # Also, ``tiledb.from_pandas`` has ``column_types`` but that _forces_ things to string to a
-        # particular if they shouldn't be.
-        #
-        # Instead, we use ``dataframe.convert_dtypes`` to get a little jump on what ``tiledb.from_pandas``
-        # is going to be doing anyway, namely, type-inferring to see what is going to be a string.
-        #
-        # TODO: when UTF-8 attributes are queryable using TileDB-Py's QueryCondition API we can remove this.
+        # Force ASCII storage if string, in order to make obs/var columns queryable.
```
aaronwolen, line 435: and improve interop with tiledbsoma-r?

johnkerl:
fyi, now that python unit-test cases are passing, there are C++ unit-test failures which i am currently debugging

johnkerl:
> fyi, now that python unit-test cases are passing, there are C++ unit-test failures which i am currently debugging

fixed on latest commit:

https://github.com/single-cell-data/TileDB-SOMA/pull/359/commits/9bdd02f9dfeadf740360ae8b55484825cf3a7b95

johnkerl:
Note tests will fail with

```
E   TypeError: dtype is not compatible with var-length attribute
```

until https://github.com/TileDB-Inc/TileDB-Py/pull/1337 wends its way into a `TileDB-Py` release we can update to -- cc @nguyenv and @ihnorton

gspowley (approved):

PR #370, opened 2022-10-04
Work around recent Mac GHA issue

johnkerl:
As detailed at
https://github.com/actions/runner-images/issues/6350

There don't appear (in my googling so far anyway) to be good options besides 'downgrade' :disappointed: so this is my best attempt at present


bkmartinjr (approved):

bkmartinjr:
@johnkerl - don't you also need to make the same change in `ci.yml`?  I'm getting errors on the PR CI checks.

johnkerl:
@bkmartinjr i had all green ... 👀

PR #371, opened 2022-10-04
correctly handle empty query results in SparseNdArray read operations

### LINKED ISSUE #373: sparse array read may raise error
SOMASparseNdArray.read_sparse_tensor, with format="coo" may raise an error if the underlying TileDB query returns an emtpy Table. This is due to a bug in Arrow SparseCOOTensor.  Details: https://issues.apache.org/jira/browse/ARROW-17933

Format "csr" and "csc" work fine, as does read_table().

Add a work-around and unit test.

---
Example:
```
In [40]: a = soma.SOMASparseNdArray("./foobar")

In [41]: a.create(type=pa.float32(), shape=(1,1))

In [42]: for i in a.read_sparse_tensor((slice(None)), format="coo"): print(i)
ArrowInvalid                              Traceback (most recent call last)
Cell In [42], line 1
----> 1 for i in a.read_sparse_tensor((slice(None)), format="coo"): print(i)

File ~/projects/TileDB-SOMA/apis/python/src/tiledbsoma/soma_sparse_nd_array.py:224, in SOMASparseNdArray.read_sparse_tensor(self, coords, format)
    217     coo_data = arrow_tbl.column("data").to_numpy()
    218     coo_coords = np.array(
    219         [
    220             arrow_tbl.column(f"__dim_{n}").to_numpy()
    221             for n in range(self.ndims)
    222         ]
    223     ).T
--> 224     yield pa.SparseCOOTensor.from_numpy(
    225         coo_data, coo_coords, shape=A.shape
    226     )
    228 elif format in ("csr", "csc"):
    229     # Temporary: as these must be 2D, convert to scipy COO and use
    230     # scipy to perform conversions.  C++ reader will be nicer!
    231     data = arrow_tbl.column("data").to_numpy()

File ~/projects/TileDB-SOMA/venv/lib/python3.10/site-packages/pyarrow/tensor.pxi:354, in pyarrow.lib.SparseCOOTensor.from_numpy()

File ~/projects/TileDB-SOMA/venv/lib/python3.10/site-packages/pyarrow/error.pxi:100, in pyarrow.lib.check_status()

ArrowInvalid: SparseCOOIndex indices must be contiguous
```

Versus:
```
In [43]: for i in a.read_sparse_tensor((slice(None)), format="csr"): print(i)
<pyarrow.SparseCSRMatrix>
type: float
shape: (1, 1)
```

bkmartinjr:
The current read_sparse_tensor code returns an empty sparse tensor when the query result is empty. This works fine for all tensor types except SparseCOOTensor, which appears to have a [bug](https://issues.apache.org/jira/browse/ARROW-17933).

Because the read_sparse_tensor returns an iterator of tensors, the behavior in this edge case is modified to return StopIteration, rather than an empty tensor.

Added unit tests as well.

Fixes #373


apis/python/src/tiledbsoma/soma_sparse_nd_array.py
```diff
@@ -193,7 +193,19 @@ def read_sparse_tensor(
                 return_incomplete=True,
             )
             for arrow_tbl in query.df[coords]:
+                """
+                In PyArrow 9.0.0, there is a bug preventing the creation of "empty" (zero element)
```
johnkerl, line 197: nit: zero element -> zero-in-all-extents

johnkerl (approved):

aaronwolen (approved):

PR #381, opened 2022-10-07
Python collections refinement

### LINKED ISSUE #375: python tiledbsoma:  `.name` and repr issue
The python __repr__ implementation gets confused about the element key name. Digging into this, it is due to some code remaining from the prototype where each soma object had a "true name" stored in its `name` property.

example issue - note that `__repr__` return the incorrect name in the first print:

```
In [96]: c = soma.SOMACollection(uri="./collection")
    ...: c.create()
    ...: d = soma.SOMADataFrame(uri="./collection/df")
    ...: d.create(pa.schema([("A", pa.int32()), ("B", pa.string())]))
    ...: c["different_name"] = d
    ...: 
    ...: print(list(c.keys()))
    ...: print(c.__repr__())
    ...: del c
    ...: 
    ...: # re-open
    ...: c = soma.SOMACollection(uri="./collection")
    ...: print(c.__repr__())
    ...: 
    ...: 
collection SOMACollection:
    df SOMADataFrame
collection SOMACollection:
    different_name SOMADataFrame
```

The core issue is some legacy code from the prototype. Each soma object has a `name` property (in addition to `uri`), which should be removed.  Only the collection has a key/name for each otherwise unnamed object.

Removing the `name` property will also align the collection better with TileDB group, and simplify the repr.

### LINKED ISSUE #376: Python: SOMACollection.set will clobber URI if called with `relative=True`
Test case:
```
c = soma.SOMACollection(uri="./collection")
c.create()
d = soma.SOMACollection(uri="./collection/child_collection")
d.create()
c.set(d, child_name="TEST", relative=True)
```

Expected behavior: the child object (uri=='./collection/child_collection') would be added to the top-level collection with key 'TEST'

Actual behavior is that the child URI is clobbered with the name and it raises an exception.

The root cause is very closely related to issue #375 

```
TileDBError                               Traceback (most recent call last)
Cell In [4], line 5
      3 d = soma.SOMACollection(uri="./collection/child_collection")
      4 d.create()
----> 5 c.set(d, child_name="TEST", relative=True)

File ~/projects/TileDB-SOMA/apis/python/src/tiledbsoma/soma_collection.py:116, in SOMACollectionBase.set(self, member, child_name, relative)
    111 """
    112 Adds a member to the collection.  This interface allows explicit control over
    113 `relative` URI, and uses the member's default name.
    114 """
    115 name = member.name if child_name is None else child_name
--> 116 self._add_object(member, child_name=name, relative=relative)
    117 self._cached_members[name] = member

File ~/projects/TileDB-SOMA/apis/python/src/tiledbsoma/soma_collection.py:305, in SOMACollectionBase._add_object(self, obj, child_name, relative)
    303 self._cached_member_names_to_uris = None  # invalidate on add-member
    304 with self._tiledb_open("w") as G:
--> 305     G.add(uri=child_uri, relative=relative, name=child_name)
    306 # See _get_child_uri. Key point is that, on TileDB Cloud, URIs change from pre-creation to
    307 # post-creation. Example:
    308 # * Upload to pre-creation URI tiledb://namespace/s3://bucket/something/something/somaname
   (...)
    311 # * Member pre-creation URI tiledb://namespace/s3://bucket/something/something/somaname/obs
    312 # * Member post-creation URI tiledb://somaname/e4de581a-1353-4150-b1f4-6ed12548e497
    313 obj._uri = self._get_child_uri(child_name)

File ~/projects/TileDB-SOMA/venv/lib/python3.10/site-packages/tiledb/group.py:303, in Group.add(self, uri, name, relative)
    292 """
    293 Adds a member to the Group.
    294 
   (...)
    300 :type name: str
    301 """
    302 if name:
--> 303     self._add(uri, relative, name)
    304 else:
    305     self._add(uri, relative)

TileDBError: [TileDB::Group] Error: Cannot add group member file:///home/bruce/projects/TileDB-SOMA/apis/python/collection/TEST, type is INVALID. The member likely does not exist.
```

bkmartinjr:
Fixes #376 
Fixes #375 

Changes:
* remove `.name` and `._nested_name` properties from the SOMA  Python classes.  This resolves #375
* refine and simplify SOMACollection, expand unit tests. Of note:
    * the class no longer relies on the `.name` property, and all set/__setitem__ calls must specify a key
    * the class will no longer auto-create objects with a URI and type guess. Instead, both the collection and any values inserted into the collection must be created before inserted into the collection, matching the TileDB group semantics.
    * a sub-class protocol was added allowing SOMAExpression & SOMAMesurement to constrain the accepted types that may be assigned to certain keys (eg, `obs`, `ms`, etc).
    * fix SOMACollection `relative=True` handling to not clobber the user-provided URI. Instead it will determine, if possible, the correct relative URI to add to the TileDB group or raise an error.  This resolves #376.  Added test cases to validate relative=True is working correctly
    * update the __repr__ protocol to not rely on `.name`
* simplify imports throughout
* add additional explicit typing
* add very skinny first cut at a common Exception/Error class hierarchy and use to propagate some errors.  More work needed here.


apis/python/src/tiledbsoma/factory.py
```diff
@@ -7,20 +7,19 @@
...
+from .soma_measurement import SOMAMeasurement
+from .soma_sparse_nd_array import SOMASparseNdArray
 from .util import SOMA_OBJECT_TYPE_METADATA_KEY
 
-MemberType = Union[
+SOMAObjectTypes = Union[
     SOMAExperiment,
     SOMAMeasurement,
-    SOMACollectionBase,
+    SOMACollection,
```
johnkerl, line 22: duplicate entry?

apis/python/src/tiledbsoma/io.py
```diff
@@ -221,10 +228,8 @@ def from_anndata(
     #        experiment.set(experiment.uns)
 
     logging.log_io(
-        f"Wrote {experiment._nested_name}",
-        util.format_elapsed(
-            s, f"{experiment._indent}FINISH WRITING {experiment._nested_name}"
-        ),
+        "Wrote",
```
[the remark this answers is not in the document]
johnkerl: what happens is:

```
Already exists; replacing: ./tiledbsoma-data/pbmc-small
Wrote
Wrote
Wrote
Wrote
Wrote
Wrote
Wrote
Wrote
Wrote
Wrote
Wrote ./tiledbsoma-data/pbmc-small
```

johnkerl (changes requested):
@bkmartinjr this is fantastic -- a big step forward in moving this from what started as a sketch and what is rapidly becoming a more mature and easier-to-maintain package

my only ask (beyond nits) is to fix the `Wrote` statements to not omit a description of what was written (details inline)

johnkerl:
Another issue: uploads to tiledb-cloud URIs are not working with this PR. Very soon I'll develop a simple repro as well as a suggested fix.

apis/python/src/tiledbsoma/io.py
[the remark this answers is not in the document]
bkmartinjr: This was removed because it was logging the (now non-existent) `.name` property`.  I will replace it with the Experiment `uri`, as that is the only analog available in this route.

bkmartinjr:
> Another issue: uploads to tiledb-cloud URIs are not working with this PR. Very soon I'll develop a simple repro as well as a suggested fix.

@johnkerl - if the fix is easy, we should include it in the PR.  Did not intentionally break, but lack a good means of testing at this point in tie.

apis/python/src/tiledbsoma/factory.py
```diff
@@ -50,38 +50,47 @@ def _construct_member(
 
     # Get the class name from TileDB storage. At the TileDB level there are just "arrays" and
     # "groups", with separate metadata-getters.
-    class_name = None
-    object_type = tiledb.object_type(member_uri, ctx=ctx)
     if object_type is None:
+        object_type = tiledb.object_type(member_uri, ctx=ctx)
+
+    # auto-detect class name from meta-data
```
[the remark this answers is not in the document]
bkmartinjr: fixed

apis/python/src/tiledbsoma/factory.py
bkmartinjr: fixed - good catch!

johnkerl:
Clarify: by
> Very soon I'll develop a simple repro as well as a suggested fix.
I meant:
> Very soon I'll develop a simple repro as well as a suggested fix on this PR.

apis/python/src/tiledbsoma/soma_dense_nd_array.py
```diff
@@ -243,8 +227,8 @@ def from_matrix(self, matrix: Matrix) -> None:
         self._common_create()  # object-type metadata etc
 
         logging.log_io(
-            f"Wrote {self._nested_name}",
-            util.format_elapsed(s, f"{self._indent}FINISH WRITING {self._nested_name}"),
+            f"Wrote {self.uri}",
+            util.format_elapsed(s, f"{self._indent}FINISH WRITING"),
```
johnkerl, line 231: `self.uri` here too please

apis/python/src/tiledbsoma/soma_sparse_nd_array.py
```diff
@@ -323,12 +311,10 @@ def from_matrix(self, matrix: Matrix) -> None:
         # https://docs.scipy.org/doc/scipy/reference/sparse.html
 
         s = util.get_start_stamp()
-        logging.log_io(None, f"{self._indent}START  WRITING {self._nested_name}")
+        logging.log_io(None, f"{self._indent}START  WRITING")
```
johnkerl, line 314: `self.uri` here too please

apis/python/src/tiledbsoma/soma_sparse_nd_array.py
```diff
@@ -348,8 +334,8 @@ def from_matrix(self, matrix: Matrix) -> None:
         self._common_create()  # object-type metadata etc
 
         logging.log_io(
-            f"Wrote {self._nested_name}",
-            util.format_elapsed(s, f"{self._indent}FINISH WRITING {self._nested_name}"),
+            f"Wrote {self.uri}",
+            util.format_elapsed(s, f"{self._indent}FINISH WRITING"),
```
johnkerl, line 338: `self.uri` here too please

apis/python/src/tiledbsoma/io.py
```diff
@@ -111,108 +114,114 @@ def from_anndata(
     )
 
     s = util.get_start_stamp()
-    logging.log_io(
-        None, f"{experiment._indent}START  WRITING {experiment._nested_name}"
-    )
+    logging.log_io(None, f"{experiment._indent}START  WRITING")
```
johnkerl, line 117: `self.uri` here too please

apis/python/src/tiledbsoma/io.py
```diff
@@ -221,10 +230,8 @@ def from_anndata(
     #        experiment.set(experiment.uns)
 
     logging.log_io(
-        f"Wrote {experiment._nested_name}",
-        util.format_elapsed(
-            s, f"{experiment._indent}FINISH WRITING {experiment._nested_name}"
-        ),
+        f"Wrote {experiment.uri}",
+        util.format_elapsed(s, f"{experiment._indent}FINISH WRITING"),
```
johnkerl, line 234: `self.uri` here too please

johnkerl (changes requested):
found more spots where what-was-written context was being omitted

apis/python/src/tiledbsoma/soma_collection.py
```diff
@@ -270,7 +270,8 @@ def _determine_default_relative(self, uri: str) -> Optional[bool]:
         if self._tiledb_platform_config.member_uris_are_relative is not None:
             return self._tiledb_platform_config.member_uris_are_relative
         if uri.startswith("tiledb://"):
-            return True
+            # TileDB-Cloud does not use relative URIs, ever.
+            return False
         return None
```
johnkerl, line 275: why not `return True` here? & i think this should return `bool` not `Optional[bool]`

apis/python/src/tiledbsoma/util.py
```diff
@@ -32,3 +34,61 @@ def is_local_path(path: str) -> bool:
...
+    """
+    Join a path to a URI.
+
+    Supports relative paths for `file` or unspecified schemes, assuming
+    they are file system paths.  Assumes NO suport for relative paths
+    otherwise.
+    """
+    p_base = urllib.parse.urlparse(base)
+    parts = [*p_base]
+    print(parts)
```
johnkerl, line 73: debug-print to be removed

johnkerl: debug-print to be removed

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -74,24 +61,25 @@ def delete(self) -> None:
...
         """
-        Fallback string display. Will be overridden by any interesting subclasses.
+        Default repr
         """
-        return f"name={self._name},uri={self._uri}"
-
-    def _repr_aux(self) -> Sequence[str]:
-        raise Exception("Must be overridden by inherting classes.")
+        if self.exists():
+            return f"{self.type}(uri={self._uri})"
```
johnkerl, line 67: suggested:

```diff
-            return f"{self.type}(uri={self._uri})"
+            return f'{self.type}(uri="{self._uri}")'
```

johnkerl: this matches the key being quoted, & jives with the idea that a `repr` of an object should give some notion of how to construct the object

apis/python/src/tiledbsoma/soma_collection.py
bkmartinjr: It must be able to also return a None, which is not type bool (see line 275).

johnkerl (changes requested):
looks great including tiledb-cloud ingest testing well :)

just a couple more final changes requested

apis/python/src/tiledbsoma/util.py
bkmartinjr: fixed

apis/python/src/tiledbsoma/tiledb_object.py
bkmartinjr: I don't believe there is any such requirement on repr (and most reprs don't follow this suggestion as it is overly verbose).  I can provide you with numerous examples from widely used packages if you desire (tiledb, numpy, pandas, etc).

However, in the name of landing this thing, I'll push this change momentarily.

johnkerl (approved):
🚢 🚢 🚢

PR #388, opened 2022-10-07
[r] Support build against external artifact and system library

eddelbuettel:
This PR generalizes the build setup for the R package to use _either_ a system-wide installation of TileDB Core for the C++ headers and library, and a dowloaded artifact.  

This (loosely) follows the behavior of the `tiledb` R package on CRAN.  

It has been tested with system-wide TileDB and without (in Docker and on CI).  Tests in CI for macOS are not yet feasible as that setup appears to lack sufficient C++20 support.  Actual CI support may be added in a subsequent PR.

### DIFF
--- a/libtiledbsoma/cmake/Modules/FindTileDB_EP.cmake
+++ b/libtiledbsoma/cmake/Modules/FindTileDB_EP.cmake
@@ -48,6 +48,7 @@ else()
     endif()
 
     # Try to download prebuilt artifacts unless the user specifies to build from source
+    # NB When updating the pinned URLs here, please also update in file apis/r/tools/get_tarball.R
     if(DOWNLOAD_TILEDB_PREBUILT)
         if (WIN32) # Windows
           SET(DOWNLOAD_URL "https://github.com/TileDB-Inc/TileDB/releases/download/2.11.1/tiledb-windows-x86_64-2.11.1-15a1161.zip")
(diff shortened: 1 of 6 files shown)


eddelbuettel:
(Rebased and force-pushed)

aaronwolen (approved):
Nice! I verified it built locally on my mac. If @Shelnutt2 is okay with pulling in the latest core release we're okay to merge.

johnkerl (approved):

libtiledbsoma/cmake/Modules/FindTileDB_EP.cmake
```diff
@@ -48,6 +48,7 @@ else()
     endif()
 
     # Try to download prebuilt artifacts unless the user specifies to build from source
+    # NB When updating the pinned URLs here, please also update in file apis/r/tools/get_tarball.R
```
gspowley, line 51: This script updates the TileDB version for C++:
https://github.com/single-cell-data/TileDB-SOMA/blob/main/scripts/update-tiledb-version.py

Let's update the script to keep R in sync (in this or another PR is fine).

gspowley (approved):

libtiledbsoma/cmake/Modules/FindTileDB_EP.cmake
eddelbuettel: That's likely the cleanest. Marginal vote for doing it in a new and separate PR though.

gspowley: Yes, update the script in a new PR.

PR #360, opened 2022-10-03
Connect libtiledbsoma to tiledbsoma readers for dataframes [WIP]

johnkerl:
# Status

* Currently WIP
* `SOMADataFrame` and `SOMAIndexedDataFrame` are on this PR
* For reviewer mercy, `SOMASparseNdArray` and `SOMADenseNdArray` will be on a separate PR

# PR context

This is the three in a group of three related PRs:

* #355: Conform to https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md by having `read` return`pyarrow.Table` not `pyarrow.RecordBatch` (as in an outdated version of that spec) -- now merged
* #359: Foreward-port https://github.com/single-cell-data/TileDB-SOMA/pull/273 from `main-old` which will truly have ASCII columns, obviating the need for our `util_arrow.ascii_to_unicode_pyarrow_readback` -- now merged
* This PR: drop in C++ acceleration-library support for `read` methods, which will go in cleanly now
  * The C++ code returns `pyarrow.Table` and with the first PR our unit tests will be ready to go
  * When the C++ code reads Unicode cells it returns them as `pyarrow.LargeBinaryArray` (needing decode) but when we are properly writing ASCII cells via the Python write path then the C++ code will read ASCII cells and return them as strings (no longer needing decoding)

PR #400, opened 2022-10-12
Connect libtiledbsoma to tiledbsoma readers for dataframes (#360 re-do) [WIP]

johnkerl:
Re-does #360 on top of #397. This PR is significantly smaller since there was some code overlap with #397.

However, alas, the `typeguard` errors

```
E               RuntimeError: Static type (UINT64) does not match expected type (INT64)
```

persist.


apis/python/src/tiledbsoma/util.py
```diff
@@ -91,3 +91,19 @@ def uri_joinpath(base: str, path: str) -> str:
                 parts[2] = parts[2] + "/" + path
 
     return urllib.parse.urlunparse(parts)
+
+
+def slice_to_list(sl: slice) -> List[int]:
+    """
+    TODO: COMMENT GOES HERE
```
bkmartinjr, line 98: Please do the todo before landing :-)

johnkerl: i'll do the todo before taking it out of draft ;)

#1 focus is fixing the CI error about int64 vs uint64; then neatening after that

johnkerl:
@gspowley @bkmartinjr @Shelnutt2 everything seems working now, especially after some nice C++ mods which were included in #397 which was merged last evening.

The remaining fail is in Windows CI, whereat we do not build `libtiledbsoma` for Windows. And (AFAICT) the early-abort in MacOS CI is just because Windows CI failed (I think but am not certain).

Now that we have a hard dependency on `libtiledbsoma`, and since we've made the decision to remove [these lines](https://github.com/single-cell-data/TileDB-SOMA/blob/0.5.0a1/apis/python/src/tiledbsoma/soma_dataframe.py#L188-L231) which do work on Windows -- it seems we now have the decision to formally drop support for Windows, including taking it out of the CI job.

Thoughts?

gspowley:
> it seems we now have the decision to formally drop support for Windows, including taking it out of the CI job.
> 
> Thoughts?

I vote we delay supporting Windows. I believe it's more important to focus on developing SOMA features on Linux and MacOS now. We can revisit supporting Windows later.

bkmartinjr:
CC @ambrosejcarr

apis/python/src/tiledbsoma/soma_dataframe.py
```diff
@@ -150,40 +154,32 @@ def read(
...
-                query = A.query(
-                    return_arrow=True,
-                    return_incomplete=True,
-                    attr_cond=qc,
-                    order=tiledb_result_order,
-                    dims=dim_names,
-                    attrs=attr_names,
-                )
+            if ids is not None:
+                sr.set_dim_points(SOMA_ROWID, util.ids_to_list(ids))
```
gspowley, line 174: If `ids` is an Arrow array, we should pass the Arrow array to `set_dim_points` instead of converting it to a list. This will reduce memory usage and improve performance by avoiding creating a copy.

apis/python/src/tiledbsoma/soma_indexed_dataframe.py
```diff
@@ -182,41 +186,33 @@ def read(
...
-
-            if ids is None:
-                iterator = query.df[:]
-            else:
-                iterator = query.df[ids]
 
-            for table in iterator:
-                yield table
+            if ids is not None:
+                sr.set_dim_points(A.schema.domain.dim(0).name, util.ids_to_list(ids))
```
gspowley, line 207: Same comment as above.

gspowley (changes requested):

apis/python/src/tiledbsoma/util.py
```diff
@@ -91,3 +92,28 @@ def uri_joinpath(base: str, path: str) -> str:
...
 
     return urllib.parse.urlunparse(parts)
+
+
+# TODO: handle slices like `1:` or `:2`
+def ids_to_list(ids: Union[slice, List[int], pa.Array]) -> pa.ChunkedArray:
+    """
+    For the interface between ``SOMADataFrame::read`` et al. (Python) and ``SOMAReader`` (C++): the
+    ``ids`` argument to the former can be slice or list; the argument to
+    ``SOMAReader::set_dim_points`` must be a list.
```
gspowley, line 102: When setting a slice for a `SOMAReader` query, we should use `SOMAReader::set_dim_ranges` instead of converting the slice to a list.

This test shows an example:
https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/test/test_soma_reader.py#L82

apis/python/src/tiledbsoma/util.py
```diff
@@ -91,3 +92,28 @@ def uri_joinpath(base: str, path: str) -> str:
...
+            return pa.chunked_array(pa.array([]))
+        step = ids.step
+        if step is None:
+            if ids.start <= ids.stop:
+                step = 1
+            else:
+                step = -1
+        stop = ids.stop + step
+        return pa.chunked_array(pa.array(list(range(ids.start, stop, step))))
+    if isinstance(ids, pa.Array):
```
gspowley, line 117: The intention of supporting Arrow arrays is captured in this test (currently in a PR):
https://github.com/single-cell-data/TileDB-SOMA/blob/gspowley/obs-slice-x-test/libtiledbsoma/test/test_soma_reader.py#L162

In this test, the "ids" will be type `pa.ChunkedArray`, so it will fall through the `if isintance(...)` checks and raise an exception.

apis/python/src/tiledbsoma/util.py
```diff
@@ -91,3 +92,28 @@ def uri_joinpath(base: str, path: str) -> str:
...
+
+# TODO: handle slices like `1:` or `:2`
+def ids_to_list(ids: Union[slice, List[int], pa.Array]) -> pa.ChunkedArray:
+    """
+    For the interface between ``SOMADataFrame::read`` et al. (Python) and ``SOMAReader`` (C++): the
+    ``ids`` argument to the former can be slice or list; the argument to
+    ``SOMAReader::set_dim_points`` must be a list.
+    """
+    if isinstance(ids, list):
+        return pa.chunked_array(pa.array(ids))
```
gspowley, line 105: We don't need to convert a list to an Arrow array, it can remain a list.

gspowley (changes requested):
Let's sync up on Monday.

gspowley (approved):
Looks great!

Issue #353, opened 2022-09-30
[c++]  Should be bring in the std::span alternate to only require C++17 ?


eddelbuettel:
The R package happily builds as seen in #352.  By relying in `std::span` we are bring in C++20 which is not an issue _per se_ but when I set up some initial continuous integration exploration the need for proper C++20 meant that only ubuntu-22.04 was viable, not the general default of ubuntu-latest (currently still 20.04).

I don't have strong feelings here: C++20 is viable, and we don't _have to_ change.  But it worth mentioning as TileDB Embedded has similar build requirements and we know we get artifacts on platforms we care about by sticking to C++17.  So should we too here?

For reference, the alternate implementation compatible with C++17 is for example [here in TileDB](https://github.com/TileDB-Inc/TileDB/blob/dev/external/include/span/span.hpp) and consists of a single header.

PR #403, opened 2022-10-13
[r] Add Continuous Integration for the R package

eddelbuettel:
This (draft) PR adds continuous integration via GitHub Actions.  It is (at least for now) intentionally limited to Linux as the standard R setup on macOS cannot compile under C++20 (see #353 on that), and it also skips Windows.

For now two existing test files are skipped to expediate ensuring the overall mechanics of gettings dependencies including TileDB artifacts as well as source from `libtiledbsoma` setup to work.  The PR is marked as draft until we know what is going on with the two other tests which work for me locally. 

The script used for CI setup is borrowed from both tiledb-r and tiledbcloud-r.

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -28,7 +28,7 @@ jobs:
         run: mkdir ~/.R && echo "CXX20FLAGS=-Wno-deprecated-declarations -Wno-deprecated" > ~/.R/Makevars
 
       - name: BuildTgz
-        run: cd apis/r && R CMD build .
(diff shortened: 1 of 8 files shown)


johnkerl (approved):
🚀

eddelbuettel:
No worries, and "help welcome" but we just wait for @aaronwolen to come back.  No idea why two existing tests fall over.

But yeppers, good to have this in place. I just pushed one more micro-tweak to not run on each commit outside of `main`.

gspowley (approved):

PR #404, opened 2022-10-13
[c++] Add fast nnz

### LINKED ISSUE #383: investigate "fast" get nnz operation in C++ reader
Explore feasibility to use fragment cell_num to implement a fast `get stored element` count (aka get nnz) using fragment cell_num metadata.

Ideas discussed:
* if one fragment, there are no dups, so use cell_num
* if multiple, non-overlapping fragments, sum fragment cell_num
* if overlapping fragments, read/count the overlapping fragments

ideally this would be commonly faster than "read all and count", and could be exposed as a common API in the C++ layer.

gspowley:
Add fast nnz to C++ API, as discussed in #383:
* if one fragment, there are no dups, so use cell_num
* if multiple, non-overlapping fragments, sum fragment cell_num
* if overlapping fragments, read/count the overlapping fragments

Closes #383.

### DIFF
--- a/libtiledbsoma/include/tiledbsoma/managed_query.h
+++ b/libtiledbsoma/include/tiledbsoma/managed_query.h
@@ -230,6 +230,15 @@ class ManagedQuery {
      */
     std::shared_ptr<ArrayBuffers> results();
 
+    /**
+     * @brief Get the schema of the array.
+     *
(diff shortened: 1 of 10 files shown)


eddelbuettel (approved):
LGTM!

johnkerl (approved):

Issue #405, opened 2022-10-13
python tiledbsoma:  schema attribute returns incorrect type on SOMADataFrame


bkmartinjr:
when a data frame is created with a given schema, the spec says that the create should create a dataframe with the precise types specified, or generate an error if types are not supported.  The only exception is the allowed promotion of binary to large_binary and string to large_string.

Currently, all `string` and `large_string` are silently converted to `large_binary`.

test case:
```
import pyarrow as pa
import tiledbsoma as soma

sdf = soma.SOMADataFrame("./test")
sdf.create(pa.schema([
  ('string', pa.string()),
  ('binary', pa.binary()),
  ('large_binary', pa.large_binary()),
  ('large_string', pa.large_string()),
  ('int64', pa.int64())
]))

print(sdf.schema)
```

When I run it on the latest `main` branch, it prints:
```
soma_rowid: int64
soma_joinid: int64
string: large_binary
binary: large_binary
large_binary: large_binary
large_string: large_binary
int64: int64
```

johnkerl:
#406 is merged to `main`

PR #406, opened 2022-10-13
Disambiguate bytes/ascii on tiledb schema -> arrow schema

johnkerl:
Context: #405

Fantastic unit-testing was added by @bkmartinjr on #407 which was merged into this PR.

Issue (for the record): `bytes` and `ascii` both read back with `np.dtype("S") == "bytes"`. We need to use either `attr._tiledb_dtype` or `attr.isascii` when reading TileDB schema, to determine which Arrow schema type (`large_binary` or `large_string`) to report.


bkmartinjr:
I have confirmed this unblocks the census build script.

bkmartinjr (approved):
LGTM

PR #407, opened 2022-10-14
update type system test to only use public SOMA API

bkmartinjr:
Improve unit test for type system round-trip.  Now uses only the public SOMA API to minimize assumptions about internal package.


johnkerl (approved):
has a CI fail with
```
KeyError: 'soco'
```
but i'll merge this into my branch & continue debug from there

thanks @bkmartinjr this is good stuff!!!!! :)

Issue #396, opened 2022-10-11
Rename __dim and data on SOMASparseNdArray


bkmartinjr:
The latest abstract spec normalizes all builtin names in ndaarray.
* __dim_N becomes soma_dim_N
* data becomes soma_data

Need to update the Python and R API

- [x] Python, implemented in #397
- [x] R #411

PR #411, opened 2022-10-14
Match SparseNdArray dim/attr names to spec

johnkerl:
For #396


bkmartinjr (approved):
Thank you for doing this!

PR #354, opened 2022-10-02
Fix MacOS build with local TileDB core

johnkerl:
Up until now:

* On Linux, without core in `/usr/local`, builds fine
* On Mac, without core in `/usr/local`, builds fine
* On Linux, with core in `/usr/local`, builds fine
* On Mac, with core in `/usr/local`, build fails with
```
/Users/johnkerl/git/single-cell-data/TileDB-SOMA/libtiledbsoma/src/pyapi/../../include/tiledbsoma/logger_public.h:41:10: fatal error: 'spdlog/spdlog.h' file not found
#include <spdlog/spdlog.h>
         ^~~~~~~~~~~~~~~~~
1 error generated.
```

This is a crucial ability for any developers on MacOS (which is several of us).

Fixed after long debug with the incomparable @ihnorton :D

PR #413, opened 2022-10-15
Use MacOS 12 for CI

johnkerl:
On #354 we got MacOS CI working in the presence of an XCode 14 linker bug by downgrading to MacOS 11. As discussed at https://github.com/actions/runner-images/issues/6350, we can stay on MacOS 12 but force XCode 13.4 (non-buggy).

Also, it's not a bad idea or a bad time to use Python 3.9 for CI (we have been using 3.8, which is now in a 'security fixes only' phase).

Note that if/when this PR is accepted, I'll switch the required builds in the branch-protection rules for this repo to require 3.9 passes rather than 3.8 (as now).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,22 +9,30 @@ on:
 
 jobs:
   build:
-
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04, macos-11, windows-2019]
-        python-version: ['3.8']
+        # TODO: restore Windows build once we have C++/libtiledbsoma integration supported there
+        os: [ubuntu-22.04, macos-12]
+        # os: [ubuntu-22.04, macos-12, windows-2019]
+        python-version: ['3.8', '3.9', '3.10']
         include:
           - runs-on: ubuntu-22.04
             cc: gcc-11
             cxx: g++-11
-          - runs-on: macos-11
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -9,22 +9,30 @@ on:
...
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04, macos-11, windows-2019]
-        python-version: ['3.8']
+        # TODO: restore Windows build once we have C++/libtiledbsoma integration supported there
+        os: [ubuntu-22.04, macos-12]
+        # os: [ubuntu-22.04, macos-12, windows-2019]
+        python-version: ['3.8', '3.9', '3.10']
```
gspowley, line 19: Based on the PR description, do we want `3.9` only here?

.github/workflows/cpp-ci.yml
```diff
@@ -11,19 +11,26 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04, macos-11]
+        os: [ubuntu-22.04, macos-12]
         python-version: ['3.8']
```
gspowley, line 15: Do we want `3.9` here too?

gspowley (approved):

johnkerl:
Next I'll update the branch-protection rules for `main`

PR #416, opened 2022-10-15
Refactor from_anndata support

### LINKED ISSUE #340: tiledbsoma Py: util_ann.py -- all AnnData related code should be in `io` subpackage
Cleanup chore:  all AnnData-related code should be in tiledbsoma.io, not in the top-level tiledbsoma package.

bkmartinjr:
Code specific to support of soma.io is moved into io.py and out of the core classes. Previous PRs did this for dataframe -- this PR adds the same refinment for NdArray.

Changes:
* add a `create_from_matrix` method to soma.io module, used by from_anndata
* simplify `create_from_matrix` and its dependencies - there were many cut&paste functions that differed only in which dimension they incrementally chunked upon (ie, their stride axis). Created a single code path that simply has this as a parameter.
* add full set of unit tests for `create_from_matrix`
* fix a typo where `ndim` had been added to the API as `ndims`.  `ndim` is consistent with Pandas/Numpy and is preferred.

Fixes #340


johnkerl (approved):
thanks @bkmartinjr !!

PR #417, opened 2022-10-16
Use C++17 to ease build requirements 

### LINKED ISSUE #353: [c++]  Should be bring in the std::span alternate to only require C++17 ?

eddelbuettel:
This (draft) PR lowers our C++ compilation standard from C++20 (which works well on Linux and modern macOS) to C++17 (which should permit CI on macOS under the R setup, ditto for windows) and would close #353.  

It does two fairly simple things:
- replaces uses of `std::span` (which C++20 or later only) with [tcb::span](https://github.com/tcbrindle/span) (as TileDB Core does)
   - switches the instantiation from `std::span` to `tcp::span`
   - switches the include to `externals/span/span.hpp` 
   - adds a new header directory `includes/externals/`
- changes a single `contains()` for `unordered_map` (also C++20 or later) to `find() != end()`   

It also switches the R package to then use C++17.  Tests pass for me on R.

This should be considered a proposal-only.  I am not that deep into `cmake` etc so the above works without adding include directories etc but we may prefer a different location below `src/` for the added header as it is an implementation detail rather than interface.

### DIFF
--- a/apis/r/configure
+++ b/apis/r/configure
@@ -11,9 +11,10 @@ else
     fi
 
     if ! test -d inst/include/tiledbsoma; then
-        mkdir -p inst/include/tiledbsoma
+        mkdir -p inst/include/tiledbsoma inst/include/externals
         echo "** copying C++ header files"
         cp -ax ../../libtiledbsoma/include/tiledbsoma/* inst/include/tiledbsoma/
+        cp -ax ../../libtiledbsoma/include/externals/* inst/include/externals/
         echo "** copying C++ source and header files"
         cp -ax ../../libtiledbsoma/src/* src/
     fi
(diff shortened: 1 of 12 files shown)


gspowley (approved):

eddelbuettel:
Rebased, will await CI results.

Issue #410, opened 2022-10-14
Add fast nnz to R & Python packages


bkmartinjr:
We need to pull the C++ fast nnz (#404) into the Python and R packages.

- [x] Python accessing C++
- [x] R accessing C++

eddelbuettel:
Sorry, I did not catch that earlier as I was not tagged.  Given what @gspowley added this is (essentially) a one-liner:

```r
> library(tiledbsoma)
> uri <- "pre-made-test-data/soco/pbmc3k_processed/obs/"
> nnz(uri)
[1] 2638
> nnz("pre-made-test-data/soco/pbmc3k_processed/ms/mRNA/var")
[1] 1838
> nnz("pre-made-test-data/soco/pbmc3k_processed/ms/mRNA/varm/PCs/")
[1] 91900
> 
```

This (quick draft) returns the (current `uint64_t` in the `nnz()` signature in C++) as a double; we could use `int64` as well.

johnkerl:
@eddelbuettel thanks!

I think we need a little more at the Python/R level -- the C++ code throws in the case of overlapping fragments in the no-duplicates case -- what I _think_ we need in Python/R is some kind of try/catch where the try part calls the C++ code and returns it, while the catch part does what the Python/R code are currently doing.

eddelbuettel:
Cannot speak for Python but R does nothing besides taking a URI, passing it to your function and returning the result.  If the C++ code throws, we show an error.  So there is no added processing currently done.

My understanding was that `nnz()` aims to be _fast_ hence all operations are at the C++ layer.

johnkerl:
@eddelbuettel I'll think it through, doing Python first to get my bearings straight.

At
https://github.com/single-cell-data/TileDB-SOMA/pull/404/files#diff-f00fe93e3a84cedd18aaa8581c5133478d52f4576500a35eb4da1b65bdfb27ecR196
the C++ code uses a trick:

* Return `cell_num` if one fragment
* Return sum over fragment `cell_num` if the fragments don't overlap
* Return sum over fragment `cell_num` if the fragments do overlap but duplicates are allowed
* But otherwise the trick doesn't work and we'll (in Python) need something like what `main-old` is doing [here](https://github.com/single-cell-data/TileDB-SOMA/blob/main-old/apis/python/src/tiledbsoma/annotation_matrix.py#L37)

bkmartinjr:
[...]

PR #422, opened 2022-10-17
[r] Provide nnz() accessor

eddelbuettel:
This PR addresses one half of #410 by adding support for the nice `nnz()` getter (that was added by @gspowley as C++ functionality) to the R package.

Once #403 is merged we can add a simple unit test too.


johnkerl (approved):

gspowley (approved):

eddelbuettel:
Thanks -- two is an informal quorum for me so in it goes ...

PR #423, opened 2022-10-17
[c++] Build libtiledbsoma static library

gspowley:
This PR adds a `libtiledbsoma` static library build to cmake. After running `scripts/bld`, the library can be found at `dist/lib/libtiledbsoma.a` at the top of the repo.


eddelbuettel:
Could you add a mode where it produces only `dist/lib/libtiledbsoma.a` but no cpython library and no catch2 binary?  When the R package builds in isolation it only needs that static library.

johnkerl (approved):

eddelbuettel (changes requested):
The build should not require more resources or time than the narrow use it replaces so it would be desirable to a skip the build of test binary, and the build of the second shared library.

gspowley:
Closing in favor of #496.

Issue #431, opened 2022-10-17
Python/C++ read-path integration


johnkerl:
* Close out #400, pending help from TileDB folks on the MacOS CI issue tracked there
* Make a sibling PR -- #400 does the two dataframe classes; another will handle the n-d array classes

johnkerl:
Remaining work is well-covered by #470 #479 #380

johnkerl:
Needs a split-out for `DenseNdArray`

PR #446, opened 2022-10-20
Connect C++ nnz to Python

johnkerl:
Context: #431 and #410

### DIFF
--- a/apis/python/src/tiledbsoma/util.py
+++ b/apis/python/src/tiledbsoma/util.py
@@ -1,7 +1,7 @@
 import pathlib
 import time
 import urllib.parse
-from typing import TypeVar
+from typing import List, TypeVar, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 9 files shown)


eddelbuettel (commented):
I don't speak Python well enough to be of real help here so I abstain.  Thanks for thinking of me though.

apis/python/src/tiledbsoma/util.py
```diff
@@ -91,3 +91,25 @@ def uri_joinpath(base: str, path: str) -> str:
...
+            return []
+        step = ids.step
+        if step is None:
+            if ids.start <= ids.stop:
+                step = 1
+            else:
+                step = -1
+        stop = ids.stop + step
+        return list(range(ids.start, stop, step))
+    raise Exception(f"expected ids as slice or list; got {type(ids)}")
```
gspowley, line 115: We also need to support passing ids to `set_dim_points` as an Arrow array.

libtiledbsoma/CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ option(SUPERBUILD "If true, perform a superbuild (builds all missing dependencie
 option(CMAKE_IDE "(Used for CLion builds). Disables superbuild and sets the EP install dir." OFF)
 option(FORCE_BUILD_TILEDB "Forces a local build of TileDB instead of searching system paths." ON)
 option(TILEDB_FORCE_ALL_DEPS "Forces a local build of TileDB instead of searching system paths." OFF)
+# option(DOWNLOAD_TILEDB_PREBUILT "If tiledb is being super built, this controls downloading prebuilt artifacts or building from source" ON)
```
gspowley, line 49: Old comment from debugging session?

gspowley (changes requested):

libtiledbsoma/CMakeLists.txt
johnkerl: derp

apis/python/src/tiledbsoma/util.py
johnkerl: @gspowley this PR has some bits of #400 in it and it should not; sorry about that -- will rework

gspowley (approved):

PR #449, opened 2022-10-21
Correct configure by ensuring R_HOME is set early

eddelbuettel:
This PR corrects an oversight and ensure `R_HOME` is always set whereever `configure` may need it.

### DIFF
--- a/apis/r/configure
+++ b/apis/r/configure
@@ -1,5 +1,8 @@
 #!/bin/sh
 
+## Ensure R_HOME is set either explicitly or implicitly from the running R instance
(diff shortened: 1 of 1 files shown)


johnkerl (approved):
🚢

eddelbuettel:
Thanks. Will await CI for a few more minutes and may get to merge then.

eddelbuettel:
Actually corrected/expanded it. The `test -z` belongs to the top as well (as do the comments).  _Now_ it should be ready.

johnkerl (approved):

PR #450, opened 2022-10-21
Expose a faster method for member-URI enumeration

johnkerl:
The idiom `[soma.uri for soma in soco]` is marvelously expressive but instantiates each `SOMA`. This is parallelized via `TheadPoolExecutor` but even then with remotely stored data the overhead can be undesirable.

There is a `soco._get_member_uris` which _only_ consults the group meta for the parent `SOMACollection`, and is almost instantaneous. This should be made a first-class citizen, as it does only what the user wants (no more) and is much more performant.

### DIFF
--- a/apis/python/README-ingestion.md
+++ b/apis/python/README-ingestion.md
@@ -228,7 +228,7 @@ VDAC3            1.1250     30.971519               8.986513                   2
 ```
 
 ```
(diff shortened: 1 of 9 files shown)


aaronwolen (approved):
Thanks!

PR #451, opened 2022-10-21
[c++] Support passing column_names as None

gspowley:
Support passing `column_names=None` to `SOMAReader` in python.


johnkerl (approved):
✨

PR #452, opened 2022-10-21
[c++] Add column name to the arrow schema

gspowley:
Add column name to the arrow schema. The column name is optional, but needed for R.


johnkerl (approved):

eddelbuettel (approved):
This is lovely (as discussed in slack) and gets the job done!

PR #453, opened 2022-10-21
[python] Apply black/isort/flake8/mypy to all repo .py files

johnkerl:
Previously, only within `apis/python`, and even then, not for `apis/python/tests`.

Some of our challenges with #400, #446, and #447 have had to do with `typeguard`, at unit-test time, reaching into `.py` files it had not visited before. Here we try to avoid future such surprises.


johnkerl:
Closed in favor of #711

PR #454, opened 2022-10-22
[r] Enhanced soma reader

eddelbuettel:
This PR takes advantage of some recent change in the C++ layer and further enhances the R use of it: 
- Rows (or row ranges or points) and columns can now be select at will, or not at all.  
- Tests have been added.  
- An `R CMD check` nag has been silenced. 
- A minor tweak has been made to one C++ library file.


gspowley (approved):
LGTM!

eddelbuettel:
> [Silence one warning when build with R package](https://github.com/single-cell-data/TileDB-SOMA/pull/454/commits/e1fe0b09f285a5be6332fe8fe64c0e1d14151e8b)

That was sloppily written. I silences one warning when linting _i.e_ when running `R CMD check` as R dislikes `exit()` in dynamically loaded code, so we now `#ifdef` this away when in R mode.

PR #456, opened 2022-10-23
[c++] Add obs slice X test

gspowley:
Add a test that queries the `obs` array and uses the results to slice an `X/data` array query.


eddelbuettel (approved):
Love it. I will so borrow the 'points' selector for the tiledb-r package too.

PR #459, opened 2022-10-24
Correctly handle zero-length dataframes with SOMAReader

johnkerl:
Fixes #458

PR #460, opened 2022-10-24
More unit-test cases for #459

johnkerl:
Follow-up from
https://github.com/single-cell-data/TileDB-SOMA/pull/459#pullrequestreview-1153616376

### DIFF
--- a/apis/python/tests/test_soma_dataframe.py
+++ b/apis/python/tests/test_soma_dataframe.py
@@ -243,6 +243,7 @@ def test_empty_soma_dataframe(tmp_path):
     assert len(a.read_all()) == 0
     assert len(next(a.read_as_pandas())) == 0
     assert len(a.read_as_pandas_all()) == 0
+    assert isinstance(a.read_as_pandas_all(), pd.DataFrame)
 
 
 def test_soma_columns(tmp_path):
--- a/apis/python/tests/test_soma_dataframe.py
+++ b/apis/python/tests/test_soma_dataframe.py
@@ -238,7 +238,11 @@ def test_SOMADataFrame_read_column_names(simple_soma_data_frame, ids, col_names)
 def test_empty_soma_dataframe(tmp_path):
     a = soma.SOMADataFrame((tmp_path / "A").as_posix())
     a.create(pa.schema([("a", pa.bool_())]))
-    next(a.read())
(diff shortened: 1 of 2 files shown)


apis/python/tests/test_soma_dataframe.py
```diff
@@ -238,7 +238,11 @@ def _check_tbl(tbl, col_names, ids, *, demote):
 def test_empty_soma_dataframe(tmp_path):
     a = soma.SOMADataFrame((tmp_path / "A").as_posix())
     a.create(pa.schema([("a", pa.bool_())]))
-    next(a.read())
+    # Must not throw
+    assert len(next(a.read())) == 0
+    assert len(a.read_all()) == 0
+    assert len(next(a.read_as_pandas())) == 0
+    assert len(a.read_as_pandas_all()) == 0
```
bkmartinjr, line 245: Given the bug fix, we also need:
```
assert isinstance(a.read_as_pandas_all(), pd.DataFrame)
```

apis/python/tests/test_soma_indexed_dataframe.py
```diff
@@ -186,7 +186,11 @@ def _check_tbl(tbl, col_names, ids, *, demote):
 def test_empty_soma_indexed_dataframe(tmp_path):
     a = soma.SOMAIndexedDataFrame((tmp_path / "A").as_posix())
     a.create(pa.schema([("a", pa.int32())]), index_column_names=["a"])
-    next(a.read())
+    # Must not throw
+    assert len(next(a.read())) == 0
+    assert len(a.read_all()) == 0
+    assert len(next(a.read_as_pandas())) == 0
+    assert len(a.read_as_pandas_all()) == 0
```
bkmartinjr, line 193: Given the bug fix, we also need:
```
assert isinstance(a.read_as_pandas_all(), pd.DataFrame)
```

bkmartinjr (approved):
One requested addition so that the test covers the bug fix.  Otherwise looks great.

johnkerl:
> One requested addition so that the test covers the bug fix. Otherwise looks great.

@bkmartinjr commit #1 on this PR added a test, which failed as intended. Commit #2 fixed the fail. I sequenced these two commits that way by design.

Issue #427, opened 2022-10-17
SOMA name prefixing


johnkerl:
The [abstract spec](https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md) has names like `SOMADataFrame` which is fine.

In Python were package names are explicit:
```
import tiledbsoma
sdf = tiledbsoma.SOMADataFrame(uri)
```
is repetitive.

In R it's not repetitive:
```
library(tiledbsoma)
sdf <- SOMADataFrame(uri)
```

Proposal:

* For Python, omit `SOMA` prefixes from class names, so `sdf = tiledbsoma.DataFrame(uri)`
* In R, leave them so `sdf <- SOMADataFrame(uri)`
* In both cases, s-o-m-a gets uttered just once; with a dot in Python, and without in R

johnkerl:
@thetorpedodog I'm fast-tracking on request of @bkmartinjr :v: -- I will try to get this out today

johnkerl:
#464 is merged

PR #464, opened 2022-10-25
Unprefix SOMA names

johnkerl:
Implements #427 

@bkmartinjr @aaronwolen heads-up, this is a breaking change for `main` as class-name metadata is written to storage so that we can invoke the correct constructor at read time. So for example data written before this PR is merged will have `obs` with storage-metadata class name `SOMADataFrame` but code running after this PR will only accept class name `DataFrame` and will not instantiate `obs` at read time.

This is the right time in `main` development to be making this change.

The line-count on this PR is high, but it is just a mindless search and replace.


apis/python/src/tiledbsoma/__init__.py
```diff
@@ -61,15 +61,15 @@
...
-    "SOMAIndexedDataFrame",
-    "SOMAMeasurement",
-    "SOMAMetadataMapping",
-    "SOMASparseNdArray",
+    "Collection",
+    "DataFrame",
+    "DenseNdArray",
+    "DoesNotExistError",
+    "Experiment",
+    "Error",
```
bkmartinjr, line 69: `Error` (and its related child classes) may be one place where we want to continue to use `SOMAError` to clearly differentiate from other types of errors/exceptions.

For example, it is `tiledb.TileDBError`, not `tiledb.Error` 

@thetorpedodog - does the Google conventions give any guidance here?

bkmartinjr: Ditto for `DoesNotExistError`

johnkerl:
@bkmartinjr @thetorpedodog 
https://github.com/single-cell-data/TileDB-SOMA/pull/464/commits/0b4001d604e7635f681c279b3821aed70f5d1bd9

apis/python/src/tiledbsoma/__init__.py
```diff
@@ -61,15 +61,15 @@
     "TileDBPlatformConfig",
     "TileDBObject",
```
bkmartinjr, line 62: Why are TileDBObject and TileDBArray exposed in the package?

johnkerl: presumably oversight -- this is not new -- thank you for the code-review catch -- i can neaten that now or on a separate PR

johnkerl: #465

apis/python/src/tiledbsoma/factory.py
```diff
@@ -69,27 +69,27 @@ def _construct_member(
     assert class_name is not None
 
     # Now invoke the appropriate per-class constructor.
-    if class_name == "SOMAExperiment":
+    if class_name == "Experiment":
```
bkmartinjr, line 72: A nit: we are changing the metadata value due to a change in the Python API, which implies they are linked somehow.  That seems a bit awkward as we may have other Python API-only changes in the future that don't change the persistent data storage format.

What would you think about creating a separate enumeration of values that are strored in the metadata, rather than storing the (literal) class name?   Eg, store 0 for Collection, 1 for DataFrame, etc.  Put another way, a level of indirection between the class name and the metadata code.

Eg, in Python:
```
class PersistentMetadataObjectTypes(Enum):
    SOMA_COLLECTION = 0,
    SOMA_DATAFRAME = 1,
    ...
```

In practice, I'm not sure this buys us more than a more explicit separation between the per-language class name and the underlying metadata.

apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -41,13 +41,13 @@ def __init__(
         self._cached_is_sparse = None
 
     @property
-    def soma_type(self) -> Literal["SOMADataFrame"]:
-        return "SOMADataFrame"
+    def soma_type(self) -> Literal["DataFrame"]:
```
bkmartinjr, line 44: The abstract spec says that this returns "SOMADataFrame", which makes sense as that string is not the Python class, but the SOMA API "type".  I think you want all `soma_type` functions to be left as is.

CC: @thetorpedodog

apis/python/src/tiledbsoma/factory.py
johnkerl: @bkmartinjr very good point.

my recommendation is that we use the classnames from the language-independent API spec. then in `factory.py` we can have a dict mapping between API-spec names like `SOMADataFrame` and language-specific names like `DataFrame`.

johnkerl: i prefer this over int-codes as the int codes are really just a less-clear enumeration of the API-spec names; for clarity i'd prefer to store the latter as metadata

bkmartinjr: Works for me.  My main point was not assume class_name == metadata_code.  Using the abstract type string works for me, which I take to mean `SOMACollection`, `SOMADataFrame`, etc.

bkmartinjr (changes requested):
Couple of issues noted inline.  Hit me up for a second review once the design issues settle.  Thanks for doing this!

johnkerl:
@thetorpedodog remaining issue is
https://github.com/single-cell-data/TileDB-SOMA/pull/464#discussion_r1004809930
and
https://github.com/single-cell-data/TileDB-SOMA/pull/464#discussion_r1004810882
which i already (in anticipation) pushed as a commit on this PR -- if you're OK with that this should be a quick stamp IMHO 🙏

apis/python/src/tiledbsoma/__init__.py
thetorpedodog: One of the recommendations they have is to [not repeat the package name in exceptions as well](https://google.github.io/styleguide/pyguide.html#244-decision):

> Libraries or packages may define their own exceptions. When doing so they must inherit from an existing exception class. Exception names should end in `Error` and should not introduce repetition (`foo.FooError`).

However, skimming through the stdlib documentation I found that `pickle` uses [`pickle.PickleError`](https://docs.python.org/3/library/pickle.html#pickle.PickleError) as a base exception name.

Much like the `SOMAObject` special case, I think `SOMAError` as the base error would be reasonable, but with `DoesNotExistError` as one of the derived types.

Also worth noting is that some of these exception types will no doubt end up moving to the `somabase` package (though moving them there from an initial implementation here seems completely reasonable to me).

apis/python/src/tiledbsoma/dataframe.py
thetorpedodog: This is what I was thinking too; that the `soma_type` property would remain `SOMAWhatever`.

thetorpedodog (approved):
other than the previously-mentioned exception naming thing (which I am flexible on) this looks good.

johnkerl:
@bkmartinjr ready for round two! :)

apis/python/src/tiledbsoma/factory.py
```diff
@@ -56,40 +56,44 @@ def _construct_member(
...
         else:
             return None
 
     except tiledb.TileDBError:
         return None
 
-    assert class_name is not None
+    assert spec_name is not None
+    if spec_name not in SPEC_NAMES_TO_CLASS_NAMES:
+        raise Exception(f'name "{spec_name}" unrecognized')
```
bkmartinjr, line 71: this should be an error derived from SOMAError, right?  In any case, it is generally a bad idea to raise naked Exception instances

bkmartinjr (approved):
I don't want to have this be an exercise in bike shedding, so...LGTM!  :-)

(one optional inline note on SOMADoesNotExist vs DoesNotExist)

apis/python/src/tiledbsoma/factory.py
johnkerl: @bkmartinjr good call!! :)

There are more such in the repo so I filed #467 as a follow-on

apis/python/src/tiledbsoma/__init__.py
thetorpedodog: i.e. `SOMAError` and `DoesNotExist` was my suggestion. `Error` and `DoesNotExist` are also great. I think both are strictly better than `SOMAError` and `SOMADoesNotExist` (which honestly would still be acceptable).

apis/python/src/tiledbsoma/factory.py
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/464/commits/7cf886146a80027630255943d984b71bd38f8b4c

apis/python/src/tiledbsoma/__init__.py
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/464/commits/5a4cac22f860ca9f5fccf89eed834da269be7f7f

bkmartinjr: I can live with any of these.  Consider me still "approved"!  :+1:

Issue #380, opened 2022-10-07
Follow up on dense SOMADataFrame objects


johnkerl:
Reproduce, triage, and propose a solution (if a problem still exists):

* https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/python/src/tiledbsoma/soma_dataframe.py#L97
* https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/python/src/tiledbsoma/soma_dataframe.py#L482

johnkerl:
# Repro script

Repro script (adapted from [tests/test_soma_dataframe.py](https://github.com/single-cell-data/TileDB-SOMA/blob/c47314179c4d03c45db327ed9498129bfd4bc954/apis/python/tests/test_soma_dataframe.py)):

https://gist.github.com/johnkerl/bd9530ab73d4e4c99e5f5b04888d0ace

# Failure

Fails with:

```
Traceback (most recent call last):
  File "/Users/johnkerl/git/single-cell-data/TileDB-SOMA/apis/python/./repro.py", line 77, in <module>
    table = sdf.read_all(value_filter="foo == 40 or foo == 20")
  File "/Users/johnkerl/git/single-cell-data/TileDB-SOMA/apis/python/src/tiledbsoma/soma_dataframe.py", line 241, in read_all
    return pa.concat_tables(
  File "pyarrow/table.pxi", line 2623, in pyarrow.lib.concat_tables
  File "/Users/johnkerl/git/single-cell-data/TileDB-SOMA/apis/python/src/tiledbsoma/soma_dataframe.py", line 222, in read
    yield util_arrow.ascii_to_unicode_pyarrow_readback(table)
  File "/Users/johnkerl/git/single-cell-data/TileDB-SOMA/apis/python/src/tiledbsoma/util_arrow.py", line 139, in ascii_to_unicode_pyarrow_readback
    [element.as_py().decode("utf-8") for element in old_field]
  File "/Users/johnkerl/git/single-cell-data/TileDB-SOMA/apis/python/src/tiledbsoma/util_arrow.py", line 139, in <listcomp>
    [element.as_py().decode("utf-8") for element in old_field]
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte
```
at repro-script line 77:
```
table = sdf.read_all(value_filter="foo == 40 or foo == 20")
```
even though
```
table = sdf.read_all(ids=slice(1, 2))
```
[...]

PR #466, opened 2022-10-25
[python] Store SOMADataFrame as a dense TileDB array

johnkerl:
For issue #380


apis/python/src/tiledbsoma/soma_dataframe.py
```diff
@@ -285,22 +262,17 @@ def write(self, values: pa.Table) -> None:
                     )
```
bkmartinjr, line 262: Line 260 above - why convert using `to_pandas()` and then call `np.array`?  

`to_pandas` will always return a np.array as long as the type can be stored in NumPy (ie, isn't a Pandas extension type).  And `np.array` will fail given one of those extension types.  

`np.array` also make sa copy by default, so you end up with a lot of extra memory copies here.

I suspect you just want:   `values.column(name).to_pands()`

If you are concerned that your type mapper might return a type that can't be expressed in numpy, you could assert that the return value is instanceof np.array.

bkmartinjr: I should add:   `values.column(name)` returns a `ChunkedArray`.  The latest PyArrow `ChunkedArray` also contains a `to_numpy()` method - maybe just use that to make your assumptions more explicit, and skip the copy?

apis/python/src/tiledbsoma/soma_dataframe.py
```diff
@@ -285,22 +262,17 @@ def write(self, values: pa.Table) -> None:
...
-            with self._tiledb_open("w") as A:
-                A[lo : (hi + 1)] = attr_cols_map
+        # TODO: This was a quick thing to bootstrap some early ingestion tests but needs more thought.
+        # In particular, rowids needn't be either zero-up or contiguous.
+        assert len(rowids) > 0
+        rowids = sorted(rowids)
+        assert rowids[0] == 0
+
+        # dense write
+        lo = rowids[0]
```
bkmartinjr, line 272: This is not what the spec says - it allows for a scattered write (ie, each row in the RecordBatch can have a non-contiguous soma_rowid).

I don't think TileDB Dense arrays support this scattered indexing, for either read or write. We ran into the same issue in DenseNdArray, and changed the read/write operations to require coordinates be contiguous (ie, they were specified separately, as either a single `int` or as a slice).

We may need to do the same for DataFrame.

bkmartinjr (changes requested):
This PR points out a disconnect between the abstract spec and TileDB dense array capabilities.  The spec assumes the ability to do scattered read/write onto a dense array, but the underlying system does not support that.

We either need to revise the spec, or implement a shim.  Seems like an offline design discussion.

Note that we ran into the same issue with DenseNdArray, and elected to change the spec.

apis/python/src/tiledbsoma/soma_dataframe.py
johnkerl: @bkmartinjr that predates this PR -- is not new to it or modified by it -- i don't even think i wrote this bit -- i'm happy to update on this PR if we want to go there

bkmartinjr: I understand.  But a code review is a code review :-)

If you are asking me to file an issue, just let me know.

johnkerl: good call @bkmartinjr -- i think the `to_pandas()` with also `np.array` is because without the later

```
TypeError: Cannot convert Series to numpy.ndarray
```

you are correct about `to_numpy()`:

```
>>> type(table.column('i'))
<class 'pyarrow.lib.ChunkedArray'>
>>> type(table.column('i').to_pandas())
<class 'pandas.core.series.Series'>
>>> type(np.asarray(table.column('i').to_pandas()))
<class 'numpy.ndarray'>
>>> type(table.column('i').to_numpy())
<class 'numpy.ndarray'>
```

where `table` is a `pyarrow.Table`

johnkerl: @bkmartinjr thanks --

the `to_pandas()` with also `np.array` is because without the latter

```
TypeError: Cannot convert Series to numpy.ndarray
```

since

```
>>> type(table.column('i'))
<class 'pyarrow.lib.ChunkedArray'>
>>> type(table.column('i').to_pandas())
<class 'pandas.core.series.Series'>
>>> type(np.asarray(table.column('i').to_pandas()))
<class 'numpy.ndarray'>
```

you're right, there is `.to_numpy()`, which _does_ produce a `numpy.ndarray`, but unlike `to_pandas()` it doesn't take the `types_mapper` argument we need

in any event, that bit predates this PR -- is not new to it or modified by it -- my preference is to
take these (very valid!) copy-reduction concerns into a separately tracked issue/PR

johnkerl: @bkmartinjr since these are very valid concerns, outside of what this PR changed, i'll file a follow-up issue to track

johnkerl: #469

bkmartinjr:
> Have I missed anything?

Conceptually no - that should do it.  However, this implementation doesn't seem to work.  I will do some debugging and comment on the PR as I figure out what is going on.

apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -288,22 +262,17 @@ def write(self, values: pa.Table) -> None:
...
-            # dense write
-            lo = rowids[0]
-            hi = rowids[-1]
-            with self._tiledb_open("w") as A:
-                A[lo : (hi + 1)] = attr_cols_map
+        # TODO: This was a quick thing to bootstrap some early ingestion tests but needs more thought.
+        # In particular, rowids needn't be either zero-up or contiguous.
+        assert len(rowids) > 0
+        rowids = sorted(rowids)
+        assert rowids[0] == 0
```
bkmartinjr, line 269: this assert is spurious -- it doesn't allow for writes at other rows - there is nothing magic about rowid == 0

apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -288,22 +262,17 @@ def write(self, values: pa.Table) -> None:
...
-            rowids = sorted(rowids)
-            assert rowids[0] == 0
-
-            # dense write
-            lo = rowids[0]
-            hi = rowids[-1]
-            with self._tiledb_open("w") as A:
-                A[lo : (hi + 1)] = attr_cols_map
+        # TODO: This was a quick thing to bootstrap some early ingestion tests but needs more thought.
+        # In particular, rowids needn't be either zero-up or contiguous.
```
bkmartinjr, line 266: this section looks wrong to me - several issues:
* we need to change the API for write to take a single dimension sub-array coordinates (like DenseNdArray, but just a single dimension),  after which all of this stuff can be removed
* and there must be a test to confirm that the table does not contain the soma_rowid column, as that would be in conflict with the function param specifying the sub-array coordaintes

bkmartinjr: A simple unit test will test this - ie,
* create a DataFrame
* write a single row at position 0
* write 10 rows at position 1-11
* etc.

bkmartinjr (changes requested):
OK, write doesn't work as implemented in this PR.

Here is a super quick rewrite that works for me (but caveat:  I did not test it extensively, and I assumed we want to move to a model where only sub-arrays can be read/written).

Please also note that while TileDB dense array _read_ (aka `__getitem__`) supports scattered read (eg, a list of int coords), the _write_ (aka `__setitem__`) does not support such a thing - it only supports singleton ints or slices.

```
    def write(self, coord: Union[int, slice], values: pa.Table) -> None:
        """
        Write an Arrow.Table to the persistent object.

        :param values: An Arrow.Table containing all columns, including the index columns. The schema for the values must match the schema for the ``DataFrame``.

        The ``values`` Arrow Table must contain a ``soma_rowid`` (int64) column, indicating which rows are being written.
        """
        if SOMA_JOINID not in values.schema.names:
            raise ValueError(f"Write requires  {SOMA_JOINID} to be specified in table.")

        attr_cols_map = {}
        for name in values.schema.names:
            attr_cols_map[name] = np.asarray(
                values.column(name).to_pandas(
                    types_mapper=util_arrow.tiledb_type_from_arrow_type_for_write,
                )
            )

        with self._tiledb_open("w") as A:
            A[coord] = attr_cols_map
```

johnkerl:
Abandoned in favor of #538

PR #480, opened 2022-10-27
[r] Expand iterated reader via refactored row/col selectors

eddelbuettel:
This PR use the row and column selection (via dimension points, range, select column, query conditions or any combination of these) on the iterated SOMA Reader instance.  

New tests have been added, these are currently conditional on a data package we may want to make available.


aaronwolen (approved):
Awesome! The only caveat is we might need to rename some functions/args down the road but in my tests the code worked perfectly.

eddelbuettel:
Thanks -- and yes, these are 'first pass' implementations subject to renaming and more consistent parameters etc chiefly to test / demonstrate the basic (and very solid thanks to the C++ layer) architecture.

Rebased, will await test re-run of CI.

PR #481, opened 2022-10-27
Handle leading/trailing whitespace in value_filter

johnkerl:
Works around https://github.com/TileDB-Inc/TileDB-Py/issues/1396


bkmartinjr (approved):
This only works around one of the issues - it doesn't also deal with the embedded newline, etc.

I am not blocked by the bug, so my feeling is we should wait for the underlying embedded parser to fix the issue, rather than trying to work around parts of it.

Issue #479, opened 2022-10-26
Python regression: sparseNdArray.read_sparse_tensor slicing does not work


bkmartinjr:
Recent regression likely related to C++ reader integration (#447):  SparseNdArray.read_as_tensor no longer slices correctly.  If coordinates (ids) are specified, it returns *all* rather than the requested slice. `read_as_pandas_all()` still works.

Example:  I have a large 2D matrix.  Using `read_as_pandas_all` with a slice produces a different result than `read_sparse_tensor`.  They should be identical (in value)

```
In [21]: census['human'].ms['RNA'].X['raw'].read_as_pandas_all((0, slice(None)))
Out[21]: 
      soma_dim_0  soma_dim_1  soma_data
0              0          13        2.0
1              0          18        1.0
2              0          21        1.0
3              0          24        1.0
4              0          29        4.0
...          ...         ...        ...
4974           0       23715        7.0
4975           0       23719        1.0
4976           0       23724        1.0
4977           0       23729        3.0
4978           0       23745        1.0

[4979 rows x 3 columns]

In [22]: for i in census['human'].ms['RNA'].X['raw'].read_sparse_tensor((0, slice(None))): print(i)
<pyarrow.SparseCOOTensor>
type: float
shape: (176, 23747)
```

The (sliced) result above seems to be identical to the entire array, eg,

```
In [23]: for i in census['human'].ms['RNA'].X['raw'].read_sparse_tensor((slice(None), slice(None))): print(i)
<pyarrow.SparseCOOTensor>
type: float
shape: (176, 23747)
```

johnkerl:
# Debugging

@bkmartinjr interesting finding -- I spent a while debugging this today and it is (surprisingly) _not_ a regression -- the bug was there before.

Specifically I checked out commits `ae1554c800c98aca721b7632d309cac797b27230` and `15185875f9831a9164baa91052ac9e387feb56bd` which were before and after #447. 

This script https://gist.github.com/johnkerl/f1331171e1f91fc7eb7312605038cc1f has the same output on the before and after commits, which is

```
================================================================
   soma_dim_0  soma_dim_1  soma_data
0           0           0   2.424371

================================================================
<pyarrow.SparseCOOTensor>
type: float
shape: (65662, 24857)
```

[...]

Issue #470, opened 2022-10-26
IndexedDataFrame only allows read indexing on first dimension, fails for anything other than dimension 0


bkmartinjr:
The latest IndexedDataFrame will not allow a multi-index-column dataframe to be access by anything other than the first dimension.

Example:
```
import pandas as pd
import pyarrow as pa
import tiledbsoma as soma

sdf = soma.IndexedDataFrame("./test_df")
df = pd.DataFrame(data={
  'soma_joinid': [0,1,2], 'a': [0,1,2], 'b': [3,4,5], 'c': [6,7,8]
})

tbl = pa.Table.from_pandas(df)
sdf.create(schema=tbl.schema, index_column_names=['a', 'b'])
sdf.write(tbl)

print(sdf.read_all(ids=([0,1],slice(None))).column('c').to_pylist())
assert sdf.read_all(ids=([0,1],slice(None))).column('c').to_pylist() == [6,7]
```
The assert fails.  The `read_all` always returns all values (as if no coordinates had been specified).  It should return `[6,7]`
```
[6, 7, 8]
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In [3], line 15
     12 sdf.write(tbl)
     14 print(sdf.read_all(ids=([0,1],slice(None))).column('c').to_pylist())
---> 15 assert sdf.read_all(ids=([0,1],slice(None))).column('c').to_pylist() == [6,7]
```

Inspecting the code suggest that the read method is hardwired to allow only a single dimension, and expects the `ids` param to have only the IDs for that dimension, rather than a tuple/list of dimensions.

For example, this works, but only selects by the first dimension:
```
# Oddly, this work but only selects on the first index column
assert sdf.read_all(ids=[0,1]).column('c').to_pylist()==[6,7]
```

Issue #419, opened 2022-10-16
Python tiledbsoma: error when requesting index of type `bytes`


bkmartinjr:
Attempting to create a SOMAIndexedDataFrame with an index column of type python `bytes` fails.  Given that TileDB supports this underlying type as an index ([reference](https://docs.tiledb.com/main/how-to/arrays/creating-arrays/creating-dimensions)), this should work and appears to be a bug in the SOMA type handling.

See #420 for an example unit test (currently marked xfail).

Test case:
```

import pandas as pd
import pyarrow as pa
import tiledbsoma as soma


sdf = soma.DataFrame("test_dataframe")
df = pd.DataFrame(data = {
    'soma_joinid': [0,1,2],
    'bytes': [b'a', b'b', b'c'],
    'ints': [0, 1, 2]
})

tbl = pa.Table.from_pandas(df)
sdf.create(tbl.schema, index_column_names=['bytes'])
sdf.write(tbl)
```

Result:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In [28], line 14
      7 df = pd.DataFrame(data = {
      8     'soma_joinid': [0,1,2],
      9     'bytes': [b'a', b'b', b'c'],
     10     'ints': [0, 1, 2]
     11 })
     13 tbl = pa.Table.from_pandas(df)
---> 14 sdf.create(tbl.schema, index_column_names=['bytes'])
     15 sdf.write(tbl)

File ~/projects/TileDB-SOMA/apis/python/src/tiledbsoma/soma_indexed_dataframe.py:56, in SOMAIndexedDataFrame.create(self, schema, index_column_names)
     50 """
     51 :param schema: Arrow Schema defining the per-column schema. This schema must define all columns, including columns to be named as index columns. If the schema includes types unsupported by the SOMA implementation, an error will be raised.
     52 
     53 :param index_column_names: A list of column names to use as user-defined index columns (e.g., ``['cell_type', 'tissue_type']``). All named columns must exist in the schema, and at least one index column name is required.
     54 """
     55 schema = _validate_schema(schema, index_column_names)
---> 56 self._create_empty(schema, index_column_names)
     57 self._is_indexed = True
     58 self._index_column_names = tuple(index_column_names)

File ~/projects/TileDB-SOMA/apis/python/src/tiledbsoma/soma_indexed_dataframe.py:90, in SOMAIndexedDataFrame._create_empty(self, schema, index_column_names)
[...]

Issue #418, opened 2022-10-16
[python] Dataframe throws exception if index column is a string/large_string


bkmartinjr:
Attempting to `create` a SOMAIndexedDataFrame with a string index will generate an exception.

See #420 for an example unit test (currently marked xfail).

Test case:

```
import numpy as np
import pandas as pd
import pyarrow as pa
import tiledbsoma as soma


df = pd.DataFrame(
    data={
        "soma_joinid": np.arange(3, dtype=np.int64),
        "ascii": ["aa", "bbb", "cccccc"],
        "float32": np.array([0.0, 1.1, 2.2], np.float32),
    }
)

arrow_tbl = pa.Table.from_pandas(df)
sdf = soma.SOMAIndexedDataFrame("test_sparse_dataframe")
sdf.create(arrow_tbl.schema, ["ascii"])
sdf.write(arrow_tbl)
```

Result:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In [5], line 17
     15 arrow_tbl = pa.Table.from_pandas(df)
     16 sdf = soma.SOMAIndexedDataFrame("test_sparse_dataframe")
---> 17 sdf.create(arrow_tbl.schema, ["ascii"])
     18 sdf.write(arrow_tbl)

File ~/projects/TileDB-SOMA/apis/python/src/tiledbsoma/soma_indexed_dataframe.py:56, in SOMAIndexedDataFrame.create(self, schema, index_column_names)
     50 """
     51 :param schema: Arrow Schema defining the per-column schema. This schema must define all columns, including columns to be named as index columns. If the schema includes types unsupported by the SOMA implementation, an error will be raised.
     52 
     53 :param index_column_names: A list of column names to use as user-defined index columns (e.g., ``['cell_type', 'tissue_type']``). All named columns must exist in the schema, and at least one index column name is required.
     54 """
     55 schema = _validate_schema(schema, index_column_names)
---> 56 self._create_empty(schema, index_column_names)
     57 self._is_indexed = True
     58 self._index_column_names = tuple(index_column_names)

File ~/projects/TileDB-SOMA/apis/python/src/tiledbsoma/soma_indexed_dataframe.py:83, in SOMAIndexedDataFrame._create_empty(self, schema, index_column_names)
     81 hi: Any = None
     82 if dtype != str:
---> 83     if np.issubdtype(dtype, np.integer):
     84         lo = np.iinfo(dtype).min
[...]

PR #483, opened 2022-10-27
Update SOMAIndexedDataFrame indexing logic

johnkerl:
## Context

Addresses #470. See also related discussion on #479.

## Fix

As noted on #479, desired indexing semantics:

* A sequence of coordinates is accepted, one per dimension.
* Sequence length must be at least one and <= number of dimensions.
* If the sequence contains missing coordinates (length less than number of dimensions), then "slice(None)" is assumed for the missing dimensions.
* Per-dimension, explicitly specified coordinates can be one of: a value, a list/ndarray/paarray/etc of values, a slice, etc.
* `raise` on anything else.

Also note that since `SOMAIndexedDataFrame` -- unlike `SOMADataFrame` -- can have more than one dimension, necessarily coordinates are a list such as `[10]` or `[None]` or `[None, 10]`.


johnkerl:
@bkmartinjr this is ready for review

apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -179,6 +179,8 @@ def read(
                     lo_hi = util.slice_to_range(ids)
                     if lo_hi is not None:
                         sr.set_dim_ranges(SOMA_ROWID, [lo_hi])
+                else:
```
bkmartinjr, line 182: for any new code & behavior, please add a unit test to cover it.  In this case, to ensure the error ir raised as expected.

apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -84,7 +84,7 @@ def _create_empty(
             # We need domain=(None,None) for string dims
             lo: Any = None
             hi: Any = None
-            if dtype != str:
+            if dtype != str and dtype != "ascii":
                 if np.issubdtype(dtype, np.integer):
                     lo = np.iinfo(dtype).min
                     hi = np.iinfo(dtype).max - 1
```
bkmartinjr, line 90: this came up in the sprint planning meeting today - why rule out other dimension types supported by the underlying database (eg, datetime)?  If `tiledb.Dim` accepts it, seems worth supporting.

apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -200,18 +201,46 @@ def read(
...
-                    sr.set_dim_points(dim_name, ids)
-                elif isinstance(ids, pa.ChunkedArray):
-                    sr.set_dim_points(dim_name, ids)
-                elif isinstance(ids, pa.Array):
-                    sr.set_dim_points(dim_name, pa.chunked_array(ids))
-                elif isinstance(ids, slice):
-                    lo_hi = util.slice_to_range(ids)
-                    if lo_hi is not None:
-                        sr.set_dim_ranges(dim_name, lo_hi)
+                if not (isinstance(ids, list) or isinstance(ids, tuple)):
```
bkmartinjr, line 214: the comment above says `ids` can be any Sequence, but this raises an error if not a list or tuple.  There are many other useful sequences (ie. range, various iterators, etc) - I think we should support them.

you can check type by doing `isinstance(ids, collections.abc.Sequence)`

apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -200,18 +201,46 @@ def read(
...
+                    raise SOMAError(
+                        f"ids type {type(ids)} unhandled; expected list or tuple"
+                    )
+                if len(ids) < 1 or len(ids) > A.schema.domain.ndim:
+                    raise SOMAError(
+                        f"ids {ids} must have length between 1 and ndim ({A.schema.domain.ndim}); got {len(ids)}"
+                    )
+
+                for i in range(len(ids)):
```
bkmartinjr, line 223: don't assume it is a tuple or list, instead do:    `for i, dim_ids in enumerate(ids):`

apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -200,18 +201,46 @@ def read(
...
+                    dim_name = A.schema.domain.dim(i).name
+                    dim_ids = ids[i]
+                    if dim_ids is None:  # select all in this dimension
+                        pass
+                    elif isinstance(dim_ids, int):
+                        sr.set_dim_points(dim_name, [dim_ids])
+                    elif isinstance(dim_ids, list):
```
bkmartinjr, line 230: should also support tuple, i,e, `instance(dim_ids, (list, tuple))`

bkmartinjr: Also you may find it beneficial to use the multi-type version of `instance`, which would allow many of these to be done on a single line, eg,

```
elif isinstance(dim_ids, (tuple, list, pa.ChunkedArray)):
    sr.set_dim_points(dim_name, dim_ids)
```

apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -200,18 +201,46 @@ def read(
...
+                        sr.set_dim_points(dim_name, dim_ids)
+                    elif isinstance(dim_ids, pa.Array):
+                        sr.set_dim_points(dim_name, pa.chunked_array(dim_ids))
+                    elif isinstance(dim_ids, slice):
+                        lo_hi = util.slice_to_range(dim_ids)
+                        if lo_hi is not None:
+                            sr.set_dim_ranges(dim_name, [lo_hi])
+                    else:
+                        raise SOMAError(
+                            f"dim_ids type {type(dim_ids)} at slot {i} unhandled"
```
bkmartinjr, line 242: maybe "unsupported" is a better term, as "unhandled" does not make it clear why the error occured.

apis/python/tests/test_indexed_dataframe.py
```diff
@@ -105,17 +101,49 @@ def simple_indexed_data_frame(tmp_path):
...
-    sdf.write(rb)
-    yield (schema, sdf, n_data, index_column_names)
-    sdf.delete()
+    sidf.write(rb)
+    yield (schema, sidf, n_data, index_column_names)
+    sidf.delete()
+
+
+@pytest.fixture
+def doubly_indexed_data_frame(tmp_path):
```
bkmartinjr, line 110: highly recommend testing more than two dimensions.  See the NdArray test fixtures for an example that does 1 to 5 dimensions as a parameter (really any number, but that seemed suffiicient).

apis/python/tests/test_indexed_dataframe.py
```diff
@@ -289,6 +325,166 @@ def make_dataframe(request):
...
-    sdf.write(make_dataframe)
+    sidf = soma.IndexedDataFrame(tmp_path.as_posix())
+    sidf.create(make_dataframe.schema, index_column_names=["index"])
+    sidf.write(make_dataframe)
+
+
+# TODO:
+# * None
+# * int
+# * list
```
bkmartinjr, line 336: add:  tuple

bkmartinjr (dismissed):
Couple of issues:
* Write is still broken - it is hard wired to only accept 1 or 2 dimensions. Just fixing read means we can't round-trip test.
* The unit tests don't test anything with more than 2 dimensions (which is why the write bug was allowed)
* Recommend all python Sequence be supported for the id params, not just list/tuple
* recommend any single allow tuple, not just list

apis/python/src/tiledbsoma/indexed_dataframe.py
johnkerl: @bkmartinjr thank you -- not on this PR but on a follow-on -- this `and dtype != "ascii"` should not be on this PR -- I'll remove it. Follow-on tasks include #418 and #419 which are necessarily outside the scope of this PR as the `SOMAReader` will need to be modified to handle non-int.

apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -200,18 +201,46 @@ def read(
...
+                    dim_ids = ids[i]
+                    if dim_ids is None:  # select all in this dimension
+                        pass
+                    elif isinstance(dim_ids, int):
+                        sr.set_dim_points(dim_name, [dim_ids])
+                    elif isinstance(dim_ids, list):
+                        sr.set_dim_points(dim_name, dim_ids)
+                    elif isinstance(dim_ids, pa.ChunkedArray):
+                        sr.set_dim_points(dim_name, dim_ids)
+                    elif isinstance(dim_ids, pa.Array):
```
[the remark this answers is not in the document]
johnkerl: we get a `no such attribute: .chunk` from `SOMAReader`

i'd prefer to keep this Python PR as-is for this PR and make a follow-on issue to modify the C++ code to handle non-chunked arrays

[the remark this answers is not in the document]
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/issues/497

johnkerl:
@bkmartinjr ready for next round of review

apis/python/src/tiledbsoma/dataframe.py
bkmartinjr, line 182: As noted in #485, these would be better thrown as a TypeError, as this is a classic use case for that exception.  Per the Python docs for [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError):
> Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3/library/stdtypes.html#list) when an [int](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).

apis/python/src/tiledbsoma/types.py
```diff
@@ -31,3 +32,21 @@
...
+# * If the sequence contains missing coordinates (length less than number of dimensions),
+#   then "slice(None)" is assumed for the missing dimensions.
+# * Per-dimension, explicitly specified coordinates can be one of: None, a value, a
+#   list/ndarray/pyarrow.Array/etc of values, a slice, etc.
+
+# TODO: add support for non-ints once the libtiledbsoma SOMAReader class has supports
+# for non-ints. See also:
+# https://github.com/single-cell-data/TileDB-SOMA/issues/418
+# https://github.com/single-cell-data/TileDB-SOMA/issues/419
+SparseIndexedDataFrameCoordinate = Optional[
```
bkmartinjr, line 49: IMHO, this would be clearer if defined as explicitly accepting a None, i.e.
>     Union[None, int, slice, Sequence[int], pa.Array, pa.ChunkedArray, np.ndarray]

No change to the meaning, just clarity

apis/python/src/tiledbsoma/types.py
```diff
@@ -31,3 +32,21 @@
 
 SparseCoordinates = Union[int, slice, Tuple[int, ...], List[int], pa.IntegerArray]
 SparseNdCoordinates = Tuple[DenseCoordinates, ...]
+
```
bkmartinjr, line 35: We are accepting the burden of keeping these comments up to date (which isn't bad), but if you are adding them, I suggest:
* make them real docstrings so the user benefits from them as well
* add the same thing to the other coordinates, as they all share common principles.

apis/python/src/tiledbsoma/dataframe.py
johnkerl: Gah. I keep changing those back to `TypeError`. I suspect they're getting reset every time I `git merge main`. Sorry about that.

apis/python/src/tiledbsoma/types.py
johnkerl: @bkmartinjr I should remove them here -- they are duplicate information I left in as crib notes while writing out the union-types and forgot to remove them.

apis/python/src/tiledbsoma/dataframe.py
bkmartinjr: force-merge is a big, big hammer

apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -200,18 +199,57 @@ def read(
                 query_condition=query_condition,
             )
 
+            # Acceptable ways to index:
```
bkmartinjr, line 202: these comments belong in the function docstring, so the end user benefits (and we keep them up-to-date :-)).

apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -200,18 +199,57 @@ def read(
...
+                    )
+
+                for i, dim_ids in enumerate(ids):
+                    # Example: ids = [None, 3, slice(4,5)]
+                    # dim_ids takes on values None, 3, and slice(4,5) in this loop body.
+                    dim_name = A.schema.domain.dim(i).name
+                    if dim_ids is None:
+                        pass  # No constraint; select all in this dimension
+                    elif isinstance(dim_ids, int):
+                        # TO DO: Support non-int index types when we have non-int index support
```
bkmartinjr, line 228: If this is a very temporary work-around, then it is OK as is.  But if longer term, then...the spec pretty clearly says that an error should be raised in `create` if the schema is unsupported.  Non-int index types would fall into that category.

In this case (not a short-term work-around), I recommend you add this check (for int) to `create` - ie, if we don't currently support non-int-indexed dataframes, then it should fail _immediately_ in create, not wait until a read or write.

apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -200,18 +199,57 @@ def read(
...
+                        sr.set_dim_points(dim_name, [dim_ids])
+                    elif isinstance(dim_ids, np.ndarray):
+                        if dim_ids.ndim != 1:
+                            raise ValueError(
+                                f"only 1D numpy arrays may be used to index; got {dim_ids.ndim}"
+                            )
+                        sr.set_dim_points(dim_name, dim_ids)
+                    elif isinstance(dim_ids, slice):
+                        lo_hi = util.slice_to_range(dim_ids)
+                        if lo_hi is not None:
```
bkmartinjr, line 241: Needs clarification in docstrings -- are the slice ranges inclusive or exclusive.  Ie., like TileDB or like NumPy?   Do they support negative indexing?  Etc.

apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -200,18 +199,57 @@ def read(
...
+                            sr.set_dim_ranges(dim_name, [lo_hi])
+                        # Else, no constraint in this slot. This is `slice(None)` which is like
+                        # Python indexing syntax `[:]`.
+                    elif isinstance(
+                        dim_ids, (collections.abc.Sequence, pa.Array, pa.ChunkedArray)
```
bkmartinjr, line 246: Does the reader support ranges with a step != 1?   This implies that it does.  but the code above explicitly disallows slices with a step != 1.   That seems contradictory.

apis/python/tests/test_indexed_dataframe.py
```diff
@@ -289,6 +294,267 @@ def make_dataframe(request):
...
+
+    n_data = len(data["index1"])
+    rb = pa.Table.from_pydict(data)
+    sidf.write(rb)
+    return (schema, sidf, n_data)
+
+
+@pytest.mark.parametrize(
+    "io",
+    [
```
bkmartinjr, line 341: missing several important tests:
* ranges with a step != 1,  e.g.,   ids=[range(0,10,2)]
* empty ids list or tuple, e.g.,   ids=[]
* slices with step != 1
* slices with negative indexing
there may be more - this was a quick look....

bkmartinjr: Oh, I would add some additional expected errors, e.g.,  array of floats or other unsupported primitives, datetime, etc.

apis/python/tests/test_indexed_dataframe.py
```diff
@@ -289,6 +294,267 @@ def make_dataframe(request):
...
+    sidf = soma.IndexedDataFrame(tmp_path.as_posix())
+    sidf.create(make_dataframe.schema, index_column_names=["index"])
+    sidf.write(make_dataframe)
+
+
+def make_multiply_indexed_dataframe(tmp_path, index_column_names: List[str]):
+    """
+    Creates a variably-indexed IndexedDataFrame for use in tests below.
+    """
+    schema = pa.schema(
```
bkmartinjr, line 306: IMHO, this should support non-int indexing, and just xfail it for now.  That way we have an accurate representation of the gaps in the API (AKA test-driven development)

apis/python/tests/test_indexed_dataframe.py
```diff
@@ -289,6 +294,267 @@ def make_dataframe(request):
...
+        # 3D: indexing slot is mixed
+        {
+            "index_column_names": ["index2", "index3", "index4"],
+            "ids": [range(400, 600), None, np.asarray([2000, 9999])],
+            "A": [11],
+            "throws": None,
+        },
+    ],
+)
+def test_indexing(tmp_path, io):
```
bkmartinjr, line 536: name suggestion:  `test_read_indexing`.   We will need a write indexing test as well.

apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -291,20 +327,13 @@ def write(self, values: pa.Table) -> None:
                 attr_cols_map[name] = values.column(name).to_pandas()
         assert n is not None
 
-        dim_cols_list = [list(dim_col) for dim_col in dim_cols_list]
+        dim_cols_tuple = tuple([list(dim_col) for dim_col in dim_cols_list])
```
bkmartinjr, line 330: this wastes memory / cpu by converting to list then to a tuple.  Better form:
```
dim_cols_tuple = tuple(list(dim_col) for dim_col in dim_cols_list)
```

Also, per the docs, TileDB-Py.__setitem__ already accepts a numpy.ndarray, so you don't need to convert it to a list, so ultimately:
```
dim_cols_tuple = tuple(dim_cols_list)
```

We need to be careful with all this extra alloc & copy activity, as it can get expensive and generates GC pressure.

apis/python/tests/test_indexed_dataframe.py
johnkerl: @bkmartinjr good idea but it won't pass `typeguard`

I guess I can put non-int in the type-specs to say that string etc _are_ allowed at that level, then xfail them in the unit-test code, and do run-time assertions in `create` ...

apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -291,20 +327,13 @@ def write(self, values: pa.Table) -> None:
...
         with self._tiledb_open("w") as A:
-            # TODO: find the right syntax for vardims ... it's not the ``*`` operator ...
-            # A[*dim_cols_list] = attr_cols_map
-            if len(dim_cols_list) == 1:
-                A[dim_cols_list[0]] = attr_cols_map
-            elif len(dim_cols_list) == 2:
-                A[dim_cols_list[0], dim_cols_list[1]] = attr_cols_map
-            else:
-                raise SOMAError("ndim >= 2 not currently supported")
+            A[dim_cols_tuple] = attr_cols_map
```
bkmartinjr, line 332: I like this!  It lets us build on top of the solid core DB features....

The only caveat:  we will need to ensure that read & write coordinates stay aligned and consistent.  I'm thinking specifically of the edge cases in TileDB, eg.
* negative indexing
* inclusive vs exclusive ranges
and the like.  Fortunately, most of these are nonsense in writes, which makes this fairly simple.

This is a good argument for TDD - having tests for each case that validate read & write.

apis/python/src/tiledbsoma/types.py
bkmartinjr: You are correct as far as the typing goes, but the implication of `Optional` is that the value can be omitted.  That is _not_ the case here -- the value can never be omitted, but it can be a None.  Quite different semantics.

I'll put it another way, index values are not optional, as they are _positional_.

apis/python/tests/test_indexed_dataframe.py
bkmartinjr: I'm not sure what this means other than the interface needs to be improved.  If typeguard is throwing an error, then lets fix the interface...

Can you elaborate?  Are you claiming a bug in typeguard?

apis/python/src/tiledbsoma/types.py
johnkerl: OK @bkmartinjr I'll do as you ask; I've gotten feedback in the past from more Pythonic devs to use `Optional[X]` when what I meant was `Union[None, X]`. It's your review. :)

johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/483/commits/d0edba8d5bebc0337179359642d6b1b2d7bba9b9

apis/python/tests/test_indexed_dataframe.py
johnkerl: @bkmartinjr I'm saying if I typedef to say (as it is at present)
```
SparseIndexedDataFrameCoordinate = Union[
    None, Union[int, slice, Sequence[int], pa.Array, pa.ChunkedArray, np.ndarray]
                                     ^^^ right here
]
```
then any tests in which I try passing a string (say) coordinate will fail typeguard.

This isn't a bug in typeguard; it's typeguard doing it's job.

What I'll need to do is replace that that typedef with
```
SparseIndexedDataFrameCoordinate = Union[
    None, Union[int, slice, Sequence[xxx], pa.Array, pa.ChunkedArray, np.ndarray]
                                     ^^^ something more broad
```
and then do runtime checks as you suggest.

apis/python/src/tiledbsoma/types.py
bkmartinjr: Can i suggest you consult with Paul or Isaiah - I am not an authority, just giving you feedback about how I read the spec :-)

apis/python/src/tiledbsoma/indexed_dataframe.py
johnkerl: Temporary check added in `create()` pending #418 and #419 which have issues dedicated to them

apis/python/tests/test_indexed_dataframe.py
johnkerl: Non-int indexing will be handled in #418 and #419.

johnkerl:
@bkmartinjr ready for next round of review

apis/python/src/tiledbsoma/types.py
```diff
@@ -30,11 +30,22 @@
...
 
+# TODO: add support for non-ints once the libtiledbsoma SOMAReader class has supports
+# for non-ints. See also:
+# https://github.com/single-cell-data/TileDB-SOMA/issues/418
+# https://github.com/single-cell-data/TileDB-SOMA/issues/419
+#
 # Note: we intentionally use `Union[None, ...]` in place of `Optional[...]` since
 # we choose to emphasize that the argument-slots this is used in are not "optional"
 # arguments -- they're required argments, which can take the `None` value.
+SparseIndexedDataFrameCoordinate = Union[
```
bkmartinjr, line 41: This may be a naive question (but here goes :-):  how is the Union of a Union different than flattening it into a single Union?

I.e., why not remove one level, as in:
```
SparseIndexedDataFrameCoordinate = Union[
    None, int, slice, Sequence[int], pa.Array, pa.ChunkedArray, np.ndarray
]
```

bkmartinjr: Also, NumPy typing should use their explicit typing support IMHO.   See the [docs](https://numpy.org/doc/stable/reference/typing.html?highlight=data%20type%20api)

The difference here is that you would first 
> import numpy.typing as npt

And then declare this as:
```
SparseIndexedDataFrameCoordinate = Union[
    None, int, slice, Sequence[int], pa.Array, pa.ChunkedArray, npt.NDArray[np.int_]
]
```

If you don't want to bite this off in this PR, just capture as an open issue. I would understand!

bkmartinjr (approved):
A couple of cleanup requests (the duplicated error checks, etc), and some optional stuff that I'm fine if you capture in an Issue for future work (eg, numpy.typing).  

No major concerns, so marking  Approved on the assumption the minor stuff will get be disposed of :-)

Thanks for the great work!

apis/python/src/tiledbsoma/types.py
johnkerl: #517 -- thank you!

apis/python/tests/test_indexed_dataframe.py
```diff
@@ -289,6 +294,283 @@ def make_dataframe(request):
...
+    """
+    Creates a variably-indexed IndexedDataFrame for use in tests below.
+    """
+    schema = pa.schema(
+        [
+            # TO DO: Support non-int index types when we have non-int index support
+            # in libtiledbsoma's SOMAReader. See also
+            # https://github.com/single-cell-data/TileDB-SOMA/issues/418
+            # https://github.com/single-cell-data/TileDB-SOMA/issues/419
+            ("index1", pa.int64()),
```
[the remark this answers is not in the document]
johnkerl: Updated `test_index_types` -- thank you!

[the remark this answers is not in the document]
johnkerl: See also #518

apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -201,17 +215,62 @@ def read(
...
+                        if dim_ids.ndim != 1:
+                            raise ValueError(
+                                f"only 1D numpy arrays may be used to index; got {dim_ids.ndim}"
+                            )
+                        sr.set_dim_points(dim_name, dim_ids)
+                    elif isinstance(dim_ids, slice):
+                        lo_hi = util.slice_to_range(dim_ids)
+                        if lo_hi is not None:
+                            lo, hi = lo_hi
+                            if lo < 0 or hi < 0:
```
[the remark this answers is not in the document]
johnkerl: #519 and #520

apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -201,17 +215,62 @@ def read(
...
+                            sr.set_dim_ranges(dim_name, [lo_hi])
+                        # Else, no constraint in this slot. This is `slice(None)` which is like
+                        # Python indexing syntax `[:]`.
+                    elif isinstance(
+                        dim_ids, (collections.abc.Sequence, pa.Array, pa.ChunkedArray)
+                    ):
+                        if (
+                            dim_ids == []
+                        ):  # TileDB-Py maps [] to all; we want it to map to none.
+                            # TODO: find out how to best implement this
```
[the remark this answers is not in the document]
johnkerl: @bkmartinjr indeed, and the work is returning an empty Arrow table with the correct schema, which is non-trivial to get from `TileDB-Py` without doing some kind of query -- cc @nguyenv -- issue linked

PR #485, opened 2022-10-27
Update slicing logic for SparseNdArray

johnkerl:
## Context

Addresses #479. See also related discussion on #470.

## Fix

As noted on #479, desired indexing semantics:

* A sequence of coordinates is accepted, one per dimension.
* Sequence length must be at least one and <= number of dimensions.
* If the sequence contains missing coordinates (length less than number of dimensions), then "slice(None)" is assumed for the missing dimensions.
* Per-dimension, explicitly specified coordinates can be one of: a value, a list/ndarray/paarray/etc of values, a slice, etc.
* `raise` on anything else.


apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -179,8 +180,33 @@ def read_sparse_tensor(
...
 
+        # The full shape of the data is a template for the shape of the answer.
+        # For example if the data is 2D and coords is (None, [5,7,9]), then
+        # output shape[0] will be taken from the full shape, and output shape[1] will
+        # be 3.
         with self._tiledb_open("r") as A:
-            shape = A.shape
+            shape = list(A.shape)
+
+        for i in range(len(coords)):
```
bkmartinjr, line 190: Rewriting the shape fairly significantly changes the behavior of this method, and I believe is not something we want to do.

Consider the following example:
* I have a sparse nd array that is shaped (100, 100)
* I call this method with:     `nda.read_sparse_tensor(( 8, 99 ), format="coo")`

This will return a SparseCOOTensor with:
* one value, at coordinate (8, 99)
* a shape of (1,1)
That doesn't make much sense.

I believe the original code was correct - the returned tensor should have the same shape as the soma.SparseNdArray, just fewer values.  This preserves the coordinate information, and keeps it self-consistent.

I'm also fairly certain that the current code will generate a SciPy error if you were to ask for `format="csr"`, as Scipy Sparse actually checks that coordinates are inside the shape limits.

bkmartinjr (changes requested):
See comment in read_sparse_tensor, which I believe introduces a bug

apis/python/src/tiledbsoma/sparse_nd_array.py
bkmartinjr: Also, the fact that this didn't fail in our unit tests highlights another case of where we need additional coverage!

johnkerl:
Failing unit-test cases are due to #493 which blocks this PR

johnkerl:
> Failing unit-test cases are due to https://github.com/single-cell-data/TileDB-SOMA/issues/493 which blocks this PR

Resolved; the bug was in my unit-test code on this PR which uses `pyarrow.SparseCOOTensor` as a data-populator mechanism which removes zero cells.

johnkerl:
Needs test cases for `read_sparse_tensor`

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -179,8 +180,13 @@ def read_sparse_tensor(
         if format not in ("coo", "csr", "csc"):
             raise NotImplementedError("format not implemented")
 
+        # The full shape of the data is a template for the shape of the answer.
+        # For example if the data is 2D and coords is (None, [5,7,9]), then
+        # output shape[0] will be taken from the full shape, and output shape[1] will
+        # be 3.
         with self._tiledb_open("r") as A:
-            shape = A.shape
+            shape = list(A.shape)
```
bkmartinjr, line 188: no reason to convert the tuple to a list - the API below where you use it all accept a tuple AFAIK

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -230,23 +236,29 @@ def read_table(self, coords: SparseNdCoordinates) -> Iterator[pa.Table]:
...
-                raise ValueError(
-                    f"coordinate length {len(coords)} != array ndim {A.schema.domain.ndim}"
+            # * A sequence of coordinates is accepted, one per dimension.
+            # * Sequence length must be at least one and <= number of dimensions.
+            # * If the sequence contains missing coordinates (length less than number of dimensions),
+            #   then "slice(None)" is assumed for the missing dimensions.
+            # * Per-dimension, explicitly specified coordinates can be one of: a value, a
+            #   list/ndarray/paarray/etc of values, a slice, etc.
+
+            if not (isinstance(coords, list) or isinstance(coords, tuple)):
```
bkmartinjr, line 246: a better way to write this:   `isinstance(coords, (list, tuple))`

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -230,23 +236,29 @@ def read_table(self, coords: SparseNdCoordinates) -> Iterator[pa.Table]:
...
+                # if not isinstance(coords, tuple):
```
bkmartinjr, line 247: remove comment

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -230,23 +236,29 @@ def read_table(self, coords: SparseNdCoordinates) -> Iterator[pa.Table]:
...
 
+            for i in range(len(coords)):
                 dim_name = A.schema.domain.dim(i).name
-                if isinstance(coord, int):
+                coord = coords[i]
+                if coord is None:
+                    pass
+                elif isinstance(coord, int):
                     sr.set_dim_points(dim_name, [coord])
                 elif isinstance(coord, list):
```
bkmartinjr, line 263: this if/else doesn't handle tuple, which should be a supported coordinate type (just a read-only list)

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -259,9 +271,7 @@ def read_table(self, coords: SparseNdCoordinates) -> Iterator[pa.Table]:
                     if lo_hi is not None:
                         sr.set_dim_ranges(dim_name, [lo_hi])
                 else:
-                    raise ValueError(
-                        f"could not handle coordinate with value {coord} of type {type(coord)}"
-                    )
+                    raise SOMAError(f"coord type {type(coord)} at slot {i} unhandled")
```
bkmartinjr, line 274: This should be a TypeError, not a SOMAError.  In general, don't use SOMAError when there is a more corect built-in exception class.

See: https://docs.python.org/3/library/exceptions.html#TypeError

Relevant quote:
> Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3/library/stdtypes.html#list) when an [int](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -230,23 +236,29 @@ def read_table(self, coords: SparseNdCoordinates) -> Iterator[pa.Table]:
...
+            # * A sequence of coordinates is accepted, one per dimension.
+            # * Sequence length must be at least one and <= number of dimensions.
+            # * If the sequence contains missing coordinates (length less than number of dimensions),
+            #   then "slice(None)" is assumed for the missing dimensions.
+            # * Per-dimension, explicitly specified coordinates can be one of: a value, a
+            #   list/ndarray/paarray/etc of values, a slice, etc.
+
+            if not (isinstance(coords, list) or isinstance(coords, tuple)):
+                # if not isinstance(coords, tuple):
+                raise SOMAError(
```
bkmartinjr, line 248: This should be a TypeError, not a SOMAError.  See: https://docs.python.org/3/library/exceptions.html#TypeError  (last para in that section).

apis/python/src/tiledbsoma/types.py
```diff
@@ -30,4 +30,4 @@
 DenseNdCoordinates = Tuple[DenseCoordinates, ...]
 
 SparseCoordinates = Union[int, slice, Tuple[int, ...], List[int], pa.IntegerArray]
-SparseNdCoordinates = Tuple[DenseCoordinates, ...]
+SparseNdCoordinates = Union[Tuple[DenseCoordinates, ...], List[DenseCoordinates]]
```
bkmartinjr, line 33: why not use Sequence, ie,   Sequence[DenseCoordinates] rather than the union of tuple & list?

apis/python/tests/test_sparse_nd_array.py
```diff
@@ -257,6 +257,43 @@ def test_sparse_nd_array_read_write_sparse_tensor(
     assert tensors_are_same_value(t, data)
 
 
+@pytest.mark.parametrize(
```
bkmartinjr, line 260: Unless I'm missing something, this test is an almost verbatim copy of the previous test.  The same thing could be accomplished by adding the "corner read" test to the above, eg, tacking onto the above test:

```
    # Read back and validate the "corner"
    b = soma.SparseNdArray(tmp_path.as_posix())
    t = next(b.read_sparse_tensor((0,) * len(shape), format=format))
    assert t.shape == shape
```

Rather than duplicating the whole thing.

Apologies if my visual scan missed something else different...

apis/python/tests/test_sparse_nd_array.py
```diff
@@ -388,3 +425,159 @@ def test_sparse_nd_array_reshape(tmp_path):
...
+                90301,
+                90302,
+                90303,
+                90304,
+                90305,
+            ],
+        },
+    ],
+)
+def test_sparse_nd_array_table_slicing(tmp_path, io):
```
bkmartinjr, line 559: Comments on this test:
1. this needs to test ndim other than 2.   Ie, 1, 2, 3, ...
2. You could simplify this a lot if you programatically generated the data contents (eg, use numpy random) - that would radically simplify the parameterization.
3. There are legal coordinates that are not tested, eg, hybrid combinations of types, pa.chunked_array, pa.array, tuple, etc.   Example:      (1, slice(2)).

bkmartinjr (dismissed):
Several requested changes specified inline.

Also, PR still marked as a draft :-)

johnkerl:
Thanks for the feedback @bkmartinjr ! I moved this back to draft while I work off action items ... I'm still working them off iteratively.

In addition to having moved this back to draft status, I'll remove you as reviewer for a while until I 'catch up', to avoid you getting GitHub notifications for things I'm still iterating on.

johnkerl:
@bkmartinjr Gah! GitHug won't let me remove you as reviewer since you've reviewed and requested changes. Even with re-request review and the attempt to remove you; it doesn't stick.

Can I ask you to hold off further review comments until I address all your wonderful feedback so far? When I'm ready for you again I'll move this back out of draft status.

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -259,9 +271,7 @@ def read_table(self, coords: SparseNdCoordinates) -> Iterator[pa.Table]:
                     if lo_hi is not None:
```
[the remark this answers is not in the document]
johnkerl: added comment in the code

johnkerl:
@bkmartinjr ready for next round of review

apis/python/tests/test_sparse_nd_array.py
```diff
@@ -388,3 +395,306 @@ def test_sparse_nd_array_reshape(tmp_path):
     a.create(type=pa.int32(), shape=(10, 10, 10))
     with pytest.raises(NotImplementedError):
         assert a.reshape((100, 10, 1))
+
+
+@pytest.mark.parametrize(
```
bkmartinjr, line 400: A small nit - in the interest of making this easier to maintain.  Unrolled parameter lists like this are quite challenging to review which cases are covered. It would be easier to read if the options were constructed as a series of loops, eg,

```
for write_format in ['coo']:
    for read_format in ['table', 'csr']:
        for shape in [(4,), (4,6,)]:
            ...
```

I realize that not of the conditions can be created this way.  But even subsets done this way are far more obvious, and can then be concatenated.

As example of why it matters: the tests seem to be missing a passing (non-error case) for `read_format=csr` or `csc`.  

Code coverage would also catch this, but it would be nice if it was easier to visually inspect.

bkmartinjr (approved):
One nit about the complexity of the test case param generation (and a suggestion of how to fix it).  Otherwise, thumbs-up

Issue #467, opened 2022-10-25
Replace Exception with SOMAError or derived classes


johnkerl:
Follow-up from https://github.com/single-cell-data/TileDB-SOMA/pull/464#discussion_r1004999589

johnkerl:
#486 is merged

PR #486, opened 2022-10-27
Replace Exception with SOMA-specific exceptions

johnkerl:
Addresses #467

### DIFF
--- a/apis/python/src/tiledbsoma/io.py
+++ b/apis/python/src/tiledbsoma/io.py
@@ -24,6 +24,7 @@ from tiledbsoma import (
 )
 
 from .constants import SOMA_JOINID, SOMA_ROWID
+from .exception import SOMAError
 from .types import Path
 from .util import uri_joinpath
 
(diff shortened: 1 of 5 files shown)


apis/python/src/tiledbsoma/io.py
```diff
@@ -340,7 +341,7 @@ def from_anndata(
     Top-level writer method for creating a TileDB group for a ``Experiment`` object.
     """
     if not isinstance(anndata, ad.AnnData):
-        raise Exception(
+        raise SOMAError(
```
bkmartinjr, line 344: should be a ValueError, not a SOMAError

CC: @thetorpedodog

apis/python/src/tiledbsoma/io.py
```diff
@@ -53,7 +54,7 @@ def _from_h5ad_common(
     Common code for things we do when processing a .h5ad file for ingest/update.
     """
     if isinstance(input_path, ad.AnnData):
-        raise Exception("Input path is an AnnData object -- did you want from_anndata?")
+        raise SOMAError("Input path is an AnnData object -- did you want from_anndata?")
```
bkmartinjr, line 57: ValueError or TypeError, not SOMAError

bkmartinjr (commented):
Generally LGTM except we should not be using SOMAError when Python already has a bulit-in type with the right semantics (eg, parameter value has the wrong type).

Those should be (I think) a TypeError.

bkmartinjr (approved):
approving, with the assumption that the two correction (TypeError, rather than SOMAError) are made.

PR #496, opened 2022-11-01
Build flow improvements

gspowley:
A number of build flow improvements based on developer feedback:
1. Document a reliable way to build `tiledbsoma`, including after pulling the latest source code.
2. Add a developer `Makefile` to automate build commands.
3. Add more documentation on components of build flow (see [libtiledbsoma/README.md](https://github.com/single-cell-data/TileDB-SOMA/blob/gspowley/sc-22584/build-flow-improvements/libtiledbsoma/README.md)).
4. Provide a way to build only a `libtiledbsoma` static library for R.
5. Merge C++ tests into Python CI to reduce redundant CI runs.

Example C++ dev commands:
```bash
# setup python virtual env
python -m venv test/tiledbsoma
source test/tiledbsoma/bin/activate

# install developer requirements
pip install -r apis/python/requirements_dev.txt

# install and test tiledbsoma
make install test
```

R build command:
```bash
# requires cmake >= 3.21
make r-build
```


apis/python/tests/__init__.py
```diff
@@ -1,6 +1,9 @@
 import pyarrow as pa
 from typeguard.importhook import install_import_hook
 
+# avoid typeguard by importing before calling install_import_hook
```
johnkerl, line 4: ❤️

libtiledbsoma/README.md
```diff
@@ -1,59 +1,86 @@
...
+sudo apt install -y python3 python-is-python3 python3.10-venv python-dev-is-python3
+```
+---
+## Clone the source code
+```
+git clone https://github.com/single-cell-data/TileDB-SOMA.git
+cd TileDB-SOMA
+```
+---
+## Setup a Python Virtual Environment
```
johnkerl, line 21: nit: Set up a ...

eddelbuettel:
Looks fabulous at a first glance.  Will test-drive after dinner.  Love the `Makefile` which I find easier to read / scan / understand that the verbosity of 'that other flavor'.

libtiledbsoma/README.md
```diff
@@ -1,59 +1,86 @@
...
 python -m venv test/tiledbsoma
 source test/tiledbsoma/bin/activate
+pip install -r apis/python/requirements_dev.txt
 ```
-
-## Clean
-
-Remove old build files:
+---
+## Python Only Development
```
johnkerl, line 29: nit: Python-only Development

scripts/bld
```diff
@@ -2,61 +2,15 @@
...
 
-BUILD_TYPE=${1:-Release} # default to Release build
-if [ ! -z "$LIBTILEDBSOMA_DEBUG_BUILD" ]; then
-  echo
-  echo ================================================================
-  echo LIBTILEDBSOMA SCRIPTS/BLD BUILD_TYPE ${BUILD_TYPE}
-  echo ================================================================
-else
-  echo "Building ${BUILD_TYPE} build"
-fi
```
johnkerl, line 19: I found the output from this verbose setup _far_ easier to parse -- while it lasted. :( 

👋 ✌️ 🖖

Makefile
```diff
@@ -0,0 +1,62 @@
+# This Makefile captures common developer build and test use cases.
+
+MAKEFLAGS += --no-print-directory
+
+# - incrementally compile and install the c++ code
+# - copy the shared objects on top of the python dev build
+.PHONY: update
```
johnkerl, line 7: Linux error:

```
$ make update
make[1]: /tmp/pip-build-env-imfu1ozp/overlay/local/lib/python3.10/dist-packages/cmake/data/bin/cmake: No such file or directory
make[1]: *** [Makefile:180: cmake_check_build_system] Error 127
make: *** [Makefile:9: update] Error 2
```

MacOS error:

```
johnkerl@Kerl-MBP[prod][gspowley/sc-22584/build-flow-improvements][TileDB-SOMA]$ make update
[ 50%] Built target ep_spdlog
[ 56%] Performing configure step for 'libtiledbsoma'
-- Using default install prefix /Users/johnkerl/git/single-cell-data/TileDB-SOMA/dist. To control CMAKE_INSTALL_PREFIX, set OVERRIDE_INSTALL_PREFIX=OFF
-- Install prefix is /Users/johnkerl/git/single-cell-data/TileDB-SOMA/dist.
-- Skipping search for TileDB, building it as an external project. To use system TileDB, set FORCE_BUILD_TILEDB=OFF
-- Starting TileDB-SOMA regular build.
-- Performing Test HAVE_AVX2
-- Performing Test HAVE_AVX2 - Success
-- Starting TileDB-SOMA build.
-- Found TileDB: TILEDB_LIB-NOTFOUND
-- Building with commit hash 4ba0d1e
CMake Warning at src/CMakeLists.txt:57 (message):
  Building TileDB without deprecation warnings


CMake Error at src/pyapi/CMakeLists.txt:3 (find_package):
  Could not find a package configuration file provided by "pybind11" with any
  of the following names:

    pybind11Config.cmake
    pybind11-config.cmake

  Add the installation prefix of "pybind11" to CMAKE_PREFIX_PATH or set
  "pybind11_DIR" to a directory containing one of the above files.  If
  "pybind11" provides a separate development package or SDK, be sure it has
  been installed.


-- Configuring incomplete, errors occurred!
See also "/Users/johnkerl/git/single-cell-data/TileDB-SOMA/build/libtiledbsoma/CMakeFiles/CMakeOutput.log".
See also "/Users/johnkerl/git/single-cell-data/TileDB-SOMA/build/libtiledbsoma/CMakeFiles/CMakeError.log".
make[3]: *** [libtiledbsoma-prefix/src/libtiledbsoma-stamp/libtiledbsoma-configure] Error 1
make[2]: *** [CMakeFiles/libtiledbsoma.dir/all] Error 2
make[1]: *** [all] Error 2
make: *** [update] Error 2
```

Note this happened on MacOS even after `pip install -r apis/python/requirements_dev.txt`

Makefile
```diff
@@ -0,0 +1,62 @@
...
+# - install the python in editable mode (dev build)
+.PHONY: install-debug
+install-debug: clean
+	BUILD_TYPE=Debug ./scripts/bld
+	pip install -v -e apis/python
+
+# - run the clean rule
+# - build the c++ static library for r
+.PHONY: r-build
+r-build: clean
```
johnkerl, line 33: ✅

Makefile
```diff
@@ -0,0 +1,62 @@
...
+	R_BUILD=1 ./scripts/bld
+
+# - run c++ unit tests
+# - run pytests
+.PHONY: test
+test: test/soco
```
johnkerl, line 39: ✅

johnkerl (changes requested):

scripts/bld
gspowley: I found all of the debug messages made the code harder to modify/maintain.

Makefile
```diff
@@ -0,0 +1,62 @@
...
+		--soco \
+		-o test/soco \
+		-n \
+		data/pbmc3k_processed.h5ad \
+		data/10x-pbmc-multiome-v1.0/subset_100_100.h5ad
+
+# - remove the c++ build and dist, and the test data
+.PHONY: clean
+clean:
+	rm -rf build dist test/soco
```
eddelbuettel, line 56: Removing `test/soco` here may not be a good idea. I have treated this as a 'constant' and relied on it for testing.

Following some other idioms we could add `distclean` to really wipe things away (including this) but make the test data persist.

eddelbuettel (changes requested):
The `make r-build` target is close in that it gets the desired static library and the header -- but it misses the `externals/` library we need for `span` under C++17.

Makefile
gspowley: Good point. The goal was to force recreating the test data when moving to newer code, in case the format changed.

A better approach may be to always write new test data when running `make install`.

scripts/bld
johnkerl: @gspowley which I knew would happen hence 🖖

johnkerl (approved):

gspowley:
> The `make r-build` target is close in that it gets the desired static library and the core headers -- but it misses the `externals/` header directory we need for `span` under C++17.

cmake >= 3.23 makes this easier to support. Does this install tree capture everything?

```
dist/
├── include
│   ├── externals
│   │   └── span
│   │       └── span.hpp
│   └── tiledbsoma
│       ├── array_buffers.h
│       ├── column_buffer.h
│       ├── common.h
│       ├── logger_public.h
│       ├── managed_query.h
│       ├── soma_reader.h
│       ├── tiledbsoma
│       ├── tiledbsoma_export.h
│       └── util.h
└── lib
    ├── libtiledb.so.2.12
    └── libtiledbsoma.a
```

eddelbuettel:
I will need a newer `cmake` first to see.  The image looks good.

eddelbuettel:
_Iff_ the additional headers are added we can build, and link, and load.  Note that in the what I quote below we now only compile the r-api source files (for now all starting with `r` namely `rinterface.cpp`, `riterator.cpp` and `rutilities.cpp`) plus the Rcpp-auto-generated glue `RcppExports.cpp`.  We no longer compile the `libtiledbsoma` files, instead we link with `libtiledbsoma.a`.  

And the R package does a full 'does it turn on?' test by trying to load at the penultimate step so this really is good.

(But I did have to add extra headers in directory `thread_pool/` and `pyapi`.  So we are close.  We can see the cigar.)

eddelbuettel:
Gave it another two spins to ascertain that

1. It does not break the current use, which it passes (and which is of course what CI shows us)

2. It does what it is supposed to do namely provide a `libtiledbsoma.a` plus headers we can use in `apis/r` and that passes too.

So :ship: from me!  Nice work.

PR #504, opened 2022-11-03
[r] Add coverage reports in script and yaml

eddelbuettel:
This PR enables code coverage for the R package and submits the report under flag 'r' with the repo token for codecov.io.

The actual coverage analysis is provided by the widely-used CRAN package [covr](https://cloud.r-project.org/package=covr) which is loaded when coverage is requested (when running on Linux).

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -5,6 +5,9 @@ on:
   push:
     branches: [main]
 
+env:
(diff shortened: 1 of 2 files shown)


eddelbuettel:
Turns out it works _if and only if_ we do *not* set a codecov token.  Lesson learned, coverage uploaded, merging.

PR #505, opened 2022-11-03
[r] Arrow conversion utilities

aaronwolen:
This adds internal helpers for converting TileDB dimensions, attributes, and schemas to their Arrow equivalents. We'll use these in a future PR to return Arrow schemas from the foundational classes.

Also tweaked `TileDBArray$dimensions()` to return a named list, which matches the behavior of `TileDBArray$attributes()`.

### DIFF
--- a/apis/r/tests/testthat/test-Arrow-utils.R
+++ b/apis/r/tests/testthat/test-Arrow-utils.R
@@ -1,0 +1,70 @@
+test_that("TileDB classes can be converted to Arrow equivalents", {
+
+  # Dimension to Arrow field
+  dim0 <- tiledb::tiledb_dim(
+    name = "dim0",
+    domain = NULL,
+    tile = NULL,
+    type = "ASCII"
+  )
+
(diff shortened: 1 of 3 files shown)


eddelbuettel (approved):
Most of `arrow` and its types is still a little mysterious to me.  I am at ease with some of the 'from C' constructors.  

These functions seem fine too.

apis/r/tests/testthat/test-Arrow-utils.R
```diff
@@ -0,0 +1,82 @@
...
+  )
+
+  dim1 <- tiledb::tiledb_dim(
+    name = "dim1",
+    domain = bit64::as.integer64(c(0, 100)),
+    tile = bit64::as.integer64(10),
+    type = "INT64"
+  )
+
+  # Error if not a dimension
```
eddelbuettel, line 18: I think you want 'Error if not an attribute' there.

johnkerl (approved):

PR #506, opened 2022-11-03
Parallelize obs & var within soma.query

johnkerl:
These two are independent of one another and may be run in parallel.

For a very small SOMA and small query sizes, this takes the total `soma.query` time (averaged over 10 trials of before and after) from 5.7 seconds to 3.8 seconds.

For a larger SOMA and larger query sizes, the difference was not noticeable with variation of query times.


aaronwolen (approved):
Good idea!

PR #511, opened 2022-11-08
Parallelized ingest

johnkerl:


gspowley (approved):
LGTM!

apis/python/src/tiledbsoma/io.py
```diff
@@ -189,66 +190,123 @@ def _from_anndata_aux(
...
-        layer_name=X_layer_name,
-    )
-
-    if anndata.layers is not None:
-        for layer_name in anndata.layers:
-            soma.X.add_layer_from_matrix_and_dim_values(
-                matrix=anndata.layers[layer_name],
+    futures = []
+    with ThreadPoolExecutor(max_workers=max_workers) as executor:
+        futures.append(
```
thetorpedodog, line 198: Any reason this isn't

```python
futures = [
    executor.submit(...),
    executor.submit(...),
    ...,
]
```
?

apis/python/src/tiledbsoma/io.py
```diff
@@ -189,66 +190,123 @@ def _from_anndata_aux(
...
-        soma.varp.add_matrix_from_matrix_and_dim_values(
-            anndata.varp[key], anndata.var_names, key
-        )
-    soma._add_object(soma.varp)
 
-    # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+    futures = []
+    with ThreadPoolExecutor(max_workers=max_workers) as executor:
+        for key in anndata.obsm.keys():
+            futures.append(
```
thetorpedodog, line 251: consider

```python
futures.extend(
    executor.submit(
        ...,
        obsm_value,
        ...,
        key,
    )
    for key, value in anndata.obsm.items()
)

File: apis/python/src/tiledbsoma/tiledb_group.py [hunk 3]
```diff
@@ -53,8 +53,15 @@ def create_unless_exists(self) -> None:
...
-            self._set_object_type_metadata()
+            try:
+                tiledb.group_create(uri=self.uri, ctx=self._ctx)
+                self._set_object_type_metadata()
+            except tiledb.cc.TileDBError as e:
+                stre = str(e)
+                if (
+                    "already exists" not in stre
+                ):  # This is fine in case of parallel creates
+                    raise e
```
thetorpedodog, line 64: Use bare `raise` here; that will raise the current exception without rewriting the stack trace.
```suggestion
                    raise
```

apis/python/src/tiledbsoma/tiledb_group.py
```diff
@@ -53,8 +53,15 @@ def create_unless_exists(self) -> None:
...
-            tiledb.group_create(uri=self.uri, ctx=self._ctx)
-            self._set_object_type_metadata()
+            try:
+                tiledb.group_create(uri=self.uri, ctx=self._ctx)
+                self._set_object_type_metadata()
+            except tiledb.cc.TileDBError as e:
+                stre = str(e)
+                if (
+                    "already exists" not in stre
+                ):  # This is fine in case of parallel creates
```
thetorpedodog, line 63: Formatting and comment here is kind of weird to me. Is this intended to mean something like:

```python
if "already exists" not in str(e):
    # Parallel creation is OK; other cases are genuine errors.
    raise
```

apis/python/src/tiledbsoma/uns_group.py
```diff
@@ -145,71 +213,23 @@ def from_anndata_uns(self, uns: Mapping[str, Any]) -> None:
...
-                array.from_pandas_dataframe(value)
-                self._add_object(array)
-
-            elif isinstance(value, sp.csr_matrix):
-                array.from_scipy_csr(value)
-                self._add_object(array)
-
-            elif array._maybe_from_numpyable_object(value):
-                self._add_object(array)
+            threadsafe_args[key] = (key, child_uris[key], uns[key])
```
thetorpedodog, line 221: I don't follow why we're doing this here; it seems like you could do the same thing with `executor.submit` without building this dict first.

thetorpedodog (commented):
Correctness-wise this all seems good. I have some organization/style suggestions and would be happy to see them addressed, but they’re not critical and need not block submission if you would prefer not to.

apis/python/src/tiledbsoma/io.py
johnkerl: Thanks for sharing your stylistic preference!

apis/python/src/tiledbsoma/io.py
johnkerl: Thanks for sharing your stylistic preference!

PR #522, opened 2022-11-10
[c++] Improve build experience

gspowley:
Update build scripts based on more feedback, including better support for building with a custom build of TileDB.

```
Usage: make rule [options]

Rules:
  install [options]   Build C++ library and install python module
  r-build [options]   Build C++ static library with "#define R_BUILD" for R
  update              Incrementally build C++ library and update python module
  test                Run tests
  clean               Remove build artifacts

Options:
  build=BUILD_TYPE    Cmake build type = Release|Debug|RelWithDebInfo|Coverage [Release]
  prefix=PREFIX       Install location [dist]
  tiledb=TILEDB_DIST  Absolute path to custom TileDB build 

Examples:
  Install Release build

    make install

  Install Debug build of libtiledbsoma and libtiledb

    make install build=Debug

  Install Release build with custom libtiledb

    make install tiledb=$PWD/../TileDB/dist
    export LD_LIBRARY_PATH=$PWD/../TileDB/dist/lib:$LD_LIBRARY_PATH     # linux
    export DYLD_LIBRARY_PATH=$PWD/../TileDB/dist/lib:$DYLD_LIBRARY_PATH # macos

  Incrementally build C++ changes and update the python module

    make update

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,46 +2,34 @@
 
 MAKEFLAGS += --no-print-directory
 
-# - incrementally compile and install the c++ code
-# - copy the shared objects on top of the python dev build
-.PHONY: update
-update:
-	@cd build && make -j && make install-libtiledbsoma
-	cp dist/lib/* apis/python/src/tiledbsoma/
+# print help by default
+help:
 
-# - run the clean rule
-# - build the c++ code using the environment's cmake (not the python build-system cmake)
-#   - builds a Release build
-# - install the python in editable mode (dev build)
+# install 
+# -------------------------------------------------------------------
 .PHONY: install
 install: clean
-	./scripts/bld
-	pip install -v -e apis/python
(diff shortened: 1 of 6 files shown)


eddelbuettel:
Nice!  

Now ... just thinking out loud looking at the options:  `r-build` is really two things namely 
- build a static library (nothing R specific about that) and 
- build with a `#define` to accomodate R.  

Should we split those?   

(Likely answer: let's not as nobody needs it, so better answer may just to document what `r-build` does....)

gspowley:
Right, other than R, we don't have a need for a static library yet. We can document what `r-build` does now, and split out the static library build in the future as needed.

eddelbuettel (approved):
:+1:

Also nice to learnsome new `make` tricks after 3+ decades of using it.

PR #525, opened 2022-11-10
[c++] Cmake bug fix and improvements

gspowley:
Cmake will search the system for TileDB. If not found, it will build (or download) the version specified in `libtiledbsoma/cmake/Modules/FindTileDB_EP.cmake`.

Use this command to explicitly build using a version of TileDB installed on the system:
```
make install tiledb=<absolute-path-to-tiledb-dist>
```

Setting `LD_LIBRARY_PATH` is no longer required.


Shelnutt2:
Why do we want to block system installs of TileDB by default? This is not how we handle other applications. I believe we should always support detecting system installs automatically.

eddelbuettel:
I generally only ever install one TileDB Core version (and use `dev` for it) and generally have not had any issues with that approach.  And `make r-build` still works as it did before so no objection from me.

johnkerl (approved):

johnkerl (changes requested):
Sorry, premature approve -- I missed Seth's very on-point feedback

gspowley:
The latest commit reverts to searching the system for TileDB by default. 

Cmake VERSION ranges could help avoid linking to an old TileDB version, but they don't support specifying a minimum version number.

johnkerl (approved):

PR #526, opened 2022-11-10
Update R version in prep for package tag

johnkerl:


eddelbuettel (approved):
We may want a script to update both API package versions.

aaronwolen (approved):
Thanks

PR #528, opened 2022-11-14
DataFrame read result_order

mlin:


apis/python/src/tiledbsoma/types.py
```diff
@@ -16,7 +16,7 @@
 BatchFormat = Literal["dense", "coo", "csr", "csc", "record-batch", "table"]
 ReadPartitions = Literal["IofN"]
 BatchSize = Literal["count", "size", "auto"]
-ResultOrder = Literal["row-major", "column-major", "unordered", "rowid-ordered"]
+ResultOrder = Literal["row-major", "column-major", "auto"]
```
mlin, line 19: @bkmartinjr @johnkerl I didn't see an obvious way to implement "rowid-ordered" by pushing down into TileDB; care to correct me on that? Is it important enough to support to either (i) sort ourselves in-memory or (ii) detect if soma_joinid is the first or second indexed dimension and set row-major or column-major accordingly?

General background question...say we've indexed three or more dimensions...are the "rows" and "columns" simply the first two, as far as TileDB is concerned? No way to ask for results ordered by the third or higher dimension? (I think I can guess reasons for this based on the first two dimensions determining the on-disk layout, but just curious if something else distinguishes the two dimensions "rows" and "columns")

libtiledbsoma/src/soma_reader.cc
```diff
@@ -97,6 +96,18 @@ SOMAReader::SOMAReader(
     if (!column_names.empty()) {
         mq_->select_columns(column_names);
     }
+
+    if (result_order != "auto") {  // default "auto" is set in soma_reader.h
```
mlin, line 100: @gspowley @johnkerl style question...on the one hand, in pure C/C++ usually one tries to use an enum or something instead of comparing string constants. On the other hand, I can see that taking `result_order` (and eventually `batch_size`) as strings is convenient for calling in from Python & R though. Is that the winning "hand"?

apis/python/src/tiledbsoma/types.py
bkmartinjr: > "rowid-ordered"

Correct - this is unsupported for anything other than DenseNdArray (or the original "dense" offset-indexed dataframe, which is scuttled for now).

> General background question

Nope..."row" is a synonym for C order, "col" for Fortran order.  All dimensions (ie. the entire hypercube) are ordered, either last dimension to first dimension (C or row order), or first dimension to last dimension (Fortran or column order)

libtiledbsoma/src/soma_reader.cc
gspowley: LGTM, I like the convenience of comparing string constants in this case.

libtiledbsoma/include/tiledbsoma/managed_query.h
```diff
@@ -146,6 +146,10 @@ class ManagedQuery {
         query_->set_condition(qc);
     }
 
+    void set_layout(tiledb_layout_t layout) {
```
gspowley, line 149: Please add a brief comment similar to others in this file. They're handy for code editors that read the comments.

gspowley (changes requested):
LGTM! Requesting a comment for a new function.

libtiledbsoma/include/tiledbsoma/managed_query.h
mlin: good catch @gspowley, updated

libtiledbsoma/src/soma_reader.cc
johnkerl: +1

gspowley (approved):
LGTM, thanks @mlin!

johnkerl (approved):

eddelbuettel:
I am coming it a little late but if this is exposed in `managed_query.h`, can be also (by copying the interace) in `soma_reader.h` ?  That is one I use from R to instantiate a `SOMAReader` object amd I do not see `managed_query` directly.

mlin:
@eddelbuettel The `result_order` argument was already stubbed in the `SOMAReader` constructor, it just wasn't actually applying the setting yet. So after landing this you should be able to pass "row-major", "column-major" or (default) "auto" into that existing constructor argument.

PR #529, opened 2022-11-14
improving IndexedDataFrame test coverage

mlin:
Nothing too exciting here...just some small/cheesy tweaks to the unit tests to hit remaining uncovered code paths in indexed_dataframe.py

### DIFF
--- a/apis/python/src/tiledbsoma/indexed_dataframe.py
+++ b/apis/python/src/tiledbsoma/indexed_dataframe.py
@@ -251,10 +251,7 @@ class IndexedDataFrame(TileDBArray):
                                 raise ValueError(
                                     f"slice start and stop may not be negative; got ({lo}, {hi})"
                                 )
-                            if lo > hi:
-                                raise ValueError(
-                                    f"slice start must be <= slice stop; got ({lo}, {hi})"
-                                )
+                            assert lo <= hi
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -251,10 +251,7 @@ def read(
                                 raise ValueError(
                                     f"slice start and stop may not be negative; got ({lo}, {hi})"
                                 )
-                            if lo > hi:
```
mlin, line 254: this check was redundant with `util.slice_to_range()`

apis/python/src/tiledbsoma/indexed_dataframe.py
```diff
@@ -418,10 +415,10 @@ def _validate_schema(schema: pa.Schema, index_column_names: Sequence[str]) -> pa
     # verify that all index_column_names are present in the schema
     schema_names_set = set(schema.names)
     for index_column_name in index_column_names:
-        if index_column_name.startswith("soma_") and index_column_name != SOMA_JOINID:
```
mlin, line 421: this check appeared to be redundant given that we're checking that the `index_column_names` are a subset of the `schema.names` where we made a similar check

johnkerl (approved):
🚢

PR #531, opened 2022-11-14
[python] Use larger default py.init_buffer_bytes

johnkerl:
This improves query performance


apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -55,6 +58,17 @@ def __init__(
...
         # Null ctx is OK if that's what they wanted (e.g. not doing any TileDB-Cloud ops).
 
+    def _default_ctx(self) -> tiledb.Ctx:
+        """
+        Default TileDB configuration parameters, for when none other has been specified by the
+        user.
+        """
+        return tiledb.Ctx(
+            {
+                "py.init_buffer_bytes": 1 * 1024**3,
```
gspowley, line 68: Can you please add a line for `soma.init_buffer_bytes` with the same value?

FYI, we may converge these into one parameter in the future.

gspowley (dismissed):

apis/python/src/tiledbsoma/tiledb_object.py
johnkerl: @gspowley I would happily but this is for `main-old`! :)

gspowley: Sorry, didn't notice that the first time. :)

gspowley (approved):

johnkerl:
The C++ SOMA reader has its own config

PR #536, opened 2022-11-15
[python] More checks for parallel ingest

johnkerl:
Follow-on work from #511: more handling for the race condition in `if not exists: create` which occurs in parallel-upload contexts.


johnkerl:
Switching over to instead have a create-schema-only ingest step

PR #537, opened 2022-11-16
[r] Remove DataFrame in favor of IndexedDataFrame

johnkerl:
Experiments have shown (dense) `DataFrame` offers no compelling performance reasons for its continued existence.

Issue: #380


eddelbuettel (commented):
How about keeping `SOMADataFrame` as an alias into `SOMAIndexedDataFrame`?  I am big fan of API continuity.

johnkerl:
@eddelbuettel thanks! On discussion with @bkmartinjr an explicit non-goal for the `main` branch -- which is not yet released -- is to not have any backward-compatibility bits in place for the alpha release. Backward-compatibility issues will be taken into consideration only after alpha.

bkmartinjr:
> an explicit non-goal

True, but I think Dirk's idea is a good one. If we are going to support a single "dataframe" class, calling it `soma.DataFrame` makes more sense to me than `soma.IndexedDataFrame`.

Perhaps alias both to the single implementation, temporarily, and then obsolete `IndexedDataFrame` soon-ish?

johnkerl:
@bkmartinjr @eddelbuettel in that case I'd prefer to rename `IndexedDataFrame` to `DataFrame` on this PR.

bkmartinjr:
The name change is fine with me (for `main`)

johnkerl:
@bkmartinjr @eddelbuettel https://github.com/single-cell-data/SOMA/pull/49

aaronwolen (approved):
Thanks for taking care of this! I added two minor suggestions: one fixes a potential buglet of my own making and the other makes use of a assertion helper I added after the original implementation of `SOMAIndexedDataFrame`.

PR #541, opened 2022-11-16
Introduce schema-only-mode for parallel append-ingest

johnkerl:
Follow-on work from #511


Shelnutt2 (approved):
Looks great!

johnkerl:
Thanks @Shelnutt2 !! :)

PR #542, opened 2022-11-16
Add the `platform_config` parameter to create methods.

thetorpedodog:
This adds a `platform_config` parameter, which allows a user to tune options relating to the creation of their TileDB array.

The structure of the data it expects can be described with this TypeScript specification:

```typescript
interface PlatformConfig {
  tiledb?: TDBConfig;
}

interface TDBConfig {
  create?: TDBCreateOptions;
}

interface TDBCreateOptions {
  dims?: { [dim: string]: TDBDimension };
  attrs?: { [attr: string]: TDBAttr };
  offsets_filters?: TDBFilter[];
  capacity?: number;
  cell_order?: string;
  tile_order?: string;
}

interface TDBDimension {
  filters?: TDBFilter[];
  tile?: number;
}

interface TDBAttr {
  filters?: TDBFilter[];
}

/**
 * Either the name of a filter (in which case it will use
 * the default arguments) or a specification with filter args.
 */
type TDBFilter = string | TDBFilterSpec;

interface TDBFilterSpec {
  /** The name of the filter. */
  _name: string;
  /** kwargs that are passed when constructing the filter. */
  [kwarg: string]: any;
}
```

### DIFF
--- a/apis/python/src/tiledbsoma/sparse_nd_array.py
+++ b/apis/python/src/tiledbsoma/sparse_nd_array.py
@@ -10,11 +10,11 @@ import tiledb
 # This package's pybind11 code
 import tiledbsoma.libtiledbsoma as clib
 
+from . import tiledb_platform_config as tdbpc
 from . import util, util_arrow
 from .collection import CollectionBase
 from .tiledb_array import TileDBArray
-from .tiledb_platform_config import TileDBPlatformConfig
-from .types import NTuple, SparseNdCoordinates
+from .types import NTuple, PlatformConfig, SparseNdCoordinates
 
 
 class SparseNdArray(TileDBArray):
@@ -27,7 +27,7 @@ class SparseNdArray(TileDBArray):
         uri: str,
         *,
         parent: Optional[CollectionBase[Any]] = None,
-        tiledb_platform_config: Optional[TileDBPlatformConfig] = None,
+        tiledb_platform_config: Optional[tdbpc.TileDBPlatformConfig] = None,
         ctx: Optional[tiledb.Ctx] = None,
     ):
         """
@@ -49,6 +49,7 @@ class SparseNdArray(TileDBArray):
         self,
         type: pa.DataType,
         shape: Union[NTuple, List[int]],
+        platform_config: Optional[PlatformConfig] = None,
     ) -> "SparseNdArray":
         """
         Create a ``SparseNdArray`` named with the URI.
@@ -70,15 +71,19 @@ class SparseNdArray(TileDBArray):
             )
 
         level = self._tiledb_platform_config.string_dim_zstd_level
+        create_options = tdbpc.from_param(platform_config).create_options()
 
         dims = []
         for n, e in enumerate(shape):
+            dim_name = f"soma_dim_{n}"
             dim = tiledb.Dim(
-                name=f"soma_dim_{n}",
+                name=dim_name,
                 domain=(0, e - 1),
-                tile=min(e, 2048),  # TODO: PARAMETERIZE,
+                tile=create_options.dim_tile(dim_name, min(e, 2048)),
                 dtype=np.int64,
-                filters=[tiledb.ZstdFilter(level=level)],
+                filters=create_options.dim_filters(
+                    dim_name, [dict(_type="ZstdFilter", level=level)]
+                ),
             )
             dims.append(dim)
         dom = tiledb.Domain(dims, ctx=self._ctx)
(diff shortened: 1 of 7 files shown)


johnkerl (approved):

thetorpedodog:
@bkmartinjr please take a look at this as well, to make sure this fills the needs you were looking for. thanks!

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -99,14 +104,10 @@ def create(
...
-                tiledb.BitWidthReductionFilter(),
-                tiledb.ZstdFilter(),
-            ],
-            capacity=100000,
-            cell_order="row-major",
-            tile_order="row-major",
+            offsets_filters=create_options.offsets_filters(),
+            capacity=create_options.get("capacity", 100000),
+            cell_order=create_options.get("cell_order", "row-major"),
+            tile_order=create_options.get("tile_order", "row-major"),
```
bkmartinjr, line 110: will TileDB-Py ignore the `tile_order` value when `cell_order = "hilbert"`?  Or do we have to choose a default based upon `cell_order`?   (I think it will ignore, but have not confirmed)

apis/python/src/tiledbsoma/tiledb_platform_config.py
```diff
@@ -1,5 +1,9 @@
 from dataclasses import dataclass
-from typing import Optional
+from typing import Any, Dict, Iterator, Mapping, Optional, Sequence, Type, Union, cast
+
+import tiledb
+
+from .types import PlatformConfig
 
 
 @dataclass(frozen=True)
```
bkmartinjr, line 9: I believe many of the values in this class are obsolete and could be deleted.  Several only make sense in `main-old`, or are obsoleted by this new work (with new defaulting).

I believe the obsolete items are:  X_capacity, X_tile_order, obs_extent, var_extent, and string_dim_zstd_level

bkmartinjr (changes requested):
Based on a site read, it looks very good and I only had a couple of small comments inline.

My only "big" ask is that you add a few unit tests, so that we don't regress going forward.

bkmartinjr:
ps. props for "TS as an IDL" :-)

apis/python/src/tiledbsoma/tiledb_platform_config.py
johnkerl: More things _should be brought over from_ `main-old`

johnkerl: That said, we can strip down to what's actually in use in today's foundational classes in `main`, then add to `platform_config` as we add to those classes

bkmartinjr: I agree, but they need to be reworked and moved elsewhere.   They are mostly soma.io preferences/defaults, not DataFrame, SparseNdArray, or DenseNdArray.  Ie, they don't belong in this file.

apis/python/src/tiledbsoma/sparse_nd_array.py
johnkerl: @bkmartinjr we need to say `tile_order=None` explicitly (I know because I've tried :) )

apis/python/src/tiledbsoma/tiledb_platform_config.py
johnkerl: All good! I remember they were added "for reasons" back in the `main-old` days, so I just want to make sure those reasons aren't lost -- thanks for checking and reassuring me! :)

apis/python/src/tiledbsoma/sparse_nd_array.py
bkmartinjr: As long as we can do that via this API, all is well.  I just know we will end up trying hilbert at some point

thetorpedodog: I've updated the behavior here to use the default only if *both* `cell_order` and `tile_order` are unset.

So:

`{"cell_order": "x"}` = `cell_order = "x"`, `tile_order = None`
`{"tile_order": None}` = `cell_order = None`, `tile_order = None`
`{}` = `cell_order = "row-major"`, `tile_order = "row-major"`

apis/python/src/tiledbsoma/tiledb_platform_config.py
thetorpedodog: This seems like something that should get taken care of in a follow-up.

bkmartinjr: That is completely OK with me.  I just personally like to stay on top of detritus :-)

bkmartinjr (approved):

thetorpedodog:
@johnkerl In response to a comment you left in a previous version of the change about how to separate parameters for X vs. parameters for other arrays:

The `platform_config` format described above works well for the single-array case, but I think we need to extend it for multiple arrays. A function which ends up creating multiple arrays can then process it and turn it into the appropriate `platform_config` for each array.

I’m thinking something like this for the format:

```javascript
{
  "create": { /* default create options */ },
  "create:X": { /* create overrides for the X array */ },
  "create:whatever": { /* create overrides for the whatever array */ },
}
```

Then, when passed to a function that creates multiple arrays, the handling looks something like:

```python
class ParamWrapper(...):
  def for_create(self, arr_name: str) -> "ParamWrapper":
    new_conf = dict(self)
    create_options = dict(self.get("create", {}))
    create_options.update(self.get("create:" + name, {}))
    if create_options:
      new_conf["create"]  = create_options
    return ParamWrapper(new_conf)
```

PR #543, opened 2022-11-16
[r] Use X_capacity=1000 to optimize reads

johnkerl:
Larger-scale queries (in the megacell range) have revealed an opportunity to more closely align tile capacity with remote-read performance.

This PR ports #533 from Python to R on `main-old`.


Shelnutt2 (approved):

johnkerl:
Note: the `config["sm.mem.reader.sparse_global_order.ratio_array_data"] = "0.3"` bit was causing a segfault in CI which needs separate debugging. For now I'll land this PR with the `capacity=1000` setting, and address the remaining issue separately.

PR #546, opened 2022-11-17
[python] Update to schema-only mode for subsequent single-soma ingests [main-old]

johnkerl:
Follow-on to #541.

On #541:

* Have inputs say `001.h5ad` .. `100.h5ad`, intending to ingest them all in append mode to `mysoma`
* The caller can ingest `001.h5ad` with `schema_only=True` and wait for completion
  * Since the destination soma is not existing yet, this schema-only step completes successfully
* Then ingest `001.h5ad` .. `100.h5ad` all in parallel

On this PR:

* Have inputs say `001.h5ad` .. `100.h5ad`, intending to ingest them all in append mode to `mysoma` -- but in batches
* First batch:
  * The caller can ingest `001.h5ad` with `schema_only=True` and wait for completion
  * Then ingest `001.h5ad` .. `002.h5ad` (for example) all in parallel (or sequential)
* Second batch:
  * The caller can ingest `003.h5ad` with `schema_only=True` and wait for completion
  * Then ingest `003.h5ad` .. `008.h5ad` all in parallel
  * The oversight on #541 was that the schema-only pass here should accept that the schema already exists.
* Etc. for remaining batches from`009.h5ad` .. `100.h5ad`

Note: The argument could be made that this is on the caller: that the caller can and should do

```
    # in a loop over i starting from 0
    if i == 0 and not soma.exists():
        print(f"-- Writing schema-only for {os.path.basename(file)}")
        tiledbsoma.io.from_anndata(soma, anndata=adata, schema_only=True)
    print(f"-- Writing data for {os.path.basename(file)}")
    tiledbsoma.io.from_anndata(soma, anndata=adata, schema_only=False)
```


aaronwolen (approved):

johnkerl:
Thanks @aaronwolen !!

PR #547, opened 2022-11-17
[r] Add a capacity-related parameter to the default config

johnkerl:


johnkerl:
This parameter was more important in `main-old` than `main`. Will close, and will resubmit if needed.

PR #548, opened 2022-11-17
[r] Expand batched reader support to AnnotationMatrix arrays

aaronwolen:
This PR expands batched reader support to the `AnnotationMatrix` class. It's off by default but can be enabled with the `batch_mode` argument that was added to the `to_matrix()` method.

### DIFF
--- a/apis/r/R/TileDBArray.R
+++ b/apis/r/R/TileDBArray.R
@@ -264,6 +264,43 @@ TileDBArray <- R6::R6Class(
     },
 
     # @description Ingest data into the TileDB array.
-    ingest_data = function() return(NULL)
+    ingest_data = function() return(NULL),
+
+    # @description Retrieve data from the TileDB array
+    # @param batch_mode logical, if `TRUE`, batch query mode is enabled, which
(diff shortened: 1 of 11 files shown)


apis/r/R/TileDBArray.R
```diff
@@ -264,6 +264,43 @@ TileDBArray <- R6::R6Class(
...
+        i <- 1
+        while (isFALSE(tiledb::completedBatched(batcher))) {
+          if (self$verbose) message(sprintf("...retrieving batch %d", i))
+          results[[i]] <- tiledb::fetchBatched(arr, batcher)
+          i <- i + 1
+        }
+
+        # TODO: currently tiledb-r's batched reader ignores return_as and a
+        # data.frame is always returned. When this is addressed we'll need to
+        # add class-specific concatenation logic here.
```
johnkerl, line 298: Do we have a tracking Shortcut story open?

johnkerl (approved):

apis/r/R/TileDBArray.R
aaronwolen: We do indeed: [sc-23436]

PR #551, opened 2022-11-18
Apply ctx to tiledb.array_exists

johnkerl:
Scenario:

* Append-to-single-soma ingestion contexts
* Some configs are present only in `ctx` and not in the shell environment
* After successful schema-only ingest, the data-ingest fails [here](https://github.com/single-cell-data/TileDB-SOMA/blob/93855155949ee97b222877bdb9e1268bbaf474a2/apis/python/src/tiledbsoma/annotation_dataframe.py#L378) because the array-exists method [here](https://github.com/single-cell-data/TileDB-SOMA/blob/93855155949ee97b222877bdb9e1268bbaf474a2/apis/python/src/tiledbsoma/tiledb_array.py#L44) is not using the `self._ctx` as it should

This PR fixes that.


johnkerl:
R CI fail needs investigation but is unrelated to this PR

PR #552, opened 2022-11-18
[r] Ignore directory with TileDB Core headers in coverage analysis

eddelbuettel:
The code coverage analysis also descends into a directory containing headers from TileDB Core which are not part of this repo, and should be excluded from coverage analysis.  Setting an `ignore:` directive achieves this.

No code changes in this PR.


johnkerl (approved):
🚢

PR #553, opened 2022-11-18
[r] Regularize test-file naming [main]

johnkerl:
Unmix `test-*` and `test_*`


eddelbuettel (approved):
Like the review for main-old, it's fine as it does not damage but I see no upside in worrying about _ vs - in filenames.  The test runner takes both.

PR #554, opened 2022-11-18
[r] Regularize test-file naming [main-old]

johnkerl:
Unmix `test-*` and `test_*`


eddelbuettel (approved):
This does not do any harm, I am unsure what benefit it has.

PR #561, opened 2022-11-22
[c++] Map 8-bit TILEDB_BOOL to 1-bit Arrow boolean

bekadavis9:
Problem: `TILEDB_BOOL` is stored as a `uint8_t`, while Arrow uses 1-bit `bool`. 
Solution: Overwrite the `ColumnBuffer`'s bytemap in-place with its corresponding byte-aligned bitmap.

Test case:
```python
#!/usr/bin/env python

import os
import shutil

import numpy as np
import pandas as pd
import pyarrow as pa
import tiledbsoma as soma
import tiledbsoma.libtiledbsoma as clib

uri = "489"
if os.path.exists(uri):
    shutil.rmtree(uri)

df = pd.DataFrame(
    {
        "soma_joinid": np.asarray([1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20], dtype=np.int64),
        "b": np.array([False,True,True,False,True,False,True, False,False,False,True,False,True,False,False,False,True,False,True,False], dtype=bool),
    }
)
tbl = pa.Table.from_pandas(df)
sdf = soma.DataFrame(uri)
sdf.create(schema=tbl.schema, index_column_names=['soma_joinid'])
sdf.write(tbl)

print("DataFrame.schema:")
print(sdf.schema)

print()
print("PANDAS")
read_df = sdf.read_as_pandas_all()
for k in read_df:
  print(k, read_df[k].dtype)
assert sdf.read_as_pandas_all().b.dtype == bool

print()
print("ARROW")
read_df = sdf.read_all()
for v in read_df:
  print(v.to_pylist())
  print(v.type)

print()
print("SOMAReader")
sr = clib.SOMAReader('489')
sr.submit()
t = sr.read_next()
print(t.to_pylist())
print(t['b'].type)
```

Results:
```bash
DataFrame.schema:
soma_joinid: int64
b: bool

PANDAS
soma_joinid int64
b bool

ARROW
[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
int64
[False, True, True, False, True, False, True, False, False, False, True, False, True, False, False, False, True, False, True, False]
bool

SOMAReader
[{'soma_joinid': 1, 'b': False}, {'soma_joinid': 2, 'b': True}, {'soma_joinid': 3, 'b': True}, {'soma_joinid': 4, 'b': False}, {'soma_joinid': 5, 'b': True}, {'soma_joinid': 6, 'b': False}, {'soma_joinid': 7, 'b': True}, {'soma_joinid': 8, 'b': False}, {'soma_joinid': 9, 'b': False}, {'soma_joinid': 10, 'b': False}, {'soma_joinid': 11, 'b': True}, {'soma_joinid': 12, 'b': False}, {'soma_joinid': 13, 'b': True}, {'soma_joinid': 14, 'b': False}, {'soma_joinid': 15, 'b': False}, {'soma_joinid': 16, 'b': False}, {'soma_joinid': 17, 'b': True}, {'soma_joinid': 18, 'b': False}, {'soma_joinid': 19, 'b': True}, {'soma_joinid': 20, 'b': False}]
bool
```

### DIFF
--- a/libtiledbsoma/include/tiledbsoma/arrow_adapter.h
+++ b/libtiledbsoma/include/tiledbsoma/arrow_adapter.h
@@ -3,6 +3,7 @@
 
 #include <tiledbsoma/tiledbsoma>
 #include "carrow.h"
+#include <stdio.h>
 
 // https://arrow.apache.org/docs/format/CDataInterface.html
 // https://arrow.apache.org/docs/format/Columnar.html#buffer-listing-for-each-layout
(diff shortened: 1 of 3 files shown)


johnkerl (approved):
🚢 🚢 🚢 

Looks great!!! :)

It wouldn't hurt to add another test -- maybe in `libtiledbsoma/test` -- which tries:

* A longer array (more than 8 bits)
* The shortest possible Arrow array (length 0 or length 1)

libtiledbsoma/include/tiledbsoma/arrow_adapter.h
```diff
@@ -3,6 +3,7 @@
 
 #include <tiledbsoma/tiledbsoma>
 #include "carrow.h"
+#include <stdio.h>
```
gspowley, line 6: nit: please remove this include if not needed.

gspowley (approved):
LGTM, thanks @bekadavis9!

gspowley:
Right, the reasoning behind this PR was to match the Arrow ABI for bool.

bkmartinjr:
> outside the scope of this PR IMO

I am not asking for this PR to be refined.  I'm just asking about the rationale for the "ask" as George put it :-).  I can see the benefit for Arrow ABI users, and the cost for Pandas/NumPy users - wondering how the decision was made wrt to this tradeoff, and if there is backlog filed somewhere to deal with the Pandas/NumPy use case.

(my motivation:  our cell census includes some boolean columns, and most of our Python users will be in the Pandas/AnnDAta API space)

johnkerl:
> If we want direct-to-pandas out of the C++ API that will involve some redesign and rework, outside the scope of this PR IMO.

> if there is backlog filed somewhere to deal with the Pandas/NumPy use case

Sounds like we're all agreed we should make a backlog issue :)

Issue #540, opened 2022-11-16
API: concurrency support / point-in-time consistency


bkmartinjr:
The current SOMA API lacks any support to access arrays/dataframes at a consistent state, and assumes that all objects are immutable for the purposes of reading.  For example, the current Python `SparseNdArray` "read" code contains (psuedo-code):
```
...
open array to get shape
close array
...
open array to get data
close array
...
```

A similar issue can arise when there are simultaneous writes and the API user performs more than one operation on an object over time, or has expectations of multi-object consistency (eg, for a SOMAExperiment).

Despite the underlying storage engine containing sophisticated features to support point-in-time consistency (and more) on _single_ objects, there is no way to access these features via the SOMA API.  I believe there are several options, in order of increasing complexity & utility:
1. leave as is - no API support. It is the responsibility of SOMA users to manage all concurrency and consistency.
2. for a single object (e.g., ndarray, dataframe), add the ability to read/write with a "recent" consistent state
3. \# 2, but with time-travel and roll-backs
4. Multi-object consistency (either "recent", aka \# 2 or with time-travel/roll-backs)

I propose we shoot for # 2 in the initial SOMA API release.  We can consider the others for future API versions, as needs dictate. The rest of this issue is a sketch of how the API might change.

CC: @thetorpedodog @johnkerl @aaronwolen @eddelbuettel @mlin @gspowley 

----

The current SOMA objects are stateless URI containers, which open the underlying storage objects upon each API call (sometimes more than once per API call).  Proposed API changes:
* each SOMA object (eg SOMASparseNdArray) becomes stateful, with two chunks of internal state:
    * an "open state", which is logically one of `not open`, `open for read`, `open for write`
    * a private handle to the underlying storage engine state require
* each SOMA object has two new methods - `open` and `close`, which put the object into one of the above states
* the read & write methods (and related methods such as `nnz`, `shape`, etc) all require that the object be in the correct mode for the operation.  Eg.,   somaNdArray.read() will raise an error if the object is not opened for reading.
* each SOMA implementation provides concurrency/consistency guarantees as able

The above should be sufficient to allow a TIileDB SOMA to guarantee that:
* objects that are opened for reading can perform multiple read-like operations at a given point of consistency
* objects that are opened for writing can treat open/close as barriers for noting eventual consistency
In other words, the same consistency contract made by the existing native TileDB-Py and TileDB-R API.

For example, with this new API, reading an dataframe would look like:
```
    uri = ...
[...]

PR #562, opened 2022-11-27
[python] platform_config for DataFrame.read()

mlin:
Use to test paginated reading.


apis/python/tests/test_dataframe.py
```diff
@@ -783,3 +783,33 @@ def test_result_order(tmp_path):
...
+    }
+    sidf.write(pa.Table.from_pydict(data))
+
+    # on defaults, table should read out in one iteration
+    assert sum(x for x in next(sidf.read_as_pandas())["i"].to_list()) == N * (N - 1) / 2
+
+    # setting in platform_config cause it to come out in multiple pages
+    pages = 0
+    total = 0
+    for table in sidf.read_as_pandas(platform_config={"soma.init_buffer_bytes": "64"}):
```
mlin, line 811: @thetorpedodog @johnkerl @gspowley @bkmartinjr I can use some advice here: wiring DataFrame.read() to take platform_config. Our libtiledbsoma SOMAReader receives platform_config as a `Mapping[str,str]`/`map<string,string>`. I've wired it up here as such, and it works for the special libtiledbsoma-specific setting I needed to override...

But I saw in #542 there's some wrapping around `platform_config` where the mapping values have nested structure, and there are some predefined attributes/options that get processed. I couldn't quite tell how/if I should extend that to this use case, to make it consistent and a little higher-level than `Mapping[str,str]`. Any advice on that please? (It's also awkward that I had to write `"64"` as a str here...)

Also noting we may end up revising this in #540 if we end up making the reader more stateful to avoid re-opening the context for each read.

bkmartinjr: The [spec](https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md#platform-specific-configuration) states that:
> The platform_config parameter is defined as a key–value mapping from strings to configuration data. Each key in the mapping corresponds to the name of a particular SOMA implementation (i.e., the same string returned by the get_storage_engine call).

So I don't know if the C++ code is expecting the _entire_ map, or just the `TileDB`-specific section.  That seems like an internal decision which @gspowley can answer.  Off the top of my head, I can't think of a reason to prefer one other the other - passing down `platform_condig["tliedb"]` seems as good as passing down the entire dict.

gspowley: The C++ code is expecting the `TileDB`-specific section of the `platform_config`. 

It seems like in the python code where a `SOMAReader` object is created, is a good place to pass `platform_config["tiledb"]` to the C++ code.

thetorpedodog: The `tiledb` key of the `platform_config` dictionary is not a raw dict for passing to `tiledb.Config`. The format as defined so far is in the commit message that added it: https://github.com/single-cell-data/TileDB-SOMA/commit/4b4b2645d876bb28789f8ca39a1026361844ada4

In this case I think the thing to do is to add a key like `tiledb_config: {[name: string]: string}` that we will use to construct the config/context itself.

johnkerl (approved):
🚢

mlin:
re https://github.com/single-cell-data/TileDB-SOMA/pull/562#discussion_r1035216423

@thetorpedodog thanks, I thought we'd have to do more to mesh this with the create config...I didn't completely follow the suggested approach though. Can I ask you to elaborate...where are we putting this new `tiledb_config` key, just in the dict passed here, or in the higher-level `TileDBPlatformConfig`?

cc @gspowley

mlin:
Withdrawing as this particular implementation should be superseded by @atolopko-czi work on #644

PR #568, opened 2022-11-29
Add documentation for `platform_config` to the Python README.

thetorpedodog:
Based on a note I left on #562, I figured the `platform_config` format should probably be documented someplace more visible and durable than the contents of a single commit message.


bkmartinjr (approved):
this is fantastic - thank you!

@mlin - when you connect the read() path `platform_config`, could you extend this section?

PR #572, opened 2022-12-02
[r] Make source build more robust

eddelbuettel:
This PR aims to robustify building from source out of the repo reflecting some of the experience Pablo had.  We add an explicit test for `cmake` which, while listed as required, is less commonly present in R environments, and reactivate a section using the build script if we are in fact in the repo.

No new code.


aaronwolen (approved):
Good call, I ran into this too.

Issue #560, opened 2022-11-21
Connect `SOMAReader` to `DenseNdArray`


johnkerl:
Needs a split-out for `DenseNdArray`

_Originally posted by @johnkerl in https://github.com/single-cell-data/TileDB-SOMA/issues/431#issuecomment-1322703660_

Issue #544, opened 2022-11-17
[python] Support Python 3.11 (requires numba 0.57.0rc1)


bkmartinjr:
Our setup.cfg says Python versions 3.7+ are supported.  The CI currently only tests 3.8 and 3.9.  We need to add to the matrix 3.7, 3.10 and 3.11

---

**Update as of 13-Mar-2023:** numba support (see numba/numba#8304) is still WIP.  Guesstimate: H123

johnkerl:
Looks like we'll need to wait on Python 3.11. From #599:

At https://github.com/single-cell-data/TileDB-SOMA/actions/runs/3661926647/jobs/6190605155 we see `numba` is not yet supported in Python 3.11. cc @thetorpedodog @bkmartinjr @Shelnutt2 .

```
Collecting numba>=0.41.0
  Downloading numba-0.56.4.tar.gz (2.4 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 2.4/2.4 MB 146.6 MB/s eta 0:00:00
  Preparing metadata (setup.py): started
  Running command python setup.py egg_info
  Traceback (most recent call last):
    File "<string>", line 2, in <module>
    File "<pip-setuptools-caller>", line 34, in <module>
    File "/tmp/pip-install-0tpo_aju/numba_b1ec8a6b557b450ea19a316e9668617a/setup.py", line 51, in <module>
      _guard_py_ver()
    File "/tmp/pip-install-0tpo_aju/numba_b1ec8a6b557b450ea19a316e9668617a/setup.py", line 48, in _guard_py_ver
      raise RuntimeError(msg.format(cur_py, min_py, max_py))
  RuntimeError: Cannot install on Python version 3.11.0; only versions >=3.7,<3.11 are supported.
```

See also https://github.com/numba/numba/issues/8304

Looks like they're hoping for a release candidate by end of 2022: https://github.com/numba/numba/issues/8304#issuecomment-1336218610

johnkerl:
#620 (merged) implements Python 3.7 support

Python 3.11 is still blocked on the aforementioned numba-dependency issue

PR #574, opened 2022-12-02
Use C++ reader for SOMA.DenseNdArray

johnkerl:
Context: #560


apis/python/src/tiledbsoma/dense_nd_array.py
```diff
@@ -148,16 +151,64 @@ def read_tensor(
...
+            if not isinstance(coords, (list, tuple)):
+                raise TypeError(
+                    f"coords type {type(coords)} unsupported; expected list or tuple"
+                )
+            if len(coords) < 1 or len(coords) > schema.domain.ndim:
+                raise ValueError(
+                    f"coords {coords} must have length between 1 and ndim ({schema.domain.ndim}); got {len(coords)}"
+                )
+
+            for i, coord in enumerate(coords):
```
johnkerl, line 175: There's some (but not entire) code-dup between this and `SparseNdArray`.

I could extract a helper function, with the differences iffed.

apis/python/src/tiledbsoma/dense_nd_array.py
```diff
@@ -148,16 +151,64 @@ def read_tensor(
...
+                    coord, (collections.abc.Sequence, pa.Array, pa.ChunkedArray)
+                ):
+                    sr.set_dim_points(dim_name, coord)
+                else:
+                    raise TypeError(f"coord type {type(coord)} at slot {i} unsupported")
+
+        sr.submit()
+
+        arrow_table = sr.read_next()
+        # XXX assert no more data remaining
```
johnkerl, line 208: @bkmartinjr : in the spec we don't have a generator for the read in `DenseNdArray`.

@gspowley : what's your advice on how to handle ensuring we really do have all the data?

gspowley: You can call `sr.results_complete()`. It returns `true` if the last call to `sr.read_next()` returned all results for the query.

eddelbuettel: And FYI a working (if minimal) implementation of `setup; while !complete: read next;` is in the R package [in this file](https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/r/src/riterator.cpp).

apis/python/src/tiledbsoma/dense_nd_array.py
```diff
@@ -148,16 +151,64 @@ def read_tensor(
...
+                raise TypeError(
+                    f"coords type {type(coords)} unsupported; expected list or tuple"
+                )
+            if len(coords) < 1 or len(coords) > schema.domain.ndim:
+                raise ValueError(
+                    f"coords {coords} must have length between 1 and ndim ({schema.domain.ndim}); got {len(coords)}"
+                )
+
+            for i, coord in enumerate(coords):
+                #                # Example: coords = [None, 3, slice(4,5)]
```
bkmartinjr, line 176: Tiny nit (OK to ignore): these types of comments make the code hard to read as they lack context.  They are useful though, as end-user docstrings, so I suggest a nice, clean and thorough docstring at the top of the method, which includes examples such as these.

apis/python/src/tiledbsoma/dense_nd_array.py
```diff
@@ -148,16 +151,64 @@ def read_tensor(
...
+                            )
+                        if lo > hi:
+                            raise ValueError(
+                                f"slice start must be <= slice stop; got ({lo}, {hi})"
+                            )
+                        sr.set_dim_ranges(dim_name, [lo_hi])
+                    # Else, no constraint in this slot. This is `slice(None)` which is like
+                    # Python indexing syntax `[:]`.
+                elif isinstance(
+                    coord, (collections.abc.Sequence, pa.Array, pa.ChunkedArray)
```
bkmartinjr, line 199: Apologies in advance if I'm confused - but we changed the DenseNdArray to _only_ allow sub-array read/write, and not allow point indexing. In other words, passing in a pa.Array for coordinates is bogus for a dense array.  Point indexing is only supported for SparseNdArray and DataFrame.

This code (lies 198-201) will never be executed, as an exception will be raised in _dense_index_to_shape() if the user submits coordinates as an array.

Assuming my recollection (sub-array only) is correct, then there appear to be several bugs here (one is left over from my own PR - blush):
1. the `DenseNdCoordinates` type should only allow int and slice, and should not include `pa.Array`.
2. for consistency, the type signature of `_dense_index_to_shape()` should type the `coords` param with `DenseNdCoordinates`
3. the code supporting pa.Array (lines 198-201) should be removed.
4. And, a unit test should confirm all this (lacking one, we failed to notice these issues)

apis/python/src/tiledbsoma/dense_nd_array.py
bkmartinjr: > in the spec we don't have a generator for the read in DenseNdArray

Correct - we decided to remove it as:
* we don't support point indexing
* the user can determine a priori the size of the result
* and therefore do their own partitioning/iteration 

We also agreed that we would revisit this if it ever became onerous

bkmartinjr: And I think you are implicitly acknowledging it in your comment, but this code does not appear to work correctly for reads that exceed the soma buffer size.  it needs do the iteration, concat the tables, and then create the tensor.

@gspowley - I believe using tables in this way results in an extra copy as you can't create a tensor from a table with zero copy.  At some point in the future, we would benefit from a "read-as-tensor" C++ interface.

bkmartinjr (dismissed):
Two requests:
* the point indexing issue needs resolution (ie, remove the unreachable code, fix the type signatures)
* the SOMAReader result needs to be read until completion, concatenated, then tuned into a tensor.

Edit -- third request:
* unit tests to verify that both of the above work correctly

apis/python/src/tiledbsoma/dense_nd_array.py
```diff
@@ -148,18 +151,80 @@ def read_tensor(
...
+                    # Else, no constraint in this slot. This is `slice(None)` which is like
+                    # Python indexing syntax `[:]`.
+                else:
+                    raise TypeError(f"coord type {type(coord)} at slot {i} unsupported")
+
+        sr.submit()
+
+        arrow_tables = []
+        while arrow_table_piece := sr.read_next():
+            arrow_tables.append(arrow_table_piece)  # XXX address the copy question
```
johnkerl, line 210: @gspowley unit tests on this PR validate that we are not getting data-clobbers on what I _thought_ would be reuse of the same Arrow buffers ... thoughts?

johnkerl: Never mind, I misunderstood -- I had thought that the narrative was always there were clobberable buffers we would need to be careful of -- now I understand that every call to `read_next` will get new buffers

johnkerl:
@bkmartinjr @eddelbuettel @gspowley ready for review round two

apis/python/src/tiledbsoma/dense_nd_array.py
```diff
@@ -148,18 +151,80 @@ def read_tensor(
...
+                        sr.set_dim_ranges(dim_name, [lo_hi])
+                    # Else, no constraint in this slot. This is `slice(None)` which is like
+                    # Python indexing syntax `[:]`.
+                else:
+                    raise TypeError(f"coord type {type(coord)} at slot {i} unsupported")
+
+        sr.submit()
+
+        arrow_tables = []
+        while arrow_table_piece := sr.read_next():
```
bkmartinjr, line 209: I believe the walrus operator is Python 3.8+.  I was under the impression SOMA was supporting 3.7+?

Site note: need a GH Action to validate the full test matrix at some point.

johnkerl: > Site note: need a GH Action to validate the full test matrix at some point.

@bkmartinjr can you please elaborate?

bkmartinjr: We should choose & and enforce it.  I was unaware of Literal :blush:

If it is 3.8+, I"m fine with walrus.

apis/python/tests/test_dense_nd_array.py
```diff
@@ -146,7 +168,12 @@ def test_dense_nd_array_slicing(tmp_path, io):
     SOMA's doubly-inclusive slice indexing semantics against Python's singly-inclusive slicing
     semantics, ensuring that none of the latter has crept into the former.
     """
-    a = soma.DenseNdArray(tmp_path.as_posix())
+    cfg = {}
```
bkmartinjr, line 171: with all of the new error checks being performed in the read path, can we add some trivial tests that hit all of the exceptions/error checks?  Ie.,
* bogus coord type (eg., a string)
* bogus slice (eg,   slice(-1))
* points < 0, > size
etc

just want to make sure we cover the expected edge cases

bkmartinjr (approved):
Changes look good.  I have suggestions, but up to you if you want to do them in this PR or later:
* clarify Python supported version matrix and its impact (as this PR adds use of walrus operator)
* better unit tests for the code added in this PR

apis/python/tests/test_dense_nd_array.py
johnkerl: @bkmartinjr bogus slice, points < 0, > size, sure!

Bogus coord type, no we cannot because of your request on
https://github.com/single-cell-data/TileDB-SOMA/pull/581#discussion_r1041366130
namely typeguard _itself_ would fail our attempt to call with a type not matching the signature

apis/python/src/tiledbsoma/dense_nd_array.py
johnkerl: @bkmartinjr I prefer to let #544 do its assigned duties

apis/python/tests/test_dense_nd_array.py
bkmartinjr: Our users do not run with typeguard, so I believe you still need to test it!

In other words, it is a weak unit test that only tests the "supproted" types!

johnkerl:
@bkmartinjr we're at 4 days and 33 commits on this PR ... re unit-test expansion I'll do that on a follow-up PR, and I won't close #560 until I do.

apis/python/tests/test_dense_nd_array.py
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/issues/584

Issue #576, opened 2022-12-04
[python] DataFrame value_filter failing on bool column


bkmartinjr:
A value filter (query condition) on a bool-typed column fails.  It works fine on a string column.

Example:
```python
In [34]: census = soma.Collection("s3://cellxgene-data-public/cell-census/2022-12-01/soma", ctx=tiledb.Ctx(config={"vfs.s3.region": "us-west-2"}))

In [35]: census['census_data']['mus_musculus'].obs.read_as_pandas_all(column_names=['soma_joinid', 'is_primary_data', 'cell_type'])
Out[35]: 
         soma_joinid  is_primary_data                cell_type
0                  0             True          oligodendrocyte
1                  1             True          oligodendrocyte
2                  2             True          oligodendrocyte
3                  3             True          oligodendrocyte
4                  4             True          oligodendrocyte
...              ...              ...                      ...
3409714      3409714             True  cerebellar granule cell
3409715      3409715             True  cerebellar granule cell
3409716      3409716             True  cerebellar granule cell
3409717      3409717             True  cerebellar granule cell
3409718      3409718             True  cerebellar granule cell

[3409719 rows x 3 columns]

In [36]: census['census_data']['mus_musculus'].obs.read_as_pandas_all(column_names=['soma_joinid', 'is_primary_data', 'cell_type'], value_filter="is_primary_data == True")
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[36], line 1
----> 1 census['census_data']['mus_musculus'].obs.read_as_pandas_all(column_names=['soma_joinid', 'is_primary_data', 'cell_type'], value_filter="is_primary_data == True")

File ~/TileDB-SOMA/apis/python/src/tiledbsoma/dataframe.py:388, in DataFrame.read_as_pandas_all(self, ids, value_filter, column_names, result_order)
    381 def read_as_pandas_all(
    382     self,
    383     ids: Optional[SparseDataFrameCoordinates] = None,
   (...)
    386     result_order: Optional[ResultOrder] = None,
    387 ) -> pd.DataFrame:
--> 388     return pd.concat(
    389         self.read_as_pandas(
    390             ids=ids,
    391             value_filter=value_filter,
    392             column_names=column_names,
    393             result_order=result_order,
    394         ),
    395         ignore_index=True,
    396     )

[...]

PR #579, opened 2022-12-06
Support value-filter on boolean columns [pending core 2.13 / tiledb-py 0.19]

johnkerl:
Context: #576

NOTE: passes on my sandbox with core 2.13.0; fails on my sandbox with core 2.12.2 as follows:

```
tiledb.cc.TileDBError: Error: Internal TileDB uncaught exception; Cannot perform query comparison; Unsupported query conditional type on is_primary_data
```

This is the same error we're seeing in CI at present (with core 2.12). This means we should hold off this PR until 2.13 is out and propagated through a `TileDB-Py` 0.19.0 etc. (which should be this week).

For the record, how to add a column to `pbmc3k_processed.h5ad` (I added a `bool` column for this PR):

```
import anndata as ad
print(adata.obs)
adata = ad.read_h5ad('data/pbmc3k_processed.h5ad')
adata.obs['is_b_cell'] = adata.obs['louvain'] == "B cells"
adata.write_h5ad('data/pbmc3k_processed.h5ad')
```

Also note that this works: https://gist.github.com/johnkerl/23da515f60cdcdee3b6204d815eb1279


libtiledbsoma/test/test_query_condition.py
```diff
@@ -70,6 +70,16 @@ def test_query_condition_float():
...
 
+def test_query_condition_bool():
+    uri = os.path.join(SOMA_URI, "obs")
+    condition = "is_b_cell == True"
+
+    pandas = pandas_query(uri, condition)
+
+    soma_arrow = soma_query(uri, condition)
+
+    assert len(pandas.index) == 342
```
gspowley, line 81: We should compare the pandas query to the soma query:
```python
assert len(pandas.index) == soma_arrow.num_rows
```

gspowley (changes requested):

johnkerl:
@gspowley ready for review round two! :)

gspowley (approved):
LGTM!

PR #581, opened 2022-12-06
Fix DenseNdArray slicing

johnkerl:
Porting the `DenseNdArray` reader from `tiledb-py` to `tiledbsoma.libtiledsoma` for PR #574 / issue #560 surfaced a prerequisite error. Namely, `slice(None)` was the only slicing ever used to test `DenseNdArray`, and the code for converting Python to SOMA slices was buggy in a way unrevealed by the absence of non-trivial slice data in unit tests. In particular, at present we have SOMA slice semantics being doubly inclusive on `SparseNdArray` which (until something changes) be carried through to `DenseNdArray`. (See also https://github.com/single-cell-data/SOMA/issues/51 for discussions on this point.)

* On this PR we bugfix `DenseNdArray` slicing with respect to doubly-inclusive slice semantics.
  * If it should later prove to be the case on https://github.com/single-cell-data/SOMA/issues/51 that we choose singly-inclusive semantics, the presence of tests introduced here will make implementing that change a test-driven process (which is not the case currently).
* After this the C++ conversion for issue #560 / PR #574 will be simplified, and as well that port will now be test-driven.


apis/python/src/tiledbsoma/util.py
```diff
@@ -127,3 +129,52 @@ def slice_to_range(ids: slice) -> Optional[Tuple[int, int]]:
...
+    """
+    Given a subarray index specified as a tuple of per-dimension slices or scalars
+    (eg, ``([:], 1, [1:2])``), and the shape of the array, return the shape of
+    the subarray.
+
+    See read_tensor for usage.
+    """
+    shape: List[int] = []
+    for n, coord in enumerate(coords):
+        shape.append(dense_index_to_shape(coord, array_shape[n]))
```
thetorpedodog, line 148: A comprehension with `zip` might be more natural here:

```suggestion
    shape = [
        dense_index_to_shape(coord, dim_shape)
        for (coord, dim_shape) in zip(coords, array_shape)
    ]
```

apis/python/src/tiledbsoma/util.py
```diff
@@ -127,3 +129,52 @@ def slice_to_range(ids: slice) -> Optional[Tuple[int, int]]:
...
+    if type(coord) is int:
+        return 1
+
+    if type(coord) is slice:
+        start, stop, step = coord.indices(array_length)
+        if step != 1:
+            raise ValueError("stepped slice ranges are not supported")
+        # This is correct for doubly-inclusive slices which SOMA uses.
+        if stop >= array_length:
+            stop = array_length - 1
```
thetorpedodog, line 177:
```suggestion
        stop = min(stop, array_length - 1)
```

thetorpedodog (commented):
Code overall looks pretty good. I have a few minor suggestions and am wondering if you intended to fill in the skeleton of the test case.

apis/python/src/tiledbsoma/util.py
```diff
@@ -127,3 +129,58 @@ def slice_to_range(ids: slice) -> Optional[Tuple[int, int]]:
     if ids.step is not None and ids.step != 1:
         raise ValueError("slice step must be 1 or None")
     return (ids.start, ids.stop)  # XXX TEMP
+
+
+def dense_indices_to_shape(
+    coords: Tuple[Union[int, slice], ...],
```
bkmartinjr, line 135: coords should be typed with the DenseNdCoordinates type used by callers - that will help keep all of the various function signatures in sync (eg, would point out that dense coords can't take a pa.Array :)

johnkerl: @bkmartinjr good call! Just know that since we use `typeguard` in our unit tests for #574, we'll no longer be able to write a unit test that passes in `pa.Array` and expect it to fail -- since `typeguard` will fail our _attempt_ to do that. :(

apis/python/src/tiledbsoma/dense_nd_array.py
```diff
@@ -152,7 +153,7 @@ def read_tensor(
             result_order, accept=["column-major", "row-major"]
         )
         with self._tiledb_open("r") as A:
-            target_shape = _dense_index_to_shape(coords, A.shape, result_order)
+            target_shape = dense_indices_to_shape(coords, A.shape, result_order)
```
bkmartinjr, line 156: there is a bug here - `coords` param can contain a pa.Array, which is not handled by this helper.

The actual bug is the definition of `DenseNdCoordinates`, which should not accept a pa.Array (a bug I left behind when I did the sub-array changes - sorry!)   the correct fix IMHO is to change the definition of `DenseNdCoordinates` to be only a list of int | slice

bkmartinjr (approved):
One issue relating to point indexing - which should be illegal in a DenseNdArray, but is allowed by the DenseNdCoordinates type signature.  Other, this is a great PR - thank you for doing it!  TDD rules

apis/python/src/tiledbsoma/dense_nd_array.py
johnkerl: @bkmartinjr I fixed this on #574: I'll backport

bkmartinjr: gotcha - I'm OK either way as long as it gets fixed at some point ;)  Do whichever is easiest for you!

PR #582, opened 2022-12-06
Support half-slices in SOMAReader

bekadavis9:
Use `nonempty_domain` to convert half-slices of `(None, int)` or `(int, None)` to valid ranges interpreted by the SOMAReader.

### DIFF
--- a/apis/python/src/tiledbsoma/dataframe.py
+++ b/apis/python/src/tiledbsoma/dataframe.py
@@ -259,8 +259,7 @@ class DataFrame(TileDBArray):
                         sr.set_dim_points(dim_name, dim_ids)
                     elif isinstance(dim_ids, slice):
                         ned = A.nonempty_domain()
-                        lo_hi = util.slice_to_range(
-                            dim_ids, ned[i]) if ned else None
+                        lo_hi = util.slice_to_range(dim_ids, ned[i]) if ned else None
                         if lo_hi is not None:
                             lo, hi = lo_hi
                             if lo < 0 or hi < 0:
--- a/apis/python/src/tiledbsoma/dataframe.py
+++ b/apis/python/src/tiledbsoma/dataframe.py
@@ -258,7 +258,9 @@ class DataFrame(TileDBArray):
                             )
                         sr.set_dim_points(dim_name, dim_ids)
                     elif isinstance(dim_ids, slice):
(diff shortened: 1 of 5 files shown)


apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -258,7 +258,8 @@ def read(
                             )
                         sr.set_dim_points(dim_name, dim_ids)
                     elif isinstance(dim_ids, slice):
-                        lo_hi = util.slice_to_range(dim_ids)
+                        ned = A.nonempty_domain()
+                        lo_hi = util.slice_to_range(dim_ids, ned[i]) if ned else None
```
johnkerl, line 262: Not sure if you have another commit coming -- please pardon if these remarks are premature :)

Let's add a comment saying that non-empty domain is None precisely when there is no data in the array at all

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -290,7 +290,8 @@ def read_table(self, coords: SparseNdCoordinates) -> Iterator[pa.Table]:
                             )
                         sr.set_dim_points(dim_name, coord)
                     elif isinstance(coord, slice):
-                        lo_hi = util.slice_to_range(coord)
+                        ned = A.nonempty_domain()
+                        lo_hi = util.slice_to_range(coord, ned[i]) if ned else None
```
johnkerl, line 294: ditto

apis/python/src/tiledbsoma/util.py
```diff
@@ -113,17 +113,24 @@ def uri_joinpath(base: str, path: str) -> str:
...
     if ids.start is None and ids.stop is None:
         return None
-    if ids.start is None or ids.stop is None:
-        # TODO: https://github.com/single-cell-data/TileDB-SOMA/issues/457
-        raise ValueError("slice start and stop must be both specified, or neither")
-    if ids.start > ids.stop:
+    start = ids.start
+    stop = ids.stop
+
+    if start is None:
```
johnkerl, line 129: Just a thought ... maybe add a to-do comment that with more work on the C++ reader we wouldn't need to do this in Python ...

johnkerl (approved):
🚀

PR #586, opened 2022-12-07
Update to TileDB 2.13 [main-old branch]

johnkerl:
Core 2.13 aligns with TileDB-Py 0.19 and TileDB-R 0.17. Those are already released.

After this we can tag TileDB-SOMA 0.1.19.

Once TileDB Cloud is updated to 2.13 (REST server) then we can include TileDB-SOMA 0.1.19 in the UDF/notebook-image repos.

### DIFF
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -36,6 +36,7 @@ Imports:
     R6,
     methods,
     Matrix,
+    tiledb,
     SeuratObject,
     jsonlite,
     glue,
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -36,7 +36,6 @@ Imports:
     R6,
     methods,
     Matrix,
-    tiledb (>= 0.17.0),
     SeuratObject,
     jsonlite,
     glue,
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -36,7 +36,7 @@ Imports:
     R6,
     methods,
     Matrix,
-    tiledb (>= 0.14.0),
+    tiledb (>= 0.17.0),
     SeuratObject,
     jsonlite,
     glue,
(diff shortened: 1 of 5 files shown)


tools/ci/install-r-dependencies.sh
```diff
@@ -2,7 +2,21 @@
...
+#
+# See also https://github.com/single-cell-data/TileDB-SOMA/issues/654
+#
+# * Enable repository from tiledb-inc
+# * Download and install tiledb in R
+Rscript -e '
+    options(repos = c(tiledbinc = "https://tiledb-inc.r-universe.dev", CRAN = "https://cloud.r-project.org"));
+    install.packages("tiledb")
+'
+
```
eddelbuettel, line 17: Nit level comment:  I would write/indent it more tightly (two lines max) and possibly also place it below the `remotes` install.  But most importantly it works.

apis/r/DESCRIPTION
```diff
@@ -36,7 +36,7 @@ Imports:
     R6,
     methods,
     Matrix,
-    tiledb (>= 0.14.0),
+    tiledb (>= 0.17.0),
```
eddelbuettel, line 39: I think we can remove the versioned dependency alltogether.

eddelbuettel (approved):
Looks fine to me apart from a stylistic nit or two.  So you may want to wait for a second approval before merging.

johnkerl:
> Looks fine to me apart from a stylistic nit or two. So you may want to wait for a second approval before merging.

At this point this PR's sole job is to unbreak CI -- thanks for your review! :)

apis/r/DESCRIPTION
```diff
@@ -36,7 +36,6 @@ Imports:
     R6,
     methods,
     Matrix,
-    tiledb (>= 0.17.0),
```
eddelbuettel: That is not right. We want tiledb, just without a version. So please make it

     tiledb,

eddelbuettel (changes requested):

apis/r/DESCRIPTION
eddelbuettel: Unless `main-old` does not use `tiledb` at all?  I am rather surprised this survived the CI for R.

johnkerl: @eddelbuettel already fixed -- thanks!

eddelbuettel (approved):
LGTM

PR #587, opened 2022-12-07
Update to TileDB 2.13 [main]

johnkerl:
Core 2.13 aligns with TileDB-Py 0.19 and TileDB-R 0.17. These are already released.


eddelbuettel (approved):
Thanks for updating the R getter too

gspowley (approved):

PR #591, opened 2022-12-08
[r] Convert arrow::large_utf8() to ascii

### LINKED ISSUE #578: cell & tile order swapped in SparseNdArray
There is a bug in SparseNdArray - it swaps the tile_order and cell_order assignment (see code below which is copied from the current `main`).  This has the effect of breaking the `platform_config` if the user tries to override one or the other.

```python
        sch = tiledb.ArraySchema(
            domain=dom,
            attrs=attrs,
            sparse=True,
            allows_duplicates=False,
            offsets_filters=create_options.offsets_filters(),
            capacity=create_options.get("capacity", 100000),
            tile_order=cell_order,
            cell_order=tile_order,
            ctx=self._ctx,
        )
```

aaronwolen:
This updates the internal `tiledb_type_from_arrow_type()` utility to convert Arrow `large_utf8` to TileDB `STRING_ASCII`, just as it does for the `utf8` and `string` types. 

These conversions are temporary until utf8 filtering support is added to core.

### DIFF
--- a/apis/r/R/utils-arrow.R
+++ b/apis/r/R/utils-arrow.R
@@ -47,7 +47,6 @@ tiledb_type_from_arrow_type <- function(x) {
     double = "FLOAT64",
     boolean = "BOOL",
     bool = "BOOL",
-    # large_utf8 = "large_string",
     # large_string = "large_string",
     # binary = "binary",
     # large_binary = "large_binary",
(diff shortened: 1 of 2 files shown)


eddelbuettel (approved):
Looks good.  

With this, maybe try a rebase and then while at make the schema creation and use in the test file that much simpler?

aaronwolen:
> With this, maybe try a rebase and then while at make the schema creation and use in the test file that much simpler?

Done in [e33265d](https://github.com/single-cell-data/TileDB-SOMA/pull/591/commits/e33265db0d1e45e2a81c2d8fe2f982dfd34a8c23)

eddelbuettel (approved):
:heart:

Issue #484, opened 2022-10-27
Specify semantics of slot-indexing by [ ]


johnkerl:
See discussion at https://github.com/single-cell-data/TileDB-SOMA/issues/479#issuecomment-1292772632

cc @bkmartinjr @thetorpedodog

PR #593, opened 2022-12-09
Make index by empty list return zero-length results

johnkerl:
For #484.

Note that now that this PR has demonstrated feasibility, I need to put up a SOMA-spec PR as well: this is https://github.com/single-cell-data/SOMA/pull/53.

Context and prior art:

* In TileDB-Py, indexing by `[]` returns _all_ elements, not none. In a recent discussion with @ihnorton along with @bkmartinjr and others, we determined that that was a historical accident, not a precedent to be followed.
* For the return-nothing case TileDB-Py uses `tiledb.EmptyRange`. This is just a sentinel object within the package; what it is used for within the TileDB-Py query logic is much as is done on this PR. Namely, if that sentinel is set, then a core query is not done, and (typed) empty buffers are returned to the caller.
* This PR mimics TileDB-Py's logic that it uses for indexing by `tiledb.EmptyRange`, and connects that syntactically to indexing by `[]` within TileDB-SOMA-Py.


johnkerl:
Abandoned in favor of #618. I'll submit a separate PR for the one little Python bit which will remain.

PR #598, opened 2022-12-09
[python] Use soma_reader in SOMADenseNdArray read method

eddelbuettel:
This ~(still somewhat 'draft' though not marked as one)~ PR adds read method support via the TileDB SOMA C++ layer to the existing SOMADenseNdArray.  

It passes the existing tests, but we may want to expand the tests before merging.  It also can do with a bit of extra polish such as hiding the required conversion to inside the method (as opposed to the interface).   This also required some minor extensions to the `soma_reader` interface which can do with additional tests.  So we may want to let this sit for another day or two add polish.

It also open some more questions for which 'string values' we want. The side had so far stuck with 'ROW_MAJOR' and 'COL_MAJOR' (as TileDB Core seems to do), this now follows @gspowley and uses 'row-major' and 'column-major'.  The lower-case strings are actually a little easier on the eye so we can also default to these.  To be discussed...

### DIFF
--- a/apis/r/tests/testthat/test-SOMAReader-Arrow.R
+++ b/apis/r/tests/testthat/test-SOMAReader-Arrow.R
@@ -30,7 +30,7 @@ test_that("Arrow Interface from SOMAReader", {
     soma_reader(uri, columns, qc@ptr) |>
         arch::from_arch_array(arrow::RecordBatch) |>
         arrow::as_arrow_table() |>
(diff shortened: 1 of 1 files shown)


eddelbuettel:
This now ready.  Whenever you have a moment @aaronwolen @awenocur @johnkerl @gspowley

johnkerl (approved):

PR #599, opened 2022-12-10
[python/ci] Include Python 3.11 in CI matrix

johnkerl:
This is a trial balloon for #544, to see what we would need to do to support Python 3.11.

First result: at https://github.com/single-cell-data/TileDB-SOMA/actions/runs/3661926647/jobs/6190605155 we see `numba` is not yet supported in Python 3.11. cc @thetorpedodog @bkmartinjr @Shelnutt2 .

```
Collecting numba>=0.41.0
  Downloading numba-0.56.4.tar.gz (2.4 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 2.4/2.4 MB 146.6 MB/s eta 0:00:00
  Preparing metadata (setup.py): started
  Running command python setup.py egg_info
    File "<string>", line 2, in <module>
    File "<pip-setuptools-caller>", line 34, in <module>
    File "/tmp/pip-install-0tpo_aju/numba_b1ec8a6b557b450ea19a316e9668617a/setup.py", line 51, in <module>
      _guard_py_ver()
    File "/tmp/pip-install-0tpo_aju/numba_b1ec8a6b557b450ea19a316e9668617a/setup.py", line 48, in _guard_py_ver
      raise RuntimeError(msg.format(cur_py, min_py, max_py))
  RuntimeError: Cannot install on Python version 3.11.0; only versions >=3.7,<3.11 are supported.
```

See also https://github.com/numba/numba/issues/8304

Looks like they're hoping for a release candidate by end of 2022: https://github.com/numba/numba/issues/8304#issuecomment-1336218610

### DIFF
--- a/.github/workflows/python-ci-minimal.yml
+++ b/.github/workflows/python-ci-minimal.yml
@@ -18,7 +18,9 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-22.04]
-        python-version: ['3.10', '3.7']
+        # XXX TEMP ONLY FOR PRE-MERGE
+        # python-version: ['3.10', '3.7']
+        python-version: ['3.11']
         include:
           - runs-on: ubuntu-22.04
             cc: gcc-11
--- a/.github/workflows/python-ci-minimal.yml
+++ b/.github/workflows/python-ci-minimal.yml
@@ -18,9 +18,7 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-22.04]
-        # XXX TEMP ONLY FOR PRE-MERGE
-        # python-version: ['3.10', '3.7']
-        python-version: ['3.11']
+        python-version: ['3.10', '3.7']
         include:
           - runs-on: ubuntu-22.04
             cc: gcc-11
--- a/.github/workflows/python-ci-full.yml
+++ b/.github/workflows/python-ci-full.yml
@@ -21,7 +21,7 @@ jobs:
         # TODO: restore Windows build once we have C++/libtiledbsoma integration supported there
         os: [ubuntu-22.04, macos-12]
         # os: [ubuntu-22.04, macos-12, windows-2019]
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
         include:
           - runs-on: ubuntu-22.04
             cc: gcc-11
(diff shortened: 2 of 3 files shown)


johnkerl:
Putting this one on pause for a while

gsakkis:
Alternative suggestion: make numba an optional dependency and fall back to no-op if it's not installed.

bkmartinjr:
> Alternative suggestion: make numba an optional dependency and fall back to no-op if it's not installed.

This would have a dramatic (negative) effect on performance if we did this without alternative implementations of the ExperimentAxisQuery CSR handling code.  The numba-ized code is very, very, very slow if run in cpython, and there are "less slow" alternative code paths we could add as the bail-out option.

I think we should wait, but that is my "keep it simple" viewpoint

johnkerl:
@gsakkis we require `numba` rather crucially in `somacore`'s `query`

I am running this every so often as a (very) lazy way to see if Python 3.11 support for `numba` has shipped yet

johnkerl:
@gsakkis merging @bkmartinjr 's comments and mine:

* `numba` is crucial
* We can announce `tiledbsoma` support for Python 3.11 only when there is `numba` support for Python 3.11

gsakkis:
Apologies for the drive-by comment, just thought that "slow is better than nothing" :slightly_smiling_face:

johnkerl:
Really appreciate it & keep the drive-bys coming @gsakkis ! :) In _this situation_ nothing is better than slow (for 3.11 release) but in other situations, that might be just fine :) 

Having the fresh perspective is really awesome & I'm glad you're bringing it :)

bkmartinjr:
> Apologies for the drive-by comment, just thought that "slow is better than nothing" slightly_smiling_face

Keep the comments coming - we all appreciate the skills & expertise you bring!

bkmartinjr:
FYI, Numba status on Python 3.11 support is here: numba/numba#8304

TL;DR - latest estimate is sometime in late Q123

johnkerl:
There are now wheels at https://pypi.org/project/numba/#history

.github/workflows/python-ci-minimal.yml
```diff
@@ -18,7 +18,9 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-22.04]
-        python-version: ['3.10', '3.7']
+        # XXX TEMP ONLY FOR PRE-MERGE
```
johnkerl, line 21: i will change this after PR accept

atolopko-czi (approved):

PR #601, opened 2022-12-10
python package sdist

mlin:
WIP


mlin:
@johnkerl @gspowley @eddelbuettel shopping this for early feedback (sdist-only) as it got a little gnarly...

The main challenge is that inside the sdist tarball, we need `TileDB-SOMA/apis/python` to be the root directory (with setup.py etc.), and the libtiledbsoma extension code to be nested within that directory -- contrary to our current repo organization where `TileDB-SOMA/libtiledbsoma/` and `TileDB-SOMA/apis/` are sibling directories.

I added symlinks under `TileDB-SOMA/apis/python/dist_links` to the necessary sibling directories so that their contents get included in the sdist, albeit in this new location; and made our `setup.py:find_or_build()` figure out the right place to find them depending whether we're in the repo or an extracted sdist.

The next problem is that `setuptools_scm`, which we use to auto-populate the package version based on the git tags, seems to have some issues that break the sdist when the python package lives in a subdirectory of the git repo rather than the root. https://github.com/pypa/setuptools_scm/issues/788 https://github.com/pypa/setuptools_scm/issues/188 

I dropped `setuptools_scm` in favor of a [version.py script](https://gist.github.com/mina86/8782771) I've used in other past projects. It serves a similar function for the most part, with one subtle difference that it wants the git tags to start with 'v' e.g. `v0.5.1` instead of just `0.5.1`.

With these workarounds, I modified the CI to build and install the sdist before most of the other steps, ensuring it includes all the necessary libtiledbsoma sources. I don't exactly love the workarounds though, so welcome any feedback/suggestions..

gspowley:
It looks reasonable to me, pretty minimal considering the directory structure requirements. People will need to get used to tagging with 'v', or maybe the script can be modified?

mlin:
Continued in #605

PR #605, opened 2022-12-12
Build python sdist & wheels

mlin:
1. Build working sdist with `cd apis/python; python setup.py sdist`, working around some complications related to the location of the libtiledbsoma extension source in our repo.
2. Use [cibuildwheel](https://github.com/pypa/cibuildwheel) in GitHub Actions to build bdist wheels from the sdist. 

Details elaborated in PR comments/review.


mlin:
This branch builds on #601 with the sdist build; reproducing comment there on the roadbumps/workarounds for the sdist part:

> The main challenge is that inside the sdist tarball, we need TileDB-SOMA/apis/python to be the root directory (with setup.py etc.), and the libtiledbsoma extension code to be nested within that directory -- contrary to our current repo organization where TileDB-SOMA/libtiledbsoma/ and TileDB-SOMA/apis/ are sibling directories.
> To avoid reorganizing the repo, I added symlinks under TileDB-SOMA/apis/python/dist_links to the necessary sibling directories so that their contents get included in the sdist, albeit in this new location; and made our setup.py:find_or_build() figure out the right place to find them depending whether we're in the repo or an extracted sdist.
> The next problem is that setuptools_scm, which we use to auto-populate the package version based on the git tags, seems to have some issues that break the sdist when the python package lives in a subdirectory of the git repo rather than the root. https://github.com/pypa/setuptools_scm/issues/788 https://github.com/pypa/setuptools_scm/issues/188
> I dropped setuptools_scm in favor of a [version.py script](https://gist.github.com/mina86/8782771) I've used in other past projects. It serves a similar function for the most part, with one subtle difference that it wants the git tags to start with 'v' e.g. v0.5.1 instead of just 0.5.1.
> With these workarounds, I modified the CI to build and install the sdist before most of the other steps, ensuring it includes all the necessary libtiledbsoma sources. I don't exactly love the workarounds though, so welcome any feedback/suggestions..

mlin:
Starting from the sdist (#601), this adds a new GitHub Actions workflow using [cibuildwheel](https://github.com/pypa/cibuildwheel) to build the wheels for {various python versions} ⨯ {[manylinux2014](https://github.com/pypa/manylinux), macOS}. Just x86-64 for now, but we can add on ARM builds in the future.

[Example run](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/3683030215) with the sdist and wheels found in the Artifacts. The build takes 30-45 minutes, so I enabled it only for main and releases, as well as manual dispatch.

cibuildwheel provides [helpful instructions](https://cibuildwheel.readthedocs.io/en/stable/deliver-to-pypi/#delivering-to-pypi) for publishing that all to PyPI, either manually or automatically. Perhaps we'll start out manually and then automate it once we're happy with the products. Once we're up on PyPI, then we should be able to set up the conda feedstock to pull from the published sdist.

@johnkerl, one thing throwing the CI off, it seems that the Python Linting builds for this PR are pulling in a newer version of `isort`, which wants changes relative to to builds on main and other branches, even though they're interleaved in time and I didn't (intentionally) change the requested version. Seen anything like that before, is there some kind of stateful caching that could be going on?

bkmartinjr:
I'm not really qualified to review this, but I did review, and it looks reasonable to me. Please feel free to land it with approvals from others on the team.  Thank you @mlin !

scripts/bld
```diff
@@ -86,7 +86,7 @@ printf "Building ${build} build\n"
 
 # cd to the top level directory of the repo
 if [ "${r_build}" == "OFF" ]; then 
-  cd "$(git rev-parse --show-toplevel)"
+  cd "$(dirname "$0")/.."
```
johnkerl, line 89: nice!

johnkerl (approved):
🚢

scripts/bld
gspowley: Thanks for cleaning that up.

gspowley (approved):
LGTM!

libtiledbsoma/src/pyapi/CMakeLists.txt
```diff
@@ -1,5 +1,12 @@
+include(CMakePrintHelpers)
+
 # https://pybind11.readthedocs.io/en/stable/compiling.html#findpython-mode
-find_package(Python COMPONENTS Interpreter Development)
+# But FindPython mode doesn't necessarily find the "activated" version. Therefore set
+# PYTHON_EXECUTABLE explicitly to ensure the correct python version is used (as the manylinux
+# container used in cibuildwheel will have multiple python versions installed).
```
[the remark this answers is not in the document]
johnkerl: +1

eddelbuettel (approved):
LGTM -- with that caveat that I know much less about Python packaging...

mlin:
Got my issues with the linters sorted out, they were just following the new `dist_links` symlinks into python code we weren't covering with them. Added those to the respective exclusion lists for now.

bkmartinjr (approved):

PR #608, opened 2022-12-13
[python] Rename NdArray -> NDArray

johnkerl:
Conforms with spec update https://github.com/single-cell-data/SOMA/pull/55


thetorpedodog (approved):
find and replace. one of the best commands. we love find and replace, don’t we?

bkmartinjr:
FYI, this breaks (changes) the on-disk format, immediately rendering old data incompatible.  Any way we can make it backwards compat (ie. look for a soma metadata key of `NdArray` or `NDArray`?)

johnkerl:
https://github.com/single-cell-data/TileDB-SOMA/pull/610

PR #609, opened 2022-12-13
[r] Rename NdArray -> NDArray

johnkerl:
Conforms with spec update https://github.com/single-cell-data/SOMA/pull/55


eddelbuettel (approved):
Looks good, I trust you 'grepped globally' and as it still passes tests ...

All small steps towards the spec are good steps.

aaronwolen (approved):

PR #610, opened 2022-12-14
Support reading data that uses old NdArray name

bkmartinjr:
Revert part of #608, allowing the continued use of older datasets.


thetorpedodog:
Could we make it entirely case-insensitive?

johnkerl (approved):

bkmartinjr:
> Could we make it entirely case-insensitive?

We could, but that really doesn't solve the underlying problem (which IMHO is a missing test and code that doesn't clearly differentiate between the class name and the storage format "magic string").

Probably better to create a set of global Literal variables that contain the magic strings used in the storage format, rather than assuming they are eq with the class name and/or SOMA type....

johnkerl:
> Could we make it entirely case-insensitive?

Discussed off-GitHub: everyone is OK with this as-is.

PR #614, opened 2022-12-15
[r] Making package R CMD check clean

eddelbuettel:
This PR removes an `exportPattern()` declaration added earlier. Its effect is to render a number of internal helper functions 'exported' and hence in need of documentation.  With the change, the internal functions stay internal, and the package passes `R CMD check --as-cran` cleanly.


aaronwolen (approved):
Thanks!

PR #616, opened 2022-12-16
Shorten class constants using `Final` typing thing.

thetorpedodog:
It turns out you can fulfil the requirements of abstract properties using just members. `Final` gives users some static analysis hints that maybe they shouldn't be writing to that field.

---

As seen in somabase: https://github.com/single-cell-data/SOMA/pull/59

### DIFF
--- a/apis/python/src/tiledbsoma/io.py
+++ b/apis/python/src/tiledbsoma/io.py
@@ -332,7 +332,7 @@ def create_from_matrix(
         platform_config=platform_config,
     )
 
-    if soma_ndarray.soma_type == "SOMADenseNDArray":
+    if isinstance(soma_ndarray, DenseNDArray):
         _write_matrix_to_denseNDArray(soma_ndarray, src_matrix)
     else:  # SOMmASparseNDArray
         _write_matrix_to_sparseNDArray(soma_ndarray, src_matrix)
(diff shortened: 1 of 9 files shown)


thetorpedodog:
the countdown part is because it’s fewer lines than it used to be

apis/python/src/tiledbsoma/io.py
```diff
@@ -332,7 +332,7 @@ def create_from_matrix(
         platform_config=platform_config,
     )
 
-    if soma_ndarray.soma_type == "SOMADenseNDArray":
+    if isinstance(soma_ndarray, DenseNDArray):
```
thetorpedodog, line 335: For a different change, but this seems like it would be better served by putting a `_write_matrix(self, src_matrix)` method on each of `SparseNDArray` and `DenseNDArray`

johnkerl: @thetorpedodog please file a GH issue! :)

johnkerl (approved):
❤️ ❤️ ❤️ ❤️ ❤️ ❤️ 

🚢

thetorpedodog:
rebased and also pushed [a change to move what I think is the remaining thing that should be moved to `typing_extensions`](https://github.com/single-cell-data/TileDB-SOMA/compare/37548b3422615025c0b985af25a5fdee45e973fa..c7c2308449b4e904ce08e43aa2adfd561acff03e)

PR #617, opened 2022-12-19
[c++] Add clang-format check to CI

gspowley:


.github/workflows/format.yml
```diff
@@ -33,3 +33,8 @@ jobs:
           # python -m pip install mypy types-setuptools
           python -m pip install mypy==0.982 types-setuptools
           python -m mypy .
+      - name: Check C++ Format
+        run: |
+          cd ../..
+          ./scripts/run-clang-format.sh . clang-format 0 \
```
eddelbuettel, line 39: Should we make that the 'versioned' `clang-format` as _e.g._ `clang-format-14`?  Which may change if/when GitHub changes the runner.  Or did you leave it open on purpose to be upgrade-safe?

gspowley: This check was lifted from TileDB-VCF, so I can't claim I left it open on purpose. We can make it versioned, if that's preferred.

eddelbuettel: I honestly do not know how I feel about it. In Core, with the version having been `clang-format-9` for so long I had to jump though hoops to get to that.  So `clang-format` (no qualifiers) is preferred. For me, on this 22.04 machine, it is currently `clang-format-14`. And I think before too long we would have to jump to `-15`. I have no good feeling how much the output chanages but I _suspect_ not so much.  So what you have is 'fine by me' even if different from Core.

gspowley: Let's leave it un-versioned. Based on VCF using an un-versioned `clang-format` for 3+ years, I agree, the output should not change much.

johnkerl (approved):

PR #618, opened 2022-12-19
[c++] Support empty range query

gspowley:
This PR provides the empty range query support described in #593 (a query containing only empty ranges should return empty results).

While reviewing #593, I realized some C++ refactoring would make empty range queries easier to support.

**Note** - This PR does not contain the python API changes in #593, in case we want to keep those changes separate.

### DIFF
--- a/libtiledbsoma/test/test_soma_reader.py
+++ b/libtiledbsoma/test/test_soma_reader.py
@@ -78,6 +78,24 @@ def test_soma_reader_dim_points():
     assert sr.results_complete()
     assert arrow_table.num_rows == len(obs_id_points)
 
+def test_soma_reader_empty_dim_points():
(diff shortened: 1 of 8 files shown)


libtiledbsoma/test/test_soma_reader.py
```diff
@@ -78,6 +78,24 @@ def test_soma_reader_dim_points():
     assert sr.results_complete()
     assert arrow_table.num_rows == len(obs_id_points)
 
+def test_soma_reader_empty_dim_points():
```
johnkerl, line 81: Nice -- as this is direct from my #593 this proves you're accomplishing the same thing I was

johnkerl (approved):
🚢

johnkerl:
Was reverted & needs to be re-submitted

PR #622, opened 2022-12-20
[c++] Support empty range query and Arrow nullable fields

gspowley:
This PR:
* Provides the empty range query support described in #593 (a query containing only empty ranges should return empty results).

* Fixes an issue in #618 related to nullable, variable length fields (a TileDB core library check fired if the offsets buffer and the validity buffer were not the same size). We still allocate an extra offset buffer entry as required for Arrow, but hide the extra buffer entry from TileDB core.

* Adds support for Arrow nullable fields, by reusing the bytemap to bitmap code written for TileDB `bool` to Arrow `bool` conversion.


eddelbuettel (approved):
Looks good to me -- R package builds and tests fine without any issues.

PR #624, opened 2022-12-21
[c++] Add SOMAReader reset for stateful API changes

gspowley:
This PR adds the capability of "resetting" a SOMAReader while holding the underlying TileDB array open.
* These changes work with the existing python API. 
* Additional minor C++ changes will be needed to match the stateful open/close API when implemented in python.


eddelbuettel (approved):
This looks good to me and appears to be something I should be able to make use of from R. 

One suggestion may be to also commit a png or svg of the uml diagram.

gspowley:
Good point @eddelbuettel, let me know if you find any changes that would help adoption in R.

I'll commit an svg of the diagram too.

eddelbuettel:
'Native code' called from R is usually asynchronous: you call a compiled routine, it sets up what it needs, does work, returns.  Intermediate are usually gone.   (Of course stateful ops happen to, standard example is a DB handle.)  So I would have to find a use case beyond what we already have where we init an 'iterating reader' and call 'next chunk' til done. It may fit best there. But @aaronwolen and I have not yet exposed that one...

All good stuff.

Issue #391, opened 2022-10-10
Ingestor lacks support for AWS regions outside us-east


bkmartinjr:
If I try to write to a bucket in another region, it fails to look at the default region or allow a CLI override.  Ideally it would utilize the standard AWS_ env variables.

```
$ tools/ingestor anndata/pbmc-small.h5ad s3://bruce-uswest-tmp/tmp/pbmc-small-2022-10-10-n
Traceback (most recent call last):
  File "/home/bruce/projects/TileDB-SOMA/apis/python/tools/ingestor", line 264, in <module>
    main()
  File "/home/bruce/projects/TileDB-SOMA/apis/python/tools/ingestor", line 167, in main
    ingest_one(
  File "/home/bruce/projects/TileDB-SOMA/apis/python/tools/ingestor", line 242, in ingest_one
    tiledbsoma.io.from_h5ad(exp, input_path, "mRNA")
  File "/home/bruce/projects/TileDB-SOMA/apis/python/src/tiledbsoma/io.py", line 42, in from_h5ad
    _from_h5ad_common(experiment, input_path, measurement_name, from_anndata, ctx=ctx)
  File "/home/bruce/projects/TileDB-SOMA/apis/python/src/tiledbsoma/io.py", line 74, in _from_h5ad_common
    handler_func(experiment, anndata, measurement_name, ctx)
  File "/home/bruce/projects/TileDB-SOMA/apis/python/src/tiledbsoma/io.py", line 121, in from_anndata
    experiment.create()
  File "/home/bruce/projects/TileDB-SOMA/apis/python/src/tiledbsoma/soma_experiment.py", line 56, in create
    super().create()
  File "/home/bruce/projects/TileDB-SOMA/apis/python/src/tiledbsoma/soma_collection.py", line 106, in create
    tiledb.group_create(uri=self._uri, ctx=self._ctx)
  File "/home/bruce/projects/TileDB-SOMA/venv/lib/python3.10/site-packages/tiledb/group.py", line 270, in create
    lt.Group._create(cctx, uri)
tiledb.cc.TileDBError: [TileDB::S3] Error: Error while listing with prefix 's3://bruce-uswest-tmp/tmp/pbmc-small-2022-10-10-n/__group/' and delimiter '/'
Exception:  PermanentRedirect
Error message:  Unable to parse ExceptionName: PermanentRedirect Message: The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.
[...]

PR #625, opened 2022-12-22
Honor AWS_DEFAULT_REGION if no overriding ctx is specified

johnkerl:
Addresses issue #391

This is temporary pending TileDB-internal Shortcut task 23827

Validation/use-case:

```
$ export AWS_DEFAULT_REGION=us-west-2
$ python
>>> import tiledbsoma as soma
>>> exp = soma.Experiment("s3://bucket-which-is-in-us-west-2/path/to/soma/experiment")
>>> print(exp)
```

This is crucial for good out-of-the-box UX and uptake/exploration.

(Of course, if a config region is supplied,it is honored -- this mod applies solely for users who don't have a `ctx`/`cfg` and perhaps haven't even learned what those are yet.)

### DIFF
--- a/apis/python/src/tiledbsoma/tiledb_object.py
+++ b/apis/python/src/tiledbsoma/tiledb_object.py
@@ -1,5 +1,6 @@
+import os
 from abc import ABC, abstractmethod, abstractproperty
-from typing import Optional, Union
+from typing import Optional, Union, cast
 
 import tiledb
 
@@ -33,6 +34,9 @@ class TileDBObject(ABC):
         """
         Initialization-handling shared between ``TileDBArray`` and ``Collection``.  Specify ``tiledb_platform_config`` and ``ctx`` for the top-level object; omit them and specify parent for non-top-level objects. Note that the parent reference is solely for propagating options, ctx, display depth, etc.
         """
+
+        if ctx is None:
+            ctx = self._default_ctx()
         self._uri = uri
         if parent is None:
             self._ctx = ctx
@@ -48,6 +52,29 @@ class TileDBObject(ABC):
         # Null ctx is OK if that's what they wanted (e.g. not doing any TileDB-Cloud ops).
 
         self.metadata = MetadataMapping(self)
+
+    def _default_ctx(self) -> tiledb.Ctx:
+        """
+        The TileDB context used when no other is supplied. Must have good defaults for positive
+        out-of-the-box UX.
+        """
+
+        cfg = {}
+
+        # Crucial for reducing number of HTTP requests for tiledb-cloud URIs.
+        # With default (small) value, it can take many HTTP-request round trips to satisfy
+        # a single query.
+        cfg["py.init_buffer_bytes"] = 4 * 1024**3
+
+        # This is necessary for smaller tile capacities when querying with a smaller memory budget.
+        cfg["sm.mem.reader.sparse_global_order.ratio_array_data"] = 0.3  # type: ignore
+
+        # Temp workaround pending https://app.shortcut.com/tiledb-inc/story/23827
+        region = os.getenv("AWS_DEFAULT_REGION")
+        if region is not None:
+            cfg["vfs.s3.region"] = cast(str, region)  # type: ignore
+
+        return tiledb.Ctx(cfg)
 
     def delete(self) -> None:
         """
--- a/apis/python/src/tiledbsoma/tiledb_object.py
+++ b/apis/python/src/tiledbsoma/tiledb_object.py
@@ -61,13 +61,8 @@ class TileDBObject(ABC):
 
         cfg = {}
 
-        # Crucial for reducing number of HTTP requests for tiledb-cloud URIs.
-        # With default (small) value, it can take many HTTP-request round trips to satisfy
-        # a single query.
-        cfg["py.init_buffer_bytes"] = 4 * 1024**3
-
         # This is necessary for smaller tile capacities when querying with a smaller memory budget.
-        cfg["sm.mem.reader.sparse_global_order.ratio_array_data"] = 0.3  # type: ignore
+        cfg["sm.mem.reader.sparse_global_order.ratio_array_data"] = 0.3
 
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -49,6 +53,29 @@ def __init__(
...
+        The TileDB context used when no other is supplied. Must have good defaults for positive
+        out-of-the-box UX.
+        """
+
+        cfg = {}
+
+        # Crucial for reducing number of HTTP requests for tiledb-cloud URIs.
+        # With default (small) value, it can take many HTTP-request round trips to satisfy
+        # a single query.
+        cfg["py.init_buffer_bytes"] = 4 * 1024**3
```
bkmartinjr, line 67: Nit:  shouldn't this be managed centrally by SOMAReader (C++ layer)?

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -49,6 +53,29 @@ def __init__(
...
+
+        # This is necessary for smaller tile capacities when querying with a smaller memory budget.
+        cfg["sm.mem.reader.sparse_global_order.ratio_array_data"] = 0.3  # type: ignore
+
+        # Temp workaround pending https://app.shortcut.com/tiledb-inc/story/23827
+        region = os.getenv("AWS_DEFAULT_REGION")
```
bkmartinjr, line 73: very happy to have this change!  however, this only fixes it when the  user is not providing a configu/ctx (i.e., it sets it in the default context, but not otherwise).

Ideally, it would "honor" the environment variable in any situation where the context does not have "vfs.s3.region" set, and the environment variable is set...    Is such a thing possible, or does it need to wait for the real fix?

In the case of the cell census, we _always_ provide a tiledb context when opening the census, so the above fix doesn't help us....

bkmartinjr (approved):
As far as this change goes, it looks good. But it doesn't resolve or work around the issue for any user that provides a config/context (for other reasons).  Nice to have a more general workaround, if that is feasible

apis/python/src/tiledbsoma/tiledb_object.py
johnkerl: For `main` yes I suppose so. We don't have _all_ read paths going through C++, but yes, all _bulk_ read paths should. (Remaining reads through TileDB-Py are metadata etc.) So yes this can go away.

apis/python/src/tiledbsoma/tiledb_object.py
johnkerl: > Is such a thing possible, or does it need to wait for the real fix?

Yes, very much doable -- good feedback, thanks!!

johnkerl: ... oops, correction to what I said, @bkmartinjr -- even if you make a `ctx = tiledb.Ctx({})` totally plain-vanilla and _you_ (the caller) aren't setting `vfs.s3.region` ... still `ctx.Config()['vfs.s3.region']` _does_ have `us-east-1` set as the default. (Which is the core bug.) So at the spot on this PR we can't find out if the caller explicitly specified `vfs.s3.region` or not. :(

So what I think we need to do for this PR -- temporarily, for now, only pending TileDB-internal tracking task 23827, and only for soma -- is maybe _always_ use `AWS_DEFAULT_REGION` if it exists in the env -- ? What do you think?

cc @ihnorton

johnkerl: Also if we don't have closure today, I'd like to merge this PR as-is -- as it is an improvement on the out-of-the-box UX which will help uptake -- and try to go farther on a follow-on PR.

johnkerl:
Also sorry for the confusion @bkmartinjr -- just now _six_ of my comments appeared -- three old, three new -- it turns out I had pending comments from yesterday I hadn't poked the 'submit' button on :(

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -49,6 +53,29 @@ def __init__(
...
+        """
+
+        cfg = {}
+
+        # Crucial for reducing number of HTTP requests for tiledb-cloud URIs.
+        # With default (small) value, it can take many HTTP-request round trips to satisfy
+        # a single query.
+        cfg["py.init_buffer_bytes"] = 4 * 1024**3
+
+        # This is necessary for smaller tile capacities when querying with a smaller memory budget.
```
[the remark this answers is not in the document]
bkmartinjr: I don't understand why we would have a different default than  TileDB-Py.   4GB per column is a LOT of memory for a default....

apis/python/src/tiledbsoma/tiledb_object.py
bkmartinjr: It seems like there are edge cases either way - I would rather not have the environment variable override the explicitly set context (that would screw up the cell-census package in a big way ,as our bucket is in us-west-2, and who knows what value the user will have their environment variable set too).

My pref:  land this as is.

atolopko-czi:
Should `AWS_REGION` be considered as well, and be given precedence over `AWS_DEFAULT_REGION`, similar to [AWS CLI behavior](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html#envvars-list)? 

I see boto3 only considers `AWS_DEFAULT_REGION`, and not `AWS_REGION`.  Though boto3 _does_ also use `AWS_PROFILE`, from which it looks up the region in `~/.aws/config`.

bkmartinjr:
IMHO: we should be 100% compat/aligned with boto3.

eddelbuettel:
Thanks @atolopko-czi I just circled that back to our internal ticket so that we make sure that the relevant region variable get automagically handled as we do with the key pair (by essentially letting the AWS SDK do its things in not getting in the way).

johnkerl:
More follow-ups to come on separate PR(s) as discussed here, thanks folks!!

PR #627, opened 2022-12-22
[python] Less-in-core ingest from H5AD [main-old]

johnkerl:
Context: Shortcut story 24294 and #641


apis/python/src/tiledbsoma/io.py
```diff
@@ -196,7 +197,6 @@ def _from_anndata_aux(
 
     anndata.obs_names_make_unique()
     anndata.var_names_make_unique()
-    anndata = tiledbsoma.util_ann._decategoricalize(anndata)
```
johnkerl, line 199: Since we got `anndata = ad.read_h5ad(h5ad_file_name, "r")` with the `"r"` this entire data structure is no longer reconstructible (below). So we split out the decategoricalizations to where they're needed.

apis/python/src/tiledbsoma/raw_group.py
```diff
@@ -82,7 +82,7 @@ def from_anndata(
         self._add_object(self.var)
```
aaronwolen, line 82: Do `raw`'s `var`/`varm` slots need to be decategoricalized?

aaronwolen (approved):
I had one question but otherwise this is great.

apis/python/src/tiledbsoma/raw_group.py
johnkerl: good call, thanks @aaronwolen ! :)

johnkerl:
R CI fails are KPs pending core 2.13.1 and TileDB-R 0.17.1

PR #629, opened 2022-12-23
Fix and re-enable CI macOS builds

aaronwolen:
Fixes a minor issue with `copy_source.sh` resulting from the different behavior of `cp` on macOS and Linux. 

For posterity (and I because I will surely forget almost immediatley), the issue was that `cp -a` on macOS copies the contents of a directory rather than the directory itself if the source ends in a `/`.

Also bumps the minimum required version of _arch_ to 0.0.2 (thanks to @eddelbuettel).

### DIFF
--- a/apis/r/copy_source.sh
+++ b/apis/r/copy_source.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-test -d ../../libtiledbsoma && cp -a ../../libtiledbsoma/ src/
+test -d ../../libtiledbsoma && cp -a ../../libtiledbsoma src/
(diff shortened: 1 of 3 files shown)


apis/r/copy_source.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-test -d ../../libtiledbsoma && cp -a ../../libtiledbsoma/ src/
+test -d ../../libtiledbsoma && cp -a ../../libtiledbsoma src/
```
eddelbuettel, line 3: The darn `cp` command on macOS had tripped me up before.  It complains about the trailing slash?

eddelbuettel (approved):
A big cup of taadaa!

apis/r/copy_source.sh
aaronwolen: Ugh yeah, what a pain. I documented exactly what was going b/c I've also been bit by this before.

PR #631, opened 2022-12-29
[python] Update to setuptools 65.5.1 [main-old]

johnkerl:


johnkerl:
R CI failures are due to an unrelated KP: We get TileDB-R 0.17.0 which depends on TileDB Embedded 2.13.0 which does not have a fix for https://github.com/TileDB-Inc/TileDB/pull/3713 -- this should be fixed in a core 2.13.1

Self-stamping

PR #632, opened 2022-12-29
[python] ExperimentAxisQuery

### LINKED ISSUE #570: Locate Python soma-query logic within the TileDB-SOMA package
PRs:

* https://github.com/single-cell-data/TileDB-SOMA/pull/632

bkmartinjr:
Initial migration of the ExperimentQuery class to tiledbsoma.

Fixes #570 

*Reviewers*: please review naming, as I'm not entirely satisfied with it.


apis/python/src/tiledbsoma/experiment_query/axis.py
```diff
@@ -0,0 +1,71 @@
...
+    value_filter: Optional[AxisValueFilter] = None
+    coords: Optional[AxisCoordinates] = None
+
+    def __post_init__(self) -> None:
+        # TODO: Error class - should not be throwing generic Exceptions
+        if (self.value_filter is None) and (self.coords is None):
+            self.coords = (slice(None),)
+        if not (self.value_filter is None) != (self.coords is None):
+            raise Exception(
+                "FilterSpec - value_filter or coords may be specified, but not both."
```
johnkerl, line 50: IIUC @aaronwolen it's a legit use-case to specify both -- ?

apis/python/src/tiledbsoma/experiment_query/anndata.py
```diff
@@ -0,0 +1,40 @@
...
+import anndata
+import pyarrow as pa
+import scipy.sparse as sparse
+
+from .types import ExperimentQueryReadArrowResult
+
+
+def arrow_to_scipy_csr(X: pa.Table, shape: Tuple[int, int]) -> sparse.csr_matrix:
+    return sparse.csr_matrix(
+        (X["soma_data"].to_numpy(), (X["_dim_0"].to_numpy(), X["_dim_1"].to_numpy())),
```
johnkerl, line 12:
```
        (X["soma_data"].to_numpy(), (X["soma_dim_0"].to_numpy(), X["soma_dim_1"].to_numpy())),
```
?

bkmartinjr: Can you expand on `?`?  Happy to discuss, but not sure what the question is :-)

johnkerl: @bkmartinjr sorry I was recommending you replace

```
        (X["soma_data"].to_numpy(), (X["_dim_0"].to_numpy(), X["_dim_1"].to_numpy())),
```

with

```
        (X["soma_data"].to_numpy(), (X["soma_dim_0"].to_numpy(), X["soma_dim_1"].to_numpy())),
```

apis/python/src/tiledbsoma/experiment_query/axis.py
bkmartinjr: I think you are correct.  I will look into adding it.  It is not a use case for the cell census, which indexes only on the joinids, but I can see it being useful elsewhere.

apis/python/src/tiledbsoma/experiment_query/anndata.py
bkmartinjr: Actually, this code is correct as written.  See query.py::_rewrite_X_for_positional_indexing().

the idea is that AnnData requires positional indexing, not joinids.

johnkerl:
@bkmartinjr re

>  Verify feature parity with 0.1

I assert I've test-driven this and find it a more-than-adequate substitute for 1.0's `soma.query` (which I wrote)

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,499 @@
...
+
+    """
+    Private helper methods
+    """
+
+    @property
+    def _threadpool(self) -> concurrent.futures.ThreadPoolExecutor:
+        """Private threadpool cache"""
+        if self.__threadpool is None:
+            # TODO: the user should be able to set their own threadpool, a la asyncio's
```
johnkerl, line 373: ✅

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,499 @@
...
+            if X_collection[_xname].soma_type != "SOMASparseNDArray" or not isinstance(
+                X_collection[_xname], SOMASparseNDArray
+            ):
+                raise NotImplementedError("Dense array unsupported")
+            all_X_arrays[_xname] = cast(SOMASparseNDArray, X_collection[_xname])
+
+        obs_table, var_table = self._read_both_axis(column_names)
+
+        X_tables = {
+            # TODO: could also be done concurrently
```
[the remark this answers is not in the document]
johnkerl: I see at `def _threadpool` below we're on the same page here :)

apis/python/tests/test_experiment_query.py
```diff
@@ -0,0 +1,592 @@
...
+@pytest.mark.xfail(
+    # see comment on test_experiment_query_all
+    sys.version_info.major == 3 and sys.version_info.minor >= 10,
+    reason="typeguard bug #242",
+)
+@pytest.mark.parametrize("n_obs,n_vars", [(1001, 99)])
+def test_query_cleanup(soma_experiment: soma.Experiment):
+    """
+    Verify soma.Experiment.query works as context manager and stand-alone,
+    and htat it cleans up correct.
```
johnkerl, line 372: and that it cleans up correctly

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,499 @@
...
+    underlying resources. The ExperimentAxisQuery is a context manager, and it is recommended
+    that you use the following pattern to make this easy and safe:
+    ```
+        with ExperimentAxisQuery(...) as query:
+            ...
+    ```
+    """
+
+    experiment: "Experiment"
+    ms: str
```
johnkerl, line 74: This is super-confusing. Please change this from `ms` to `measurement_name`.

johnkerl (changes requested):
Amazing body of work @bkmartinjr !!!! :)

apis/python/src/tiledbsoma/experiment.py
```diff
@@ -70,3 +71,18 @@ def ms(self) -> CollectionBase[Measurement]:
         A collection of named measurements.
         """
         return cast(CollectionBase[Measurement], self["ms"])
+
+    def query_by_axis(
```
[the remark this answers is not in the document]
bkmartinjr: OK, will change.

apis/python/src/tiledbsoma/experiment_query/query.py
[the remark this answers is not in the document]
bkmartinjr: I am happy to do all the threads post-PR - would prefer to land this one so we can update users to this API (eg, the cell_census package), and then pivot to perf use cases.

What you found is exactly what logic would suggest, which is why I think this needs to be handled with a user-configured threadpool, and the above change to queue all X layers so they can be concurrent, if the user allows it.

Follow-up Issue to remind us: #684

apis/python/src/tiledbsoma/experiment_query/axis.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def __post_init__(self) -> None:
+        if self.value_filter is not None:
+            if not isinstance(self.value_filter, str):
+                raise TypeError("AxisQuery - value_filter must be a str type")
+
+        if self.coords is None:
+            self.coords = (slice(None),)
+        else:
+            if not isinstance(self.coords, tuple):
```
thetorpedodog, line 58: This seems stricter than you want?

apis/python/src/tiledbsoma/experiment_query/axis.py
```diff
@@ -0,0 +1,68 @@
...
+import numpy as np
+import numpy.typing as npt
+import pyarrow as pa
+
+AxisCoordinate = Union[slice, int, npt.ArrayLike]
+AxisCoordinates = Tuple[AxisCoordinate, ...]
+AxisValueFilter = str
+
+
+@dataclass()
```
thetorpedodog, line 13: in somacore we use `attrs`, which offers some niceties that might be useful here:

- `@some_attr.validator` adds a check that is run after initialization
- [converters](https://www.attrs.org/en/stable/init.html#converters) to transform the attribute into the form you want it

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,503 @@
...
+    ```
+    """
+
+    experiment: "Experiment"
+    measurement_name: str
+
+    _query: MatrixAxisQuery
+    _joinids: AxisJoinIds
+    _indexer: "AxisIndexer"
+    __threadpool: Optional[
```
thetorpedodog, line 79: As much as we probably don’t want the user to modify this, in my experience double-underscore-prefixed properties are more trouble than they’re worth.

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,503 @@
...
+        soma_dim_0: [[1538112,1538112,1538112,1538112,1538112,...,1538308,1538308,1538308,1538308,1538308], ...]
+        soma_dim_1: [[5,19,26,34,37,...,10577,10603,10616,10617,10655], ...]
+        soma_data: [[2,2,1,1,1,...,1,1,1,2,2], ...]
+        """
+        if not (
+            layer_name and layer_name in self.experiment.ms[self.measurement_name].X
+        ):
+            raise ValueError("Must specify X layer name")
+
+        X = self.experiment.ms[self.measurement_name].X[layer_name]
```
thetorpedodog, line 199: I know that we love `X` as a member name and an array name, but I’m still not a fan of using capitals for variables.

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,503 @@
...
+
+    experiment: "Experiment"
+    measurement_name: str
+
+    _query: MatrixAxisQuery
+    _joinids: AxisJoinIds
+    _indexer: "AxisIndexer"
+    __threadpool: Optional[
+        concurrent.futures.ThreadPoolExecutor
+    ]  # lazily created, use property accessor
```
thetorpedodog, line 81: It’s fine to have these declarations here, but you could add alternately omit them and add necessary annotations in the `__init__` method (which there will be fewer of, since many of the types can be inferred). Up to you.

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,503 @@
...
+        """
+        if not (
+            layer_name and layer_name in self.experiment.ms[self.measurement_name].X
+        ):
+            raise ValueError("Must specify X layer name")
+
+        X = self.experiment.ms[self.measurement_name].X[layer_name]
+        if X.soma_type != "SOMASparseNDArray":
+            raise NotImplementedError("Dense array unsupported")
+        assert isinstance(X, SOMASparseNDArray)
```
thetorpedodog, line 202: The equivalent:
```python
if not isinstance(x, SOMASparseNDArray):
  raise NotImplementedError(...)  # maybe TypeError?
```
would save the duplicate logic here.

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,503 @@
...
+    ```
+        with ExperimentAxisQuery(...) as query:
+            ...
+    ```
+    """
+
+    experiment: "Experiment"
+    measurement_name: str
+
+    _query: MatrixAxisQuery
```
johnkerl, line 76: This has been confusing me a bit, just in terms of readability. _After_ reading a few times I get it, but my concern is how to make it more immediately graspable on a first read.

Below we have `query = self._query[axis]` ... this is confusing since query is singular, queries (which we're not saying) is plural, and I was at first inclined to suggest `query = self._queries[axis]`. But `self._query` isn't a (plural) queries object, it's a `MatrixAxisQuery` ...

So my suggestion is to rename `_query` to `_matrix_axis_query` and then below have `axis_query = self._matrix_axis_query[axis]`

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,503 @@
...
+    ):
+        if not experiment.exists():
+            raise ValueError("Experiment does not exist")
+        if measurement_name not in experiment.ms:
+            raise ValueError("Measurement does not exist in the experiment")
+
+        self.experiment = experiment
+        self.measurement_name = measurement_name
+
+        self._query = {
```
johnkerl, line 99: `self._matrix_axis_query = ...`

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,503 @@
...
+        return self
+
+    def __exit__(self, *excinfo: Any) -> None:
+        self.close()
+
+    def obs(
+        self, *, column_names: Optional[Sequence[str]] = None
+    ) -> somacore.ReadIter[pa.Table]:
+        """Return obs as an Arrow table iterator."""
+        query = self._query["obs"]
```
johnkerl, line 130: `obs_query = self._matrix_axis_query["obs"]`

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,503 @@
...
+            ids=query.coords,
+            value_filter=query.value_filter,
+            column_names=column_names,
+        )
+
+    def var(
+        self, *, column_names: Optional[Sequence[str]] = None
+    ) -> somacore.ReadIter[pa.Table]:
+        """Return obs as an Arrow table iterator."""
+        query = self._query["var"]
```
johnkerl, line 141: `var_query = self._matrix_axis_query["var"]`

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,503 @@
...
+            X_tables = self._rewrite_X_for_positional_indexing(X_tables)
+
+        X = X_tables.pop(X_name)
+        query_result: ExperimentAxisQueryReadArrowResult = {
+            "obs": obs_table,
+            "var": var_table,
+            "X": X,
+        }
+        if len(X_layers) > 0:
+            assert len(X_layers) == len(X_tables)
```
thetorpedodog, line 325: Just to make sure: This is an internal sanity check, and external user code should never be able to trigger this (because if so, it should instead be a `raise SomeKindOfError`, since `assert`ions are assumed to be able to be stripped out with no impact on correctness).

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,503 @@
...
+            query_result["X_layers"] = X_tables
+
+        return query_result
+
+    def to_anndata(
+        self,
+        X_name: str,
+        *,
+        column_names: Optional[AxisColumnNames] = None,
+        X_layers: Optional[List[str]] = None,
```
thetorpedodog, line 335: Does this need to be exactly a list?

johnkerl (changes requested):
Sorry, didn't see this on my first read which was GitHub-UI only -- doing another read with local checkout ...

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,503 @@
...
+                    X_table[
+                        "soma_data"
+                    ].to_numpy(),  # as a side effect, consolidates chunks
+                ),
+                names=("_dim_0", "_dim_1", "soma_data"),
+            )
+        return new_X_tables
+
+
+class AxisIndexer:
```
thetorpedodog, line 473: This can be an `attrs` class (or a `dataclass`) with the `_obs/_var_index` defaults being `None`.

thetorpedodog (commented):
Something I notice here is that there’s actually nothing TileDB-specific about this query. It seems like (once the `Experiment` interface gets created) this could go into somacore itself.

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,503 @@
...
+    from ..experiment import Experiment
+from ..sparse_nd_array import SparseNDArray as SOMASparseNDArray
+from .anndata import _make_anndata
+from .axis import AxisQuery
+from .eq_types import AxisColumnNames, ExperimentAxisQueryReadArrowResult
+
+AxisJoinIds = TypedDict(
+    "AxisJoinIds",
+    {
+        "obs": pa.Array,
```
johnkerl, line 36: Here you have `pa.Array` not `Union[None, pa.Array` -- but at line 97 there is

```
        self._joinids = {
            "obs": None,
            "var": None,
        }
```

I don't know why the linter isn't saying anything ... 🤔

thetorpedodog: It doesn’t know about any pyarrow types, so it just makes them all `Any`s.

johnkerl: OK

Even for readability I think it would make sense to type these as `Union[None, pa.Array]`

apis/python/src/tiledbsoma/experiment_query/axis.py
bkmartinjr: We use `dataclasses` elsewhere in the package, which is why I went this direction.  If our intent  is to move to `attrs`, I'm happy to make this change.

Can I take the suggestion as indication we are migrating (over time of course)?

apis/python/src/tiledbsoma/experiment_query/query.py
bkmartinjr: yes, sanity.  will remove as it is really unnecessary.

apis/python/src/tiledbsoma/experiment_query/query.py
bkmartinjr: Nope.  I fixed it elsewhere, as well.

apis/python/src/tiledbsoma/experiment_query/query.py
bkmartinjr: I will add the optional typing to the declaration, for clarity.

bkmartinjr:
> Something I notice here is that there’s actually nothing TileDB-specific about this query. It seems like (once the Experiment interface gets created) this could go into somacore itself.

@thetorpedodog - this is true today, but I _suspect_ it will eventually specialize to take advantage of performance characteristics of underlying storage engines.  That said, I'm not sure which is better - subclass interfaces for this, or entirely separate implementations.

apis/python/src/tiledbsoma/experiment_query/query.py
bkmartinjr: change made

apis/python/src/tiledbsoma/experiment_query/axis.py
bkmartinjr: I converted to attrs.

Please note: I also simplified this by removing all conversion, and delegating it to `DataClass` where it belongs.  This removes some nice features (eg, automatic conversion of `numpy.typing.ArrayLike`), but I filed issues to request that those be added to the DataFrame & NDArray classes where they belong (#688)

apis/python/src/tiledbsoma/experiment_query/query.py
bkmartinjr: updated

apis/python/src/tiledbsoma/experiment_query/axis.py
```diff
@@ -1,34 +1,56 @@
...
-@dataclass()
+def normalize_coords(
+    coords: SparseDataFrameCoordinates,
+) -> Tuple[SparseDataFrameCoordinate]:
+    """
+    Private. Convert incoming query coordinates into pyarrow Arrow int64,
+    slices or ints.
+
+    NOTE: this should arguably be delegated to the DataFrame class, as the
+    ExperimentAxsQuery class has no dependency on this. However, the error
```
johnkerl: `ExperimentAxsQuery` -> `ExperimentAxisQuery`

bkmartinjr: was dead code - now removed

johnkerl (approved):

apis/python/src/tiledbsoma/__init__.py
```diff
@@ -1,3 +1,4 @@
+from . import experiment_query
```
thetorpedodog, line 1: This doesn’t seem like it should be necessary, since `from tiledbsoma import experiment_query` or `import tiledbsoma.experiment_query` will work without it.

apis/python/src/tiledbsoma/experiment_query/anndata.py
```diff
@@ -0,0 +1,52 @@
...
+    obs = obs.to_pandas()
+    obs.index = obs.index.map(str)
+
+    var = query_result["var"]
+    var = var.to_pandas()
+    var.index = var.index.map(str)
+
+    shape = (len(obs), len(var))
+
+    X = query_result.get("X", None)
```
thetorpedodog, line 41: nit: `, None` not needed here

(also, still lots of Capital letter variables around which irk me slightly (think of poor, innocent pep8!) but ultimately unimportant and fine in the grand scheme of things and I just need to breathe slowly and relax)

apis/python/src/tiledbsoma/experiment_query/axis.py
```diff
@@ -0,0 +1,76 @@
...
+    coords: SparseDataFrameCoordinates = attrs.field(
+        default=(slice(None),),
+        validator=attrs.validators.instance_of(tuple),
+    )
+
+    @coords.validator
+    def _validate_coords(self, _: Any, value: Any) -> None:
+        """
+        This should arguably be delegated to DataClass, but that would
+        def error reporting to the user. Doing validation proactively has
```
thetorpedodog, line 57: typo here?

apis/python/src/tiledbsoma/experiment_query/anndata.py
```diff
@@ -0,0 +1,52 @@
...
+    if X is not None:
+        X = _arrow_to_scipy_csr(X, shape)
+
+    X_layers = query_result.get("X_layers", {})
+    layers: Dict[str, sparse.csr_matrix] = {}
+    for X_layer_name, X_layer_table in X_layers.items():
+        layers[X_layer_name] = _arrow_to_scipy_csr(X_layer_table, shape)
```
thetorpedodog, line 48:
```suggestion
    layers = {name: _arrow_to_scipi_csr(table, shape) for name, table in X_layers.items()}
```

apis/python/src/tiledbsoma/experiment_query/anndata.py
```diff
@@ -0,0 +1,52 @@
...
+
+    return anndata.AnnData(
+        X=X, obs=obs, var=var, layers=(layers if len(layers) else None)
```
thetorpedodog, line 51:
```suggestion
        X=X, obs=obs, var=var, layers=layers or None
```
…to take advantage of the implicit falsity of an empty collection.

apis/python/src/tiledbsoma/experiment_query/eq_types.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+AxisColumnNames = TypedDict(
+    "AxisColumnNames",
+    {
+        "obs": Optional[Sequence[str]],  # None is all
+        "var": Optional[Sequence[str]],
+    },
+    total=False,
+)
```
thetorpedodog, line 23: Any reason for the difference between this TypedDict and the above one? Not a major problem but might be worth making consistent. This would also let you put docstring-style strings below each entry which tooling can pick up on in the future.

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,483 @@
...
+MatrixAxisQuery = TypedDict(
+    "MatrixAxisQuery",
+    {
+        "obs": AxisQuery,
+        "var": AxisQuery,
+    },
+)
+
+
+class ExperimentAxisQuery(ContextManager["ExperimentAxisQuery"]):
```
thetorpedodog, line 50: You could use [`contextlib.AbstractContextManager`](https://docs.python.org/3/library/contextlib.html) as the base here and avoid needing to implement your own `__enter__` method.

apis/python/tests/test_experiment_query.py
```diff
@@ -0,0 +1,591 @@
...
+
+@pytest.mark.xfail(
+    # This test fails on Python 3.10+ due to a bug in typeguard. The bug
+    # is tripped due to use of a TypedDict. Remove
+    # work-around when the typeguard issue is fixed AND released.
+    # Underlying issue:
+    #   https://github.com/agronholm/typeguard/issues/242
+    # Related to _when_ we might see a release:
+    #   https://github.com/agronholm/typeguard/issues/257
+    sys.version_info.major == 3 and sys.version_info.minor >= 10,
```
thetorpedodog, line 76: `sys.version_info >= (3, 10)`

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,483 @@
...
+from ..sparse_nd_array import SparseNDArrayRead
+
+if TYPE_CHECKING:
+    from ..experiment import Experiment
+from ..sparse_nd_array import SparseNDArray as SOMASparseNDArray
+from .anndata import _make_anndata
+from .axis import AxisQuery
+from .eq_types import AxisColumnNames, ExperimentAxisQueryReadArrowResult
+
+AxisJoinIds = TypedDict(
```
thetorpedodog, line 33: It looks like these are only used internally. Should they be `_`-prefixed? (Also, since they aren’t used for their dict-ness at all I would make them attrs classes but that’s up to you.)

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,483 @@
...
+        self._indexer: "AxisIndexer" = AxisIndexer(self)
+
+        self._threadpool_: Optional[concurrent.futures.ThreadPoolExecutor] = None
+
+    def close(self) -> None:
+        """
+        Clean up and close all resources. This must be called or the thread pool
+        will not be released.
+        """
+        if self._threadpool_ is not None:
```
thetorpedodog, line 106: nit: this can be `if self._threadpool_` since all threadpools are truthy

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,483 @@
...
+        obs_table, var_table = (f.result() for f in futures)
+        return obs_table, var_table
+
+    def read(
+        self,
+        X_name: str,
+        *,
+        use_position_indexing: bool = False,
+        column_names: Optional[AxisColumnNames] = None,
+        X_layers: Optional[Sequence[str]] = None,
```
thetorpedodog, line 270: Making this default an empty tuple...
```suggestion
        X_layers: Sequence[str] = (),
```

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,483 @@
...
+    ) -> ExperimentAxisQueryReadArrowResult:
+        """
+        Read the _entire_ query result into (in-memory) Arrow Tables. This is a
+        low-level routine intended to be used by loaders for other in-core
+        formats, such as AnnData, which can be created from the resulting Tables.
+        """
+        if column_names is None:
+            column_names = AxisColumnNames(obs=None, var=None)
+        X_collection = self.experiment.ms[self.measurement_name].X
+        X_layers = [] if X_layers is None else X_layers
```
thetorpedodog, line 280: ...means this line can be eliminated entirely.

apis/python/src/tiledbsoma/experiment_query/query.py
```diff
@@ -0,0 +1,483 @@
...
+        all_X_names = [X_name] + list(X_layers)
+        all_X_arrays: Dict[str, SOMASparseNDArray] = {}
+        for _xname in all_X_names:
+            if not isinstance(_xname, str) or not _xname:
+                raise ValueError("X layer names must be specified as a string.")
+            if _xname not in X_collection:
+                raise ValueError("Unknown X layer name")
+            if not isinstance(X_collection[_xname], SOMASparseNDArray):
+                raise NotImplementedError("Dense array unsupported")
+            all_X_arrays[_xname] = cast(SOMASparseNDArray, X_collection[_xname])
```
thetorpedodog, line 290: Because of the `isinstance` check above, I believe this `cast` is no longer necessary.
```suggestion
            all_X_arrays[_xname] = X_collection[_xname]
```

apis/python/src/tiledbsoma/experiment_query/eq_types.py
```diff
@@ -0,0 +1,23 @@
+from typing import Dict, Optional, Sequence
+
+import pyarrow as pa
+from typing_extensions import TypedDict
+
+
+class ExperimentAxisQueryReadArrowResult(TypedDict, total=False):
```
thetorpedodog, line 7: Looking at this again: is there a reason this is a TypedDict rather than an attrs class? It seems like that would let you make `_make_anndata` as a member function and also let users access `result.obs`/`result.var`/`result.X` etc. (unless there is a reason that they definitely *want* to do it using dict access patterns.

thetorpedodog (approved):
I have a variety of comments, most to all of which are small, and the rest of which could be addressed in a follow-up if you would prefer.

An overarching thing I noticed was using `TypedDict`s as *return* types from some of the query functions, which seems odd to me—I can see taking them as parameters to make calling more straightforward, but in general for a return I find using a real concrete type to be better, because it provides some measure of runtime type safety (rather than just being “a dict”, it’s an object with its own semantics).

That said, none of these are blockers and it’s probably best to get this *in* for a start so we can polish and build upon it.

.github/workflows/format.yml
```diff
@@ -34,8 +34,8 @@ jobs:
         run: |
           # mypy 0.990 has false positives on `import anndata as ad` and `Module has no attribute
           # "read_h5ad"` etc :(
-          # python -m pip install mypy types-setuptools
-          python -m pip install mypy==0.982 types-setuptools
+          # python -m pip install mypy types-setuptools typeguard
+          python -m pip install mypy==0.982 types-setuptools typeguard 'attrs>=22.1'
```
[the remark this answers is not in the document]
bkmartinjr: I believe this is detritus - will remove.

apis/python/tests/test_experiment_query.py
bkmartinjr: :man_facepalming: someday i need to learn Python...

apis/python/src/tiledbsoma/experiment_query/query.py
bkmartinjr: Actually, the suggested method fails mypy static analysis, whereas the explicit None test passes.  Leaving as is.

apis/python/src/tiledbsoma/experiment_query/query.py
bkmartinjr: made this change.  it cleans up some areas, but results in a lot of getattr/setattr code where the axis was parameterized.  Mixed result, but I think probably worth doing the change.

apis/python/src/tiledbsoma/experiment_query/query.py
bkmartinjr: mypy isn't smart enough unless I introduce an intermediate variable.  which I will do.

apis/python/src/tiledbsoma/experiment_query/eq_types.py
bkmartinjr: There isn't a good reason - I'll give it a try.  It feels like it may need a later round of refinement to align better with the new fluent method, but it is a move in that direction....

apis/python/src/tiledbsoma/experiment_query/query.py
bkmartinjr: I am unable to get AbstractContextManager to work reliably. On older versions of Python, it is not a subscriptable (generic), but mypy is failing when I omit the parameter.

For now, reverting to the contextlib.ContextManager, which seems to work correctly across all versions, albeit requiring an `__enter__` implementation.

PR #634, opened 2023-01-03
[r] Adjust SparseNDArray test to renamed soma_data method

eddelbuettel:
This PR updates one test to refer to the now-renamed `soma_data` method (see #397 for original rationale).


aaronwolen (approved):
Thanks!

Issue #602, opened 2022-12-12
[python] Python doc-string audits


johnkerl:
Two phases:

* Make sure everything has something at all -- something that will appear when a user types `help(soma.foo)` as rendered by the Python interpreter itself
* Tweaking style (single vs double backticks, etc etc etc) which control how docstrings are rendered by various systems including `readthedocs`

This issue is about the former; see #331 for the latter.

johnkerl:
Re-opening as
https://github.com/single-cell-data/TileDB-SOMA/blob/a7b33be1c7f70bfc6c01fb6e9d902a0ab383a977/apis/python/src/tiledbsoma/options/tiledb_create_options.py#L51-L96
reveals a great many functions without docstrings.

Morever, I should probably make this a keep-open-until-release issue.

johnkerl:
P.S. I am taking a look at
https://pypi.org/project/docstr-coverage/

johnkerl:
P.S. I am taking a look at
https://pypi.org/project/docstr-coverage/

bkmartinjr:
@johnkerl - regarding the above statement:
> Tweaking style (single vs double backticks, etc etc etc) which control how docstrings are rendered by various systems including readthedocs

Is there a specific list of requirements?  Just double-backtick, or other baseline requirements for a docstring?

johnkerl:
@bkmartinjr I had anecdotal observations & shared what I knew at the time -- I don't have encyclopedic knowledge to share.

bkmartinjr:
> I had anecdotal observations & shared what I knew at the time -- I don't have encyclopedic knowledge to share.

Oh, I was hoping we had a existing standard that you could point me at, rather than having us make a new one.  I've sent some options (pandas, numpy) via Slack. No reason to think those are right, but there needs to be a target for us to shoot at :-)

PR #635, opened 2023-01-03
[python] Add some docstrings

johnkerl:
Context: #602

### DIFF
--- a/apis/python/src/tiledbsoma/dense_nd_array.py
+++ b/apis/python/src/tiledbsoma/dense_nd_array.py
@@ -126,7 +126,7 @@ class DenseNDArray(TileDBArray):
 
     def reshape(self, shape: NTuple) -> None:
         """
-        Unsupported operation for this object type.
+        Unsupported operation for this object type [lifecycle: experimental].
         """
         raise NotImplementedError("reshape operation not implemented.")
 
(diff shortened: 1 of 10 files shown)


apis/python/src/tiledbsoma/dense_nd_array.py
```diff
@@ -125,6 +125,9 @@ def shape(self) -> NTuple:
             return cast(NTuple, A.schema.domain.shape)
 
     def reshape(self, shape: NTuple) -> None:
+        """
```
bkmartinjr, line 128: Optional/nit:  please add maturity tags whenever you update docstrings (even though `experimental` is the default, we should still eventually add them).

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -99,10 +99,16 @@ def __eq__(self, other: object) -> bool:
...
     def uri(self) -> str:
+        """
+        Accessor for the object's storage URI
+        """
         return self._uri
 
     @abstractproperty
     def soma_type(self) -> str:
+        """
+        Abstract accessor for `soma_type` string. To be implemented by child classes.
```
bkmartinjr, line 110: I don't think docstrings should have implementation details (e.g., 'to be implemented by ...').  They should focus on end-user (API user) information.

So, perhaps something like:  
> Return the SOMA object type, e.g., "SOMADataFrame"

bkmartinjr (approved):
one concrete suggest on `soma_type`.  The other comments are optional.

PR #637, opened 2023-01-03
DataFrame count accessor

johnkerl:
One of the first questions users ask about a single-cell dataset is "how many cells and how many genes are in here?"

We need to let people answer this question.

This is essential for uptake UX.

See also: https://github.com/single-cell-data/SOMA/pull/61


thetorpedodog (approved):

aaronwolen (approved):
Thanks!

apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -163,6 +163,23 @@ def get_index_column_names(self) -> Sequence[str]:
 
         return self._index_column_names
 
+    @property
```
bkmartinjr, line 166: Can you add __len__ per @thetorpedodog 's suggestion?

johnkerl: @bkmartinjr I didn't see the suggestion but it's a great one -- thanks!

johnkerl: #649

PR #639, opened 2023-01-03
[r] Fix storage of int64 values by SOMADataFrame

aaronwolen:
This updates `SOMADataFrame`'s write method to temporarily disable arrow's `int64->int32` downcasting behavior. 

Thanks to @eddelbuettel for the help tracking down and solving this issue.

### DIFF
--- a/apis/r/tests/testthat/test-SOMADataFrame.R
+++ b/apis/r/tests/testthat/test-SOMADataFrame.R
@@ -78,12 +78,14 @@ test_that("int64 values are stored correctly", {
   sidf$create(asch, index_column_names = "foo")
   tbl0 <- arrow::arrow_table(foo = 1L:10L, bar = 1L:10L, schema = asch)
 
+  orig_downcast_value <- getOption("arrow.int64_downcast")
+
   sidf$write(tbl0)
   tbl1 <- sidf$read()
   expect_true(tbl1$Equals(tbl0))
 
   # verify int64_downcast option was restored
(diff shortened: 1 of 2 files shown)


johnkerl (commented):
LGTM but I defer to @eddelbuettel for R expertise

apis/r/tests/testthat/test-SOMADataFrame.R
```diff
@@ -67,6 +67,25 @@ test_that("SOMADataFrame creation", {
...
+  sidf <- SOMADataFrame$new(uri)
+  sidf$create(asch, index_column_names = "foo")
+  tbl0 <- arrow::arrow_table(foo = 1L:10L, bar = 1L:10L, schema = asch)
+
+  sidf$write(tbl0)
+  tbl1 <- sidf$read()
+  expect_true(tbl1$Equals(tbl0))
+
+  # verify int64_downcast option was restored
+  expect_true(is.null(getOption("arrow.int64_downcast")))
```
eddelbuettel, line 86: What if it was set before? Not unlikely on CI, but you or I could have set it for your system. Then this test goes *boom*.

eddelbuettel (approved):
LGTM but the unit test may be overzealous.  _Maybe_ retrieve the value before writing and compare to that?

apis/r/tests/testthat/test-SOMADataFrame.R
aaronwolen: Good point. Just updated in 27ed9a04

PR #640, opened 2023-01-03
[python] Less-in-core ingest from H5AD [main]

johnkerl:
Context: Shortcut story 24294 and #641

See also #627 which is for the `main-old` branch.


apis/python/src/tiledbsoma/io.py
```diff
@@ -370,7 +370,6 @@ def from_anndata(
 
     anndata.obs_names_make_unique()
     anndata.var_names_make_unique()
-    anndata = util_ann._decategoricalize(anndata)
```
johnkerl, line 373: Since we got `anndata = ad.read_h5ad(h5ad_file_name, "r")` with the `"r"` this entire data structure is no longer reconstructible (below). So we split out the decategoricalizations to where they're needed.

aaronwolen (approved):
🚀

Issue #386, opened 2022-10-07
Refine soma_joinid/soma_rowid handling


bkmartinjr:
Affects both R & Python

Related to #324 and #325

Current thinking is that we want the handling of soma_joinid/soma_rowid to be explicit, eg,
* soma_joinid/int64 - (materialized column) explicit in read, get schema, writes, etc.  create op will enforce existence and type. No magic population of the column, user must explicitly specify values.
* soma_rowid/int64 - (pseudo-column only in SOMADataFrame), explicit in schema (create, get schema ops), explicit in read (always present unless user overrides with subset of column names), explicit in writes (it is the coordinate into the dense array)

- [x] Python implementation
- [x] R implementation

bkmartinjr:
@aaronwolen - any chance this is fixed?  It is done in Python, and I suspect it is done in R at this point.

bkmartinjr:
@aaronwolen @eddelbuettel  -  I believe the R work for this is complete.  Can you close it if you agree?  Python side has been complete for a long while.  Note that with the removal of the "dense" DataFrame, much of this issue is obsolete.

aaronwolen:
This isn't done but I'll be sure to get it added to our next sprint.

aaronwolen:
With #642 in the books I think we can close. @bkmartinjr?

bkmartinjr:
> With #642 in the books I think we can close. @bkmartinjr?

OK w/ me!  Thanks for doing the lift on this...

PR #642, opened 2023-01-03
[r] Correct handling of soma_joinid and soma_rowid

aaronwolen:
Addresses #386 for the R package.

Updates `SOMADataFrame`'s `create` method to: 

- automatically add `soma_joinid` if it's missing from the schema.
- prevents the creation of `SOMADataFrames` with `soma_` prefixed columns
- assert `soma_joinid` is of type `int64`

### DIFF
--- a/apis/r/tests/testthat/helper-test-data.R
+++ b/apis/r/tests/testthat/helper-test-data.R
@@ -42,7 +42,7 @@ create_and_populate_soma_dataframe <- function(uri) {
 
   tbl <- arrow::arrow_table(
     foo = 1L:10L,
-    soma_joinid = 1L:10L,
+    soma_joinid = 11L:20L,
     bar = 1.1:10.1,
     baz = letters[1:10],
     schema = arrow_schema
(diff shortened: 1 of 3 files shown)


apis/r/tests/testthat/helper-test-data.R
```diff
@@ -27,16 +27,22 @@ create_dense_matrix_with_int_dims <- function(nrows = 10, ncols = 5, seed = 1) {
...
   )
+}
+
+create_and_populate_soma_dataframe <- function(uri) {
+
+  arrow_schema <- create_arrow_schema()
 
   tbl <- arrow::arrow_table(
     foo = 1L:10L,
+    soma_joinid = 1L:10L,
```
johnkerl, line 45: kind of a nit but i like to make `soma_joinid` not startin at 0, just to demonstrate its flexibility as a sparse key, and to make sure we're not assuming anything about it being an array index per se anywhere inadvertently

johnkerl (approved):
🚢

eddelbuettel (approved):
:+1:

apis/r/tests/testthat/helper-test-data.R
aaronwolen: Good call. Changed in 70712ab0 .

aaronwolen:
Merging because R CI is passing

PR #620, opened 2022-12-19
Python 3.7 CI

johnkerl:
* Use `typing_extensions` in place of `typing`
* Don't use the walrus operator (`:=`)
* Other various findings on this PR

PR #646, opened 2023-01-04
Fix new smoke-test CI failure

johnkerl:


.github/workflows/py-packaging.yaml
```diff
@@ -91,6 +91,7 @@ jobs:
         unzip -l $whl
         pip install wheel
         pip install $whl
+        pip install typing_extensions
```
bkmartinjr, line 94: if this is a dependency, shouldn't it go into setup.cfg rather than special case it here?

It looks to me that we use it all over, but it is missing from the package dependencies...

bkmartinjr: As a side note, easy to test by doing a clean virtual environment install and see if it picks up everything needed....

bkmartinjr (changes requested):
Pretty sure that this belongs in the package dependencies, in setup.cfg, not in the workflow

.github/workflows/py-packaging.yaml
johnkerl: @bkmartinjr it's already here:
https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/python/setup.cfg#L39

Perhaps @mlin can help explain ...

bkmartinjr: lolz, you are right about it being present - I was looking at a slightly out of date branch!  Apologies.

However, this still feels wrong - the package should install without any additional deps being pre-installed, right?  What error do you see when this is missing?

bkmartinjr: Note that it is _missing_ from setup.py.  I don't understand why we replicate this stuff (@mlin can you comment on why package metadata is in both files?)

Perhaps synching these two will resolve the symptom.  But the redundancy should be resolved IMHO or we will make the same mistake again.

johnkerl: > What error do you see when this is missing?

Ever since I merged #620 , I've been getting GitHub notifications of fails in smoke-test with package-not-found error for `typing_extensions`. However ... just now I went to get a fresh link (one of many) to send you but (bafflingly) I can't find any such links now. 🥴

johnkerl: @bkmartinjr 
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/3839870890

These happen only on merges to main, not PR commits

bkmartinjr: Looking at the logs (including the "install wheel" section), it is clear that the package is just missing from the install (i.e., all other deps are installed, why not this one?).

My best theory is above - it is missing from `setup.py`:
> Note that it is missing from setup.py. I don't understand why we replicate this stuff (@mlin can you comment on why package metadata is in both files?)

Have you tried adding it there?

johnkerl: Yup, will do after lunch -- thx!!

johnkerl:
After latest rebase, now identical to #650.

PR #650, opened 2023-01-05
Fixing typing_extensions dependency

johnkerl:


bkmartinjr (approved):
this change looks good, but points out that we are missing `numpy`!

Would you mind adding it as well?

Issue #518, opened 2022-11-09
Support pa.int8/pa.uint8 indexing


johnkerl:
```
E   tiledb.cc.TileDBError: [TileDB::Dimension] Error: Tile extent check failed; Tile extent must not be 0
```

finish typing up

johnkerl:
Fixed by #653

PR #653, opened 2023-01-05
[python] Do not use extent 2048 on 8-bit index types

### LINKED ISSUE #652: [python] Unpin TileDB-Py 0.19.0
Reminder to revert https://github.com/single-cell-data/TileDB-SOMA/pull/651 after analysis

johnkerl:
Confluence of several issues today:

* Old bug described by #518, fixed on this PR -- was not fixed until just now
  * Root cause was the 2048 part fixed on this PR -- which mods to 0 of course for 8-bit types
  * Constructing a tiledb dim with extent 0 _should_ fail -- and was failing in the xfailed cases on this PR
* TileDB-Py 0.19.1 included a (completely innocent) refactor in terms of how `tiledb.Dimension` is constructed and handled
  * Giving rise to https://github.com/pytest-dev/pytest/issues/4659
* So in summary (debug thanks to @nguyenv) we had soma extent 0 -> tiledb-py constructor fail as designed-> segv in unit test due to pytest bug -> which is what we were seeing today with #651


apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -100,10 +100,19 @@ def _create_empty(
                 else:
                     raise TypeError(f"Unsupported dtype {dtype}")
 
+            extent = create_options.dim_tile(index_column_name)
+            if (
+                dtype == pa.int8()
+                or dtype == pa.uint8()
+                or dtype == np.int8
+                or dtype == np.uint8
+            ):
```
gsakkis, line 109: Minor drive-by question: can the two first conditions happen? `dtype` here is the result of `util_arrow.tiledb_type_from_arrow_type`, which IIUC can only be a `np.dtype` instance or the string "ascii" or "bytes". So this condition can be simplified to `if isinstance(dtype, np.dtype) and dtype.itemsize == 1:` 
(or perhaps more tersely `if getattr(dtype, "itemsize", 0) == 1:`).

mlin (commented):
@johnkerl 

1. Should we put a comment inline explaining the hard-coded `extent` value?
2. How do these issues affect the new PyPI package (and upcoming conda package), which I believe still require tiledb `>=0.19.0`?

johnkerl:
@mlin 

> Should we put a comment inline explaining the hard-coded extent value?

Great idea, thanks!

> How do these issues affect the new PyPI package (and upcoming conda package), which I believe still require tiledb >=0.19.0?

I just changed it to `>=0.19.0`, thanks!

apis/python/src/tiledbsoma/dataframe.py
johnkerl: Awesome, thanks @gsakkis !

gsakkis (approved):

apis/python/setup.cfg
```diff
@@ -36,7 +36,7 @@ install_requires =
     pyarrow
     scanpy
     scipy
-    tiledb==0.19.0
+    tiledb>=0.19.0
     typing_extensions
```
bkmartinjr, line 40: I believe the correct spelling is `typing-extensions`.

See https://pypi.org/project/typing-extensions/

yes, the package name and import name are different!

bkmartinjr (commented):
Apologies for the late review - would like to see the issues fixed (but another PR is fine).

apis/python/setup.cfg
johnkerl: @bkmartinjr #655

apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -100,10 +100,15 @@ def _create_empty(
                 else:
                     raise TypeError(f"Unsupported dtype {dtype}")
 
+            # Default 2048 mods to 0 for 8-bit types and 0 is an invalid extent
+            extent = create_options.dim_tile(index_column_name)
```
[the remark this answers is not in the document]
johnkerl: @bkmartinjr no because the two other callsites both have hard-coded `dtype` of `np.int64`:

* https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/python/src/tiledbsoma/sparse_nd_array.py#L82
* https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/python/src/tiledbsoma/dense_nd_array.py#L82

PR #657, opened 2023-01-06
[r] Fix creation of SOMADataFrames with non-int32 dimensions

aaronwolen:
Minor tweaks to `SOMADataFrame` to support creation of dimensions with int64, float/double, and string data types. As per the spec, we limit the domain of numeric dimensions to non-negative values.

### DIFF
--- a/apis/r/R/SOMADataFrame.R
+++ b/apis/r/R/SOMADataFrame.R
@@ -50,7 +50,8 @@ SOMADataFrame <- R6::R6Class(
 
         tdb_dims[[field_name]] <- tiledb::tiledb_dim(
           name = field_name,
-          domain = arrow_type_range(field$type),
+          # Numeric index types must be positive values for indexing
+          domain = arrow_type_unsigned_range(field$type),
           tile = tile_extent,
           type = tiledb_type_from_arrow_type(field$type),
           filter_list = tiledb::tiledb_filter_list(c(
--- a/apis/r/R/SOMADataFrame.R
+++ b/apis/r/R/SOMADataFrame.R
@@ -48,6 +48,11 @@ SOMADataFrame <- R6::R6Class(
           tile_extent
         )
 
+        # Default 2048 mods to 0 for 8-bit types and 0 is an invalid extent
+        if (field$type$bit_width %||% 0L == 8L) {
+          tile_extent <- 64L
+        }
+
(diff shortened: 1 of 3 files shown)


apis/r/R/SOMADataFrame.R
```diff
@@ -38,11 +38,21 @@ SOMADataFrame <- R6::R6Class(
       for (field_name in index_column_names) {
         field <- schema$GetFieldByName(field_name)
 
+        # TODO: Parameterize
+        tile_extent <- 2048L
```
johnkerl, line 42: recent tiledb-soma-py bug: if index type is int8/uint8 this mods to 0 😬 -- maybe use 64 in that case

johnkerl (approved):

eddelbuettel (approved):
Looks good

apis/r/R/utils-arrow.R
```diff
@@ -112,18 +112,35 @@ arrow_type_range <- function(x) {
     uint32 = bit64::as.integer64(c(0, 4294967295)),
     # We can't specify the full range of uint64 in R so we use the max of int64
     uint64 = c(bit64::as.integer64(0), bit64::lim.integer64()[2]),
-    float32 = c(-3.4028235e+38, 3.4028235e+38),
+    # float32/float
     float = c(-3.4028235e+38, 3.4028235e+38),
-    float64 = c(-1.7976931348623157e+308, 1.7976931348623157e+308),
+    # float64/double
     double =  c(-1.7976931348623157e+308, 1.7976931348623157e+308),
```
eddelbuettel, line 118: If you want to be fancy you could use R constants here:

```r
> c(.Machine$double.xmin, .Machine$double.xmax)
[1] 2.22507e-308 1.79769e+308
> 
```

aaronwolen: I _always_ want to be fancy 😄. Great suggestion!

apis/r/R/SOMADataFrame.R
aaronwolen: Good call, thanks!

apis/r/R/utils-arrow.R
aaronwolen: Fixed in d20770aa and tested in 45b1a7ad

Issue #308, opened 2022-09-19
Python API style - chaining


bkmartinjr:
In using the nascent tiledbsoma API, one pattern that I would find useful is if the `create` and related methods returned `self`, allowing chained code like the second line:

> an_arrow_tbl  = pa.Table.from_pandas(pd.DataFrame(...))
>  soma_df = soma.SOMADataFrame(uri).create(an_arrow_tbl.schema).write(an_arrow_table)

Methods that would benefit from this:
* create
* write
* write_from_pandas
* write_all_from_pandas
etc.

This deserves some discussion with Pythonic style experts, of which I am not one....

johnkerl:
Fluent syntax is awesome, and super-easy to code up! I think we should do it (my $0.02) ...

johnkerl:
cc @thetorpedodog

thetorpedodog:
I'm not personally the biggest fan of fluent APIs but I am not morally opposed to them (it's very much a personal thing). The thing I would want to make sure of is that the methods where you get a `self` back make sense. For instance, assuming `some_data_frame.write(...)` writes something *into* the dataframe, getting the dataframe back would certainly make sense, but a method that was like `something_else.write(some_sink)` (where `something_else` is being written to `some_sink`) could sensibly return some information about the write result and might not be as good a candidate for chaining. Based on my (still fairly new) understanding, all the methods you've listed seem like they are reasonable methods to chain.

If only we could jump to Python 3.11 which has the [`Self` annotation](https://docs.python.org/3.11/library/typing.html#typing.Self)!

bkmartinjr:
> make sense

fully agree. There are many methods where it does _not_ make sense (eg, delete).  I am supportive of doing this selectively and iteratively.

bkmartinjr:
This change has been applied to `create()` with good effect.  We should decide if we want to retain that, and if other functions would benefit.  Should be resolved alongside single-cell-data/SOMA#41

johnkerl:
@bkmartinjr @thetorpedodog anything else to cover here, given #663? Otherwise I could close this task.

thetorpedodog:
I don’t think there’s anything else that needs to be done on this at the moment.

johnkerl:
OK -- @bkmartinjr (or anyone else) -- if you think of something we can re-open this.

PR #663, opened 2023-01-06
[python] Fluent API changes for read methods

### LINKED ISSUE #596: Feature request:  SparseNdArray "read all" variant returning a Arrow sparse tensor
Feature request: it would be very useful to have a "read all" variant of the "read_sparse_tensor" method, ie., that performs the tensor concatenation and returns a single sparse tensor (rather than an iterator of sparse tensors)

### LINKED ISSUE #656: [python] Rename read methods
Split-out issue to track discussion here:

https://github.com/single-cell-data/SOMA/pull/56#discussion_r1049885231

bkmartinjr:
Adopt the API changes proposed by `somacore` for `read()`, and partial changes for `write`.

Where we used to do stuff like: `sparse_array.read_tensor(format='csc')`, we now say `sparse_array.read().coo()`.  Types are defined by [somacore](https://github.com/single-cell-data/SOMA/tree/main/python-spec).

*NOTES*:
* this PR _only_ makes full set of changes to the `read()` API.  A subset of changes to `write()` (method naming, not params).
* a copy of a subset of `somacore` is included.  to be removed when the full package can be installed as a dependency
* there is a slight behavior change to SparseNDArray, when calling `read()`:   when the return value is empty, *and* `coos()` are requested, the old behavior would return None (rather than the expected empty SparseCOOMatrix).  The new behavior removes the workaround for ARROW-17933 (see #668), which means that this narrow case will receive an ArrowInvalid exception until that bug is fixed. I did this as there is no real work around (returning None is no better given that the use expects an Arrow object).

There is still work to be done to align the classes with the ABCs being defined by @thetorpedodog 

Fixes #596
Fixes #656 

Partially addresses #308


apis/python/src/tiledbsoma/util_iter.py
```diff
@@ -0,0 +1,119 @@
...
+    def concat(self) -> pa.Table:
+        """Concatenate remainder of iterator, and return as a single Arrow Table"""
+        return pa.concat_tables(self)
+
+    def close(self) -> None:
+        """TODO: Implement when there is a stateful handle that needs to be closed"""
+        pass
+
+
+RT = TypeVar("RT")
```
johnkerl, line 37: If `RT` is intended to abbreviate `ReadType` or `ReturnType`, IMHO we could comfortably spell out either of those instead of abbreviating

apis/python/src/tiledbsoma/util_iter.py
```diff
@@ -0,0 +1,119 @@
...
+        if arrow_table is None:
+            raise StopIteration
+
+        return arrow_table
+
+    def concat(self) -> pa.Table:
+        """Concatenate remainder of iterator, and return as a single Arrow Table"""
+        return pa.concat_tables(self)
+
+    def close(self) -> None:
```
johnkerl, line 32: I imagine you have good reason for including a stub `close` but not a stub `open` -- I don't understand though why we have one and not the other

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -154,260 +159,150 @@ def nnz(self) -> int:
...
 
-    def read_sparse_tensor(
+    def read(
         self,
-        coords: SparseNdCoordinates,
-        *,
-        format: Literal["coo", "csr", "csc"] = "coo",
-    ) -> Iterator[Union[pa.SparseCOOTensor, pa.SparseCSCMatrix, pa.SparseCSRMatrix]]:
-        """
-        Read a use-defined slice of the SparseNDArray and return as an Arrow sparse tensor.
```
johnkerl, line 164: It appears I've missed it -- I'm not seeing where all this great docstring content went to ... was it somehow omitted from this PR, or did it go to somewhere I missed seeing it?

johnkerl (changes requested):

apis/python/src/tiledbsoma/sparse_nd_array.py
bkmartinjr: A brain fault - will fix.

apis/python/src/tiledbsoma/util_iter.py
bkmartinjr: It is a context manager cleanup protocol, not an array/reader close - see somabase.  There will never be an open in this class - when you get one of these classes, it is a handle to an already open object.

apis/python/src/tiledbsoma/util_iter.py
bkmartinjr: Typical python style for generics uses short variables.  See @thetorpedodog code example in [somacore](https://github.com/single-cell-data/SOMA/blob/257db861deca17eca1bb0637951dedec1e025e1c/python-spec/src/somacore/base.py#L50) for example.

Or another example [here](https://github.com/single-cell-data/SOMA/blob/257db861deca17eca1bb0637951dedec1e025e1c/python-spec/src/somacore/data.py#L200)

bkmartinjr: BTW, if any consolation - I am unsure _why_ this is the style.  I had exactly the same reaction as you when I first saw it!

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -154,96 +159,23 @@ def nnz(self) -> int:
...
-        *,
-        format: Literal["coo", "csr", "csc"] = "coo",
-    ) -> Iterator[Union[pa.SparseCOOTensor, pa.SparseCSCMatrix, pa.SparseCSRMatrix]]:
+        coords: Optional[SparseNdCoordinates] = None,
+        # *,
+        # TODO: missing parameters
+    ) -> "SparseNDArrayRead":
         """
-        Read a use-defined slice of the SparseNDArray and return as an Arrow sparse tensor.
+        Read a use-defined slice of the SparseNDArray.
```
johnkerl, line 169: nit: user-defined

bkmartinjr: fixed

johnkerl (approved):
🚢

thetorpedodog (approved):

PR #664, opened 2023-01-06
[python] Resume mode [main]

johnkerl:
Ports #628 from `main-old` to `main`. This implements important user feedback for `tiledbsoma.io.from_h5ad` and `tiledbsoma.io.from_anndata`.

When an ingest is interrupted, for any reason whatsoever -- network event for remote writes, disk-full for local writes, OOM, control-C, etc etc -- it's crucial that we give users an option to not rewrite data that has already been written. When redundant timestamp-layers are written, subsequent query performance is negatively impacted -- resulting in a doubly negative user experience, both at ingest time and at query time.

This PR addresses this key UX and performance issue.


apis/python/tests/test_dataframe.py
```diff
@@ -552,7 +552,7 @@ def make_multiply_indexed_dataframe(tmp_path, index_column_names: List[str]):
             "index_column_names": ["index2", "index3"],
             "ids": [[True], slice(None)],
             "A": None,
-            "throws": RuntimeError,
+            "throws": (RuntimeError, tiledb.cc.TileDBError),
```
johnkerl, line 555: Fails one way in sandbox, another way in CI

aaronwolen (approved):
Awesome work!

johnkerl:
Thanks @aaronwolen !!!! :)

PR #666, opened 2023-01-07
[python] Improve resume-mode/schema-only testing [main-old]

johnkerl:
The unit-test code on this PR was, erroneously, using a new tempdir on each rewrite pass -- not actually testing rewrite of the same soma directory. This PR fixes that oversight.


nguyenv (approved):

johnkerl:
Thanks @nguyenv ! :)

Issue #603, opened 2022-12-12
[python] Initial stateful Python implementation


johnkerl:
Make the API stateful, so it can represent an explicit point-in-time handle to a durable collection and/or object.

#540 tracks the full gamut of statefulness:

* Right now: open all arrays with time.now() timestamp (implcitly)
* Sooner:
  * `experiment.open()` with implicit `time.now()` put into some variable/config/something
  * All component opens will use that timestamp
* Easy to do after:
  * `experiment.open(specified_timestamp)`
  * All component opens will use that timestamp
* More after that:
  * Grouped commit / fragment coalesce / rollback ideas 

This issue is specifically about moving from the first bullet (the status quo) to the second bullet.
 
Follow-on from https://github.com/single-cell-data/SOMA/issues/44

maniarathi:
@mlin @thetorpedodog Is this complete? Can we close this issue?

thetorpedodog:
Going over why we marked this done:

We're currently passing around the `context` object every time we open (and soon, when we create) a SOMA object from a parent. This is where we are sticking the timestamp to support consistency.

Further work includes finishing out the `add_new_x` methods (https://github.com/single-cell-data/TileDB-SOMA/issues/840), and making it so that when we do open child objects their ownership is tracked for closing them all (https://github.com/single-cell-data/TileDB-SOMA/issues/843).

thetorpedodog:
And @mlin is working on timestamp setting in https://github.com/single-cell-data/TileDB-SOMA/pull/819, which I stomped all over in https://github.com/single-cell-data/TileDB-SOMA/commit/93050f1ca76700b6038033befe953b6f6085bee0.

PR #671, opened 2023-01-08
[python] Stateful (open/close) objects -- abstract first pass

mlin:
Context: #540

### DIFF
--- a/apis/python/src/tiledbsoma/tiledb_object.py
+++ b/apis/python/src/tiledbsoma/tiledb_object.py
@@ -1,6 +1,8 @@
 import os
+import time
 from abc import ABC, abstractmethod, abstractproperty
-from typing import Optional, Union, cast
+from types import TracebackType
+from typing import ContextManager, Optional, Type, TypeVar, Union, cast
 
 import tiledb
 
@@ -8,8 +10,11 @@ from . import util
 from .metadata_mapping import MetadataMapping
 from .tiledb_platform_config import TileDBPlatformConfig
 
+# type variable for methods returning self; see PEP 673
+TTileDBObject = TypeVar("TTileDBObject", bound="TileDBObject")
 
-class TileDBObject(ABC):
+
+class TileDBObject(ContextManager["TileDBObject"], ABC):
     """
     Base class for ``TileDBArray`` and ``Collection``.
 
@@ -19,6 +24,7 @@ class TileDBObject(ABC):
     _uri: str
     _tiledb_platform_config: TileDBPlatformConfig
     _metadata: MetadataMapping
+    _parent_timestamp: Optional[int] = None
 
     def __init__(
         self,
@@ -32,7 +38,10 @@ class TileDBObject(ABC):
         ctx: Optional[tiledb.Ctx] = None,
     ):
         """
-        Initialization-handling shared between ``TileDBArray`` and ``Collection``.  Specify ``tiledb_platform_config`` and ``ctx`` for the top-level object; omit them and specify parent for non-top-level objects. Note that the parent reference is solely for propagating options, ctx, display depth, etc.
+        Initialization-handling shared between ``TileDBArray`` and ``Collection``.  Specify
+        ``tiledb_platform_config`` and ``ctx`` for the top-level object; omit them and specify
+        parent for non-top-level objects. Note that the parent reference is solely for propagating
+        options, ctx, display depth, etc.
         """
 
         if ctx is None:
@@ -43,6 +52,7 @@ class TileDBObject(ABC):
         else:
             tiledb_platform_config = parent._tiledb_platform_config
             self._ctx = parent._ctx
+            self._parent_timestamp = parent._effective_timestamp()
 
         self._tiledb_platform_config = tiledb_platform_config or TileDBPlatformConfig()
         # Null ctx is OK if that's what they wanted (e.g. not doing any TileDB-Cloud ops).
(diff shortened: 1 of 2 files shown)


mlin:
@johnkerl @thetorpedodog @bkmartinjr @atolopko-czi 

This is just an early stub to codify https://github.com/single-cell-data/SOMA/issues/44#issuecomment-1332780970 and ensuring discussion. Please comment whether it seems to be heading in the right direction. A few discussion questions for example:

1. Is `TileDBObject` the right place to define this -- such that all the collection and array classes would inherit it -- or should we stick to individual arrays for 1.0?
2. If so, then it seems like we want to say that once we open a collection with a given timestamp, we're going to propagate that timestamp into any other TileDBObjects we access from the open collection. True? (Need to think about when accessed objects are themselves to be "opened" in this case.)
3. I wrote some strong language about how the "write" mode works in the docstrings, which I'm not sure we can actually achieve, even for an individual array. Please review and comment; are `write_*` calls on TileDBArrays 1:1 with creating a TileDB fragment, and if so, are we able to group-commit/rollback multiple fragments? 

(tickets #540 #603)

johnkerl:
Re 1: `TileDBObject` may work -- it's a base for groups and arrays both. Another option is `TileDBArray` and `Collection` -- may be necessary since the former are opened with `tiledb.Group` and the latter with `tiledb.open`

Re 2 and 3, I'd reiterate from https://github.com/single-cell-data/TileDB-SOMA/issues/603:

> #540 tracks the full gamut of statefulness:
> 
> * Right now: open all arrays with time.now() timestamp (implcitly)
> * Sooner:
>   * `experiment.open()` with implicit `time.now()` put into some variable/config/something
>   * All component opens will use that timestamp
> * Easy to do after:
>   * `experiment.open(specified_timestamp)`
>   * All component opens will use that timestamp

For 1.0 I don't see any reason to document stronger language than we can pull off. IMO, for this initial PR, we should do the minimal syntax to open the experiment at the current timestamps and ensure that is passed on to opens of member groups and arrays. This is the first PR on this subject; it needn't be the last, and it needn't do everything.

mlin:
@johnkerl Great input, thank you.

It does seem like it's going to be useful to push a few core bits all the way down into `TileDBObject`:

1. Storing the timestamp and current open mode (if any)
2. Validating those arguments to `open()`
3. Context manager sugar

But, certainly further specialization in `TileDBArray` and `CollectionBase` as well as the concrete classes will be needed...the changes won't be "confined" to `TileDBObject`, but there's just enough of a shared core to factor there.

For the write mode, thanks and that aligns with my limited understanding. Specifically addressing https://github.com/single-cell-data/SOMA/issues/44#issuecomment-1332865182 and following comments, it seems like we say the write mode will make multiple writes share the same timestamp, but we can't (in the 1.0 timeframe) claim atomic commit of multiple writes -- even to a single object. Fair @bkmartinjr @thetorpedodog?

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -125,6 +136,80 @@ def exists(self) -> bool:
         except tiledb.cc.TileDBError:
             return False
 
+    _open_mode: Optional[str] = None
+    _open_timestamp: Optional[int] = None
+
+    def open(
```
Shelnutt2, line 142: Timestamps should be a tuple/list for `[start, end]`.

apis/python/tests/test_collection.py
```diff
@@ -249,3 +250,38 @@ def test_exceptions_on_not_created(tmp_path):
...
+    create_and_populate_dataframe(dataframe)
+    sparse_nd_array = soma.SparseNDArray(
+        os.path.join(basedir, "snda"), parent=collection
+    )
+    create_and_populate_sparse_nd_array(sparse_nd_array)
+
+    collection.set("sdf", dataframe)
+    collection.set("snda", sparse_nd_array)
+
+    time.sleep(0.01)
```
Shelnutt2, line 274: I'd recommend against sleeping and instead use write + read timestamps to guarantee ordering.

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -18,6 +23,7 @@ class TileDBObject(ABC):
 
     _uri: str
     _tiledb_platform_config: TileDBPlatformConfig
+    _parent_timestamp: Optional[int] = None
```
Shelnutt2, line 26: Timestamps should be a tuple/list for [start, end].

bkmartinjr:
@Shelnutt2 - is there a way we can anticipate what a multi-object commit would look like in this API?  Or is that too ambitious (and we can safely ignore for now)?   I'm thinking about multiple updates to an Experiment/Measurement/Collection that want to be treated as a snapshot/commit.

mlin:
@thetorpedodog @bkmartinjr another question arising in discussion with @johnkerl -- the current direction makes using open/close suggested but not required; if you don't open then you can still do read/write operations without any snapshot guarantees between them. If this is a good direction, then this isn't a breaking change as drafted. But please comment whether we want to enforce the pattern, thanks

bkmartinjr:
@mlin - I touched base with @Shelnutt2 offline, and am satisfied that this PR will handle future multi-object consistency.  I misspoke above when I said `multi-object commit`, which we don't actually need IMHO.  I am simple looking for a way to publish a self-consistent set of soma.Experiment/Measurement changes, i.e., all are visible or none are visible.  The group-level timestamp support should satisfy that down the road.

> open/close suggested but not required

I don't feel strongly, but my gut instinct is to make it required so that consistent behavior is the *norm*, not the exception.

apis/python/src/tiledbsoma/tiledb_object.py
mlin: @thetorpedodog @bkmartinjr @johnkerl @Shelnutt2 Are we ok with using timestamp *ranges* at the SOMA API level? Thought I'd double-check as the concept didn't come up in https://github.com/single-cell-data/SOMA/issues/44#issuecomment-1332780970 and strikes me as quite specific to advanced TileDB capabilities. 
Should we say the timestamp type is `Any` ("token") at the somacore level and `Tuple[int,int]` in the TileDB implementation? A simpler approach might be to take `timestamp: int` and take that to mean `(0, t)` when we call into TileDB. But perhaps `int` isn't general enough at the somacore level, like the example of using a commit digest?

mlin: Another idea from @johnkerl, take *two* arguments with defaults `start=0` and `end=[now]`.

mlin:
@thetorpedodog @johnkerl @Shelnutt2  @bkmartinjr @atolopko-czi FYI I'm working on implementation on another branch while keeping this one largely limited to the interface changes, to hopefully focus discussion (for the Friday meeting but hopefully beforehand too!).

Trying to summarize the threads so far as well as what I've been stewing on:

1. I think we have the right mechanism for the key ability of an array/sub-collection to inherit a read timestamp from its parent, to ensure consistent reads across all those objects (and of course, multiple reads from any one object).
2. Unresolved- is it *required* to open an object before reading from it?
    - Merely accessing an element of an open collection is sufficient for the accessed element to inherit the read timestamp
    - But opening the object gets you the potential reuse of handles/metadata for multiple reads of the object
3. If I close a collection but still hold references to accessed elements -- perhaps open themselves -- are they affected in any way?
     * and if I then reopen the collection at a different timestamp or mode?
4. If we're `open('w',t)`, do we reject attempts to read, and if not then what timestamp would we read from?
5. Trying to inherit open write timestamps in a similar way as read timestamps seems to me quite fraught, as it multiplies the above corner cases.
    - It may be better to always specify the desired write timestamp to each individual object one is writing.
    - Considering it's possible to write an object fully before even adding it to a parent collection, I think?
7. What will be the type of a timestamp in somacore?
    * Bears on whether/how we make the full power of TileDB time travel available through TileDB-SOMA

johnkerl:
> doesn't that suggest different timestamp datatypes for reading vs. writing?

@mlin yes -- for read there is `(start, end)` (where `start` will almost always default to 0 epoch) but for write there is just `t`

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -133,6 +143,80 @@ def exists(self) -> bool:
...
+            return self._open_timestamp
+        return self._parent_timestamp
+
+    def close(self) -> None:
+        """
+        Release any resources held while the object is open.
+        """
+        if not self.is_open():
+            raise RuntimeError("TileDBObject is not open")
+        self._open_mode = None
```
atolopko-czi, line 206: Should this be cascaded to children?

atolopko-czi: Also, should the underlying tiledb object be closed as well at this point?

atolopko-czi: To avoid some of the edge cases discussed, e.g., client code holding onto TileDbObject references and using them outside of the open() call, I'm wondering if TileDBObjects should raise an error on future reads if the object has been explicitly closed.

thetorpedodog: I am *moderately* in favor of cascaded close operations. My thought there is that given that you’ll mostly be implicitly opening children by accessing them from a parent, closing that parent should also close the children. Provided that we make `close` idempotent, a user closing a child and then closing a parent later would not be a problem.

I am also strongly in favor of saying “once you `close`, you can’t reopen”, because that seems like the general pattern in e.g. built-in `open`.

mlin:
Closing this "abstract first pass" as implementation advances in #730

johnkerl:
:v:

PR #672, opened 2023-01-08
[python] Import somacore [WIP / not ready for review / please do not review]

johnkerl:
Purpose: Validates that the Python TileDB-SOMA implementation is following the SOMA spec at the class/function/signature level.

Context: #669 

Status:

* WIP;  not ready for review.
* Currently branched off `bkmartinjr/fluent-api-changes` which itself is out of data with `main` (as of 2023-01-08). Rebasing will be needed here.
* Needs `somacore` pip-installable
* Depends on SOMA PR https://github.com/single-cell-data/SOMA/pull/67

Issue #669, opened 2023-01-07
[python] Import and use `somacore`


johnkerl:
When `somacore` lands as a package (see for example https://github.com/single-cell-data/SOMA/pull/65), we'll need to start using it in this package: import the ABCs and make sure we follow their signatures and types.

johnkerl:
Status / plan of action:

* This will be multi-repo and multi-PR
* As is developing in Slack and at the following we'll need a per-python-version `*-lint-requirements.txt` _in addition to and separately maintained from_ `setup.cfg` within this repo
  * https://github.com/single-cell-data/SOMA/pull/71
  * https://github.com/single-cell-data/SOMA/pull/72
  * https://github.com/single-cell-data/SOMA/pull/73
  * https://github.com/single-cell-data/TileDB-SOMA/pull/683
  * https://github.com/single-cell-data/TileDB-SOMA/pull/678
  * https://github.com/single-cell-data/TileDB-SOMA/pull/685
* We'll need to pip-install the lint-requirements packages before running `mypy`
* Even then we get things like `foo: np.ndarray` getting `Missing type parameters for generic type "ndarray"`
  * Fix these via `foo: np.ndarray[Any, Any]` and then we get `TypeError: Type subscription requires python >= 3.9` for Python 3.7 and 3.8 CI
  * Which will in turn require mypy-suppress comments line by line at those callsites if we're to enforce PR-blocking typechecks between `somacore` and `tiledbsoma` (which is the entire purpose of having `somacore`
* Also to address: function signatures such as
  * https://github.com/single-cell-data/SOMA/blob/a7b623eb61418b8bfa8a66c13823c15b358468fe/python-spec/src/somacore/data.py#L25-L35
  * vs
  * https://github.com/single-cell-data/TileDB-SOMA/blob/8a0a356721a64cf0875f6dd03652d93fa68eeb0d/apis/python/src/tiledbsoma/dataframe.py#L194-L202
  * combined with package flux due to #644 and #603
 
Just to set a realistic time estimate, I don't see this issue closing for another week

cc @bkmartinjr @thetorpedodog @mlin @atolopko-czi

PR #673, opened 2023-01-09
[python] Prep for import somacore

johnkerl:
PR #672 (see also issue #669) is not ready to merge, as noted there -- but what we can do now is clear up some of the necessary changes that don't specifically involve doing `import somacore`. Namely:

* Define `obj.metadata` in a way that lets static analysis realize it's a property not a function (there was an FP until I replaced `obj.metadata` attribute with `obj._metadata` and then also made a `def metadata()` with `@property`)
* `get_index_column_names()` in this package is spelt `index_column_names()` in `somacore`

With this PR merged first, #672 will be much simpler and clearer.


apis/python/src/tiledbsoma/io.py
```diff
@@ -519,7 +519,7 @@ def to_anndata(
     s = util.get_start_stamp()
     logging.log_io(None, "START  Experiment.to_anndata")
 
-    measurement = experiment.ms[measurement_name]
+    measurement: Measurement = experiment.ms[measurement_name]
```
thetorpedodog, line 522: fyi: in the future this should not be necessary since `Experiment.ms` will be defined to return a `Collection[Measurement]`.

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -67,6 +67,10 @@ def _default_ctx(self) -> tiledb.Ctx:
 
         return tiledb.Ctx(cfg)
 
+    def metadata(self) -> MetadataMapping:
```
thetorpedodog, line 70: `metadata` is defined in `somacore` as a `@property`, not a function.

johnkerl: @thetorpedodog I cannot explain then why I had to change _from property to function_ to make errors go away.

Maybe not everything is at `HEAD` in both repos -- ???

johnkerl: @thetorpedodog addressed on commit
https://github.com/single-cell-data/TileDB-SOMA/pull/673/commits

thetorpedodog: I think this *might* be the issue: if an abstract `@property` is defined, it needs to be defined as a property on the implementing class:

```python
class Base(metaclass=abc.ABCMeta):
  @property
  @abc.abstractmethod
  def it(self): pass

class ThisWillWork(Base):
  @property
  def it(self): return 50

class ThisFails(Base):
  def __init__(self):
    self.it = 50
```

thetorpedodog (approved):

johnkerl:
> I think this might be the issue: if an abstract @property is defined, it needs to be defined as a property on the implementing class:

@thetorpedodog you discovered what I did :) -- yay! :) 

I did precisely this on the most recent commit

Issue #675, opened 2023-01-09
[python] Fix wheels CI on merge to main


johnkerl:
Tracks https://github.com/single-cell-data/TileDB-SOMA/pull/674#issuecomment-1375631332

PR #677, opened 2023-01-09
Fix setup.py to play well with cibuildwheel

gsakkis:
Fixes https://github.com/single-cell-data/TileDB-SOMA/issues/675:
- Remove `if __name__ == "__main__"`; it [confuses cibuildwheel](https://github.com/pypa/cibuildwheel/blob/7c4bbf8cb31d856a0fe547faf8edf165cd48ce74/cibuildwheel/projectfiles.py#L41-L46)
- Add back `ext_modules` option; it's needed (even if `sources` are empty) otherwise the wheel is considered pure-Python.

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -15,15 +15,15 @@
 
 # type: ignore
 
+import pathlib
 import shutil
 import subprocess
 import sys
-from pathlib import Path
 
+import setuptools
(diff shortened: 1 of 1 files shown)


johnkerl (approved):
🚀 

Thanks @gsakkis !!!!! :D

PR #680, opened 2023-01-09
Remove duplicate `test_general_utilities.py` file.

thetorpedodog:
This somehow found its way into the main source when it should be (and already is!) in `tests/`.


johnkerl (approved):
🪓

Issue #644, opened 2023-01-04
Add explicit context to the API


johnkerl:
Adding an explicit context to the API, which is used to represent session state and session configuration. Requires integration of existing `tiledb.Ctx` and `platform_config`.

PR #681, opened 2023-01-09
[python] Add SomaSessionContext to all SOMA objects

atolopko-czi:
For #644 

* SomaSessionContext class added and taken by all SOMA class constructors.
* SomaSessionContext includes a tiledb.Ctx, so replaced "ctx" params with SomaSessionContext
* Moved the `member_uris_are_relative` PlatformConfig to SomaSessionContext
* PlatformConfig is replaced by just CreateOptions for now. CreateOptions is now passed through all "creation" calls, and so isn't maintained as an instance variable on TileDBObjects, as was PlatformConfig. Anything needed by a TileDBObject from PlatformConfig that was *not* creation-related can now just become an attribute of SomaSessionContext.


apis/python/src/tiledbsoma/factory.py
```diff
@@ -73,22 +74,22 @@ def _construct_member(
     # Now invoke the appropriate per-class constructor.
     if class_name == "Experiment":
         assert object_type is None or object_type == "group"
-        return Experiment(uri=member_uri, parent=parent, ctx=ctx)
+        return Experiment(uri=member_uri, parent=parent, session_context=session_context)
```
mlin, line 77: @atolopko-czi I made some minor revisions to this big switch in #671 although those changes ended up not being necessary in the ~final approach of that PR, and maybe belong here instead.

The [TileDBObject constructor](https://github.com/single-cell-data/TileDB-SOMA/blob/3ad6cf7cc6e3939851bb77c8ab30c6bf8dcef56d/apis/python/src/tiledbsoma/tiledb_object.py#L28-L32) says to pass in *at most one of* parent or context, because if you supply the parent then the new object inherits the parent context. As currently coded on main, if you pass in both (as we're doing here), the constructor *ignores* the explicitly-provided context in favor of inheriting the parent context. This is pretty confusing so unless there's a case where we need the new object to have a different context than its parent, I suggest adding an assert in the TileDBObject constructor that parent and context are mutually exclusive, and not passing in the context (separately from the parent) here.

mlin: Just clarifying "unless there's a case" -- I don't know of such a case, but I'm pretty new at this too, so it merits investigation. I do know in #671 that all the CI passes with such an assert.

johnkerl: @mlin there's nothing profound here -- I can say this as I wrote the original version :) -- the idea was at each level the normal behavior is take context from parent. With option to have a 'special' per-object context if desired.

Also note what _is_ the parent object -- nominally either a `soma.Experiment`, or a `soma.Collection` consisting of `soma.Experiment` objects. So for `soma.Experiment` in particular it was important to get the context from the parent -- if there is one -- or for it to get its own context since it is the top-level object.

How we handle the case "what if parent and context are both specified" -- yes currently it takes a non-null context as override of the parent. And I like your idea of asserting one XOR the other. (Which is perhaps the way it should have been done all along.)

atolopko-czi: Thanks, both! I'll pass just `parent`, to align with #671. And I'll add the param XOR validation as well.

mlin: Terrific, thanks @atolopko-czi. I've backed those changes out of #671 which also helps clarify/simplify what's going on there (since they ended up being extraneous).

apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -51,16 +53,19 @@ def create(
         self,
         schema: pa.Schema,
         index_column_names: Sequence[str] = (SOMA_JOINID,),
-        platform_config: Optional[PlatformConfig] = None,
+        platform_config: Optional[TileDBCreateOptions] = None,
```
thetorpedodog, line 56: The `platform_config` parameter is designed to be extensible for future operations and (as of right now) we want to be able to accept the same `PlatformConfig` value across multiple calls (e.g. a user could pass the same `platform_config` dict to create and query etc.). Right now it’s only *used* on `create` but we’ll be adding more integration points (cf. the fact that the [format description](https://github.com/single-cell-data/TileDB-SOMA/tree/main/apis/python#platform_config-format) has only the `tiledb.create` field available.

*Internally* we can transform it into something more compact (which was the goal with the way I implemented the thing which wraps dictionaries, though it is admittedly somewhat inelegant).

apis/python/src/tiledbsoma/tiledb_create_options.py
```diff
@@ -14,85 +14,50 @@
...
-            "ZstdFilter",
-        ),
+        default: Sequence[StrOrMap] = DEFAULT_FILTERS,
     ) -> Sequence[tiledb.Filter]:
         return _build_filters(self.get("offsets_filters", default))
 
-    def cell_tile_orders(
-        self,
-        default_cell: Optional[str] = "row-major",
-        default_tile: Optional[str] = "row-major",
```
thetorpedodog, line 94: The reason I did this here is in case we wanted to use different default cell/tile reading orders in different places. It is entirely possible that we don’t, and removing this is honestly probably fine given that—it can always be added back later.

apis/python/src/tiledbsoma/tiledb_session_context.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # Temp workaround pending https://app.shortcut.com/tiledb-inc/story/23827
+    region = os.getenv("AWS_DEFAULT_REGION")
+    if region is not None:
+        cfg["vfs.s3.region"] = cast(str, region)  # type: ignore
+
+    return tiledb.Ctx(cfg)
+
+
+@dataclass(frozen=True)
```
thetorpedodog, line 28: optional: in a previous change I suggested the use of `attrs` instead of `dataclasses` (too lazy to dig it up right now). In this case it would be simply switching from `@dataclass` to `@attrs.define` here (after adding the import). See also [the general case for attrs over dataclasses](https://www.attrs.org/en/stable/why.html#data-classes).

(Other than that, using a dataclass/attrs class here is perfect.)

apis/python/src/tiledbsoma/tiledb_session_context.py
```diff
@@ -0,0 +1,39 @@
...
+class TileDBSessionContext:
+    """
+    Maintains context across an entire SOMA session that can be shared across multiple SOMA objects.
+    """
+
+    tiledb_ctx: tiledb.Ctx = build_default_tiledb_ctx()
+
+    # Allows "relocatability" for local disk / S3, and correct behavior for TileDB Cloud
+    member_uris_are_relative: Optional[bool] = None
```
thetorpedodog, line 37: If you add the comment as a docstring-like string below the member, some tooling (e.g. intellisense) will pick it up as if it were one:

```suggestion
    member_uris_are_relative: Optional[bool] = None
    """Allows "relocatability" for local disk / S3, and correct behavior for TileDB Cloud."""

**Review by thetorpedodog — COMMENTED** [review]
Liking the way `SessionContext` is working, but I think the changes to `platform_config` might not be what the API is looking for in general. (More details below.) It seems like those two things are mostly different concerns here, so it should be possible to have the one without the other.

Apart from that I think the direction is good, as we discussed earlier today.

File: apis/python/src/tiledbsoma/dataframe.py (hunk 2 above)
Comment (atolopko-czi) [reply apis/python/src/tiledbsoma/dataframe.py:56]: Since the somacore spec is just asking for a `Mapping`-typed param, would it be wrong to introduce a new `TileDbQueryOptions(soma.PlatformConfig)` class for query methods? Consider that both create() and query() methods (et. al.) would still adhere to the the somaspec (taking a soma.PlatformConfig-derived type), even though in this TileDB platform implementation, we'd be using separate classes for these individual cases.

File: apis/python/src/tiledbsoma/collection.py [hunk 6]
```diff
@@ -84,19 +84,13 @@ def __init__(
         *,
         # Non-top-level objects can have a parent to propagate context, depth, etc.
         parent: Optional[CollectionBase[Any]] = None,
-        # Top-level objects should specify these:
-        tiledb_platform_config: Optional[TileDBPlatformConfig] = None,
-        ctx: Optional[tiledb.Ctx] = None,
+        # Top-level objects should specify this:
+        session_context: Optional[TileDBSessionContext] = None,
```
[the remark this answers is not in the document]
atolopko-czi: Makes sense, will add this!

apis/python/src/tiledbsoma/tiledb_create_options.py
atolopko-czi: As every `create()` call takes a `TileDbCreateOptions` arg, wouldn't every `TileDbArray` be able to have its own cell/tile order specified?  As shown in the [test](https://github.com/single-cell-data/TileDB-SOMA/blob/cce50edf05f37da122564d92cc08b9dd66baec64/apis/python/tests/test_create_options.py#L47), specifying `cell_order` and `row_order` is still supported.

apis/python/src/tiledbsoma/tiledb_session_context.py
atolopko-czi: I do remember seeing some discussion of that!  :)  Will change it to `attrs` for consistency.

apis/python/src/tiledbsoma/dataframe.py
atolopko-czi: > a user could pass the same platform_config dict to create and query 
So there could _per-call_ TileDB options that are not create- or query-specific? I was imaging that any options that are not create- or query-specific would be specified as part of the SessionContext. But of course, that implies session-long settings, not call-specific setting.

apis/python/src/tiledbsoma/tiledb_session_context.py
atolopko-czi: done

apis/python/src/tiledbsoma/collection.py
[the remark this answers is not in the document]
atolopko-czi: done

apis/python/src/tiledbsoma/soma_tiledb_context.py
```diff
@@ -0,0 +1,40 @@
...
+    # Temp workaround pending https://app.shortcut.com/tiledb-inc/story/23827
+    region = os.getenv("AWS_DEFAULT_REGION")
+    if region is not None:
+        cfg["vfs.s3.region"] = cast(str, region)  # type: ignore
+
+    return tiledb.Ctx(cfg)
+
+
+@attrs.define(frozen=True, kw_only=True)
+class SomaTileDBContext:
```
atolopko-czi, line 30: I want to name this in way that captures it is a SOMA-level context and specific to this TileDB-implementation of SOMA, but without confusing it with `tiledb.Ctx`.  Not sure if this name captures the intent, suggestions welcome.

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -1,71 +1,60 @@
...
-        else:
-            tiledb_platform_config = parent._tiledb_platform_config
-            self._ctx = parent._ctx
 
-        self._tiledb_platform_config = tiledb_platform_config or TileDBPlatformConfig()
-        # Null ctx is OK if that's what they wanted (e.g. not doing any TileDB-Cloud ops).
+        if parent is not None:
+            assert (
+                context is None
+            ), "Only one of `context` and `parent` params can be passed as an arg"
```
thetorpedodog, line 43: `assert` statements should only ever be used for internal sanity checks, so:

```suggestion
            if context is not None:
                raise TypeError("Only one of `context` and `parent` params can be passed as an arg")
```

apis/python/src/tiledbsoma/dataframe.py
atolopko-czi: Makes sense re:allowing multiple platform impl options to be specified together. I'll re-introduce that.

What I'm not understanding is why we would allow both "create" and "read" sibling dicts into a `create()` call. The "read" sub-dict wouldn't be needed here, presumably. I was imagining platform_config would be specific to the operation being called.

thetorpedodog: We went with something like this for a few reasons:

1. It would allow a library to use the same `platform_config` object across multiple calls if need be, particularly for complex operations that might have multiple steps (create an array, fill it up, etc.).
2. If two operations have a similarly-named setting that does something different, it won’t be confused between operations (something like read buffer size vs. write buffer size).

atolopko-czi: @thetorpedodog Thanks for explaining those design justifications.  I've reinstated support for the original structure of platform_config. Have a look when you can, thanks!

apis/python/src/tiledbsoma/tiledb_create_options.py
atolopko-czi: keeping as is for now

apis/python/src/tiledbsoma/io.py
```diff
@@ -8,7 +8,7 @@
 import pandas as pd
 import pyarrow as pa
 import scipy.sparse as sp
-import tiledb
+import somacore
```
thetorpedodog, line 11: maybe `from somacore import options` or `from somacore.options import PlatformConfig` so you can use `options.PlatformConfig` or just `PlatformConfig` down the line

thetorpedodog (approved):
I would recommend also getting an approval from somebody to check the TileDB specifics, but from what I can tell this looks good!

apis/python/src/tiledbsoma/io.py
atolopko-czi: done!

atolopko-czi:
> I would recommend also getting an approval from somebody to check the TileDB specifics, but from what I can tell this looks good!

@johnkerl Would that be you? :) I didn't change much in terms of how TileDB create options are extracted from the platform_config and then used, but I did touch the code in small ways. Let me know if you can review at least the `tiledbsoma.options.tiledb_create_options` package. Also note that I will be testing Cell Census builder against this branch as another means of verifying correctness, since it makes use of the `platform_config`.

johnkerl (approved):
🚢

PR #682, opened 2023-01-09
[python] More memory-reduction for ingest [main]

johnkerl:
This is crucial good UX for uptake of the SOMA product.

Here we leverage two things:

* `adata = anndata.read_h5ad(path_to_h5ad, "r")` with the `"r"` already in place (not new)
* Replace `adata.X[:]` with `adata.X` (new) -- this is a paginable object we can use to read CSR/CSC into memory a chunk at a time without needing to read the entire CSR/CSV into memory

Sandbox experiments using `htop` and a 58GB `.h5ad` file have shown peak memory on ingest drops from over 30GB without this PR, to about 15GB with this PR.

Note: currently stacked atop `kerl/resume-mode-main-2` which is also awaiting code review.

### DIFF
--- a/apis/python/src/tiledbsoma/io.py
+++ b/apis/python/src/tiledbsoma/io.py
@@ -1,8 +1,9 @@
 import math
 import time
-from typing import Any, Optional, Sequence, Union
+from typing import Any, List, Optional, Sequence, Union
 
 import anndata as ad
+import h5py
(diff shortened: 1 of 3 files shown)


aaronwolen (approved):
🚀

johnkerl:
Thanks @aaronwolen !

PR #683, opened 2023-01-10
Add a few hints so mypy knows what type things are.

thetorpedodog:


apis/python/src/tiledbsoma/util_tiledb.py
```diff
@@ -66,6 +66,7 @@ def to_tiledb_supported_array_type(x: T) -> T:
     # If the array contains NaN/NA, and the primitive is unable to represent
     # a reasonable facsimile, i.e. not string or float, raise.
     if pd.api.types.is_categorical_dtype(x.dtype):
+        assert isinstance(x, pd.Series)
```
thetorpedodog, line 69: FYI I’m not positive about this assertion. Is this guaranteed to be the case?

johnkerl: Not sure

johnkerl (approved):

PR #685, opened 2023-01-10
add somacore 0.0.0a2 to typecheck run

thetorpedodog:


johnkerl (approved):
🚀

thetorpedodog:
For reasons I don’t fully understand and am not quite sure where is the right place to fix, the tests do not pass on this, for what appears to be that the `SOMAMeasurement`s are not being created with the `SOMAMeasurement` tag in their metadata. @johnkerl I don’t want to push this onto your branch in that broken state so please pull it in yourself if/when you like.

johnkerl:
> For reasons I don’t fully understand and am not quite sure where is the right place to fix, the tests do not pass on this, for what appears to be that the SOMAMeasurements are not being created with the SOMAMeasurement tag in their metadata. @johnkerl I don’t want to push this onto your branch in that broken state so please pull it in yourself if/when you like.

@thetorpedodog #709 -- old bug, just newly detected here -- a fun puzzle! :)

PR #687, opened 2023-01-10
[python] No more memory reduction [main-old]

johnkerl:
This would have been a port of #682 from `main` to `main-old`.

However (as noted in the comments here) `h5py` doesn't suppor the kind of permuted reads that `main-old` requires (and `main` does not require).


aaronwolen (approved):

johnkerl:
Merged with CI bypass since this is a comment-only PR and it's appropriate to free up CI compute slots for PRs that truly need them

PR #674, opened 2023-01-09
Simplify setup

gsakkis:
- Remove duplicated info from setup.cfg and move remaining relevant bits to setup.py.
- Simplify setup.py a bit while at it.

PR #689, opened 2023-01-10
[python] Unbreak clean install

johnkerl:
Several of us found today that if we did `pip uninstall tiledbsoma`, then clean (below), then `pip install -v -e apis/python`, we got a very quick and "successful" build, but on `import tiledbsoma` we got

```
ImportError: dynamic module does not define module export function (PyInit_libtiledbsoma)
```

Bisecting commits between `0.5.0a2` (good) and `main` (bad) identified #674 as the culprit. This PR restores code that was deleted on #674.

Fix validated on laptop with:

```
rm -rf  $(find . -name '*.so' -o -name '*.dylib')
rm -rvf $(find . -name dist -o -name build)
rm -rf venv
rm -rvf test/soco
rm -rvf test/tiledbsoma

pip uninstall -y tiledbsoma

pip install -v -e apis/python

python
>>> import tiledbsoma
```

Everything makes sense here in that the `build_ext` was simply _gone_, so of course post-clean builds were "quick" (not actually compiling C++ code), and of course there was an import fail (C++ code wasn't compiled), and of course this only happened after a clean (C++ code was already compiled in sandbox from before #674) -- this all makes sense _in sandbox_. What I don't understand though is how we had green CI on #674 and ever since -- since these _also_ do a from-scratch install ...

Update: @gspowley points out that this only happens on an _editable_ install (`pip install -e`) which is what we devs do but which CI does not. So now _everything_ makes sense. :)

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -22,14 +22,19 @@ import sys
 
 import setuptools
 import wheel.bdist_wheel
+from setuptools.command.bdist_egg import bdist_egg
+from setuptools.command.build_ext import build_ext
 
 this_dir = pathlib.Path(__file__).parent.absolute()
 sys.path.insert(0, str(this_dir))
 import version  # noqa E402
 
+MODULE_NAME = "tiledbsoma"
+EXT_NAME = "tiledbsoma.libtiledbsoma"
 
-def find_or_build_package_data():
-    # Setup paths
+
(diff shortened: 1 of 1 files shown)


gspowley (approved):
LGTM! Thanks @johnkerl!

PR #691, opened 2023-01-11
Remove bdist_egg command & unused names from setup.py

gsakkis:
Neaten #689; only the `build_ext` command has to be restored.


johnkerl (approved):
Thanks @gsakkis !

Issue #700, opened 2023-01-11
[python] Coalesce wheels-build and PyPI publish


johnkerl:
This needs to be resolved before we publish a 0.5.0a4.

Currently this is split into two files, one of them stubbed:

* https://github.com/single-cell-data/TileDB-SOMA/blob/786cecce398692b14ce55d002b4738fabd0fd197/.github/workflows/py-packaging.yaml#L33-L70
* https://github.com/single-cell-data/TileDB-SOMA/blob/786cecce398692b14ce55d002b4738fabd0fd197/.github/workflows/ci.yml#L88-L132

This needs to be done in one place, and done completely.

See also
https://github.com/single-cell-data/TileDB-SOMA/wiki#branches-and-releases

Update 2023-01-12: see also
https://github.com/single-cell-data/TileDB-SOMA/wiki/PyPI-packaging-WIP
for important information about the migration path

johnkerl:
cc @mlin

johnkerl:
#704 takes care of the dedupe per se.

However there's more work here to be done than I realized. Issue #706 has been opened to track that.

PR #701, opened 2023-01-11
[ci/python] Coalesce PyPI-publish for main branch

johnkerl:
Context: #700


.github/workflows/python-packaging.yml
```diff
@@ -100,3 +117,42 @@ jobs:
...
+    needs: wheels
+
+    # Until general release of the main branch of tiledbsoma we'll follow the convention of tagging releases
+    # 0.5.0a1, 0.5.1a1, etc -- always with the "a1" suffix -- that way PyPI will automagically make these
+    # "prereleases". Then:
+    # pip install tiledbsoma -> 0.1.12 (or whatever the main-old branch is at)
+    # pip install --pre tiledbsoma -> 0.5.0a1 (or whatever the main branch is at)
+
+    # For prime time, once we're in 1.0 release:
+    #if: matrix.os == 'ubuntu-22.04' && github.event_name == 'release' && !github.event.release.prerelease
```
thetorpedodog: Once we do release, I recommend still uploading prerelease packages, so users can install them with `pip install --pre`.

.github/workflows/python-packaging.yml
```diff
@@ -100,3 +117,42 @@ jobs:
...
+      packages_dir: sdist, dist-wheel
+      verbose: true
+
+  # ----------------------------------------------------------------
+  publish-to-pypi:
+    name: Publish package to PyPI
+    needs: wheels
+
+    # Until general release of the main branch of tiledbsoma we'll follow the convention of tagging releases
+    # 0.5.0a1, 0.5.1a1, etc -- always with the "a1" suffix -- that way PyPI will automagically make these
```
thetorpedodog, line 145: Do you mean `0.5.0aX`?

thetorpedodog (commented):
In general this looks reasonable. I am not the expert on the complexities of Python builds with native libraries so I will just drop a couple suggestions here:

johnkerl:
@mlin @thetorpedodog @bkmartinjr I could use a lot of feedback here 🙏

.github/workflows/python-packaging.yml
johnkerl: derp thx :D

.github/workflows/python-packaging.yml
johnkerl: I thought PyPI did this automagically with the `a` in the release name -- PyPI made it a pre-release even if it wasn't a pre-release at the GitHub level?

* https://github.com/single-cell-data/TileDB-SOMA/wiki#branches-and-releases
* When I made [0.5.0a3](https://github.com/single-cell-data/TileDB-SOMA/releases/tag/0.5.0a3) today I didn't mark it as pre-release at the GitHub-tag level and I was able to install it later with `pip install --pre tiledbsoma`

johnkerl: Or oh wait I get it now, nvm -- I'll update the comment

johnkerl: @thetorpedodog 
https://github.com/single-cell-data/TileDB-SOMA/pull/701/commits/6cced85663a648fe34c549420515573f9f82133e

.github/workflows/python-packaging.yml
```diff
@@ -24,11 +36,15 @@ jobs:
       run: python setup.py sdist
       working-directory: ./apis/python
     - name: Upload sdist artifact
+      # One might ask "Upload" -- to where? This means you can go to
```
bkmartinjr, line 39: Tiny, tiny nit:  this is well covered by the docs for the action.  Not sure recapitulating it here is helpful (and if it is, why is it not done for all of the other canned actions)?

bkmartinjr (commented):
I am going to sit out this review, as I don't have sufficient context about the desired build process.  I don't see anything objectionable, but defer to @mlin.

.github/workflows/python-packaging.yml
```diff
@@ -100,3 +117,63 @@ jobs:
...
       # TODO: more-thorough test
+
+  # ----------------------------------------------------------------
+  # Nominally this is on the nightly schedule
+  publish-to-test-pypi:
+    name: Publish package to TestPyPI
+    needs: wheels
+    runs-on: ubuntu-20.04
+
+    steps:
```
mlin, line 128: @johnkerl The main gap I see right now is that these `publish-to-test-pypi` and `publish-to-pypi` steps will each run on a fresh github worker, separate from the workers where we built the sdist and wheels in prior steps (one of which was a macOS worker, QED). Therefore when they start up they don't yet have the sdist tarball and wheels available on their local filesystem.

We can use the download-artifacts action to fetch them -- this is one of the main use cases of uploading to artifacts (in addition to making them available to download from github yourself). The smoke-test step above this one does so for example, and [another example from the cibuildwheel docs](https://cibuildwheel.readthedocs.io/en/stable/deliver-to-pypi/#github-actions).

Two tricky bits that will follow,

1. download-artifacts extracts the zip of artifacts automatically, and I can never remember what their filesystem layout is going to be on the github worker (they might just end up flat in the working directory). I end up temporarily sticking a `find . -type f` in the script to run and find out from the logs. The files may then need to be reorganized (moved into subdirectories) according to what you need to do with them.
2. The first step in this workflow builds an sdist tarball which is then used as the 'gold master' from which all the wheels are then built. It'd be nice to make sure exactly this sdist tarball is also what we upload to PyPI. In contrast I think the gh-action-pypi-publish is probably building a new sdist tarball from the local source checkout. They should be equivalent, but you never know. Maybe there's some way to fiddle with the config to make sure it publishes the sdist tarball that download-artifacts will localize for us.

.github/workflows/python-packaging.yml
johnkerl: I assert it would have been helpful had it been here -- either that, or
as an act of kindness to our co-workers

.github/workflows/python-packaging.yml
johnkerl: @mlin thanks for sharing
which my personal failure to find yesterday cost me an hour of confusion and despair. I'll be sure to consult that link in the future, and I'll work to make it easier for others to find as well.

Now that I'm properly contextualized, I can see why my approach here won't work.

Thanks for your review!

johnkerl: Also: @mlin I believe at the very least (as an initial PR) we should remove this entirely:
https://github.com/single-cell-data/TileDB-SOMA/blob/main/.github/workflows/python-ci.yml#L89-L104
as it's a duplicate of code here:
https://github.com/single-cell-data/TileDB-SOMA/blob/main/.github/workflows/python-packaging.yml#L33-L70

I'll put up a separate PR to remove this redundancy

johnkerl:
#704 does some reorg and #700. If this PR  were to be rebased on top of now-merged #704, little would be left other than my erroneous implementation.

PR #704, opened 2023-01-12
[ci/python] Dedupe PyPI stubs

johnkerl:
This is a preliminary PR to go underneath the (now WIP) #701. (See also tracking issue #700.)

Unlike #701, it doesn't attempt to implement anything that doesn't currenly exist. Rather, it maintains the manual wheel flow from
and simply:

* Removes a redundant step from `python-ci.yml`
* Leaves CI per se in `python-ci.yml` and leaves Python package in `python-packaging.yml`
* Updates the to-do stubs with my current understanding of where we're going


thetorpedodog (commented):
Looks pretty good overall but I defer to @mlin who is much better at this packaging thing than I am.

mlin (approved):
LGTM this comes down to just comments, right? To be clear, I definitely think we should automate the PyPI publishing once we're comfortable all the cibuildwheel stuff is working reliably and to our satisfaction. Now would be fine, except it will probably take a couple of hours of trial-and-error and maybe we're all juggling too much this week. Also happy to circle back on it later this month.

johnkerl:
> this comes down to just comments

@mlin no, at https://github.com/single-cell-data/TileDB-SOMA/pull/704/files there are guidance (in the form of comments) as well as deduping -- removing no-longer-used (legacy from `main-old`) wheels stanza from `python-ci.yml`

PR #717, opened 2023-01-13
Separate pip installation of prereqs and tiledbsoma itself.

thetorpedodog:
- Separates the `pip install` steps for the build-/test-only deps (pytype, typeguard, etc.) from the package itself.
- Installs the `wheel` package so that we use wheels to install where possible (editor's note: why is this not included by default???).

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -46,7 +46,12 @@ jobs:
         python-version: ${{ matrix.python-version }}
 
     - name: Install Python packages
-      run: python -m pip -v install pytest pytest-cov typeguard typing_extensions apis/python
+      run: |
+        # Python package install commands
+        # 1. Testing prereqs.
+        python -m pip -v install --upgrade pip wheel pytest pytest-cov typeguard
+        # 2. The package itself.
+        python -m pip -v install apis/python
       env:
         CC: ${{ matrix.cc }}
         CXX: ${{ matrix.cxx }}


johnkerl:
> editor's note: why is this not included by default???

@thetorpedodog I lack the Python expertise (which you have) to know what this means to you

Please spell this out at a more elementary level and I'll try to answer your question

thetorpedodog:
> > editor's note: why is this not included by default???
> 
> @thetorpedodog I lack the Python expertise (which you have) to know what this means to you
> 
> Please spell this out at a more elementary level and I'll try to answer your question

This is a complaint about the fact that the `wheel` package is not part of Python core / installed by default despite being almost essential for package management using pip etc., not a question that needs answering.

johnkerl (approved):
@thetorpedodog I'll accept this PR (it clearly does no harm) but AFAICT you've educated no one as to why this is a beneficial change -- I'd love to learn from your expertise -- please tell me what the goodness is here. What will work differently/quicker/better/[?] after this PR compared to before?

thetorpedodog:
The short version is I was looking at the logs from some CI run and noticed a message to the effect of "wheel is not installed so running setup.py manually" and figured now was a good time to make sure wheel is installed.

johnkerl:
Thanks @thetorpedodog ! :) (This is the kind of thing I want in PR descriptions moving forward -- there's awesome context here -- let's get it at the top, from the start, & help level one another's knowledge up.)

PR #721, opened 2023-01-13
Add quote marks around a few strings in C++ errors.

thetorpedodog:
A few strings in error messages were unquoted, which meant it was sometimes ambiguous as to what was going on or what the invalid value might be.


gspowley (approved):
LGTM!

PR #723, opened 2023-01-13
[ci] Use fewer CI resources for PR-pushes, more for merge [pending more research, please do not review]

johnkerl:
## Issue and/or context:

#693

## Changes:

Narrative is in the comments

## Notes for Reviewer: 

Non-admins of this repo may lack access to this status page but it's routinely pegged at 5 running MacOS jobs:
https://github.com/organizations/single-cell-data/settings/actions/hosted-runners

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -24,6 +24,10 @@ jobs:
         # Free/Pro/Team/Enterprise, respectively, are 5/5/5/50 -- not say 5/10/25/50.  In practice
         # we find that the number of concurrent MacOS runners is the number one throttle on our CI.
         # Running MacOS CI immediately post-merge is an acceptable tradeoff.
+        # See also
+        # * https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows
+        isPR:
+          - ${{ github.event_name == "pull_request" }}
         include:
           - runs-on: ubuntu-22.04
             cc: gcc-11
@@ -31,9 +35,11 @@ jobs:
           - runs-on: macos-12
             cc: gcc-11
             cxx: g++-11
-          - if: github.event_name == "pull_request" or github.event_name == "push"
-            os: [ubuntu-22.04]
-            python-version: ['3.7', '3.10']
+        exclude:
+          - isPR: true
+            os: macos-12
+          - isPR: true
+            python-version: ['3.8', '3.9']
 
     steps:
     - name: Select XCode version
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -28,7 +28,7 @@ jobs:
         # See also
         # * https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows
         isPR:
-            - ${{ github.event_name == 'pull_reauest' }}
+            - ${{ github.event_name == 'pull_request' }}
         include:
           - runs-on: ubuntu-22.04
             cc: gcc-11
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -4,10 +4,13 @@ on:
   pull_request:
     paths-ignore:
       - 'apis/r/**'
+    isPR: true
   push:
     branches: [main]
+    isPR: false
   release:
     types: [published]
+    isPR: false
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/python-ci.yml
```diff
@@ -15,17 +15,30 @@ jobs:
...
+              cc: gcc-11
+              cxx: g++-11
+        if: github.event_name == "pull_request"
+          # GHA MacOS caps for tiers Free/Pro/Team/Enterprise are 5/5/5/50 :(
+          # -- not 5/10/25/50 or anything reasonable :(.
+          # Since most of our CI activity is people pushing multiple iterations
+          # of PRs, let's ration some resources for that use-case, and save
+          # the full coverage for PR merges and releases.
+          os: [ubuntu-22.04]
+          python-version: ['3.7', '3.10']
```
atolopko-czi, line 37: maybe even just a single python version `3.10`?

.github/workflows/python-ci.yml
```diff
@@ -15,17 +15,30 @@ jobs:
...
+          # os: [ubuntu-22.04, macos-12, windows-2019]
+          python-version: ['3.7', '3.8', '3.9', '3.10']
+          include:
+            - runs-on: ubuntu-22.04
+              cc: gcc-11
+              cxx: g++-11
+            - runs-on: macos-12
+              cc: gcc-11
+              cxx: g++-11
+        if: github.event_name == "pull_request"
```
atolopko-czi, line 30: I don't pretend to fully understand the github workflow eventing semantics, but I would have thought we want to trigger a build per `push` event, rather than `pull_request`, since the latter seems to involve "activity types" like 
- assigned
- unassigned
- labeled
- unlabeled
- opened
- edited
- closed
...

johnkerl: Super-good call @atolopko-czi ... I found [this](https://stackoverflow.com/questions/70767090/github-actions-on-push-and-on-pull-request-difference#:~:text=In%20general%2C%20push%20will%20trigger,from%20a%20fork%20for%20example.) which suggests that `pull_request` and `push` are the same unless we're using a fork-repo model (which we aren't)

I'll add a comment to that effect for the benefit of the next person to read through here.

atolopko-czi: GTK! Thanks for the clarifying comment.

.github/workflows/python-ci.yml
atolopko-czi: Fair enough!

atolopko-czi (approved):
LGTM! Thanks for improvement!

thetorpedodog:
unfortunately some of what you're trying to do is unsupported:

- there is no `if` syntax for exclusions from a matrix: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#excluding-matrix-configurations
- the `matrix` information is not available in the `if` block for a job: https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability

the best idea I can think of is to effectively set a `run-on-pr` variable, and then add `if: ${{ github.event_name != "pull_request" || run-on-pr }}` to each step of the job, something like:

```yaml
jobs:
  some-job:
    matrix:
      os: [ubuntu-latest, macos-latest]
      python-version: ["3.7", "3.8", "3.9", "3.10"]
      include:
        - os: ubuntu-latest
          python-version: "3.10"
          run-on-pr: true

    steps:
      - uses: blah/blah
        if: github.event_name != "pull_request" || matrix.run-on-pr
      - run: echo "blah blah"
        if: github.event_name != "pull_request" || matrix.run-on-pr
```

which is not ideal.

atolopko-czi:
What if python-ci.yml is split into two workflows, one for pull_requests triggers and one for push/release triggers?  This would introduce code duplication, which is less than ideal.

thetorpedodog:
It also might be possible to make a single workflow, which [we can then reuse](https://docs.github.com/en/actions/using-workflows/reusing-workflows) and launch it from separate matrices for each individual event type.

johnkerl:
> which is not ideal

@thetorpedodog this is far better than the status quo, or any of my failed attempts -- and also better than outright code duplication

@atolopko-czi let's do it! 😎

atolopko-czi:
Closing, since we now have other PRs for CI performance improvements: https://github.com/single-cell-data/TileDB-SOMA/issues/693#issuecomment-1397555637

johnkerl:
Thanks @atolopko-czi !!! 😎

PR #730, opened 2023-01-17
[python] Stateful (open/close) objects

mlin:
Adds open/close/enter/exit methods to TileDBObject and (for Arrays) reuse `tiledb.Array` handle for consecutive reads or writes.

- Using open/close is optional. If the developer calls a read/write method without first opening the object, it self-opens and closes in the appropriate mode, just for the duration of that operation.
- Arrays reuse the python `tiledb.Array` handle while the object is open. We want to reuse handle within `SOMAReader` too, but some libtiledbsoma refactoring will be needed to accomplish that (in follow-up PRs).
- Collection also keeps open the `tiledb.Group` handle, although it's a bit pointless right now because of how the implementation (i) already caches the group contents for reads and (ii) invalidates and repopulates that cache after *each* write. Future follow-up improvements/refactoring possible. 
- Nothing about timestamps in this PR; that work is proceeding in parallel, and the open mode will determine the appropriate timestamps to use for each operation.


mlin:
@gspowley re SOMAReader & https://github.com/single-cell-data/SOMA/issues/44#issuecomment-1385552927

I'm a little concerned about how the SparseNDArrayReader in the new fluent API (or indeed the iterator style we used before that) retains a reference to, and keeps using, the SOMAReader we initialize it with. Isn't it possible to run multiple overlapping queries, i.e. to have multiple SparseNDArrayReader alive at once, in which case each one would need its own SOMAReader in order to remain coherent, right?

Is there some way to separate the TileDB array handle/metadata SOMAReader uses, from the query-specific state/stack it maintains over the course of one query?

gspowley:
> Isn't it possible to run multiple overlapping queries, i.e. to have multiple SparseNDArrayReader alive at once, in which case each one would need its own SOMAReader in order to remain coherent, right?

Yes, this is possible and does require multiple SOMAReaders. The way it is implemented today, each SOMAReader opens the array separately. 

This statement from https://github.com/single-cell-data/SOMA/issues/44#issuecomment-1385552927 was a too strong:
> we need to keep one SOMAReader object instantiated per array to hold that array open for reads.

It's not a requirement to have one SOMAReader per array. That is just one python implementation that could take advantage of the existing SOMAReader implementation.

> Is there some way to separate the TileDB array handle/metadata SOMAReader uses, from the query-specific state/stack it maintains over the course of one query?

Yes, we could refactor the SOMAReader to hold a map of opened arrays (URI -> array) and reuse the open array handles. In this case, we could hold arrays open until they were explicitly closed from python.

I'll look into that.

gspowley:
A better design would be to create a `clib.Array` (analogous to `tiledb.Array`) that is held open in python and passed to the `clib.SOMAReader`.

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -20,6 +26,11 @@ class TileDBObject(ABC, somacore.SOMAObject):
     _context: SOMATileDBContext
     _metadata: MetadataMapping
 
+    # "r", "w", or falsey
+    _open_mode: str = ""
```
bkmartinjr, line 30: tiny nit:  can we type "mode" throughout, e.g., `Optional[Literal["r", "w"]]`?  It would help annotate the API upstream, e.g. in the `open()` method.

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -113,15 +125,100 @@ def exists(self) -> bool:
...
+                    self._metadata.get(util.SOMA_OBJECT_TYPE_METADATA_KEY)
+                    == self.soma_type
+                )
         except tiledb.cc.TileDBError:
             return False
 
-    @abstractmethod
-    def _tiledb_open(self, mode: str = "r") -> Union[tiledb.Array, tiledb.Group]:
-        """Open the underlying TileDB array or Group"""
+    def open(self: TTileDBObject, mode: str = "r") -> TTileDBObject:
```
bkmartinjr, line 136: Should abstract variants of these methods/types be pushed upstream to `somacore`?  Or is that a later PR?

apis/python/src/tiledbsoma/tiledb_object.py
johnkerl: More than I nit, I think this is important.

* In the open, we can default an _argument_ which is `None` to be `"r"` -- as in Python's `open(file, mode="r", ...)`, or `tiledb.open` which also defaults `mode` to `"r"`
* But (here) we store always either`"r"` or `"w"` as an attribute

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -113,15 +125,100 @@ def exists(self) -> bool:
...
+        ), f"use {self.__class__.__name__}.open() as context manager"
+        return self
+
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_value: Optional[BaseException],
+        traceback: Optional[TracebackType],
+    ) -> None:
+        if self._open_mode:
```
bkmartinjr, line 180: given that close is idempotent, I'd just call it without the conditional.  Style choice only, so totally up to you.

apis/python/tests/test_dense_nd_array.py
```diff
@@ -80,17 +80,18 @@ def test_dense_nd_array_read_write_tensor(tmp_path, shape: Tuple[int, ...]):
...
 
-    t = b.read((slice(None),) * ndim, result_order="row-major")
-    assert t.equals(pa.Tensor.from_numpy(data))
+        t = b.read((slice(None),) * ndim, result_order="row-major")
+        assert t.equals(pa.Tensor.from_numpy(data))
 
-    t = b.read((slice(None),) * ndim, result_order="column-major")
-    assert t.equals(pa.Tensor.from_numpy(data.transpose()))
+        t = b.read((slice(None),) * ndim, result_order="column-major")
+        assert t.equals(pa.Tensor.from_numpy(data.transpose()))
```
bkmartinjr, line 95: on at least a few of these, worth testing that the __exit__ actually called close()

apis/python/tests/test_tiledb_object.py
```diff
@@ -40,3 +41,65 @@ def test_mutually_exclusive_create_args():
...
+    o.open("r")
+    assert o._open_mode == "r"
+    o.close()
+    assert not o._open_mode
+    o.close()  # no-op
+    assert not o._open_mode
+
+    with o.open("w"):
+        assert o._open_mode == "w"
+    assert not o._open_mode
```
bkmartinjr, line 59: I think this motivates the creation of a public `closed` (or `is_open`) property on the object, much like a Python file.

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -113,15 +125,100 @@ def exists(self) -> bool:
...
+                raise RuntimeError(
+                    self.__class__.__name__ + " is already open read-only"
+                )
+            yield
+        else:
+            with self.open(mode):
+                yield
+
+    def __del__(self) -> None:
+        self.close()
```
bkmartinjr, line 215: Need to call __del__ on super() if it exists, as this definitely may get specialized in a sub-class.

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -113,15 +125,100 @@ def exists(self) -> bool:
...
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_value: Optional[BaseException],
+        traceback: Optional[TracebackType],
+    ) -> None:
+        if self._open_mode:
+            self.close()
+
+    @contextmanager
+    def _maybe_open(self, mode: str = "r") -> Iterator[None]:
```
bkmartinjr, line 184: Nit: I find this method name super confusing. Outside of error cases, it _always_ performs an `open` - there is no "maybe".  Perhaps a better name would capture the "open if not already open" concept?

Spitballs:
* _ensure_open()
* _open_if_not_open()
* _verify_open()

johnkerl: +1 for `_ensure_open()` :)

apis/python/src/tiledbsoma/tiledb_array.py
```diff
@@ -36,31 +36,36 @@ def schema(self) -> pa.Schema:
...
-        # and our return value's __exit__ on exit from the body of the with-block. The tiledb
-        # array object does both of those things. (And if it didn't, we'd get a runtime AttributeError
-        # on with-as, flagging the non-existence of the __enter__ or __exit__.)
-        return tiledb.open(self._uri, mode=mode, ctx=self._ctx)
+    # lazy tiledb.Array handle for reuse while self is "open"
+    _open_tiledb_array: Optional[tiledb.Array] = None
+
+    @property
+    def _tiledb_obj(self) -> tiledb.Array:
+        "get the open tiledb.Array handle (opening it if needed)"
```
bkmartinjr, line 44: I"m curious why we use a lazy open of the underlying resource, rather than doing it at the time of the original `open()` call?  

The delayed call to `tiledb.open` means that errors associated with this will be delayed.  For example, if the `tiledb.open` fails (say due to bad URI), it will raise during a `read`, rather than during the `open`.  Raising early (during `open`) feels more natural (principle of lease surprise?) to me.

Shelnutt2: I'd also add that lazy opening negates some of the performance optimizations we were looking for by making the 
"first" query slower than subsequent ones.

I think that as post-1.0 nearly all of this logic must be moved into C++ outside of python/R so I'm not going to push for one or the other here. Moving this to C++ will yield many more improvements for Reads and longer term writes.

apis/python/src/tiledbsoma/collection.py
```diff
@@ -109,6 +110,25 @@ def _create(self, soma_type: str) -> "CollectionBase[CollectionElementType]":
         self._cached_values = {}
         return self
 
+    # lazy tiledb.Group handle for reuse while self is open
+    _open_tiledb_group: Optional[tiledb.Group] = None
+
+    @property
+    def _tiledb_obj(self) -> tiledb.Group:
```
bkmartinjr, line 117: same question as with tiledb_arra.py::_tiledb_obj -- why is this lazy?   I'd rather have the early error reporting and latency win from early....

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -113,15 +125,100 @@ def exists(self) -> bool:
...
+            raise RuntimeError(self.__class__.__name__ + " is already open")
+        self._open_mode = mode
+        return self
+
+    def close(self) -> None:
+        """
+        Release any resources held while the object is open. Closing an already-closed object is a
+        no-op.
+        """
+        if self._open_mode:
```
bkmartinjr, line 164: Assuming that somacore.SOMAObject will have open/close methods, there should probably be a super().close() here.

bkmartinjr (changes requested):
I generally like this - with two issues we should resolve:
* _open_mode - the typing of this param/arg, and whether it is better represented as two bits of state (is_open, and open_mode)
* lazy vs eager opening of the underlying tiledb array handle (I prefer eager)

apis/python/src/tiledbsoma/collection.py
```diff
@@ -109,6 +110,25 @@ def _create(self, soma_type: str) -> "CollectionBase[CollectionElementType]":
...
+            self._open_tiledb_group = self._close_stack.enter_context(
+                tiledb.Group(
+                    self._uri, mode=self._open_mode, ctx=self.context.tiledb_ctx
+                )
+            )
+        return self._open_tiledb_group
+
+    def close(self) -> None:
+        self._open_tiledb_group = None
+        super().close()  # closes self._open_tiledb_group via self._close_stack
```
gsakkis, line 130: `_tiledb_obj` and `close` are almost identical between `TileDBArray` and `CollectionBase`. The common logic can be pulled to the base `TileDBObject` by:
1. replacing `_open_tiledb_array` and `_open_tiledb_group` with a `__tiledb_obj` private attribute of TileDBObject (`Union[tiledb.Array, tiledb.Group, None]`)
2. adding a bool class attribute, say `TileDBObject._open_group`, and override it with `TileDBArray._open_group = False` and `CollectionBase._open_group = True`
3. In `TileDBObject._tiledb_obj` use `self._open_group` to determine whether to call `tiledb.open` or `tiledb.Group`

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -49,6 +62,7 @@ def __init__(
             self._context = context or SOMATileDBContext()
 
         self._metadata = MetadataMapping(self)
+        self._close_stack = ExitStack()
```
gsakkis, line 65: AFAICT this is not really needed; at most a single object is ever added to the stack.

apis/python/src/tiledbsoma/tiledb_object.py
mlin: Added `OpenMode = Literal["r","w"]`, now the internal attribute is `Optional[OpenMode]` and the open() argument is `mode: OpenMode`. I don't see appeal in separating them, since there are three and not four valid states.

apis/python/src/tiledbsoma/tiledb_object.py
mlin: Can I defer to @thetorpedodog to look at how much of this should lift up into `somacore` since he plans some ~immediate refactoring anyway?

(Possibly most of the tiledb_object.py diff should lift up, since it's just managing the state transitions and exception handling in a totally generic way.)

apis/python/src/tiledbsoma/tiledb_object.py
mlin: There's no default `__del__` to call unless we know the superclass definitely provides one; left a comment to this effect. `__del__` is weird in general with a [tortured history](https://peps.python.org/pep-0442/), I'm not sure if it's conventional is to try to compensate for developer forgetting to close like this.

apis/python/src/tiledbsoma/collection.py
mlin: Made them eager.

apis/python/src/tiledbsoma/tiledb_object.py
mlin: @gsakkis That is true considering this diff in isolation but,

1. Arrays: the C++ SOMAReader currently has a totally separate TileDB handle from the python code; this PR reuses the latter and only anticipates reusing the former, because [some C++ refactoring](https://github.com/single-cell-data/TileDB-SOMA/pull/730#issuecomment-1397493151) is needed (in follow-up PR).
2. Collections: when we write to a collection we then also refresh our cached view of it. But, the tiledb.Group handle opened in write mode doesn't allow us to read it, so we need to open a separate handle in read mode to do the cache refresh. Currently we are not reusing that read-mode handle, but it'd be nice to as another minor follow-up improvement.

mlin: There is also that ExitStack helps to ensure reasonable things happen when exceptions occur in/around the unwind.

apis/python/src/tiledbsoma/collection.py
mlin: I think the subclasses will end up needing to override them anyway to (i) specialize the typing, i.e. `TileDBArray._tiledb_obj -> tiledb.Array` not `Union[tiledb.Array, tiledb.Group] and (ii) to reuse multiple handles as mentioned.

I'm personally averse to the `TileDBObject._open_group` which would imbue the base class with knowledge of its subclasses, not to say that's absolutely forbidden but I think one should usually avoid.

apis/python/src/tiledbsoma/collection.py
```diff
@@ -341,8 +377,8 @@ def _del_element(self, key: str, skip_cache_reload: bool = False) -> None:
         if self._cached_values is not None:
             del self._cached_values[key]
         try:
-            with self._tiledb_open("w") as G:
-                G.remove(key)
+            with self._maybe_open("w"):
+                self._tiledb_obj.remove(key)
```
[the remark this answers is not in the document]
mlin: Then subclasses would need to override `_ensure_open` (nee `_maybe_open`) just to specialize the type -- seems like a push to me? Can start the context body with `G = self._tiledb_obj` if using it a lot.

apis/python/src/tiledbsoma/collection.py
gsakkis: It doesn't require knowledge of its subclasses, it just reduces the differences of `_tiledb_object` (or `_sub_open` currently) down to a single boolean.

I guess it comes down to personal preferences, for me DRY code > super strictly-typed code (even more so now that tiledb-py type annotations are not exported; `tiledb.Array == tiledb.Group == Any`)

apis/python/src/tiledbsoma/tiledb_object.py
bkmartinjr: Works for me.  I also noticed that @thetorpedodog made the same change upstream in the SOMA ABC, so things are aligned.

apis/python/src/tiledbsoma/tiledb_object.py
bkmartinjr: I am going to defer to @thetorpedodog on this one.  In other places in our code base, we do some fairly convoluted stuff to handle this edge case, eg, 

```
    def __del__(self) -> None:
        """Ensure that we're closed when our last ref disappears."""
        # If any superclass in our MRO has a __del__, call it.
        sdel = getattr(super(), "__del__", lambda: None)
        sdel()
```

apis/python/src/tiledbsoma/tiledb_object.py
bkmartinjr: Yes, this seems sensible, and in another PR @thetorpedodog mentioned this was his plan.  Resolving this comment.

apis/python/src/tiledbsoma/tiledb_array.py
bkmartinjr: Resolving, as this is now eager.

bkmartinjr (approved):

apis/python/tests/test_tiledb_object.py
```diff
@@ -40,3 +47,77 @@ def test_mutually_exclusive_create_args():
...
+    # closed after an exception during open()
+    bad = False
+    o.mock_open_error = True
+    try:
+        with o.open("r"):
+            bad = True
+        bad = True
+    except Exception:
+        pass
+    assert not bad
```
thetorpedodog, line 122: lol

thetorpedodog (approved):
I am adding a few notes here BUT I strongly feel this should be submitted so that it can go in and then be adapted for the new abstraction style, rather than remaining in limbo and having to be repeatedly re-synchronized with HEAD.

These are **not** suggestions on what to change now, just notes for us for future work.

johnkerl (approved):

gspowley (approved):

Issue #686, opened 2023-01-10
[python/r/spec] Normalize parameter names between various classes


bkmartinjr:
There are a number of places parameter names should be the same, but are not.

Good example:
> DataFrame.read(ids=...)
> SparseNDArray.read(coords=...)

`ids` and `coords` are the same thing, and should have the same name IMHO

[ ] Python -- on #732 
[ ] R -- on #748

johnkerl:
Note this also affects the spec repo https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md

johnkerl:
Has been done

PR #732, opened 2023-01-17
[python] Update to version 0.0.0a3 of somacore.

thetorpedodog:
- Moves `experiment_query` to `somacore`.
- Renames all `ids`/`slices`/`coords` parameters to `coords` (fixing #686 for Python)
- Explicitly identify every parameter to read methods (no more `**_`).
- Replaces internal result-order strings with enums (the API still allows the use of the string names).
- Remove enum types described by `somacore`.


johnkerl (approved):
🚢 
Awesome thanks @thetorpedodog !!!

Issue #433, opened 2022-10-17
Implement a clean error-handling class/protocol and propagate through the code


johnkerl:
Audit tiledbsoma-py for calls to `Exception` and `assert`. These should be raising `TypeError`, `ValueError`, etc., or `SOMAError`.

Exception: `assert` is totally fine in test code.

See also #328 which is a duplicate of this but which has its own interesting narrative.

johnkerl:
#742 is the result of an audit for Python, and `stopifnot` is culturally acceptable for R.

johnkerl:
#742 handles bare `Exception` in Python but I neglected to handle bare `assert` in non-test code as well -- re-opening

PR #742, opened 2023-01-17
[python] Audit for bare Exception

johnkerl:
## Issue and/or context:

#433 

## Changes: 

Just one found! :)

### DIFF
--- a/apis/python/tools/ingestor
+++ b/apis/python/tools/ingestor
@@ -218,7 +218,7 @@ def ingest_one(
         if not soco.exists():
             soco.create()
         if not soco.exists():
-            raise Exception(f"Could not create SOCO at {soco.uri}")
+            raise tiledbsoma.exception.SOMAError(f"Could not create SOCO at {soco.uri}")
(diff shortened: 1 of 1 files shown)


atolopko-czi (approved):
LGTM, as a code change. Is this is sufficient to address to #433?  Are all exceptions propagating back to the user via top-level calls known to be `SOMAError` or a subclass? I assume that's the intention for #433?  And if so, are unit tests verifying this?

johnkerl:
@atolopko-czi what we talked about today in the sprint-planning call was a recursive grep -- which I did & which found only this

Unit-testing would require wrapping everything in the API that can throw, and checking its exception class isn't `Exception` -- I don't see a more compact way to do it, and this sketch seems an over-wordy way to go about it ...

johnkerl:
@atolopko-czi I just realized I forgot about "base assert" as well

I updated the description of #433 and I'll have a follow-on PR for `assert` in package (totally fine to leave `assert` in test code of course)

PR #743, opened 2023-01-18
Add pre-commit hook for mypy and remove explicit mypy step from python-ci

gsakkis:
Run `mypy` via `pre-commit` so that typing errors are caught before being committed.

One small downside is the need to specify explicitly the dependencies to be installed in the mypy virtualenv (e.g. `somacore`), instead of having all the `tiledbsoma` dependencies already installed as on https://github.com/single-cell-data/TileDB-SOMA/pull/726. That's because AFAICT the `additional_dependencies` option can install only PyPI packages, not local dependencies (e.g. `apis/python/`).

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -45,12 +45,12 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
-    - name: Install testing/linting prereqs
-      run: python -m pip -v install -U pip pytest-cov typeguard pre-commit mypy types-setuptools
+    - name: Install testing prereqs
+      run: python -m pip -v install -U pip pytest-cov typeguard
 
     - name: Run pre-commit hooks on all files
       if: matrix.os == 'ubuntu-22.04' && matrix.python-version == '3.9'
-      run: pre-commit run -a -v
+      run: python -m pip -v install pre-commit && pre-commit run -a -v
 
     - name: Check C++ Format
       if: matrix.os == 'ubuntu-22.04' && matrix.python-version == '3.9'
@@ -67,11 +67,6 @@ jobs:
 
     - name: Show XCode version
       run: clang --version
-
-    - name: Check type annotations with mypy
-      if: matrix.os == 'ubuntu-22.04' && matrix.python-version == '3.9'
-      working-directory: apis/python
-      run: python -m mypy .
 
     - name: Generate test data
       shell: bash
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,3 +15,10 @@ repos:
     - id: flake8
       additional_dependencies: ["flake8-bugbear==22.12.6"]
       args: ["--config", "apis/python/setup.cfg"]
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.991
+    hooks:
+    - id: mypy
+      additional_dependencies: ["types-setuptools", "somacore==0.0.0a3"]
+      args: ["--config-file=apis/python/setup.cfg", "apis/python/src", "apis/python/tools"]
+      pass_filenames: false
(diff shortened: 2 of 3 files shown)


thetorpedodog:
The mypy configuration in `pre-commit` does not install dependencies and so will miss out on any errors involving libraries that are imported. It’s useful as a step in the commit process but does *not* replace a full mypy run. For instance, this would be allowed:

```python
import numpy

numpy.here.there.everywhere
```

cf. the comment in https://github.com/single-cell-data/SOMA/commit/d0df9ed67647d7c0d9d12d191addceebf7d7ed39

gsakkis:
It does install any dependencies listed in `additional_dependencies`, as well as their dependencies recursively. Numpy is a dependency of somacore so this just works as is:
```
$ pre-commit run mypy -a 
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

apis/python/src/tiledbsoma/io.py:36: error: Module has no attribute "here"  [attr-defined]
Found 1 error in 1 file (checked 29 source files)
```

thetorpedodog:
> It does install any dependencies listed in `additional_dependencies`, as well as their dependencies recursively. Numpy is a dependency of somacore so this just works as is:
> 
> ```
> $ pre-commit run mypy -a 
> mypy.....................................................................Failed
> - hook id: mypy
> - exit code: 1
> 
> apis/python/src/tiledbsoma/io.py:36: error: Module has no attribute "here"  [attr-defined]
> Found 1 error in 1 file (checked 29 source files)
> ```

Ah, because `somacore` includes so many of the dependencies of `tiledbsoma`, it will be much better covered. However, any other dependencies of `tiledbsoma` that are *not* included in that list will not be included.

If it were me I would still keep the full mypy run to ensure something like that doesn’t happen, and use the `pre-commit` step just as a lighter-weight (useful but not complete) step to run when doing an actual commit.

gsakkis:
> However, any other dependencies of tiledbsoma that are not included in that list will not be included.

In practice this doesn't seem to be a problem here:
- Not including `somacore` in `additional_dependencies` results in a dozen errors; you'll get a loud reminder if you forget it.
- All other missing dependencies don't have type annotations; you can see this by commenting out  `ignore_missing_imports = True`:
```
...
Skipping analyzing "pandas": module is installed, but missing library stubs or py.typed marker  [import]
Skipping analyzing "scipy": module is installed, but missing library stubs or py.typed marker  [import]
Skipping analyzing "pyarrow": module is installed, but missing library stubs or py.typed marker  [import]
Skipping analyzing "anndata": module is installed, but missing library stubs or py.typed marker  [import]
Skipping analyzing "tiledb": module is installed, but missing library stubs or py.typed marker  [import]
Skipping analyzing "h5py": module is installed, but missing library stubs or py.typed marker  [import]
...
```

So the type checking will be incomplete only if you forget to add a dependency that both (a) comes with type annotations and (b) does not complain if you omit it, which IMO is rather unlikely.

johnkerl:
@gsakkis @thetorpedodog this is all wondeful engagement, although sadly somewhat over my particular head

In terms of detailed correctness, I defer to @thetorpedodog 's accept/reject

In terms of big picture, my concern is that we are still running `mypy` in CI. That appears to be the intention; but on this PR the "Run pre-commit hooks on all files" step is skipped in CI so I can't double-check ...

gsakkis:
Pre-commit is running on only one combination of the test matrix (`if: matrix.os == 'ubuntu-22.04' && matrix.python-version == '3.9'`) as before this PR. Pretty easy to change to run always if you prefer.

thetorpedodog (approved):
It looks like this will probably work OK. I would still *prefer* to see mypy run in the regular venv with all the deps installed, since the additional runtime is negligible, but will not insist on it.

johnkerl:
Merging ... any second thoughts, we can file a follow-on PR ...

Issue #744, opened 2023-01-18
[python] Switch from `ndims` to `ndim` (now to-do-tracker only)


johnkerl:
NumPy, SciPy, Pandas, et al. use `ndim` -- so should we. Affects spec, Python impl, and R impl.

See also:

* https://github.com/single-cell-data/SOMA/issues/84
* https://github.com/single-cell-data/SOMA/pull/85
* #747 

Update:

* #747 has a back-compat alias of `ndims` to `ndim` just until the `somacore` 0.0.0a4` release
* This task now remains open solely as a reminder to act on the TODO there and remove that back-compat alias

bkmartinjr:
`somacore//data.py` defines NDArray class with `ndims`

https://github.com/single-cell-data/SOMA/blob/d0df9ed67647d7c0d9d12d191addceebf7d7ed39/python-spec/src/somacore/data.py#L91

johnkerl:
@bkmartinjr yes, that's correct.

More details to come on https://github.com/single-cell-data/SOMA/issues/84

PR #747, opened 2023-01-18
[python] Take ndim accessor from somacore

johnkerl:
## Issue and/or context:

Issues https://github.com/single-cell-data/SOMA/pull/85 and https://github.com/single-cell-data/TileDB-SOMA/issues/744

## Changes:

Following discussion at https://github.com/single-cell-data/SOMA/pull/85

## Notes for Reviewer:

To make this mergeable, we'll need a 0.0.0a4 of `somacore` first


thetorpedodog:
I would add `ndim` as an alias to `ndims` (so that it can be used in downstream code from now) and add a TODO to remove the implementation once the next release is out.

bkmartinjr (approved):
LGTM

PR #749, opened 2023-01-19
[python] Apply timestamps through SOMATileDBContext

mlin:
#540

Add `read_timestamp` and `write_timestamp` properties to `SOMATileDBContext`. Then, feed these through to all array read and write operations (including the C++ `SOMAReader`).

NOTE: timestamps appear not yet supported by TileDB for (i) array creation and (ii) any group operation. Please correct me if I've got that wrong!


apis/python/src/tiledbsoma/options/soma_tiledb_context.py
```diff
@@ -39,4 +40,29 @@ class SOMATileDBContext:
...
     """Allows "relocatability" for local disk / S3, and correct behavior for TileDB Cloud."""
 
-    # read_timestamp, e.g.
+    read_timestamp_start: Optional[int] = None
+    "Timestamp range start for all array read operations. Usually, implicitly zero."
+
+    read_timestamp_end: Optional[int] = None
+    """
+    Timestamp range end for all array write operations.
+    If unspecified, then implictly the latest data as of the first access to each array.
```
mlin, line 49: Q: currently leaving the default timestamp to None meaning no consistency between different arrays by default. If we want consistency by default then we could make the read_timestamp_end default to the time of context initialization. Any opinions?

libtiledbsoma/src/soma_reader.cc
```diff
@@ -78,15 +81,33 @@ SOMAReader::SOMAReader(
...
+            array->set_open_timestamp_end(timestamp->second);
+            // FIXME: tiledb::Array::set_open_timestamp_{start,end} are only
+            // effective if called before [re]opening the array. The
+            // tiledb::Array constructor internally opens the array already, so
+            // we need to reopen() for this to have effect.
+            // https://github.com/TileDB-Inc/TileDB/blob/6f92d2864440cec17474f2aae22f223147828c6a/tiledb/sm/cpp_api/array.h#L97
+            // We might need to open the array using the C API and then use the
+            // alternate tiledb::Array constructor that takes ownership of the C
+            // array.
+            array->reopen();
```
mlin, line 104: @gspowley I ran into this weird quirk with the TileDB C++ API, appreciate your input if you have a POV or if I missed something. The way the tiledb::Array constructor already opens the array seems to constrain us to reopen() in order to set the start & end timestamps. (Without using deprecated constructor variants supporting the end timestamp only.)

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -124,6 +124,7 @@ def create(
             ctx=self._ctx,
         )
 
+        # TODO: confirm that no timestamp is needed or supported for tiledb.Array.create()
         tiledb.Array.create(self._uri, sch)
```
mlin, line 128: timestamps appear not yet supported for (i) Array creation and (ii) any group operation -- please correct me if I've got that wrong? @thetorpedodog @Shelnutt2

libtiledbsoma/src/soma_reader.cc
gspowley: I agree, we need to set the start and end timestamps and reopen the array with the current TileDB C++ API (we have to do this in TileDB-VCF too).

apis/python/src/tiledbsoma/options/soma_tiledb_context.py
```diff
@@ -39,4 +40,29 @@ class SOMATileDBContext:
...
+    """
+
+    write_timestamp: Optional[int] = None
+    "Timestamp applied to all array write operations."
+
+    def _tiledb_read_timestamp_arg(self) -> Optional[Tuple[int, int]]:
+        "(internal) form the read timestamp tuple arg for TileDB methods"
```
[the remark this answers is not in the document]
atolopko-czi: seems reasonable! easy enough to refactor if it feels ugly later on.

apis/python/src/tiledbsoma/options/soma_tiledb_context.py
atolopko-czi: consistency by default supports the principle of least astonishment, so I'm for it.

libtiledbsoma/src/soma_reader.cc
```diff
@@ -162,6 +185,14 @@ uint64_t SOMAReader::nnz() {
...
     }
 
+    if (timestamp_ &&
+        (timestamp_->first > 0 || timestamp_->second < UINT64_MAX)) {
+        // Use count_cells for timestamped queries for now.
+        // TODO: make optimized method filter fragment(s) on timestamp range
+        // https://github.com/TileDB-Inc/TileDB/blob/6f92d2864440cec17474f2aae22f223147828c6a/tiledb/sm/cpp_api/fragment_info.h#L314
+        return nnz_slow();
+    }
+
```
gspowley, line 195: If we enable timestamps by default, we'll need the optimized nnz to filter filter on timestamp range (TODO) or all nnz calls will be slow.

Let me know if you want and help with that.

gspowley (changes requested):
The C++ changes look good. We need to address the optimized nnz with timestamps, so all nnz calls don't take the slow path.

libtiledbsoma/src/soma_reader.cc
mlin: @gspowley OK, agree...does the TODO sound like the correct approach? iterate through the fragment info, filter out those not overlapping the read timestamp range, then proceed with the rest of the current logic using only the remaining fragments.

gspowley: Yes, iterating through the fragment info and filtering on read timestamp range sounds right. We also need to replace the `fragment_info.total_cell_num()` call, because it computes the total number of cells for all loaded fragments. We can manually compute the total_cell_num by calling `fragment_info.cell_num(index)` for each fragment that passed the timestamp filter.

mlin: @gspowley `FragmentInfo.timestamp_range()` makes it sound like one fragment can contain multiple writes with distinct timestamps. Is that true, or otherwise why would a fragment have a timestamp *range* instead of just a timestamp? My concern is the optimized method is in trouble if the desired read timestamp falls in the middle of a fragment's "timestamp range".

mlin:
@gspowley in [2cbf38f](https://github.com/single-cell-data/TileDB-SOMA/pull/749/commits/2cbf38f675daf35ba391586ed5c56f8bc3dc2ae8) I took a crack at adding timestamp-awareness into the fragment-optimized nnz. I do feel I'm in a bit over my head with the TileDB internals, so definitely appreciate if you have time to make a pass over it. I made it still punt to nnz_slow() if a fragment's timestamp range is split by one end of the read timestamp range. I'm not clear how that would arise in practice though, and therefore how to test it...

apis/python/src/tiledbsoma/options/soma_tiledb_context.py
mlin: I ran into some rough waters making the read timestamp active by default, as it makes writers unable to read their own writes (thanks @johnkerl for late-night debugging help). Hoping this doesn't blow up my argument to separate the timestamps from open(r/w), but need to noodle on it :sweat_smile:

apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -247,13 +238,10 @@ def read(
             if value_filter is not None:
                 query_condition = QueryCondition(value_filter)
 
-            sr = clib.SOMAReader(
-                self._uri,
-                name=self.__class__.__name__,
+            sr = self._soma_reader(
```
johnkerl, line 241: @mlin just a thought, to split this PR up & make it easier to iterate on, maybe have a PR that _only_ changes `sr = clib.SOMAReader` to `self._soma_reader` -- with no timestamps mods at all -- just a 'code motion' PR ...

libtiledbsoma/src/soma_reader.cc
gspowley: @mlin Right, a fragment will have a timestamp range if the fragment was created by a consolidation operation. If no consolidation operations were performed on the array, fragments will have a single timestamp (start and end of the range are the same). I agree, we will need to fall back to a slow nnz if the desired read timestamp falls in the middle of a fragment's timestamp range.

apis/python/src/tiledbsoma/dataframe.py
gspowley: @mlin, I'm happy to take pass over the optimized nnz. It sounds like we are moving that to a separate PR.

mlin:
Closing this in favor of #819, which is far simpler and cleaner now that we merged a lot of the parts separately.

PR #750, opened 2023-01-19
[python] Add GHA workflow for minimal CI jobs

atolopko-czi:
## Issue and/or context:
#723 

## Changes:
- add `python-ci-minimal.yml` GHA workflow for minimal CI jobs, invoked by pushes to feature branches
- rename `python-ci.yml` to `python-ci-full.yml`, to differentiate from the "minimal" workflow; invoked by pushes to main and for releases
- add `python-ci-single.yml`, which is a reusable workflow called by the "minimal" and "full" workflows.
- factored out linting steps into a separate job from the build/tests steps to avoid having conditional logic on those steps

## Notes for Reviewer:
Successful workflow runs:
* "minimal": https://github.com/single-cell-data/TileDB-SOMA/actions/runs/3961177942
* "full" https://github.com/single-cell-data/TileDB-SOMA/actions/runs/3961201306
As these requires some temporary hacks to trigger, our final test of the "full" workflow can only be run upon merging this PR to main. 🤞


.github/workflows/python-ci-full.yml
```diff
@@ -0,0 +1,36 @@
...
+        # os: [ubuntu-22.04, macos-12, windows-2019]
+        python-version: ['3.7', '3.8', '3.9', '3.10']
+        include:
+          - runs-on: ubuntu-22.04
+            cc: gcc-11
+            cxx: g++-11
+          - runs-on: macos-12
+            cc: gcc-11
+            cxx: g++-11
+    uses: single-cell-data/TileDB-SOMA/.github/workflows/python-ci-single.yml@main
```
thetorpedodog, line 25: I believe you should be able to call this (and below) `./.github/...` so that it refers to the code in the current revision (i.e., the same way things work now).

atolopko-czi: GHA complained when a version wasn't specified, but then the workflow file didn't yet exist on main.  Can't test "no version defaults to main" until we merge.  Let me merge this to main and then I can open a new PR to address this.

atolopko-czi: oh, I see now; need to use relative path.  Will try that on _this_ branch now. Thx!

thetorpedodog: yeah, the documentation could be a little clearer that the `./` on the beginning is what is important, and that `.github/...` and `./.github/...` are not the same thing

.github/workflows/python-ci-minimal.yml
```diff
@@ -3,9 +3,8 @@ name: TileDB-SOMA Python CI (Minimal)
 on:
   # TODO: 'pull_request' not triggering on subsequent pushes; is 'push' acceptable?
```
johnkerl, line 4: From what I googled I believe so

.github/workflows/python-ci-minimal.yml
```diff
@@ -0,0 +1,30 @@
+name: TileDB-SOMA Python CI (Minimal)
+
+on:
+  push:
```
atolopko-czi, line 4: This now differs from r-ci.yml, which is not ideal.  And the GHA console shows push- and pull_request-triggered workflow runs somewhat differently.  I'm going to test once more with pull_request...

atolopko-czi:
@johnkerl @thetorpedodog I'd say this is ready, if we want to move forward with it. I merged in the caching improvements that Paul made earlier today. 

Iever quite figured out why only `push` works to trigger Python workflows, while `pull_request` is working to trigger R workflows. Interestingly, the R workflow triggers even if just a `apis/python/**` file changes, which makes me believe it's not the `pull_request` trigger that is working, but rather the `push` trigger that also exists in the `r-ci.yml` file. It must be that the a transient merge _into_ `main` is being done to test the PR against main, and that is what's triggering the R workflow. It's a bit mind-boggling and GHA docs are not clear on the matter at all.  But they hint at such things as I've described...

Not going to spend more time on this, so we can take it or leave it as is. 😄

johnkerl:
Yes @atolopko-czi re known-broken R CI pending https://github.com/single-cell-data/TileDB-SOMA/pull/766 :)

johnkerl (approved):
@atolopko-czi thank you thank you thank you 🙏 

I know this was a long slog, and a puzzler, with lots of revs trying to untangle various complexities

But I really do apprecate it -- the more people are actively contributing to SOMA right now, the more we're using CI -- and your work here will help us _all_move faster. :D

johnkerl:
See also #693

Issue #707, opened 2023-01-12
Close old PRs/issues/etc


johnkerl:
- [ ] Close/merge old PRs
- [ ] Close old issues when possible
- [x] Remove old tags/labels

maniarathi:
@johnkerl Think we can close this?

johnkerl:
@maniarathi if you like!

It's a "never done" thing closing issues & PRs; this was just an "extra" reminder for pre-release neatening.

If you're happy with it I'm happy with it & we can go on with life which will (as always) have more closing of things :).

PR #758, opened 2023-01-19
Remove pre-main spec file

johnkerl:
## Issue and/or context:

#707 

## Changes:

Stale spec file, relevant to `main-old`

The `main` spec is at https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md


aaronwolen (approved):
Good call.

johnkerl:
Cancelling CI as this affects no code, and we should free up CI compute resources for PRs that would benefit from it.

PR #760, opened 2023-01-19
[python] Cache builds of libtiledbsoma in Python CI

thetorpedodog:
Context: #693

This caches the `build` and `dist` directories [was: `dist/lib`], which is the output from the libtiledbsoma build process. It uses the hash of all files in the `libtiledbsoma` directory (as well as `setup.py`) to ensure that when any file is changed, the cache is invalidated and the build is done anew.

---

I'm not sure how I would set this up on the R side but something analogous would be good there, too.

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -47,6 +47,14 @@ jobs:
         cache: pip
         cache-dependency-path: ./apis/python/setup.py
 
+    - name: Cache native libraries
+      uses: actions/cache@v3
+      with:
+        path: |
+          build
+          dist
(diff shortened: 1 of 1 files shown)


thetorpedodog:
This is looking good. A change where I didn't change any C++ library stuff was much faster (though it still did one build from source due to flakiness on cache setup):

https://github.com/single-cell-data/TileDB-SOMA/actions/runs/3961880845/usage

atolopko-czi (approved):

thetorpedodog:
Update: the dummy branch (#761) is also showing that when changes DO occur, the cache is not used. Given that covers both the positive and negative cases I’m going to merge so this can start saving us time immediately. R failures are unrelated.

Issue #752, opened 2023-01-19
Update to core/C++ 2.14, tiledb-r 0.18, tiledb-py 0.20


johnkerl:
All synchronized with the 2.14 core release

PR #764, opened 2023-01-19
[r] Caveat for TileDB-R dependency

johnkerl:
## Issue and/or context:

Found via group debug regarding today's CI failures -- core 2.14 is out, and `tiledb-r` 0.18.0 is out (at CRAN), but `tiledb-py` 0.20 is not yet and we haven't bumped `libtiledbsoma`'s core dependency from 2.13 to 2.14 yet. So R CI fails with `tiledb-r` writing a 2.14 array that `libtiledbsoma` can't read because it's a too-new version.

See also #752

### DIFF
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -33,6 +33,9 @@ Imports:
     Matrix,
     stats,
     bit64,
+    # We can't pin on 0.17.*, 0.18.*, etc since CRAN forbids it.  This means that whenever tiledb-r
+    # bumps, we can have CI fails in the time window between that release and our bumping
+    # libtilesoma's core-dependency version.
     tiledb (>= 0.17.0),
     arrow,
     utils,


apis/r/DESCRIPTION
```diff
@@ -33,6 +33,9 @@ Imports:
     Matrix,
     stats,
     bit64,
+    # We can't pin on 0.17.*, 0.18.*, etc since CRAN forbids it.  This means that whenever tiledb-r
+    # bumps, we can have CI fails in the time window between that release and our bumping
+    # libtilesoma's core-dependency version.
```
eddelbuettel, line 38: 1.  We are not on CRAN.  This is not a current binding constraint, and may not for some time.  So why make it an issue.

2.  One cannot leave comments in the DESCRIPTION file.  Try `R CMD check .` on _any_ package where you add something like that.  It will blow up immediately.

I suggest closing and withdrawing the PR.

eddelbuettel (changes requested):

PR #767, opened 2023-01-20
[python] Remove unnecessary steps from lint job

thetorpedodog:
Because pre-commit–based `mypy` doesn’t depend upon the packages in the installed directory, the only thing we need to install is pre-commit.

Context: #693


atolopko-czi (approved):
Simpler still, nice!

Issue #768, opened 2023-01-20
[python] ExperimentAxisQuery.get_indexer() disappeared


bkmartinjr:
The public API (method) `ExperimentAxisQuery.get_indexer` seems to have disappeared in the migration to `somacore`.  This is used by downstream code -- need to revert this deletion.

PR #769, opened 2023-01-20
[python] Add test case for issue 768 -- existence of `ExperimentAxisQuery.get_indexer()`

bkmartinjr:
## Issue and/or context:

Public API for ExperimentAxisQuery indexer was accidentally removed in the `somacore` refactoring.  See #768 

single-cell-data/SOMA#90 resolves the underlying issue.

This PR adds a unit test.

## Changes:

Add unit test of public API.

## Notes for Reviewer:

The test change will fall back on the old (private) API, as a temporary work-around until such time as `tiledbsoma` is updated to use a `somacore` release that contains this fix.


johnkerl (approved):
✨

PR #772, opened 2023-01-20
Add pandas-stubs for mypy and fix uncovered typing issues

gsakkis:
Add public type stubs for `pandas` via [pandas-stubs](https://pypi.org/project/pandas-stubs/) and fix the 9 resulting mypy errors.

### DIFF
--- a/apis/python/src/tiledbsoma/types.py
+++ b/apis/python/src/tiledbsoma/types.py
@@ -10,12 +10,12 @@ from typing_extensions import Literal
 if TYPE_CHECKING:
     NPInteger = np.integer[npt.NBitBase]
     NPFloating = np.floating[npt.NBitBase]
-    NDArray = npt.NDArray[np.number[npt.NBitBase]]
+    NPNDArray = npt.NDArray[np.number[npt.NBitBase]]
     PDSeries = pd.Series[Any]
 else:
     NPInteger = np.integer
     NPFloating = np.floating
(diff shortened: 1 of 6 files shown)


apis/python/src/tiledbsoma/types.py
```diff
@@ -11,10 +11,12 @@
     NPInteger = np.integer[npt.NBitBase]
     NPFloating = np.floating[npt.NBitBase]
     NDArray = npt.NDArray[np.number[npt.NBitBase]]
```
johnkerl, line 13: While we're here this needs to be `NPNDArray`

We already have `SparseNDArray` and `DenseNDArray` for SOMA, and this really needs the `NP` prefix like the other NumPy types already have, and as the Pandas types have the `PD` prefix

johnkerl (changes requested):
This is awesome @gsakkis -- thanks!!

One requested change which I think is important.

If you don't want to do it on this PR I'll do it on another.

apis/python/src/tiledbsoma/types.py
gsakkis: Makes sense; renamed it [here](https://github.com/single-cell-data/TileDB-SOMA/pull/772/commits/cd21d417d4c2bdc42f3263c959dfa2a1735446d4) :+1:

johnkerl (approved):
🚢

Issue #737, opened 2023-01-17
[r] Unicode attributes


johnkerl:
R counterpart to #415

eddelbuettel:
This is taken care of for us by TileDB Core and R.  The following snippet from the unit test directory was added by @johnkerl on Jan 20 based on something I had shared over slack.   So this can be closed unless I am missing something.

```r
  tbl0 <- arrow::arrow_table(foo = 1L:36L,
                             soma_joinid = 1L:36L,
                             bar = 1.1:36.1,
                             baz = c("á", "ą", "ã", "à", "å", "ä", "æ", "ç", "ć", "Ç", "í",
                                     "ë", "é", "è", "ê", "ł", "Ł", "ñ", "ń", "ó", "ô", "ò",
                                     "ö", "ø", "Ø", "ř", "š", "ś", "ş", "Š", "ú", "ü", "ý",
                                     "ź", "Ž", "Ż"),
                             schema = asch)

  sdf$write(tbl0)

  # read back the data (ignore attributes)
  expect_equivalent(
    tiledb::tiledb_array(sdf$uri, return_as = "asis")[],
    as.list(tbl0),
    ignore_attr = TRUE
  )

  # Read result should recreate the original Table
  tbl1 <- sdf$read()
  expect_true(tbl1$Equals(tbl0))

  # Slicing by foo
  tbl1 <- sdf$read(coords = list(foo = 1L:2L))
  expect_true(tbl1$Equals(tbl0$Slice(offset = 0, length = 2)))

  # Subselecting columns
  expect_error(
    sdf$read(column_names = "foobar"),
    "'column_names' must only contain valid dimension or attribute columns"
  )

  tbl1 <- sdf$read(column_names = "bar")
  expect_true(tbl1$Equals(tbl0$SelectColumns(2L)))

  # Attribute filters
  tbl1 <- sdf$read(value_filter = "bar < 5")
  expect_true(tbl1$Equals(tbl0$Filter(tbl0$bar < 5)))
```

PR #773, opened 2023-01-20
[r] Simple R UTF-8 test

johnkerl:
## Issue and/or context:

#737 

## Changes:

Simple write-and-read-back tests for UTF-8 strings in `SOMADataFrame`.

## Notes for Reviewer:

There's more we can check, of course. While doing a quick health check with core 2.14 (#770) I typed this up, and it works, and I figured, why not put up an initial PR -- ?

### DIFF
--- a/apis/r/tests/testthat/test_SOMADataFrame.R
+++ b/apis/r/tests/testthat/test_SOMADataFrame.R
@@ -1,3 +1,4 @@
+
 test_that("Basic mechanics", {
   uri <- withr::local_tempdir("soma-indexed-dataframe")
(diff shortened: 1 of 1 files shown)


eddelbuettel:
If you care there are more characters here:

https://github.com/TileDB-Inc/TileDB-R/blob/a7b8e229c6530c82d47b0e61dabc4f8b2a2b1979/inst/tinytest/test_querycondition.R#L365-L368

(I actually copied those from a real-life application of mine but that is a whole different story ....)

eddelbuettel (approved):

PR #775, opened 2023-01-20
[python] Change _CachedElement._TdbInfo.type from str to type

### LINKED ISSUE #774: [Python] Non-overlapping equality check
After somehow[1] instructing `mypy` to take into account the `tiledb-py` type annotations , it finds an apparently actual error at [this line](https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/python/src/tiledbsoma/collection.py#L266), in particular the `tdb.type == cached.tdb.type` condition:
> apis/python/src/tiledbsoma/collection.py:267: error: Non-overlapping equality check (left operand type: "type", right operand type: "str")  [comparison-overlap]

Running the tests with the debugger reveals that mypy is indeed correct and this condition is always False:
```py
(Pdb++) tdb.type
<class 'tiledb.libtiledb.Array'>
(Pdb++) cached.tdb.type
'array'
(Pdb++) tdb.type == cached.tdb.type
False
(Pdb++) 
```

[1] by creating manually an empty `py.typed` file at the tiledb installation dir

gsakkis:
Fixes #774

### DIFF
--- a/apis/python/src/tiledbsoma/collection.py
+++ b/apis/python/src/tiledbsoma/collection.py
@@ -40,7 +40,7 @@ class _CachedElement:
 
     @dataclass
     class _TdbInfo:
-        type: str
+        type: Type[Any]
         uri: str
         name: str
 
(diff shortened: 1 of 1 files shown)


johnkerl (approved):
✨

PR #779, opened 2023-01-20
[python] Some low-hanging fruit for code coverage

johnkerl:
## Issue and/or context:

#778

## Changes:

Just some quick end-of-week neatens on some low-hanging fruit, e.g. functions no longer having callsites. There's more to do beyond this particular quick/easy PR.

PR #781, opened 2023-01-20
[python] A bit more deadstrip

johnkerl:
## Issue and/or context:

Follow-on to #779 

## Changes:

Just a couple missed items. More to come! :)


bkmartinjr (commented):
Suggest updating `devtools/README.md` to indicate that it is tools for SOMA developers

bkmartinjr (commented):
Another possible bit of dead code:   _show_metadata method (in two files)

johnkerl:
> Another possible bit of dead code: _show_metadata method (in two files)

@bkmartinjr thanks, added! But again, as in the description -- there's no claim to completeness here. If we look for more we'll find more. I just wanted to finish up with `util_tiledb.py` from #779, and fix something incomplete I did there which was rename `tools` to `devtools` in one place but not all.

bkmartinjr (approved):

Issue #740, opened 2023-01-17
[c++] SOMA open-at-timestamp support


johnkerl:

mlin:
@gspowley In #788 I extracted just the C++ bits of #740...I'd be glad if you're able to take #788 from there, including tossing it if you see fit :smiley_cat:

PR #788, opened 2023-01-22
[c++] SOMAReader open-at-timestamp support

mlin:
#740

Isolated just the C++ WIP from #749. Still needs work on making our optimized nnz method timestamp-aware, and testing.


gspowley:
@mlin One case we did not consider so far was how to handle consolidated fragments, which can contain duplicates. I added code to fall back to counting cells if we find a relevant, consolidated fragment.

I also moved the check for `allows_dups` into `nnz_slow`, as your comment mentioned. It's the right thing to do because if we are counting cells to compute nnz, we should not count duplicates. This is a good use case for a TileDB read query that allows disabling duplicates.

New `nnz` unit tests were added to cover all of the permutations.

gspowley (approved):

libtiledbsoma/include/tiledbsoma/soma_reader.h
```diff
@@ -320,6 +329,9 @@ class SOMAReader {
 
     // True if the query was submitted
     bool submitted_ = false;
+
+    // Unoptimized method for computing nnz() (issue count_cells query)
```
johnkerl, line 333:
```
    // Unoptimized method for computing nnz() (issues a `count_cells` query)
```

johnkerl (approved):
LGTM, but I defer to @gspowley

PR #790, opened 2023-01-23
Simplify to_tiledb_supported_array_type

gsakkis:
[functools.singledispatch](https://docs.python.org/3/library/functools.html#functools.singledispatch) has its usages as it can help simplify conditional logic based on the type of the first parameter of a function. In case of `to_tiledb_supported_array_type` however, the logic becomes easier to follow IMO without `singledispatch` (which as @johnkerl [points out](https://github.com/single-cell-data/TileDB-SOMA/commit/8e0a7ea751a58386d5d62d06c0386fa2fd5f2f45#diff-294cb93189555455defa256cc57a6654cbf09ae4fa06608e51051f031e3e6c22R26-R29), is a relatively obscure feature).

### DIFF
--- a/apis/python/src/tiledbsoma/util_tiledb.py
+++ b/apis/python/src/tiledbsoma/util_tiledb.py
@@ -1,11 +1,10 @@
-import functools
 from typing import TypeVar, cast
 
 import numpy as np
-import pandas as pd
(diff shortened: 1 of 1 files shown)


johnkerl (approved):
Thanks @gsakkis !!

PR #796, opened 2023-01-24
[r] Prevent one test script from failing under `rcmdcheck`

eddelbuettel:
#### Issue and/or context:

`rcmdcheck` breaks one of our `testthat` scripts

#### Changes:

Bail when we detect we test under `rcmdcheck`.  

#### Notes for Reviewer:

See https://app.shortcut.com/tiledb-inc/story/24829/avoid-heisen-test-bug-under-rcmdcheck

### DIFF
--- a/apis/r/tests/testthat/test-SOMAMeasurement.R
+++ b/apis/r/tests/testthat/test-SOMAMeasurement.R
@@ -3,6 +3,10 @@ test_that("Basic mechanics", {
   # TODO: Determine why this fails only on linux w/ rcmdcheck::rcmdcheck()
   testthat::skip_on_covr()
 
+  ## this variable appears to be set under rcmdcheck, but not R CMD CHECK
+  ## as the testscript appears to fail here for rcmdcheck only (why?), we bail
+  testthat::skip_if(Sys.getenv("CALLR_IS_RUNNING", "") != "")
+
   uri <- withr::local_tempdir("soma-ms")
 
   measurement <- SOMAMeasurement$new(uri)


apis/r/tests/testthat/test-SOMAMeasurement.R
```diff
@@ -3,6 +3,10 @@ test_that("Basic mechanics", {
   # TODO: Determine why this fails only on linux w/ rcmdcheck::rcmdcheck()
   testthat::skip_on_covr()
 
+  ## this variable appears to be set under rcmdcheck, but not R CMD CHECK
+  ## as the testscript appears to fail here for rcmdcheck only (why?), we bail
+  testthat::skip_if(Sys.getenv("CALLR_IS_RUNNING", "") != "")
```
aaronwolen, line 8: Should we delete the `skip_on_covr()` above since you're avoiding the same issue with a more precise marker?

aaronwolen (approved):
One question/suggestion but thanks for looking into this!

mojaveazure (approved):

apis/r/tests/testthat/test-SOMAMeasurement.R
eddelbuettel: I wondered about that same question.  "Probably".  But having been burned by CI and its friends I try to be as minimal as possible, at least on a first attempt.

Let me make that change and see how it fares.

eddelbuettel: (The bigger question is course still by the combination of `testthat` and `rcmdcheck` decides to be mean to us.  The error makes no sense to me but it seems related to how you construct the test that.  Then again we do just that in other test files.  Very very strange.)

eddelbuettel: Coverage still blows up.  That sucks.  So I revert that part.

Issue #643, opened 2023-01-04
[python] Feature-parity gap analysis


johnkerl:
* Between `main-old` and `main`
* ~Between Python and R -- @mlin good POC for R context~ (now tracked in #839 )

Perhaps best to make a cloud-shared spreadsheet we can put X's  in to check off as done -- ? Whatever works best.

maniarathi:
Splitting this issue into two. This issue is focused on performing the gap analysis between `main` and `main-old` in Python. A separate ticket, #839 , captures the work to perform the gap analysis between the Python and R packages.

nguyenv:
The gap analysis between Python's `main` and `old-main` can be found at [this spreadsheet](https://docs.google.com/spreadsheets/d/1vmS5YGJP2_c3PtnCRveHwK3sk-h4Ahk7uu42ZVk1WaI/edit#gid=0). We compared the branches in two ways:

  1. Running through the example notebooks locally and recording necessary changes that needed to made in order to run under main
  2. Recording all the features listed in the tiledbsoma release notes for old-main and marking whether or not they are still present in main
  
There were some minor convience functions that had not been migrated yet like `add_X_layer` though this has since been added. Though overall, no major gaps were found.

PR #797, opened 2023-01-24
[python] Do more is-Arrow checks on create and write

johnkerl:
## Issue and/or context:

#643 

## Changes:

Doing feature-parity gap analysis with one of our computational biologists, @ckrilow , we found that trying to write back the results of `scanpy` analyses to SOMA storage results in highly confusing error messages about missing attrs and so on:

```
  File "/Users/johnkerl/git/single-cell-data/TileDB-SOMA/apis/python/src/tiledbsoma/sparse_nd_array.py", line 72, in create
    if not pa.types.is_primitive(type):
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pyarrow/types.py", line 550, in is_primitive
    return lib._is_primitive(t.id)
AttributeError: 'numpy.dtype[float32]' object has no attribute 'id'
```

```
  File "/Users/johnkerl/git/single-cell-data/TileDB-SOMA/apis/python/src/tiledbsoma/sparse_nd_array.py", line 303, in write
    raise TypeError("Unsupported Arrow type")
TypeError: Unsupported Arrow type <------ Note: supplied type was not _any_ Arrow type ...
```

By doing more proactive type-checking we can provide a useful/informative error message that helps the user self-resolve the typing issues more quickly.

## Note to reviewer:

One might counter that if we're going to run-time type-check _some_ args to _some_ functions/methods, we should run-time type-check _all_ args to _all_ functions/methods. That proposal is not without merit. However, since Python-using computational biologists are particularly likely to be reading/writing `scanpy` analysis results, while SOMA's native type system is Arrow, there is a _particularly_ wide and perilous error surface for people naïvely trying to write non-Arrow data. This surface  warrants an extra level of user guidance.

### DIFF
--- a/apis/python/src/tiledbsoma/util.py
+++ b/apis/python/src/tiledbsoma/util.py
@@ -195,7 +195,7 @@ def dense_index_to_shape(
 
 def check_type(
     name: str,
-    actual_value: Type[Any],
+    actual_type: Type[Any],
     expected_types: Tuple[Type[Any], ...],
 ) -> None:
     """
@@ -213,7 +213,8 @@ def check_type(
     name: str
         Name of the argument in question, as used by the caller.
 
-    actual_value: Any
+    XXX CHANGE ME
+    actual_type: Any
         Value of the argument. Note that it would be much easier for us to pass the _type_ of the
         value rather than the full value. Unfortunately, we cannot: within the SOMA API we have
         argument names like `type`. Then the caller would have to pass us `type(type)`. But this is
@@ -227,13 +228,13 @@ def check_type(
     -------
     None. Raises `TypeError` on mismatch.
     """
-    # if type(actual_value) not in expected_types:
-    if not isinstance(actual_value, expected_types):
+    # if not isinstance(actual_value, expected_types):
+    if actual_type not in expected_types:
         if len(expected_types) == 1:
             raise TypeError(
-                f"expected {name} argument to be of type {expected_types[0]}; got {type(actual_value)}"
+                f"expected {name} argument to be of type {expected_types[0]}; got {actual_type}"
             )
         else:
             raise TypeError(
-                f"expected {name} argument to be one of {expected_types!r}; got {type(actual_value)}"
+                f"expected {name} argument to be one of {expected_types!r}; got {actual_type}"
             )
--- a/apis/python/src/tiledbsoma/util.py
+++ b/apis/python/src/tiledbsoma/util.py
@@ -1,7 +1,7 @@
 import pathlib
 import time
 import urllib.parse
-from typing import List, Optional, Tuple, Union
+from typing import Any, List, Optional, Sequence, Tuple, Union
 
 import somacore
 from somacore import options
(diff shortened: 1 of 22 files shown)


apis/python/src/tiledbsoma/util.py
```diff
@@ -191,3 +191,48 @@ def dense_index_to_shape(
         return stop - start + 1
 
     raise TypeError("coordinates must be tuple of int or slice")
+
+
+def check_type(
+    name: str,
+    actual_value: Any,
```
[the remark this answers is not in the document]
johnkerl: Clever indeed ... :)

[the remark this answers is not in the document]
johnkerl: @thetorpedodog thanks, this is super-educational, but isn't something I want to use here ... preferring the "level-down" syntax as I think it solves this problem adequately

johnkerl:
@thetorpedodog ready for round 2 🙏

apis/python/src/tiledbsoma/util.py
```diff
@@ -191,3 +191,50 @@ def dense_index_to_shape(
...
+
+    expected_type: Any
+        Things like `pyarrow.Tensor`, etc.
+
+    Returns
+    -------
+    None. Raises `TypeError` on mismatch.
+    """
+    # if not isinstance(actual_value, expected_types):
+    if actual_type not in expected_types:
```
thetorpedodog, line 232: `isinstance` instead?

apis/python/src/tiledbsoma/util.py
```diff
@@ -191,3 +191,50 @@ def dense_index_to_shape(
...
+        if len(expected_types) == 1:
+            raise TypeError(
+                f"expected {name} argument to be of type {expected_types[0]}; got {actual_type}"
+            )
+        else:
```
thetorpedodog, line 237: nit: no need for else after a raise

apis/python/src/tiledbsoma/util.py
```diff
@@ -191,3 +191,50 @@ def dense_index_to_shape(
...
+        argument names like `type`. Then the caller would have to pass us `type(type)`. But this is
+        a name clash between Python's `type` as a function name, and SOMA's `type` as an argument
+        name.
+
+    expected_type: Any
+        Things like `pyarrow.Tensor`, etc.
+
+    Returns
+    -------
+    None. Raises `TypeError` on mismatch.
```
thetorpedodog, line 229: I would make the first line of this describe its function more precisely:

`Verifies the type of an argument or produces a useful error message.`

I also think that everything including and after `Parameters` can be deleted—the names are clear, the function is internal-only, and the implementation is a number of lines you can count on one hand.

(also consider checking margins so things wrap around 88 chars to match black’s formatting of the rest of the code)

apis/python/src/tiledbsoma/util.py
```diff
@@ -191,3 +191,50 @@ def dense_index_to_shape(
         return stop - start + 1
 
     raise TypeError("coordinates must be tuple of int or slice")
+
+
+def check_type(
+    name: str,
+    actual_type: Type[Any],
```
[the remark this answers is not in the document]
johnkerl: @thetorpedodog and I prefer to pass type of thing. I know most everything should be pass by reference, but, I don't want any chance of overhead of value-copies. I prefer it this way.

apis/python/src/tiledbsoma/util.py
johnkerl: Only works if I pass values, and I'm not passing values.

`isinstance(type, types)` fails unit test.

apis/python/src/tiledbsoma/util.py
johnkerl: Sorry this is not ready for review -- this entire doc block with `XXX CHANGE ME` needs updating.

I should have left this in draft, sorry.

apis/python/src/tiledbsoma/util.py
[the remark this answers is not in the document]
thetorpedodog: Everything in Python is a pointer; no matter what you pass in a function call—int, bytearray, table—it’s always passing a reference to some object.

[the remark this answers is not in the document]
johnkerl: @thetorpedodog change of heart, I give in to help smooth this along.

Also, if high-memory-footprint things were being passed by value someaccidentallyhow, it would have been causing issues elsewhere in the callstack already.

apis/python/src/tiledbsoma/util.py
johnkerl: ibidx

johnkerl:
@thetorpedodog ready for round three 🙏

gsakkis:
Alternative idea: add [typeguard](https://typeguard.readthedocs.io/en/latest/) as a dependency to do the job and sprinkle `check_type`/`check_argument_types` calls and/or `typechecked` decorators at will.

thetorpedodog (approved):

Issue #801, opened 2023-01-24
[r/python] SOMA objects from the future do not raise an error


bkmartinjr:
If I create a TileDB object which looks like a SOMA object, but has a `soma_encoding_version` from the future (ie., not "1"), I would expect an error to be raised.

However, the code in `_factory.py` and elsewhere does not raise an error, and treats the object as if it were supported.

Places we should consider a check:
* `_factory.py` (will be fixed by #802)
* possibly anywhere we do an Array/Group open

jp-dark:
Already implemented.

PR #802, opened 2023-01-24
More unit tests, plus partial fix for #801

bkmartinjr:
## Issue and/or context:

We need to increase unit tests

## Changes:
* Improved unit tests for SparseNDArray
* New unit tests for factory.py 
* Fixed a couple of minor bugs related to error handling:
    * Partial fix for #801 - check encoding version in reification. We still need to check it for other code paths
    * Correctly handle the case of missing object metadata (use `meta.get(key, None)` rather than `meta[key]`)

## Notes for Reviewer:
None

### DIFF
--- a/apis/python/src/tiledbsoma/factory.py
+++ b/apis/python/src/tiledbsoma/factory.py
@@ -64,11 +64,11 @@ def _construct_member(
         if object_type == "array":
             with tiledb.open(member_uri, ctx=context.tiledb_ctx) as A:
                 spec_name = A.meta.get(SOMA_OBJECT_TYPE_METADATA_KEY, None)
-                encoding_version = A.meta.get(SOMA_ENCODING_VERSION_METADATA_KEY, None)
+                encoding_version = A.meta.get(SOMA_ENCODING_VERSION_METADATA_KEY)
         elif object_type == "group":
             with tiledb.Group(member_uri, mode="r", ctx=context.tiledb_ctx) as G:
                 spec_name = G.meta.get(SOMA_OBJECT_TYPE_METADATA_KEY, None)
-                encoding_version = G.meta.get(SOMA_ENCODING_VERSION_METADATA_KEY, None)
+                encoding_version = G.meta.get(SOMA_ENCODING_VERSION_METADATA_KEY)
         else:
             return None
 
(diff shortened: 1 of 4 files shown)


apis/python/src/tiledbsoma/factory.py
```diff
@@ -58,10 +63,12 @@ def _construct_member(
     try:
         if object_type == "array":
             with tiledb.open(member_uri, ctx=context.tiledb_ctx) as A:
-                spec_name = A.meta[SOMA_OBJECT_TYPE_METADATA_KEY]
+                spec_name = A.meta.get(SOMA_OBJECT_TYPE_METADATA_KEY, None)
```
thetorpedodog, line 66: nit: `, None)` not needed; it’s already the default

apis/python/src/tiledbsoma/factory.py
```diff
@@ -70,6 +77,10 @@ def _construct_member(
 
     if spec_name is None:
         raise SOMAError("internal error: spec_name was not found")
+    if encoding_version is None:
+        raise SOMAError("internal error: encoding_version not found")
+    if encoding_version != SOMA_ENCODING_VERSION:
+        raise SOMAError("Unsupported SOMA object encoding version")
```
thetorpedodog, line 83: Should this be a ValueError?

apis/python/tests/test_sparse_nd_array.py
```diff
@@ -754,3 +756,63 @@ def test_sparse_nd_array_table_slicing(tmp_path, io, write_format, read_format):
...
+    with pytest.raises(ValueError):
+        next(a.read((slice(10, -2),)).tables())
+    with pytest.raises(TypeError):
+        next(a.read("hi").tables())
+    with pytest.raises(ValueError):
+        next(a.read(coords=()).tables())
+    with pytest.raises(ValueError):
+        next(a.read(coords=[]).tables())
+    with pytest.raises(ValueError):
+        next(a.read(coords=((slice(None), slice(None)))).tables())
```
thetorpedodog, line 818: https://youtube.com/clip/Ugkxi9EpXTo55k3mOGXWfXMEDJB0uYbOiBTO

thetorpedodog (approved):
one minor thing (ok if left as is), one question. looks good!

apis/python/src/tiledbsoma/factory.py
bkmartinjr: Maybe...I'm a bit unclear on our exception UX approach with code like this.  I'll change for now, as we definitely have some work to do in this area.

apis/python/tests/test_sparse_nd_array.py
bkmartinjr: yeah, will parameterized - got a bit carried away there!

Issue #806, opened 2023-01-25
[r/python] Package dependencies need max version for `tiledb` dependency


bkmartinjr:
The Python dependency on `tiledb`, used for the write path, must have a maximum version specified.  The max version must be a version that is compatible with the SOMAReader-linked version used for the read path.

In other words, need to make sure that you can't write data which is then not readable due to version skew.

[x] Python #807
[ ] R #812

johnkerl:
Per @maniarathi , the R ecosystem does not actually make it possible to do the pinning easily. Instead, at a later point in time, the long term solution (C++ pinning) will be prioritized.

PR #807, opened 2023-01-25
[python] Pin to specific TileDB core major+minor

johnkerl:
## Issue and/or context:

#806 

## Notes for Reviewer:

I'm looking for feedback from @Shelnutt2 @ihnorton @eddelbuettel on whether this approach is sound or not.

This is a Python-only PR -- given @eddelbuettel's notes below, it'll take some time for me to digest/incorporate this into my fledgling knowledge of R. There will be a separate R PR.

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -111,7 +111,7 @@ setuptools.setup(
         "scanpy",
         "scipy",
         "somacore==0.0.0a7",
-        "tiledb>=0.20.0,<0.21.0",
+        "tiledb==0.20.*",
         "typing-extensions",  # Note "-" even though `import typing_extensions`
     ],
     extras_require={
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -111,7 +111,7 @@ setuptools.setup(
         "scanpy",
         "scipy",
         "somacore==0.0.0a7",
-        "tiledb>=0.20.0",
+        "tiledb==0.20.*",
         "typing-extensions",  # Note "-" even though `import typing_extensions`
     ],
     extras_require={
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -111,7 +111,7 @@ setuptools.setup(
         "scanpy",
         "scipy",
         "somacore==0.0.0a7",
-        "tiledb==0.20.*",
+        "tiledb>=0.20.0,<0.21.0",
         "typing-extensions",  # Note "-" even though `import typing_extensions`
     ],
     extras_require={
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -33,7 +33,7 @@ Imports:
     Matrix,
     stats,
     bit64,
-    tiledb (>= 0.18.0),
+    tiledb (>= 0.18.0, < 0.19.0),
     arrow,
     utils,
     fs,
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -33,7 +33,7 @@ Imports:
     Matrix,
     stats,
     bit64,
-    tiledb (== 0.18),
+    tiledb (>= 0.18.0),
     arrow,
     utils,
     fs,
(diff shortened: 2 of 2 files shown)


apis/r/DESCRIPTION
```diff
@@ -33,7 +33,7 @@ Imports:
     Matrix,
     stats,
     bit64,
-    tiledb (>= 0.18.0),
+    tiledb (== 0.18),
```
eddelbuettel, line 36: Can't do in `DESCRIPTION`.  

```sh
[... earlier lines from package check omitted ... ]
✔  checking package namespace information ...
E  checking package dependencies (1.7s)
   Package required and available but unsuitable version: ‘tiledb’
   
   See section ‘The DESCRIPTION file’ in the ‘Writing R Extensions’
   manual.
   
   See
     ‘/tmp/file2fc4054d83de9/tiledbsoma.Rcheck/00check.log’
   for details.
   
   
$ grep " tiledb " DESCRIPTION 
    tiledb (== 0.18.0),
$ 
```


Could approximate in `configure` and error if we wanted to.

johnkerl: @eddelbuettel @aaronwolen @mojaveazure if anyone has priors on how to approximate in `configure` I'd be grateful. (As a non-R-expert, I'm googling now ...)

eddelbuettel: Here is what I just did in tiledb-r to see 'am I at 2.15.0 or not' (which has deprecation implications, so i 'silence' compilation a little).  Recall the joker we hold that the script `configure` can be anything that runs when in mode 0755 so could be `/bin/sh` (but not `/bin/bash`, do.not.get.me.started....).  In this case it is `autoconf` but again that is _not_ a requirement:

```sh
## If we have pkg-config, use it to test minimal version
if test x"${have_pkg_config}" != x"no"; then
    AC_MSG_CHECKING([for pkg-config checking TileDB version])
    if pkg-config --atleast-version=2.0.0 tiledb; then
        AC_MSG_RESULT([>= 2.0.0])
    else
        AC_MSG_RESULT([insufficient])
        AC_MSG_ERROR([TileDB Embedded 2.0.0 or later is required.])
    fi
    AC_MSG_CHECKING([for pkg-config checking TileDB deprecation silencer])
    if pkg-config --atleast-version=2.15.0 tiledb; then
        ## For TileDB 2.15 or newer are fine
        AC_MSG_RESULT([not required])
        AC_SUBST([TILEDB_SILENT_BUILD], "")
    else
        ## For TileDB older than 2.15 we use this to silence a deprecation warning
        AC_MSG_RESULT([enabled])
        AC_SUBST([TILEDB_SILENT_BUILD], "-DTILEDB_SILENT_BUILD")
    fi
fi
```

eddelbuettel: And I should have checked first we already _do_ have `configure` as a shell script in `apis/r`.

eddelbuettel: And I was absent-minded.  `pkg-config` is to check TileDB Embedded.  You want to check the _R_ package.  For that we would use something akin to

```sh
$ Rscript -e 'packageVersion("tiledb") == "0.18.0"'          # I am at 0.18.0.1 even though CRAN is at 0.18.0
[1] FALSE
$ Rscript -e 'packageVersion("tiledb") >= "0.18.0" && packageVersion("tiledb") < "0.19.0"'
[1] TRUE
$ 
```

eddelbuettel: Or even

```sh
$ Rscript -e 'tiledb::tiledb_version(TRUE) == "2.15.0"'
[1] TRUE
$ 
```

eddelbuettel (approved):

apis/python/setup.py
```diff
@@ -111,7 +111,7 @@ def run(self):
         "scanpy",
         "scipy",
         "somacore==0.0.0a7",
-        "tiledb>=0.20.0",
+        "tiledb>=0.20.0,<0.21.0",
```
thetorpedodog, line 114: You can use the [`~=` operator or `*` matcher](https://peps.python.org/pep-0440/#compatible-release) here:

`tiledb~=0.20.0` or `tiledb==0.20.*`

PR #808, opened 2023-01-25
Import somacore v0.0.0a7, without making any functional changes.

thetorpedodog:
This change imports the new version of somacore, but does not actually implement any of the new APIs specified by it.

- Because `create` and `open` are now class methods with different semantics, the existing methods are renamed `create_legacy` and `open_legacy`.
- Duplicate methods (reimplementing the context manager protocol) have been removed and are delegated to somacore.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,6 +13,9 @@ repos:
     rev: v0.991
     hooks:
     - id: mypy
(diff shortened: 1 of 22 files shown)


bkmartinjr (approved):

thetorpedodog:
Also a note: I am planning on getting the rest of this, wherein the new stuff is actually implemented, out for review today. (At least the `create`/`open` part; `add_new_X` will be another followup after that.)

PR #812, opened 2023-01-25
[r] Pin to specific TileDB-R major+minor

johnkerl:
## Issue and/or context:

#806, R version

## Notes for Reviewer:

Taken from @eddelbuettel 's excellent advice on #807

### DIFF
--- a/apis/r/configure
+++ b/apis/r/configure
@@ -43,8 +43,8 @@ else
 fi
 
 ## check for tiledb-r 0.18.*. In DESCRIPTION we can pin only a minimum version, not a maximum.
-tiledb_r_ok=`Rscript -e 'packageVersion("tiledb") >= "0.18.0" && packageVersion("tiledb") < "0.19.0"'`
-if expr "$tiledb_r_ok" : '.*FALSE.*'; then
+Rscript -e 'q(save="no", status=if(packageVersion("tiledb") >= "0.18.0" && packageVersion("tiledb") < "0.19.0") FALSE else TRUE)'
+if [ $? -ne 0 ]; then
     actual=`Rscript -e 'packageVersion("tiledb")'`
     echo "** need tiledb-r 0.18.*; got $actual -- exiting"
     exit 1
--- a/apis/r/configure
+++ b/apis/r/configure
@@ -43,9 +43,9 @@ else
 fi
 
 ## check for tiledb-r 0.18.*. In DESCRIPTION we can pin only a minimum version, not a maximum.
-Rscript -e 'q(save="no", status=if(packageVersion("tiledb") >= "0.18.0" && packageVersion("tiledb") < "0.19.0") FALSE else TRUE)'
+${R_HOME}/bin/Rscript -e 'q(save="no", status=if(packageVersion("tiledb") >= "0.18.0" && packageVersion("tiledb") < "0.19.0") FALSE else TRUE)'
 if [ $? -ne 0 ]; then
-    actual=`Rscript -e 'packageVersion("tiledb")'`
+    actual=`${R_HOME}/bin/Rscript -e 'cat(packageVersion("tiledb"))'`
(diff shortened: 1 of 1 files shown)


eddelbuettel:
You can return the boolean result via the `q()` function and it `status` argument, it is a little more solid than grepping for the printed status.   The only use I found on my machine with a minimal search is again done from `configure.ac` but you can generalise to `sh`.

```sh
## Check if R has RcppMsgPack
$("${R_HOME}/bin/Rscript" --vanilla -e 'hasPkg <- "RcppMsgPack" %in% rownames(installed.packages()); q(save="no", status=if (hasPkg) packageVersion("RcppMsgPack") >= "0.2.0" else FALSE)')
if test x"$?" == x"1"; then
    AC_MSG_RESULT([yes])
    msgpackincdir=$("${R_HOME}/bin/Rscript" --vanilla -e 'cat(system.file("include", package="RcppMsgPack"))')
    msgpack_cxxflags="-I${msgpackincdir} -DHAVE_MSGPACK"
    AC_MSG_NOTICE([Found RcppMsgPack, using '${msgpack_cxxflags}'])
else
    AC_MSG_RESULT([no])
    AC_MSG_NOTICE([Install (optional) RcppMsgPack (>= 0.2.0) from CRAN via 'install.packages("RcppMsgPack")'])
fi
```

Deep down I am not sure I like the hard(er) pinning.   Could we settle for a (stern) warning but not an error?

johnkerl:
> Deep down I am not sure I like the hard(er) pinning. Could we settle for a (stern) warning but not an error?

I'm here solely to implement @bkmartinjr 's ask from #806.

@bkmartinjr ?

bkmartinjr:
> Deep down I am not sure I like the hard(er) pinning. Could we settle for a (stern) warning but not an error?

How does that warning, however stern, help the user in the common case where the writer & reader paths are completely independent?

At least in our use case (cell census), we expect that when we install `tiledbsoma`, we will always end up in an environment where data we write is readable.  That is not true if the user installs it into an existing Python environment where there is a TileDB version "from the future" (ie, more recent than the version linked into SOMA).  We need some way to make sure that the writer version <= reader version.

I guess we could layer more strict dependency requirements in our own cell-census package (ie., specify both tiledbsoma and tiledb version requirements), but that seems like a hack.

The real solution has already been pointed out by @Shelnutt2 - we need a single version of TileDB in the package.

johnkerl:
@eddelbuettel I don't get it either :(

eddelbuettel:
And unless one does extra tricks one does not see the install failure log :-/

eddelbuettel:
Did another full run here in a Ubuntu 22.04 container, no issues.  I now tried to add some debugging logic, we will see if that helps.

eddelbuettel:
So it was spurious.  Doh.

eddelbuettel:
Got it.  It's standard R fare:

image

eddelbuettel:
@johnkerl It's repaired.

johnkerl:
> @johnkerl It's repaired.

Thanks @eddelbuettel !! :) 

Do you want for us to merge this or no? I'm sensitive to the concerns articulated on #806, as well as to your own concerns -- and also I know we'll ultimately address this issue once everything's going through `libtiledbsoma` ...

mojaveazure:
I don't think this PR actually solves #806; `configure` only affects the package during installation; if tiledbsoma-r requires tiledb-r `A.B` but tiledb-r `X.Y` (where `X.Y` > `A.B`) already exists and is installed, a user could:
- downgrade to tiledb-r `A.B`
- pass the requirements in `configure` and install tiledbsoma-r
- upgrade to tiledb-r `X.Y`
- successfully `library(tiledbsoma)` because `configure` isn't affecting the package anymore

johnkerl:
Group decision earlier this week was that this doesn't work as intended in the R ecosystem. Instead, at a later point in time, the long term solution (C++ pinning) will be prioritized.

PR #814, opened 2023-01-26
[python] Generalize `tiledbsoma.io.add_X_layer` to `obsm`, `varp`, etc.

johnkerl:
## Issue and/or context:

#643 

Based on computational-biologist feedback comparing `main-old` to `main`

### DIFF
--- a/apis/python/src/tiledbsoma/io.py
+++ b/apis/python/src/tiledbsoma/io.py
@@ -504,14 +504,26 @@ def add_matrix_to_collection(
 
     Use `ingest_mode="resume"` to not error out if the schema already exists.
     """
+
+    if collection_name not in exp.ms[measurement_name]:
+        # E.g. this is the first obsm matrix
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/io.py
```diff
@@ -477,20 +477,43 @@ def add_X_layer(
...
+    scanpy.pp.log1p, etc.
+
+    Use `ingest_mode="resume"` to not error out if the schema already exists.
+    """
+    cls = (
+        DenseNDArray
+        if isinstance(matrix_data, (np.ndarray, h5py.Dataset))
+        else SparseNDArray
+    )
+    collection = cast(Collection, exp.ms[measurement_name][collection_name])
```
johnkerl, line 512: needs some handling for the case where this is the _first_ `obsm` etc -- this code (as currently written) only handles adding _more_ when the collection already exists

PR #817, opened 2023-01-26
Move test data to `testdata` folder.

thetorpedodog:
Using `anndata` as the folder name confuses tools like mypy who think that it's a Python package. Making it `testdata` ensures that we don't clash with package names and reduces the amount of manual configuration we have to do.


gsakkis (approved):
+1 for resolving the name ambiguity, with a minor preference to `tests/data` instead of `testdata`

thetorpedodog:
> +1 for resolving the name ambiguity, with a minor preference to `tests/data` instead of `testdata`

I actually think this is a good idea but in the interest of avoiding additional work on this change I'm going to submit as-is. I would welcome a followup that does this if anybody (i.e., not me) finds it useful to do so.

Issue #708, opened 2023-01-12
[r/python] Ensure that all public APIs have docstrings/roxygen/etc including lifecycle-maturity tags


johnkerl:
Not yet to worry about formatting for `readthedocs` or other tools -- rather, here just to focus on whether all APIs have available information accessible from on-line help.

See https://github.com/single-cell-data/SOMA/pull/47/files for the list of values

- [x] Python #818
- [x] R #829

johnkerl:
Appears to be a duplicate of #602

PR #818, opened 2023-01-26
Add lifecycle maturity tags to public API docstrings (Python)

atolopko-czi:
## Issue and/or context:
#708 

## Changes:
- Added to all publicly exported classes and their methods, including constructor
- Added to top-level functions in io, logging, general_utilties modules
- Not added to @property methods
- Not added to exceptions
- All are "experimental" maturity for now

Unrelated:
- Removed dead code: `TileDBPlatformConfig` class
- No longer export `TileDBCreateOptions` or `MetadataMapping` (they're for internal use only)

## Notes for Reviewer:
If this is approved, will tackle R changes in a separate PR.

### DIFF
--- a/apis/python/src/tiledbsoma/__init__.py
+++ b/apis/python/src/tiledbsoma/__init__.py
@@ -12,7 +12,6 @@ from .general_utilities import (
     get_storage_engine,
 )
 from .measurement import Measurement
-from .metadata_mapping import MetadataMapping
 from .sparse_nd_array import SparseNDArray
 from .tiledb_array import TileDBArray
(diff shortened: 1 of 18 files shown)


apis/python/src/tiledbsoma/__init__.py
```diff
@@ -35,6 +35,6 @@
     "SOMAError",
     "DataFrame",
     "Measurement",
-    "MetadataMapping",
+    "MetadataMapping",  # TODO: should this be exported?
```
johnkerl, line 38: tbh i think not

johnkerl (approved):
🚢

PR #819, opened 2023-01-26
[python] Apply timestamps through SOMATileDBContext

mlin:
#540

Add `read_timestamp` (and `read_timestamp_start`) properties to `SOMATileDBContext`. Then, feed these through to all array read and write operations (including the C++ `SOMAReader`).

(The context provides a slot to set the write timestamp too, but only for internal testing purposes right now, pending discussion of some corner cases.)


mlin:
@johnkerl @gspowley @atolopko-czi @thetorpedodog @bkmartinjr At last, thanks to all your help, this is reduced to a small diff that's mostly tests :smile_cat: 

A little bit of reconciliation with the new open() API will be needed, but this is now so small that I don't think it matters too much which one we merge first. Thoughts?

Issue #638, opened 2023-01-03
[python] Restructure object opening/creation protocols


johnkerl:
Implementation work following spec PR https://github.com/single-cell-data/SOMA/pull/96.

Tracking issue: #638

* Implement `collection.add(...)` to allow building new collections from the top down.
* Rework `open`/`create` operations to make using the API look more like this :
```
  new_df = tiledbsoma.DataFrame.create("some/uri", ...)
  some_coll = tiledbsoma.SOMACollection.open("existing/uri", ...)

  # Usually a user will want to use a global "smart open":
  whatever_obj = tiledbsoma.open("uri/pointing/at/a/ndarray", ...)
  # whatever_obj is now a NDArray
```
* The `create` and `open` methods will now be static AKA class-level methods
* No longer will `Experiment` / `Measurement` / etc be able to be in a does-not-exist-yet state

johnkerl:
Status from @thetorpedodog is this is closeable once associated doc-work is done.

johnkerl:
🚀

PR #823, opened 2023-01-27
[python] Create `ReadWriteHandle` to manage TileDB handles

thetorpedodog:
Context: Part of #638 and https://github.com/single-cell-data/SOMA/issues/41. Split out from largerPR #824.

Unlike typical file objects, when a TileDB object is opened in `w` mode, it is *only* a writer; it can't actually read anything at all. This introduces a class that allows you to manage both of them in one place, and more easily flush writes and ensure that they are visible.

---

It feels pretty silly that this very small change is about the best thing that I can separate out from the other [HUUUUGE](https://www.youtube.com/watch?v=pQO0v_HEGU0#t=27s) change but here we are.

This ends up being important for the immediately forthcoming big huge massive awful change that restructures creation and opening. If I’ve missed something obvious here and it’s all much simpler than I’m making it, then that’s good, and the amount of churn shouldn’t be too great because this encapsulates all the logic I was dealing with pretty nicely.

### DIFF
--- a/apis/python/src/tiledbsoma/util_tiledb.py
+++ b/apis/python/src/tiledbsoma/util_tiledb.py
@@ -1,11 +1,15 @@
-from typing import TypeVar, cast
+from typing import Any, Generic, Optional, TypeVar, cast
 
+import attrs
 import numpy as np
 import pandas._typing as pdt
 import scipy.sparse as sp
 import tiledb
 from pandas.api.types import infer_dtype, is_categorical_dtype
+from somacore import options
 
+from .exception import DoesNotExistError, SOMAError
+from .options import SOMATileDBContext
 from .types import NPNDArray, PDSeries
 
 SOMA_OBJECT_TYPE_METADATA_KEY = "soma_object_type"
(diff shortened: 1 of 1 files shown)


apis/python/src/tiledbsoma/util_tiledb.py
```diff
@@ -96,3 +100,114 @@ def is_duplicate_group_key_error(e: tiledb.TileDBError) -> bool:
...
+@attrs.define()
+class ReadWriteHandle(Generic[_TDBO]):
+    """Paired read/write handles to a TileDB object.
+
+    You can (and should) access the members, but do not write to them.
+    """
+
+    uri: str
+    context: SOMATileDBContext
+    read: _TDBO
```
johnkerl, line 118: Very awkwardly named. `read` is a verb and a noun and an adjective; and a Python builtin.

Please call this something (anything) else: maybe `object_for_read`, or `read_object`, ...

johnkerl (changes requested):
Other than my ask, I like this :)

apis/python/src/tiledbsoma/util_tiledb.py
thetorpedodog: These are now `reader` and `writer`.

johnkerl (approved):

Issue #822, opened 2023-01-27
[python] TileDB error when creating `SparseNDArray` due to mis-handling of `platform_config`


bkmartinjr:
The SparseNDArray create method (and possible create for other foundational types) incorrectly determines the tile extent when:
* platform_config specifies a tile extent
* and, the NDArray shape is smaller than that extent

The  code in error is in the create() method:
>                 tile=tiledb_create_options.dim_tile(dim_name, min(e, 2048)),
It should be applying the `min` to the return value of `dim_tile`, eg.,
>                 tile=min(e, tiledb_create_options.dim_tile(dim_name, 2048)),

Test case:
```
In [1]: import pyarrow as pa
   ...: import tiledbsoma as soma
   ...: 
   ...: snda = soma.SparseNDArray("foo")
   ...: snda.create(
   ...:     pa.float32(),
   ...:     (100, 10000),
   ...:     platform_config={
   ...:         "tiledb": {
   ...:             "create": {
   ...:                 "dims": {
   ...:                     "soma_dim_0": {"tile": 2048},
   ...:                     "soma_dim_1": {"tile": 2048},
   ...:                 }
   ...:             }
   ...:         }
   ...:     },
   ...: )
   ...: 
---------------------------------------------------------------------------
TileDBError                               Traceback (most recent call last)
Cell In[1], line 5
      2 import tiledbsoma as soma
      4 snda = soma.SparseNDArray("foo")
----> 5 snda.create(
      6     pa.float32(),
      7     (100, 10000),
      8     platform_config={
      9         "tiledb": {
     10             "create": {
     11                 "dims": {
     12                     "soma_dim_0": {"tile": 2048},
     13                     "soma_dim_1": {"tile": 2048},
     14                 }
     15             }
     16         }
     17     },
     18 )

File ~/cell-census/venv/lib/python3.9/site-packages/tiledbsoma/sparse_nd_array.py:84, in SparseNDArray.create(self, type, shape, platform_config)
     82 for n, e in enumerate(shape):
     83     dim_name = f"soma_dim_{n}"
---> 84     dim = tiledb.Dim(
     85         name=dim_name,
     86         domain=(0, e - 1),
     87         tile=tiledb_create_options.dim_tile(dim_name, min(e, 2048)),
     88         dtype=np.int64,
     89         filters=tiledb_create_options.dim_filters(
     90             dim_name,
     91             [
     92                 dict(
     93                     _type="ZstdFilter",
[...]

PR #824, opened 2023-01-27
[python] Create and open objects using `Thing.create()` and `Thing.open()`

thetorpedodog:
This is the massive change to rework the whole workflow of creating and opening SOMA objects to work more like files. `.create()` creates an object and gives you back an open handle. `.open()` opens an object that already exists.

This does *not* implement the `.add_whatever` part of Collections, which should reduce a lot of test (and client!) code but I am going to save for a future change.

Unfortunately, this kind of has to go through as one big thing because it's a change to the semantics of the interior, and if you want the tests to keep passing, well...

Additional notes:

- I had to sprinkle a few `@typeguard_ignore` comments around because TypeGuard chokes very badly on some of the generics stuff we're doing.
- Most of the tests for `whatever.exists()` and the like are gone, because being able to open an object means that it exists.
- Flushing could certainly be handled better. There are cases where the new code does it automatically, and cases where it needs to happen manually, and there's not too much logic as to how I've chosen to do one or the other apart from "maybe kind of trying to match the behavior to change the tests as little as possible".

## Issue and/or context:

check out #823 which is a cool change related to this

Part of #638 and https://github.com/single-cell-data/SOMA/issues/41

## Notes for Reviewer:

i am so sorry for everything i have done


thetorpedodog:
@bkmartinjr some of the `experiment_query` tests are now failing and the reason is not obvious to me. I’m going to need some help figuring this one out.

bkmartinjr:
@thetorpedodog - is there an example of the canonical (best practice) way to create/write/close a simple object (eg., a DataFrame or SparseNDArray)?   I see all of the create/write/flush examples - is that the API pattern we are suggesting people use (ie., bottoms up)?  

I had expected to see some top-down patterns, eg., using `collection.add_new_foo()`.   Or is that still TBD?

thetorpedodog:
First off: Bruce, you are a lifesaver.

> @thetorpedodog - is there an example of the canonical (best practice) way to create/write/close a simple object (eg., a DataFrame or SparseNDArray)? I see all of the create/write/flush examples - is that the API pattern we are suggesting people use (ie., bottoms up)?
> 
> I had expected to see some top-down patterns, eg., using `collection.add_new_foo()`. Or is that still TBD?

No, this is still something of a transitional state. The `.add_new_whatever` is coming in a follow-up change (which should be *much* more manageable), just because this one is already so huge.

thetorpedodog:
Based on discussion just now, I am going to be reworking this to completely separate read and write modes. The changes shouldn't be too huge and will mainly be in the tests.

apis/python/src/tiledbsoma/constants.py
```diff
@@ -3,3 +3,6 @@
 """
 
 SOMA_JOINID = "soma_joinid"
```
bkmartinjr, line 5: probably worth adding `soma_data`, `soma_dim_` as these are in the spec, and widely used in the NDArray code

bkmartinjr: at some point, these + soma_joinid should all  upstream into `somacore` as they are independent of the storage engine

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -73,12 +67,13 @@ def _ctx(self) -> tiledb.Ctx:
...
+        # This needs to be implemented as a @property because Python's ABCs
+        # require that abstract properties be implemented on the object itself,
+        # rather than being a field (i.e., creating self.whatever in __init__).
         return self._metadata
-        # Note: this seems trivial, like we could just have `metadata` as an attribute.
-        # However, we've found that since in `somacore` it's implemented as `@property`,
-        # to avoid a static-analysis failure we have to do the same here.
 
+    # TODO: This needs reconsidering, since it means we're deleting something
+    # out from underneath ourselves.
```
bkmartinjr, line 76: should be a class method

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -87,20 +82,16 @@ def delete(self) -> None:
...
 
     def __repr__(self) -> str:
-        """
-        Default repr
-        """
-        if self.exists():
-            return f'{self.soma_type}(uri="{self._uri}")'
-        else:
-            return f"{self.soma_type}(not created)"
+        return f'{self.soma_type}(uri="{self._uri}")'
```
bkmartinjr, line 94: I can't express how much joy I feel with the removal of the `(not created)` string!

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -124,134 +115,41 @@ def exists(self) -> bool:
 
         [lifecycle: experimental]
         """
+        # We already opened this, so it should always exist.
+        return not self._was_deleted
```
bkmartinjr, line 119: to confirm: the API is not supporting the edge case where the underlying data was deleted after open. 

I consider this acceptable, but the old code would catch that case, so there is a change in semantics.

apis/python/src/tiledbsoma/tiledb_array.py
```diff
@@ -1,35 +1,49 @@
...
 from .util_arrow import get_arrow_schema_from_tiledb_uri
+from .util_tiledb import ReadWriteHandle
 
+_Self = TypeVar("_Self", bound="TileDBArray")
 
-class TileDBArray(TileDBObject):
+
+class TileDBArray(TileDBObject[tiledb.Array]):
     """
     Wraps arrays from TileDB-Py by retaining a URI, options, etc.  Also serves as an abstraction layer to hide TileDB-specific details from the API, unless requested.
```
bkmartinjr, line 20: comment slightly misleading, as it doesn't just wrap TileDB-Py.  I would ammend to "wraps TileDB"

apis/python/src/tiledbsoma/tiledb_array.py
```diff
@@ -1,35 +1,49 @@
...
 from .tiledb_object import TileDBObject
 from .util_arrow import get_arrow_schema_from_tiledb_uri
+from .util_tiledb import ReadWriteHandle
 
+_Self = TypeVar("_Self", bound="TileDBArray")
 
-class TileDBArray(TileDBObject):
+
+class TileDBArray(TileDBObject[tiledb.Array]):
     """
```
bkmartinjr, line 19: Nit:  the `TileDB...` classes are private implementation details, not intended for reuse.  Should they be marked private, and/or named with an underscore prefix?

apis/python/src/tiledbsoma/tiledb_array.py
```diff
@@ -38,57 +52,31 @@ def schema(self) -> pa.Schema:
...
         """
         Reads the attribute names from the schema: for example, the list of column names in a dataframe.
         """
-        with self._ensure_open():
-            A = self._tiledb_obj
-            return [A.schema.attr(i).name for i in range(A.schema.nattr)]
+        arr = self._handle.reader
+        return tuple(arr.schema.attr(i).name for i in range(arr.schema.nattr))
 
     def _soma_reader(
```
bkmartinjr, line 81: I (naively?) expected the SOMAReader instance, which is really just a read-only TileDB handle, to be integrated into _handle.reader (or something similar).  At least for TileDB "array" objects, shouldn't all "handles" to the underlying array be managed in a consistent manner?

I'm thinking of stuff like the `flush(update_read=True)` call -- wouldn't we want the SOMAReader re-opened as well?

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -207,54 +136,53 @@ def read(
...
+            elif isinstance(coord, int):
+                sr.set_dim_points(dim_name, [coord])
+            elif isinstance(coord, np.ndarray):
+                if coord.ndim != 1:
+                    raise ValueError(
+                        f"only 1D numpy arrays may be used to index; got {coord.ndim}"
+                    )
+                sr.set_dim_points(dim_name, coord)
+            elif isinstance(coord, slice):
+                ned = arr.nonempty_domain()  # None iff the array has no data
```
bkmartinjr, line 167: This could cause a major bug due to SOMAReader being handled differently than _handle.reader.

Example:

1. Create new array (_handle has open read and write, but there is no SOMAReader handle).  
2. Write to array (don't close or flush).
4. Call _this_ read method with a coordinate slice:
    * It will create a new SOMAReader
    * It will call _handle.reader.nonempty_domain() *which will return None* because _handle.reader was opened before th write
    * It will therefore skip setting the SOMAReader dim_range
    * and will return the entire array, rather than the slice the user requested

Example of this with DataFrame - where it asks for 2 rows in a slice, but gets all10 back:
```

import pyarrow as pa
import tiledbsoma as soma

path = 'foo'
sz = 10

df = soma.DataFrame.create(
    path,
    schema=pa.schema(
        [
            ("soma_joinid", pa.int64()),
            ("label", pa.large_string()),
        ]
    ),
    index_column_names=["soma_joinid"],
)
df.write(
    pa.Table.from_pydict(
        {
            "soma_joinid": [i for i in range(sz)],
            "label": [str(i) for i in range(sz)],
        }
    )
)

# the result of this slice read is unexpected
# (note lack of flush before doing the read)

print(repr(df.read((slice(1,3),)).concat()))
```

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -374,3 +301,68 @@ def tables(self) -> TableReadIter:
...
+            " SOMA NDArrays only support primtive Arrow types"
+        )
+
+    dims = []
+    for n, e in enumerate(shape):
+        dim_name = f"soma_dim_{n}"
+        dim = tiledb.Dim(
+            name=dim_name,
+            domain=(0, e - 1),
+            tile=create_options.dim_tile(dim_name, min(e, 2048)),
```
bkmartinjr, line 332: this is missing a fix John just applied to main in #822

apis/python/src/tiledbsoma/tiledb_object.py
bkmartinjr: Second thought:  why isn't this a class method, which takes a URI?  The idea being that create/delete/exists all should be class methods, outside of the lifecycle of the actual instance.

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -374,3 +301,68 @@ def tables(self) -> TableReadIter:
         [lifecycle: experimental]
         """
         return TableReadIter(self.sr)
+
+
+def _build_tiledb_schema(
```
bkmartinjr, line 306: super tiny nit: given that this is used by *NDArray, you might consider moving it into util_tiledb.py - it threw me when I saw the DenseNDArray implementation importing it :-)

apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -67,105 +50,47 @@ def create_legacy(
...
+            context,
+        )
+        handle = cls._create_internal(uri, tdb_schema, context)
+        is_sparse = tdb_schema.sparse
+        return cls(
+            uri,
+            "w",
+            handle,
+            context,
+            _is_sparse=is_sparse,
```
bkmartinjr, line 68: given that all SOMA DataFrames are sparse (as of Nov 2022), this attribute seems like dead code.  Can we remove it?  I couldn't find anything that actually uses the value.....

apis/python/src/tiledbsoma/tiledb_array.py
thetorpedodog: True but something I want to leave out of this.

thetorpedodog (commented):
Thank you very much for the great feedback on the review. I've done a few of the smaller fixes and many of the other ones are things I have already been thinking about for direct followups or other cleanup efforts.

Thus far it seems like some of the concerns you had are made irrelevant by the fact that we do read xor write, which is good because it says to me that this files a lot of sharp corners off the API.

There are a few comments marked as "outdated" that I have to figure out how to get to since clicking on them just takes me to nowhere.

apis/python/src/tiledbsoma/tiledb_object.py
thetorpedodog: I've removed it entirely for now; when we restore it it should just be `tiledbsoma.delete` since it's not attached to a type.

bkmartinjr (commented):
I am still working on the review, but so far there are two major bits of feedback:
* consider making delete & exists class methods that operate on URL, like create.
* SOMAReader is a read handle, and needs to be handled consistently with the TileDB-Py read handle or you will get weird behavior (eg, the slicing bug introduced in all three array-based types, due to unsynchronized use of the TileDB-Py and SOMAReader handles)

apis/python/src/tiledbsoma/tiledb_object.py
thetorpedodog: I've removed this method, since as you note we're not supporting the weird "pull the rug out" edge case. It could be added back in as a classmethod, but I feel like most uses of an "exists" function would be in the form of "does it exist, if so open, else create", which is better handled by the easier-to-ask-forgiveness model of "try and open, otherwise create".

HOWEVER!

I do think there is room for a unified "open-or-create" function which does both of these at the same time since that's a common use case (compare: opening a file in `r+` mode), and it could save a lot of user duplication in that respect.

apis/python/src/tiledbsoma/sparse_nd_array.py
thetorpedodog: Picked this up upon rebasing.

apis/python/src/tiledbsoma/tiledb_array.py
bkmartinjr: I think until that is done, there are bugs aplenty to be found during the create process where handles get out of sync.  I put one trivial example into the comments, but I bet there are more due to the opportunity for the two separate read handles to get out of sync.

We really have a worst-case scenario right now - _three_ handles:
* TileDB-Py write handle
* TileDB-Py read handle for `.schema` and `.nonempty_domain`
* SOMAReader read handle for the rest of the read path
Ooof....

I wonder if @gspowley could make your life easier by providing all "read" path stuff in a single SOMAReader handle?

thetorpedodog:
another thing: once we get the `add_new_[whatever]` methods in, we can trace object ownership and use that for cascading close, so you open a root collection, create / modify / whatever the children, then close() the root collection and *everything* is closed for you. when that happens, the developer experience should be really nice (way fewer with-blocks, just create and add and go)

apis/python/src/tiledbsoma/types.py
```diff
@@ -39,5 +40,7 @@
     pa.SparseCSCMatrix,
 ]
 
-PlatformConfig = Mapping[str, Any]
-"""The platform-configuration dictionary. May contain a ``tiledb`` key."""
+StorageType = Literal["array", "group"]
+"""How a SOMA object is stored."""
+TDBHandle = Union[tiledb.Array, tiledb.Group]
```
bkmartinjr, line 45: given how the `tiledb` types are defined, this might be better as:
> TDBHandle = Union[tiledb.DenseArray, tiledb.SparseArray, tiledb.Group]

apis/python/src/tiledbsoma/tiledb_array.py
thetorpedodog: I fixed a few spots where I was not properly checking for read mode when using the SOMAReader. This means that we have one of two situations:

1. Only a write handle, with an auxiliary read handle. The SOMAReader is never used.
2. A tiledb-py read handle, and a SOMAReader (temporary) read handle.

Because the SOMAReader is constructed with the same Context (technically, config) as the tiledb-py read handle, it will get the same view of the dataset as the read handle (pending ensuring that we sensibly set timestamps on everything).

apis/python/src/tiledbsoma/types.py
thetorpedodog: Probably true. There is a lot of type-annotation–related stuff that I pretty much got to the point of "it works and I don't want to touch it" and I want to go back through and clean that up after this implementation work settles down.

apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -67,105 +50,34 @@ def create_legacy(
...
         )
-        self._is_sparse = sch.sparse
 
-        tiledb.Array.create(self._uri, sch)
+    def __init__(
+        self,
+        handle: tiledb.Array,
+        *,
+        # Hints to pre-fill cache entries.
+        _index_column_names: Optional[Tuple[str, ...]] = None,
```
bkmartinjr, line 73: I'm not sure caching the index column names makes any sense given the change to stateful handles.  I believe that it was originally put in place to avoid repeated calls to `open` on the array (just to get the dimension names).  With an open handle guaranteed, getting dim names is cheap, ergo caching is not worth the complexity.

apis/python/tests/test_experiment_query.py
```diff
@@ -704,41 +694,41 @@ def test_experiment_query_column_names(soma_experiment):
...
+            pa.Table.from_pydict(
+                {
+                    "soma_joinid": [i for i in range(sz)],
+                    "label": [str(i) for i in range(sz)],
+                }
+            )
         )
-    )
-    return df
+    return factory.open(path)
```
bkmartinjr, line 715: this is making me think we need a `tiledbsoma.open(uri)` method, published at the top-level...

apis/python/src/tiledbsoma/constants.py
bkmartinjr: Works for me.  On the deferred items, are you tracking the backlog?  (I'm pretty sure you are, just confirming)

apis/python/src/tiledbsoma/tiledb_object.py
bkmartinjr: I agree.  I also think there is room for an "open URI and return it" without knowing the type a priori.

A la `tiledbsoma.open(uri)`

apis/python/src/tiledbsoma/tiledb_array.py
bkmartinjr: That is true for now, when we can only open the 'latest'.  As soon as @mlin lands his timestamp work, it will be quite easy to get them out of sync.  For the next PR :-)

bkmartinjr (approved):
Made it through - overall this is a very big step in the right direction.  I think there are a bunch of items we need to progressively fix in the next, next-next, ..., PRs.  Marking this as approved to let Paul move forward with that progression.

apis/python/src/tiledbsoma/collection.py
```diff
@@ -1,98 +1,113 @@
...
+        *,
+        context: Optional[SOMATileDBContext] = None,
+        platform_config: Optional[options.PlatformConfig] = None,
+    ) -> _Self:
+        """Opens this specific type of collection."""
+        del platform_config  # unused
+        context = context or SOMATileDBContext()
+        handle = ReadWriteHandle.open_group(uri, mode, context)
+        # TODO: Verify that we have the right type using metadata.
+        return cls(handle, _this_is_internal_only="tiledbsoma-internal-code")
```
johnkerl, line 93: _pedo mellon a minno_ :D

apis/python/src/tiledbsoma/collection.py
```diff
@@ -367,41 +293,38 @@ def _set_element(
                 #
                 # The standard solution is a negligible but non-zero delay.
                 time.sleep(0.001)
+        # HACK: There is no way to change a group entry without deleting it and
```
johnkerl, line 296: comment gold -- future generations thank you

apis/python/tests/test_dense_nd_array.py
```diff
@@ -30,16 +23,16 @@ def test_dense_nd_array_create_ok(
     """
     assert pa.types.is_primitive(element_type)  # sanity check incoming params
 
-    a = soma.DenseNDArray(uri=tmp_path.as_posix())
     with pytest.raises(TypeError):
-        a.create_legacy(element_type.to_pandas_dtype(), shape)  # XXX
-    a.create_legacy(element_type, shape)
+        soma.DenseNDArray.create(
+            tmp_path.as_posix(), type=element_type.to_pandas_dtype(), shape=shape
+        )  # XXX
```
johnkerl, line 29: what is `XXX` a reminder for?

johnkerl (approved):
🚢

apis/python/src/tiledbsoma/tiledb_array.py
mlin: Right now a fresh `SOMAReader` is needed for each query, and some refactoring is needed in order to keep its TileDB handle open for multiple operations. @gspowley and I have discussed a couple possible approaches. Suggest this can be left to follow-up work.

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -28,230 +24,95 @@ class TileDBObject(ABC, somacore.SOMAObject):
...
+        handle.writer.meta.update(
+            {
+                constants.SOMA_OBJECT_TYPE_METADATA_KEY: cls.soma_type,
+                constants.SOMA_ENCODING_VERSION_METADATA_KEY: constants.SOMA_ENCODING_VERSION,
+            }
+        )
+        # HACK: We need this so that the metadata appears on the read handle.
+        handle._flush_hack()
+
+    def _ensure_open_read(self) -> None:
```
mlin, line 113: Nit: "ensure" might suggest it's going to open it if not already, maybe this is just a "check"

apis/python/src/tiledbsoma/collection.py
```diff
@@ -102,81 +117,41 @@ def _not_implemented(self, *args: Any, **kwargs: Any) -> NoReturn:
...
+            else:
+                raise SOMAError(
+                    f"internal error: unrecognized group entry type {entry.cls}"
+                )
+            entry.soma = opener(
+                uri=entry.uri,
+                mode=self.mode,
+                context=self.context,
+                soma_type=None,
+            )
```
mlin, line 153: What's later going to close the thing? (If it happens implicitly and I missed it, then it's probably worth an explanatory comment?)

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -28,230 +24,95 @@ class TileDBObject(ABC, somacore.SOMAObject):
...
         """
         Release any resources held while the object is open. Closing an already-closed object is a
         no-op.
         """
-        if self._open_mode:
-            self._open_mode = None
-            self._close_stack.close()
+        my_cs: Optional[ExitStack] = getattr(self, "_close_stack", None)
+        if my_cs:
+            my_cs.close()
```
mlin, line 92:
```suggestion
        self._close_stack.close()
```

apis/python/src/tiledbsoma/util_tiledb.py
```diff
@@ -150,11 +139,21 @@ def open_group(
...
+        cls,
+        tiledb_cls: Callable[..., _TDBO],
+        uri: str,
+        mode: options.OpenMode,
+        context: SOMATileDBContext,
+    ) -> "ReadWriteHandle[_TDBO]":
         try:
-            rd = tiledb.Group(uri, "r", ctx=context.tiledb_ctx)
+            rd = tiledb_cls(uri, "r", ctx=context.tiledb_ctx)
             try:
```
mlin, line 154: Might be nice to let an `ExitStack` manage unwinding the handles on exception paths here and in `close()`, esp if we're soon going to have a third thing `SOMAReader` to unwind.

thetorpedodog: This is going to end up further changing pretty soon (better handling of write-only handles) but I will see if that makes sense after those revisions.

apis/python/tests/test_dense_nd_array.py
thetorpedodog: Not sure; [it was there when I got here](https://github.com/single-cell-data/TileDB-SOMA/blame/da1456d72fdcc1619a7139bdb45c640a0010a72d/apis/python/tests/test_dense_nd_array.py).

apis/python/tests/test_experiment_query.py
thetorpedodog: Exposing this as `tiledbsoma.open` is my exact goal once this all goes through.

apis/python/src/tiledbsoma/dataframe.py
thetorpedodog: I’ll leave this as-is for now to reduce churn here.

thetorpedodog:
Leaving a comment here so that the record shows commit https://github.com/single-cell-data/TileDB-SOMA/commit/1642ebd8ad35e97db48aec028ac4a82e6e12c908 as handling the feedback above before I rebase and squash everything together.

Thank you all very much for the excellent review feedback!

Issue #820, opened 2023-01-26
[r] Look into `remotes::install_github`


johnkerl:
Primary use-case at present is trying to install a new version of `tiledbsoma-r` in a notebook environment which has an old version installed (namely, `main-old`, `v0.1.*`).

This needs to be analyzed.

<img width="944" alt="Screenshot 2023-01-26 at 5 26 14 PM" src="https://user-images.githubusercontent.com/2008794/214964804-80ee18cf-94d4-413b-812c-8401757ab741.png">

johnkerl:
cc @aaronwolen @eddelbuettel @mojaveazure

eddelbuettel:
As discussed at length on our slack this is a 'Venn Diagram' intersection of an usual repo layout meeting tools expecting standards conformance.

PR #826, opened 2023-01-27
[r] Support remotes::install_github use

eddelbuettel:
#### Issue and/or context: #820

R users are accustomed to installing from Git(Hub) repositories via `remotes::install_github()`.  For reasons specific to this project, this repository has a somewhat non-standard layout and build process which has been seen to be conflicting.  This overcomes this.

#### Changes:

1. The `cleanup` scripts (invoked by standard `R CMD ...` processes in a source repository) is altered to not only copy the `libtiledbsoma/` sources but also keep a copy as tarball in the toplevel directory where a run of `R CMD build .` (to create a release tar.gz) will by default not include it -- meaning 'it exists' for us to build, it is not redundantly shipped.

2. The `configure` script (invoked by each `R CMD INSTALL ...`) now checks for tarball and unpacks it if need be.

3. The rest of the build script is as before.

#### Notes for Reviewer:

See https://app.shortcut.com/tiledb-inc/story/24921/support-remotes-install-github-by-making-apis-r-sufficient-by-placing-a-copy-of-libtiledbsoma-there-and-refreshing-it


johnkerl (approved):
Works great on Linux! :)
https://gist.github.com/johnkerl/7dad15b5a4c95cc5ae7ff8786ad80736

MacOS, apparent rpath issue -- ?
https://gist.github.com/johnkerl/1fcfec15e278e69c18abeff25e3c1e93

Regardless, having Linux support for this feature is, alone, a nice improvement. IMO we should merge this, and deal with the MacOS issue as a follow-on task.

eddelbuettel:
We should probably clean this up before merging as it has the potential to disrupt CI and standard use on macOS.

PR #827, opened 2023-01-27
[python] Use the contents of `bld`, not `setup.py`, for binary cache

thetorpedodog:
We were (I was) using the contents of `setup.py` for the cache key when building the tiledbsoma C++ library. This was not ideal since it meant the cache would be busted when unrelated changes happened. We should instead use the `bld` script, which does the actual blding, as the key.

### DIFF
--- a/.github/workflows/python-ci-single.yml
+++ b/.github/workflows/python-ci-single.yml
@@ -10,7 +10,7 @@ on:
       python_version:
         required: true
         type: string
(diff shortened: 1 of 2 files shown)


johnkerl (approved):
LGTM but I would need a @gsakkis or an @mlin to assess correctness ...

gsakkis (approved):
LGTM

Issue #821, opened 2023-01-26
[python] Relative-URI flag is not being respected


johnkerl:
Context: #643 

This worked in `main-old`, and was broken at some point in `main`.

* Relative URs are important for anyone who does `aws sync` between local disk S3, and/or moves data around on local disk, and/or moves data around on S3.
* We didn't detect this earlier because we didn't unit-test it earlier
  * https://github.com/TileDB-Inc/TileDB-Py/pull/1386 now exists and will help us with our unit tests
  * We also as a unit-test case need to write an experiment to directory `.../a`, then `mv` to `.../b`, then make sure it's traversable before and after

bkmartinjr:
They are still working for me (in the cell census builder). You may want to post a test case as it is likely that a narrow version is failing.

I use `collection.set(..., relative=True)` and it works great.

johnkerl:
Indeed

The issue involves the `relative` flag as argument vs as a context param -- we allow both, and the former takes precedence over the latter

That would explain why you do not see the issue

I'll post test cases as part of the PR that fixes the problem (this issue is already assigned to me)

PR #828, opened 2023-01-27
[python] Fix `use_relative_uri` for tiledbsoma.io [pre-824]

johnkerl:
## Issue and/or context:

#821 

## Changes:

In addition to the problem described in #821, namely, S3 and local-disk SOMA data is being written with absolute, not relative, URIs: having an option-flag within the `context` struct is a hold-over from the `main-old` days when we had multiple things propagated from parent to child -- including the (now long-gone) `indent` strings for `__repr__`. It's now redundant with `use_relative_uri`; we don't need both.

## Notes to the reviewer:

This is the same as 833 but this is _not_ is atop the branch create-create from PR https://github.com/single-cell-data/TileDB-SOMA/pull/824

### DIFF
--- a/apis/python/devtools/ingestor
+++ b/apis/python/devtools/ingestor
@@ -102,19 +102,19 @@ select `relative=True`. (This is the default.)
         tiledbsoma.logging.info()
 
     if args.relative is None:
-        rel_member_uris = None
+        use_relative_uri = None
     else:
         relative = args.relative[0]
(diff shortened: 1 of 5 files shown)


atolopko-czi (approved):
Looks reasonable, thanks for the simplification!

johnkerl:
Closed in favor of #833, as #824 has been merged

PR #832, opened 2023-01-27
[python] Fix legacy misuse of`urllib.parse.urljoin`

johnkerl:
## Issue and/or context:

Prep-work for #821

## Changes:

```
>>> from urllib.parse import urljoin
>>> urljoin("A", "B")
'B'

>>> urljoin("A"+"/", "B")
'A/B'

>>> import os
>>> os.path.join("A", "B")
'A/B'
```

`urljoin` is confusing and error-prone. (Its use for unit tests tracks back to #381.). We should not use it for unit tests.

`os.path.join` is correct and trustworthy. We should use it for unit tests.

Barring that, we should work around the idiosyncracies of `urljoin`.

### DIFF
--- a/apis/python/tests/test_experiment_basic.py
+++ b/apis/python/tests/test_experiment_basic.py
@@ -1,4 +1,4 @@
-from urllib.parse import urljoin
+import os
 
 import numpy as np
 import pyarrow as pa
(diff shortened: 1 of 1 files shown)


thetorpedodog:
This code is going to change dramatically when #824 is in, so I would hold off until then.

johnkerl:
@thetorpedodog it's 3 lines ... 🙏

bkmartinjr (dismissed):
Given that these _are_ URLs, I would argue that urljoin is actually the correct call, and that the input to the urljoin is the bug.  os.path.join does not correctly resolve relative urls....

You could just fix it with:   urljoin("a/', 'b')

Put another way, an even _better_ fix is just to give urljoin the right first parameter.  Docs are [here](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin), where it shows the first arg is a base path.

bkmartinjr (approved):
LGTM

PR #833, opened 2023-01-29
[python] Fix `use_relative_uri` for tiledbsoma.io

johnkerl:
## Issue and/or context:

#821 

## Changes:

In addition to the problem described in #821, namely, S3 and local-disk SOMA data is being written with absolute, not relative, URIs: having an option-flag within the `context` struct is a hold-over from the `main-old` days when we had multiple things propagated from parent to child -- including the (now long-gone) `indent` strings for `__repr__`. It's now redundant with `use_relative_uri`; we don't need both.

## Notes to the reviewer:

Currently stacked atop #832, until that is approved and merged.

This is the same as #828 but is atop the branch `create-create` from PR #824


johnkerl:
I'm getting a completely unconstrucive `ruff` error in CI that I don't see in sandbox, either MacOS or Ubuntu.

```
$ ruff --version
ruff 0.0.231
```

in both places.

johnkerl:
I'm getting a completely unconstructive error message from `ruff`

```
ruff.....................................................................Failed
- hook id: ruff
- duration: 0.09s
- files were modified by this hook

Found 5 error(s) (5 fixed, 0 remaining).
```

which I cannot repro in my sandbox in spite of my having the same `ruff` version `0.0.226` as [.pre-commit-config.yaml](https://github.com/single-cell-data/TileDB-SOMA/blob/create-create/.pre-commit-config.yaml#L7-L8) :(

atolopko-czi (approved):
Looks reasonable, thanks for the simplification!

atolopko-czi:
> I'm getting a completely unconstructive error message from `ruff`
>  ...
> which I cannot repro in my sandbox in spite of my having the same `ruff` version `0.0.226` as [.pre-commit-config.yaml](https://github.com/single-cell-data/TileDB-SOMA/blob/create-create/.pre-commit-config.yaml#L7-L8) :(

fixed

PR #836, opened 2023-01-30
Pull the common parts of NDArrays into the base NDArray class.

thetorpedodog:
Eliminates some duplicate code and some redundant type-assertions/casts.

### DIFF
--- a/apis/python/src/tiledbsoma/common_nd_array.py
+++ b/apis/python/src/tiledbsoma/common_nd_array.py
@@ -1,9 +1,6 @@
-"""Common code shared by both NDArray implementations.
+"""Common code shared by both NDArray implementations."""
 
-TODO: Extract more stuff.
-"""
-
-from typing import Optional, Sequence, Type, TypeVar
+from typing import Optional, Sequence, Tuple, Type, TypeVar, cast
 
 import numpy as np
 import pyarrow as pa
@@ -20,6 +17,8 @@ _Self = TypeVar("_Self", bound="NDArray")
 
 
 class NDArray(TileDBArray, somacore.NDArray):
+    """Abstract base for the common behaviors of both kinds of NDArray."""
+
     @classmethod
     def create(
         cls: Type[_Self],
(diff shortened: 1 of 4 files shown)


mlin:
@thetorpedodog Is there supposed to be a bigger diff/deletion to `sparse_nd_array.py` like there is in `dense_nd_array.py`?

thetorpedodog:
> @thetorpedodog Is there supposed to be a bigger diff/deletion to `sparse_nd_array.py` like there is in `dense_nd_array.py`?

That part already happened in https://github.com/single-cell-data/TileDB-SOMA/commit/93050f1ca76700b6038033befe953b6f6085bee0#diff-ff1513d37e7cf18c204fe02ac36f85121c928794afc8012e438fe3cba63b2a2e. Unfortunately, there’s nothing more in common left to extract.

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -26,7 +25,7 @@
 _UNBATCHED = options.BatchSize()
 
 
-class SparseNDArray(TileDBArray, somacore.SparseNDArray):
+class SparseNDArray(NDArray, somacore.SparseNDArray):
```
atolopko-czi, line 28: Can the `create()` class method be removed from `SparseNDArray` as well?

apis/python/src/tiledbsoma/common_nd_array.py
```diff
@@ -3,15 +3,54 @@
...
 
 from . import util, util_arrow
 from .options.soma_tiledb_context import SOMATileDBContext
 from .options.tiledb_create_options import TileDBCreateOptions
+from .tiledb_array import TileDBArray
+
+_Self = TypeVar("_Self", bound="NDArray")
+
+
+class NDArray(TileDBArray, somacore.NDArray):
```
atolopko-czi, line 22: could this be documented as being a base class and is useful for polymorphic typing?

apis/python/src/tiledbsoma/sparse_nd_array.py
thetorpedodog: How did I manage to miss that? And the other things that are also shared in common? Feeling pretty silly about that right now.

apis/python/src/tiledbsoma/common_nd_array.py
thetorpedodog: Added a brief note. If you think it needs a little more context, please do say so.

atolopko-czi (approved):
LGTM!

PR #837, opened 2023-01-31
Add `add_new_collection` implementation.

thetorpedodog:
This adds, in three steps, the `add_new_collection` implementation to Collection. The use of a relative vs. absolute URI is derived from the URI passed into the function (if it's relative, it's relative, if it's absolute, it's absolute). This does not currently support adding to a collection stored on TileDB Cloud without a URI specified.

As I understand it, because the `io.from_whatever` functions all generated their own URIs of the form `parent_uri/child_name`, this was never going to work with groups/arrays stored on TileDB Cloud, because it would try to generate a URI that looked like `tiledb://some_ns/some_group/child_name`, which is not allowed. This meant the `use_relative_uri` function was moot inside those functions anyway, since you can’t actually do the thing you want to use absolute URIs for. Is this understanding correct?

### DIFF
--- a/apis/python/src/tiledbsoma/collection.py
+++ b/apis/python/src/tiledbsoma/collection.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
+import re
 import time
 from typing import (
     Any,
+    ClassVar,
     Dict,
     Generic,
     Iterator,
@@ -12,6 +14,7 @@ from typing import (
     Type,
     TypeVar,
     cast,
+    overload,
 )
 
 import attrs
@@ -24,7 +27,7 @@ from .exception import SOMAError
 from .options import SOMATileDBContext
 from .tiledb_object import AnyTileDBObject, TileDBObject
 from .types import StorageType, TDBHandle
-from .util import make_relative_path
+from .util import is_relative_uri, make_relative_path, uri_joinpath
 from .util_tiledb import (
     ReadWriteHandle,
     is_does_not_exist_error,
@@ -33,6 +36,7 @@ from .util_tiledb import (
 
 # A collection can hold any sub-type of TileDBObject
 CollectionElementType = TypeVar("CollectionElementType", bound=AnyTileDBObject)
+_Coll = TypeVar("_Coll", bound="CollectionBase[AnyTileDBObject]")
 _Self = TypeVar("_Self", bound="CollectionBase[AnyTileDBObject]")
 
 
@@ -57,6 +61,7 @@ class CollectionBase(
     ``DataFrame``, ``DenseNDArray``, ``SparseNDArray`` or ``Experiment``.
     """
 
+    __slots__ = ()
     _tiledb_type = tiledb.Group
 
     # TODO: Implement additional creation of members on collection subclasses.
@@ -92,7 +97,7 @@ class CollectionBase(
 
     # Subclass protocol to constrain which SOMA objects types  may be set on a
     # particular collection key. Used by Experiment and Measurement.
-    _subclass_constrained_soma_types: Dict[str, Tuple[str, ...]] = {}
+    _subclass_constrained_soma_types: ClassVar[Dict[str, Tuple[str, ...]]] = {}
 
     def __init__(
         self,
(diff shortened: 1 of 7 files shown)


apis/python/src/tiledbsoma/collection.py
```diff
@@ -107,10 +112,53 @@ def __init__(
...
+    @overload
+    def add_new_collection(
+        self,
+        key: str,
+        cls: Type[_Coll],
+        *,
+        uri: Optional[str] = ...,
+        platform_config: Optional[options.PlatformConfig] = ...,
+    ) -> _Coll:
+        ...
```
mlin, line 135: @thetorpedodog My only comment here is it's pretty hard to understand what's going on with the class hierarchy typing in `collection.py`. I think it'd be worth leaving a few breadcrumb comments to help the reader understand the goals & constraints that lead to the unusual constructs here (the three declarations of `add_new_collection`, `_real_class`, etc.). I don't doubt there's good reason for them, just need some signposts :smile_cat: 

I'm not familiar enough with TileDB Cloud to address the absolute/relative URI question I'm afraid...

johnkerl:
> Is this understanding correct?

No. Creation URIs are like `tiledb://namespace/s3://path/to/name` which do work. This is used routinely in `main-old`. We don't have unit tests in `main` that depend on tiledb-cloud, for dependency reasons, but these creation URIs were working. I need to re-test them (including notebook runs) now that we've done some refactoring.

thetorpedodog:
> > Is this understanding correct?
> 
> That understanding is incorrect. Creation URIs are like `tiledb://namespace/s3://path/to/name` which do work. This is used routinely in `main-old`. We don't have unit tests in `main` that depend on tiledb-cloud, for dependency reasons, but these creation URIs were working. I need to re-test them (including notebook runs) now that we've done some refactoring.

In this case what I would expect to happen is:

1. You create an array like `tiledb://namespace/s3://path/to/name`.
2. You open `tiledb://namespace/s3://path/to/name`. (this...might work?)
3. The URL of the resulting array is `tiledb://namespace/[something]`.
4. You try to create `tiledb://namespace/[something]/[child]`. (this would fail)

johnkerl:
@thetorpedodog this works in `main-old`:

```
ingestor anndata/pbmc3k_processed.h5ad tiledb://johnkerl-tiledb/s3://tiledb-johnkerl/test-20230131-104117
```

and this _used_ to work in `main` until recently it was broken:

```
ingestor testdata/pbmc3k_processed.h5ad tiledb://johnkerl-tiledb/s3://tiledb-johnkerl/test-20230131-104118
```

We do need to find a way to test tiledb-cloud URIs without introducing a tiledb-cloud dependency into the open-source `tiledbsoma` package.

apis/python/devtools/ingestor
```diff
@@ -106,6 +106,7 @@ select `relative=True`. (This is the default.)
     if args.relative is None:
         use_relative_uri = None
     else:
+        raise NotImplementedError("use_relative_uri is currently not implemented")
```
johnkerl, line 109: It is not OK to break this.

johnkerl (dismissed):
It is not OK to break `is_relative`.

I will work with you on this.

thetorpedodog:
After some discussion and research:

- Creating a collection like `tiledb://namespace/backend://path/to/wherever` will work, as will opening it.
- The reported URI of the opened collection will be `tiledb://namespace/backend://path/to/wherever`.
- Adding a child element URI like `tiledb://namespace/backend://path/to/wherever/child` to a collection will also work. The child element will be stored with a URI like `tiledb://namespace/some-random-uuid` and will *not* include the full backend path.

What this means is:

- Creating a full tree of ingesting things should be able to work correctly the first time, because all the created members will have URIs with their storage backend included.
- Resuming an ingestion is troublesome:
  - The base collection can be opened at `tiledb://namespace/location/backend://path/to/wherever`.
  - However, child members will have a URI that looks like `tiledb://namespace/some-random-uuid`. This means that we can’t automatically create child URIs based on the actual stored group URIs upon resumption.

thetorpedodog:
note to self: I think reworking the URI abstraction layers in `_set_element` will help make this work.

apis/python/src/tiledbsoma/collection.py
atolopko-czi: +1. Not familiar with the multiple overrides. Is this just for dev-time type-inference support for method return types?

atolopko-czi (commented):
Core changes LGTM.

Not approving since the relative/absolute uri issue that John raised seems to still need resolution, and so I don't want to encourage merging this yet.

Since a non-trivial amount of new code is concerned with checking that correct child collection types are being added, that could be worth having some unit tests for.

apis/python/tests/test_collection.py
```diff
@@ -207,3 +210,40 @@ def test_collection_update_on_set(tmp_path):
...
+
+
+@pytest.mark.parametrize(
+    ("key", "want"),
+    [
+        ("hello", "hello"),
+        ("good bye", "good_bye"),
+        ("../beas/tie@boyz", "_beas_tie_boyz"),
+        ("g0nna~let-the.BEAT", "g0nna_let_the_BEAT"),
+        ("____DROP", "_DROP"),
```
atolopko-czi, line 245: 🎧

thetorpedodog:
> Since a non-trivial amount of new code is concerned with checking that correct child collection types are being added, that could be worth having some unit tests for.

It’s more of a reworking of the existing child-collection-type–checking code that was there before and so is covered by many of the same tests, both positive and negative. I will definitely do another check to see if there are any new cases it makes sense to add.

thetorpedodog:
In the interest of getting The Good Parts of this change in while leaving the complicated stuff for later, I’ve decided to pull the changes to `tiledbsoma.io` from this change. (Originally I intended to have this both make `add_new_collection` available and use/demonstrate it all in one go, but those two steps are separable.) I have a couple ideas for how to make the whole process work but don’t want it to block the rest of the work.

atolopko-czi (approved):
LGTM, now with only core changes in this PR (excluding io.py changes)

Issue #494, opened 2022-11-01
[python] Conda builds


johnkerl:
* Follow https://blog.gishub.org/how-to-publish-a-python-package-on-conda-forge
* `. ~/miniconda/bin/activate`
* `conda skeleton pypi tiledbsc` (use the name `tiledbsc` from https://pypi.org/project/tiledbsc/)
* Get `Error: No source urls found for tiledbsc`
  * https://stackoverflow.com/questions/49735655/conda-skeleton-fails

> Conda Skeleton Requires a source distribution. Uploading a wheel to pypi is not sufficient.
> 
> In addition to a build distribution
> ```
>     python setup.py bdist_wheel --universal
> ```
> create a source distribution
> ```
>     python setup.py sdist
> ```
> 
> before uploading to PyPI

mlin:
Following input from @Shelnutt2 we need to decide between two approaches for packaging libtiledbsoma in conda:

1. Minimal conda-forge recipes build by reference to the PyPI and CRAN packages, each using its own built-in recipe for compiling libtiledbsoma. If you install both tiledbsoma-py and tiledbosoma-r conda packages, then you have two copies -- and potentially different versions -- of libtiledbsoma, possibly needing deconfliction (TBD).
2. Dedicated libtiledbsoma [sub]package for conda, which becomes a dependency of tiledbsoma-py and tiledbsoma-r, and can be shared between them. 

The second approach is cleaner considering the conda package structure on its own. However, in the bigger picture, it'd necessitate maintaining a third distinct way of compiling libtiledbsoma into packaged form, given that the standalone PyPI and R packages must have their own compile recipes for it.

Shelnutt2:
> However, in the bigger picture, it'd necessitate maintaining a third distinct way of compiling libtiledbsoma into packaged form, given that the standalone PyPI and R packages must have their own compile recipes for it.

Can you provide more details on what would need to be changed? `libtiledbsoma` can already be compiled and installed standalone. It's a proper CMake project with install targets so I don't think we need to change anything to allow it to be packaged for conda?

mlin:
[...]

PR #838, opened 2023-01-31
Modify `setup.py` To Find System Installed `libtiledbsoma` On POSIX

nguyenv:
## Issue and/or context:

When `libtiledbsoma` was installed to a system directory, `setup.py` was not able to find the shared library object and build from source. This is one of the first steps we need to solve in order to get conda builds to work.

## Changes:

Ported over certain functions from TileDB-Py's `setup.py` to fix the above issue. The ctypes's `CDLL` function finds global installs of the shared library object. On Linux, `CDLL` only returns a relative path. We need to use ctypes's `dlinfo` to get the absolute path.

## Notes for Reviewer:

N/A

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -64,10 +64,10 @@ def find_libtiledbsoma_full_path_on_linux(lib_name):
 
 def libtiledbsoma_exists():
     """
-    Checks the given list of paths and returns true if any contain the TileDB-SOMA library.
+    Returns the path to the globally installed TileDB-SOMA library, if it exists.
     :return: The path to the TileDB-SOMA library, or None.
     """
-    # Check to see if TileDB is globally installed.
+    # Check to see if TileDB-SOMA is globally installed.
     lib_name = get_libtiledbsoma_library_name()
 
     try:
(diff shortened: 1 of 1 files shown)


apis/python/setup.py
```diff
@@ -27,9 +30,65 @@
...
+    libc = ctypes.CDLL(lib_name)
+    lmptr = ctypes.c_void_p()
+    dlinfo(libc._handle, 2, ctypes.byref(lmptr))
+
+    return ctypes.cast(lmptr, ctypes.POINTER(LINKMAP)).contents.l_name.decode()
+
+
+def libtiledbsoma_exists():
+    """
+    Checks the given list of paths and returns true if any contain the TileDB library.
```
Shelnutt2, line 67: `contain the TileDB library` -> `contain the TileDB-SOMA library`

apis/python/setup.py
```diff
@@ -27,9 +30,65 @@
...
+    :return: The path to the TileDB library, or None.
```
Shelnutt2, line 68: `The path to the TileDB library, or None.` -> `The path to the TileDB-SOMA library, or None.`

Shelnutt2 (approved):
Two minor nits, else lgtm! Great work

nguyenv:
Related to #494.

PR #845, opened 2023-02-01
[c++] Create shared library for conda builds

nguyenv:
## Issue and/or context:

The conda builds require a shared library.

## Changes:

These are changes made by @Shelnutt2 to enable shared library builds. `TILEDBSOMA_BUILD_STATIC` is used to toggle between building a shared library ('OFF') or building a static library ('ON'). By default, `TILEDBSOMA_BUILD_STATIC` is set to `OFF`. 

Note that the shared library currently only works for Python. We plan to support for R in the future. Setting `TILEDBSOMA_BUILD_R` sets  `TILEDBSOMA_BUILD_STATIC` to `OFF`.

## Notes for Reviewer:

1. The unit tests for R have been modified so that the `libtiledbsoma.a` static library is not hardcoded, and the shared library can be more readily tested in the future. 
2. There are two issues with the shared library on R. First, on Linux, the shared library does actually link to the R API and is usable, but there are still issues with macOS. Second, as incidiated by @eddelbuettel, we will need more planning and work to separate out the logger as R uses `spdlog_with_R.h` whereas Python does not.


apis/r/configure
```diff
@@ -39,7 +39,7 @@ else
         cmake -B build -S libtiledbsoma -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/local -DOVERRIDE_INSTALL_PREFIX=OFF -DTILEDBSOMA_BUILD_R=ON > cmake_log.txt
         cmake --build build -j 2 >> cmake_log.txt
         sudo cmake --build build --target install-libtiledbsoma >> cmake_log.txt
-        rm -rf build
+        #rm -rf build
```
eddelbuettel, line 42: I found it helps to remove prior builds.

apis/r/configure
```diff
@@ -39,7 +39,7 @@ else
...
     	export LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
```
eddelbuettel, line 43: BTW that should not be needed (but may not do harm).  On Debian/Ubuntu and alike:

```sh
$ cat /etc/ld.so.conf.d/* | grep /usr/local/lib
/usr/local/lib/i386-linux-gnu
/usr/local/lib/i686-linux-gnu
/usr/local/lib
/usr/local/lib/x86_64-linux-gnu
$ 
```

libtiledbsoma/include/tiledbsoma/spdlog_with_R.h
```diff
@@ -11,8 +11,8 @@
 
 // define R_R_H and USING_R so that REprintf and R_FlushConsole are used
 // by spdlog, and also declare them hereby
-#define R_R_H
-#define USING_R
+//#define R_R_H
+//#define USING_R
```
eddelbuettel, line 15: Can we make this conditional, and/or you just exclude this header file from your Python build?

eddelbuettel (commented):
Some inline comments

apis/r/configure
```diff
@@ -30,9 +36,9 @@ else
         ## else with sources build by hand
         cd src
         mkdir -p build
-        cmake -B build -S libtiledbsoma -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=. -DOVERRIDE_INSTALL_PREFIX=OFF -DTILEDBSOMA_BUILD_R=ON > cmake_log.txt
+        cmake -B build -S libtiledbsoma -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/local -DOVERRIDE_INSTALL_PREFIX=OFF -DTILEDBSOMA_BUILD_R=ON > cmake_log.txt
         cmake --build build -j 2 >> cmake_log.txt
-        cmake --build build --target install-libtiledbsoma >> cmake_log.txt
+        sudo cmake --build build --target install-libtiledbsoma >> cmake_log.txt
```
Shelnutt2, line 41: Do we want sudo here?

apis/r/configure
```diff
@@ -30,9 +36,9 @@ else
         ## else with sources build by hand
         cd src
         mkdir -p build
-        cmake -B build -S libtiledbsoma -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=. -DOVERRIDE_INSTALL_PREFIX=OFF -DTILEDBSOMA_BUILD_R=ON > cmake_log.txt
+        cmake -B build -S libtiledbsoma -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/local -DOVERRIDE_INSTALL_PREFIX=OFF -DTILEDBSOMA_BUILD_R=ON > cmake_log.txt
```
Shelnutt2, line 39: Should we hard code this to `/usr/local`?

.github/workflows/r-ci.yml
```diff
@@ -39,10 +39,18 @@ jobs:
...
         run: cd apis/r && tools/r-ci.sh run_tests
 
+      - name: View Install Output
+        run: cat /home/runner/work/TileDB-SOMA/TileDB-SOMA/apis/r/tiledbsoma.Rcheck/00install.out
+        if: failure() && matrix.os == 'ubuntu-latest'
+
+      - name: View Install Output
+        run: cat /Users/runner/work/TileDB-SOMA/TileDB-SOMA/apis/r/tiledbsoma.Rcheck/00install.out
+        if: failure() && matrix.os == 'macOS-latest'
+
```
eddelbuettel, line 53: Would `cat /*/runner/work/TileDB-SOMA/TileDB-SOMA/apis/r/tiledbsoma.Rcheck/00install.out` for both allowing to collapse this into just one block?  

Or maybe also `cat ~runner/work/....`   or `cat $HOME/runner/...` ?

apis/r/configure
```diff
@@ -10,8 +10,14 @@ if test -z "${R_HOME}"; then
 fi
 
 ### for single-cell-data/apis/r repo, consider completing sources
-if test -f src/lib/libtiledbsoma.a; then
-    echo "** static library present"
+pkg-config --exists tiledbsoma
+if [ $? -eq 0 ]; then
```
eddelbuettel, line 14: If we order it this way, R will first try to find the shared library.  That we now decided should not have the 'R_BUILD' enabled.  This risks breaking R builds.

apis/r/configure
```diff
@@ -10,8 +10,14 @@ if test -z "${R_HOME}"; then
...
+    lib=`pkg-config --libs-only-L tiledbsoma` | cut -d'_' -f 2
+    export LD_LIBRARY_PATH=$lib:$LD_LIBRARY_PATH
+    sudo ldconfig
```
eddelbuettel, line 17: Notwithstanding the previous comment, we should not need the change to LD_LIBRARY_PATH, or the call to `ldconfig`.

apis/r/configure
```diff
@@ -10,8 +10,14 @@ if test -z "${R_HOME}"; then
...
+    echo "** system installed static library present"
+elif test -f src/lib/libtiledbsoma.a; then 
+    echo "** static library present in src/lib"
```
eddelbuettel, line 20: This second block is what is currently used. I think it should remain the default.

apis/r/configure
eddelbuettel: That location also differs from what we currently use.

apis/r/configure
```diff
@@ -45,10 +51,10 @@ have_tiledb="false"
 ## check for pkg-config and use it to inquire about tiledb build options
 pkg-config --version >/dev/null 2>&1
 if [ $? -eq 0 ]; then
-    pkg-config --exists tiledb
+    pkg-config --exists tiledb tiledbsoma
     if [ $? -eq 0 ]; then
-        pkgcflags=`pkg-config --cflags tiledb`
-        pkglibs=`pkg-config --libs tiledb`
+        pkgcflags=`pkg-config --cflags tiledb tiledbsoma`
+        pkglibs=`pkg-config --libs tiledbsoma tiledb`
```
eddelbuettel, line 57: All this is in the configure script for R when we decided overall to not use a shared library with R.  So I think we should also not ask `pkg-config` about the (shared, system) library for tiledbsoma.

eddelbuettel (changes requested):
I am having a few concerns with respect to (possibly unintended) changes to the R build. 

I have not had a chance to test the branch in a container.

nguyenv:
I will revert the R unit test changes to what they were originally.

PR #846, opened 2023-02-01
Remove `__eq__` method from base `TileDBObject` class.

thetorpedodog:
Because SOMA objects have a lot of complex state, saying that they have the same URI isn't really enough for equality, and the existence of the `__eq__` method is more likely to cause confusion if the user has two separate handles open on the same backing object.


atolopko-czi (approved):
Seems reasonable to remove this. I'm not sure if any code currently even uses equality check for any `TileDBObject`s, but assuming not.

thetorpedodog:
Nothing in our tests does, and if there is external code (a) it is easy to fix and (b) it might not have been doing what it was intended to…

Issue #843, opened 2023-01-31
[python] Track object ownership to allow for cascading close


thetorpedodog:
When an child object is created or accessed through the handle of a specific SOMA object, closing the parent should close the child object:

Reading:

```python
experiment = tiledbsoma.Experiment.open("path/to/my/exp")
obs = experiment.obs
# do stuff with obs
# query the experiment
# etc.
experiment.close()  # This should close experiment, obs, etc.
```

Writing:

```python
experiment = tiledbsoma.Experiment.create("path/to/new/exp")
ms = experiment.add_new_collection("ms")
# add things to ms
# ...
experiment.close()  # This should close experiment, ms, everything.
```

PR #847, opened 2023-02-01
[python] Track ownership of accessed and added collection members.

thetorpedodog:
Collections are now responsible for the lifecycle of elements created using `collection.add_new_whatever` and items accessed through `collection["something"]`. This allows close operations to cascade from parent to child and allows a coherent API structure where the user opens a collection, has access to everything inside, and closes everything when they close the collection.

Part of #843.

Huge shout-out to @mlin for introducing `ExitStack`, which I had not heard of before, and which made this work almost trivial.

### DIFF
--- a/apis/python/src/tiledbsoma/collection.py
+++ b/apis/python/src/tiledbsoma/collection.py
@@ -164,6 +164,8 @@ class CollectionBase(
         self._set_element(
             key, cast(CollectionElementType, new_group), use_relative_uri=was_relative
         )
+        # Since we own this object, set it to be closed when we are closed.
+        self._close_stack.enter_context(new_group)
         return new_group
 
     def _not_implemented(self, *args: Any, **kwargs: Any) -> NoReturn:
@@ -209,6 +211,8 @@ class CollectionBase(
                 tiledb_type=storage_type,
                 soma_type=None,
             )
+            # Since we just opened this object, we own it and should close it.
+            self._close_stack.enter_context(entry.soma)
         return cast(CollectionElementType, entry.soma)
 
     def set(
--- a/apis/python/src/tiledbsoma/tiledb_object.py
+++ b/apis/python/src/tiledbsoma/tiledb_object.py
@@ -45,7 +45,14 @@ class TileDBObject(somacore.SOMAObject, Generic[_HandleType]):
         self._handle = handle
         self._metadata = MetadataMapping(self._handle)
         self._close_stack = ExitStack()
+        """An exit stack to manage closing handles owned by this object.
+
+        This is used to manage both our direct handle (in the case of simple
+        TileDB objects) and the lifecycle of owned children (in the case of
+        Collections).
+        """
         self._close_stack.enter_context(self._handle)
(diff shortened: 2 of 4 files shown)


thetorpedodog:
> LGTM.
> 
> Worth also testing a redundant close? I see the comment saying it "is a no-op", but can't hurt to test.

This is in fact covered in the new collection test:

https://github.com/single-cell-data/TileDB-SOMA/blob/a0656c748e382411696877f26eecb077d64f8a46/apis/python/tests/test_collection.py#L283-L288
 
> Out-of-scope concern for this PR, but if we allow parent _open for write_ to have unowned collections added to it, then after the parent closes the children can still be modified. Would the PIT timestamps (the "commit") then not propagate to the parent? That is, we could end up with children having later timestamps than the parent. Is that ok?

Timestamps aren’t propagated up per se; the timestamp of a write on a collection just records when an element was added or removed from it (kind of like the modification time on a filesystem directory). If there is no change to the list of things which is in the collection, there’s nothing to write (by analogy, opening and writing a file does not change the modification time of the directory it lives in). The other (un-owned) array is living on its own timeline, and a reference to it was added to the collection.

Opening a collection for reading at a given timestamp means opening the collection, and then each individual item in the collection, at a snapshot in time of all the writes that took place before that (logical) timestamp. So if you do the following things in wall-time order:

 1. Logical timestamp 10:30: Create array A with data M
 2. Logical timestamp 11:30: Write data N to A
 3. Logical timestamp 11:00:
    - Create collection X
    - Add owned arrays B and C
    - Add un-owned array A
 4. Logical timestamp 12:00: Write data P to B

After all of the above writes are complete, what you will see when you open the dataset at various times:

- 10:45: X does not yet exist (it’s slightly more complicated than this, but this is a good enough approximation).
- 11:15: X contains:
  - A, which has data M
  - B and C
- 11:45: X contains:
  - A, which has data N
  - B and C
- 12:15: X contains:
  - A, which has data N
  - B, which has data P
  - C

In step 4, nothing about group X changes; only B is modified.

Of course, if you do:

 1. Logical timestamp 11:30: Create array A
 2. Logical timestamp 11:00:
    - Create collection X
    - Add un-owned array A

Then things start getting really weird and at that point you’re on your own.

> (I was considering whether it's worth allowing unowned collections to be added in the old manner.)

I think we’re going to need to allow this somehow, but the exact thing you see on the Python API side is up for debate:

- The current model, where the object’s URI is added to the collection, and the exact object you passed in is available at `collection[name]`.
- An ownership-based model, where the object’s URI is added to the collection, and a new handle, owned by and with the same context as the collection is created and is available at `collection[name]`.

atolopko-czi:
@thetorpedodog Thank you for the detailed response on timestamp semantics! Makes sense.  As writer logic is likely to be developed in a more concerted manner than reader logic, it's probably do not need to cover every edge case of odd behaviors that could arise. Careful documentation of the semantics and known edge cases should suffice. Do we have enough details in the abstract spec re: these writer semantics, or if not, do we at least plan to add such docs?

PR #848, opened 2023-02-01
[python] Add `add_new_dataframe` and `_sparse`/`_dense_ndarray`

thetorpedodog:
This completes the `add_new_` methods to allow users to add not just collections but also data objects.  Just like `add_new_collection`, the newly-added members' lifecycle is owned by the collection itself.

### DIFF
--- a/apis/python/src/tiledbsoma/collection.py
+++ b/apis/python/src/tiledbsoma/collection.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import functools
 import re
 import time
 from typing import (
@@ -10,6 +11,7 @@ from typing import (
     Iterator,
     List,
     Optional,
+    Sequence,
     Tuple,
     Type,
     TypeVar,
@@ -18,13 +20,18 @@ from typing import (
 )
 
 import attrs
+import pyarrow as pa
 import somacore
 import tiledb
 from somacore import options
-from typing_extensions import NoReturn
 
+from .common_nd_array import NDArray
+from .constants import SOMA_JOINID
+from .dataframe import DataFrame
+from .dense_nd_array import DenseNDArray
 from .exception import SOMAError
 from .options import SOMATileDBContext
+from .sparse_nd_array import SparseNDArray
 from .tiledb_object import AnyTileDBObject, TileDBObject
 from .types import StorageType, TDBHandle
 from .util import is_relative_uri, make_relative_path, uri_joinpath
@@ -38,6 +45,7 @@ from .util_tiledb import (
 CollectionElementType = TypeVar("CollectionElementType", bound=AnyTileDBObject)
 _Coll = TypeVar("_Coll", bound="CollectionBase[AnyTileDBObject]")
 _Self = TypeVar("_Self", bound="CollectionBase[AnyTileDBObject]")
+_NDArr = TypeVar("_NDArr", bound=NDArray)
 
 
 @attrs.define()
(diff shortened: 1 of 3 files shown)


mlin (approved):

apis/python/src/tiledbsoma/collection.py
```diff
@@ -168,12 +176,70 @@ def add_new_collection(
...
+        *,
+        uri: Optional[str] = None,
+        schema: pa.Schema,
+        index_column_names: Sequence[str] = (SOMA_JOINID,),
+        platform_config: Optional[options.PlatformConfig] = None,
+    ) -> DataFrame:
+        if key in self:
+            raise KeyError(f"{key!r} already exists in {type(self)}")
+        absolute_uri, was_relative = self._new_absolute_uri(key=key, uri=uri)
+        self._check_allows_child(key, DataFrame)
```
atolopko-czi, line 191: nit: keep key-related checks together (move up 1 line)

apis/python/src/tiledbsoma/collection.py
```diff
@@ -168,12 +176,70 @@ def add_new_collection(
...
+        new_df = DataFrame.create(
+            absolute_uri,
+            index_column_names=index_column_names,
+            schema=schema,
+            platform_config=platform_config,
+            context=self.context,
+        )
+        # A dataframe might not be the declared type of this collection,
+        # but we can't really handle that within the type system.
```
atolopko-czi, line 200: curious, what typing check are we unable to validate in `_check_allows_child()`, above?

atolopko-czi (approved):
LGTM! 1 nit, 1 question, but approving as is.

apis/python/src/tiledbsoma/collection.py
thetorpedodog: Let's say you have a `Collection[Measurement]`. You can do `meas_coll.add_new_df(...)` despite the fact that (in theory) the only thing you're *supposed* to be able to add is a `Measurement`. The type system won't warn you that you're about to do something dumb in that case (and we don't have a mechanism to do so at runtime, in part because doing so would add additional complexity).

johnkerl: Uhh ... I didn't touch this file. Just ran `python -m pre_commit run -a -v` ...

apis/python/src/tiledbsoma/collection.py
```diff
@@ -168,12 +176,70 @@ def add_new_collection(
...
+        self._check_allows_child(key, cls)
+        absolute_uri, was_relative = self._new_absolute_uri(key=key, uri=uri)
+        new_arr = cls.create(
+            absolute_uri,
+            type=type,
+            shape=shape,
+            platform_config=platform_config,
+            context=self.context,
+        )
+        # A NDArray might not be the declared type of this collection,
```
johnkerl, line 228: Nit: _An NDArray_ ...

apis/python/tests/test_experiment_query.py
```diff
@@ -43,22 +33,29 @@ def soma_experiment(
...
-        var,
-        X_layer_names,
-        obsp_layer_names,
-        varp_layer_names,
-    )
+    with soma.Experiment.create((tmp_path / "exp").as_posix()) as exp:
+        add_dataframe(exp, "obs", n_obs)
+        ms = exp.add_new_collection("ms")
+        rna = ms.add_new_collection("RNA", soma.Measurement)
+        add_dataframe(rna, "var", n_vars)
```
johnkerl, line 44: I conceptually don't understand why this is not

```
rna.add_dataframe("var", n_vars)
```

It seems to me that the `add_dataframe` et al. which you're creating -- as helper functions solely for unit-test -- are the kinds of helper functions _people would actually want to use_. I don't see why we're hiding this good stuff inside unit-test files instead of delivering it and letting people benefit from it.

More generally I think there's an impedance mismatch  between tools created _by_ software engineers, and tools created _for_ computational biologists. The former might be inclined to say _I'm here to give you flexibility! You can separate make-schema from write-data!_ And the latter might be inclined to say _OK I'll use that some fraction of the time but most of the time I want to do both -- why force me to have to write two separate create-schema and write-data function calls in the happy case which form the bulk of my work?_

cc @mojaveazure @ckrilow @aaronwolen -- please let me know if I'm off-base. I'm trying to advocate for users of these tools, not just creators of these tools -- but I'm also in fact really just another software engineer, not a computational biologist. I may be misprepresenting the needs of our intended users.

apis/python/src/tiledbsoma/collection.py
```diff
@@ -168,12 +176,70 @@ def add_new_collection(
...
+        # An NDArray might not be the declared type of this collection,
+        # but we can't really handle that within the type system.
+        self._set_element(
+            key, new_arr, use_relative_uri=was_relative  # type: ignore[arg-type]
+        )
+        self._close_stack.enter_context(new_arr)
+        return new_arr
+
+    # These are both correct but mypy doesn't get it.
+    add_new_dense_ndarray = functools.partialmethod(  # type: ignore[assignment]
```
johnkerl, line 237: These look exposed -- they're called `add_new_dense_ndarray` and `add_new_sparse_nd_array`, not `_add_new_dense_ndarray` and `_add_new_sparse_nd_array`. But they lack docstrings. They're only used in this PR as called from unit-test files which suggests to me they're user-facing.

There's Python magic going on here which as a non-Python-expert person I don't immediately grasp. But the reader shouldn't have to be a Python expert to grasp what we're doing here.

Please dedicate more wording to explaining, for the rest of us, what it _is_ that's _correct_ and what _is_ the "it" that mypy doesn't get it. Because I don't get it either.

johnkerl (changes requested):

apis/python/src/tiledbsoma/collection.py
thetorpedodog: Added some commentary.

johnkerl: @thetorpedodog thanks for addressing one of my questions!

Why are there not docstrings? Please add them or make clear why these apparently-user-facing methods don't actually need docstrings.

thetorpedodog: The docstrings, while they are not literally inherited or available from the `__doc__` member, can be extracted by tooling from the root `somacore.BaseCollection` object. However, putting them there will be a cue that these are indeed the user-facing functions you're looking for.

johnkerl: @thetorpedodog please create a tracking issue to make sure we don't forget, and link to it here when you've done so.

johnkerl (approved):
I find the online help profoundly confusing but I've filed #858 to track

Issue #851, opened 2023-02-02
[r,python] Create a helper function to comprehensively show version numbers


eddelbuettel:
As seen in the lively discussion in #728 it surprisingly complex to _concisely_ describe the complete software stack for SOMA.  Moreover, how to do so varies between implementation languages.  A helper function `soma_versions()` or `soma_software_versions()` may be helpful.  From #728 we see 

```py
import tiledbsoma, tiledb, sys
tiledbsoma.__version__
tiledb.__version__
sys.version_info
```

and

```r
packageVersion("tiledbsoma")
packageVersion("tiledb")
tiledb::tiledb_version(TRUE)      # for the TileDB Core version used by package tiledb
R.version.string 
```

which one could arrange in a pretty-fied string.

johnkerl:
The PR #855 is merged but once we tag our next release let's update our bug-reporting template to tell people they can use `tiledbsoma.show_package_versions()` (Python) or `tiledbsoma::show_package_versions()` (R)

PR #855, opened 2023-02-02
[r,python] Add `show_package_versions` for bug-report use

johnkerl:
## Issue and/or context:

#851 

## Changes:

As on #851


apis/python/src/tiledbsoma/general_utilities.py
```diff
@@ -38,3 +41,22 @@ def get_storage_engine() -> str:
     [lifecycle: experimental]
     """
     return "tiledb"
+
+
+def show_package_versions() -> None:
+    """
+    Nominal use is for bug reports, so issue filers and issue fixers can be on the same page.
+    """
+    line1 = "tiledbsoma.__version__   " + get_implementation_version()
```
thetorpedodog, line 50: why not print these immediately? no need to make vars, build a list, join the list, then print the result.

apis/python/src/tiledbsoma/general_utilities.py
```diff
@@ -38,3 +41,22 @@ def get_storage_engine() -> str:
...
+    line3 = "python__version__        " + ".".join(
+        [
+            str(e)
+            for e in [
+                sys.version_info.major,
+                sys.version_info.minor,
+                sys.version_info.micro,
+            ]
+        ]
+    )
```
thetorpedodog, line 61:
```python
".".join(str(v) for v in sys.version_info)
```

apis/python/src/tiledbsoma/general_utilities.py
johnkerl: @thetorpedodog d'oh

johnkerl:
@eddelbuettel OK on the R side?

thetorpedodog (approved):
python looks good

apis/r/R/utils-tiledb.R
```diff
@@ -18,3 +18,14 @@ map_query_layout <- function(layout) {
...
+#' show_package_versions
+#'
+#' @description Prints package information suitable for assisting with bug reports.
+#'
+#' @export
+show_package_versions <- function() {
+    cat("tiledbsoma:   ", toString(utils::packageVersion("tiledbsoma")), "\n")
+    cat("tiledb:       ", toString(utils::packageVersion("tiledb")), "\n")
+    cat("R:            ", R.Version()$version.string, "\n")
+}
```
eddelbuettel, line 31: Suggest to change to

```r
#' show_package_versions
#'
#' @description Prints package information suitable for assisting with bug reports.
#'
#' @export
#' @importFrom utils packageVersion
show_package_versions <- function() {
    cat("tiledbsoma:   ", toString(utils::packageVersion("tiledbsoma")), "\n")
    cat("tiledb:       ", toString(utils::packageVersion("tiledb")), "\n")
    cat("tiledb core:  ", tiledb::tiledb_version(TRUE), "\n")
    cat("R:            ", R.version.string, "\n")
}
```

(untested)

eddelbuettel (changes requested):

apis/r/R/utils-tiledb.R
johnkerl: @eddelbuettel added in latest commit

@thetorpedodog I added the core-version line to the Python side as well (great idea!)

eddelbuettel: Cool, thanks, still haven't had a chance to run it so not sure we need `toString()`, part of me thinks that formatting should be automagic in the context.   

Oh, and I just checked, we already have `importFrom(utils,packageVersion)` in `NAMESPACE` so my add in what is now line 27 above wasn't needed (but won't hurt).

eddelbuettel (approved):

apis/r/R/utils-tiledb.R
johnkerl: All good @eddelbuettel , I ran it :)

PR #861, opened 2023-02-02
[python] Support indices of type `bytes`

johnkerl:
## Issue and/or context:

#419

## Changes:

All core-level support now exists. This is just a matter of un-xfailing the relevant unit-test cases.


atolopko-czi (approved):
LGTM

Issue #539, opened 2022-11-16
API: missing ability to introspect given only a URI


bkmartinjr:
The current API has the ability to (examples use Python syntax):
1. introspect the type of an open object (e.g., the `obj.soma_type` property)
2. get a typed handle to an object IF the user already knows the object type (e.g.,  `handle = soma.DenseNdArray(uri)`)
3. given a collection entry, get a typed handle (aka "reify"), e.g., `handle = collection[key]`

However, the API _lacks_ any ability to introspect a URI and determine which underlying type it is.  In other words, you can't do:
```
uri = '...'  # URI to SOMA object of unknown type
soma_handle = soma.open_a_uri_as_soma_object(uri)  # where soma_handle is a SOMACollection, NdArray, etc
```

This makes it challenging to write code that is fully assumption free about SOMA objects/collections.

This missing operation is _implied_ by example (3) above, at least for the current specification where object typing is an object attribute, and the collection will automatically introspect any element in the soma collection, returning a handle.

I believe we should propose a top-level utility function that would expose this functionality, i.e.,:
* given a URI, return the `soma_type` of that URI (or an error if not a SOMA object)
* and/or, open and return a SOMA object representing that URI

Concrete example:

```
uri = '...'

# Lookup soma_type for URI
soma_type = soma.get_soma_type(uri)  # -> "SOMASparseNdArray" or other legal type, or None if not a SOMA object

# open SOMA object for URI
soma_handle = soma.open_uri(uri)  # returns instance of SOMACollection, SOMAIndexedDataFrame, etc, or None

assert soma_type == soma_handle.soma_type
```

Please comment.  CC:  @thetorpedodog @johnkerl @aaronwolen @eddelbuettel @mlin

johnkerl:
@thetorpedodog can we dup this to one of your open tasks? IIUC this functionality will  be appearing soon. (Please correct me if I'm wrong.)

thetorpedodog:
yes, this is as simple as making sure `factory.open` is ready for general consumption and exposing it as `tiledbsoma.open`.

johnkerl:
OK @thetorpedodog I'll leave this open as a checklisting/tracking issue for that.

thetorpedodog:
Reopening after linking this spuriously closed it.

PR #863, opened 2023-02-02
[python] Keep only the read or write TileDB handles open long-term

thetorpedodog:
Context: #638 and #540

This reorganizes the way TileDB handles are managed to ensure that they are only opened for as long as strictly necessary.

Both handle and metadata management are both moved into `handles.py`. The `Handle` object is renamed `Wrapper` (because otherwise it got very confusing) and its construction process has been revamped to store only the single raw TileDB handle it owns (along with metadata). For write handle (wrappers), any reading is done by a short-lived read handle opened onto the same TileDB object.

This also let me overhaul the way `factory.open` works, which I had way overcomplicated on the first go round. It is now much simpler.

---

Once again with apologies to @mlin. This used to reduce >100 lines of code but then I found a bug in metadata handling (that may have already existed?) and added a bunch of tests to characterize it. Then I found some duplicate code that could be merged and got it back to net-negative, which is something.


thetorpedodog:
This change will also in part address https://github.com/single-cell-data/TileDB-SOMA/issues/539; with a couple test cases in a follow-up `factory.open` should be ready for export

thetorpedodog:
I got a little lost in the sauce working on this so in the morning I’m going to spend a few minutes decomposing this into a few smaller commits as best I can. The overall diff should still be the same (maybe with a few more comments).

thetorpedodog:
Finally, [I did it to myself](https://twitter.com/dril/status/873264183281889281):

This branch has conflicts that must be resolved

apis/python/src/tiledbsoma/tiledb_object.py
```diff
@@ -23,9 +25,27 @@ class TileDBObject(somacore.SOMAObject, Generic[_HandleType]):
...
+    @classmethod
+    def open(
+        cls: Type[_Self],
+        uri: str,
+        mode: options.OpenMode = "r",
+        *,
+        context: Optional[SOMATileDBContext] = None,
+        platform_config: Optional[options.PlatformConfig] = None,
+    ) -> _Self:
+        """Opens this specific type of collection."""
```
mlin, line 37: Is this docstring correct?

apis/python/src/tiledbsoma/handles.py
```diff
@@ -0,0 +1,239 @@
+"""Abstractions to more easily manage read and write access to TileDB data.
```
mlin, line 1: Filename- how about smth like `tiledb_handle_wrappers.py`?

mlin (approved):
LGTM subject to two nits inline

The nice clean factorization of the Wrapper.open logic is going to get a little messier when we add timestamps unfortunately, because of the way the timestamps have to be set in starkly different ways on group vs array. But I'm pretty sure we'll be able to contain that within the implementation details of ArrayWrapper/GroupWrapper, and hopefully TileDB-Py will make the group timestamps more straightforward in the future.

apis/python/src/tiledbsoma/tiledb_object.py
thetorpedodog: It is (but only when you are calling this on a Collection type).

apis/python/src/tiledbsoma/handles.py
thetorpedodog: I went with `tdb_handles.py` in the interest of keeping it short and reducing repetition.

PR #869, opened 2023-02-03
[python] Use the array we already have open to generate the Arrow schema

thetorpedodog:
Not related to anything in particular, just something I noticed in the course of other stuff.

### DIFF
--- a/apis/python/src/tiledbsoma/tiledb_array.py
+++ b/apis/python/src/tiledbsoma/tiledb_array.py
@@ -9,7 +9,7 @@ import tiledbsoma.libtiledbsoma as clib
 
 from .options.soma_tiledb_context import SOMATileDBContext
 from .tiledb_object import TileDBObject
-from .util_arrow import get_arrow_schema_from_tiledb_uri
+from .util_arrow import tiledb_schema_to_arrow
 from .util_tiledb import ReadWriteHandle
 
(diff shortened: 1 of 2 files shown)


johnkerl (approved):
absolutely baller 🎉 

cc @gspowley @bekadavis9 @eddelbuettel -- this is one more array-open no longer having to happen in `tiledb-py`, one step closer to the longer-term goal of all array opens happening via `libtiledbsoma`

atolopko-czi (approved):
Looks reasonable!  One Q, one nit, but approving as is.

What happens if the array is closed? Can the schema still be accessed w/o reopening (as the earlier implementation would have done)?

Nit: we should update the docstring in `_tiledb_array_schema()`, which reads "Returns the TileDB array schema. Not part of the SOMA API; for dev/debug/etc.".

thetorpedodog:
> Looks reasonable! One Q, one nit, but approving as is.
> 
> What happens if the array is closed? Can the schema still be accessed w/o reopening (as the earlier implementation would have done)?

The schema cannot be accessed while the array is closed, no; however I think this is a pretty reasonable limitation (i.e., you can only do things with an opened array).

> Nit: we should update the docstring in `_tiledb_array_schema()`, which reads "Returns the TileDB array schema. Not part of the SOMA API; for dev/debug/etc.".

Updated. Thank you!

PR #870, opened 2023-02-03
[python] Directly use backend TileDB dimension names for `DataFrame`

thetorpedodog:
We were pre-caching the dimension names for the DataFrame because we were previously opening and closing the TileDB array handle (if it was not already open) when they were requested. However, since we now have a long-lived handle that we keep around, that is no longer needed.

---

Based on @bkmartinjr’s [a comment on an earlier review](https://github.com/single-cell-data/TileDB-SOMA/pull/824#discussion_r1089607732):

> I'm not sure caching the index column names makes any sense given the change to stateful handles. I believe that it was originally put in place to avoid repeated calls to open on the array (just to get the dimension names). With an open handle guaranteed, getting dim names is cheap, ergo caching is not worth the complexity.

### DIFF
--- a/apis/python/src/tiledbsoma/dataframe.py
+++ b/apis/python/src/tiledbsoma/dataframe.py
@@ -61,24 +61,8 @@ class DataFrame(TileDBArray, somacore.DataFrame):
         handle = cls._create_internal(uri, tdb_schema, context)
         return cls(
             handle,
-            _index_column_names=tuple(index_column_names),
             _dont_call_this_use_create_or_open_instead="tiledbsoma-internal-code",
         )
-
-    def __init__(
-        self,
-        handle: tiledb.Array,
-        *,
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -158,7 +138,6 @@ def read(
         * Negative indexing is unsupported.
         """
         del batch_size, partitions, platform_config  # Currently unused.
-        self._check_open_read()
```
atolopko-czi, line 161: Is this related to the PR?

thetorpedodog: This was something from another change that leaked into this one. Removing (un-removing?) it.

atolopko-czi (approved):
Less code! 👏

PR #873, opened 2023-02-03
[python] Fix ingest to cloud URIs (part 1 of 2)

johnkerl:
## Issue and/or context:

In conjunction with #863, this allows ingest to tiledb-cloud URIs to again succeed.

```
uri = 'tiledb://johnkerl-tiledb/s3://tiledb-johnkerl/scratch/test-cloud-20230201'
soma = tiledbsoma.SOMA(uri)
tiledbsoma.io.from_h5ad(soma, '/tmp/test.h5ad')
```

Note this test scenario does not pass until and unless #863 is also included. I'm putting this up as a PR on `main` instead of atop #863 at the explicit request of @thetorpedodog .

## Note

Since this is an open-source package we do not have unit tests taking the `tiledb.cloud` dependency.

### DIFF
--- a/apis/python/src/tiledbsoma/util_tiledb.py
+++ b/apis/python/src/tiledbsoma/util_tiledb.py
@@ -84,7 +84,12 @@ def is_does_not_exist_error(e: tiledb.TileDBError) -> bool:
     stre = str(e)
     # Local-disk/S3 does-not-exist exceptions say 'Group does not exist'; TileDB Cloud
     # does-not-exist exceptions are worded less clearly.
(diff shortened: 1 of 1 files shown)


thetorpedodog (approved):
Something like "improve detection of nonexistent array errors from TileDB Cloud" might be a better change description, since that's the broader effect of the change (with an addendum that the eventual goal is that this will help fix ingestion to TileDB Cloud URIs). The code itself is good.

johnkerl:
Thanks @thetorpedodog !!

Issue #876, opened 2023-02-04
[FEATURE REQUEST] NDArray `create` does not allow override of all TileDB schema params


bkmartinjr:
**Is your feature request related to a problem? Please describe.**

The `platform_config`  parameter of NDArray.create() does not allow the user to override several (potentially useful) TileDB schema attributes. 
Request that these be enabled.  The missing items I noticed as hard-wired, and which should be over-ridable.
* `allows_duplicates` 
* `validity_filters`

**Describe the solution you'd like**

Expand support of platform_config to incorporate these additional parameters.

bkmartinjr:
Upon further inspection, it appears that DataFrame has the same issue - several schema parameters are hard-wired and can not be overridden with `platform_config`

PR #878, opened 2023-02-06
[python] No allow-duplicates for sparse N-d arrays

### LINKED ISSUE #875: [BUG] Python regression -- `SparseNDArray` allows duplicates
**Describe the bug**
Recent releases of `tiledbsoma` `SparseNDArray` create underlying TileDB arrays with `allows_duplicates=True`, which is a regression from the spec and past behavior.

**To Reproduce**

```python
In [7]: A = soma.SparseNDArray.create("./foo", type=pa.float32(), shape=(10, 100))

In [8]: A.uri
Out[8]: './foo'

In [9]: with tiledb.open(A.uri) as ta: print(ta.schema)
ArraySchema(
  domain=Domain(*[
    Dim(name='soma_dim_0', domain=(0, 9), tile=10, dtype='int64', filters=FilterList([ZstdFilter(level=3), ])),
    Dim(name='soma_dim_1', domain=(0, 99), tile=100, dtype='int64', filters=FilterList([ZstdFilter(level=3), ])),
  ]),
  attrs=[
    Attr(name='soma_data', dtype='float32', var=False, nullable=False, filters=FilterList([ZstdFilter(level=-1), ])),
  ],
  cell_order='row-major',
  tile_order='row-major',
  capacity=100000,
  sparse=True,
  allows_duplicates=True,
)
```

**Versions (please complete the following information):**

Latest main branch HEAD.

**Additional context**

The code is in `common_nd_array.py`, and seems to have been intentional - but is definitely a regression. Need a test case to control for this in future revisions.

johnkerl:
## Issue and/or context:

Fixes #875

## Changes:

Undoes the inadvertent https://github.com/single-cell-data/TileDB-SOMA/pull/681/commits/fa08391e68cc1555275a92362c1971c189c9acab#diff-c57e3497da1ff73a9f6544da46c2129f132e4c4a73dbc4c96ee972d9fb3bdf39L112

## Notes for Reviewer:

Does not address #876, which will be handled on a separate PR.

### DIFF
--- a/apis/python/src/tiledbsoma/common_nd_array.py
+++ b/apis/python/src/tiledbsoma/common_nd_array.py
@@ -122,11 +122,13 @@ def build_tiledb_schema(
 
     cell_order, tile_order = create_options.cell_tile_orders()
 
+    # TODO: accept more TileDB array-schema options from create_options
+    # https://github.com/single-cell-data/TileDB-SOMA/issues/876
     return tiledb.ArraySchema(
         domain=dom,
         attrs=attrs,
         sparse=is_sparse,
-        allows_duplicates=is_sparse,
+        allows_duplicates=False,
         offsets_filters=create_options.offsets_filters(),
         capacity=create_options.get("capacity", 100000),
         tile_order=tile_order,
--- a/apis/python/tests/test_dense_nd_array.py
+++ b/apis/python/tests/test_dense_nd_array.py
@@ -41,6 +41,10 @@ def test_dense_nd_array_create_ok(
         assert a.schema.field(f"soma_dim_{d}").type == pa.int64()
     assert a.schema.field("soma_data").type == element_type
 
+    # Validate TileDB array schema
+    with tiledb.open(tmp_path.as_posix()) as A:
+        assert not A.schema.sparse
+
 
 @pytest.mark.parametrize("shape", [(10,)])
 @pytest.mark.parametrize("element_type", NDARRAY_ARROW_TYPES_NOT_SUPPORTED)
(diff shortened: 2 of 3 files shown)


atolopko-czi (approved):
LGTM

apis/python/src/tiledbsoma/common_nd_array.py
```diff
@@ -122,11 +122,13 @@ def build_tiledb_schema(
...
     cell_order, tile_order = create_options.cell_tile_orders()
 
+    # TODO: accept more TileDB array-schema options from create_options
+    # https://github.com/single-cell-data/TileDB-SOMA/issues/876
     return tiledb.ArraySchema(
         domain=dom,
         attrs=attrs,
         sparse=is_sparse,
-        allows_duplicates=is_sparse,
+        allows_duplicates=False,
```
bkmartinjr, line 131: I would much prefer this was:
> allows_duplicates=create_options.get("allows_duplicates", False),

That would allow user control over this param, much like the other platform-dependent configuration.  I realize this _might_ allow the user to do sketchy things from a SOMA standpoint, but there are very legit use cases (mostly performance related, where datasets are known a priori to not contain dups)

johnkerl: @bkmartinjr above I wrote

> # Notes for Reviewer:
> 
> Does not address https://github.com/single-cell-data/TileDB-SOMA/issues/876, which will be handled on a separate PR.

johnkerl: I believe this answers your question -- ? Please let me know if it does not.

bkmartinjr: I was making a statement - there is no question to answer :-)

I understand you are planning a separate PR.

atolopko-czi: I thought we had to decided to keep allow_dups as non-parameterized and deal with overriding this via schema evolution? I'm not opposed to parameterizing it, though.

bkmartinjr: I don't recall that decision - I thought it was only that the default (and implicit behavior) was 'no-dups'.  At least in our use case, there is a significant performance win from having a higher layer (app) take responsibility for the no-dups requirement, rather than doing it at the lower (TileDB) layer.  Allowing for that seems important.

PR #879, opened 2023-02-06
[r] Validate no allow-duplicates for sparse N-d arrays and dataframes

johnkerl:
## Issue and/or context:

#875 

## Changes:

Counterpart to Python PR #878 

## Notes for Reviewer:

Does not address https://github.com/single-cell-data/TileDB-SOMA/issues/876, which will be handled on a separate PR.

### DIFF
--- a/apis/python/src/tiledbsoma/common_nd_array.py
+++ b/apis/python/src/tiledbsoma/common_nd_array.py
@@ -122,11 +122,13 @@ def build_tiledb_schema(
 
     cell_order, tile_order = create_options.cell_tile_orders()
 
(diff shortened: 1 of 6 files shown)


eddelbuettel (approved):
The new test predicates are fine. 

(Minor aside: The temporary test directories are ephemeral, changing the name seems fairly unimportant to me as does changing them so :man_shrugging:  ... )

johnkerl:
@eddelbuettel thank you!

We changed from `SOMAIndexedDataFrame` to `SOMADataFrame` some time ago -- having the test names more closely match the code is an (admittedly very minor) positive step in general neatening toward 1.0 release.

PR #882, opened 2023-02-06
Pull in `somacore==0.0.0a12` and associated improvements.

thetorpedodog:
- Enforces use of object-identity–based equality for SOMA collections.
- Adds support for `bytes` and `str` in experiment axis queries.
- Improves type annotations for better static analysis.


thetorpedodog:
merging because R failures are unrelated (also: why did it run here?)

Issue #877, opened 2023-02-04
[BUG] Python -- `NDArray` dims zstd level uses `string_dim_zstd_level`


bkmartinjr:
**Describe the bug**

The schema-creation code for NDArray dimensions utilizes the `string_dim_zstd_level` for default compression level, when the dimensions _must_ be int64.   This may lead to unexpected behavior if the user expects that setting the config value for `string_dim_zstd_level` will only affect string dimensions.

Code in question is in `common_nd_array.py::build_tiledb_schema`:

```python
            filters=create_options.dim_filters(
                dim_name,
                [
                    dict(
                        _type="ZstdFilter",
                        level=create_options.string_dim_zstd_level(),
                    )
                ],
            ),
```

**Versions (please complete the following information):**

Latest main head

**Additional context**

There is currently no default config param for int64 dimensions.  Fixes to this might include:

* change the config key name to "default_dim_std_level"
* add a config specifically for soma_joinid/int64 indexed dimensions

PR #883, opened 2023-02-06
[python] `string_dim_zstd_level` -> `int64_dim_zstd_level`

johnkerl:
## Issue and/or context:

#877

### DIFF
--- a/apis/python/src/tiledbsoma/dataframe.py
+++ b/apis/python/src/tiledbsoma/dataframe.py
@@ -338,7 +338,7 @@ def _build_tiledb_schema(
                 [
                     dict(
                         _type="ZstdFilter",
-                        level=tiledb_create_options.int64_dim_zstd_level(),
+                        level=tiledb_create_options.dataframe_dim_zstd_level(),
                     )
                 ],
             ),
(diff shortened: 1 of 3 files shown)


apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -338,7 +338,7 @@ def _build_tiledb_schema(
                 [
                     dict(
                         _type="ZstdFilter",
-                        level=tiledb_create_options.string_dim_zstd_level(),
+                        level=tiledb_create_options.int64_dim_zstd_level(),
```
bkmartinjr, line 341: the dtype of DataFrame dimensions are not hard-wired to any given primitive type - they could be string, int32, float, etc.  This really should be a more generic name.

Perhaps:  `default_datarame_dim_zstd_level`?

bkmartinjr: In keeping with that idea, maybe the config shoudl jsut have two defaults:
* DataFrame.default_dim_zstd_level
* SparseNDArray.default_dim_zstd_level

bkmartinjr (commented):
nits (OK to skip if you prefer):
* missing docstrings
* missing unit test (i.e., testing the ability to change the config)

johnkerl:
@bkmartinjr the doc-string and unit-test coverage for options issues are bigger than this PR. I've created #885, and re-opened #602, to track these.

bkmartinjr (approved):
LGTM

PR #887, opened 2023-02-06
[python] Capture stdout/stderr from `stats_dump` output

thetorpedodog:
Because of missing `stats_reset` calls in previous code, there would be huge volumes (megabytes) of stats output if you did `pytest -vs ./tests` with no options. This captures the output from `stats_dump` in a fixture and verifies its presence, and removes debugging stats management calls.

### DIFF
--- a/apis/python/tests/test_util_tiledb.py
+++ b/apis/python/tests/test_util_tiledb.py
@@ -1,13 +1,29 @@
+import pyarrow as pa
 import pytest
 
 import tiledbsoma
 
 
-def test_stats(capfd: pytest.CaptureFixture[str]):
+def test_stats(tmp_path, capfd: pytest.CaptureFixture[str]):
     """Make sure these exist, don't throw, and write correctly."""
     tiledbsoma.stats_enable()
-    tiledbsoma.stats_dump()
     tiledbsoma.stats_reset()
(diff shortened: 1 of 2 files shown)


johnkerl (changes requested):
This is OK -- and -- I'd prefer we _also_ do:

* Do a reset
* Enable stats
* Do some trivial query
* Get stats

No reason to do compute-and-discard of MB of info when we could be doing compute-and-discard of KB of info

-- and --

thanks for doing this -- good catch -- this issue is my fault & I appreciate you dealing with it

(You've got a lot on your plate right now & tbh I'd have preferrred you had asked me to deal with this once you noticed the problem -- but, thanks again)

johnkerl:
> Re my request changes: my strong preference is you let me put up a commit focusing on my requested changes -- you have a lot to do right now that you are the best person to do -- this PR isn't it & it would be better I finished off this PR & you work on awesomepaulthings

Changes made on commit

https://github.com/single-cell-data/TileDB-SOMA/pull/887/commits/5013f5908d502964d503f68a388c5a97a3442de8

johnkerl (approved):

thetorpedodog:
looks good, merging

PR #888, opened 2023-02-06
[python] Add additional create-time `platform_config` hooks for schema mods

### LINKED ISSUE #876: [FEATURE REQUEST] NDArray `create` does not allow override of all TileDB schema params

bkmartinjr:
## Issue and/or context:

Fixes #876

The create `platform_config` options did not support a handful of useful TileDB schema options, most notably allow_duplicates.  While SOMA requires that there are no duplicates in SparseNDArray dimensions or DataFrame joinids, it does not specify where this must be enforced/implemented.  There are performance-inspired use cases where an application may wish to enforce this constraint at _write_ time, rather than having TileDB enforce it a _read_ time.

## Changes:
Add create platform_config support for:
* allows_duplicates
* validity_filters

## Notes for Reviewer:
None


johnkerl (changes requested):
#878 and #880 (and #879) added some quick unit-test checks validating the platform config that goes into a create against the tiledb schema that comes out -- we should have a brief unit-test case here too expanding on the same theme

CC @Shelnutt2 regarding maybe we need a big red warning sign for new users of the system:

* Your use-case for `allow_duplicates=True` is important, and you understand it, and it needs to be supported. (This is an important PR and it needs to move forward.)
* That said, while other options such as compression level result in performance changes (at write time and/or read time), they result in the same readback semantics. Not so `allow_duplicates` -- it helps for read-time performance _and also modifies read-time semantics_. Users who are less contextualized than you need a clear warning to use this only if they understand what it does.

Shelnutt2 (approved):

bkmartinjr:
@johnkerl - I think the warning is a bad idea, unless you plan to do it for all other sharp edges (and I don't think that is possible).  Platform-specific extensions are just that - use with knowledge and caution.

I will add a couple of test cases.

johnkerl (approved):
🚢

PR #889, opened 2023-02-07
[python] Refactor processing of URIs for collection creation/setting

thetorpedodog:
This separates the way URIs are processed when creating a new child element from the way they are processed when adding an existing SOMA object into the collection.

When creating a new child element, we already know the new relative or absolute URI that we want to use to set it as a member, but we might need to figure out what URI to use to create the new element.

    coll.add_new_thing("new_key", ...)
    # We want the new thing to be added at the relative URI `new_key`.
    # We need to figure out the absolute URI to create it at.
    coll.add_new_thing("abs_key", uri="file:///abs/path")
    # We want the new thing to be created at `file:///abs/path`.
    # We want to add it to the collection with that absolute URI.

Conversely, when adding an existing object into a collection, we know the absolute URI of the collection that we are adding, but we *don't* know the relative URI we might want to use to set the object as a member of the collection.

    to_add = tiledbsoma.open("file:///path/to/something")
    coll["new_key"] = to_add
    # We know the absolute path of `to_add`.
    # We need to figure out whether to use a relative URI,
    # and what URI to use, depending upon our own URI.

In both cases, `tiledb://` URLs do require some special handling.

### DIFF
--- a/apis/python/src/tiledbsoma/collection.py
+++ b/apis/python/src/tiledbsoma/collection.py
@@ -5,6 +5,7 @@ import re
 import time
 from typing import (
     Any,
+    Callable,
     ClassVar,
     Dict,
     Iterator,
@@ -38,6 +39,7 @@ from .util_tiledb import is_does_not_exist_error, is_duplicate_group_key_error
 
 # A collection can hold any sub-type of TileDBObject
 CollectionElementType = TypeVar("CollectionElementType", bound=AnyTileDBObject)
+_TDBO = TypeVar("_TDBO", bound=AnyTileDBObject)
 _Coll = TypeVar("_Coll", bound="CollectionBase[AnyTileDBObject]")
 _Self = TypeVar("_Self", bound="CollectionBase[AnyTileDBObject]")
 _NDArr = TypeVar("_NDArr", bound=NDArray)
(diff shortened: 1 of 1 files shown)


johnkerl (changes requested):
it's 8pm my time and i want to look at this tomorrow

there may be nothing to change but i don't want this to be merged on an accept until i look at it -- i don't play this card on _all_ PRs but i play it on a _few_ PRs and this is one of them

johnkerl:
* The workpath toward storing creation URIs in metadata is interesting but needs to be first brought up for discussion either in the Friday spec-discussion meeting, or, in a company-internal discussion.
* There are no new unit-test cases here -- demonstrating nothing was broken -- but having no PR at all also breaks nothing. To demonstrate the value-add, this PR needs to have additional test cases which test-drive and validate the functionality that's introduced here.
* There's no tracking issue associated with this PR leading me to think this is not in the critical path for 1.0rc0. Let's focus on closing our #638 and #540 this week.

bkmartinjr:
@johnkerl - the _original_ code was also not tested, as is many other areas (e.g. soma.io). I do not think it is reasonable to selectively change the acceptance criteria without broader team discussion. Given that we have hit our nominal coverage goal, I suggest you use the criteria of not regressing on coverage, at least for the RC.  I am a *huge* fan of tests, and want to see them soon, but getting these critical path PRs landed seems highest priority to me.

johnkerl:
@bkmartinjr I don't see this as a critical-path PR -- if either you or @thetorpedodog can help me understand, I would appreciate it.

johnkerl (approved):
@thetorpedodog @bkmartinjr : 

* This was presented to me by @thetorpedodog (outside of GitHub) as a workpath toward modifying the UX to get rid of the `use_relative_uri` parameter.
  * I stand by my bullet items above, namely, we should first have a discussion about intended UX; this sounds nice but is not critical-path work for 1.0rc0; if this makes possible something that wasn't possible before, let's demonstrate its use
  * If this is indeed a step on that workpath, I stand by my reject -- the UX can be discussed, and PRs can be done after 1.0rc0
* This is presented to me by @bkmartinjr as a critical-path item
  * Indeed, this refactor does not break anything
  * If indeed this does unblock critical-path work for 1.0rc0, I'd like to understand how

I will accept to unblock -- with the understanding that this is critical-path work as presented by @bkmartinjr and not a UX-update-by-PR as presented by @thetorpedodog

bkmartinjr:
> If indeed this does unblock critical-path work for 1.0rc0, I'd like to understand how

I view Paul's massive code improvements (clarity, factoring) as important for 1.0.  Critical is an overstatement on my part.

Given that it is already done, I'm not sure why you are opposed to it at this point.

thetorpedodog:
wait there was one more thing to check. not merging YET.

thetorpedodog:
> wait there was one more thing to check. not merging YET.

This “thing to check” is that I was advised that TileDB Cloud *may* support adding members via relative URI (e.g. if you have `tiledb://myns/backend://path/to/whatever`, you might be able to add `tiledb://myns/backend://path/to/whatever/else` with the relative URI `else`) but after testing, this is not the case.

PR #890, opened 2023-02-07
Add static `exists` method to TileDBObject.

### LINKED ISSUE #800: [python] Backwards compat with old `soma_object_type values` is broken
Late 2022, the TileDB metadata vales for dense and sparse NDArray were changed:
* DenseNdArray became DenseNDArray
* SparseNdArray became SparseNDArray

Backwards compatiability was maintained in the factory.py reification code.

However, backwards compat was broken in TiledBObject.exists(), which only supports the most recent metadata values allowed in the `soma_object_type` metadata key.

We should either:
* remove support for the legacy metata keys from factory.py
* or, support them everywhere (eg, in exists()).

To reproduce:
* create and save soma.Collection, containing a soma.DenseNDArray.
* using tiledb, rewrite the metadata for the dense array so that `soma_object_type` is `SOMADenseNdArray`
* open the dense array (eg., `a = soma.DenseNDArray(uri)`)
* the exists method will return the wrong value (eg, `a.exists()` returns False)

The bug is in the `exists()` method in TileDBObject, which does not properly account for the old `soma_object_type` values.

thetorpedodog:
Fixes #800.

### DIFF
--- a/apis/python/src/tiledbsoma/tiledb_object.py
+++ b/apis/python/src/tiledbsoma/tiledb_object.py
@@ -6,6 +6,7 @@ import tiledb
 from somacore import options
 
 from . import constants, tdb_handles
+from .exception import SOMAError
 from .options import SOMATileDBContext
 
 _WrapperType_co = TypeVar(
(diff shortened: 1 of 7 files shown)


bkmartinjr (approved):
This change looks good.

However, it raises an (optional, may be ignored) comment - much like we want to add a `soma.open()` generic factory, we should also consider a `soma.exists()` generic version of this method (eg, it would be equivalent to `soma.open(uri).exists()`)

PR #892, opened 2023-02-07
[python] Add timestamp slots in `SOMATileDBContext` and apply throughout

mlin:
#540 

Add `read_timestamp` and `write_timestamp` slots to `SOMATileDBContext`. Then, feed these through to all read and write operations (including the C++ `SOMAReader`). `read_timestamp` defaults to the moment when the `SOMATileDBContext` is initialized.


apis/python/src/tiledbsoma/tdb_handles.py
```diff
@@ -81,7 +83,11 @@ def open(
     @classmethod
     @abc.abstractmethod
     def _opener(
-        cls, uri: str, mode: options.OpenMode, context: SOMATileDBContext
+        cls,
+        uri: str,
+        mode: options.OpenMode,
+        context: SOMATileDBContext,
+        auxiliary_reader: bool = False,
```
thetorpedodog, line 90: It might make sense to call this `use_latest_read_timestamp` instead of `auxililary_reader`, so that it’s clearer what the effect, and add the comment about *why* we do it at line 71.

apis/python/src/tiledbsoma/options/soma_tiledb_context.py
```diff
@@ -29,3 +30,64 @@ class SOMATileDBContext:
...
+    # (internal) tiledb.Ctx specifically for tiledb.Group operations; unlike arrays, tiledb.Group
+    # needs its timestamps set in the tiledb.Ctx.
+    _group_read_tiledb_ctx: tiledb.Ctx = attrs.field(init=False)
+    _group_write_tiledb_ctx: tiledb.Ctx = attrs.field(init=False)
+
+    def __attrs_post_init__(self) -> None:
+        "initialization hook invoked by the attrs-generated __init__"
+
+        if self.read_timestamp < 0:
+            object.__setattr__(self, "read_timestamp", int(time.time() * 1000))
```
thetorpedodog, line 68: This might be better accomplished with a `factory` function:

```python
read_timestamp = attrs.field(type=int, factory=lambda: int(time.time() * 1000))
```

apis/python/src/tiledbsoma/options/soma_tiledb_context.py
```diff
@@ -29,3 +30,64 @@ class SOMATileDBContext:
...
+        assert (
+            self.read_timestamp_start >= 0
+            and self.read_timestamp >= self.read_timestamp_start
+        )
+        assert self.write_timestamp is None or self.write_timestamp >= 0
```
thetorpedodog, line 73: and similarly, you can use a validator function here:

```python
@read_timestamp.validator
def _validate_timestamps(self, _: Any, __: Any) -> None:
    ...  # do checks in here
```
(despite the fact that it's decorated as `@read_timestamp.validator`, it is run after initialization is complete and so has access to the full state of `self`. cf. [`@i.validator` on `options.IOfN`](https://github.com/single-cell-data/SOMA/blob/6df198854cd5c91da51640b3e93d9e6f7cddc961/python-spec/src/somacore/options.py#L43-L49))

The exceptions should probably also be raised as `ValueError`s, since these are user-settable values and not purely internal consistency checks.

apis/python/src/tiledbsoma/options/soma_tiledb_context.py
```diff
@@ -29,3 +30,64 @@ class SOMATileDBContext:
...
+
+        # prepare the pair of tiledb.Ctx for groups
+        group_read_config = self.tiledb_ctx.config().dict()
+        group_read_config["sm.group.timestamp_start"] = self.read_timestamp_start
+        group_read_config["sm.group.timestamp_end"] = self.read_timestamp
+        object.__setattr__(
+            self, "_group_read_tiledb_ctx", tiledb.Ctx(group_read_config)
+        )
```
thetorpedodog, line 81: Given this, I think it makes sense to unmark `frozen=True`, and possibly add a note in the docstring saying "don't mutate this after creation" (which the spec already says the user should not do). I'm not sure the little bit of safety we get from `frozen` is worth the hoops we're jumping through here.

johnkerl (approved):
Nothing to add on top of @thetorpedodog 's great, detailed feedback :)

apis/python/src/tiledbsoma/options/soma_tiledb_context.py
mlin: Thanks @thetorpedodog; incorporated all of the above but I kept `frozen` because I do think it's helpful to guard against mutation of `SOMATileDBContext` when it might be shared by a lot of different SOMA objects. Also all the group-specific tiledb.Ctx stuff, I hope we can get rid of at some point in the future when `tiledb.Group()` may accept timestamps directly like `tiledb.Array()`.

johnkerl:
🚀

apis/python/tests/test_sparse_nd_array.py
```diff
@@ -921,3 +922,58 @@ def test_create_platform_config_overrides(
...
+    # read with (timestamp_start, timestamp_end) = (15, 25) & see only the second write
+    with soma.SparseNDArray.open(
+        tmp_path.as_posix(),
+        context=SOMATileDBContext(read_timestamp_start=15, read_timestamp=25),
+    ) as a:
+        assert a.read().coos().concat().to_scipy().todense().tolist() == [
+            [0, 0],
+            [0, 1],
+        ]
+        assert a.nnz == 1
```
atolopko-czi, line 979: Rather than having each of the PIT read tests on individual objects, it could be worthwhile to place all of them together in a feature-specific test module, `test_point_in_time_semantics.py` (e.g.).  This makes it clear that we have many related tests for the various TileDBObjects that should be considered together when making future changes to PIT logic.  This would also allow us to DRY up the test code for the dense & spare array cases, by using a single pytest test method using `@pytest.mark.parametrize` to parameterize the classes/objects for the two cases.

atolopko-czi (commented):
A few minor, post-merge comments. If any resonate, can add to new issue(s).

Issue #872, opened 2023-02-03
[python] Top-level namespace needs review


bkmartinjr:
The top-level exported namespace for the python package needs refinement.  Current status as of `0.5.0a8` contains some private stuff (eg, TileDBArray), some missing items (eg, #791, and possibly some items from `.options`), etc.  Our prior design discussion concluded that any name _commonly_ used by end-users should be promoted to the top level.  Uncommonly used items could stay within sub-namespaces.

A particular gap is the re-export of some useful `somacore` names (one example is highlighted in #791).  Another pair we use in `cell_census` are:
```
from somacore.query.query import AxisColumnNames
from somacore.options import SparseDFCoord
```

Current state of the top-level package:

```
In [2]: dir(soma)
Out[2]: 
['AxisQuery',
 'Collection',
 'DataFrame',
 'DenseNDArray',
 'DoesNotExistError',
 'Experiment',
 'ExperimentAxisQuery',
 'Measurement',
 'MetadataMapping',
 'SOMAError',
 'SparseNDArray',
 'TileDBArray',
 'TileDBObject',
 '__all__',
 '__builtins__',
 '__cached__',
 '__doc__',
 '__file__',
 '__loader__',
 '__name__',
 '__package__',
 '__path__',
 '__spec__',
 '__version__',
 'collection',
 'constants',
 'dataframe',
 'dense_nd_array',
 'exception',
 'experiment',
 'general_utilities',
 'get_SOMA_version',
 'get_implementation',
 'get_implementation_version',
 'get_storage_engine',
 'libtiledbsoma',
 'measurement',
 'metadata_mapping',
 'options',
 'query_condition',
 'sparse_nd_array',
 'tiledb_array',
 'tiledb_object',
 'types',
 'util',
 'util_arrow',
 'util_iter',
 'util_tiledb']
```

johnkerl:
@bkmartinjr some things have been done since 0.5.0a8 (in `HEAD`); more to do, though -- thanks for pointing this out.

johnkerl:
@bkmartinjr here is `HEAD` as of right now:

import tiledbsoma as soma

module = soma
# module = soma.options

for e in sorted(dir(module)):
    if not e.startswith("__"):
            #print(e)
            print(e, type(module.__getattribute__(e)))

[then some hand-editing to organize by type]

AxisQuery                  <class 'type'>
DoesNotExistError          <class 'type'>
ExperimentAxisQuery        <class 'type'>
[...]

PR #871, opened 2023-02-03
[python] Add user-level stats hooks, to reach parity with tiledbsoma-r

johnkerl:
## Issue and/or context:

Issue: #514 

## Changes:

Bring this one step closer to the user, following the R lead at #865

PR #865, opened 2023-02-03
[r] Add stats API accessors

eddelbuettel:
#### Issue and/or context:

Issue #514 asked for stats support from the TileDB engine, this had fallen to the side because I had conflated it with requiring an interface to the query object (which it has: for subsets of the data).   

#### Changes:

This PR adds four direct interfaces to the TileDB::Stats class and compiles them directly.  We could equally well call the same functionality from the tiledb-r package but it may be preferable to minimise access to it.  

We return a JSON string and do not parse it for the user.  This could be added, but would require a whole new dependency, 

#### Notes for Reviewer:

See https://app.shortcut.com/tiledb-inc/story/25087/expose-stats-api and https://github.com/single-cell-data/TileDB-SOMA/issues/514

#### Demo

<details>

```r
> library(tiledbsoma)
> uri <- "/tmp/tiledb/penguins"    # local array used for testing
> sdf <- SOMADataFrame$new(uri)
> sdf$stats_enable()
> arr <- sdf$read()
> txt <- sdf$stats_dump()
> cat(txt, "\n")
[
  {
    "timers": {
      "Context.StorageManager.sm_load_fragment_metadata.sum": 0.000637428,
      "Context.StorageManager.sm_load_fragment_metadata.avg": 0.000637428,
      "Context.StorageManager.sm_load_array_schemas_and_fragment_metadata.sum": 0.00142219,
      "Context.StorageManager.sm_load_array_schemas_and_fragment_metadata.avg": 0.00142219,
      "Context.StorageManager.sm_load_array_schema_from_uri.sum": 0.000610096,
      "Context.StorageManager.sm_load_array_schema_from_uri.avg": 0.000610096,
      "Context.StorageManager.sm_load_all_array_schemas.sum": 0.000683514,
      "Context.StorageManager.sm_load_all_array_schemas.avg": 0.000683514,
      "Context.StorageManager.array_open_read_load_schemas_and_fragment_meta.sum": 0.00143929,
      "Context.StorageManager.array_open_read_load_schemas_and_fragment_meta.avg": 0.00143929,
      "Context.StorageManager.array_open_read_load_directory.sum": 0.00197803,
      "Context.StorageManager.array_open_read_load_directory.avg": 0.00197803,
      "Context.StorageManager.array_open_READ.sum": 0.00347599,
      "Context.StorageManager.array_open_READ.avg": 0.00347599,
      "Context.StorageManager.VFS.ArrayDirectory.list_root_uris.sum": 0.000945713,
[...]

PR #896, opened 2023-02-07
Export `open` function, to universally open everything.

### LINKED ISSUE #539: API: missing ability to introspect given only a URI

thetorpedodog:
Adds some tests to `factory.open` to make sure it is ready for universal consumption and, having done so, makes it available for universal consumption.

---

Fixes #539.

### DIFF
--- a/apis/python/src/tiledbsoma/__init__.py
+++ b/apis/python/src/tiledbsoma/__init__.py
@@ -5,6 +5,7 @@ from .dataframe import DataFrame
 from .dense_nd_array import DenseNDArray
 from .exception import DoesNotExistError, SOMAError
 from .experiment import Experiment
+from .factory import open
 from .general_utilities import (
     get_implementation,
     get_implementation_version,
@@ -19,24 +20,25 @@ from .sparse_nd_array import SparseNDArray
 __version__ = get_implementation_version()
 
 __all__ = [
-    "get_implementation",
-    "get_implementation_version",
-    "get_SOMA_version",
-    "get_storage_engine",
-    "show_package_versions",
-    "stats_enable",
-    "stats_disable",
-    "stats_reset",
-    "stats_dump",
     "AxisColumnNames",
     "AxisQuery",
-    "ExperimentAxisQuery",
     "Collection",
+    "DataFrame",
     "DenseNDArray",
     "DoesNotExistError",
     "Experiment",
-    "SOMAError",
-    "DataFrame",
+    "ExperimentAxisQuery",
+    "get_implementation_version",
+    "get_implementation",
+    "get_SOMA_version",
+    "get_storage_engine",
     "Measurement",
+    "open",
+    "show_package_versions",
+    "SOMAError",
     "SparseNDArray",
+    "stats_disable",
+    "stats_dump",
+    "stats_enable",
+    "stats_reset",
 ]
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/__init__.py
```diff
@@ -19,24 +20,25 @@
...
+    "get_implementation_version",
+    "get_implementation",
+    "get_SOMA_version",
+    "get_storage_engine",
     "Measurement",
+    "open",
+    "show_package_versions",
+    "SOMAError",
     "SparseNDArray",
+    "stats_disable",
```
bkmartinjr, line 40: I missed the PR that added these -- IMHO they should not be at the top-level namespace, as they are TileDB-specific.

thetorpedodog: This is not part of the current change, but I disagree on this assessment—I think it makes sense for `tiledbsoma` to directly export additional functionality *in addition to* the basic SOMA functionality. If somebody is importing `tiledbsoma`, they already are binding their implementation to `tiledbsoma` specifically.

bkmartinjr: I do not feel strongly about this, so I leave it to a consensus decision....

thetorpedodog:
> core changes look good. Suggest that the tests include validation of the context manager characteristics, eg., add `with soma.open(...): ...`

In this case we know that the types returned support the context manager protocol (`isinstance` checks), and the context manager behavior is covered elsewhere.

apis/python/src/tiledbsoma/__init__.py
johnkerl: Agreed with @thetorpedodog that this is unrelated to this PR. The work was done on #865 and #871. Putting stats routines at top level was worked out in the PRs, but indeed could (and should) have been discussed more broadly than on the PRs.

I've added this to the agenda for our Friday spec-discussion meeting, and have also upated issue #872.

johnkerl (commented):
Wow, quick review @bkmartinjr ! I didn't get to chime in before the accept and merge.

This is awesome, delighted to see it! 🎉

Issue #858, opened 2023-02-02
[python] Improve on-line help for `add_new_*`


johnkerl:
Follow-on from https://github.com/single-cell-data/TileDB-SOMA/pull/848

https://github.com/single-cell-data/TileDB-SOMA/pull/848#pullrequestreview-1281883231

I find the existing wording insufficient for me to understand how to use the relative feature.

johnkerl:
Since I filed this task I'll leave it open until I go through #898 #899 #909

PR #899, opened 2023-02-07
Move `io` into submodule and organize various `util_` files.

thetorpedodog:
This change does some reorganization to make dependencies and names a little clearer:

- Moves `io.py` into `io/ingest.py`. By re-exporting the already-used names at `io/__init__.py`, calling code does not need to be changed.
- Takes the `util_ann.py`, `util_scipy.py`, and functions from `util_tiledb.py` which are only used by `io` and moves them into `io/conversions.py`.
- Renames the `util_arrow` and `util_iter` modules to `arrow_types` and `read_iters`, respectively, to better reflect their usage.
- Moves the exception-related methods of `util_tiledb` (now the rest of it) into `exception`.

No behavior changes; just moving things around. This both does immediate reorg and sets up for having a clearer place for decorators to help improve https://github.com/single-cell-data/TileDB-SOMA/issues/858 (and some other related documentation stuff).


bkmartinjr (approved):
Much cleaner - thanks for doing this.

johnkerl (approved):
🚀

Issue #900, opened 2023-02-07
[Bug] Reading slices beyond shape of an `NDArray` errors incorrectly


bkmartinjr:
**Describe the bug**

Requesting a slice to be read, where the slice extends beyond the end of the dimension, will error.  This is not correct - it should simply read any portion of the array defined by the slice.

There seems to be a logic error in the interface which converts the slice into a point list for SOMAReader.

**To Reproduce**

Test case:
```python
import pyarrow as pa
import numpy as np
import scipy.sparse
import tiledbsoma as soma

with soma.Collection.create("./testcollection") as collection:
    snda = collection.add_new_sparse_ndarray(
        "snda", type=pa.float32(), shape=(100, 100)
    )
    snda.write(
        pa.SparseCOOTensor.from_scipy(
           scipy.sparse.random(100, 100, dtype=np.float32)
        )
    )

with soma.SparseNDArray.open("testcollection/snda", mode="r") as snda:
    some_data = snda.read((slice(0, 101),)).tables().concat()
```

Result:
```
RuntimeError                              Traceback (most recent call last)
Cell In[24], line 17
     10     snda.write(
     11         pa.SparseCOOTensor.from_scipy(
     12            scipy.sparse.random(100, 100, dtype=np.float32)
     13         )
     14     )
     16 with soma.SparseNDArray.open("testcollection/snda", mode="r") as snda:
---> 17     some_data = snda.read((slice(0, 101),)).tables().concat()

File ~/TileDB-SOMA/apis/python/src/tiledbsoma/sparse_nd_array.py:125, in SparseNDArray.read(***failed resolving arguments***)
    121         if lo < 0 or hi < 0:
    122             raise ValueError(
    123                 f"slice start and stop may not be negative; got ({lo}, {hi})"
    124             )
--> 125         sr.set_dim_ranges(dim_name, [lo_hi])
    126     # Else, no constraint in this slot. This is `slice(None)` which is like
    127     # Python indexing syntax `[:]`.
    128 elif isinstance(
    129     coord, (collections.abc.Sequence, pa.Array, pa.ChunkedArray)
    130 ):

RuntimeError: [TileDB::Subarray] Error: Cannot add range to dimension 'soma_dim_0'; Range [0, 101] is out of domain bounds [0, 99]
```

**Versions (please complete the following information):**

tiledbsoma.__version__    0.5.0a8.post62.dev2571341179
tiledb.__version__        0.20.0
core version              2.14.0
[...]

Issue #519, opened 2022-11-09
Consider negative slicing


johnkerl:
I'm wondering if util.slice_to_range should not enhanced to do the negative checks, etc, as they are common to all TileDB indexing, and much of it is duplicated in SparseNdArray.

_Originally posted by @bkmartinjr in https://github.com/single-cell-data/TileDB-SOMA/pull/483#discussion_r1017316024_

To do:

* Evaluate feasibility of negative indexing within `libtiledbsoma` with @gspowley
* Assuming everything makes sense, implement this in C++
* Connect that up to Python

johnkerl:
Note: discussion on https://github.com/single-cell-data/SOMA/issues/51, if accepted, would result in us disallowing negative indexing

johnkerl:
Reasons to not support it:

* What's natural for Python users is `a[-2]` is an alias for `a[len(a)-2]`; but what's natural for R users is `a[-2]` means the whole array _except for_ element 2. Thus there's no one semantic meaning to choose.
* In sparse-array contexts, it's not even clear what "two from the end" would even mean.
* We reserve the right to re-think this some time down the road, should new perspectives arise in the fullness of time.

PR #901, opened 2023-02-07
[python] Fix and test over-indexing semantics

johnkerl:
## Issue and/or context:

#900

### DIFF
--- a/apis/python/src/tiledbsoma/util.py
+++ b/apis/python/src/tiledbsoma/util.py
@@ -128,6 +128,10 @@ def slice_to_range(
     if stop is None:
         stop = nonempty_domain[1]
 
+    # Indexing beyond end of array should "trim" to end of array
+    if stop > nonempty_domain[1]:
+        stop = nonempty_domain[1]
+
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/util.py
```diff
@@ -128,6 +128,10 @@ def slice_to_range(
     if stop is None:
         stop = nonempty_domain[1]
 
+    # Indexing beyond end of array should "trim" to end of array
```
bkmartinjr, line 131: Possibly naive question:  do we need the same for `start` coordinate?   I.e., bound it to zero?

bkmartinjr (approved):
One question about start range bounding.  Fix for end of range looks great.

apis/python/src/tiledbsoma/util.py
johnkerl: sure

johnkerl:
@bkmartinjr second thoughts -- we already decided we would not accept negative indices: #519

PR #905, opened 2023-02-08
Update to mypy 1.0.0 and FINALLY use `Self`!

thetorpedodog:
I discovered this when I was trying to figure out why there were inconsistencies between the errors reported by mypy running in pre-commit and the mypy I had installed in my development venv, which were driving me thoroughly insane while I was trying to do completely innocuous things. This seems to have fixed it, and finally, I can eliminate all those `TypeVar("_Self")`s!


bkmartinjr (approved):
Yea!

johnkerl (commented):
🚀

PR #907, opened 2023-02-08
[r] Update man/ directory by running roxygenize (closes #906)

### LINKED ISSUE #906: [r] Synchronize `man`/ `roxygenise` updates
In the R package, `man/*.Rd` files are currently out of date with respect to `R/*.R` source.

There are two opportunities:

* Get these up to date manually for the next release and/or head
* Automate this:
  * Maybe some release-time CI perhaps -- ?
  * Maybe some pre-commit hook that runs `roxygenise` on every PR that alters `roxygen` inputs -- ?

eddelbuettel:
#### Issue and/or context:

`man/` was out of sync

#### Changes:

`roxygenize()` has been called, files have been updated

#### Notes for Reviewer:

#906 and https://app.shortcut.com/tiledb-inc/story/25203/run-roxygenize

### DIFF
--- a/apis/r/cleanup
+++ b/apis/r/cleanup
@@ -9,5 +9,6 @@ rm -r -f \
    src/include/ \
    src/lib \
    src/cmake_log.txt \
-   src/build 
+   src/build \
+   src/libtiledbsoma
 
--- a/apis/r/man/SOMAArrayBase.Rd
+++ b/apis/r/man/SOMAArrayBase.Rd
@@ -4,7 +4,7 @@
 \alias{SOMAArrayBase}
 \title{SOMA Array Base Class}
 \description{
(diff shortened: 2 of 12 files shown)


johnkerl (approved):
LGTM but IIUC this does the _first half of_ #906 which is the run-once-manually-to-catch-up -- do we want some way to auto-stay current? (If we don't then I wrote the second half of #906 based on a misunderstanding ...)

eddelbuettel:
And I said, when someone edits roxygen text they are expected to run roxygenize because R CMD check tests for (say) 'are these new arguments documented'.  What we could do (but what nobody does) is to _fail_ in CI on a _WARNING_ (which ends up creating too many positives).  So to me your suggestion misses the core issue.

johnkerl:
OK got it I did indeed miss the core issue @eddelbuettel . Thanks for teaching me! :)

PR #908, opened 2023-02-08
[python] Add `replace` method to `SOMATileDBContext`

### LINKED ISSUE #904: [FEATURE REQUEST] `SOMATileDBContext` needs merge/evolve capability

It is not easy to update/evolve a SOMATileDBContext (or merge two of them).  This is useful in some cases, for example:
* you have a SOMATileDBContext with a TileDB Ctx
* Create a new TileDB Ctx with some updated KV pairs
* return an evolved SOMATileDBContext with the TileDB Ctx you just built

From discussion with @thetorpedodog and @johnkerl:  Add a class method which takes a dict of KV pairs to update, other new values to replace on the context, and use that to build a new context object.

**Additional context**

Needed by cell census package.

bkmartinjr:
## Issue and/or context:

Fixes #904 

## Changes:

Added `replace` method and unit tests, allowing contexts to be updated/evolved.

### DIFF
--- a/apis/python/src/tiledbsoma/options/soma_tiledb_context.py
+++ b/apis/python/src/tiledbsoma/options/soma_tiledb_context.py
@@ -1,9 +1,9 @@
 import time
 from typing import Any, Dict, Optional, Union
-from typing_extensions import Self
 
 import attrs
 import tiledb
+from typing_extensions import Self
 
 
 def _build_default_tiledb_ctx() -> tiledb.Ctx:
@@ -99,7 +99,7 @@ class SOMATileDBContext:
         assert isinstance(self._group_write_tiledb_ctx, tiledb.Ctx)
 
     @classmethod
-    def evolve(cls, context: Self, **changes) -> Self:
+    def evolve(cls, context: Self, **changes: dict[str, Any]) -> Self:
         """
         Create a copy of the context, merging changes.
 
--- a/apis/python/src/tiledbsoma/options/soma_tiledb_context.py
+++ b/apis/python/src/tiledbsoma/options/soma_tiledb_context.py
@@ -1,5 +1,6 @@
 import time
 from typing import Any, Dict, Optional, Union
+from typing_extensions import Self
 
 import attrs
 import tiledb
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/options/soma_tiledb_context.py
```diff
@@ -96,3 +97,16 @@ def __attrs_post_init__(self) -> None:
         else:
             object.__setattr__(self, "_group_write_tiledb_ctx", self.tiledb_ctx)
         assert isinstance(self._group_write_tiledb_ctx, tiledb.Ctx)
+
+    @classmethod
+    def evolve(cls, context: Self, **changes: Dict[str, Any]) -> Self:
```
johnkerl, line 102: personally I prefer `merge` but this is fine

johnkerl (approved):
✅

apis/python/src/tiledbsoma/options/soma_tiledb_context.py
thetorpedodog: I am fine with any name choice. `merge` is also good, and `replace` is another frequently-used one (cf. `inspect.Parameter`).

Two levels of suggestion here:

FIRST: A minor change that doesn't change your intent but saves lots of typing:

```python
@attrs.define
class SomeAttrsClass:
  ...

  evolve = attrs.evolve

  # This ends up working equivalently to:
  def evolve(self, **kwargs):
    return attrs.evolve(self, **kwargs)
  # ... because `self` gets bound to the first param of `attrs.evolve`
  # when accessed as `some_instance.evolve(...)`  
```

Given that this solves an immediate need I would be fine with it going in like this (or even the slightly more verbose `def evolve(...)` version).

POTENTIALLY DEEPER SUGGESTION:

That said I think in this case something like this might be a bit more useful (and completely fine if you save this for a follow-up):

```python
@attrs.define(...)
class SOMATileDBContext:

  def evolve(self, *, tiledb_config: Optional[Dict[str, Any]] = None, **changes: Any) -> Self:
    if tiledb_config:
      # build a new tiledb.Ctx here
      changes["tiledb_ctx"] = new_ctx
    attrs.evolve(self, **changes)
```

That would let you say something like `some_inst.evolve(tiledb_config={"put": "your", "overrides": "here"})`

ALSO A MINOR NOTE: When annotating things like `*args` and `**kwargs`, the annotation should reflect the type of *one* of those things, so:

```python
def any_number_of_strings(*whatever: str) -> result:
  ...

any_number_of_strings("this", "that", "the other")

def some_ints(**kwargs: int) -> result:
  ...

some_ints(a=1, b=2, c=3)
```

johnkerl: oohhh `replace` even better :D

bkmartinjr: All super good suggestions.  Will implement all, evolving to replace :-)

bkmartinjr:
@thetorpedodog - please take another look.  I believe I addressed all of your ideas/concerns (good ideas BTW)

thetorpedodog (approved):

PR #909, opened 2023-02-08
[python] Use `**kwargs` and a function decorator to forward arguments

### LINKED ISSUE #858: [python] Improve on-line help for `add_new_*`

thetorpedodog:
In #848 the main reason I used `functools.partialmethod` was to avoid having to repeat the same parameter list multiple times (and specify the same defaults multiple times, etc. etc.). However, `partialmethod` doesn't pick up the function's `__doc__` and related attributes, which meant that doing `help(coll.add_new_sparse_ndarray)` just gave you the documentation for `partialmethod`, which is completely useless.

This is one potential solution, which IMO has a lot of advantages:

- less repetition, and useful in more places than anticipated.
- no way for arg lists/defaults to go out of sync.
- pretty much the intended use case for `**kwargs`.

However, it does have some disadvantages:

- less static analysis support.
- somewhat complex internals (though they are well-tested and hidden behind a simple abstraction).

I'm not deeply attached to this idea (though I think it is pretty neat) and would be willing to go with the simpler solution (just repeat everything three+ times).

Fixes #858.

See the narrative below for way more detail:

---

For methods which primarily consist of forwarding arguments from themselves to another backing function, this change replaces the repeated argument list with the use of `**kwargs` to reduce repetition and the possibility of drifting defaults. Additionally, it creates a new `@forwards_kwargs_to` decorator to update the `__signature__` of the wrapper function to ensure that online help still provides useful information.

Background:

One of the primary reasons to avoid using `**kwargs` to pass arguments directly through functions is that doing so means users no longer receive useful help information when using a function in an interactive environment like ipython. Instead of seeing the parameters the function expects (and type annotations and defaults), the help text (and parameter hints, etc.) will only show the `**kwargs`.

Conversely, using `**kwargs` is often useful and beneficial for a library implementor (like us). Directly using `**kwargs` means that arguments can be passed through untouched, and there is no need to repeat argument lists and default parameters, all of which can get out of sync from caller to callee.

These two goals are somewhat at odds. However, the standard library provides a semi-documented workaround: a callable can provide a `__signature__` member which is used *instead of* the actual function signature of the called object in online help (e.g. `help(some_fn)`) and is also used to provide parameter information in interactive shells.

How it works:

The new `@forwards_kwargs_to` decorator inspects the argument list of both the wrapped function and the destination function and generates a Signature which accurately reflects the combined parameters, default values, and ordering of a function that uses `**kwargs` to directly forward arguments to a backing function. It sets the `__signature__` member on the wrapped function. This value is then used at runtime.

Example (from an actual ipython session with this code)

    $ ipython
    In [1]: import tiledbsoma

    In [2]: coll = tiledbsoma.open("some-collection")

    In [3]: coll.add_new_dense_ndarray?
    Signature:
    coll.add_new_dense_ndarray(
        key: 'str',
        *,
        uri: 'Optional[str]' = None,
        type: pyarrow.lib.DataType,
        shape: Sequence[int],
        platform_config: Optional[Mapping[str, Any]] = None,
    ) -> 'DenseNDArray'
    Docstring:
    Adds a new DenseNDArray to this Collection.

    For details about the behavior of ``key`` and ``uri``, see
    :meth:`add_new_collection`. The remaining parameters are passed to
    the :meth:`DenseNDArray.create` method unchanged.

    In [4]: coll.add_new_dense_ndarray('something', <TAB>
    platform_config= ...
    shape=           ...
    type=            ...
    uri=             ...

Cons:

The `__signature__` value is generated at runtime and cannot be used at static analysis time. This means that tools like `mypy` and other static analysis–based development tools (e.g. Pylance's intellisense) aren't able to provide detailed type checking.

### DIFF
--- a/apis/python/src/tiledbsoma/collection.py
+++ b/apis/python/src/tiledbsoma/collection.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import functools
 import re
 import time
 from typing import (
@@ -11,7 +10,6 @@ from typing import (
     Iterator,
     List,
     Optional,
-    Sequence,
     Tuple,
     Type,
     TypeVar,
@@ -20,16 +18,14 @@ from typing import (
 )
 
 import attrs
-import pyarrow as pa
 import somacore
 import somacore.collection
 import tiledb
 from somacore import options
 from typing_extensions import Self
 
-from . import tdb_handles
+from . import funcs, tdb_handles
 from .common_nd_array import NDArray
-from .constants import SOMA_JOINID
 from .dataframe import DataFrame
 from .dense_nd_array import DenseNDArray
 from .exception import is_does_not_exist_error, is_duplicate_group_key_error
@@ -187,14 +183,9 @@ class CollectionBase(
             uri,
         )
 
+    @funcs.forwards_kwargs_to(DataFrame.create, exclude=("context",))
     def add_new_dataframe(
-        self,
-        key: str,
-        *,
-        uri: Optional[str] = None,
-        schema: pa.Schema,
-        index_column_names: Sequence[str] = (SOMA_JOINID,),
-        platform_config: Optional[options.PlatformConfig] = None,
+        self, key: str, *, uri: Optional[str] = None, **kwargs: Any
     ) -> DataFrame:
         """Adds a new DataFrame to this collection.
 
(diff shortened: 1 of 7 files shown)


johnkerl (approved):
Thanks @thetorpedodog for the write-up and comparision of approaches in the description section -- I _very_ much appreciate your outreach to the less-python-expert members of the team (i.e. all of us) -- this is empowering, inclusive, and supportive.

Re the tradeoffs -- this approach is indeed "python magic" but it's encapsulated within the single `@funcs.forwards_kwargs_to` which IMO not only _contains_ the magic well but also _explains_ the magic well.

bkmartinjr (approved):
am good with this trade!

surprising that functools (or the like) doesn't have support for this given that the module must do something similar internally...

thetorpedodog:
Since I’ve got a couple votes in favor, I’m going to merge this. If anybody else here has comments or concerns, I am happy to continue discussing here and can resolve further issues in a followup.

PR #538, opened 2022-11-16
[python] Remove DataFrame in favor of IndexedDataFrame

johnkerl:
Experiments have shown (dense) `DataFrame` offers no compelling performance reasons for its continued existence.

Issue: #380

PR #910, opened 2023-02-09
Allow passing 0-length lists of coordinates to DataFrame.

thetorpedodog:
This is an oddity I noticed when working on documentation for the `DataFrame.read` method in somacore. It's not a bug per se but is a special case that I found unexpected. I dug through back to #538 where this was introduced and didn't find the reasoning behind this particular behavior. It's fine to keep as-is but I figured I would see if we do want to keep this behavior.

---

Currently, DataFrame allows passing a sequence of coordinates shorter than the number of indexed dimensions:

    # df has 4 dimensions

    df.read(coords=[1, 2, 3, 4])
    # -> [1, 2, 3, 4]
    df.read(coords=[1, 2])
    # -> [1, 2, unrestricted, unrestricted]
    df.read(coords=[1])
    # -> [1, unrestricted, unrestricted, unrestricted]

However, this pattern does not continue down to 0 dimensions:

    df.read(coords=[])
    # -> Not allowed

despite the fact that not passing coordinates at all is allowed:

    df.read(coords=None)
    # -> [unrestricted, unrestricted, unrestricted, unrestricted]

This change treats the zero-length array equivalent to an array of any other length:

    df.read(coords=[])
    # -> [unrestricted, unrestricted, unrestricted, unrestricted]

### DIFF
--- a/apis/python/src/tiledbsoma/dataframe.py
+++ b/apis/python/src/tiledbsoma/dataframe.py
@@ -140,7 +140,7 @@ class DataFrame(TileDBArray, somacore.DataFrame):
 
         * None
         * A sequence of coordinates is accepted, one per dimension.
-        * Sequence length must be at least one and <= number of dimensions.
+        * Sequence length must be <= number of dimensions.
         * If the sequence contains missing coordinates (length less than number of dimensions),
           then `slice(None)` -- i.e. no constraint -- is assumed for the missing dimensions.
         * Per-dimension, explicitly specified coordinates can be one of: None, a value, a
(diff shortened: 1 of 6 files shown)


bkmartinjr (approved):
thumbs up.  this makes is consistent with numpy indexing as well, e.g., if you have a 2D array called `a`, then `a[()]` returns the entire beast.

johnkerl (approved):
This jives with #484 which is tangentially related.

P.S. Good catch @thetorpedodog !

PR #912, opened 2023-02-10
Put more info in the `__repr__` of SOMA objects.

thetorpedodog:
This adds more information to the general `__repr__` of SOMA objects, including their open mode and status, and slightly reformats the representation of collections.

It also ensures that `repr` will work on objects and collections that have been closed, and that `repr` will only recurse into sub-elements of a collection that have already been opened (so it doesn't eagerly open up an entire collection tree).

---

Yet another entry from the List Of Things I Notice While Doing Other Things.

### DIFF
--- a/apis/python/src/tiledbsoma/collection.py
+++ b/apis/python/src/tiledbsoma/collection.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import itertools
 import re
 import time
 from typing import (
@@ -7,8 +8,8 @@ from typing import (
     Callable,
     ClassVar,
     Dict,
+    Iterable,
     Iterator,
-    List,
     Optional,
     Tuple,
     Type,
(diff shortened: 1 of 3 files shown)


bkmartinjr (approved):
I am a bit torn on this change. On one hand, I like that it resolves the significant performance issues with the current repr.  On the other hand, because it will show cached state when it is available, but will not preemptively load un-opened objects, the Collection repr behaves differently at different times. This may be confusing to users.

Example (look at commands `13` and `15` which are "the same" but show different (cached) results:
```
In [11]: import tiledbsoma as soma

In [12]: census = soma.open("s3://cellxgene-data-public/cell-census/2023-02-06/soma")

In [13]: census
Out[13]: 
<Collection 's3://cellxgene-data-public/cell-census/2023-02-06/soma' (open for 'r') (2 items)
    'census_info': 's3://cellxgene-data-public/cell-census/2023-02-06/soma/census_info' (unopened)
    'census_data': 's3://cellxgene-data-public/cell-census/2023-02-06/soma/census_data' (unopened)>

In [14]: census['census_data']['homo_sapiens'].obs
Out[14]: <DataFrame 's3://cellxgene-data-public/cell-census/2023-02-06/soma/census_data/homo_sapiens/obs' (open for 'r')>

In [15]: census
Out[15]: 
<Collection 's3://cellxgene-data-public/cell-census/2023-02-06/soma' (open for 'r') (2 items)
    'census_info': 's3://cellxgene-data-public/cell-census/2023-02-06/soma/census_info' (unopened)
    'census_data': Collection 's3://cellxgene-data-public/cell-census/2023-02-06/soma/census_data' (open for 'r') (2 items)
        'homo_sapiens': Experiment 's3://cellxgene-data-public/cell-census/2023-02-06/soma/census_data/homo_sapiens' (open for 'r') (2 items)
            'ms': 's3://cellxgene-data-public/cell-census/2023-02-06/soma/census_data/homo_sapiens/ms' (unopened)
            'obs': DataFrame 's3://cellxgene-data-public/cell-census/2023-02-06/soma/census_data/homo_sapiens/obs' (open for 'r')
        'mus_musculus': 's3://cellxgene-data-public/cell-census/2023-02-06/soma/census_data/mus_musculus' (unopened)>
```

I don't have a better idea - and _really_ like the speed and overhead improvements.  So am approving the PR.   Just wanted to point out a sharp edge in the UX of this change.

thetorpedodog:
> I don't have a better idea - and _really_ like the speed and overhead improvements. So am approving the PR. Just wanted to point out a sharp edge in the UX of this change.

There’s definitely no perfect solution here. The inconsistency is a bit strange, but the alternative is, in the worst case, “I tried to print this thing and it started reading from disk / made a bajillion network requests and took 10 seconds”.

This makes me think that in the future an “always open everything eagerly” knob (possibly in the context) or just an “open everything in this collection” method might be useful but that’s neither here nor there.

bkmartinjr:
> This makes me think that in the future an “always open everything eagerly” knob (possibly in the context) or just an “open everything in this collection” method might be useful but that’s neither here nor there.

Here are options I see:
1. print _only_ the children of the collection.  Always fast, always consistent.
2. Current implementation - opportunistically print cached grandchildren, etc.   Always fast, sometimes 'huh'?
3. Do fast & dirty in the __repr__ (i.e. option 1), but have a verbose method (`mohr_repr()`) that prints entire tree and does greedy open if data not cached
4. Control behavior of _repr_ #3 with a flag, toggling between modes

nothing brilliant - really different strokes for different use cases.

johnkerl (approved):
+1 re the discussion, this is non-ideal, an no other option is ideal, and it's a balancing act.

Issue #914, opened 2023-02-10
[r] Re-punctuate `[lifecycle: ...]` tags


johnkerl:
For PR #829 (issue #708) I added `[lifecycle: experimental]` to TileDB-SOMA-R code.

I did this in exuberant naïveté; in turns out the `[...]` are more semantic in R than I realized.

IIUC we need to say `(lifecycle: experimental)` in R.

cc @eddelbuettel @bkmartinjr @thetorpedodog @atolopko-czi

eddelbuettel:
> are more semantic in R than I realized

If and only if one turns on the option for markdown processing via the already optional roxygen2 (a popular add-on documentation converter / generator).  In plain base R and _hand-edited_ `.Rd` files you can write squares as long and as much as you please.   But we do this and so it affected your change as pointed out over slack.

PR #916, opened 2023-02-10
[r] Re-punctuate lifecycle tags

johnkerl:
## Issue and/or context:

#914


eddelbuettel (approved):
:+1:  -- and I looked briefly, there is apparently no good way to escape square backets in standard (non-GitHub flavoured) markdown.

PR #917, opened 2023-02-10
[python] Check to ensure we're not doing partitioned reads

thetorpedodog:
Since we don't yet support partitioning, we shouldn't return results which aren't partitioned to callers who expect it.


thetorpedodog:
Updated to remove the `dense_nd_array.py` change since we'll remove that parameter entirely pretty soon and I don't want to create merge conflicts for myself.

PR #928, opened 2023-02-13
[python] Add names to `NDArray` tests, too

thetorpedodog:


apis/python/tests/test_sparse_nd_array.py
```diff
@@ -722,16 +737,7 @@ def test_csr_csc_2d_read(tmp_path, shape):
             "throws": None,
         },
         {
-            "shape": (4, 5, 6),
```
johnkerl, line 725: I imagine you have a good reason for this ...

This red-lining matches nothing in the PR title, description, etc.

Please add a note what you're doing here & why. (I can `git checkout` the diff and spend some time figuring out why ... but you should add a note to help the reviewer understand in the first place.)

johnkerl (changes requested):

apis/python/tests/test_sparse_nd_array.py
thetorpedodog: This is a duplicate test.

johnkerl: @thetorpedodog thanks!

PR #945, opened 2023-02-15
Make the pull-request template a little less shouty

johnkerl:
No tracking issue -- just aesthetics.

Having used this template for a while I find the `##` a bit much, and unnecessary. Also, it's error-prone as if one does at first

```
## Issue: foo
```

one needs to immediately re-edit to

```
## Issue:

foo
```

which is error-prone and (micro-)time-consuming.

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,6 +1,6 @@
-## Issue and/or context:
+**Issue and/or context:**
 
-## Changes:
+**Changes:**
 
-## Notes for Reviewer:
+**Notes for Reviewer:**
 


pull_request_template.md
```diff
@@ -1,6 +1,6 @@
-## Issue and/or context:
+**Issue and/or context:**
```
eddelbuettel, line 1: I use `####` when I edit them.  Appears to be equivalent.

johnkerl: @eddelbuettel that would solve only one of the two problems I noted in the description, namely:

```
#### Issue: foo
```

is still a mistake one needs to immediately re-edit to

```
#### Issue:

foo
```

I want to make this less error-prone and less (micro-)time-consuming.

atolopko-czi (approved):
LGTM

PR #946, opened 2023-02-15
[python] Re-export `somacore.options.SparseDFCoord`

johnkerl:
**Issue and/or context:** #872

### DIFF
--- a/apis/python/src/tiledbsoma/__init__.py
+++ b/apis/python/src/tiledbsoma/__init__.py
@@ -1,4 +1,5 @@
 from somacore import AxisColumnNames, AxisQuery, ExperimentAxisQuery
+from somacore.options import SparseDFCoord
 
 from .collection import Collection
 from .dataframe import DataFrame
@@ -28,6 +29,7 @@ __all__ = [
     "DoesNotExistError",
     "Experiment",
     "ExperimentAxisQuery",
+    "SparseDFCoord",
     "get_implementation_version",
     "get_implementation",
     "get_SOMA_version",


bkmartinjr (approved):
Using the criteria set by @thetorpedodog, I am not sure we need to re-export these helper types, as users don't create them directly in common usage  Rather they create the types inside the union.  In this case, it would be a tuple of slices, etc. In that regard, they are much like the types declared for return values, which we also don't re-export.

The only case I can think of is where a "user" is really proxying the signature, and needs to use the type to ensure future compliance with somacore. In which case, using somacore.options.SparseDFCoord seems fine.

This is definitely an edge case, and I'm down to the edges of that edge case.  So consider this a suggestion, and I'll approve the PR.

Would like to see Paul's opinion on this as a last check.

thetorpedodog:
From https://github.com/single-cell-data/TileDB-SOMA/issues/872#issuecomment-1431543978: this type should not be re-exported.

johnkerl:
@bkmartinjr I'm fine canceling this PR.

It was in direct response to your ask on #872 

Screenshot 2023-02-15 at 10 09 59 AM

johnkerl:
@bkmartinjr I'm fine canceling this PR.

It was in direct response to your ask on #872 

Screenshot 2023-02-15 at 10 57 57 AM

bkmartinjr:
> It was in direct response to your ask on https://github.com/single-cell-data/TileDB-SOMA/issues/872

right, and we have had a much more substantive discussion since then.  I'm very glad we arrived at a nice design principle to apply (thanks Paul!)

johnkerl:
OK @bkmartinjr I'm closing this PR given that updated feedback on your interpretation of the export rule & whether you expect users to call this directly. Thanks!

PR #948, opened 2023-02-15
Allow slicing with any type and with negative values.

### LINKED ISSUE #933: [Bug] Support negative and string indices for `DataFrame`
Accessing a negative index of a dataframe currently causes an error. Per https://github.com/single-cell-data/SOMA/issues/121#issuecomment-1426485914 it should not (only when the domain is non-negative).

https://github.com/single-cell-data/TileDB-SOMA/blob/d1feac930f3f9218a46d605e9c0ddc583c5f1cbd/apis/python/src/tiledbsoma/dataframe.py#L201-L204

thetorpedodog:
- Removes the "slice indices must be positive" requirement,  per discussion on https://github.com/single-cell-data/SOMA/pull/119.
- Enables slicing for string dimensions.

This change also follows up on the previous change to allow"over-indexing" off the low end of a domain to work just as it does when indexing off the high end of a domain: the low end is clipped to the bottom of the domain. So for a dimension with domain [−3, 8]:

    slice(-100, 100) -> -3, 8
    slice(-5, 0) -> -3, 0
    slice(5, 10) -> 5, 8
    slice(-100, -50) -> error (too low)
    slice(50, 100) -> error (too high)

Since this uses the New And Improved `is_slice_of` function, this also pulls in the new somacore (where that is the only change present).

Fixes #933.

### DIFF
--- a/apis/python/tests/test_dataframe.py
+++ b/apis/python/tests/test_dataframe.py
@@ -378,6 +378,7 @@ def make_multiply_indexed_dataframe(tmp_path, index_column_names: List[str]):
             ("strings_aaa", pa.string()),
             ("zero_one", pa.int64()),
             ("thousands", pa.int64()),
+            ("both_signs", pa.int64()),
             ("soma_joinid", pa.int64()),
             ("A", pa.int64()),
         ]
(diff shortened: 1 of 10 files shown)


apis/python/tests/test_dataframe.py
```diff
@@ -536,6 +536,27 @@ def make_multiply_indexed_dataframe(tmp_path, index_column_names: List[str]):
             "A": [12, 13, 14, 15],
             "throws": None,
         },
+        {
```
bkmartinjr, line 539: one thing this does not test is a case where:
* an indexed column allows negative int values
* and the user slices negatively

Eg., a DataFrame with an indexed column of type int64, actual column values of [-10..10], and a read that slices out `[-2:2]`, etc.

Unless I am misreading the code in `make_multiply_indexed_dataframe`, all numeric coordinates are currently positive, which means there may still be dragons in coordinate values <= 0

bkmartinjr (approved):
Other than the testing suggestion, this LGTM

apis/python/tests/test_dataframe.py
thetorpedodog: Added.

Issue #931, opened 2023-02-14
[python] Fix non-appendability of `SparseNDArray`


johnkerl:
## Use-case analysis

* For cell census, arrays are immutable and people want to take their shapes
* For `soma.io`, arrays are mutable and people want to take their shapes
* Currently only the first use-case is supported

## Why this affects only `SOMASparseNDArray` and not `SOMADataFrame`

* `DataFrame` is a sparse 1D, n-attr TileDB array
  * Currently this is written with full-int64 domain at the schema level
  * You could get its non-empty domain at the data level to get the 'shape' i.e. to figure out the number of rows ...
  * ... but what we actually use to get the 'shape' aka `df.count` is
    * https://github.com/single-cell-data/TileDB-SOMA/blob/0.5.0a11/apis/python/src/tiledbsoma/dataframe.py#L94-L100
    * https://github.com/single-cell-data/TileDB-SOMA/blob/0.5.0a11/libtiledbsoma/src/soma_reader.cc#L176
  * This does:
    * If there is one fragment it can quickly return cell count from `FragmentInfo`
    * Else if there is more than one fragment but all with non-overlapping non-empty domains it can quickly return sum of cell count over each FragmentInfo
    * Else the computation is a bit slower
    * See also: https://github.com/single-cell-data/TileDB-SOMA/issues/383
    * See also: https://github.com/single-cell-data/TileDB-SOMA/pull/404
* `SparseNDarray` is a sparse 2D, 1-attr TileDB array
  * Currently this is written with "tight" domain at the schema level, e.g. if it's 100x200 sparse the domain will be `((0,99),(0,199))`
  * You can look at non-empty domain at the data level but that might be say `((2,99),(0,198))` or somesuch if an entire row or column is entirely unoccupied
  * You can get its nnz using the same function above
    * However this (unlike in the fully occuped 1D case) does not_ give back the 100x200 info
    * Suppose the array is 7.4% occupied -- this will give you the single number 1,480 which is the number of IJV triples in the sparse array but nothing about the original 100 or 200
  * The issues with the previous two bullets are why we're writing "tight" domain -- so we can read back the 100 x 200 info from the domain and the domain is the place we "store" this information

TL;DR:

* For `DataFrame` which is 1D sparse but fully occupied:
  * domain, non-empty domain, or nnz all could be used to store the 'shape'
  * we use nnz, and we leave domain all-int64, and we don't use non-empty domain programatically
[...]

PR #952, opened 2023-02-16
[python] Mutable-X mods, and `.shape` as explicit `SparseNDArray` capacity

johnkerl:
**Issue and/or context:**

Tracking issue: #931

This is the Python-impl counterpart to https://github.com/single-cell-data/SOMA/pull/131; an R-impl counterpart is upcoming.

This PR addresses only explicit shaping of `SparseNDArray` -- explicit shaping of `DataFrame` will be on a separate PR.

### DIFF
--- a/apis/python/src/tiledbsoma/_common_nd_array.py
+++ b/apis/python/src/tiledbsoma/_common_nd_array.py
@@ -39,7 +39,7 @@ class NDArray(TileDBArray, somacore.NDArray):
             If the type is unsupported, an error will be raised.
 
         :param shape: The maximum capacity of the dataframe, including room for any
-            intended future appends,as a sequence. E.g. ``[100, 10]``.  All
+            intended future appends,as a sequence. E.g. ``(100, 10)``.  All
             lengths must be in the positive int64 range, or `None`.  If a slot
             is `None` -- only supported for `SparseNDArray`, not `DenseNDArray`
             -- then the maximum possible size will be used. This makes a
--- a/apis/python/src/tiledbsoma/_common_nd_array.py
+++ b/apis/python/src/tiledbsoma/_common_nd_array.py
@@ -1,6 +1,6 @@
 """Common code shared by both NDArray implementations."""
 
-from typing import Optional, Sequence, Tuple, cast
+from typing import Optional, Sequence, Tuple, Union, cast
 
 import numpy as np
 import pyarrow as pa
@@ -26,7 +26,7 @@ class NDArray(TileDBArray, somacore.NDArray):
         uri: str,
         *,
         type: pa.DataType,
-        shape: Sequence[int],
+        shape: Sequence[Union[int, None]],
         platform_config: Optional[options.PlatformConfig] = None,
         context: Optional[SOMATileDBContext] = None,
     ) -> Self:
@@ -38,8 +38,12 @@ class NDArray(TileDBArray, somacore.NDArray):
         :param type: The Arrow type to be stored in the NDArray.
             If the type is unsupported, an error will be raised.
 
-        :param shape: the length of each dimension as a sequence, e.g., ``[100, 10]``.
-            All lengths must be in the positive int64 range.
+        :param shape: The maximum capacity of the dataframe, including room for any
+            intended future appends,as a sequence. E.g. ``[100, 10]``.  All
+            lengths must be in the positive int64 range, or `None`.  If a slot
+            is `None` -- only supported for `SparseNDArray`, not `DenseNDArray`
+            -- then the maximum possible size will be used. This makes a
+            `SparseNDArray` growable.
 
         :param platform_config: Platform-specific options used to create this Array,
             provided via ``{"tiledb": {"create": ...}}`` nested keys.
@@ -61,7 +65,7 @@ class NDArray(TileDBArray, somacore.NDArray):
     @property
     def shape(self) -> Tuple[int, ...]:
         """
-        Return length of each dimension, always a list of length ``ndim``
+        Return capacity of each dimension, always a list of length ``ndim``
         """
         return cast(Tuple[int, ...], self._handle.schema.domain.shape)
 
@@ -76,7 +80,7 @@ class NDArray(TileDBArray, somacore.NDArray):
 
 def _build_tiledb_schema(
     type: pa.DataType,
-    shape: Sequence[int],
+    shape: Sequence[Union[int, None]],
     create_options: TileDBCreateOptions,
     context: SOMATileDBContext,
     *,
@@ -85,8 +89,16 @@ def _build_tiledb_schema(
     _util.check_type("type", type, (pa.DataType,))
 
     # check on shape
-    if len(shape) == 0 or any(e <= 0 for e in shape):
-        raise ValueError("SOMA NDArray shape must be non-zero length tuple of ints > 0")
+    if is_sparse:
+        if len(shape) == 0 or any(e is not None and e <= 0 for e in shape):
+            raise ValueError(
+                "SOMASparseNDArray shape must be non-zero length tuple of positive ints or Nones"
+            )
+    else:
+        if len(shape) == 0 or any(e is None or e <= 0 for e in shape):
+            raise ValueError(
+                "SOMADenseNDArray shape must be non-zero length tuple of positive ints"
+            )
 
     if not pa.types.is_primitive(type):
         raise TypeError(
@@ -97,10 +109,20 @@ def _build_tiledb_schema(
     dims = []
     for n, e in enumerate(shape):
         dim_name = f"soma_dim_{n}"
+
+        slot_capacity = 2**63 if e is None else e
+        extent = min(slot_capacity, create_options.dim_tile(dim_name, 2048))
+        # TileDB requires that each signed-64-bit-int domain slot, rounded up to
+        # a multiple of the tile extent in that slot, be representable as a
+        # signed 64-bit int. So if the tile extent is 999, say, this will exceed
+        # 2**63 - 1.
+        if e is None:
+            slot_capacity -= extent
+
         dim = tiledb.Dim(
             name=dim_name,
-            domain=(0, e - 1),
-            tile=min(e, create_options.dim_tile(dim_name, 2048)),
+            domain=(0, slot_capacity - 1),
+            tile=extent,
             dtype=np.int64,
             filters=create_options.dim_filters(
                 dim_name,
--- a/apis/python/src/tiledbsoma/_common_nd_array.py
+++ b/apis/python/src/tiledbsoma/_common_nd_array.py
@@ -110,35 +110,16 @@ class NDArray(TileDBArray, somacore.NDArray):
         if not shape:
             raise ValueError("SOMA NDArrays must have a nonzero number of dimensions")
 
-        # check on shape
-        if is_sparse:
-            if any(e is not None and e <= 0 for e in shape):
-                raise ValueError(
-                    "SOMASparseNDArray shape must be non-zero length tuple of positive ints or Nones"
-                )
-        else:
-            if any(e is None or e <= 0 for e in shape):
-                raise ValueError(
-                    "SOMADenseNDArray shape must be non-zero length tuple of positive ints"
-                )
-
         dims = []
-        for n, e in enumerate(shape):
-            dim_name = f"soma_dim_{n}"
-
-            slot_capacity = 2**63 if e is None else e
-            extent = min(slot_capacity, create_options.dim_tile(dim_name, 2048))
-            # TileDB requires that each signed-64-bit-int domain slot, rounded up to
-            # a multiple of the tile extent in that slot, be representable as a
-            # signed 64-bit int. So if the tile extent is 999, say, this will exceed
-            # 2**63 - 1.
-            if e is None:
-                slot_capacity -= extent
-
+        for dim_idx, dim_shape in enumerate(shape):
+            dim_name = f"soma_dim_{dim_idx}"
+            dim_capacity, dim_extent = cls._dim_capacity_and_extent(
+                dim_name, dim_shape, create_options
+            )
             dim = tiledb.Dim(
                 name=dim_name,
-                domain=(0, slot_capacity - 1),
-                tile=extent,
+                domain=(0, dim_capacity - 1),
+                tile=dim_extent,
                 dtype=np.int64,
                 filters=create_options.dim_filters(
                     dim_name,
@@ -178,3 +159,12 @@ class NDArray(TileDBArray, somacore.NDArray):
             cell_order=cell_order,
             ctx=context.tiledb_ctx,
         )
+
+    @classmethod
+    def _dim_capacity_and_extent(
+        cls,
+        dim_name: str,
+        dim_shape: Optional[int],
+        create_options: TileDBCreateOptions,
+    ) -> Tuple[int, int]:
+        raise NotImplementedError("must be implemented by child class.")
--- a/apis/python/src/tiledbsoma/_common_nd_array.py
+++ b/apis/python/src/tiledbsoma/_common_nd_array.py
@@ -56,7 +56,7 @@ class NDArray(TileDBArray, somacore.NDArray):
         # the upper limit is exactly 2**63-1.
 
         context = context or SOMATileDBContext()
-        schema = _build_tiledb_schema(
+        schema = cls._build_tiledb_schema(
             type,
             shape,
             TileDBCreateOptions.from_platform_config(platform_config),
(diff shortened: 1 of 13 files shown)


apis/python/src/tiledbsoma/common_nd_array.py
```diff
@@ -38,8 +38,12 @@ def create(
         :param type: The Arrow type to be stored in the NDArray.
             If the type is unsupported, an error will be raised.
 
-        :param shape: the length of each dimension as a sequence, e.g., ``[100, 10]``.
-            All lengths must be in the positive int64 range.
+        :param shape: The maximum capacity of the dataframe, including room for any
+            intended future appends,as a sequence. E.g. ``[100, 10]``.  All
```
bkmartinjr, line 42: please use a tuple, not a list, in the example as it is not intended to be a mutable arg

apis/python/src/tiledbsoma/common_nd_array.py
```diff
@@ -38,8 +38,12 @@ def create(
...
+            lengths must be in the positive int64 range, or `None`.  If a slot
+            is `None` -- only supported for `SparseNDArray`, not `DenseNDArray`
+            -- then the maximum possible size will be used. This makes a
```
bkmartinjr, line 45: given that SOMA has a single dimension type for ndarray, and therefore a well-defined domain range, lets just say what the default size is going to be (ie.., the default, if None is specified, is a capacity of 2^64).

apis/python/src/tiledbsoma/common_nd_array.py
```diff
@@ -85,8 +89,16 @@ def _build_tiledb_schema(
     util.check_type("type", type, (pa.DataType,))
 
     # check on shape
-    if len(shape) == 0 or any(e <= 0 for e in shape):
-        raise ValueError("SOMA NDArray shape must be non-zero length tuple of ints > 0")
+    if is_sparse:
+        if len(shape) == 0 or any(e is not None and e <= 0 for e in shape):
+            raise ValueError(
+                "SOMASparseNDArray shape must be non-zero length tuple of positive ints or Nones"
```
bkmartinjr, line 95: NIT: encoding subclass knowledge in the parent is slightly "ewwww" :-)   Possible fix: do the sparse-specific defaulting in a subclass specialization of this method, and then call the super()._build_tiledb_schema() method.

apis/python/src/tiledbsoma/common_nd_array.py
```diff
@@ -97,10 +109,20 @@ def _build_tiledb_schema(
     dims = []
     for n, e in enumerate(shape):
         dim_name = f"soma_dim_{n}"
+
+        slot_capacity = 2**63 if e is None else e
```
bkmartinjr, line 113: see comment above - it feels to me that this defaulting belongs in the sparse subclass

maybe worth asking @thetorpedodog .

apis/python/src/tiledbsoma/common_nd_array.py
bkmartinjr: put another way - the class-specific requirements should not be pushed up into this general-purpose superclass

apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -41,6 +41,7 @@ def create(
         *,
         schema: pa.Schema,
         index_column_names: Sequence[str] = (SOMA_JOINID,),
+        shape: Optional[int] = None,
```
bkmartinjr, line 44: This signature (and the docstring below) do not handle cases where the indexed columns (aka dimensions) are non-integer.  How are (for example) date/time and float handled?

apis/python/src/tiledbsoma/dataframe.py
```diff
@@ -96,7 +103,7 @@ def count(self) -> int:
         """
         Return the number of rows in the dataframe. Same as `len(df)`.
         """
-        # A.domain.shape at the tiledb level gives us the 0..2^63 range which is not what we want
+        # A.domain.shape at the tiledb level gives us the 0..2^63-1 range which is not what we want
```
bkmartinjr, line 106: not sure I understand this comment - it may be obsolete.  How does `shape` or domain relate to count/nnz?

apis/python/src/tiledbsoma/dataframe.py
bkmartinjr: This also pointed out a major gap in the DataFrame unit test -- there is no test of indexing on column types other than int and string.

As I understand the spec, any user-defined dataframe column can be indexed, which means we need tests for float and other (non-string, non-int) types.

bkmartinjr: Additional issue - there can be more than one index column on a DataFrame.  So shouldn't the capacity be a sequence?

apis/python/src/tiledbsoma/sparse_nd_array.py
```diff
@@ -177,32 +174,6 @@ def __init__(self, sr: clib.SOMAReader, shape: NTuple):
         self.sr = sr
         self.shape = shape
 
-    def coos(self) -> SparseCOOTensorReadIter:
```
bkmartinjr, line 180: why did you delete `coos()`?  This should remain, as discussed with Seth.  It does not suffer from the same "scales with size of domain" issue that csr/csc have....

apis/python/src/tiledbsoma/common_nd_array.py
johnkerl: @bkmartinjr I found via unit test that tiledb core requires that the capacity, rounded up to an exact multiple of the extent, needs to be representable as a signed 64-bit integer

so in particular when a unit test (or anyone else) sets extent to a not-power-of-two number like 999 or 1000 then the create fails

so i went with this wording instead since users _cannot_ always get 2**63

i can add a comment to that effect if that helps

apis/python/src/tiledbsoma/common_nd_array.py
johnkerl: oh my bad @bkmartinjr i have tried to be _very_ discipliined about using tuples not lists!!!! thank you!

apis/python/src/tiledbsoma/common_nd_array.py
johnkerl: @bkmartinjr we had code that was de-duplicated and now it needs to be re-duplicated

i can move the code around, no problem

apis/python/src/tiledbsoma/dataframe.py
bkmartinjr: Added #956 956 on the issue of poor unit test coverage, which IMHO would help shake out this PR.

apis/python/src/tiledbsoma/sparse_nd_array.py
johnkerl: @bkmartinjr it uses `.shape`:

```diff
 -        assert query.X("raw").coos().concat() == pa.SparseCOOTensor.from_numpy(
-            raw_X["soma_data"].to_numpy(),
-            np.array(
-                [
-                    raw_X["soma_dim_0"].to_numpy(),
-                    raw_X["soma_dim_1"].to_numpy(),
-                ]
-            ).T,
-            shape=soma_experiment.ms["RNA"].X["raw"].shape,
-        )
- 
-```

-- we can keep it but we'll need a bit of a refactor I think (or, store current not-necessarily-tight bounding box as array metadata ...)

File: apis/python/src/tiledbsoma/read_iters.py [hunk 8]
```diff
@@ -60,55 +57,3 @@ def concat(self) -> RT:
         """
         arrow_tables = pa.concat_tables(TableReadIter(self.sr))
         return self._from_table(arrow_tables)
-
-
-class SparseCOOTensorReadIter(SparseTensorReadIterBase[pa.SparseCOOTensor]):
```
bkmartinjr, line 65: this should not be deleted, as it is still very much needed.  Only CSC/CSR suffer from the memory-use-scale-with-domain-size issue

apis/python/src/tiledbsoma/common_nd_array.py
thetorpedodog: I think this might work best broken into two parts:

 1. Here, just validate non-emptiness:
    ```python
    if not shape:
      raise ValueError("ndarrays must have a nonzero number of dimensions")
    ```
 2. Validate the values when building dimensions below…

apis/python/tests/test_basic_anndata_io.py
```diff
@@ -123,11 +123,11 @@ def test_import_anndata(adata, ingest_modes):
             with pytest.raises(ValueError):
                 exp.ms["RNA"].X["data"].read(coords=all2d)
 
-        # Check raw/X/data (sparse)
+        #        # Check raw/X/data (sparse)
         assert exp.ms["raw"].X["data"].metadata.get(metakey) == "SOMASparseNDArray"
         if have_ingested:
-            matrix = exp.ms["raw"].X["data"].read(coords=all2d).coos().concat()
```
bkmartinjr, line 129: please leave all support and tests for COO tensors in place.  They should _not_ be removed.

Adding tests for `tables` alongside COO Tensor is great!

apis/python/tests/test_dataframe.py
```diff
@@ -111,6 +111,53 @@ def test_dataframe(tmp_path, arrow_schema):
         assert not A.schema.allows_duplicates
 
 
+@pytest.mark.parametrize("appendable", [True, False])
```
bkmartinjr, line 114: as noted in #956, this lacks coverage for the DataFrame indexing, specifically,  multiple indexing columns (and corresponding correct result from `shape`), and the full gamut of indexed column types (eg., datetime, float, etc).

bkmartinjr: At a minimum, please add tests related to shape: ie., that if a dataframe has >1 index columns, it generates the right shape

apis/python/src/tiledbsoma/read_iters.py
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/952#discussion_r1108907410

apis/python/tests/test_basic_anndata_io.py
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/952#discussion_r1108907410

apis/python/src/tiledbsoma/common_nd_array.py
thetorpedodog: …so that it looks something like

```python
# Here...
for n, size in enumerate(shape):
  # ...
  domain, extent = cls._dim_domain_extent(size)
  dims.append(tiledb.Dim(
    name=f"soma_dim_{n}",
    domain=(0, domain),
    extent=extent,
  ))

# Elsewhere...

class DenseNDArray:
  @classmethod
  def _dim_domain_extent(cls, name: str, size: Optional[int]) -> Tuple[int, int]:
    if not isinstance(size, int) or size <= 0:
      raise ValueError(...)
    return ...

class SparseNDArray:
  @classmethod
  def _dim_domain_extent(cls, str, size):
    ...
```

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -436,7 +436,11 @@ def _write_dataframe(
         soma_df = factory.open(df_uri, "w", soma_type=DataFrame)
     except DoesNotExistError:
         soma_df = DataFrame.create(
-            df_uri, schema=arrow_table.schema, platform_config=platform_config
+            df_uri,
+            schema=arrow_table.schema,
+            platform_config=platform_config,
+            ### XXX TEMP
```
thetorpedodog, line 442: Something to note?

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -487,10 +491,12 @@ def create_from_matrix(
     try:
         soma_ndarray = cls.open(uri, "w", platform_config=platform_config)
     except DoesNotExistError:
+        # A SparseNDArray must be appendable in soma.io.
+        shape = matrix.shape if cls == DenseNDArray else (None,) * len(matrix.shape)
```
thetorpedodog, line 495: `[None for _ in matrix.shape]`, maybe?

Also maybe use `cls.is_sparse` instead of the class check.

(Neither of these are huge issues.)

apis/python/src/tiledbsoma/io/ingest.py
thetorpedodog: what was I saying here? maybe "did something get missed here?" instead

apis/python/tests/test_dense_nd_array.py
```diff
@@ -114,6 +114,24 @@ def test_dense_nd_array_reshape(tmp_path):
         assert a.reshape((100, 10, 1))
 
 
+@pytest.mark.parametrize("appendable", [True, False])
```
bkmartinjr, line 117: the test is fine, but the naming/concept of "appendable" is misleading.

The test should simply confirm that DenseNDArray requires a shape to be specified.  I.e., there is no defaulting on the shape param to create().

apis/python/tests/test_experiment_query.py
```diff
@@ -170,16 +170,6 @@ def test_experiment_query_coords(soma_experiment):
             .concat()
         )
         assert query.X("raw").tables().concat() == raw_X
-        assert query.X("raw").coos().concat() == pa.SparseCOOTensor.from_numpy(
```
bkmartinjr, line 173: please do _not_ delete COO Tensor support!

apis/python/tests/test_sparse_nd_array.py
```diff
@@ -205,69 +205,6 @@ def tables_are_same_value(a: pa.Table, b: pa.Table) -> bool:
     return True
 
 
-@pytest.mark.parametrize(
-    "shape,format",
-    [
-        ((10,), "coo"),
```
bkmartinjr, line 211: please do _not_ delete COO Tensor support.

apis/python/tests/test_sparse_nd_array.py
```diff
@@ -323,6 +260,50 @@ def test_sparse_nd_array_read_as_pandas(
         assert not A.schema.allows_duplicates
 
 
+@pytest.mark.parametrize("appendable", [True, False])
```
bkmartinjr, line 263: appendable is not the correct concept, and is very misleading.

We have simply added a default "shape" to the create() method.  All sparse arrays are "appendable" as you are using the term, and _no_ sparse arrays allow their domain/shape to be changed after creation.

For clarity, I suggest picking a different term.  All the code below is doing is incremental writes of different parts of the sparse array (i.e., adding explicit values to the coordanates).

apis/python/tests/test_sparse_nd_array.py
```diff
@@ -341,10 +322,6 @@ def test_empty_read(tmp_path):
         # These work as expected
         coords = (slice(None),)
         assert sum(len(t) for t in a.read(coords).tables()) == 0
-        # Fails due to ARROW-17933
-        # assert sum(t.non_zero_length for t in a.read(coords).coos()) == 0
```
bkmartinjr, line 345: please do not delete COO Tensor support

apis/python/tests/test_sparse_nd_array.py
```diff
@@ -469,7 +431,7 @@ def test_csr_csc_2d_read(tmp_path, shape):
     # We want to test read_format == "none_of_the_above", to ensure it throws NotImplementedError,
     # but that can't be gotten past typeguard.
     "read_format",
-    ["table", "coo", "csr", "csc"],
```
bkmartinjr, line 472: please do not delete COO tensor support (I'll stop noting this in further comments, but there are other places it is deleted)

apis/python/src/tiledbsoma/sparse_nd_array.py
bkmartinjr: it is fine if it uses shape -- why is this a problem?

I suspect you are misunderstanding why use of `.shape` is problematic.  I'm happy to do a IRT discussion to detail it if useful.

bkmartinjr (dismissed):
Several large issues:
* COO tensor support was incorrectly removed. It is no different than Table support under the covers (i.e., handles arbitrarily big domains just fine), and needs to be maintained.
* There are some fairly big issues with the DataFrame changes - it needs to support multi-column indexing (ie., len(shape) > 1), and it needs to handle non-integer columns (e.g., float)
* As an aside, part of the issue with DataFrame is that the unit tests lack any coverage of these important indexing capabilities (not the fault of this PR, but certainly should be fixed before 1.0).

Some other minor code organizational issues noted as well, in particular the migration of ndarray class-specific code & knowledge into the ndarray superclass.

apis/python/src/tiledbsoma/dataframe.py
johnkerl: @bkmartinjr stale indeed

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: temp/debug; deleted; thank you

apis/python/src/tiledbsoma/sparse_nd_array.py
johnkerl: @bkmartinjr I did indeed misunderstood:

Before, my (clearly mistaken) belief from our conversation was "Uses `.shape`? Let it go."

Now, I know this proof of concept does work quite well:

```
#!/usr/bin/env python

import tiledbsoma as soma
import pyarrow as pa

exp = soma.Experiment.open("/Users/johnkerl/s/v/ileum2")

x = exp.ms["RNA"].X["data"]

arrow_table = x.read().tables().concat()

coo_data = arrow_table.column("soma_data").to_numpy()

coo_coords = np.array([
    arrow_table.column(f"soma_dim_{n}").to_numpy() for n in range(len(x.shape))
]).T

scoot = pa.SparseCOOTensor.from_numpy(coo_data, coo_coords, shape=x.shape)
```

even with

```
>>> x.shape
(9223372036854773760, 9223372036854773760)
>>> scoot.shape
(9223372036854773760, 9223372036854773760)
```

So I can confidently revert all the redlines of the COO bits on this PR.

johnkerl: COO read-iterator restored on latest commit

apis/python/tests/test_experiment_query.py
johnkerl: Restored on most recent commit as discussed above

apis/python/tests/test_sparse_nd_array.py
johnkerl: Fixed

apis/python/tests/test_basic_anndata_io.py
johnkerl: Restored on most recent commit as discussed above

apis/python/src/tiledbsoma/read_iters.py
johnkerl: Restored on most recent commit as discussed above

apis/python/src/tiledbsoma/common_nd_array.py
johnkerl: done

apis/python/src/tiledbsoma/common_nd_array.py
johnkerl: done

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: done

apis/python/src/tiledbsoma/dataframe.py
johnkerl: As of the most recent commit, this PR retracts any attempts at explicit shaping of `DataFrame`. Those questions will now be addressed on a separate PR.

apis/python/tests/test_dataframe.py
johnkerl: As of the most recent commit, this PR retracts any attempts at explicit shaping of DataFrame. Those questions will now be addressed on a separate PR.

The asks here will not be forgotten; they are tracked on #961.

apis/python/src/tiledbsoma/_common_nd_array.py
```diff
@@ -38,14 +38,25 @@ def create(
...
+            intended future appends,as a sequence. E.g. ``(100, 10)``.  All
+            lengths must be in the positive int64 range, or `None`.  If a slot
+            is `None` -- only supported for `SparseNDArray`, not `DenseNDArray`
+            -- then the maximum possible int64 size will be used. This makes a
+            `SparseNDArray` growable.
 
         :param platform_config: Platform-specific options used to create this Array,
             provided via ``{"tiledb": {"create": ...}}`` nested keys.
         """
+        # Implementor note: we carefully say "maximum possible in64 size" rather than 2**63-1. The
```
atolopko-czi, line 51:
```suggestion
        # Implementor note: we carefully say "maximum possible int64 size" rather than 2**63-1. The
```

apis/python/src/tiledbsoma/_common_nd_array.py
```diff
@@ -73,70 +84,92 @@ def reshape(self, shape: Tuple[int, ...]) -> None:
...
+            dim_name = f"soma_dim_{n}"
+
+            slot_capacity = 2**63 if e is None else e
+            extent = min(slot_capacity, create_options.dim_tile(dim_name, 2048))
+            # TileDB requires that each signed-64-bit-int domain slot, rounded up to
+            # a multiple of the tile extent in that slot, be representable as a
+            # signed 64-bit int. So if the tile extent is 999, say, this will exceed
+            # 2**63 - 1.
+            if e is None:
+                slot_capacity -= extent
```
thetorpedodog, line 131: This should probably still be separated out as in [my previous suggestion](https://github.com/single-cell-data/TileDB-SOMA/pull/952#discussion_r1108916830), so that this doesn’t have to special case sparse vs. not sparse, and the "infinite capacity" case is easier to follow.

johnkerl: @thetorpedodog something is wrong -- i very much split these up late last night per your (excellent) suggestions

let me check for a merge fail or somesuch b/c i _did_ do this ... 😬

thetorpedodog: [such things](https://github.com/single-cell-data/TileDB-SOMA/pull/964) befall us all from time to time

johnkerl: looking at https://github.com/single-cell-data/TileDB-SOMA/commit/1a25808c8b78951be59c7a94423dfd67d70c26b5 ...

johnkerl: @thetorpedodog https://github.com/single-cell-data/TileDB-SOMA/pull/952/commits/e0e31e7ed54d15eb88726265d50fa49aa3352caf -- sorry for the confusion!

apis/python/tests/test_dataframe.py
johnkerl: Action plan here:
https://github.com/single-cell-data/TileDB-SOMA/pull/952#issuecomment-1434690695

apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -161,6 +160,38 @@ def _set_reader_coord(
...
+        dim_shape: Optional[int],
+        create_options: TileDBCreateOptions,
+    ) -> Tuple[int, int]:
+        """
+        Given a user-specified shape (maybe ``None``) along a particular dimension,
+        returns a tuple of the TileDB capacity and extent for that dimension, suitable
+        for schema creation. If the user-specified shape is None, the largest possible
+        int64 is returned for the capacity.
+        """
+        if dim_shape is not None:
```
thetorpedodog, line 176: maybe put the `dim_shape is None` case first to reduce the number of negations in the `if`

thetorpedodog (approved):
One small suggestion about if-case ordering. Other than that the implementation looks good, and just sync up the docs when the spec change is in.

apis/python/src/tiledbsoma/_sparse_nd_array.py
johnkerl: done

johnkerl:
Merging on accept from @thetorpedodog, with all issues raised by @bkmartinjr resolved ... @bkmartinjr if there is anything more you want done here please let me know and I will action it ASAP.

PR #953, opened 2023-02-16
Remove matplotlib pin since scanpy is fixed.

thetorpedodog:
Ensures installation of scanpy 1.9.2, which includes the matplotlib fix: https://github.com/scverse/scanpy/releases/tag/1.9.2


johnkerl (approved):
🚀

PR #958, opened 2023-02-16
[c++/python] Use `py::print` in libtiledbsoma11 to print, not `cout`

thetorpedodog:
Even after #887 was submitted, I would occasionally see stats output written to the console after tests completed (when running with `-s`, to disable output capture by default). I believe this is due to buffering within `std::cout` where the text may not actually hit the fd by the time the function returns and the test ends.

It could have also potentially introduced confusing output:

    print("before")
    tiledbsoma_stats_dump()
    print("after")

Depending upon how Python internally buffers `sys.stdout` and how that interacts with the system, the "before" could potentially appear after the `tiledbsoma_stats_dump` output (and it may be possible for "after" to appear *before* it? not sure about that).

Just letting Python handle it all with `py::print` is way easier.

### DIFF
--- a/apis/python/tests/test_util_tiledb.py
+++ b/apis/python/tests/test_util_tiledb.py
@@ -4,7 +4,7 @@ import pytest
 import tiledbsoma as soma
 
 
-def test_stats(tmp_path, capfd: pytest.CaptureFixture[str]):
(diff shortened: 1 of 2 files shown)


johnkerl (approved):
🚢

PR #962, opened 2023-02-17
[python] Use `setuptools_scm` to get `TileDB-SOMA-Py` version

nguyenv:
**Issue and/or context:**

The initial motivation behind this change is that the conda builds for the Python API were not pip installing properly due to the issues with the `version.py` ad-hoc script (source tarballs were missing a `RELEASE-VERSION` file). As noted in the previous `version.py` script, we prefer to use `setuptools_scm` anyways as it is more standard and used across all the TileDB repos.

**Changes:**

Previously we were using an ad-hoc script to calculate the version number due to issues where the Python package is in a subdirectory rather than the top level. The `setuptools_scm` setup from `TileDB-VCF`, which uses the same directory structure as `TileDB-SOMA`,  has now been ported over in order to solve this issue.

**Notes for Reviewer:**

N/A


johnkerl (commented):
LGTM but I defer to @thetorpedodog (especially in the absence of @gsakkis)

apis/python/setup.py
```diff
@@ -196,5 +196,11 @@ def run(self):
     },
     python_requires=">=3.7",
     cmdclass={"build_ext": build_ext, "bdist_wheel": bdist_wheel},
-    version=version.getVersion(),
+    use_scm_version={
+        "version_scheme": "guess-next-dev",
+        "local_scheme": "dirty-tag",
```
thetorpedodog, line 201: Can we use the defaults here? That way, it will look like `0.5.0a12.dev13+g[short revision sha]` (or `+g[short revision sha].d[datecode]` for dirty checkouts), which is useful for if you are doing work across repos.

nguyenv:
> LGTM but I defer to @thetorpedodog (especially in the absence of @gsakkis)

Ty - adding them to the list of reviewers.

apis/python/version.py
```diff
@@ -1,135 +1,4 @@
...
-def getVersion():
-    release_version = readReleaseVersion()
-    version = readGitVersion() or release_version
-    if not version:
-        raise ValueError("Cannot find the version number")
-    if version != release_version:
-        writeReleaseVersion(version)
-    return version
+# file generated by setuptools_scm
+# don't change, don't track in version control
```
thetorpedodog, line 2: don’t forget to `git rm` this and add it to `.gitignore`.

thetorpedodog (changes requested):
I’m not sure why this seems to be working now, given that we weren’t using it before (cf. the notice previously in `version.py`). However, I’m not complaining.

thetorpedodog:
nb. I think the build failure on the latest rev is from the `import version` here, which no longer needs to happen:

https://github.com/single-cell-data/TileDB-SOMA/blob/c79ee23bbb40d0e90b2a2aab9a2be9e6358a1123/apis/python/setup.py#L28-L30

(at the same time the `this_dir` variable can be made a local in `find_or_build_package_data`, since that’s the only place it’s used anymore)

nguyenv:
> nb. I think the build failure on the latest rev is from the `import version` here, which no longer needs to happen:
> 
> https://github.com/single-cell-data/TileDB-SOMA/blob/c79ee23bbb40d0e90b2a2aab9a2be9e6358a1123/apis/python/setup.py#L28-L30
> 
> (at the same time the `this_dir` variable can be made a local in `find_or_build_package_data`, since that’s the only place it’s used anymore)

Yes thank you! That solved it.

apis/python/setup.py
```diff
@@ -82,6 +80,9 @@ def libtiledbsoma_exists():
 
 
 def find_or_build_package_data(setuptools_cmd):
+    this_dir = pathlib.Path(__file__).parent.absolute()
+    sys.path.insert(0, str(this_dir))
```
thetorpedodog, line 84: the `sys.path` modification is no longer needed either

apis/python/setup.py
```diff
@@ -82,6 +80,9 @@ def libtiledbsoma_exists():
 
 
 def find_or_build_package_data(setuptools_cmd):
+    this_dir = pathlib.Path(__file__).parent.absolute()
```
thetorpedodog, line 83: maybe move this under `# Set up paths` so that it’s right next to `scripts_dir`?

thetorpedodog (approved):
Two small things and this is good to go. I’m extremely pleased that the standard `setuptools_scm` setup now works for us. Thanks!

PR #964, opened 2023-02-17
Add comments on why we expect replace tests to fail.

thetorpedodog:
area man forgets to actually commit changes he intended to make

### DIFF
--- a/apis/python/tests/test_collection.py
+++ b/apis/python/tests/test_collection.py
@@ -142,10 +142,12 @@ def test_collection_mapping(soma_object, tmp_path):
     assert len(c) == 1
     assert [k for k in c] == ["mumble"]
 
-    pytest.xfail(reason="currently supports one entry modification at a time")
-    del c["mumble"]
-    assert "mumble" not in c
-    assert not c.get("mumble", False)
+    # TEMPORARY: This should no longer raise an error once TileDB supports
+    # replacing an existing group member.
+    with pytest.raises(soma.SOMAError):
+        del c["mumble"]
+        assert "mumble" not in c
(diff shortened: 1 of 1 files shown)


apis/python/tests/test_collection.py
```diff
@@ -155,8 +157,10 @@ def test_delete_add(soma_object, tmp_path: pathlib.Path):
 
     with soma.open(tmp_uri, "w", soma_type=soma.Collection) as update:
         del update["porkchop sandwiches"]
-        pytest.xfail(reason="currently does not support removal/re-addition")
-        update["porkchop sandwiches"] = soma_object
+        # TEMPORARY: This should no longer raise once TileDB supports replacing
+        # an existing group member.
+        with pytest.raises(soma.SOMAError):
+            update["porkchop sandwiches"] = soma_object
```
atolopko-czi, line 163: If this _method_ had the @pytest.xfail annotation, wouldn't that forgo the need to catch the exception? (And per `xfail`,  it would fail once it actually works, i.e. TileDB provides the necessary supports)

thetorpedodog: The reason I want to avoid `xfail` on the method is because that will swallow *any* failure in the test (i.e., if something else breaks, it’s not visible). This instead verifies that only the part we want to not work doesn’t work. This should be temporary (replace support should come sooner than later).

This also highlights a weird mismatch between the two types of `xfail`: `@pytest.mark.xfail(...)` says “execute this, but if it fails that’s OK”, where `pytest.xfail()` in a test says “stop executing immediately mark the test xfailed”.

atolopko-czi (approved):

Issue #960, opened 2023-02-17
[c++/python] Handle `DataFrame` index types beyond int/string


johnkerl:
Split out from #922 

Currently:

* `libtiledbsoma.cc` only allows int and string indices: https://github.com/single-cell-data/TileDB-SOMA/blob/0.5.0a11/libtiledbsoma/src/pyapi/libtiledbsoma.cc#L328-L350
* The underlyling `soma_reader.h` / `soma_reader.cc` support templated types: https://github.com/single-cell-data/TileDB-SOMA/blob/0.5.0a11/libtiledbsoma/include/tiledbsoma/soma_reader.h#L152-L153

Therefore this should be a straightforward elaboration in `libtiledbsoma.cc`.

Work here will, along the way, address the unit-test checklist from #956.

Status:

* [x] `int64` and `string` -- status quo before this issue
* `int32`, `int16`, `int8`, `uint64`, `uint32`, `uint16`, `uint8`, `float32`, `float64`
  * [x] Python #968 
  * [ ] R #984 -- tracked in a separate issue
* `pyarrow.timestamp("s")`, `pyarrow.timestamp("ms")`, `pyarrow.timestamp("us")`, `pyarrow.timestamp("ns")`
  * [x] Python #985 
  * [ ] R #984 -- tracked in a separate issue
* [ ] `bool`: TileDB currently does not support `bool` as an index type
  * This is a low-pri feature request for TileDB Embedded: SC 25805

johnkerl:
Closing as all remaining issues have their own tracking tasks

Issue #956, opened 2023-02-16
[python] Augment `DataFrame` unit tests once `SOMAReader` supports index types beyond int/string


bkmartinjr:
The current DataFrame unit tests are lacking in a couple of areas:
* tests that utilize more than one indexing column
* tests that cover the full gamut of types in an indexed column

Examples:
* indexing on a float-typed column
* indexing on both a string, float and int-typed column

etc.

johnkerl:
* Tests that utilize more than one indexing column are here:
  * https://github.com/single-cell-data/TileDB-SOMA/blob/0.5.0a11/apis/python/tests/test_dataframe.py#L369-L384
  * https://github.com/single-cell-data/TileDB-SOMA/blob/0.5.0a11/apis/python/tests/test_dataframe.py#L678-L703
* These cover the full gamut of types currently supported by `SOMAReader` in an indexed column 
  * As tracked at #922 `SOMAReader` currently only handles int and string indices
  * https://github.com/single-cell-data/TileDB-SOMA/blob/0.5.0a11/libtiledbsoma/src/pyapi/libtiledbsoma.cc#L328-L350
  * Indexing on float depends on 922
  * Indexing on a combination of string/int exists as linked above
  * Indexing on a combination of string/float/int depends on 922
 
cc @maniarathi I believe this is a non-blocker for the rc _unless_ we decide that we want `SOMAReader`'s current inability to index `DataFrame` on other than int/string to itself be a blocker for the rc. No more work can be done on this particular issue until that `SOMAReader` work is done.

johnkerl:
Dup of #960, as every remaining ask on this PR is being implemented in issue #960 / PR #968

PR #968, opened 2023-02-18
[python] Handle `DataFrame` index types beyond int/string -- except timestamps and bools

johnkerl:
**Issue and/or context:** #960 and #956

**Changes:**

Arrow types `timestamp_s`, `timestamp_ms`, `timestamp_us`, and `timestamp_ns` handled on a subsequent PR. This current PR is big enough for now.

**Notes for Reviewer:**

Summary:

See the block comments in `libtiledbsoma.cc`. TL;DR is with the combination of mixable types like `float` and `double`, or `int16` and `int64`, etc., `pybind11`'s overloading by type _does not resolve these types_  -- it does first-found and _always_ finds `float` before `double` if the overloaded `float` is defined first in the source code. I proved `pybind11` was misbehaving by logging `sizeof(T)` in `soma_reader.h`.

More information on why overloading a single name `set_dim_points` (and likewise `set_dim_ranges`) in `pybind11` does not and cannot work now that we have multiplie int and float types:

* See [this commit](https://github.com/single-cell-data/TileDB-SOMA/tree/a8120b7d5b368904374b50b0b05809b728504152) on the branch [kerl/pybind11-overload-failure-example ](https://github.com/single-cell-data/TileDB-SOMA/tree/kerl/pybind11-overload-failure-example)
  * On this example-only commit I overloaded only `int64` and `int16`, and `float32` and `float64` -- to keep this example simple
  * For the branch used on this PR, I am handling all 8 int types
* [Here](https://github.com/single-cell-data/TileDB-SOMA/blob/a8120b7d5b368904374b50b0b05809b728504152/libtiledbsoma/src/pyapi/libtiledbsoma.cc#L290-L318) I try to overload `set_dim_points` solely by type
  * For the branch used on this PR, for the reasons described here, I am no longer relying solely on name-overloading
* Here is a sample invocation: https://gist.github.com/johnkerl/6f7f25ee8749aea9951ce6212d43124d
  * Note that `help(sr.set_dim_points)` shows these, which show that it sends `string` and `int` to different places (the status quo before this PR) but the two different int types in this example are not differentiated, nor the two different float types in this exaample:
 ```
set_dim_points(self: tiledbsoma.libtiledbsoma.SOMAReader, arg0: str, arg1: List[str]) -> None # once
set_dim_points(self: tiledbsoma.libtiledbsoma.SOMAReader, arg0: str, arg1: List[int]) -> None # twice
set_dim_points(self: tiledbsoma.libtiledbsoma.SOMAReader, arg0: str, arg1: List[float]) -> None
```
  * As a result we get the error `RuntimeError: Static type (FLOAT64) does not match expected type (FLOAT32)` since the `double` impl is first-found but the array schema has `float`.
* Also (for bonus) here's a [longer repro script to experiment with](https://gist.github.com/johnkerl/2571f2fc10680e03e8576d5852019315)

### CODE REVIEW

File: libtiledbsoma/src/pyapi/libtiledbsoma.cc [hunk 1]
```diff
@@ -283,24 +283,110 @@ PYBIND11_MODULE(libtiledbsoma, m) {
...
                 &SOMAReader::set_dim_points))
 
         .def(
             "set_dim_points",
+            static_cast<void (SOMAReader::*)(
+                const std::string&, const std::vector<int64_t>&)>(
+                &SOMAReader::set_dim_points))
+
+        .def(
+            "set_dim_points_string_or_bytes",
```
gspowley, line 320: This looks like the same as the `set_dim_points` with `std::vector<std::string>` above. Are they both needed for pybind11 to bind properly?

libtiledbsoma/src/pyapi/libtiledbsoma.cc
```diff
@@ -322,18 +408,113 @@ PYBIND11_MODULE(libtiledbsoma, m) {
                     uintptr_t arrow_array_ptr = (uintptr_t)(&arrow_array);
 
                     // Call array._export_to_c to get arrow array and schema
+                    //
+                    // If ever a NumPy array gets in here, there will be an
+                    // exception like "AttributeError: 'numpy.ndarray' object
+                    // has no attribute '_export_to_c'".
+                    // TODO: try/catch with something more helpful of the form
```
gspowley, line 415: Did you plan to address the TODO in this PR?

libtiledbsoma/src/pyapi/libtiledbsoma.cc
```diff
@@ -322,18 +408,113 @@ PYBIND11_MODULE(libtiledbsoma, m) {
...
+                    // "Maybe not an Arrow array".
                     array.attr("_export_to_c")(
                         arrow_array_ptr, arrow_schema_ptr);
 
-                    // 'l' = arrow data type int64
                     if (!strcmp(arrow_schema.format, "l")) {
                         tcb::span<int64_t> data{
                             (int64_t*)arrow_array.buffers[1],
```
gspowley, line 422: We have been using `arrow_array.buffers[1]` as the data buffer. It would be safer to use something like the code below, because the data buffer will be in `arrow_array.buffers[2]` if offsets are present.

```c++
int data_index = arrow_array.n_buffers - 1;
...
(int64_t*)arrow_array.buffers[data_index],
```

Existing if/else logic may already avoid using the wrong buffer index, but explicitly avoiding the issue is easy to add.

libtiledbsoma/src/pyapi/libtiledbsoma.cc
```diff
@@ -283,24 +283,110 @@ PYBIND11_MODULE(libtiledbsoma, m) {
...
+            "set_dim_points_uint8",
+            static_cast<void (SOMAReader::*)(
+                const std::string&, const std::vector<uint8_t>&)>(
+                &SOMAReader::set_dim_points))
+
+        // The above short functions are expected to be invoked when the coords
+        // are Python list/tuple, or NumPy arrays.  Arrow arrays are in this
+        // long if-else-if function.
+        .def(
+            "set_dim_points_arrow_array",
```
gspowley, line 389: Maybe call the function `set_dim_points_arrow` to be less specific? It could be an `Array` or a `ChunkedArray`.

gspowley (approved):
LGTM, nice work!

libtiledbsoma/src/pyapi/libtiledbsoma.cc
johnkerl: @gspowley good call & i was thinking the same thing ... best to just "let go" of the plain `set_dim_points` everywhere ...

apis/python/src/tiledbsoma/_tiledb_array.py
```diff
@@ -124,19 +124,17 @@ def _set_reader_coord(
...
             return True  # No constraint; select all in this dimension
 
-        # TODO: support index types other than int/string once libtiledbsoma.cc does
-        # https://github.com/single-cell-data/TileDB-SOMA/issues/960
         if isinstance(coord, int):
-            sr.set_dim_points(dim.name, [coord])
+            sr.set_dim_points_int64(dim.name, [coord])
             return True
-        if isinstance(coord, slice):
+        if isinstance(coord, slice):  # XXX and need to check if this is an integer type
```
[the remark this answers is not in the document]
johnkerl: yes very much

libtiledbsoma/src/pyapi/libtiledbsoma.cc
johnkerl: Probably not now that I understand more

I'll take this out

libtiledbsoma/src/pyapi/libtiledbsoma.cc
johnkerl: good call

libtiledbsoma/src/pyapi/libtiledbsoma.cc
johnkerl: @gspowley oops, I merged with this issue unresolved. Putting up a follow-up PR now ...

johnkerl: @gspowley #986

PR #969, opened 2023-02-18
[python] Fixing PyPI packaging

mlin:


mlin:
@nguyenv @thetorpedodog Without prejudice on the way forward with version.py / setuptools_scm -- I got the PyPI [packaging build working again](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/4209979028) with slight modifications to Vivian's lib/lib64 patch and reverting back to version.py.

So, we just need to decide whether to (i) adapt version.py for the conda build or (ii) adapt the PyPI sdist build to work with setuptools_scm, given the current repo directory structure (or changing it).

To be clear, I have no love for the version.py; I pulled that in only to unblock myself when I wasn't able to figure out (ii) a couple months back, and would be equally happy to use setuptools_scm if we find a way now!

If the version.py problem with conda is just the tarball missing the `RELEASE-VERSION` file, then that's probably an easy fix. That file merely tells the script what version to report, as a single text line, once the code is no longer in a git repo from which that information can be "detected."

nguyenv:
If we can use `setuptools_scm` instead then that would be great but maybe best for now to just revert back to the previous `version.py` script to get us all unblocked. We can think about how to replace it with `setuptools_scm` in the future. 

I believe all I need to get unblocked for the conda work is to have the `RELEASE-VERSION` in the  source tarball.

mlin:
@nguyenv Ok, spitballing here, any of this might be wildly incorrect:

The new `libtiledbsoma_global_exists()` path in setup.py bundles a systemwide libtiledbsoma.so (if present) instead of compiling it from the included source code. That will happen in the course of doing one of two things, installing the package locally or building a bdist/wheel.

1. If setup.py is installing the package locally, then we don't want it to make *another copy* of the libtiledbsoma.so, we just want to use the systemwide one at runtime. This includes the conda packaging case.
2. If building a bdist/wheel, then we would *never* want to include a systemwide libtiledbsoma.so of unknown (to setup.py) provenance.

So I think what we want instead is a path in which `setup.py install` will *do nothing* with libtiledbsoma, just install the python code and leave the caller responsible to ensure a compatible system-wide version will be available. That's what conda will do for us at both packaging and runtime, right? 

Maybe this path is simply activated by an environment variable you set whilst building the conda package. Then you can use the sdist tarball from PyPI, which will have the `RELEASE-VERSION` file already.

mlin:
I should say, my latent assumption in the preceding comment is that there's a separate conda [sub]package for the libtiledbsoma C++ library, which we assume is already installed when we're building the python-specific package, and at runtime. (I think that was the plan, but thought I should add this since nothing I wrote above makes sense otherwise)

nguyenv:
Yes, you are correct about everything you've said above. If we already have `libtiledbsoma.so` installed, then don't rebuild it and link to that. And yes for the conda packages, we build the C++ libtiledbsoma library first, which is its own package, and then we separately build the Python-specific package which depends on the C++ lib.

How I have `meta.yaml` set up for that now is that it's using a single source for both the C++ and Python package:
```
source:
  url: https://github.com/single-cell-data/TileDB-SOMA/archive/{{ version }}.tar.gz
  sha256: {{ sha256 }}
```

When we were only building the Python package before, it used the PyPI sdist tarball and everything worked fine. The Github source code tarball is missing that `RELEASE-VERSION` file in `apis/python`. To get around it, I think I can [define multiple sources](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#source-from-multiple-sources) where the C++ package uses the Github source code tarball and the Python package uses the PyPI sdist tarball.

mlin:
@nguyenv Cool thanks...I promoted this diff out of draft in hopes of unblocking the main build, and for conda, I also looked more at how TileDB-VCF works and suggest we put together a follow-up PR borrowing further from it:

1. The one complication with excluding libtiledbsoma.so from the conda python package is ensuring we can later find & import it *at runtime*. [TileDB-VCF has runtime init code](https://github.com/TileDB-Inc/TileDB-VCF/blob/d3390e2a77fabf73f882989692406af02ce2f2d0/apis/python/src/tiledbvcf/__init__.py#L9-L26) that achieves this by finding it and transiently instantiating a dummy `ctypes.CDLL` to load it. *Then*, [`from . import libtiledbvcf`](https://github.com/TileDB-Inc/TileDB-VCF/blob/d3390e2a77fabf73f882989692406af02ce2f2d0/apis/python/src/tiledbvcf/dataset.py#LL7C2-L7C2) works, even though libtiledbvcf.so isn't bundled with the package, because it's already been dynlinked into the python process (I'm only 80% sure of this part). We have [`from . import libtiledbsoma`](https://github.com/single-cell-data/TileDB-SOMA/blob/213f9bdf4a160e525caf576c834af5a07aed196b/apis/python/src/tiledbsoma/_tiledb_array.py#L9) that we'd need to front-run similarly.
2. The TileDB-VCF setup.py takes a [command-line argument `LIBTILEDBVCF_PATH`](https://github.com/TileDB-Inc/TileDB-VCF/blob/d3390e2a77fabf73f882989692406af02ce2f2d0/apis/python/setup.py#L52-L53) (could be an environment variable too). That (i) [overrides the in-tree search path](https://github.com/TileDB-Inc/TileDB-VCF/blob/d3390e2a77fabf73f882989692406af02ce2f2d0/apis/python/setup.py#L97-L111) and also (ii) [excludes the shared library file from the package](https://github.com/TileDB-Inc/TileDB-VCF/blob/d3390e2a77fabf73f882989692406af02ce2f2d0/apis/python/setup.py#L191) to avoid making the extra copy of it. Then the conda python package builder [sets that to $CONDA_PREFIX](https://github.com/TileDB-Inc/TileDB-VCF/issues/47#issuecomment-595021300) so that it uses the version installed by the related conda libtiledbvcf subpackage. I think this is a good idea to make the operator be quite explicit if they want to use some other version of the library instead of building from in-tree source.
3. Regarding RELEASE-VERSION: the multiple sources approach sounds promising. Another possibility: if it's possible to (i) set the source to the actual git repo + hash/tag instead of a source tarball, and (ii) run `pushd; python setup.py check; popd` in the resulting repo checkout, that will have the side-effect of generating RELEASE-VERSION in the expected location, without "dirtying" the repo since it's in .gitignore. This has to be done in a git repo checkout (with tags fetched), not a source tarball, because version.py consults the git history.

nguyenv (approved):

johnkerl (approved):

johnkerl:
Merging after conversation with @nguyenv

PR #979, opened 2023-02-21
[python] Move `options` implementations modules to private

thetorpedodog:
This sets the sub-modules of `options` to be "private". The name `SOMATileDBContext` remains exported, but internal-only things are now explicitly internal-only.


johnkerl (approved):
🚢

Issue #919, opened 2023-02-11
[python] Point-in-time child object access fails when using explicit write_timestamp


atolopko-czi:
**Describe the bug**
When a _single_ process attempts to access a DataFrame as a child of a Collection that was created by an earlier `create()` (using a independent context manager block) , the child access _transiently_ fails to "see" the DataFrame _if_ 1) a sibling collection was created before the DataFrame, and 2) a `write_timestamp` is explicitly provided via the `SOMATileDBContext`.

Apologies, but I can't whittle this down to a simpler case.  All of these conditions seem to be necessary.

**To Reproduce**
https://gist.github.com/atolopko-czi/26683305258a9f77a57ccc364916338f

**Versions (please complete the following information):**
TileDB-SOMA version: 0.0.5a9
Language and language version (e.g. Python 3.8, R 4.2.2): Py 3.10.6, but shouldn't matter
OS (e.g. MacOS, Ubuntu Linux): MacOS, of course

**Additional context**

Test code results look like:
```
test_case 0 FAILED: "Experiment has no item 'df'", read_timestamp=1676138942753, write_timestamp=100
test_case 1 PASSED, read_timestamp=18446744073709551615, write_timestamp=None
test_case 2 FAILED: "Experiment has no item 'df'", read_timestamp=18446744073709551615, write_timestamp=100
```
Sometimes they all pass, but 0 and 2 fails transiently, sometimes with only one of them failing. 😵‍💫 

See reproduction test case, above, to interpret.

I am seeing that some of the TileDB groups and metadata files do _not_ have the specified explicit `write_timestamp`.
```
❯ find /tmp/pit/ -regex '.*/__[0-9]+_.*'
# Test case 0, failure: The expt/df/__schema timestamp is NOT set to the specified write_timestamp
/tmp/pit/0/__meta/__100_100_2424680daedc4bb6839a5f6507335d89
/tmp/pit/0/expt/__meta/__100_100_3621c4e641b14a42950df4cb4e9b8696
/tmp/pit/0/expt/df/__meta/__100_100_40fd7a800ca041d3a4bb82d3d3c43bec
/tmp/pit/0/expt/df/__schema/__1676140033065_1676140033065_cb4dd07d67bf4635a6140c082e200b27
/tmp/pit/0/expt/causes_bug/__meta/__100_100_09451752238549b696401ea472a452d2
/tmp/pit/0/expt/__group/__100_100_82aab386278a47a6bcf328f8eb097e86
[...]

PR #981, opened 2023-02-21
[python] Use the same timestamp across read and write operations

thetorpedodog:
This eliminates the distinction between the timestamps used to read and write TileDB data. This ensures that:

- All data is written with the same timestamp.
- When one context is used to write and then read the same data, the same timestamp is used for both operations.
- A consistent view of a TileDB object is used for metadata when opening a handle for writing.

---

Given that all our tests pass, and this eliminates the "auxiliary read at time now" hack, I think this might be the right thing to do for this case.

### DIFF
--- a/apis/python/src/tiledbsoma/_tdb_handles.py
+++ b/apis/python/src/tiledbsoma/_tdb_handles.py
@@ -166,7 +166,7 @@ class ArrayWrapper(Wrapper[tiledb.Array]):
         return tiledb.open(
             uri,
             mode,
-            timestamp=(context.timestamp_start, context.timestamp),
+            timestamp=context._timestamp_arg(),
             ctx=context.tiledb_ctx,
         )
 
--- a/apis/python/src/tiledbsoma/_tdb_handles.py
+++ b/apis/python/src/tiledbsoma/_tdb_handles.py
@@ -24,7 +24,7 @@ from somacore import options
 from typing_extensions import Literal, Self
 
 from ._exception import DoesNotExistError, SOMAError, is_does_not_exist_error
-from .options import SOMATileDBContext
+from .options._soma_tiledb_context import SOMATileDBContext
 
 RawHandle = Union[tiledb.Array, tiledb.Group]
 _RawHdl_co = TypeVar("_RawHdl_co", bound=RawHandle, covariant=True)
@@ -67,12 +67,7 @@ class Wrapper(Generic[_RawHdl_co], metaclass=abc.ABCMeta):
             tdb = cls._opener(uri, mode, context)
             handle = cls(uri, mode, context, tdb)
             if mode == "w":
-                # Briefly open a read-mode handle to populate metadata/schema/group contents/etc.,
-                # ignoring any read_timestamp set in the context to get an up-to-date view in
-                # preparation for writing.
-                with cls._opener(
-                    uri, "r", context, use_latest_read_timestamp=True
-                ) as auxiliary_reader:
+                with cls._opener(uri, "r", context) as auxiliary_reader:
                     handle._do_initial_reads(auxiliary_reader)
             else:
                 handle._do_initial_reads(tdb)
@@ -89,7 +84,6 @@ class Wrapper(Generic[_RawHdl_co], metaclass=abc.ABCMeta):
         uri: str,
         mode: options.OpenMode,
         context: SOMATileDBContext,
-        use_latest_read_timestamp: bool = False,
     ) -> _RawHdl_co:
         """Opens and returns a TileDB object specific to this type."""
         raise NotImplementedError()
@@ -168,19 +162,13 @@ class ArrayWrapper(Wrapper[tiledb.Array]):
         uri: str,
         mode: options.OpenMode,
         context: SOMATileDBContext,
-        use_latest_read_timestamp: bool = False,
     ) -> tiledb.Array:
-        if not use_latest_read_timestamp:
-            timestamp_arg = (
-                context.write_timestamp
-                if mode == "w"
-                else (context.read_timestamp_start, context.read_timestamp)
-            )
-        else:
-            # array opened in write mode should initialize with latest metadata
-            assert mode == "r"
-            timestamp_arg = None
-        return tiledb.open(uri, mode, timestamp=timestamp_arg, ctx=context.tiledb_ctx)
+        return tiledb.open(
+            uri,
+            mode,
+            timestamp=(context.timestamp_start, context.timestamp),
+            ctx=context.tiledb_ctx,
+        )
 
     @property
     def schema(self) -> tiledb.ArraySchema:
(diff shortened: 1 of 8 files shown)


apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -209,21 +197,8 @@ def _opener(
...
-        if not use_latest_read_timestamp:
-            # As of Feb 2023, tiledb.Group() has no timestamp arg; instead its timestamps must be
-            # set in the tiledb.Ctx config. SOMATileDBContext prepares the suitable tiledb.Ctx.
-            ctx_arg = (
-                context._group_write_tiledb_ctx
-                if mode == "w"
-                else context._group_read_tiledb_ctx
-            )
-        else:
-            # Group opened in write mode should initialize with latest contents & metadata
```
mlin, line 223: I think this is kind of a thorny question, if for whatever reason we've opened the object for writing at a *past* timestamp (preceding other, existing, later-timestamped writes), what view of the contents/metadata should we start with? But that's a weird enough thing to be doing in the first place that I'm not too worried about it. Simplifying the code is good.

mlin (approved):
Subjective opinions written in slack. No objections if this is the direction we want to go. Note this is a late, breaking change to 1.0rc0 for @atolopko-czi @bkmartinjr

thetorpedodog:
preserving some of the stuff I have said in chat for posterity:

The model I have in my head, though, is that when interacting with a dataset, you are interacting with TileDB at *one specific timestamp*. If you want to read at a different timestamp, that belongs in a different context. (This would be even more salient with the proposed reopen API.)

Moreover, based on the following events:

- 10:00: Create collection `a`. This sets the write timestamp of all changes we do to `a`.
- 10:02: `a.add_new_dataframe('b', ...)`. Write lots of data to `b`.
- 10:04: `a.add_new_dataframe('c', ...)`. Write lots more data to `c`.
- 10:06: Close everything.

With the existing model, the timestamp of a write is determined by when the open occurs. That means that the write to `a`, which includes adding `b` and `c` as sub-elements, occurs at 10:00. However, the writes to `b` and `c` only occur at 10:02 and :04 (respectively), meaning that, if after all writes complete:

- If you open `a` as of 10:01, `b` and `c` will both appear as members of the group, but *will not contain any data* (in fact, the metadata will not be set, which I am now thinking “write metadata with the timestamp of object creation” is probably the right thing to do).
- If you open `a` as of 10:03, only `b` will contain data.
- Only if you open `a` as of 10:04 or later, will the full object be there.

With this new model, all of `a`, `b`, and `c` will be created and written as of the same timestamp (10:00, by default, when the context was created).

> I think this is kind of a thorny question, if for whatever reason we've opened the object for writing at a past timestamp (preceding other, existing, later-timestamped writes), what view of the contents/metadata should we start with?

My strong opinion is that it should always be as of the only specified timestamp, so that you get a temporally coherent view no matter what. If the user wants to act based on the “current” metadata, they can open the array/group/whatever as of Now (or The Infinite Future, as the case may be).

I’m going to leave this open until sometime tomorrow so that more discussion can happen if need be.

mlin:
Then also for the record, my opinion is that setting the same timestamp across a group of writes is currently not advisable (especially as a default behavior) due to the randomization behavior exhibited in #919. We *think* we know how to avoid that (TBD -- it [may not yet be fixed](https://github.com/chanzuckerberg/cell-census/pull/210)), but it's a deep and wide pitfall, which we ourselves fell into without delay. Again, I defer to you as long as this has been considered and deemed an acceptable risk (hopefully a temporary one).

thetorpedodog:
What I don’t entirely understand is what circumstances are leading to randomized writes—are you opening, writing, and closing the same SOMA object multiple times over the course of a single execution? Because if you’re doing multiple writes on the same `open`, the behavior here will be unchanged on a per–SOMA object basis, since all write calls on the same `open` are recorded at the same timestamp anyway.

The other problem I was aiming to solve here is that there is an asymmetry between the way read timestamps and write timestamps are handled. For instance, with the previous behavior:

1. T+0:00: Create a new context with default options.
2. T+0:01: Use the context to write a bunch of data. Close.
3. T+0:02: Use that same context to *read* the data you just wrote.

When you read the data, it will not actually see the data you just wrote, because the read timestamp stored in the context is T+0:00, before the (automatic) timestamp used to write the data at T+0:01.

My feeling is that if we’re writing at the current timestamp, we should also be *reading* at the current timestamp, so the behavior is consistent on both sides. This can be achieved by allowing the user to opt out of timestamp consistency by passing `None` as a timestamp. (Alternately, we could make it opt-in, where `None` is the default but a user can pass in the current timestamp at context creation.)

thetorpedodog:
Issuing a correction on my previous statements regarding write timing: The timestamps of writes are recorded as of the **close** call, not as of the open. This means that some of the worse symptoms described above do not occur, but the asymmetry between read and write would remain.

mlin:
My caution is an acknowledgment that I don't think I fully understand all the circumstances when it may be hazardous to set the same timestamp across my logical group of writes -- it's already surprised me a couple of times. Since we're demonstrably in the vicinity of the worst possible kind of bug -- nondeterministic, silent, and undetected data corruption -- my instinct is to just avoid the whole area, if possible, rather than attempt to navigate it safely.

On the other side of the scale, I understand we want to share the write timestamps mainly so that there will exist no timestamp when a reader would see some but not all of our writes. That's a virtuous property, which however I don't think necessary for SOMA & Cell Census. It should suffice that we can publish a known-good timestamp when the reader will see all of our writes (and nothing written after). I think we can simply let the timestamps default to "now" as we actually apply each write, per TileDB default behavior, and report when we finished.

Therefore I wouldn't set any shared write timestamp by default given our experience with the current version of TileDB. That implies I think the write timestamp is distinct from the read timestamp, where a shared default seems reasonable for some snapshot consistency (but always better to set a known-good one explicitly). 

This isn't to say the current implementation doesn't have its own sharp edges that could be improved, which you've pointed out, but I feel your full model for read/write temporal coherence across a group of objects would be more safely adopted by TileDB-SOMA at a future date, when it's fully supported by TileDB itself, without even the slightest proximity to a data corruption hazard area.

Finally, to belabor it once again, this is all my personal opinion; I'm not suggesting I'm right, and I'm fine with agreeing to disagree on any of it, with you choosing what to do next.

thetorpedodog:
Getting the opt-out tested ended up taking a little more time than I anticipated (mostly, the part where I was learning more about how timestamps work that led to my last comment), so I wasn’t able to get it out yesterday.

*For now* I would still prefer to make consistency opt-out, but we still have the option to make it opt-in if we need to before release.

mlin (approved):
With aforementioned discomfort -- agreeing to disagree

thetorpedodog:
After some discussion with Bruce, I’m going to submit this with timestamp consistency being the default (so that by doing nothing you get consistency). I would like to discuss at the big meeting tomorrow whether that *should* be the default, and if we want to go to “no timestamp by default” it will be a one-line change, just so that the rest of the implementation isn’t blocked on the default decision.

PR #985, opened 2023-02-22
[python] Handle timestamp index types for `DataFrame`

johnkerl:
**Issue and/or context:**

Issue #960. Follow-on PR to #968.

### DIFF
--- a/libtiledbsoma/src/pyapi/libtiledbsoma.cc
+++ b/libtiledbsoma/src/pyapi/libtiledbsoma.cc
@@ -530,6 +530,171 @@ PYBIND11_MODULE(libtiledbsoma, m) {
                 const std::string&, const std::vector<uint8_t>&)>(
                 &SOMAReader::set_dim_points))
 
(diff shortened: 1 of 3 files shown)


libtiledbsoma/src/pyapi/libtiledbsoma.cc
```diff
@@ -512,6 +505,20 @@ PYBIND11_MODULE(libtiledbsoma, m) {
                             reader.set_dim_point(dim, value);
                         }
 
+                    } else if (
+                        !strcmp(arrow_schema.format, "tss:") ||
+                        !strcmp(arrow_schema.format, "tsms:") ||
+                        !strcmp(arrow_schema.format, "tsus:") ||
+                        !strcmp(arrow_schema.format, "tsns:")) {
+                        tcb::span<int64_t> data{
+                            (int64_t*)arrow_array.buffers[1],
```
gspowley, line 514: Can we apply the index approach from #986?

gspowley (approved):

Issue #988, opened 2023-02-22
[Bug] [python] `nnz` raises `RuntimeError` when `allows_duplicates=True`


bkmartinjr:
**Describe the bug**

SOMAReader.nnz raises an error when the underlying array has `allow_duplicates=True`.  It should not do so.

Example:
```
In [10]: A = tiledb.open('file:///home/bruce/cell-census/tmp/census/2023-02-22/soma/census_data/homo_sapiens/obs')

In [11]: A.schema
Out[11]: 
ArraySchema(
  domain=Domain(*[
    Dim(name='soma_joinid', domain=(0, 9223372036854775806), tile=2048, dtype='int64', filters=FilterList([DoubleDeltaFilter(), ZstdFilter(level=19), ])),
  ]),
  attrs=[
    Attr(name='dataset_id', dtype='<U0', var=True, nullable=False, filters=FilterList([DictionaryFilter(), ZstdFilter(level=19), ])),
    Attr(name='assay', dtype='<U0', var=True, nullable=False, filters=FilterList([DictionaryFilter(), ZstdFilter(level=19), ])),
    Attr(name='assay_ontology_term_id', dtype='<U0', var=True, nullable=False, filters=FilterList([DictionaryFilter(), ZstdFilter(level=19), ])),
    Attr(name='cell_type', dtype='<U0', var=True, nullable=False, filters=FilterList([DictionaryFilter(), ZstdFilter(level=19), ])),
    Attr(name='cell_type_ontology_term_id', dtype='<U0', var=True, nullable=False, filters=FilterList([DictionaryFilter(), ZstdFilter(level=19), ])),
    Attr(name='development_stage', dtype='<U0', var=True, nullable=False, filters=FilterList([DictionaryFilter(), ZstdFilter(level=19), ])),
    Attr(name='development_stage_ontology_term_id', dtype='<U0', var=True, nullable=False, filters=FilterList([DictionaryFilter(), ZstdFilter(level=19), ])),
    Attr(name='disease', dtype='<U0', var=True, nullable=False, filters=FilterList([DictionaryFilter(), ZstdFilter(level=19), ])),
    Attr(name='disease_ontology_term_id', dtype='<U0', var=True, nullable=False, filters=FilterList([DictionaryFilter(), ZstdFilter(level=19), ])),
    Attr(name='donor_id', dtype='<U0', var=True, nullable=False, filters=FilterList([DictionaryFilter(), ZstdFilter(level=19), ])),
[...]

PR #989, opened 2023-02-23
[c++] Support `nnz` when `allow_duplicates=True`

johnkerl:
**Issue and/or context:**

#988 

**Changes:**

We've established the pattern that:

* `allow_duplicates=False` is the default
* `allow_duplicates=True` is supported with user caveats that uniqueness is the application's responsibility

Here we honor that pattern.

### DIFF
--- a/libtiledbsoma/src/soma_reader.cc
+++ b/libtiledbsoma/src/soma_reader.cc
@@ -213,7 +213,7 @@ uint64_t SOMAReader::nnz() {
 
         // If any relevant fragment is a consolidated fragment, fall back to
         // counting cells, because the fragment may contain duplicates.
-        if (frag_ts.first != frag_ts.second) {
+        if (!mq_->schema()->allows_dups() && frag_ts.first != frag_ts.second) {
             return nnz_slow();
         }
     }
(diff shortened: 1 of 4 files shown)


libtiledbsoma/src/soma_reader.cc
```diff
@@ -274,32 +274,34 @@ uint64_t SOMAReader::nnz() {
...
-    // If duplicates are allowed, we cannot count simply count cells.
     if (mq_->schema()->allows_dups()) {
-        throw TileDBSOMAError(
-            "[SOMAReader] nnz not supported when duplicates are allowed");
+        LOG_DEBUG(
+            "[SOMAReader] computing nnz() when duplicates are allowed: this is "
+            "accurate only when the application has otherwise ensured "
+            "uniqueness");
     }
+}
```
gspowley, line 283: Extra `{`?

gspowley (approved):
Thanks @johnkerl, for making it clear we are honoring the caveat that uniqueness is the application's responsibility.

bkmartinjr:
@johnkerl - see my comments on #988. I'm still not sure this is right, and would appreciate a chance to work through the consistency issue

bkmartinjr:
Related to my comments in #988, and this code:
```
        // If any relevant fragment is a consolidated fragment, fall back to
        // counting cells, because the fragment may contain duplicates.
        if (frag_ts.first != frag_ts.second) {
            return nnz_slow();
        }
```
I believe that when `allows_duplicates=True`, the call to nnz_slow is going to do the same thing as if we drop through to code below (i.e., do not call `nnz_slow`).  Rationale:  the code in nnz_slow does not de-dup in this corner case.

If you buy my argument in #988, the fix is to not bother calling `nnz_slow` in this case when `allows_duplicates==True`. Suggested modification is something like:
```
        // If any relevant fragment is a consolidated fragment, fall back to
        // counting cells, because the fragment may contain duplicates.
        if (!mq_->schema()->allows_dups() && frag_ts.first != frag_ts.second) {
            return nnz_slow();
        }
```

johnkerl:
> Suggested modification is something like:

@bkmartinjr yes indeed -- @mlin and I have been thinking the same thing -- I'll do this

libtiledbsoma/src/soma_reader.cc
johnkerl: @gspowley yeah I had a bad merge/rebase. Fixed on commit #2.

johnkerl:
@bkmartinjr I'm ready to merge this on @gspowley's approval -- double-checking this works for you -- if not please let me know if you have any suggestions for how to improve the C++ and/or Python unit-test cases to remove any misalignment between the unit-test cases' articulated expectations and yours -- thank you!

PR #993, opened 2023-02-24
Don't do R CI jobs on pre-commit changes.

thetorpedodog:
I was puzzled as to why, whenever we updated dependencies, the R build workflow would run despite R code not being changed. Turns out it was because the pre-commit configuration (which lives outside of the ignored Python directory) was triggering it to run.

Since R currently does not use pre-commit, we can safely ignore it, meaning that Python dependency changes will no longer trigger mysterious R workflows.


johnkerl (approved):
✨

PR #994, opened 2023-02-24
ccache test

thetorpedodog:


.github/workflows/r-ci.yml
```diff
@@ -35,7 +39,11 @@ jobs:
         uses: lukka/get-cmake@latest
 
       - name: MkVars
-        run: mkdir ~/.R && echo "CXX17FLAGS=-Wno-deprecated-declarations -Wno-deprecated" > ~/.R/Makevars
+        run: |
+          set -e
+          mkdir ~/.R
+          echo "CXX17FLAGS=-Wno-deprecated-declarations -Wno-deprecated" > ~/.R/Makevars
+          echo "CCACHE=ccache" > ~/.R/Makevars
```
eddelbuettel, line 46: It needs a bit more because of how R lays out files. I [once wrote about that](https://dirk.eddelbuettel.com/blog/2017/11/27/) and still copy the snippet for `~/.ccache/ccache.conf` 
around:

```
max_size = 5.0G
# important for R CMD INSTALL *.tar.gz as tarballs are expanded freshly -> fresh ctime
sloppiness = include_file_ctime
# also important as the (temp.) directory name will differ
hash_dir = false

eddelbuettel: Also your second echo wants a `>>`

eddelbuettel (commented):
See the inline comments re `ccache` with R packages

The bigger issue is that the cache _usually_ does not persist.  I only use `ccache` in one of the many CI yaml files for my packages because only one is really expensive (so much so that I hit timeouts at Travis back in the day).

thetorpedodog:
thanks for the info! this was very much a quick experiment based on [something some guy wrote on Stack Overflow](https://stackoverflow.com/a/45512708/39808). i was hoping that maybe a fairly simple change could get caching benefits similar to those we saw on the python side, but as you point out it’s more complex than that. because this was just a shot in the dark, and doing it Right™ is little beyond my current expertise, i’ll probably let this go.

eddelbuettel:
Hah!  Guilty as charged. I just added the snippet there too.  Drove me nuts when I saw lightning fast `ccache` driven compilations _from source directories with fixed paths and timestamps_ but not when testing (for which one is strongly encouraged to create *.tar.gz which then get updated below `$TMP`.

PR #998, opened 2023-02-24
[python Make the default timestamp in `SOMATileDBContext` be `None`

thetorpedodog:
Storing timestamps in the Context is problematic since it is likely that users will want to share many parts of the Context across multiple open calls.  This would result in every object opened on that shared context having the same timestamp, and not just the child tree of a single open call from user code.  This is already an issue with Cell Census, which uses a single shared context for the entire duration of the program.

Because of this, it's better to have the timestamp be unset rather than having it set to context creation.


mlin (approved):
+1 to putting this change on main. Separately I'll digest today's illuminating discussions and reform opinions on what to do in 1.0. Thanks!!

PR #1001, opened 2023-02-24
[python] Allow all operations on a collection to share a timestamp

thetorpedodog:
This change is intended as one cohesive unit, but its three parts can be separated from each other if not wanted:

1. Adding a `tiledb_timestamp` parameter to `open`s to allow us to open entire collections in a timestamp-coherent manner.
2. Allowing users to pass that `tiledb_timestamp` as a Python datetime.
3. Making the `tiledb_timestamp` member on TileDB objects into a Python datetime (while the raw msec value is still available at `tiledb_timestamp_ms`.

This also removes the `timestamp_start` parameter, which seems like an advanced feature on top of the already advanced feature that is timestamps.

The overall goal is to provide the “expected” behavior, where opening a collection of objects acts like opening one object, and when you don’t specify a timestamp, things just work and you don’t have to think about it.

This is a follow-up for issues discovered in #919.

### DIFF
--- a/apis/python/src/tiledbsoma/_util.py
+++ b/apis/python/src/tiledbsoma/_util.py
@@ -252,7 +252,7 @@ def ms_to_datetime(millis: int) -> datetime.datetime:
     """Returns the millisecond timestamp as a timezone-aware UTC datetime.
 
     This may raise an exception, since millis may be outside the representable
-    range for
+    range for a Python datetime.
     """
     secs, millis = divmod(millis, 1000)
     dt = datetime.datetime.fromtimestamp(secs, tz=datetime.timezone.utc)
--- a/apis/python/src/tiledbsoma/_util.py
+++ b/apis/python/src/tiledbsoma/_util.py
@@ -1,3 +1,4 @@
+import datetime
 import pathlib
 import time
 import urllib.parse
@@ -7,7 +8,7 @@ from typing import Any, Optional, Tuple, Type, TypeVar
 import somacore
 from somacore import options
 
-from ._types import Slice, is_slice_of
+from ._types import OpenTimestamp, Slice, is_slice_of
 
 
 def get_start_stamp() -> float:
@@ -224,3 +225,24 @@ def check_unpartitioned(partitions: Optional[options.ReadPartitions]) -> None:
     if not partitions or partitions == options.IOfN(0, 1):
         return
     raise ValueError("Paritioned reads are not currently supported")
+
+
+_ETERNITY_MS = 2**64 - 1
+
+
+def to_timestamp_ms(input: OpenTimestamp) -> int:
+    """Converts a timestamp input type to Unix millis."""
+    check_type("tiledb_timestamp", input, (int, datetime.datetime))
+    if isinstance(input, int):
+        timestamp_ms = input
+    else:
+        # Manually pull out the milliseconds so that nothing funny happens
+        # like 12:00:00.300 turning into 299 ms.
+        milli_part = input.microsecond // 1000
+        input = input.replace(microsecond=0)
+        seconds_part = int(input.timestamp())
+        timestamp_ms = seconds_part * 1000 + milli_part
+
+    if not 0 <= timestamp_ms <= _ETERNITY_MS:
+        raise ValueError("open timestamp must be between 0 (Unix epoch) and 2**64-1 ms")
+    return timestamp_ms
--- a/apis/python/src/tiledbsoma/_util.py
+++ b/apis/python/src/tiledbsoma/_util.py
@@ -246,3 +246,14 @@ def to_timestamp_ms(input: OpenTimestamp) -> int:
     if not 0 <= timestamp_ms <= _ETERNITY_MS:
         raise ValueError("open timestamp must be between 0 (Unix epoch) and 2**64-1 ms")
     return timestamp_ms
+
+
+def ms_to_datetime(millis: int) -> datetime.datetime:
+    """Returns the millisecond timestamp as a timezone-aware UTC datetime.
+
+    This may raise an exception, since millis may be outside the representable
+    range for
+    """
+    secs, millis = divmod(millis, 1000)
+    dt = datetime.datetime.fromtimestamp(secs, tz=datetime.timezone.utc)
+    return dt.replace(microsecond=millis * 1000)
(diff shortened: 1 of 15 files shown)


apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -31,15 +41,18 @@ class SOMATileDBContext:
 
     tiledb_ctx: tiledb.Ctx = _build_default_tiledb_ctx()
 
-    timestamp: Optional[int] = attrs.field(default=None)
+    timestamp_ms: Optional[int] = attrs.field(
+        default=None, converter=_maybe_timestamp_ms
+    )
     """
-    Timestamp for operations on SOMA objects, in milliseconds since Unix epoch.
+    Default timestamp for operations on SOMA objects, in Unix millis.
```
bkmartinjr, line 48: Not sure what "Unix millis" indicates.  If actually `milliseconds since the Python (aka Unix) epoch`, it may be clearer to fully state that. That would match the Python `time.time()` documentation.

bkmartinjr (approved):
Looks good to me.  One optional incredibly small nit noted.

thetorpedodog:
> Not sure what "Unix millis" indicates. If actually milliseconds since the Python (aka Unix) epoch, it may be clearer to fully state that. That would match the Python time.time() documentation.

I will be sure to take care of that in an upcoming rev. In the meantime, I’m leaving this open so others can give feedback since this does have a bit more impact than a simple bugfix.

johnkerl (approved):
I think it's safe to update the comment text as requested by @bkmartinjr

atolopko-czi:
If I understand correctly, the new tiledb_timestamp parameter is intended for overriding the common case default of "now". If overriding is the less common case, can't that also be done through the context object? Do we need the new parameter or is it just a convenience?

thetorpedodog:
> If I understand correctly, the new tiledb_timestamp parameter is intended for overriding the common case default of "now". If overriding is the less common case, can't that also be done through the context object? Do we need the new parameter or is it just a convenience?

The reason we pulled the timestamp out of the context is described in https://github.com/single-cell-data/TileDB-SOMA/pull/998. In short, in cases where you want to use the same context many times, storing the timestamp there is bad because it means multiple open/write/close cycles will have no defined order.

mlin (approved):
Hi all, the new approach SGTM given our new understanding that the specific behavior @atolopko-czi isolated in #919 was not consistent with the intended design where shared timestamps are supposed to merge more benignly (at least for setting distinct group keys & similar straightforward cases). That reduces the risk where previously I'd been quite averse to even being in the same zipcode!

The default read timestamp is a clear win. For write timestamps I still note that the group temporal coherence model @thetorpedodog has developed for SOMACollection seems well out in front of TileDB itself, at least as documented. The [docs page on write timestamps](https://docs.tiledb.com/main/how-to/arrays/writing-arrays/writing-at-a-timestamp) merely says "There are cases when it is very useful" to set them, without elaborating. I'd have slightly preferred to leave all write-mode timestamps unset for now and let the model bake more fully inside TileDB itself. I hope that will happen but I think we're good to :ship: now!

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -31,15 +42,18 @@ class SOMATileDBContext:
...
-    ``None``, the default, does not provide cross-object timestamp consistency:
+    This is used when a timestamp is not provided to an ``open`` operation.
 
-    - **Reads** will see data as of the time that the individual TileDB object
-      is opened.
-    - **Writes** will be recorded as of when the TileDB object is closed.
+    ``None``, the default, sets the timestamp on each root ``open`` operation.
+    That is, if you ``open`` a collection, and access individual members of the
+    collection through indexing or ``add_new``, the timestamp of all of those
+    operations will be that of the time you called ``open``.
```
atolopko-czi, line 56: Per your response, "in cases where you want to use the same context many times, storing the timestamp there is bad because it means multiple open/write/close cycles will have no defined order.", do we want to warn against setting explicit timestamps in `SOMATileDBContext`? Or is that too edge-casey?

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -49,54 +63,11 @@ class SOMATileDBContext:
...
-                timestamp_start=self.timestamp_start,
-                timestamp=self.timestamp,
-            )
-        object.__setattr__(self, "_group_tiledb_ctx_", ctx)
-        return ctx
-
-    def _timestamp_arg(self) -> Optional[Tuple[int, int]]:
-        """The value to use as the ``timestamp`` argument to ``tiledb.open``."""
-        if self.timestamp is None:
+    def timestamp(self) -> Optional[datetime.datetime]:
```
atolopko-czi, line 67: Is this method necessary or just a convenience for users?

atolopko-czi (approved):
LGTM, with a few minor comments.

apis/python/src/tiledbsoma/_util.py
```diff
@@ -224,3 +225,35 @@ def check_unpartitioned(partitions: Optional[options.ReadPartitions]) -> None:
...
+    if not 0 <= timestamp_ms <= _ETERNITY_MS:
+        raise ValueError("open timestamp must be between 0 (Unix epoch) and 2**64-1 ms")
+    return timestamp_ms
+
+
+def ms_to_datetime(millis: int) -> datetime.datetime:
+    """Returns the millisecond timestamp as a timezone-aware UTC datetime.
+
+    This may raise an exception, since millis may be outside the representable
+    range for
```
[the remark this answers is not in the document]
thetorpedodog: thank

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
thetorpedodog: It is not strictly necessary. It's here to parallel the structure of `timestamp`/`timestamp_ms` in `TileDBObject`s.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
thetorpedodog: This is a great point. I’ve added a warning here, recommending that users not do this. I would personally be OK with removing the timestamp from the context entirely, but didn’t want to go that far on this change.

PR #1024, opened 2023-02-28
[python] Pin stub versions used in type checking

thetorpedodog:
`pandas-stubs` released a new version on 2023-02-27, which caused builds to break despite no changes.  We should fix the new issues, but in the short term, just pin the stubs to the then-current version.

(See the build breakage from the last rev of #1001.)


johnkerl (approved):
🚢

Issue #997, opened 2023-02-24
[docs] Audit the developer wiki


johnkerl:
Check for broken links, etc. at https://github.com/single-cell-data/TileDB-SOMA/wiki.

johnkerl:
Updated -- no broken links remain.

To do: some how-to information for sandbox development.

johnkerl:
The only remaining work item is #1032 which is an open issue tagged `1.0`. Closing this issue.

PR #1025, opened 2023-02-28
Update C++ sandbox-build instructions

johnkerl:
Context: #997

### DIFF
--- a/libtiledbsoma/README.md
+++ b/libtiledbsoma/README.md
@@ -103,15 +103,24 @@ Examples:
 
 ## Notes
 
-A build system with `libfmt` and `libspdlog` installed may conflict with the required versions for `tiledbsoma`. Suggestion:
+A build system with `libfmt`, `libspdlog`, or `tiledb` installed may conflict with the required
+versions for `tiledbsoma`. Suggestion is for you to not install these, and to let this project's
+superbuild install compatiable versions for you.
 
-* Uninstall `libfmt-dev` and `libspdlog-dev`
-* Check the following is gone (and manually remove if necessary) `/usr/lib/*/cmake/spdlog/spdlogConfig.cmake`
+Specifically, uninstall the `libfmt-dev` and `libspdlog-dev` packages (or perhaps named `spdlog` and `fmt`, depending
+on the OS). As well, uninstall the `tiledb` package if you have that separately installed.
 
-Similarly, for MacOS:
+As a pro-tip: check the following is gone (and manually remove if necessary) `/usr/lib/*/cmake/spdlog/spdlogConfig.cmake`
 
-* `brew uninstall spdlog` (if you have it)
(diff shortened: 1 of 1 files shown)


libtiledbsoma/README.md
```diff
@@ -103,21 +103,10 @@ Examples:
 
 ## Notes
 
-A build system with `libfmt` and `libspdlog` installed may conflict with the required versions for `tiledbsoma`. If that is the case, this configuration is known-good on Ubuntu 22.04:
+A build system with `libfmt` and `libspdlog` installed may conflict with the required versions for `tiledbsoma`. Suggestion:
```
eddelbuettel, line 106: It's unclear language and it is not entirely unclear what our recommendation should be. I for one have _always_ had system `libfmt*` and `libspdlog*` and I have not been bitten.  Should be maybe just remove the sentence here as well as the suggestion?  At TileDB Core, the recommendation always has been (paraphrasing) "if in doubt, use a superbuild".

eddelbuettel (commented):
I added a comment.  It's a complicated topic, across all operating systems and versions and hardware variants :face_with_head_bandage:

libtiledbsoma/README.md
johnkerl: @eddelbuettel you've had those libraries separately installed and it's worked for you. So we should retain the _option_ of having them -- as well as the known-good logic.

@gspowley has had those libraries _not_ separately installed and it's worked for him. So we should retain that option.

@johnkerl (yours truly) tried it both your way and George's way and neither one worked for me -- in part due to the dangling spdlog CMake file left around after uninstall. That caused untold amounts of frustration and failure to resolve the issue, even with multiple smart people involved. Civic duty compels me to include the hard-learned lesson.

I'll update the wording to reflect both use-cases.

eddelbuettel: I am trying to point at the bigger picture: you already delineated two to three distinct cases at just one Linux distro. Are we sure this generalizes? Are we sure it covers older distros? What about Conda, Brew, ...?  For as long as we cannot say anything _definitive_ I lean towards saying little.

johnkerl: @eddelbuettel when I can be definitive, I will; as a second choice, I lean toward helpfulness and sharing known issues / hard-won lessons.

eddelbuettel: I hear you and I appreciate the intent to be maximally helpful. But recall how what was once an off-hand comment by me on what worked here stuck around 'forever' (til you cleaned) in prominent position.  I think I learned over the years that saying less may be better than saying something that may end up being misleading to some, but it is far from clear cut.  And .... this was just a comment and not an objection.  Full steam ahead.

johnkerl:
Updated wording @nguyenv @gspowley @eddelbuettel

gspowley (approved):

Issue #961, opened 2023-02-17
[python] Explicit shaping of `DataFrame`


johnkerl:
Counterpart to #931 which addresses explicit shaping of `SparseNDArray`.

In particular, to be addressed:

* https://github.com/single-cell-data/TileDB-SOMA/pull/952#discussion_r1108911306

PR #1027, opened 2023-02-28
[python] Explicit shaping for `DataFrame`

johnkerl:
**Issue and/or context:** #961 

Requires as a prerequisite: https://github.com/single-cell-data/SOMA/pull/143

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -556,15 +556,14 @@ def _build_tiledb_schema(
             )
 
     dims = []
-    for i, index_column_name in enumerate(index_column_names):
+    for index_column_name, domain in zip(index_column_names, domains):
         pa_type = schema.field(index_column_name).type
         dtype = _arrow_types.tiledb_type_from_arrow_type(
             pa_type, is_indexed_column=True
         )
 
-        domain: Tuple[Any, Any]
-        if domains[i] is not None:
-            domain = domains[i]  # type: ignore[assignment]
+        if domain is not None:
+            pass  # User-specified; go with it
 
         elif isinstance(dtype, str):
             domain = None, None
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -98,15 +98,16 @@ class DataFrame(TileDBArray, somacore.DataFrame):
             All named columns must exist in the schema, and at least one
             index column name is required.
 
-        :param domains: An optional sequence of [XXX TO DO] positive int64 values specifying the
-            shape of each index column, including room for any intended future
-            appends. If provided, this must have the same length as
+        :param domains: An optional sequence of min-max tuples specifying the
+            domain of each index column, including room for any intended future
+            appends.  If provided, this sequence must have the same length as
             `index_column_names`, and the index-column domains will be as specified.
-            If omitted, the index-column domains will use the maximum possible int64
-            value.  This makes a `SOMADataFrame` growable.
+            If omitted entirely, or if ``None`` in a given dimension, the
+            corresponding index-column domain will use the maximum possible values.
+            This makes a `SOMADataFrame` growable.
 
-        :param platform_config: Platform-specific options used to create this DataFrame,
-            provided via ``{"tiledb": {"create": ...}}`` nested keys.
+        :param platform_config: Platform-specific options used to create this
+            DataFrame, provided via ``{"tiledb": {"create": ...}}`` nested keys.
 
         :param tiledb_timestamp: If specified, overrides the default timestamp
             used to open this object. If unset, uses the timestamp provided by
@@ -543,7 +544,16 @@ def _build_tiledb_schema(
     context: SOMATileDBContext,
 ) -> tiledb.ArraySchema:
     """Converts an Arrow schema into a TileDB ArraySchema for creation."""
-    # XXX check domains is None, or, length matches index_column_names
+
+    if domains is None:
+        domains = tuple([None for _ in index_column_names])
+    else:
+        ndom = len(domains)
+        nidx = len(index_column_names)
+        if ndom != nidx:
+            raise ValueError(
+                f"if domains are specified, they must have the same length as index_column_names; got {ndom} != {nidx}"
+            )
 
     dims = []
     for i, index_column_name in enumerate(index_column_names):
@@ -551,9 +561,10 @@ def _build_tiledb_schema(
         dtype = _arrow_types.tiledb_type_from_arrow_type(
             pa_type, is_indexed_column=True
         )
+
         domain: Tuple[Any, Any]
-        if domains is not None and domains[i] is not None:
-            domain = domains[i]
+        if domains[i] is not None:
+            domain = domains[i]  # type: ignore[assignment]
 
         elif isinstance(dtype, str):
             domain = None, None
@@ -582,10 +593,9 @@ def _build_tiledb_schema(
         else:
             raise TypeError(f"Unsupported dtype {dtype}")
 
-        # Default 2048 mods to 0 for 8-bit types and 0 is an invalid extent
-        extent = tiledb_create_options.dim_tile(index_column_name)
-        if isinstance(dtype, np.dtype) and dtype.itemsize == 1:
-            extent = 64
+        extent = _find_extent_for_domain(
+            index_column_name, tiledb_create_options, dtype, domain
+        )
 
         dim = tiledb.Dim(
             name=index_column_name,
@@ -636,3 +646,29 @@ def _build_tiledb_schema(
         tile_order=tile_order,
         ctx=context.tiledb_ctx,
     )
+
+
+def _find_extent_for_domain(
+    index_column_name: str,
+    tiledb_create_options: TileDBCreateOptions,
+    dtype: Any,
+    domain: Tuple[Any, Any],
+) -> Any:
+    """Helper function for _build_tiledb_schema."""
+
+    # Default 2048 mods to 0 for 8-bit types and 0 is an invalid extent
+    extent = tiledb_create_options.dim_tile(index_column_name)
+    if isinstance(dtype, np.dtype) and dtype.itemsize == 1:
+        extent = 64
+
+    if isinstance(dtype, str):
+        return extent
+
+    if not (np.issubdtype(dtype, NPInteger) or np.issubdtype(dtype, NPFloating)):
+        return extent
+
+    lo, hi = domain
+    if lo is None or hi is None:
+        return extent
+
+    return min(extent, hi - lo + 1)
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -98,13 +98,16 @@ class DataFrame(TileDBArray, somacore.DataFrame):
             All named columns must exist in the schema, and at least one
             index column name is required.
 
-        :param domains: An optional sequence of min-max tuples specifying the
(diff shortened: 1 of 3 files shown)


apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -512,22 +522,40 @@ def _canonicalize_schema(
...
+        ndom = len(domains)
+        nidx = len(index_column_names)
+        if ndom != nidx:
+            raise ValueError(
+                f"if domains are specified, they must have the same length as index_column_names; got {ndom} != {nidx}"
+            )
+
     dims = []
-    for index_column_name in index_column_names:
+    for i, index_column_name in enumerate(index_column_names):
```
thetorpedodog, line 542: You might zip these together:

```python
for index_column_name, domain in zip(index_column_names, domains):
```

apis/python/tests/test_dataframe_index_columns.py
```diff
@@ -54,1354 +55,2037 @@ def arrow_table():
...
         schema=arrow_table.schema,
         index_column_names=index_column_names,
+        domains=domains,
     )
 
+    if "write" in errors:
+        with pytest.raises(errors["write"]):
+            with soma.DataFrame.open(uri, "w") as sdf:
+                sdf.write(arrow_table)
+        return
```
thetorpedodog, line 2061: It might be easier to follow the tests if the good cases are tested separately from each type of bad case, so there aren’t a bunch of parameters going unused in every test case.

johnkerl: @thetorpedodog what does "tested separately" mean to you? I don't want to have completely separate `test_` functions if that's what you mean -- that would result in too much duplicated code.

thetorpedodog: I actually am suggesting having different `test_types`/`test_types_create_fail`/`test_types_write_fail`/`test_types_read_fail` functions. Given that the existing test function already duplicates the creation/writing/reading code for the good and bad cases, there’s very little additional boilerplate, and the absolute line count would be significantly lower. Each test would be focused on one individual behavior rather than one big test which tries to handle all cases.

I am generally of the opinion that a little duplication of test code is preferable to having a test that behaves significantly differently depending upon the input.

johnkerl: @thetorpedodog https://github.com/single-cell-data/TileDB-SOMA/pull/1027/commits/54c7f04730a380e5c433746e5d7a98a11961a2a9

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -97,8 +98,16 @@ def create(
             All named columns must exist in the schema, and at least one
             index column name is required.
 
-        :param platform_config: Platform-specific options used to create this DataFrame,
-            provided via ``{"tiledb": {"create": ...}}`` nested keys.
+        :param domains: An optional sequence of min-max tuples specifying the
```
bkmartinjr, line 101: is the "min-max tuple" inclusive or exclusive of the values specified?  I.e., open, closed or half-open ranges?

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -97,8 +98,16 @@ def create(
...
+            domain of each index column, including room for any intended future
+            appends.  If provided, this sequence must have the same length as
+            `index_column_names`, and the index-column domains will be as specified.
+            If omitted entirely, or if ``None`` in a given dimension, the
+            corresponding index-column domain will use the maximum possible values.
```
bkmartinjr, line 106: max possible values - what does this mean?  Is it "all possible values for the column type"?  (I believe this is the intent)

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -78,6 +78,7 @@ def create(
         *,
         schema: pa.Schema,
         index_column_names: Sequence[str] = (SOMA_JOINID,),
+        domains: Optional[Sequence[Optional[Tuple[Any, Any]]]] = None,
```
bkmartinjr, line 81: given how fundamental this is to the operation of the DataFrame, suggest adding a paragraph detailing the model in the class docstring.

apis/python/src/tiledbsoma/_dataframe.py
johnkerl: updated

apis/python/src/tiledbsoma/_dataframe.py
johnkerl: updated

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -517,22 +527,39 @@ def _canonicalize_schema(
...
-    for index_column_name in index_column_names:
+    for index_column_name, domain in zip(index_column_names, domains):
         pa_type = schema.field(index_column_name).type
         dtype = _arrow_types.tiledb_type_from_arrow_type(
             pa_type, is_indexed_column=True
         )
-        domain: Tuple[Any, Any]
-        if isinstance(dtype, str):
+
+        if domain is not None:
```
bkmartinjr, line 556: I'm not sure the code path correctly handles `bool` typed index columns.  I believe they are unsupported as in index column, but a (superficial) quick skim looks like it will not generate an error.

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -97,8 +98,16 @@ def create(
             All named columns must exist in the schema, and at least one
             index column name is required.
 
-        :param platform_config: Platform-specific options used to create this DataFrame,
-            provided via ``{"tiledb": {"create": ...}}`` nested keys.
+        :param domains: An optional sequence of min-max tuples specifying the
+            domain of each index column, including room for any intended future
+            appends.  If provided, this sequence must have the same length as
```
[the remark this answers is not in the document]
johnkerl: updated

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -517,22 +530,39 @@ def _canonicalize_schema(
...
+        if domain is not None:
+            pass  # User-specified; go with it
+
+        elif isinstance(dtype, str):
             domain = None, None
         elif np.issubdtype(dtype, NPInteger):
             iinfo = np.iinfo(cast(NPInteger, dtype))
             domain = iinfo.min, iinfo.max - 1
+            if index_column_name == SOMA_JOINID:
+                domain = (0, domain[1])
```
bkmartinjr, line 565: I'm not sure this approach is consistent with how we handle schema checks for soma_joinid.

In the schema, if the user specifies the type, it must be correct or we throw an error.  We don't silently correct it.

In this case, if the domain is wrong, the code silently corrects the domain.  I don't think there is a strong case one way or the other, but perhaps we should be consistent.  I.e., check that a user-specified domain is correct, and then simply raise an error if it is not.

optional

apis/python/tests/test_dataframe_index_columns.py
```diff
@@ -54,18 +55,27 @@ def arrow_table():
 
 
 @pytest.mark.parametrize(
-    "name,index_column_names,coords,expecteds",
+    "name,index_column_names,domains,coords,expecteds",
```
bkmartinjr, line 58: I believe this is missing a test for the case where the domain of a `soma_joinid` indexed dataframe has a user-specified domain that is illegal.

I.e., something like `index_column_names=['soma_joinid']` and `domains=[(-1000,-1)]`

apis/python/src/tiledbsoma/_dataframe.py
johnkerl: @bkmartinjr this isn't correcting a user error.

This block only applies when the user has _not_ specified the domain. Here we _choose_ one, and we specifically disallow negative `soma_joinid` which is in conformance with the spec.

https://github.com/single-cell-data/SOMA/pull/124
https://github.com/single-cell-data/SOMA/issues/121

bkmartinjr: You are absolutely correct - apologies.

That does raise a possible follow-up: do we error check that the domain for `soma_joinid` is bounded by [0, maxint)?

johnkerl: @bkmartinjr good call & i'll resolve this comment in favor of that one

apis/python/tests/test_dataframe_index_columns.py
johnkerl: committed

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -517,52 +539,100 @@ def _canonicalize_schema(
...
+    for index_column_name, slot_domain in zip(index_column_names, domain):
         pa_type = schema.field(index_column_name).type
         dtype = _arrow_types.tiledb_type_from_arrow_type(
             pa_type, is_indexed_column=True
         )
-        domain: Tuple[Any, Any]
-        if isinstance(dtype, str):
-            domain = None, None
+
+        if slot_domain is not None:
```
thetorpedodog, line 565: It might be worth pulling this whole bit where you figure out the real domain into its own function.

thetorpedodog (approved):
This looks good to me.

thetorpedodog:
> Requires as a prerequisite: [single-cell-data/SOMA#143](https://github.com/single-cell-data/SOMA/pull/143)

As a note, in situations where functionality is being added to the implementation in a backwards-compatible way, adding it to the base spec is not, strictly speaking, a prereq. I would use the approach of getting things done and landed in the implementation first, to nail down the interface, and then go back and put the interface we landed on in the spec, just so that we’re not juggling things across two separate changes at the same time.

(That said, if you prefer separating the discussion of “this is what the interface should look like” versus “this is how the implementation does it”, that’s also fine.)

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -133,6 +147,14 @@ def index_column_names(self) -> Tuple[str, ...]:
         """
         return self._tiledb_dim_names()
 
+    @property
+    def domain(self) -> Tuple[Tuple[Any, Any], ...]:
+        """
```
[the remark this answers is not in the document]
thetorpedodog: Is a separate "returns" section needed when the text of the docstring describes the returned value?

apis/python/tests/test_dataframe_index_columns.py
```diff
@@ -1346,69 +1592,308 @@ def arrow_table():
...
+        actual_table = sdf.read(coords=coords).concat()
+        for query_column_name, expected_array in expecteds.items():
+            actual_array = actual_table[query_column_name].combine_chunks()
+            # The output from the first one is easier to read when it fails
+            assert actual_array.to_pylist() == expected_array.to_pylist()
+            assert actual_array == expected_array
+        if domain is not None:
+            actual_domain = sdf.domain
+            for i in range(len(domain)):
+                if (
```
bkmartinjr, line 1653: can we add a test that verifies what `domain` returns when the index column type is `string` or `bytes`?

bkmartinjr (approved):
Couple of small requests around docstrings and test coverage.  No concerns with the change itself.

apis/python/src/tiledbsoma/_dataframe.py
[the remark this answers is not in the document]
bkmartinjr: good call - the style guide says about the Returns docstring section:

> It may also be omitted if the docstring starts with Returns or Yields (e.g. """Returns row from Bigtable as a tuple of strings.""") and the opening sentence is sufficient to describe the return value.


[Reference](https://google.github.io/styleguide/pyguide.html#doc-function-returns)

PR #1037, opened 2023-03-01
[r] Revert #1026

johnkerl:
Reverts #1026 as we are seeing segfaults in CI


aaronwolen (approved):
Noting we will still see CI failures until https://github.com/single-cell-data/TileDB-SOMA/pull/1033 is merged but it should _not_ segfault.

PR #1038, opened 2023-03-01
[r] Implement `PlatformConfig` and `SOMATileDBContext` for R

johnkerl:
This is almost entirely @mojaveazure 's work -- a re-post of #1026 which we reverted after we started seeing segfaults in `main` CI. The changes here are:

* Make the `ctx` passed to `tiledb_group_create` be long-lived rather than ephemeral
* Some name-changes along the way to more clearly delineate which things are tiledbsoma-level context objects and which are tiledb-level context objects


.gitignore
```diff
@@ -65,6 +65,7 @@ apis/r/src/libtiledbsoma/*
 apis/r/src/Makevars
 apis/r/src/cmake_log.txt
 apis/r/.editorconfig
+!apis/r/man/roxygen/*
```
eddelbuettel, line 68: What does this do? I have never seen a leading `!` in `.gitignore`, nor have I seen man/roxygen/* files.

apis/r/R/roxygen.R
```diff
@@ -0,0 +1,17 @@
+
+#' @export
```
eddelbuettel, line 2: Is this a user-facing and exported class, or a roxygen helper construct?

apis/r/R/roxygen.R
```diff
@@ -0,0 +1,17 @@
...
+#'
+rd_return_virtual <- function() {
+  return("This is a \\strong{virtual} class and cannot be directly instantiated")
+}
+
+#' @export
+#'
+rd_atomic <- function() {
```
eddelbuettel, line 10: See previous comment

apis/r/R/utils.R
```diff
@@ -1,3 +1,62 @@
...
+}
+
+#' Check if a vector is named
+#' @noRd
+is_named <- function(x) {
+  !is.null(names(x))
+}
+
+is_named2 <- function(x) {
+  return(!is.null(x = names(x = x)) && all(nzchar(x = names(x = x))))
```
eddelbuettel, line 17: Why a return here and not in the other ones?  Can is_named and is_named2 be collapsed or do we need two?

apis/r/R/roxygen.R
mojaveazure: Not necessarily user-facing, but exported so that it can be used in `\Sexpr`

`R CMD check` might (should) throw a warning about it, so I will end up writing docs for it marking it with `@keywords internal`

eddelbuettel (commented):
This is a fairly large and comprehensive PR, and I don't think I have had a chance to catch up on any design specs for all this or see what Python does so I would need a bit more time.

That said it all looks fine at a first glance.  I added some initial questions / comments from a first pass -- mostly just curious questions.

apis/r/R/utils.R
mojaveazure: `return` was a stylistic choice that I wrote out of habit. I did not write `is_named` and can remove the `return` (and named parameters)

`is_named2` is more ~complete~ strict than `is_named`; I don't know who wrote `is_named` and what their requirements were, whereas I desired the strictness that comes from also checking with `nzchar()`

```R
is_named(c("" = 1)) # TRUE
is_named2(c("" = 1)) # FALSE
```

eddelbuettel: Hm. Maybe `git blame` can help, and keeping it to one would be nice.  Else maybe use a more evocative name that simply suffixing with '2'?

apis/r/R/roxygen.R
mojaveazure: I was worried that R CMD check would complain about `:::` being used in `\Sexpr` but I just checked and saw that [{generics} gets away with it](https://github.com/r-lib/generics/blob/main/R/accuracy.R) so I'll remove the exports

eddelbuettel: _Within the package it is being used_ the `:::` is allowed (and I would say recommended).  It's a very good (it less well understood idiom.  I just used it today (on a branch of the same repo).

The only downside is when you copy and paste code to debug, ie outside the package namespace.  But hiding and clean top-level namespaces are worth it so maybe git it a go!

mojaveazure: I wasn't aware if that included dynamically-generated docs (where `\Sexpr` is being used), but I'm glad to know that it does!

eddelbuettel: I cannot speak to that. But I can assure you code within package `foo` does not need `foo:::myfunc` and can access `myfunc`.  I would think related / generated docs in the same package behave the same way but neither know nor asserted they do.  I would try it.

apis/r/R/utils.R
aaronwolen: Totally agree, @eddelbuettel.

mojaveazure: I've merged `is_named` and `is_named2` by adding a new parameter to enable the checking that `is_named2` provides. By default, it follows `is_named` behavior (although running tests using `is_named2` behavior still passed)

apis/r/R/ConfigList.R
```diff
@@ -0,0 +1,91 @@
...
+    # },
+    # set_from_list = function(key, value) {
+    #   stopifnot(is_named_list(x = value))
+    #   cfg <- ScalarMap$new()
+    #   for (i in names(x = value)) {
+    #     cfg$set(key = i, value = value[[i]])
+    #   }
+    #   self$add_map(key = key, value = cfg)
+    #   return(invisible(x = self))
+    # }
```
eddelbuettel, line 89: Core prefers to remove commented-out code in PRs prior to merge.  Just noting it here -- I don't think we must remove this, but we could.

apis/r/R/MappingBase.R
```diff
@@ -0,0 +1,234 @@
...
+      #     '  }',
+      #     paste0('  private$.data[[', sQuote(x = key, q = FALSE), ']] <- value'),
+      #     '  return(invisible(x = NULL))',
+      #     '}',
+      #     sep = '\n'
+      #   )
+      #   f <- eval(expr = str2expression(text = f), envir = self$.__enclos_env__)
+      #   makeActiveBinding(sym = key, fun = f, env = self$.__enclos_env__)
+      #   # self$.__enclos_env__$.__active__[[key]] <- f
+      # }
```
eddelbuettel, line 101: See above re comments.

apis/r/R/PlatformConfig.R
```diff
@@ -0,0 +1,130 @@
...
+    #' and \code{value} as a \code{\link{ScalarMap}} to update \code{op} for
+    #' \code{platform} with the keys and values from \code{value}
+    #'
+    #' @return \[chainable\] Invisibly returns \code{self} with \code{value}
+    #' added for \code{key} in \code{op} for \code{platform}
+    #'
+    set = function(platform, op, key, value) {
+      stopifnot(
+        is.character(x = platform) && length(x = platform) == 1L,
+        is.character(x = op) && length(x = op) == 1L
```
eddelbuettel, line 114: You could use the form  `stopifnot("text to be displayed if error" = test1 && test2)` which is nicer on the user.  Even if it is stricly an internal `assert()` style test it still provides comments.

apis/r/R/SOMATileDBContext.R
```diff
@@ -0,0 +1,145 @@
...
+        !length(x = config) || is_named(x = config, allow_empty = FALSE)
+      )
+      config['sm.mem.reader.sparse_global_order.ratio_array_data'] <- '0.3'
+      # Add the TileDB context
+      # TODO: Reenable this with newer version of tiledb-r
+      # if (!length(x = config)) {
+      #   config <- NA_character_
+      # }
+      # cfg <- tiledb::tiledb_config(config = config)
+      # TODO: replace this when patched version of tiledb-r is released
```
eddelbuettel, line 44: (May need a versioned Imports: then just to be safe the bug-fixed version is used.)

apis/r/DESCRIPTION
```diff
@@ -43,7 +43,8 @@ Imports:
     arch (>= 0.0.2),
     dplyr,
     data.table,
-    spdl
+    spdl,
+    rlang
```
eddelbuettel, line 47: This is missing the matching use in `NAMESPACE`:

```r
N  checking dependencies in R code (1.7s)
   Namespace in Imports field not imported from: ‘rlang’
     All declared Imports should be used.
```

eddelbuettel: Fixed in fb87cd2

eddelbuettel (approved):
This is a large and comprehensive PR.  It looks good, I will need some more time with the new classes to be become more familiar with the use but nice work and congrats to @mojaveazure .

There are a number of nags from `R CMD check` some of which (on _e.g._ the Rd references) are addressed in my open PR #1052 so we can overlook them here.  But it would be good always keep PRs `R CMD check --as-cran` clean, not merely at 'not breaking CI'.

:shipit:

apis/r/R/ConfigList.R
johnkerl: +1

apis/r/R/MappingBase.R
```diff
@@ -0,0 +1,234 @@
...
+        .NotYetImplemented()
+      }
+      private$.data <- list()
+    },
+    #' @return The keys of the map
+    #'
+    keys = function() {
+      return(names(x = private$.data))
+    },
+    #' @return The values of the map
```
aaronwolen, line 38:
```suggestion
    #' @return A `list` containing the map values
```

apis/r/R/SOMATileDBContext.R
```diff
@@ -0,0 +1,146 @@
+#' SOMA TileDB Context
+#'
+#' Context map for TileDB-backed SOMA objects
+#'
+#' @export
+#'
+#' @noMd
```
aaronwolen, line 7: Why?

apis/r/R/SOMATileDBContext.R
```diff
@@ -0,0 +1,146 @@
...
+  classname = 'SOMATileDBContext',
+  inherit = SOMAContextBase,
+  public = list(
+    #' @template param-config
+    #' @param cached Force new creation
+    #'
+    #' @return An instantiated \code{SOMATileDBContext} object
+    #'
+    initialize = function(config = NULL, cached = TRUE) {
+      # super$initialize()
```
aaronwolen, line 19: Commented code should be removed

aaronwolen (changes requested):
This is great. I love how its architected. I left a few comments in the code that mostly revolve around encouraging consistency with existing soma-r code/paradigms.

High-level comments:

- I'm on the fence as to whether we need the `ScalarMap` class or not but I'm not opposed to it.
- I'd like to see some tests in which `PlatformConfig` is used to create an array or a `SOMAExperiment`. You could potentially modify `create_and_populate_experiment()` to accept a `platform_config` arg for this purpose.
- Does `MappingBase` need a `remove()` method?
- Not sure we're getting enough value from rlang to add it as a dep

Minor stylistic nits:

- For `PlatformConfig` I'd like to substitute `op`/`ops` for either `option`/`options` or `param`/`params` (I prefer the latter).
- For readability I think we should drop argument names for single argument functions (e.g., `length(x)` instead of `length(x = x)`)

Other:

- Let's remove the html rendered docs from this PR
- Looks like there was a rebase issue at some point, see [comment below](https://github.com/single-cell-data/TileDB-SOMA/pull/1038#discussion_r1126889939)

aaronwolen (changes requested):
This is great. I love how its architected. I left a few comments in the code that mostly revolve around encouraing consistency with existing soma-r code/paradigms.

High-level comments:

- I'm on the fence as to whether we need the `ScalarMap` class or not but I'm not opposed to it.
- I'd like to see some tests in which `PlatformConfig` is used to create an array or a `SOMAExperiment`. You could potentially modify `create_and_populate_experiment()` to accept a `platform_config` arg for this purpose.
- Does `MappingBase` need a `remove()` method?
- Not sure we're getting enough value from rlang to add it as a dep

Minor stylistic nits:

- For `PlatformConfig` I'd like to substitute `op`/`ops` for either `option`/`options` or `param`/`params` (I prefer the latter).
- For readability I think we should drop argument names for single argument functions (e.g., `length(x)` instead of `length(x = x)`)

Other:

- Let's remove the html rendered docs from this PR

apis/r/R/MappingBase.R
```diff
@@ -0,0 +1,234 @@
...
+    #'
+    #' @return The value of \code{key} in the map, or \code{default} if
+    #' \code{key} is not found
+    #'
+    get = function(key, default = rlang::missing_arg()) {
+      key <- key[1L]
+      key <- tryCatch(
+        expr = match.arg(arg = key, choices = self$keys()),
+        error = \(...) NULL
+      )
```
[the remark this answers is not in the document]
mojaveazure: Implemented in e35142e

apis/r/R/MappingBase.R
```diff
@@ -0,0 +1,234 @@
...
+    },
+    #' @param key Key to fetch
+    #' @templateVar key key
+    #' @templateVar default NULL
+    #' @template param-default
+    #'
+    #' @return The value of \code{key} in the map, or \code{default} if
+    #' \code{key} is not found
+    #'
+    get = function(key, default = rlang::missing_arg()) {
```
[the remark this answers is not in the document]
mojaveazure: I've replaced the default with `quote(expr = )` to simulate `missing_arg()`; this way errors are still thrown by default and the user can pass `NULL` to return `NULL` when `key` is not found

apis/r/R/PlatformConfig.R
```diff
@@ -0,0 +1,130 @@
...
+      platform,
+      op = NULL,
+      key = NULL,
+      default = rlang::missing_arg()
+    ) {
+      if (!length(x = self)) {
+        warning("No platforms configured", call. = FALSE)
+        return(NULL)
+      }
+      platform <- platform[1L] %||% self$platforms()[1L]
```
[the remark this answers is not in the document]
mojaveazure: Implemented in e35142e

apis/r/R/PlatformConfig.R
```diff
@@ -0,0 +1,130 @@
...
+    #'
+    platforms = function() {
+      return(self$keys())
+    },
+    #' @param platform The \dQuote{platform} to pull option names (middle keys)
+    #' for; pass \code{TRUE} to return all possible option names
+    #'
+    #' @return The option names (middle keys) for \code{platform}
+    #'
+    ops = function(platform = NULL) {
```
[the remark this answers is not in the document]
mojaveazure: Changed to `param`/`params` in e35142e

apis/r/DESCRIPTION
mojaveazure: Dependency on rlang removed in e35142e

apis/r/R/MappingBase.R
```diff
@@ -0,0 +1,234 @@
...
+      prd <- ifelse(
+        test = tolower(x = substr(x = cls, start = 1L, stop = 2L)) %in% vowels,
+        yes = 'An',
+        no = 'A'
+      )
+      lng <- self$length()
+      ent <- ifelse(test = lng == 1L, yes = 'entry', no = 'entries')
+      cat(prd, cls, "map with", lng, ent, sep = ' ')
+      return(invisible(x = self))
+    }
```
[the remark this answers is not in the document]
mojaveazure: Implemented in 9d82635

eddelbuettel (approved):
Looks good.  And thanks for removing docs/ from this round.

It is still too long and large which makes reviewing difficult.  In hindsight, more and shorter PRs may have been easier to deal with.  "Next time."

aaronwolen (approved):
Great work, @mojaveazure!

Issue #1035, opened 2023-03-01
[Bug][python] `exists` raises exception in cases when it should return False


bkmartinjr:
**Describe the bug**

`tiledbsoma.*.exists()` is documented as return a False for any URI that does not exist.  However, in many cases it will raise an exception instead of returning False.

Simple example: an unknown URI scheme, e.g, `tiledbsoma.DataFrame.exists("foo://bar")`

The typical pattern in Python (e.g., `fsspec`) is to swallow all exceptions and return False, e.g.,
```
In [26]: fsspec.AbstractFileSystem().exists("foo")
Out[26]: False

In [27]: fsspec.AbstractFileSystem().exists("foo://bar")
Out[27]: False
```

Root cause is the explicit `except SOMAError` catch in the `exists` implementation.

**Versions (please complete the following information):**
```
tiledbsoma.__version__    1.0.0rc2
tiledb.__version__        0.20.0
core version              2.14.0
python version            3.10.6.final.0
```

PR #1040, opened 2023-03-03
[python] Return False on various `.exists()`

johnkerl:
**Issue and/or context:** #1035

### DIFF
--- a/apis/python/src/tiledbsoma/_tiledb_object.py
+++ b/apis/python/src/tiledbsoma/_tiledb_object.py
@@ -164,9 +164,23 @@ class TileDBObject(somacore.SOMAObject, Generic[_WrapperType_co]):
         context: Optional[SOMATileDBContext] = None,
         tiledb_timestamp: Optional[OpenTimestamp] = None,
     ) -> bool:
-        """Finds whether an object of this type exists at the given URI.
-        [lifecycle: experimental].
         """
+        Finds whether an object of this type exists at the given URI.
+        [lifecycle: experimental].
+
+        :param uri: The URI to open.
+
+        :param context: If provided, the ``SOMATileDBContext`` to use when creating and
+            attempting to access this object.
+
+        :param tiledb_timestamp: The TileDB timestamp to open this object at,
+            measured in milliseconds since the Unix epoch.
+            When unset (the default), the current time is used.
+
+        :raises TypeError: if the ``uri`` is not a string.
+        """
+        if not isinstance(uri, str):
+            raise TypeError(f"uri argument must be of type string; got {type(uri)}")
         context = context or SOMATileDBContext()
         try:
             with cls._wrapper_type.open(uri, "r", context, tiledb_timestamp) as hdl:
@@ -174,9 +188,9 @@ class TileDBObject(somacore.SOMAObject, Generic[_WrapperType_co]):
                 if not isinstance(md_type, str):
                     return False
                 return md_type.lower() == cls.soma_type.lower()
-        except SOMAError:
+        except SOMAError:  # type mismatch
             return False
-        except tiledb.cc.TileDBError:
+        except tiledb.cc.TileDBError:  # no-such on storage
             return False
 
     @classmethod
--- a/apis/python/src/tiledbsoma/_tiledb_object.py
+++ b/apis/python/src/tiledbsoma/_tiledb_object.py
@@ -10,7 +10,7 @@ from typing_extensions import Self
 from . import _constants, _tdb_handles
 from ._exception import SOMAError
 from ._types import OpenTimestamp
-from ._util import ms_to_datetime
+from ._util import check_type, ms_to_datetime
 from .options import SOMATileDBContext
 
 _WrapperType_co = TypeVar(
@@ -179,8 +179,7 @@ class TileDBObject(somacore.SOMAObject, Generic[_WrapperType_co]):
 
         :raises TypeError: if the ``uri`` is not a string.
         """
-        if not isinstance(uri, str):
-            raise TypeError(f"uri argument must be of type string; got {type(uri)}")
+        check_type("uri", uri, (str,))
         context = context or SOMATileDBContext()
         try:
             with cls._wrapper_type.open(uri, "r", context, tiledb_timestamp) as hdl:
--- a/apis/python/src/tiledbsoma/_tiledb_object.py
+++ b/apis/python/src/tiledbsoma/_tiledb_object.py
@@ -176,6 +176,8 @@ class TileDBObject(somacore.SOMAObject, Generic[_WrapperType_co]):
                 return md_type.lower() == cls.soma_type.lower()
         except SOMAError:
             return False
+        except tiledb.cc.TileDBError:
+            return False
 
     @classmethod
     def _set_create_metadata(cls, handle: _tdb_handles.AnyWrapper) -> None:
--- a/apis/python/src/tiledbsoma/_tiledb_object.py
+++ b/apis/python/src/tiledbsoma/_tiledb_object.py
@@ -188,9 +188,7 @@ class TileDBObject(somacore.SOMAObject, Generic[_WrapperType_co]):
                 if not isinstance(md_type, str):
                     return False
                 return md_type.lower() == cls.soma_type.lower()
-        except SOMAError:  # type mismatch
-            return False
-        except tiledb.cc.TileDBError:  # no-such on storage
+        except (SOMAError, tiledb.cc.TileDBError):
             return False
 
     @classmethod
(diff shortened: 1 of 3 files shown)


atolopko-czi (changes requested):

apis/python/tests/test_dataframe.py
```diff
@@ -991,3 +991,10 @@ def test_timestamped_ops(tmp_path, allows_duplicates, consolidate):
         assert list(x.as_py() for x in tab["string"]) == ["apple"]
         assert sidf.tiledb_timestamp_ms == 1615402887987
         assert sidf.tiledb_timestamp.isoformat() == "2021-03-10T19:01:27.987000+00:00"
+
+
+def test_exists():
+    # Specifically, we're checking this doesn't raise
+    assert not soma.DataFrame.exists("/nonesuch/no/nope/never/ever")
+    assert not soma.DataFrame.exists("foo://bar")
+    assert not soma.DataFrame.exists("s3://@@@@ILLEGAL@@@@")
```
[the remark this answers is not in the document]
johnkerl: @atolopko-czi I'll make `tests/test_tiledboject.py` and add negative cases for all the foundational and composite types.

I won't add positive cases since those already have _very_ thorough coverage in existing tests already, and I'll add a comment to that effect.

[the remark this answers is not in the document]
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/1040/commits/1f9430cc56ae7d2ee00f2dfa130eb30dd21fecba

apis/python/src/tiledbsoma/_tiledb_object.py
```diff
@@ -176,6 +176,8 @@ def exists(
                 return md_type.lower() == cls.soma_type.lower()
         except SOMAError:
             return False
+        except tiledb.cc.TileDBError:
```
bkmartinjr, line 179: why restrict this to a TileDBError?

I suggest a naked `except`, otherwise we will have the same bug when some other error is raised.

(this is what fsspec and other packages do)

apis/python/tests/test_tiledbobject.py
```diff
@@ -0,0 +1,42 @@
+import tiledbsoma as soma
```
bkmartinjr, line 1: this is fantastic!

can you also add some non-string arguments?  bytes, dict, number, etc.

bkmartinjr (changes requested):
The catch(es) in the implementation should be consolidated into a single naked except statement, to catch _any_ exception.

apis/python/tests/test_tiledbobject.py
```diff
@@ -0,0 +1,42 @@
...
+# Checking that objects _do_ exist is already done (thoroughly) in other
+# tests. Here we check the negative cases.
+#
+# Specifically, we're checking these don't raise, and all return False.
+
+
+def test_dataframe_exists():
+    assert not soma.DataFrame.exists("/nonesuch/no/nope/never/ever")
+    assert not soma.DataFrame.exists("foo://bar")
+    assert not soma.DataFrame.exists("s3://@@@@ILLEGAL@@@@")
```
atolopko-czi, line 12: At the risk of belaboring the unit test aspect of this PR, all of these tests could be shortened with `@pytest.mark.parametrize`

johnkerl:
> The catch(es) in the implementation should be consolidated into a single naked except statement, to catch any 

@bkmartinjr this is an antipattern -- it would catch things like `NameError` for misspellings, etc.

It's correct to catch only the exceptions appropriate to whether the collection exists or not.

atolopko-czi (approved):
LGTM, with one more optional suggestion for refactoring tests.

apis/python/tests/test_tiledbobject.py
johnkerl: @bkmartinjr the `uri` argument is of type `str`.

`soma.DenseNDArray.exists(123.45)` should, if anything, raise `ValueError` since 123.45 is not a valid URI in any sense.

bkmartinjr:
> @bkmartinjr this is an antipattern -- it would catch things like NameError for misspellings, etc.

please provide an example of type of error you think it should raise and the use case in which that is the right solution.  Given how it is defined, I think you are wrong (and there are heavily used packages that share this "anti-pattern").

My main point here is that you aren't catching all errors that should return a False.  which is why fsspec and others use a catch-all.

apis/python/src/tiledbsoma/_tiledb_object.py
johnkerl: > The catch(es) in the implementation should be consolidated into a single naked except statement, to catch any 

@bkmartinjr this is an antipattern -- it would catch things like `NameError` for misspellings, etc.

It's correct to catch only the exceptions appropriate to whether the collection exists or not.

If this is what `fsspec` does then we're propagating an antipattern. I won't die on this hill and am willing to disagree and commit and implement a catch-everything. But I don't (and won't) like it.

bkmartinjr: lolz, no death necessary on _either_ side of the argument.  Let @thetorpedodog  decide :-)

The current implementation catches most of the cases I care about.

johnkerl: https://realpython.com/the-most-diabolical-python-antipattern/

apis/python/tests/test_tiledbobject.py
johnkerl: done

johnkerl:
@bkmartinjr `soma.DataFrame.exists(123.45)` is a type error (and should be). Anyone who thinks 123.45 is a valid URI needs to know that.

apis/python/src/tiledbsoma/_tiledb_object.py
bkmartinjr: No disagreement on the theoreticals, but please:
* catch all reasonable exceptions (i.e., OSError, etc) that _shoudl_ return false
* document what it raises so the signature is clear

johnkerl:
Changes made in accordance with off-line sync: https://github.com/single-cell-data/TileDB-SOMA/pull/1040/commits/ee4fcd76291563de49b95c422e15fef98a616d9e

apis/python/tests/test_tiledbobject.py
```diff
@@ -0,0 +1,45 @@
...
+# tests. Here we check the negative cases.
+#
+# Specifically, we're checking these don't raise, and all return False.
+
+
+@pytest.mark.parametrize(
+    "uri",
+    ["/nonesuch/no/nope/never/ever", "foo://bar", "s3://@@@@ILLEGAL@@@@"],
+)
+@pytest.mark.parametrize(
```
bkmartinjr, line 16: the other negative test not covered is the check for a validate SOMA object of the expected type.

Ie.., soma.DataFrame.exists checks for the right TileDB metadata that matches the expected class type.

johnkerl: good call @bkmartinjr , thanks for the code-review feedback

johnkerl:
Thanks @bkmartinjr ! Cross-type tests coming up :)

apis/python/src/tiledbsoma/_tiledb_object.py
```diff
@@ -164,17 +164,33 @@ def exists(
...
         try:
             with cls._wrapper_type.open(uri, "r", context, tiledb_timestamp) as hdl:
                 md_type = hdl.metadata.get(_constants.SOMA_OBJECT_TYPE_METADATA_KEY)
                 if not isinstance(md_type, str):
                     return False
                 return md_type.lower() == cls.soma_type.lower()
-        except SOMAError:
+        except SOMAError:  # type mismatch
+            return False
+        except tiledb.cc.TileDBError:  # no-such on storage
```
thetorpedodog, line 193:
```suggestion
        except (SOMAError, tiledb.cc.TileDBError):
```

apis/python/src/tiledbsoma/_tiledb_object.py
```diff
@@ -164,17 +164,33 @@ def exists(
...
+            attempting to access this object.
+
+        :param tiledb_timestamp: The TileDB timestamp to open this object at,
+            measured in milliseconds since the Unix epoch.
+            When unset (the default), the current time is used.
+
+        :raises TypeError: if the ``uri`` is not a string.
         """
+        if not isinstance(uri, str):
+            raise TypeError(f"uri argument must be of type string; got {type(uri)}")
```
thetorpedodog, line 183: `_util.check_type`?

thetorpedodog (approved):
couple tiny little things; neither mandatory

Issue #1041, opened 2023-03-03
[docs] Doc-infra updates in support of 1.0 doc tasks


johnkerl:
This is mostly orthogonal to other doc tasks for 1.0, which mainly have to do with doc _content_. The work here is on _presentation and rendering of that content_.

On this issue:

* Immediate term: rendering in `sphinx` (Python) and `pkgdown` (R)
  * Let folks know the steps for sandbox previews
    * https://github.com/single-cell-data/TileDB-SOMA/wiki/Doc-work-and-sandbox-previews
* Have a web location to publish this to
* Make the publishing happen in CI

On a subsequent PR (issue #1055):

* Unified rendering using Quarto

johnkerl:
These are live and linked to from our [README.md](https://github.com/single-cell-data/TileDB-SOMA/blob/main/README.md):

* [Python API](https://tiledbsoma.readthedocs.io/en/latest/python-api.html)
* [R API](https://single-cell-data.github.io/TileDB-SOMA)
* [Example Python notebooks](https://github.com/single-cell-data/TileDB-SOMA/tree/main/apis/python/notebooks)

PR #1042, opened 2023-03-03
[python] Trivial doc-neatens in prep for web publishing

johnkerl:
**Issue and/or context:** #1041

**Changes:** Trivial mods for things that look OK in Python on-line help but are in fact revealed to be sub-optimal when previewed in sandbox readthedocs preview.

**Information for the reviewer:** Here's how to preview this kind of work in a sandbox environment:
https://github.com/single-cell-data/TileDB-SOMA/wiki/Doc-work-and-sandbox-previews

### DIFF
--- a/apis/python/src/tiledbsoma/__init__.py
+++ b/apis/python/src/tiledbsoma/__init__.py
@@ -6,8 +6,8 @@ motivated by use cases from single-cell biology. The ``tiledbsoma``
 Python package is an implementation of SOMA using the
 [TileDB Embedded](https://github.com/TileDB-Inc/TileDB) engine.
 
-Provides
----------
+Provides:
+----------
   1. The ability to store, query, and retrieve larger-than-core datasets,
      resident in both cloud (object-store) and local (file) systems.
   2. A data model supporting dataframes, and both sparse and dense
@@ -17,8 +17,9 @@ Provides
 See the [SOMA GitHub repo](https://github.com/single-cell-data/SOMA) for more
 information on the SOMA project.
 
-Using the documentation
-------------------------
+Using the documentation:
+-------------------------
+
 Coming soon: web-based documentation site.
 
 Documentation is also available via the Python builtin ``help`` function. We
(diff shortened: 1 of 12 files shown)


apis/python/src/tiledbsoma/__init__.py
```diff
@@ -6,7 +6,7 @@
 Python package is an implementation of SOMA using the
 [TileDB Embedded](https://github.com/TileDB-Inc/TileDB) engine.
 
-Provides:
+Provides
```
johnkerl, line 9: For on-line Python CLI help, I think with or without trailing `:` is OK

But for readthedocs output (see also [here](https://github.com/single-cell-data/TileDB-SOMA/wiki/Doc-work-and-sandbox-previews)) these show up as subsection titles wherein the `:` looks odd

Hence my removing the trailing `:`

doc/source/python-api.rst
```diff
@@ -18,86 +18,17 @@ Features:
 * Enables out-of-core access to data aggregations much larger than single-host main memory
 * Enables distributed computation over datasets
 
-Modules
```
johnkerl, line 21: This is an iterative improvement but not entirely right. @ebezzi points out this is not inherting methods from `somacore`. This is not a final product. I can omit the changes on this particular from the current PR if that helps it get accepted/merged sooner.

apis/python/src/tiledbsoma/__init__.py
bkmartinjr: Google style guide implies (by example) that the colon is to be included in section headers, and does not use underline.

[Section 3.8](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings)

I have no personal opinion other than consistency is goodness

apis/python/src/tiledbsoma/__init__.py
```diff
@@ -17,51 +17,51 @@
...
 
 >>> import tiledbsoma
 >>> help(tiledbsoma.DataFrame)
 
-API maturity tags:
+API maturity tags
 ------------------
 Classes and functions are annotated with API maturity tags, for example:
 
     [lifecycle: experimental]
```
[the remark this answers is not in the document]
ebezzi: Also note that I capitalized the `E` since this looks like a paragraph, but we can also keep it lower case.

apis/python/src/tiledbsoma/__init__.py
johnkerl: SInce has been rebased atop Google-style PRs

ebezzi (approved):

PR #1043, opened 2023-03-03
[python] Support Conda builds on POSIX

nguyenv:
**Issue and/or context:**

These changes support conda builds on POSIX at https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/6.

**Changes:**

- Rename Pybind11 bindings shared object library from `libtiledbsoma` to `pytiledbsoma` to avoid name mangling
- Move Pybind11 bindings (`pytiledbsoma.*.so`) from cmake to `setup.py` so that bindings library resides in same directory as the rest of the Python files
- Front-load the C++ shared object library (`libtiledbsoma.{so,dylib}`), so it can be found and imported at runtime 
- Add `__attribute__((visibility("default")))` to relevant functions to make them visibile when used in conjunction with `-fvisibility=hidden`
- Add `version.cc` to cmake shared lib build


Shelnutt2 (approved):

johnkerl:
@nguyenv I did a merge to `main` to resolve conflicts with the base branch, however, it's not clear to me that I did this correctly

johnkerl:
🎉

Issue #834, opened 2023-01-30
[python] Review TileDB-SOMA for spec compliance


atolopko-czi:
Once TileDB-SOMA RC0 is code-complete, review the TileDB-SOMA public API to ensure compliance with the SOMA [abstract specification](https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md) w.r.t class names, class properties, and method names , and method signatures.

The review team is:
@atolopko-czi , @mlin , @thetorpedodog , @aaronwolen

- [X] Any discrepancies found can simply be noted on this issue as comments. @atolopko-czi will coalesce into one or more issues for any fixes that may be needed. 
- [X] Review discrepancies with the larger team if it is not clear whether it is the implementation or the spec that should be corrected.
- [X] Python-perspective review & impl fixes, as needed (@atolopko-czi)
~- [ ] R-perspective review & impl fixes, as needed (@aaronwolen)~

atolopko-czi:
@johnkerl @@thetorpedodog We still have many "⚠️" notes in the spec.  Do we need to address these before performing a spec compliance check or proceed as is?

- [ ] "Issue: are there parts of the Arrow type system that we wish to explicitly exclude from SOMA?"
- [ ] "Issue: it would be a good idea to factor SOMAExperiment and SOMAMeasurement into separate sections."
- [ ] "[SOMADataFrame] To be further specified: all methods need specification"
- [ ] "[SOMADenseNDArray] To be further specified: this is incomplete."
- [ ] "[Operation: read()] Issue: support for other formats, such as Arrow Table, is under discussion."
- [ ] "[General Utilities] To be further specified"
- [ ] "[Value Filters] To be further specified"
- [ ] "[Other Issues (open issues with this doc)] Issues to be resolved: ..."

atolopko-czi:
Tracking review progress and action items [here](https://docs.google.com/spreadsheets/d/1LP2mQI457WXqMUhBjLtcgSzLF72akp2x8jsCJI-U_dQ/edit?usp=sharing).

johnkerl:
cc @bkmartinjr

atolopko-czi:
@thetorpedodog @mlin Found and fixed some discrepancies in spec in https://github.com/single-cell-data/SOMA/pull/145, but looking for feedback on how to handle the following additional discrepancies:
- [ ] Collection.add_new_collection: `cls` param: cls (impl) vs `type` (spec)
[...]

PR #1045, opened 2023-03-03
[python] Improve spec/impl consistency on `cls`/`type`/`kind` parameters

atolopko-czi:
**Issue and/or context:**
#834, https://github.com/single-cell-data/TileDB-SOMA/issues/834#issuecomment-1452797584

**Changes:**
Use `kind` for collection type param in `CollectionBase.add_new_collection()`

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/_collection.py
+++ b/apis/python/src/tiledbsoma/_collection.py
@@ -138,7 +138,7 @@ class CollectionBase(
     def add_new_collection(
         self,
         key: str,
-        cls: None = None,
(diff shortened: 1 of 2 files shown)


thetorpedodog (approved):
I’m somewhat surprised that mypy doesn’t complain about the name mismatch between `cls` and `kind`. I do have a companion change for the spec repo, but don’t let that change block this if we don’t need to.

johnkerl (approved):

PR #1048, opened 2023-03-03
[python] Depend on somacore 1.0.0rc4

thetorpedodog:
Things this gives us:

- "kind" as child collection type parameter
- explicit dataframe sizing in the base spec
- more accurate types
- doc updates (maturity tags)


johnkerl (approved):
🚢

PR #1052, opened 2023-03-05
[r] Refactor arrow export

eddelbuettel:
**Issue and/or context:**

Export to Arrow via the standard C interface Arrow has been refactored.  Two key changes:
- we no longer use the interim `arch` package derived from predecessors of the [`nanoarrow`](https://github.com/apache/arrow-nanoarrow) package 
- instead we include `nanoarrow` directly 'vendored' (there is also a [`cmake` mechanism](https://github.com/apache/arrow-nanoarrow/blob/main/examples/cmake-minimal/CMakeLists.txt), 
- we now 'tag' the external pointers allowing for type-checks (as we do in the `tiledb` R package

**Changes:**

The `nanoarrow` header and C file are included verbatim in the vendored form.

We also updated and corrected a few roxygen definitions (02dacfea474778bb34ab0800b88f12e14743a16f), and added a newline to `span.hpp` (69f38ee684ffea059dfee2b0f5ecdabdda672ad6).  Both suppress nags from `R CMD check` which passes cleanly now/ 

**Notes for Reviewer:**

@gspowley I had to make a minimal change to `libtiledbsoma` to (optionally) not include `carrow.h` to avoid "already defined" errors for the Arrow schema and array definition.  The (essentially) one-liner could be avoided if I used a namespace in the `nanoarrow.*` files.  A second even more minimal change was to add a single newline to `span.hpp` as `R CMD check` complains about that too.

**Errata**

The branch underlying the PR is named sc-25958, this is in error as the ticket is [SC 25968](https://app.shortcut.com/tiledb-inc/story/25968/refactor-arrow-export)


libtiledbsoma/include/tiledbsoma/arrow_adapter.h
```diff
@@ -2,7 +2,9 @@
 #define ARROW_ADAPTER_H
 
 #include <tiledbsoma/tiledbsoma>
+#ifndef ARROW_SCHEMA_AND_ARRAY_DEFINED
```
Shelnutt2, line 5: If this is a problem, I suspect we have an issue with public vs private symbols. @nguyenv / @gspowley let's make sure we are not exporting symbols outside the public API.

eddelbuettel: It's not a symbol issue. I just balks at 'data structure already defined'.  Ie if I remove the define here we end a sudden death as in:

```r
edd@rob:~/git/tiledb-soma/apis/r(de/sc-25958/refactor_arrow)$ install.r
* installing *source* package found in current working directory ...
* installing *source* package ‘tiledbsoma’ ...
** using staged installation
** static library present
** updated src/Makevars for system library via pkg-config
** libs
ccache g++  -std=gnu++17 -I"/usr/share/R/include" -DNDEBUG -I. -I../inst/include/ -I/usr/local/include  -DR_BUILD  -Iinclude -I'/usr/local/lib/R/site-library/Rcpp/include' -I'/usr/local/lib/R/site-library/RcppSpdlog/include'    -fpic  -g -O3 -Wall -pipe  -Wno-parentheses -Wno-ignored-attributes -Wno-unused-local-typedefs -Wno-deprecated-declarations -Wno-unused-function  -c rinterface.cpp -o rinterface.o
In file included from include/tiledbsoma/arrow_adapter.h:6,
                 from include/tiledbsoma/tiledbsoma:37,
                 from rinterface.cpp:6:
include/tiledbsoma/carrow.h:16:8: error: redefinition of ‘struct ArrowSchema’
   16 | struct ArrowSchema {
      |        ^~~~~~~~~~~
In file included from rinterface.cpp:2:
./nanoarrow.h:83:8: note: previous definition of ‘struct ArrowSchema’
   83 | struct ArrowSchema {
      |        ^~~~~~~~~~~
include/tiledbsoma/carrow.h:32:8: error: redefinition of ‘struct ArrowArray’
   32 | struct ArrowArray {
      |        ^~~~~~~~~~
./nanoarrow.h:99:8: note: previous definition of ‘struct ArrowArray’
   99 | struct ArrowArray {
      |        ^~~~~~~~~~
make: *** [/usr/lib/R/etc/Makeconf:178: rinterface.o] Error 1
ERROR: compilation failed for package ‘tiledbsoma’
* removing ‘/usr/local/lib/R/site-library/tiledbsoma’
* restoring previous ‘/usr/local/lib/R/site-library/tiledbsoma’
Warning message:
In install.packages(pkgs = f, lib = lib, repos = if (isMatchingFile(f)) NULL else repos) :
  installation of package ‘.’ had non-zero exit status
edd@rob:~/git/tiledb-soma/apis/r(de/sc-25958/refactor_arrow)$ 
```

Recall that all this is about using the _C interface to Arrow_ data so we are not accessing any functions or linkable symbols (unless my coffee is too weak and I miss something here).

eddelbuettel: There are alternatives.  We can protect `struct ArrowSchema` and `struc ArrowArray` on this side inside a namespace (and using via `cmake` may even automate this).  We could possibly include `nanoarrow` in `libtiledbsoma` (it has a lot of nice and simple (fast to compile) C functionality).

apis/r/DESCRIPTION
```diff
@@ -40,14 +40,12 @@ Imports:
...
-    arch (>= 0.0.2),
     dplyr,
     data.table,
     spdl
 LinkingTo:
     Rcpp,
-    RcppSpdlog,
-    arch
+    RcppSpdlog
 Additional_repositories: https://ghrr.github.io/drat
```
aaronwolen, line 49: Is this still necessary now that arch is gone?

aaronwolen (approved):
Very nice!

And thanks for catching those residual experiment query doc issues!

libtiledbsoma/include/tiledbsoma/arrow_adapter.h
eddelbuettel: @Shelnutt2 @nguyenv @gspowley Please do let know if you see issues here.  The PR is otherwise approved and ready to be merged.

gspowley (approved):
LGTM

johnkerl (approved):
🚢

eddelbuettel:
Thanks!  @gspowley I may still ping you in a day or two because the suggested use via `cmake` has a hook for a distinct namespace (and I have no idea how that works because only the header file has a `#define` for it and not the source file.)  To be seen...

Issue #1006, opened 2023-02-27
[r] Implement a read-accessor for `SOMADataFrame$index_column_names`


johnkerl:
R/Python parity as tracked on #839.

eddelbuettel:
We can close this as as #1056 was merged.

PR #1056, opened 2023-03-06
[r] Add index_column_names() (as an aliases to dimnames())

eddelbuettel:
**Issue and/or context:**

The spec states that `index_column_names()` shall return index (dimension) names.

**Changes:**

Aliases / wrapped existing `dimnames()` accessor

**Notes for Reviewer:**

#1006
[SC 26006](https://app.shortcut.com/tiledb-inc/story/26006/provide-index-column-name-accessor)


johnkerl:
Any reason to keep dimnames? IIUC we have (by intent) zero backward-compatibility concerns until the 1.0 R release.

eddelbuettel:
Because we use it everywhere else?

johnkerl:
@eddelbuettel fine with me as long as it's intentional & not just a holdover for the sake of compability with earlier (also pre-release) versions of tiledbsoma

eddelbuettel:
Intentional.  (I feared the question would come up.) But to a 20+ user of the language, `dimnames()` is very natural and idiomatic.  Yet we _of course_ want to respect the spec (hence this PR) so to me the natural is both.  "Country and Western" as per https://www.youtube.com/watch?v=vS-zEH8YmiM

johnkerl (approved):

PR #1057, opened 2023-03-07
[python] Polish `version.py`

mlin:
Low-priority:

- more-thoroughly document why we currently use this instead of setuptools_scm
- enable use from any working directory


apis/python/version.py
```diff
@@ -1,14 +1,25 @@
 # -*- coding: utf-8 -*-
 
 """
-NOTICE (mlin 2022-12-15): we'd prefer to use setuptools_scm instead of this
-ad hoc script from gist; but as of this writing, it has issues with our repo
-structure where the python package is in a subdirectory rather than the top
-level.
-  https://gist.github.com/mina86/8782771
+NOTICE (mlin 2023-03-06): this script derives the Python package version number
```
johnkerl, line 4: ❤️

johnkerl (approved):
🚢 

Thank you for the amazing knowledge management :)

Issue #1051, opened 2023-03-04
[Bug] [python] numba/numpy requirements conflict with older pip solver


mlin:
While going through [PyPI release procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/PyPI-packaging-WIP) for 1.0.0rc4, when I tried the suggested sanity test,

```
docker run -e VERSION=1.0.0rc4 --rm -it ubuntu:20.04 bash -ec '
  apt-get -qq update && apt-get install -y python3-pip
  pip3 install tiledbsoma==${VERSION}
  python3 -c "import tiledbsoma; print(tiledbsoma.libtiledbsoma.__file__)"'
```

Note the use of ubuntu 20.04, whose pip would use the [old dependency resolver](https://pip.pypa.io/en/latest/user_guide/#changes-to-the-pip-dependency-resolver-in-20-3-2020). This led to,

```
ERROR: numba 0.56.4 has requirement numpy<1.24,>=1.18, but you'll have numpy 1.24.2 which is incompatible.
...
  File "/usr/local/lib/python3.8/dist-packages/somacore/query/_fast_csr.py", line 5, in <module>
    import numba
  File "/usr/local/lib/python3.8/dist-packages/numba/__init__.py", line 42, in <module>
    from numba.np.ufunc import (vectorize, guvectorize, threading_layer,
  File "/usr/local/lib/python3.8/dist-packages/numba/np/ufunc/__init__.py", line 3, in <module>
    from numba.np.ufunc.decorators import Vectorize, GUVectorize, vectorize, guvectorize
  File "/usr/local/lib/python3.8/dist-packages/numba/np/ufunc/decorators.py", line 3, in <module>
    from numba.np.ufunc import _internal
SystemError: initialization of _internal failed without raising an exception
```

[Our setup.py requires numpy](https://github.com/single-cell-data/TileDB-SOMA/blob/fe0cc356cd4999986f983702cd8b4a9da69ce230/apis/python/setup.py#L201) with no version constraint, and that's listed before scanpy (which is where we get the numba 0.56.4 transitive requirement). I think the old dependency resolver therefore installs the too-new version of numpy first, and then when it later gets to scanpy & numba, it isn't smart enough to backtrack & install the older, compatible numpy version.

This problem does NOT occur if we change the command to use ubuntu 22.04, whose newer pip figures out to do that.

[...]

PR #1058, opened 2023-03-07
[python] Constrain numba & numpy versions for compatibility with old pip solver

mlin:
**Issue and/or context:** #1051. The old pip solver (default on ubuntu 20.04) doesn't handle the numba's particular numpy version constraints when numba itself is only our transitive dependency (via scanpy).

**Changes:** After trying a bunch of things (see commit history) it seems that pinning the exact current numba version and also lifting up its numpy constraints is the least-bad solution. Also left a comment on when the pins can/should be relaxed, and tests the ubuntu 20.04 case in in the python-packaging workflow.

### DIFF
--- a/.github/workflows/python-packaging.yml
+++ b/.github/workflows/python-packaging.yml
@@ -104,6 +104,7 @@ jobs:
         echo "WHL=$WHL" >> $GITHUB_ENV
     - name: Smoke test ${{ matrix.os }}
       run: python -c 'import tiledbsoma; print(tiledbsoma.libtiledbsoma.__file__)'
+      # TODO: more thorough local smoke test
     - name: Smoke test in docker
       # Repeat test in stock ubuntu 20.04 docker image. This uses an older pip
       # version with the old dependency solver. see issue #1051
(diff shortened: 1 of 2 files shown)


.github/workflows/python-packaging.yml
```diff
@@ -97,13 +97,24 @@ jobs:
...
         pip install wheel
-        pip install $whl
+        pip install $WHL
+        echo "WHL=$WHL" >> $GITHUB_ENV
     - name: Smoke test ${{ matrix.os }}
       run: python -c 'import tiledbsoma; print(tiledbsoma.libtiledbsoma.__file__)'
       # TODO: more thorough local smoke test
+    - name: Smoke test in docker
+      # Repeat test in stock ubuntu 20.04 docker image. This uses an older pip
+      # version with the old dependency solver. see issue #1051
```
johnkerl, line 110:
```suggestion
      # version with the old dependency solver. See issue #1051.
```

apis/python/setup.py
```diff
@@ -198,7 +198,15 @@ def run(self):
     install_requires=[
         "anndata",
         "attrs>=22.2",
-        "numpy",
+        # Pinning numba & its particular numpy constraints:
+        # The old pip solver (<=2020) doesn't deal with the transitive
+        # requirements (scanpy -> numba -> numpy) properly resulting in broken
+        # installation of incompatible numpy>=1.24. Issue #1051
```
johnkerl, line 204:
```suggestion
        # installation of incompatible numpy>=1.24. (See also issue #1051.)
```

johnkerl (approved):
🚀

PR #1061, opened 2023-03-07
[python] Support ingest of H5AD dense X as SOMA sparse X

johnkerl:
**Issue and/or context:** Some very common and very small datasets -- e.g. the ubiquitious `pbmc3k_processed` -- have their `X` stored as dense (`numpy.ndarray`). Ingesting these as dense makes them more awkward to query.

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -48,7 +48,8 @@ from ..options._tiledb_create_options import TileDBCreateOptions
 from . import conversions
 
 SparseMatrix = Union[sp.csr_matrix, sp.csc_matrix, SparseDataset]
-Matrix = Union[NPNDArray, SparseMatrix]
+DenseMatrix = Union[NPNDArray, h5py.Dataset]
+Matrix = Union[DenseMatrix, SparseMatrix]
 _NDArr = TypeVar("_NDArr", bound=NDArray)
 _TDBO = TypeVar("_TDBO", bound=TileDBObject[RawHandle])
 
@@ -63,6 +64,7 @@ def from_h5ad(
     platform_config: Optional[PlatformConfig] = None,
     ingest_mode: IngestMode = "write",
     use_relative_uri: Optional[bool] = None,
+    sparsify_X: bool = True,
 ) -> str:
     """
     Reads an ``.h5ad`` file and writes to a TileDB group structure.
@@ -104,6 +106,7 @@ def from_h5ad(
         platform_config=platform_config,
         ingest_mode=ingest_mode,
         use_relative_uri=use_relative_uri,
+        sparsify_X=sparsify_X,
     )
 
     logging.log_io(
@@ -122,6 +125,7 @@ def from_anndata(
     platform_config: Optional[PlatformConfig] = None,
     ingest_mode: IngestMode = "write",
     use_relative_uri: Optional[bool] = None,
+    sparsify_X: bool = True,
 ) -> str:
     """
     Top-level writer method for creating a TileDB group for a ``Experiment`` object.
@@ -220,7 +224,10 @@ def from_anndata(
                 # Using the latter allows us to ingest larger .h5ad files without OOMing.
                 cls = (
                     DenseNDArray
-                    if isinstance(anndata.X, (np.ndarray, h5py.Dataset))
+                    if (
+                        not sparsify_X
+                        and isinstance(anndata.X, (np.ndarray, h5py.Dataset))
+                    )
                     else SparseNDArray
                 )
                 with create_from_matrix(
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -63,6 +64,7 @@ def from_h5ad(
     platform_config: Optional[PlatformConfig] = None,
     ingest_mode: IngestMode = "write",
     use_relative_uri: Optional[bool] = None,
+    sparsify_X: bool = True,
```
bkmartinjr, line 67: Can I suggest taking a page from @thetorpedodog and making this more flexible? Ideally, the user could specify what type they want to use for X  layers.

Proposed argument: make it match _either_ `tiledbsoma.open` which uses `soma_type`, or `tiledbsoma.Collection.add_new_collect` which uses `kind`.

I.e.,     `X_kind: Union[Type["_tiledb_object.AnyTileDBObject"], str, None] = None`

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -122,6 +125,7 @@ def from_anndata(
...
```
bkmartinjr, line 128: see above comment

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -874,10 +881,12 @@ def _coo_to_table(mat_coo: sp.coo_matrix, axis: int = 0, base: int = 0) -> pa.Ta
         t1 = time.time()
 
         # Chunk size on the stride axis
-        if isinstance(matrix, np.ndarray):
+        if isinstance(matrix, (np.ndarray, h5py.Dataset)):
```
bkmartinjr, line 884: Based on some past work, I'm not sure this is a reliable way to detect the type of the X matrix as you are using an implementation detail of AnnData. There is a way to detect this using the offiial API - just index/slice the X array and you will get one of the declared return types (ndarray, scipy.sparse).

I don't think it is critical to fix this in this PR, but the above may bite us if AnnData evolves (and I believe 0.9 is around the corner).

bkmartinjr: BTW, I have code that does just this located [here](https://github.com/chanzuckerberg/single-cell-curation/blob/4097e0c370a675917f3617b1c9ce057ddc021609/cellxgene_schema_cli/cellxgene_schema/validate.py#L393)

apis/python/tests/test_basic_anndata_io.py
```diff
@@ -47,7 +47,11 @@ def adata(h5ad_file):
         ["resume"],
     ],
 )
-def test_import_anndata(adata, ingest_modes):
+@pytest.mark.parametrize(
+    "sparsify_X",
+    [True, False],
+)
+def test_import_anndata(adata, ingest_modes, sparsify_X):
```
bkmartinjr, line 54: Given that this tests `from_anndata`, which is an in-memory format that can be read from multiple file formats (zarr, h5py, etc) and in multiple modes (backed, memory-only, etc), it would be good to expand the test matrix to include more than the single in-memory H5AD. 

See comment in `ingest.py` for how this may fail in some scenarios.

bkmartinjr (changes requested):
I'd like to see a parameter with more flexibility (eg. rather than "make it sparse", it would be nice if the user could say "store X in this type/kind")

There are other comments about issues I noted with the code, but it is OK if they are pushed to another PR as they are outside the scope of this change.

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: > just index/slice the X array
this will cause an OOM :) 

maybe slicing by `[0]`

johnkerl: Definitely a great idea and definitely for a separate PR -- thank you @bkmartinjr !

johnkerl: #1064

apis/python/tests/test_basic_anndata_io.py
johnkerl: #1064

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: done

ebezzi (approved):

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -218,11 +228,18 @@ def from_anndata(
...
+
+                # XXX TEMP
+                #                cls = (
+                #                    DenseNDArray
+                #                    if (
+                #                        not sparsify_X
+                #                        and isinstance(anndata.X, (np.ndarray, h5py.Dataset))
+                #                    )
+                #                    else SparseNDArray
+                #                )
```
atolopko-czi, line 240: [post-merge] should remove

johnkerl: @atolopko-czi good call

johnkerl: follow-up PR coming soon :)

PR #1065, opened 2023-03-07
Kerl/sparsify x follow up

johnkerl:
Addresses https://github.com/single-cell-data/TileDB-SOMA/pull/1061#discussion_r1128360745 from #1061.


atolopko-czi (approved):
LGTM

PR #1066, opened 2023-03-07
Run `scripts/update-tiledb-version.py  2.15.0`

johnkerl:
TileDB Embedded 2.15.0 is released:
https://github.com/TileDB-Inc/TileDB/releases/tag/2.15.0

TileDB-Py and TileDB-R correspondings are not both yet:
https://github.com/TileDB-Inc/TileDB-Py/releases
https://github.com/TileDB-Inc/TileDB-R/releases

However, it's best to update the `libtiledbsoma` dependency first.

This is because arrays written with version n+1 can't be read by version n. We have the reader in `libtiledbsoma` and the writers via TileDB-Py and TileDB-R.


Shelnutt2 (approved):

eddelbuettel (approved):
:+1:

johnkerl:
Attempting to repro the R CI SEGV in my Ubuntu sandbox (FYI)

apis/r/.Rbuildignore
```diff
@@ -23,3 +23,6 @@ tiledbsoma/libtiledbsoma.tar.gz.current
 # vscode
 ^\.vscode$
 ^\.devcontainer$
+^travis-tool\.sh$
+^run\.sh$
+^travis_wait_.*\.log$
```
eddelbuettel, line 28: That's a leftover smell from my `run.sh` (initially written for Travis.)  Harmless.  Could remove, could remain.

apis/r/R/RcppExports.R
```diff
@@ -105,17 +105,17 @@ check_arrow_array_tag <- function(xp) {
...
 #' }
 #' @export
-sr_setup <- function(ctx, uri, colnames = NULL, qc = NULL, dim_points = NULL, dim_ranges = NULL, config = NULL, loglevel = "auto") {
-    .Call(`_tiledbsoma_sr_setup`, ctx, uri, colnames, qc, dim_points, dim_ranges, config, loglevel)
+sr_setup <- function(uri, config, colnames = NULL, qc = NULL, dim_points = NULL, dim_ranges = NULL, loglevel = "auto") {
+    .Call(`_tiledbsoma_sr_setup`, uri, config, colnames, qc, dim_points, dim_ranges, loglevel)
 }
 
-#' @rdname sr_setup
+#' @rdname sr_complete
```
eddelbuettel, line 112: sr_setup is a forward reference. If you are going through the trouble of renaming what the man page is, should we go all in and and set `@rd_name` to something meaningful like 'soma_iterated_reader' ?

apis/r/R/RcppExports.R
```diff
@@ -105,17 +105,17 @@ check_arrow_array_tag <- function(xp) {
...
 #' @export
 sr_complete <- function(sr) {
     .Call(`_tiledbsoma_sr_complete`, sr)
 }
 
-#' @rdname sr_setup
+#' @rdname sr_next
```
eddelbuettel, line 118: Ditto. This now creates a new manual page (once you run `roxygenize()`) as you created a new tag.  It was sr_setup on purpose just like the previous one.

apis/r/R/SOMAArrayBase.R
```diff
@@ -61,7 +61,7 @@ SOMAArrayBase <- R6::R6Class(
 
     # Instantiate soma_reader_pointer with a soma_reader object
     soma_reader_setup = function() {
-      private$soma_reader_pointer <- sr_setup(tiledb::tiledb_ctx()@ptr, self$uri)
+      private$soma_reader_pointer <- sr_setup(self$uri, config=as.character(config(tiledb::tiledb_ctx())))
```
eddelbuettel, line 64: That's excellent.

apis/r/src/riterator.cpp
```diff
@@ -147,7 +140,7 @@ Rcpp::XPtr<tdbs::SOMAReader> sr_setup(Rcpp::XPtr<tiledb::Context> ctx,
     Rcpp::XPtr<tdbs::SOMAReader> xptr = make_xptr<tdbs::SOMAReader>(ptr);
     return xptr;
 }
-//' @rdname sr_setup
+//' @rdname sr_complete
```
eddelbuettel, line 143: See above.

eddelbuettel (approved):
Nice catch expanding the config out from context.

Some comments on some auxiliary changes you may or may not want here.  Might be worth fixing the Rd tags to keep the PR smaller and more focussed.

apis/r/R/RcppExports.R
johnkerl: no worries, my misunderstand -- will revert

apis/r/R/RcppExports.R
johnkerl: ditto

johnkerl:
@Shelnutt2 -- as @eddelbuettel noted this may need to wait until tomorrow for the CRAN version bump of tiledb-r to fully propagate

PR #1070, opened 2023-03-08
[docs] Convert all the modules to Google-style docstrings

ebezzi:
**Issue and/or context:**
https://github.com/single-cell-data/TileDB-SOMA/issues/977

**Changes:**
Convert all the remaining modules to Google-style docstrings

**Notes for Reviewer:**
I checked the generated docs locally and they generally look good. I think some more cleanup is in order:
- Lifecycle tags do not exist for all methods. Is that on purpose or just oversight?
- Links are mostly absent, I guess due to missing references and/or missing `intersphinx` configuration. I can fix them as part of another ticket, if necessary.
- Some docstrings aren't compliant to Google style yet since they have been written in a descriptive way. An example is `tiledbsoma.io.from_h5ad`. Do we want to rewrite it using a less descriptive way? 
- I think we need to re-audit the docstrings since some methods lack args or are outdated. In some cases, these are private functions (or non exported) so this might be less important. 
- Some types are parsed by sphinx and are rendered in a suboptimal way. See this example:
image


johnkerl (approved):
> Lifecycle tags do not exist for all methods. Is that on purpose or just oversight?
> Links are mostly absent, I guess due to missing references and/or missing intersphinx configuration. I can fix them as part of another ticket, if necessary.
> Some docstrings aren't compliant to Google style yet since they have been written in a descriptive way. An example is tiledbsoma.io.from_h5ad. Do we want to rewrite it using a less descriptive way?
> I think we need to re-audit the docstrings since some methods lack args or are outdated. In some cases, these are private functions (or non exported) so this might be less important.
> Some types are parsed by sphinx and are rendered in a suboptimal way. See this example:

Let's do these all as follow-on PRs please ...

PR #1071, opened 2023-03-08
[c++] Add mode flag to `is_complete()` and add `total_num_cells()` accessor 

gspowley:
**Issue and/or context:**
Provide a way for APIs to check if a TileDB query is complete without considering if the query is empty (no ranges were added).

**Changes:**
Added a `query_status_only` flag to `is_complete()` with a default value of `false`, which does not change the current behavior.

**Notes for Reviewer:**
This PR does not change the current behavior.


eddelbuettel:
Related SC story is [2641](https://app.shortcut.com/tiledb-inc/story/26041/prevent-beyond-end-of-data-reads)

When reading 'in iterated chunks' from R, the 'instantiation and setup' step also submitted the query.  This implies that we always check query status as 'COMPLETE' (on non-huge arrays actually needing iteration).  We can now combine this with the `total_num_cells()` accessor to identify when a) queries state 'COMPLETE' yet b) we have not yet fetched data so that c) it is safe to do so and d) more importantly are safe from the issue @pablo-gar identified where are _subsequent_ read would up in tears.  No more.

eddelbuettel (approved):
LGTM but then I already noodled over this all afternoon :)

apis/r/src/riterator.cpp
```diff
@@ -147,13 +147,17 @@ Rcpp::XPtr<tdbs::SOMAReader> sr_setup(Rcpp::XPtr<tiledb::Context> ctx,
...
+
 //' @rdname sr_setup
 //' @export
 // [[Rcpp::export]]
 bool sr_complete(Rcpp::XPtr<tdbs::SOMAReader> sr) {
    check_xptr_tag<tdbs::SOMAReader>(sr);
-   spdl::info("[sr_complete] Complete test is {}", sr->is_complete());
-   return sr->is_complete();
+   size_t nobs = sr->total_num_cells();
+   bool complt = sr->is_complete(true);
```
johnkerl, line 157: total nit but can we just say ... `complete` and `result`? what are we optimizing for by leaving letters out?

johnkerl (approved):
🚢

apis/r/src/riterator.cpp
eddelbuettel: I will defer to @gspowley here -- these are existing (internal) accessor so changing them now may be feasible but we _really_ have to be sure it is worth any change and possible disruption.

eddelbuettel: Or if you refer to just the left-hand side and the variable name: Sure. I am mostly proud I moved on from `v`.

johnkerl: i'm just talking about local variable names & this is a nit

johnkerl: @eddelbuettel please let's forget about it. it's a nit. not worth the sentences we've already expended on it. i regret even mentioning it.

Issue #922, opened 2023-02-13
[c++] C++ unification tracking (high-level/parent issue)


johnkerl:
Goal of having a single copy of tiledb core via `libtiledbsoma`, eliminating all use of `tiledb-py` and `tiledb-r`:

* Currently present in `SOMAReader` (as of 2023-02-13):
    * Arrays:
        * [x] `read` for arrays
            * supporting methods like `set_dim_points`, `submit`, `results_complete`, etc.
        * [x] `nnz`
    * [x] `stats` access
* Reads:
    * Arrays
        * [x] Rename `SOMAReader` to `SOMAArrayReader`
        * [ ] Optimize CSR I/O (for Python) or CSC I/O (for R): see also #718 
        * [x] `shape` #927
        * [ ] `schema` reads (as Arrow) #1121
        * [ ] `metadata` reads #926 
        * [ ] `open` (for read) and `close`
    * Groups:
        * [ ] Make a `SOMAGroupReader` #1122
        * [ ] `open` (for read), and `close`
        * [ ] Get member name-to-URI map #1122
* Writes:
    * Arrays:
        * [ ] Make a `SOMAArrayWriter`
        * [ ] `open` (for write), and `close`
        * [ ] update metadata
        * [ ] `write` bulk data
    * Groups:
        * [ ] Make a `SOMAGroupWriter` #1182
        * [ ] `create` #1182
        * [ ] `open` (for write), and `close`
        * [ ] add/delete members
        * [ ] update metadata

Query API:

* [ ] Port query logic to C++
  * The current query logic (in `somacore`) uses _only_ the API, with no direct implementation-level accesses.
  * This could almost all be translated into C++.
  * As pre-requisite work, we'll need `SOMAArrayReader` and `SOMAGroupReader` feature-complete (as building blocks) before we can implement this in C++
  * `SOMAArrayWriter` and `SOMAGroupWriter` development can proceed in parallel with this.

Feature completion:

* [ ] Let `SOMAReader` use all supported types (not just `int` and `string` as currently) to index `SOMADataFrame`
    * https://github.com/single-cell-data/TileDB-SOMA/blob/0.5.0a11/libtiledbsoma/src/pyapi/libtiledbsoma.cc#L328-L350
 
Read performance, and in particular all of the zero-copy or ecosystem format paths (optmized CSC/CSR, etc)

* [ ] in principle, we should consider at a zero-copy path for any format that the user will want to use from R or Python.

[...]

PR #1072, opened 2023-03-08
[c++/r/python] Push down stats to `libtiledbsoma`; pass config not context to `sr_setup`

johnkerl:
**Issue and/or context:**

This is a simplifying split-out from #1066.

**Changes:**

* As long as #922 is open we have TileDB Embedded accessed two ways -- via `libtiledbsoma` (for bulk reads) and via `tiledb-r` (for metadata/etc reads, and all writes). There was a reproducible scenario on #1066 wherein we were seeing a SEGV passing a `tiledb::Context*` from the former to the latter. This PR does away with that binary-compatibility issue, instead passing a config dict across that interface.
* Likewise, stats accessors are pushed down to `libtiledbsoma` where they can be used by Python _and_ R.

**Notes for Reviewer:**

Both these changes are worth making:

* The first mod resolved a segv on #1066. As well, `sr_setup` was taking a `ctx` _only_ to manufacture its `cfg` in the case when the `cfg` argument was not supplied. Using a single argument here not only resolves the segv but is a simpler implementation.
* The second mod does not yet resolve the stats-print unit-test fail on #1066 but (as evidenced by CI passes on this PR) it does no harm when the tiledb-core version is consistent (as it not the case on #1066). Moreover, pushing more logic down into C++ `libtiledbsoma` is _the direction in which we should be going_. Every one fewer `tiledb.foo` (in Python) or `tiledb::foo` (in R) is a step in the right direction.


libtiledbsoma/src/version.cc
```diff
@@ -0,0 +1,52 @@
...
+#else
+#define VERSION "dev"
+#endif
+
+namespace tiledbsoma::version {
+
+std::string as_string() {
+    int major, minor, patch;
+    tiledb_version(&major, &minor, &patch);
+    return fmt::format(
```
eddelbuettel, line 48: This may get fmt::format linking issues.  Can you just use snprintf along these lines

```r
> Rcpp::cppFunction(r"(std::string v(int a, int b, int c) { char txt[32]; snprintf(txt, 32, "%d.%d.%d", a, b, c); return std::string(txt); })")
> v(2,3,4)
[1] "2.3.4"
> 
```

libtiledbsoma/src/stats.cc
```diff
@@ -0,0 +1,57 @@
...
+}
+
+void reset() {
+    tiledb::Stats::reset();
+}
+
+std::string dump() {
+    std::string stats;
+    tiledb::Stats::raw_dump(&stats);
+    return stats;
```
eddelbuettel, line 54: These still just pass through to TileDB Core methinks, libtiledbsoma has no separate stats instance.  So while you move the accessor into a different namespace I have not yet convinced myself actually get different stats.

eddelbuettel (commented):
A few comments inline

libtiledbsoma/src/version.cc
johnkerl: it does work & this is just a move from `src/pyapi/libtiledbsoma.cc` where it also worked :)

libtiledbsoma/src/stats.cc
johnkerl: as in the description of this PR -- i've not yet debugged everything on #1066 but the pushdown of this code to `libtiledbsoma` is something that is worth doing

johnkerl:
should be still in draft b/c i still have R CI to fix

apis/r/R/RcppExports.R
```diff
@@ -105,8 +105,8 @@ check_arrow_array_tag <- function(xp) {
 #' summary(rl)
 #' }
 #' @export
-sr_setup <- function(ctx, uri, colnames = NULL, qc = NULL, dim_points = NULL, dim_ranges = NULL, config = NULL, loglevel = "auto") {
-    .Call(`_tiledbsoma_sr_setup`, ctx, uri, colnames, qc, dim_points, dim_ranges, config, loglevel)
+sr_setup <- function(uri, config, colnames = NULL, qc = NULL, dim_points = NULL, dim_ranges = NULL, loglevel = "auto") {
```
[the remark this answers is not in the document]
johnkerl: hm, i intended it specifically to be non-nullable since it needs to be non-null

also, i touched all callsites of `sr_setup` ...

libtiledbsoma/src/version.cc
eddelbuettel: Well I think that is a little cavalier with respect to linking issues seesn with `libfmt`.  Note that Core does not call out to fmt / libfmt either.

apis/r/R/RcppExports.R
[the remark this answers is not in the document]
eddelbuettel: If it is non-nullable you are enforcing an interface change I am not in favor of.  We can accomodate both, that is exactly what nullable does.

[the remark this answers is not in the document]
eddelbuettel: As for 'all callsites' well I am currently in another branch using it too :)

libtiledbsoma/src/version.cc
johnkerl: @eddelbuettel this is simply a move of existing code from one file to another. it works, as evidenced by CI output.

apis/r/R/RcppExports.R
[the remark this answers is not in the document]
johnkerl: ok

what would you do if the config were null?

johnkerl:
* R CI working
* `tiledbsoma:::tiledb_embedded_version()` as non-exported but available for debug on #1066 
* All goals from this PR's description achieved:
  * More pushdown of code that was in `src/pyapi/tiledbsoma.cc` -- bindings and impls were both there, so impls were not available to R. Now, bindings are still there but impls are pushed down to where R bindings can get them. This is an improvement and a step in the right direction.
  * Config-passing issue from #1066: fixes a problem found there, and is also a more elegant solution
 
There's a bit more to do on #1066 but _this_ PR achieves _its_ goals.

apis/r/R/RcppExports.R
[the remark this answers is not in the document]
eddelbuettel: I think as before: not alter ctx, and just pull the config out of the existing ctx (and hence freeing the user from have to do so).

Looking at tiledb-r, a usage pattern is this:

```c++
// [[Rcpp::export]]
XPtr<tiledb::Context> libtiledb_ctx(Nullable<XPtr<tiledb::Config>> config=R_NilValue) {
    if (config.isNull()) {
        return make_xptr<tiledb::Context>(new tiledb::Context());
    } else {
        XPtr<tiledb::Config> config_xptr(config);
        return make_xptr<tiledb::Context>(new tiledb::Context(*config_xptr.get()));
    }
}
```

There are three different parts:  `Nullable<>` has to use the proper templatem and should assign the R value of NULL (as the signature lives in both C++ here, and needs to be callable from R).  You can then then via .isNull() as shown, and branch.  The above kinda does something related:  create ctx (as an external pointer) in the two cases..  I am being called away from the keyboard but we can resume later or worst case tomorrow...

eddelbuettel:
Our discussion last eve was not quite right, as I misunderstood one aspect.  The Nullable<> for the config (text) vector, as we have it right now, is not quite (and we already had a nullable config, albeit used with context object which we now no longer have.  So config cannot be nullable, it is mandatory.

As we removed `ctx` fromt the interface, please also remove it from the documentation, and add documentation for `tiledb_embedded_version`:

```
✔  checking Rd cross-references ...
W  checking for missing documentation entries (1.4s)
   Undocumented code objects:
     ‘tiledb_embedded_version’
   All user-level objects in a package should have documentation entries.
   See chapter ‘Writing R documentation files’ in the ‘Writing R
   Extensions’ manual.
✔  checking for code/documentation mismatches (3.9s)
W  checking Rd \usage sections ...
   Documented arguments not in \usage in documentation object 'sr_setup':
     ‘ctx’
   
   Functions with \usage entries need to have the appropriate \alias
   entries, and all their arguments documented.
   The \usage entries must correspond to syntactically valid R code.
   See chapter ‘Writing R documentation files’ in the ‘Writing R
   Extensions’ manual.
✔  checking Rd contents (2.4s)
```

After those two changes we should be golden.

apis/r/R/RcppExports.R
```diff
@@ -92,9 +92,9 @@ check_arrow_array_tag <- function(xp) {
 #' uri <- "test/soco/pbmc3k_processed/obs"
 #' sr <- sr_setup(uri, config=as.character(config(ctx)), loglevel="warn")
 #' rl <- data.frame()
-#' while (nrow(rl) == 0 || !tiledbsoma:::sr_complete(sr)) {
-#'     dat <- tiledbsoma:::sr_next(sr)
-#'     dat |>
+#' while (nrow(rl) == 0 || !sr_complete(sr)) {
```
eddelbuettel, line 95: The `nrow(rl) == 0` is no longer needed.  Also, when changing the roxygen stuff please edit only in the _C++ source_ file, use `compileAttributes()` to get it here and `roxygenize()` to create the .Rd file.

eddelbuettel (changes requested):

apis/r/R/RcppExports.R
johnkerl: done

apis/r/R/RcppExports.R
```diff
@@ -69,17 +69,16 @@ check_arrow_array_tag <- function(xp) {
 #'   \item{\code{sr_next}}{returns the next chunk}
 #' }
 #'
-#' @param ctx An external pointer to a TileDB Context object
 #' @param uri Character value with URI path to a SOMA data set
+#' @param config Named chracter vector with \sQuote{key} and \sQuote{value} pairs
+#' used as TileDB config parameters. If unset default configuration is used.
```
eddelbuettel, line 74: I think we should remove "If unset default configuration is used." as it is a mandatory parameter.

eddelbuettel: And see below -- this is again just the shadow.

apis/r/src/riterator.cpp
```diff
@@ -27,17 +27,16 @@ namespace tdbs = tiledbsoma;
 //'   \item{\code{sr_next}}{returns the next chunk}
 //' }
 //'
-//' @param ctx An external pointer to a TileDB Context object
 //' @param uri Character value with URI path to a SOMA data set
+//' @param config Named chracter vector with \sQuote{key} and \sQuote{value} pairs
+//' used as TileDB config parameters. If unset default configuration is used.
```
eddelbuettel, line 32: Suggest to remove "If unset default configuration is used."

Run compileAttributes() and roxygenize() and I think you can [ci skip] commit this.

eddelbuettel (approved):
This is good.  Moving things out of pyapi/ where they are "hidden" from R use is good, and changing the iterated reader to not muck with a ctx object is an excellent improvement.

One suggested minor change.

apis/r/src/riterator.cpp
johnkerl: done ✨

apis/r/R/RcppExports.R
johnkerl: done

Issue #1014, opened 2023-02-27
[Docs] [Python] Add tutorial/notebook describing TileDB-SOMA objects


pablo-gar:
Definition of done:

- A tutorial/notebook is written in the doc-site covering the basics (description, usage) of TileDB-SOMA objects:
   - DataFrame
   - SparseNDArray
   - DenseNDArray
   - Collection
   - Experiment

PR #1074, opened 2023-03-08
[docs] Add SparseNDArray section in tutorial notebook

ebezzi:
**Issue and/or context:**
https://github.com/single-cell-data/TileDB-SOMA/issues/1014

**Changes:**
- Add a pbmc3k fixture that uses `SparseNDArray`
- Add a `SparseNDArray` section to the notebook.

**Notes for Reviewer:**

Reviewer-friendly link:
https://github.com/single-cell-data/TileDB-SOMA/blob/ebezzi/sparse-ndarray-notebook/apis/python/notebooks/tutorial_soma_objects.ipynb


apis/python/notebooks/tutorial_soma_objects.ipynb
```diff
@@ -1380,7 +1380,328 @@
...
+    "- the shape of the array, i.e., the number of dimensions and the length of\n",
+    "      each dimension"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "cf955720-db72-4536-b568-134e308d17e0",
+   "metadata": {},
+   "source": [
+    "A `SparseNDArray` is functionally similar to a `DenseNDArray` but it is implemented using a sparse array. Only elements that have a nonzero value are actually stored; elements that are not explicitly stored in the dense array are assumed to be zeros."
```
johnkerl, line 1395:
```suggestion
    "A `SparseNDArray` is functionally similar to a `DenseNDArray`, except that onl elements that have a nonzero value are actually stored. Elements that are not explicitly stored are assumed to be zeros."
```

apis/python/notebooks/tutorial_soma_objects.ipynb
```diff
@@ -1380,7 +1380,328 @@
...
+   "source": [
+    "X.shape"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "5f91e73f-aa79-4fa2-a763-ad792e5641a8",
+   "metadata": {},
+   "source": [
+    "Since sparse matrices are not represented as contiguous arrays in memory, they don't have a fixed _size_ (or shape) like a dense matrix would have. Instead, `.shape()` returns the _capacity_ of the matrix, which means that those are valid indices for reading/writing to that matrix. These are dependent on the capacity of the system rather than the actual size of the array.\n",
```
johnkerl, line 1504:
```suggestion
    "Since sparse matrices are not represented as contiguous arrays in memory, they don't have a fixed size like a dense matrix would have. Instead, `.shape()` returns the _capacity_ of the matrix, which means that those are valid indices for reading/writing to that matrix. These are dependent on the capacity of the system rather than the current bounding box of the array.\n",
```

apis/python/notebooks/tutorial_soma_objects.ipynb
```diff
@@ -1380,7 +1380,328 @@
...
+    "\n",
+    "The closest concept to size for a `SparseNDArray` is the domain (or nonzero domain), which can be defined as the largest coordinates that correspond to a nonzero value (across each dimension). There is currently no way to infer the nonzero domain of a `SparseNDArray` without materializing the array."
```
johnkerl, line 1506:
```suggestion
    "The closest concept to size for a `SparseNDArray` is the non-empty domain which can be defined as the largest coordinates that correspond to a nonzero value (across each dimension). There is currently no direct way to infer the nonzero domain of a `SparseNDArray` without materializing the array; however, `obs.count` and `var.count` provide these values."
```

johnkerl (changes requested):

johnkerl (approved):

PR #1077, opened 2023-03-09
[r] Implement `tiledbsoma:::tiledb_embedded_version`

johnkerl:
On #1072 I did too many things on a single PR. Including introducing `tiledbsoma:::tiledb_embedded_version`, with a typo: I omitted the `export` tag (correct, as I don't want this an exported symbol -- this is for debug in places like #1066), but I also omitted the `Rcpp::export` tag (incorrect, as no codegen happened). Also, on discussion with @eddelbuettel for this debug function we will expose the `libtiledbsoma` TileDB Embedded version as an int-triple.


eddelbuettel (approved):
:+1:

PR #1081, opened 2023-03-09
[python] Add optional shape parameter to `SparseNDArrayRead .coos()`

ebezzi:
**Issue and/or context:**
A `SparseNDArray` can be created with a capacity that is too high for slices to be handled when reading a slice of the array. For instance, this is what gets created by default on my machine when converting a very small h5ad:

```
>>> X.shape
(9223372036854773760, 9223372036854773760)
>>> X.read((slice(0),)).coos().concat().to_scipy().todense()
ValueError: array is too big; `arr.size * arr.dtype.itemsize` is larger than the maximum possible size.
```

This issue makes working with the slices very difficult since scipy will refuse to densify the array (a similar behavior occurs when converting to a `csr_matrix`)

**Changes:**
This adds an optional parameter to `.coos()` that allows specifying a custom shape. This will be particularly useful when the exact shape of the matrix is known beforehand (for instance, when working on a single dataset, where it can be inferred by the size of `var` and `obs`).

**Notes for Reviewer:**
- Do we want to add a constraint where the custom shape must be less than (or equal) the original capacity?

### DIFF
--- a/apis/python/src/tiledbsoma/_sparse_nd_array.py
+++ b/apis/python/src/tiledbsoma/_sparse_nd_array.py
@@ -292,14 +292,18 @@ class SparseNDArrayRead(somacore.SparseRead):
         self.sr = sr
         self.shape = shape
 
-    def coos(self) -> SparseCOOTensorReadIter:
+    def coos(self, shape: Optional[NTuple] = None) -> SparseCOOTensorReadIter:
         """
         Returns an iterator of Arrow SparseCOOTensor.
+
+        Args:
+            shape:
+                Optionally, a NTuple that overrides the default capacity.
 
         Lifecycle:
             Experimental.
         """
-        return SparseCOOTensorReadIter(self.sr, self.shape)
+        return SparseCOOTensorReadIter(self.sr, shape or self.shape)
 
     def dense_tensors(self) -> somacore.ReadIter[pa.Tensor]:
         """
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -292,14 +292,18 @@ def __init__(self, sr: clib.SOMAReader, shape: NTuple):
...
         self.shape = shape
 
-    def coos(self) -> SparseCOOTensorReadIter:
+    def coos(self, shape: Optional[NTuple] = None) -> SparseCOOTensorReadIter:
         """
         Returns an iterator of Arrow SparseCOOTensor.
 
+        Args:
+            shape:
+                Optionally, a NTuple that overrides the default capacity.
```
johnkerl, line 301:
```suggestion
                Optionally, a tuple that overrides the default capacity.
```

apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -292,14 +292,18 @@ def __init__(self, sr: clib.SOMAReader, shape: NTuple):
...
+
         Lifecycle:
             Experimental.
         """
-        return SparseCOOTensorReadIter(self.sr, self.shape)
+        return SparseCOOTensorReadIter(self.sr, shape or self.shape)
```
johnkerl, line 306: very simple and very elegant! :) 

let's add a check that (if shape is not None) `len(shape) == len(self.shape)` and raise `ValueError` if not -- with a unit-test `pytest.raises(ValueError)` for this as well

johnkerl (changes requested):

johnkerl (approved):
🚀

Issue #927, opened 2023-02-13
[c++] Add shape-getter to `SOMAReader` API


johnkerl:
Split-off from #922

PR #1083, opened 2023-03-09
[c++/python/r] Add `shape` to C++ `SOMAReader`

nguyenv:
**Issue and/or context:**
#927 

**Changes:**

- [x] Add `shape` to C++ `SOMAReader`
- [x] Bind in Pybind11; replace in Python API
- [x] Bind in Rcpp; replace in R API


eddelbuettel (approved):
Thumbs up for the R and C++ side of things, staying out of reviewing Python.

apis/r/src/rutilities.h
```diff
@@ -14,6 +14,17 @@ namespace tdbs = tiledbsoma;
...
+  Rcpp::NumericVector num(n);
+  std::memcpy(&(num[0]), vec.data(), n*sizeof(double));
+
+  num.attr("class") = "integer64";
+  return(num);
+}
+
 // Convert to a scalar int64_t
 //
 inline int64_t makeScalarInteger64(const double val) {
```
johnkerl, line 30: as we discussed offline with the three of us, this using double as an intermediate is the correct thing to do, and the straightforward/obvious cast is _not_ the correct thing to do

which is functionally fine, but very non-obvious and will surely confuse anyone else who comes along later

let's get some comments in here why this wrong-looking thing is right-functioning

johnkerl (changes requested):

apis/r/src/rutilities.h
eddelbuettel: I am sorry but you may be looking at the wrong code getting the wrong impression. Also note that this is not a user facing function, nor is it (AFAICT) called in this PR.

nguyenv: Updated: https://github.com/single-cell-data/TileDB-SOMA/pull/1083/files#diff-e450bf4b6b00306a671ed796c0f66adbc26fda96fba948fd8df54d4b2a90243cR19-R26

apis/r/src/rutilities.h
```diff
@@ -3,16 +3,36 @@
...
 // current build is encoded in TILEDB_VERSION
-#define TILEDB_VERSION TileDB_Version(TILEDB_VERSION_MAJOR,TILEDB_VERSION_MINOR,TILEDB_VERSION_PATCH)
+#define TILEDB_VERSION \
+    TileDB_Version(    \
+        TILEDB_VERSION_MAJOR, TILEDB_VERSION_MINOR, TILEDB_VERSION_PATCH)
+
+// Create a integer64 object
+//
+// Integer64 is an S3 class. Integers in R are 32-bits. To handle C++
+// signed 64-bit integers (uint64_t), the full bits may be stored using double '
```
eddelbuettel, line 22: s/uint64_t/int64_t/

apis/r/src/rutilities.h
```diff
@@ -19,7 +19,7 @@ namespace tdbs = tiledbsoma;
 // Create a integer64 object
 //
 // Integer64 is an S3 class. Integers in R are 32-bits. To handle C++
-// signed 64-bit integers (uint64_t), the full bits may be stored using double '
+// signed 64-bit integers (int64_t), the full bits may be stored using double '
```
johnkerl, line 22: ✅

johnkerl (approved):

PR #1085, opened 2023-03-10
[r] Handle config in non-iterated readers

johnkerl:
As reported by @mlin . Similar scenario as #1078 but with non-iterated readers.

Repro script:

```
#!/usr/bin/env Rscript

library(tiledbsoma)
library(tiledb)

uri <- 's3://.....'

config <- tiledb_config()
config["vfs.s3.region"] <- "us-west-2"
ctx <- tiledb_ctx(config)

cat("Loading experiment:\n")
exp <- SOMAExperiment$new(uri, ctx=ctx)

cat("\nGetting var:\n")
sdf <- exp$ms$get("RNA")$var

cat("\nDoing read:\n")
print(sdf$read())
```

This was failing to get the AWS region correct:

```
Error in soma_reader(uri = uri, colnames = column_names, qc = value_filter,  :
  Error opening array: 's3://cellxgene-data-public/cell-census/2023-03-06/soma/census_data/homo_sapiens/ms/RNA/var'
  [TileDB::Array] Error: Error while listing with prefix 's3://cellxgene-data-public/cell-census/2023-03-06/soma/census_data/homo_sapiens/ms/RNA/var/' and delimiter '/'
Exception:  PermanentRedirect
Error message:  Unable to parse ExceptionName: PermanentRedirect Message: The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.
```

With this PR it works correctly:

```
Table
60664 rows x 4 columns
$soma_joinid <int64 not null>
$feature_id <large_string not null>
$feature_name <large_string not null>
$feature_length <int64 not null>
```

I don't know a great way to unit-test this but I do know this change is the right thing to do. (It's a re-use of an existing, known-good pattern from `apis/r/src/riterator.cpp` to `apis/r/src/rinterface.cpp`.)

### DIFF
--- a/apis/r/R/SOMADataFrame.R
+++ b/apis/r/R/SOMADataFrame.R
@@ -189,7 +189,8 @@ SOMADataFrame <- R6::R6Class(
                             colnames = column_names,   # NULL is dealt with by soma_reader()
                             qc = value_filter,         # idem
                             dim_points = coords,       # idem
-                            loglevel = log_level)      # idem
+                            loglevel = log_level,      # idem
+                            config=as.character(tiledb::config(self$ctx)))
           private$soma_reader_transform(rl)
       } else {
           ## should we error if this isn't null?
--- a/apis/r/R/SOMADataFrame.R
+++ b/apis/r/R/SOMADataFrame.R
@@ -190,7 +190,7 @@ SOMADataFrame <- R6::R6Class(
                             qc = value_filter,         # idem
                             dim_points = coords,       # idem
                             loglevel = log_level,      # idem
-                            config=as.character(tiledb::config(self$ctx)))
+                            config = as.character(tiledb::config(self$ctx)))
           private$soma_reader_transform(rl)
       } else {
           ## should we error if this isn't null?
--- a/apis/r/R/SOMADenseNdArray.R
+++ b/apis/r/R/SOMADenseNdArray.R
@@ -124,7 +124,8 @@ SOMADenseNDArray <- R6::R6Class(
           rl <- soma_reader(uri = uri,
                             dim_points = coords,        # NULL is dealt with by soma_reader()
                             result_order = result_order,
-                            loglevel = log_level)       # idem
+                            loglevel = log_level,       # idem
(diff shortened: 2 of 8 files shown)


apis/r/R/SOMADataFrame.R
```diff
@@ -189,7 +189,8 @@ SOMADataFrame <- R6::R6Class(
                             colnames = column_names,   # NULL is dealt with by soma_reader()
                             qc = value_filter,         # idem
                             dim_points = coords,       # idem
-                            loglevel = log_level)      # idem
+                            loglevel = log_level,      # idem
+                            config=as.character(tiledb::config(self$ctx)))
```
eddelbuettel, line 193: Please use the same whitespace as in the lines above it.

apis/r/R/SOMADenseNdArray.R
```diff
@@ -124,7 +124,8 @@ SOMADenseNDArray <- R6::R6Class(
           rl <- soma_reader(uri = uri,
                             dim_points = coords,        # NULL is dealt with by soma_reader()
                             result_order = result_order,
-                            loglevel = log_level)       # idem
+                            loglevel = log_level,       # idem
+                            config=as.character(tiledb::config(self$ctx)))
```
eddelbuettel, line 128: As above

apis/r/R/SOMASparseNdArray.R
```diff
@@ -123,7 +123,8 @@ SOMASparseNDArray <- R6::R6Class(
           rl <- soma_reader(uri = uri,
                             dim_points = coords,        # NULL is dealt with by soma_reader()
                             result_order = result_order,
-                            loglevel = log_level)       # idem
+                            loglevel = log_level,       # idem
+                            config=as.character(tiledb::config(self$ctx)))
```
eddelbuettel, line 127: As above

eddelbuettel (changes requested):
See inline

eddelbuettel (approved):

PR #1095, opened 2023-03-10
[r] Publish TileDB-SOMA-R `pkgdown` docs to GitHub Pages

johnkerl:
**Issue and/or context:** Issue #1041. See also PR #1097 which disables publishing from `main-old`.

**Changes:**
Simply promotes `main-old` logic to `main`.

**Notes for Reviewer:**
This publishes to https://single-cell-data.github.io/TileDB-SOMA


eddelbuettel (dismissed):
Please put

```
template:
  bootstrap: 5
```

at the top of the `_pkgdown.yml` as the old look and feel looks ... old and dated.

Cf https://github.com/TileDB-Inc/TileDB-R/blob/5764bd45b798425568e0efd2343bdf11cc7ca5f2/_pkgdown.yml#L2-L3  as a reference

apis/r/tools/install-linux-dependencies.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -euo pipefail
+# Context: .github/workflows/*.yaml
+apt-get update -y && apt-get install -y libgeos-dev
```
eddelbuettel, line 4: Are we sure we need `libgeos` for `rgeos` anywhere?

johnkerl: 🤷 Probably not -- this is no more & no less than a copy from `main-old`.

johnkerl:
@eddelbuettel weird, this repo doesn't _have_ a `_pkgdown.yml` ... 👀 ...

apis/r/tools/install-linux-dependencies.sh
mojaveazure: `main-old` has (IIRC) a dependency on [SeuratObject](https://cran.r-project.org/package=SeuratObject), which (used to) have a dependency on rgeos; `main` does not have a dependency on SeuratObject, and SeuratObject now suggests rgeos rather than depends on it

eddelbuettel: Perfect, so let's nix the script, or keep it but remove its call.

Shorter CI cycles are good.  More feedback, faster.

eddelbuettel:
Then I am confused it rendered at all.  You probably want to add one, we will need it eventually methinks.

apis/r/DESCRIPTION
```diff
@@ -16,6 +16,10 @@ Authors@R: c(
            email = "dirk@tiledb.com",
            role = "aut",
            comment = c(ORCID = "0000-0001-6419-907X")),
+    person(given = "Paul",
+           family = "Hoffman",
+           email = "paul.hoffman@tiledb.com",
+           role = "aut"),
```
mojaveazure, line 22: Can you add my ORCID?
```suggestion
    person(given = "Paul",
           family = "Hoffman",
           email = "paul.hoffman@tiledb.com",
           role = "aut",
           comment = c(ORCID = "0000-0002-7693-8957")),
```

johnkerl: @mojaveazure already added you but TIL "ORCID" :)

mojaveazure (approved):

PR #1098, opened 2023-03-10
[python] Get `readthedocs` going

johnkerl:
**Issue and/or context:** #1041 

**Changes:**

Gets Python API docs into `readthedocs`. 

**Notes for Reviewer:**

Current preview link (not on main): https://tiledbsoma.readthedocs.io/en/kerl-readthedocs-initial/python-api.html

There's more work to do but this gets the bulk of the docs available on the web for users to see. In particular `readthdocs` still  needs a button poked manually; I don't have commit-triggers set up. Also some info about this (and R's `pkgdown`) needs to be written in this repo's developer wiki.

### DIFF
--- a/apis/python/src/tiledbsoma/_dense_nd_array.py
+++ b/apis/python/src/tiledbsoma/_dense_nd_array.py
@@ -25,6 +25,7 @@ class DenseNDArray(NDArray, somacore.DenseNDArray):
     """``DenseNDArray`` is a dense, N-dimensional array, with offset (zero-based)
     integer indexing on each dimension. ``DenseNDArray`` has a user-defined
     schema, which includes:
+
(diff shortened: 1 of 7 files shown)


apis/python/src/tiledbsoma/_dense_nd_array.py
```diff
@@ -3,7 +3,8 @@
 #
 # Licensed under the MIT License.
 
-"""Implementation of SOMA DenseNDArray.
+"""
+Implementation of SOMA DenseNDArray.
```
ebezzi, line 7: I also prefer the newline between the triple quote and the text, but the Google guidelines for the docstrings don't have it. If we all like this style better, maybe we should make an exception but then we should make it consistent across the codebase.

ebezzi (approved):

PR #1100, opened 2023-03-12
[r] Accomodate legacy validity setting

eddelbuettel:
**Issue and/or context:**

This PR accommodates the new legacy validity mode (for nullable string columns) added in TileDB-Inc/TileDB-R#517 

**Changes:**

A new metadata tag, `soma_legacy_validity`, is now attached to all arrays created by `SOMA` objects. By default this value is `"false"` unless the TileDB-R legacy validity mode was enabled at creation time (i.e., via config option `r.legacy_validity_mode`). When reading arrays from disk, the `AnnotationDataFrame` class will check for this tag on initialization and when performing reads or writes. If the tag is present and set to `"true"`, legacy validity mode is enabled globally (as it's not possible to set on a per-array basis). Legacy validity mode is also enabled when reading `AnnotationDataFrame` arrays that lack the tag, as this indicates the array was created with an older version of the package. These checks are limited to `AnnotationDataFrame` arrays because the incorrect validity map values only affect nullable string attributes. See TileDB-R's [release notes](https://github.com/TileDB-Inc/TileDB-R/blob/master/NEWS.md#breaking-changes) for more information.

**Notes for Reviewer:**

See [PR 99 in tiledbsc](https://github.com/TileDB-Inc/tiledbsc/pull/99) and [SC 25774](https://app.shortcut.com/tiledb-inc/story/25774/adjust-tiledbsc-and-main-old-for-legacy-validity).


aaronwolen (approved):
Thanks for carrying this over.

PR #1101, opened 2023-03-13
[c++/r/python] Add libtiledbsoma info to `show_package_versions()`

mlin:
RELATED: #1066 

Probably because I've been installing stuff from tiledb-inc.r-universe.dev, I did get myself into a situation where R is using TileDB 2.15 creating stuff that's unreadable to libtiledbsoma using 2.14. This was confusing so here's just adding a line to both the Python and R `show_package_versions()` to report the version of libtiledbsoma and the TileDB Embedded linked into it, potentially distinct from the TileDB Embedded version loaded by TileDB-Py and TileDB-R.

Just makes diagnosis a little easier next time. We could consider making our package init compare the versions and show some warning if the libtiledbsoma TileDB version appears older.


apis/r/tests/testthat/test-version.R
```diff
@@ -1,6 +1,10 @@
-test_that("TileDB Embedded is available", {
+test_that("libtiledbsoma version info is available", {
```
eddelbuettel, line 1: You are adding not replacing here so maybe keep post context setters?

eddelbuettel (approved):
Looking only at the R side of things, this seems fine even when the sha1 is simply not that informative.

johnkerl (approved):
🚢

mlin:
@johnkerl @eddelbuettel I left some breadcrumb comments in 0b3393fa24392268b04c8ee08bb2b67367eb12ad to hopefully help the next guy/gal who gets into the same trouble I did -- please check for accuracy

apis/r/tests/testthat/test-version.R
```diff
@@ -1,5 +1,5 @@
-test_that("libtiledbsoma version info is available", {
-    # Mainly testing this function exists, without introspecting
+test_that("version info for libtiledbsoma and its linked TileDB Embedded", {
+    # Mainly testing these functions exist, without introspecting
```
eddelbuettel, line 2: Thank you!

apis/r/tools/get_tarball.R
```diff
@@ -1,5 +1,11 @@
...
 
+# The TileDB Embedded version specified here will be linked to the libtiledbsoma native lib loaded
+# by our tiledbsoma R package. The R package code -also- uses TileDB-R, which links its own 'copy'
+# of TileDB Embedded, whose version we don't control here. Ideally the TileDB Embedded versions
+# should match! The show_package_versions() helper function in each package can help to diagnose
+# any mismatch.
+
 ## todo: chipset for macOS to detect arm
 isX86 <- Sys.info()["machine"] == "x86_64"
 isMac <- Sys.info()['sysname'] == "Darwin"
```
eddelbuettel, line 11: I think I used `uname -m` and comparison for `arm64` to detect non-intel macOS.

eddelbuettel (approved):
Looks good.  One quick comment re arm64 inline

Issue #991, opened 2023-02-23
[r] Unit-test accessing `nnz` when `allow_duplicates=True`


johnkerl:
Issue #988 / PR #989 is almost entirely C++ -- however there is a bit of Python unit-test validation that everything is wired all the way through to the user level.

eddelbuettel:
Waiting for #1038

PR #1107, opened 2023-03-14
[r] Add test for nnz() when duplicates==TRUE

eddelbuettel:
**Issue and/or context:**

The spec desires a test for `nnz()` when the schema contains duplicates

**Changes:**

A test has been added.  Given that the package does not yet cover _writing_ arrays from R we farm parts out to the TileDB R package.

**Notes for Reviewer:**

#991 
[SC 26467](https://app.shortcut.com/tiledb-inc/story/26467/ass-test-for-support-of-nnz-under-duplicates-true)


johnkerl (approved):
🚢

PR #1110, opened 2023-03-15
[r] `SOMASparseNDArray$read_sparse_matrix`: set dims on the returned `sparseMatrix`

mlin:
#1099

Quick hit here to set the `sparseMatrix` dims for non-iterated reads only. I have WIP on iterated reads, but ran into a few complications, so I'd like to keep working on that in additional PR(s) to come.

Issue #1099, opened 2023-03-12
[r] `SOMASparseNdArray$read_sparse_matrix`: returned matrix should have dims


mlin:
We don't yet set the dims of the `sparseMatrix` returned by `SOMASparseNdArray$read_sparse_matrix`, so they default to the max coordinates actually present in the returned data.

PR #1112, opened 2023-03-15
[r] Fix testing logic for iterated reads from `SOMASparseNDArray`

mlin:
RELATED: #1099, #1110

The existing tests "Iterated Interface from SOMA Sparse Matrix" were passing, but they were checking weird quantities in a way I doubt was intentional (but do correct me otherwise :). It checked that the sum of `dim(dat)[1]` over each returned `sparseMatrix` shard `dat` was 6596. This is a weird assertion because the whole test matrix has only 2638 rows. As we aren't yet expressly setting the dims of each `sparseMatrix` shard, `dim(dat)` defaults to (max row number with a nonzero entry, max column number with a nonzero entry) of each shard -- see `?Matrix::sparseMatrix`. As we progress in row-major order, the max row number keeps increasing to 2638 on the last iteration; thus the running total is much greater than nrow=2638.

This diff changes the test to verify more-natural quantities, the total number of nonzero rows and the total number of nonzero entries (though our actual test array is all-nonzero).

This also invites discussion of what `dim(dat)` *should* be. I think the only possible answers are the current default or the shape of the whole sparse array (even though we're only returning a shard of it). I suggest the latter since the former is more confusing and unpredictable as seen here. discussion?

### DIFF
--- a/apis/r/tests/testthat/test_SOMAReader-Iterated.R
+++ b/apis/r/tests/testthat/test_SOMAReader-Iterated.R
@@ -133,36 +133,26 @@ test_that("Iterated Interface from SOMA Sparse Matrix", {
 
     sdf$read_sparse_matrix(iterated = TRUE)
 
-    expect_false(sdf$read_complete())
-    dat <- sdf$read_next()
-    d <- dim(dat)
-    expect_equal(d[2], 1838)
-    n <- d[1]
-    expect_true(n > 0)
-
-    expect_false(sdf$read_complete())
-    dat <- sdf$read_next()
-    d <- dim(dat)
-    expect_equal(d[2], 1838)
-    n <- n + d[1]
-    expect_true(n > 0)
-
-    expect_false(sdf$read_complete())
-    dat <- sdf$read_next()
-    d <- dim(dat)
-    expect_equal(d[2], 1838)
-    n <- n + d[1]
-    expect_true(n > 0)
-
-    expect_false(sdf$read_complete())
-    dat <- sdf$read_next()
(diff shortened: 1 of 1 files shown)


eddelbuettel:
I nearly had a minor heart-attack reading just the `diff` here as it almost implies that we may not read the right data or amounts. :grinning: but all is still well.  A quick and dirty check based on the large sample data set 

```r

suppressMessages({
    library(tiledbsoma)
    library(arrow)
    library(bit64)
    library(data.table)
    library(dplyr)
    library(tiledb)
})

tdir <- tempfile()
tgzfile <- system.file("raw-data", "soco-pbmc3k.tar.gz", package="pbmc3k.tiledb")
untar(tarfile = tgzfile, exdir = tdir)

uri <- file.path(tdir, "soco", "pbmc3k_processed", "ms", "RNA", "X", "data")

ctx <- tiledb_ctx()
sr <- sr_setup(uri, config=as.character(config(ctx)))
rl <- data.frame()
while (!tiledbsoma:::sr_complete(sr)) {
    dat <- sr_next(sr)
    D <- tiledbsoma:::arrow_to_dt(dat)
    rl <- rbind(rl, D)
}

chk <- tiledb::tiledb_array(uri, return_as="data.table")[]
all.equal(dim(rl), dim(chk))
## cannot compare directly as ordering different
all.equal(summary(rl$soma_data), summary(chk$soma_data))
```

eddelbuettel (approved):
Looks good and a nice trick to read the count from `Matrix`.

PR #1113, opened 2023-03-15
[python] sdist debug on 1.1.0 [WIP / please do not review]

johnkerl:
**Issue and/or context:**

**Changes:**

**Notes for Reviewer:**

PR #1115, opened 2023-03-15
[python] Pin typeguard to < 3.0

johnkerl:
**Issue and/or context:**

https://pypi.org/project/typeguard/3.0.0/ just dropped yesterday. As a result all new CI is failing with

```
    from typeguard.importhook import install_import_hook
E   ModuleNotFoundError: No module named 'typeguard.importhook'

```

**Changes:**

Don't use typeguard 3.

### DIFF
--- a/.github/workflows/python-ci-single.yml
+++ b/.github/workflows/python-ci-single.yml
@@ -89,7 +89,7 @@ jobs:
         key: libtiledbsoma-build-dist-${{ inputs.os }}-${{ inputs.python_version }}-${{ hashFiles('libtiledbsoma', 'scripts/bld') }}
 
     - name: Install testing prereqs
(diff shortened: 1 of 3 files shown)


atolopko-czi (approved):
Pin All The Things!

bkmartinjr:
@johnkerl - why not just fix the import and update to latest?  The 3.0 release notes specify the change and therefore the likely fix:
> Moved install_import_hook to be directly importable from the typeguard module

johnkerl:
@bkmartinjr thanks! At this point as this fail (solved) and the sdist fail (not solved -- #1113) are blocking some internal flows, my #1 priority is "unbreak and continue" with mimimal disruption.

Happy to upgrade our stuff to track the typeguard upgrade as a follow-on once we've recovered the ability to tag a tiledbsoma release.

PR #1119, opened 2023-03-15
[r] Add SOMA Experiment Query -> Seurat outgestors

### LINKED ISSUE #941: [r] Implement `to_seurat`
Context: #839

mojaveazure:
Add support for creating `Seurat` objects and subobjects (eg. `Assay`, `DimReduc`) from SOMA experiment queries; all done using SeuratObject as a suggested dependency

Implemented SOMA methods:
- `SOMAExperimentAxisQuery$to_seurat_assay()`: allows loading in just expression data (`ms/X`)
- `SOMAExperimentAxisQuery$to_seurat_reduction()`: allows loading in just dimensional reduction information (`ms/obsm`, `ms/varm`)
- `SOMAExperimentAxisQuery$to_seurat_graph()`: allows loading in just nearest-neighbor graphs (`ms/obsp`)
- `SOMAExperimentAxisQuery$to_seurat()`: allows loading in entire `Seurat` objects, including partially suppressing layers from `ms/X`, dimensional reduction information, and nearest-neighbor graphs

Direct dependencies added:
- rlang (already a dependency via arrow)

Questions for reviewers:
- Method arguments are a mix of SOMA-style (`X_layer_names`, `var_column_names`) and Seurat-style (`cells_index`, `features_index`) names. Should these be set as just SOMA-style (`cells_index` &rarr; `obs_index`, `graphs` &rarr;` `obsp_arrays`) or just `Seurat` style (`X_layer_names` &rarr; `layers`)?
- Should methods for Seurat-based generics be added with [delayed method registration](https://vctrs.r-lib.org/reference/s3_register.html) (eg. `as.Seurat.SOMAExperimentAxisQuery`, called simply as `as.Seurat(query)`)? vctrs is already a dependency via arrow, so no additional dependencies will be added

Future work:
- This PR does not address multimodal SOMAs; queries are as strictly unimodal, so similar functionality will need to be ported to the `SOMAExperiment` level to support multiple modalities

addresses #941


eddelbuettel:
`rlang::is_na is currently missing from NAMESPACE as an imports.

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -279,7 +283,7 @@ SOMAExperimentAxisQuery <- R6::R6Class(
         x = self$var_df$attrnames(),
         y = features_index
       )
-      if (!(isFALSE(var_column_names) || rlang::is_na(var_column_names))) {
+      if (!(isFALSE(var_column_names) || is_na(var_column_names))) {
```
eddelbuettel, line 286: (FWIW I'd keep the prefix to make it clear where it comes from.  You need the NAMESPACE nag to avoid R CMD check from coming after you...)

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -1,3 +1,7 @@
+#' @importFrom rlang is_na
```
eddelbuettel, line 1: :+1:

johnkerl:
Just an overall context/FYI (no criticism of this PR -- just background):

* In TileDB-SOMA-Py, `tiledbsoma.io.to_anndata` takes _all_ the members of the SOMA experiment (`obs`, `var`, `X`, `obsm`, `obsp`, `varm`, `varp`) and creates an `AnnData` object, after no filtering/selecting/querying of the SOMA experiment
* `ExperimentAxisQuery.to_anndata` takes only `obs`, `var`, and `X`, and creates an `AnnData` object, after user-specified filtering/querying of the SOMA experiment

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -169,15 +169,210 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+          is.character(embeddings) ||
+          is_scalar_logical(embeddings),
+        "'loadings' must be a named character vector" = is.null(loadings) ||
+          (is.character(loadings) && is_named(loadings, allow_empty = FALSE)) ||
+          is_scalar_logical(loadings),
+        "'graphs' must be a character vector" = is.null(graphs) ||
+          is.character(graphs) ||
+          is_scalar_logical(graphs)
+      )
+      tryCatch(self$obs_df, error = \(...) stop("No 'obs' found", call. = FALSE))
```
eddelbuettel, line 227: Writing an anonymous function this way is nice but creates a hard depends on R 4.1.0.  _That is probably fine_ but if we wanted to be more defensive we could the keyword `function` in.

eddelbuettel (approved):
This looks pretty good. I am not in position to argue domain-specific things here -- I of course trust that you of  all coders know what should be in a Seurat object :grinning: 

I added one small comment re the use of `\()`.

bkmartinjr (approved):
With the standard caveats around my R skills, this looks great.  I am biasing toward landing it soon, so we can get concrete early user feedback on the open issues, prior to a 1.0 release.

> Method arguments are a mix of SOMA-style and Seurat-style

On this question: the best advice I can give is what we tried to do for the Python API equivalent.  Where args  controlled SOMA aspects or behavior (eg., "column_names"), we used the SOMA name. Where it controlled some AnnData/ScanPy behavior, we tried to use AnnData-esque names (example "X_layers", which isn't perfectly AnnData, but is pretty close).

I'm not sure there is a single answer - it depends on what the argument indicates.

CC: @pablo-gar in case you have additional comments.

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -163,6 +168,527 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+        }
+        warning(msg, call. = FALSE, immediate. = TRUE)
+        varm_layer <- NULL
+      } else {
+        names(ms_load) <- .anndata_to_seurat_reduc(ms_load, 'loadings')
+      }
+      # Check provided names
+      if (!obsm_layer %in% c(ms_embed, names(ms_embed))) {
+        stop("Cannot find embeddings ", sQuote(obsm_layer), call. = FALSE)
+      }
```
aaronwolen, line 551: Another candidate for `assert_subset()`

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -163,6 +168,527 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+      # Find Seurat name
+      seurat <- c(
+        .anndata_to_seurat_reduc(obsm_layer),
+        tryCatch(
+          expr = .anndata_to_seurat_reduc(varm_layer, 'loadings'),
+          error = null
+        )
+      )
+      if (length(seurat) == 2L && !identical(x = seurat[1L], y = seurat[2L])) {
```
aaronwolen, line 563: Might be slightly clearer as a named list or vector so downstream references can be expressed as `seurat$obsm_layer` (or similar) rather than `seurat[1L]`.

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -163,6 +168,527 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+          sQuote(varm_layer),
+          "); using the embeddings to create a Seurat name (",
+          sQuote(seurat[1L]),
+          ")"
+        )
+        warning(
+          paste(strwrap(msg), collapse = '\n'),
+          call. = FALSE,
+          immediate. = TRUE
+        )
```
aaronwolen, line 577: Should we error out here?

aaronwolen (commented):
This looks really good! I have a couple requested changes but nothing major.

aaronwolen (changes requested):
This looks really good! I have a couple requested changes but nothing major.

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -163,6 +168,527 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+            strwrap(paste(
+              "The following layers cannot be found in this ExperimentQuery:",
+              paste(sQuote(setdiff(x = X_layers, y = layers)), collapse = ', ')
+            )),
+            collapse = '\n'
+          ),
+          call. = FALSE,
+          immediate. = TRUE
+        )
+      }
```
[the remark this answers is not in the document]
mojaveazure: Typically in Seurat, we try to persevere for small issues and only error out if something actually cannot be done. In this case, if two layers were passed but one is non existent, Seurat would typically throw a warning and continue on. Seurat would only error if absolutely nothing could be done (eg. no layers existing)

In the (hopefully near) future, I want to add support for appending to existing `Seurat` objects, which would allow the user to make mistakes and correct them later on

apis/r/R/SOMAExperimentAxisQuery.R
mojaveazure: I've changed to an error; see above for rationale behind warnings instead of errors

aaronwolen (approved):
Thanks for the changes!

PR #1125, opened 2023-03-15
[r] `SOMASparseNDArray$read_sparse_matrix`: set dims on iterated `sparseMatrix` parts

mlin:
For iterated reads, I think the dims of each shard have to match the whole array shape, because in principle we could be returning entries from anywhere in the array.

It would be nice to have some other slot that would make it easy for the user to figure out the 'bounding box' of the returned shard, without having to do something cute like `nnzRows <- function(m) { sum(Matrix::rowSums(m != 0) > 0) }` (which is partly due to ChatGPT btw =). Not sure where to hang that though, any suggestions?

Should complete #1099


eddelbuettel:
[ Just as a very general comment, my $0.02 based on experience with tiledb and the core library: memory management on _sparse_ arrays is simply difficult as you never know "what box of chocolates you gonna get". So for dense array we get known fixed quantities, otherwise it may not be possible -- which is a bit frustrating but maybe par for the course as 'sparsity may differ'.  ]

I'll put your PR through the motions tomorrow morning. It looks very clean.

eddelbuettel (approved):
:+1:

PR #1127, opened 2023-03-15
[ci] Ensure macOS builds use clang instead of gcc

mlin:
Appears to fix current main CI...unclear if setting gcc was intentional, and we do NOT do that for the wheel builds.

### DIFF
--- a/.github/workflows/python-ci-full.yml
+++ b/.github/workflows/python-ci-full.yml
@@ -20,8 +20,6 @@ jobs:
             cc: gcc-11
             cxx: g++-11
           - runs-on: macos-12
-            cc: gcc-11
-            cxx: g++-11
     uses: ./.github/workflows/python-ci-single.yml
     with:
       os: ${{ matrix.os }}
--- a/.github/workflows/python-ci-full.yml
+++ b/.github/workflows/python-ci-full.yml
@@ -1,10 +1,17 @@
 name: TileDB-SOMA Python CI (Full)
 
+# This workflow calls ./python-ci-single.yml on the full {os} x {python version}
+# matrix. Since that's CI-resource-intensive, it runs only for main branch and
+# releases.
 on:
   push:
     branches: [main]
   release:
     types: [published]
+  # You can also invoke this workflow manually from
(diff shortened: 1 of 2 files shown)


Shelnutt2 (approved):

.github/workflows/python-ci-full.yml
```diff
@@ -20,8 +20,6 @@ jobs:
             cc: gcc-11
             cxx: g++-11
           - runs-on: macos-12
```
johnkerl, line 22: > Appears to fix current main CI...unclear if setting gcc was intentional, and we do NOT do that for the wheel builds.

@mlin we can merge this & that's fine but the pre-merge minimal CI doesn't run on MacOS.

Just to verify this works, please _temporarily_ include a `- runs-on: macos-12` line in `python-ci-minimal.yml` and push & make sure it builds OK, then remove it, then merge to `main` with @Shelnutt2 's accpt

johnkerl: @mlin I would do Request Changes but I want to be respectful of timezones. Please make this change as requested before merging. Thanks!

johnkerl: I'm trying this advice on #1128 ... my own advice may have been misspelt ... 👀

johnkerl (changes requested):
I tried the thing I suggested -- which is #1128 

This makes the change run on MacOS _before_ merge to `main` (not after) -- and it fails as at
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/4432210297/jobs/7775996102

Maybe we need to explicitly specify `clang` -- ?

cc @Shelnutt2

mlin:
Well spotted @johnkerl -- I didn't appreciate the minimal/full subtleties. Setting clang explicitly in the matrix [appears to work](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/4433164052/jobs/7777942295). I added some breadcrumbs in the workflow .yml files for the next guy/gal. Please merge this if you see fit & I'm not online. cc @nguyenv @Shelnutt2

nguyenv (approved):

johnkerl (approved):

PR #1129, opened 2023-03-16
[wip] sdist troubleshooting

mlin:
This is not yet working; just posting a few diffs for discussion

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -23,8 +23,8 @@ import sys
 
 import setuptools
 import setuptools.command.build_ext  # FIXME: why was this commented out?
-
 import wheel.bdist_wheel
+
 try:
     from pybind11.setup_helpers import Pybind11Extension
 except ImportError:
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -22,8 +22,8 @@ import subprocess
 import sys
 
 import setuptools
+import setuptools.command.build_ext  # FIXME: why was this commented out?
 
-# import setuptools.command.build_ext
 import wheel.bdist_wheel
 try:
     from pybind11.setup_helpers import Pybind11Extension
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -25,7 +25,12 @@ import setuptools
 
 # import setuptools.command.build_ext
 import wheel.bdist_wheel
-from pybind11.setup_helpers import Pybind11Extension
+try:
+    from pybind11.setup_helpers import Pybind11Extension
+except ImportError:
+    # explanation:
+    # https://pybind11.readthedocs.io/en/stable/compiling.html#classic-setup-requires
+    from setuptools import Extension as Pybind11Extension
 
 this_dir = pathlib.Path(__file__).parent.absolute()
 sys.path.insert(0, str(this_dir))
@@ -164,6 +169,7 @@ class bdist_wheel(wheel.bdist_wheel.bdist_wheel):
 INC_DIRS = [
     "dist_links/libtiledbsoma/include",
     "dist_links/libtiledbsoma/external/include",
+    "../../build/externals/install/include",  # FIXME
 ]
 LIB_DIRS = [
     str(libtiledbsoma_dir / "lib"),
@@ -227,6 +233,7 @@ setuptools.setup(
         )
     ],
     zip_safe=False,
+    setup_requires=["pybind11"],
     install_requires=[
         "anndata",
         "attrs>=22.2",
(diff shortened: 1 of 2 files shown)


apis/python/setup.py
```diff
@@ -22,10 +22,15 @@
 import sys
 
 import setuptools
-
-# import setuptools.command.build_ext
+import setuptools.command.build_ext  # FIXME: why was this commented out?
```
mlin, line 25: This import is+was used below, so I'm not sure why [it was commented out](https://github.com/single-cell-data/TileDB-SOMA/commit/123cb4370057eb47c5657ca2f846336a86e6032e#diff-66aed70e28174320f1db4eaccedcab15d85d14b96fb0c2e9f03417e6fc07df92R26), any feedback?

apis/python/setup.py
```diff
@@ -22,10 +22,15 @@
...
 import wheel.bdist_wheel
-from pybind11.setup_helpers import Pybind11Extension
+
+try:
+    from pybind11.setup_helpers import Pybind11Extension
+except ImportError:
+    # explanation:
+    # https://pybind11.readthedocs.io/en/stable/compiling.html#classic-setup-requires
+    from setuptools import Extension as Pybind11Extension
```
mlin, line 33: This works around a catch-22 where pybind11 is a requirement to load setup.py, yet pip cannot read&install our requirements without loading setup.py.

apis/python/setup.py
```diff
@@ -162,9 +167,9 @@ def run(self):
 
 
 INC_DIRS = [
-    "../../libtiledbsoma/include",
-    "../../libtiledbsoma/external/include",
-    "../../build/externals/install/include",
+    "dist_links/libtiledbsoma/include",
+    "dist_links/libtiledbsoma/external/include",
```
mlin, line 171: `dist_links/libtiledbsoma` is just a symlink to `../../libtiledbsoma` when we're in the git source tree. When we're in an extracted sdist, it's a materialized copy of that directory (resulting from pip following the symlink as it created the sdist tarball). There's other stuff above where we're conditionally using `dist_links/libtiledbsoma` or `../../libtiledbsoma` that can probably be changed to use the symlink unconditionally -- I think that may have been my original oversight.

apis/python/setup.py
```diff
@@ -228,6 +233,7 @@ def run(self):
         )
     ],
     zip_safe=False,
+    setup_requires=["pybind11"],
```
mlin, line 236: second part of the pybind11 workaround

apis/python/setup.py
```diff
@@ -162,9 +167,9 @@ def run(self):
 
 
 INC_DIRS = [
-    "../../libtiledbsoma/include",
-    "../../libtiledbsoma/external/include",
-    "../../build/externals/install/include",
+    "dist_links/libtiledbsoma/include",
+    "dist_links/libtiledbsoma/external/include",
+    "../../build/externals/install/include",  # FIXME
```
mlin, line 172: Here's where I got stuck -- this CMake build directory isn't bundled into the sdist tarball, and I think it'd be weird to do so. While building from the sdist, CMake should generate a similar `build` directory, just in some location other than `../..`. We can figure out what location to expect it in that case and branch this accordingly.

However, I have to say reaching into CMake's stuff from here feels a little weird to me, and I wish I understood better why it's needed now [but wasn't previously](https://github.com/single-cell-data/TileDB-SOMA/commit/123cb4370057eb47c5657ca2f846336a86e6032e#diff-66aed70e28174320f1db4eaccedcab15d85d14b96fb0c2e9f03417e6fc07df92L196)...

mlin (commented):
@nguyenv @johnkerl @Shelnutt2 sorry I didn't look over #1043 carefully before. I made some partial progress in this diff, with a number of comments & questions inline -- all up for discussion

apis/python/setup.py
nguyenv: I thought with `import setuptools` above we could just call `setuptools.command.build_ext.build_ext` below without actually needing to also `import setuptools.command.build_ext`.

apis/python/setup.py
nguyenv: We moved the Pybind11 code from `libtiledbsoma/src/pyapi` into `apis/python/src/libtiledbsoma` so that the Pybind11 bindings reside with the Python code instead of with the C++ code. When I was working on the conda builds, there were issues with `import tiledbsoma` not being able to find the `pytiledbsoma.so` Pybind11 generated shared library. We decided use a working model (TileDB-Py) and reorganized the code to get the conda builds working.

johnkerl:
@mlin can we close this?

PR #1131, opened 2023-03-16
[r] Limit pkgdown runs to releases and manual triggers

eddelbuettel:
**Issue and/or context:**

Every push or pull request currently triggers `pkgdown` rebuilds which is more than is needed.  

**Changes:**

Limits triggers to releases and manual calls (aka `workflow_dispatch` see [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch))

**Notes for Reviewer:**

[SC 26530](https://app.shortcut.com/tiledb-inc/story/26530/tiledbsoma-ci-burns-too-many-cycles-for-pkgdown)


johnkerl (approved):

johnkerl:
Thanks @eddelbuettel !!

PR #1134, opened 2023-03-16
[python] Fix wheel builds given recent Conda-related changes

johnkerl:
Recent changes resulted in wheel-build failures. This PR takes input from @nguyenv, @mlin, and myself to solve this problem.

Note this is a neaten of the exploratory PR #1113. We do not run wheel builds on every PR, pre-merge or post-merge -- only on releases, and in a nightly cron. #1113 contains the fixes here, and it also has a dev-only tweak to run wheel builds pre-merge. That is how this current PR was validated.

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -165,13 +165,21 @@ INC_DIRS = [
     "../../libtiledbsoma/include",
     "../../libtiledbsoma/external/include",
     "../../build/externals/install/include",
+    str(libtiledbsoma_dir / "include"),
+    str(
+        "./src/tiledbsoma"
+    ),  # since pytiledbsoma.cc does #include of query_condition.cc
+    str(libtiledbsoma_dir.parent / "build/externals/install/include"),
 ]
+
 LIB_DIRS = [
     str(libtiledbsoma_dir / "lib"),
 ]
 CXX_FLAGS = [
     f'-Wl,-rpath,{str(libtiledbsoma_dir / "lib")}',
 ]
+if sys.platform == "darwin":
+    CXX_FLAGS.append("-mmacosx-version-min=10.14")
 
 if os.name == "posix" and sys.platform != "darwin":
     LIB_DIRS.append(str(libtiledbsoma_dir / "lib" / "x86_64-linux-gnu"))
@@ -223,7 +231,7 @@ setuptools.setup(
             library_dirs=LIB_DIRS,
             libraries=["tiledbsoma"],
             extra_link_args=CXX_FLAGS,
-            extra_compile_args=["-std=c++17"],
+            extra_compile_args=["-std=c++17"] + CXX_FLAGS,
             language="c++",
         )
     ],
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -22,10 +22,17 @@ import subprocess
 import sys
 
 import setuptools
-
-# import setuptools.command.build_ext
+import setuptools.command.build_ext
 import wheel.bdist_wheel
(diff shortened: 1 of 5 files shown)


apis/python/setup.py
```diff
@@ -165,13 +165,21 @@ def run(self):
     "../../libtiledbsoma/include",
     "../../libtiledbsoma/external/include",
     "../../build/externals/install/include",
```
nguyenv, line 167: https://github.com/single-cell-data/TileDB-SOMA/pull/1129#discussion_r1138145214

johnkerl: done -- thanks! :)

johnkerl:
ok @mlin @nguyenv I made the suggestions @nguyenv recommended from @mlin 's #1129 

those same mods are also in #1113 which is run wheel-builds pre-merge (unlike this PR which, as  noted in its descrption, doesn't and shouldn't)

johnkerl:
Here is the diff between #1113 and #1134 (this PR) proving that since #1113 passes CI including building wheels, it's safe to merge this PR which won't build wheels until nightly cron and release tags.

```
$ git diff -r kerl/sdist-check kerl/fix-wheel-builds
diff --git a/.github/workflows/python-packaging.yml b/.github/workflows/python-packaging.yml
index 79956376..042589c1 100644
--- a/.github/workflows/python-packaging.yml
+++ b/.github/workflows/python-packaging.yml
@@ -4,8 +4,8 @@ on:
   # I.e. on request
   workflow_dispatch:
   # Normally the smoke test only runs on merge to main.  To run it on a PR CI, uncomment `pull_request:`
-  pull_request:
-    branches: [main]
+  # pull_request:
+  #   branches: [main]
   # paths-ignore:
   #   - 'apis/r/**'
   release:
diff --git a/apis/python/setup.py b/apis/python/setup.py
index 033eec09..9ff4c717 100644
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -185,13 +185,7 @@ LIB_DIRS = [
 CXX_FLAGS = [
     f'-Wl,-rpath,{str(libtiledbsoma_dir / "lib")}',
 ]
-print()
-print()
-print("================================================================")
-print("SYS.PLATFORM", sys.platform)
-print("CXX_FLAGS BEFORE", CXX_FLAGS)
 if sys.platform == "darwin":
-    # COMPILE_FLAGS.append("-mmacosx-version-min=10.14")
     CXX_FLAGS.append("-mmacosx-version-min=10.14")

 if os.name == "posix" and sys.platform != "darwin":
@@ -202,10 +196,6 @@ if os.name == "posix" and sys.platform != "darwin":
     )
     CXX_FLAGS.append(f'-Wl,-rpath,{str(libtiledbsoma_dir / "lib64")}')

-print("CXX_FLAGS AFTER", CXX_FLAGS)
-print()
-print()
-
 # Don't use `if __name__ == "__main__":` as the `python_requires` must
 # be at top level, outside any if-block
@@ -248,7 +238,6 @@ setuptools.setup(
             library_dirs=LIB_DIRS,
             libraries=["tiledbsoma"],
             extra_link_args=CXX_FLAGS,
-            # extra_compile_args=["-std=c++17"],
             extra_compile_args=["-std=c++17"] + CXX_FLAGS,
             language="c++",
         )
```

nguyenv:
https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/6

I've tested the conda builds with the latest changes applied to `setup.py` in this PR and all are still passing.

nguyenv (approved):

Issue #1135, opened 2023-03-16
Reduce peak memory usage of to_anndata()


atolopko-czi:
**Is your feature request related to a problem? Please describe.**

Using `tiledbsoma.io.ingest.to_anndata()` to load data should have similar peak memory requirements to using [AnnData.read_h5ad()](https://anndata.readthedocs.io/en/latest/generated/anndata.read_h5ad.html).

In one test, loading a [~6GB H5AD file](https://cellxgene.cziscience.com/collections/6f6d381a-7701-4781-935c-db10d30de293) on an `i3.xlarge` (30.5GB) EC2 works fine, consuming 18GB of memory at peak, and 9GB upon return. But loading the equivalent data from a TileDB-SOMA collection does not complete on the instance (the Python kernel dies); it works fine on a larger memory instance (`r4.4xlarge`, ~128GB)

**Describe the solution you'd like**

TileDB-SOMA's to_anndata() implementation should have similar peak and final memory usage as AnnData.read_h5ad().

**Describe alternatives you've considered**

Don't use AnnData at all. ;-) 

**Additional context**

None

PR #1136, opened 2023-03-16
[python] Allow `to_anndata` when `feature_id` not `var_id` is in `var` columns

johnkerl:
**Issue and/or context:** In support of #1135, as not all SOMA experiments have a `var` with a column named `var_id` -- it may be `feature_id`, or something else.

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -1056,7 +1056,10 @@ def to_anndata(
 
     var_df = measurement.var.read().concat().to_pandas()
     var_df.drop([SOMA_JOINID], axis=1, inplace=True)
-    var_df.set_index("var_id", inplace=True)
+    if "var_id" in var_df.keys():
+        var_df.set_index("var_id", inplace=True)
+    elif "feature_id" in var_df.keys():
+        var_df.set_index("feature_id", inplace=True)
 
     nobs = len(obs_df.index)
     nvar = len(var_df.index)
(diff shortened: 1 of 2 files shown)


atolopko-czi:
> Of course, the var column we need might be something other than var_id or feature_id.
>
> @pablo-gar @aaronwolen @atolopko-czi how best to handle?

Any reason not to just add a new parameter for it? If it defaults to `None`, the hardcoded `var_id` value could be used, for backward compatibility.

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1031,7 +1031,12 @@ def to_h5ad(
 
 # ----------------------------------------------------------------
 def to_anndata(
-    experiment: Experiment, measurement_name: str, X_layer_name: str = "data"
+    experiment: Experiment,
+    measurement_name: str,
+    *,
+    obs_id_name: str = "obs_id",
+    var_id_name: str = "var_id",
```
atolopko-czi, line 1038: In general, for any arbitrary SOMA collection/experiment, these will _not_ work, right? If that's the case, maybe the default should just be `soma_joinid`? Or make it a required param, with no default.

johnkerl: No, these are _distinct_ from `soma_joinid`.

The `soma_joinid` column is integers.

What we need here is strings.

atolopko-czi: OK, fair enough!  Still, if obs/var_id is not generally going to work, should we make it a require param w/o a default?

johnkerl: @atolopko-czi  I'd rather keep it the default as it's often enough correct (for anything ingested from AnnData, which is very often the case) -- and people can override the default when it's not the case

atolopko-czi (approved):
LGTM!

Issue #435, opened 2022-10-17
[r/python] Build a Python/R round-trip unit test framework


johnkerl:
Goal is two-fold:

* Write data from R; read it from Python
* Write data from Python; read it from R

johnkerl:
Update: let's not do this in the current sprint, as R classes for the `main` branch are still being built out.

ebezzi:
What we have now is:
1. A framework that can be used to do roundtrip testing (in `apis/system/test`)
2. A set of tests for: `DataFrame`, `SparseNDArray`, `DenseNDArray`
We'll create another ticket for an end to end test that uses a real dataset. Marking this as done.

PR #1139, opened 2023-03-17
R-Python roundtrip testing

ebezzi:
**Issue and/or context:**
https://github.com/single-cell-data/TileDB-SOMA/issues/435

Note: I decided against using libraries for R-Python integration (e.g. [rpy2](https://rpy2.github.io/), or [Reticulate](https://rstudio.github.io/reticulate/)). The rationale is that such libraries will convert R types to Python and vice versa using their own mechanism, and this could introduce bugs that aren't specific to the SOMA code, or even worse, conceal bugs and/or schema differences.

This approach creates an R script directly in Python, adds assertions using `testthat` and executes it. Failed assertions will be caught by Python directly. This has the advantage of being an embedded script and can assert against the original object directly, so that test cases do not need to be copied over to R.

TODOs:
1. Add test cases for all the other types
2. Set up the R environment. Right now this assumes R is installed with all the required dependencies. R can install dependencies in the code, but it's slow so the R environment should be prepared by (e.g.) a GHA step. Alternatively we could run these in a Dockerized environment.
3. Migrate this to pytest. Even if some features won't be useful, we can still benefit from it (tmp_path fixture, run it with the other tests, etc)
4. Write the opposite direction. This should be straightforward and can still be done in Python, to keep a standard.

**Changes:**
- Add a prototype file to write a dataframe in Python, read it in R and assert on it.

**Notes for Reviewer:**
- This is a draft - I want to validate the approach with the team before moving to the rest of TODOs.


R-Python-Testing/write_python_read_r.py
```diff
@@ -0,0 +1,75 @@
...
+table = soma_df$read()
+df = as.data.frame(table)
+"""
+
+
+def r_assert(code: str):
+    R_script = R_script_base + code
+    with open("test-dataframe-read.R", "w") as f:
+        f.write(R_script)
+    subprocess.run(["Rscript", "test-dataframe-read.R"])
```
bkmartinjr, line 52: I don't understand where the assertion failure is caught.  If I read this right, the Rscript process will return a non-zero exit code, but it will be ignored by `subprocess.run` because it doesn't have the `check` param set to True, AND/OR there is no check for non-zero exit code.

I was imagining that the `r_assert` would raise an exception (like `assert`) when the R script fails.

So maybe this line should be:

> assert subprocess.run(["Rscript", "test-dataframe-read.R"]).returncode == 0

or something like that?

R-Python-Testing/write_python_read_r.py
```diff
@@ -0,0 +1,75 @@
+import shutil
```
bkmartinjr, line 1: when you consider that we will need a bunch of these tests, how will the individual tests be organized and dispatched?

One possibility:  just make all of this a set of stand-alone pytest tests, and use that framework for the very top-level dispatching, logging and error handling?

bkmartinjr (commented):
repo organization: where do these API round-trip tests belong?  I don't think the top-level of the repo is right, as they are tests of the contents of the `api` subdir.

Maybe put them into `api/system/tests/` or something like that?

bkmartinjr:
General comments, with the caveat that I"m not very multi-lingual:
* using a Python framework as the orchestration tier seems OK to me. 
* if the R code gets complex, it could also be pushed out into stand-alone files (i.e., there is no reason it has to be exclusively in-line strings, although that is also convenient for small tests)
* good time to consider repo organization, and I'm not sure a top-level folder is right given that this focuses on the contents of `api`
* you need a test runner at the top level.  If Python is the top-level language, why not just use pytest?

R-Python-Testing/write_python_read_r.py
ebezzi: My original idea was to make this a top-level module and create several pytest files (one for each data type).

ebezzi:
> repo organization: where do these API round-trip tests belong? I don't think the top-level of the repo is right, as they are tests of the contents of the `api` subdir.
> 
> Maybe put them into `api/system/tests/` or something like that?

`api/system/tests/` looks like a great suggestion, thanks!

ebezzi:
@bkmartinjr I implemented most of your suggestions:
1. Moved the test suite to `api/system/tests`. Will need to check with @johnkerl to see if this makes sense as well
2. Implemented the test runner using pytest
3. Moved the common code to a common.py and switched to a class approach. This makes it easier since different test suites (for different SOMA objects) can inherit from it and re-use the common methods (like `r_assert`). This can be done using modules as well but it will require passing around more arguments.
4. Added the missing assertion error on subprocess failure.

apis/system/tests/common.py
```diff
@@ -0,0 +1,39 @@
+import subprocess
+
+
+def to_R(x):
```
johnkerl, line 4: sorry, _what_ is  being `to_R`'ed? i _think_ this is something that convertions python booleans (and maybe other things) to R string -- please let's either have the function name say something about what it does, and/or a doc-block

apis/system/tests/common.py
```diff
@@ -0,0 +1,39 @@
...
+    if isinstance(x, bool):
+        return str(x).upper()
+    elif isinstance(x, str):
+        return f'"{x}"'
+    else:
+        return str(x)
+
+
+def create_R_list(xs):
```
johnkerl, line 13: ditto

apis/system/tests/common.py
```diff
@@ -0,0 +1,39 @@
...
+    return ",".join([to_R(x) for x in xs])
+
+
+class TestWritePythonReadR:
+    @classmethod
+    def setup_class(cls):
+        import tempfile
+
+        cls.td = tempfile.TemporaryDirectory()
```
johnkerl, line 22: please have `cls.td` -> `cls.tempdir` throughout. much easier to read/understand for everyone after you.

apis/system/tests/write_python_read_r.py
```diff
@@ -0,0 +1,58 @@
...
+        rb = pa.Table.from_pydict(pydict)
+        with soma.DataFrame.open(self.uri, "w") as sdf:
+            sdf.write(rb)
+
+        df = rb.to_pandas()  # Alternatively, we could re-open the dataframe
+        return df
+
+    # Prepares an R script with the dependencies and loads the data.frame in `df`.
+    # Future assertions can be done by appending to this script
+    def base_script(self):
```
johnkerl, line 38: maybe `base_R_script`?

johnkerl (approved):
lgtm -- with some comments

apis/system/tests/common.py
ebezzi: Basically this is to inject a Python object in the auto-generated R script. If we want to be explicit, this could be `embed_into_R_code` or something like that.

bkmartinjr (commented):
In the spirit of having a skinny end-to-end demonstrator, is it worth also including the following in this PR?
* GHA workflow to run the tests (including dependency install, etc)
* an simple example of "create in R, read in Py"? (to complement the "create in Py, read in R" example)

If these are high confidence, then I am OK if you push to another PR. My only motivation for this suggestion is to flesh out any additional issues that might arise early on....

ebezzi:
> In the spirit of having a skinny end-to-end demonstrator, is it worth also including the following in this PR?
> 
> * GHA workflow to run the tests (including dependency install, etc)
> * an simple example of "create in R, read in Py"? (to complement the "create in Py, read in R" example)

I am planning to work on both. The GHA workflow might be a bit trickier due to how R handles dependencies, but it's something we need, so I'll work on it.

Do you think it's worth working on those two items before adding the other SOMA objects to this PR?

bkmartinjr:
> Do you think it's worth working on those two items before adding the other SOMA objects to this PR?

I think you can make the call.  That said, I'd probably bias toward getting the full framework set up with a single trivial test, which will make it easy to progressively add more tests (and have other folks help).  But either works for me.   @johnkerl if you have a preference, let it be known :-)

bkmartinjr (approved):
marking as approved.  If you do elect to extend this PR with more framework (eg, the GHA or full round-trip), just hit me up for an additional review.

apis/system/tests/write_r_read_python.py
```diff
@@ -0,0 +1,55 @@
...
+
+from .common import TestReadPythonWriteR
+
+
+class TestDataframeWriteRReadPython(TestReadPythonWriteR):
+    @pytest.fixture(scope="class")
+    def R_dataframe(self):
+        base_script = f"""
+        require("tiledbsoma")
+        require("arrow")
```
eddelbuettel, line 15: library() is preferred is over required().  Both work.

apis/system/tests/write_r_read_python.py
```diff
@@ -0,0 +1,55 @@
...
+        df_schema <- schema(
+        field("foo", int32()),
+        field("bar", float64()),
+        field("baz", string()),
+        field("quux", bool())
+        )
+
+        platform_config <- NULL
+        tiledbsoma_ctx <- NULL
+        sdf <- SOMADataFrameCreate("{self.uri}", df_schema, c("foo"), platform_config, tiledbsoma_ctx)
```
eddelbuettel, line 26: Because NULL is the default argyment these three lines can be written just one shorter line:

```suggestion
        sdf <- SOMADataFrameCreate("{self.uri}", df_schema, "foo")
```

ebezzi:
@johnkerl @bkmartinjr I upgraded this to a non-draft PR. As per recommendations, this PR now contains:
1. A "write R, read Python" test (for dataframes). This was particularly useful as this direction highlighted some issues that we addressed [here](https://github.com/single-cell-data/TileDB-SOMA/pull/1149).
2. A [GHA pipeline](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/4491463305/jobs/7899954161?pr=1139) that runs all the tests. This takes about 10 minutes and might slow down PRs, so we might decide to only run it more sparingly, but I think for now we should keep it.
3. An additional test that verifies that the TileDB engine library is the same on both languages (R and Python). This will force updates to TileDB to happen at the same time on R and Python, which is generally a good thing.

Please let me know if this looks good. If it does, I'll go ahead and merge the PR and finish the tests for all the data types (at minimum, DenseNDArray and SparseNDArray should be covered, in my opinion) in a separate PR.

@eddelbuettel if you have time, I'd appreciate additional feedback on the R code snippets and on the CI. I extracted some of the instructions from the `r-ci.sh` file, as I couldn't see an easy way to re-use that file for these ad-hoc tests, but I am happy to make changes if needed. Thanks!

eddelbuettel (commented):
I added a few more comments.

ebezzi:
> I am officially torn here. I admire the fine and patient work @ebezzi has done, but I cannot shake a feeling that it sets up 'too much' infrastructure for the relatively few comparison predicates we are running.
> 
> I think it would be worth our while to consider alternative. Drive it from Python via `rpy2` (and I am fairly good friends with the author, we could get help). Drive it from R and use `reticulate` (ditto for some of the early authors and creators). Or take advantage of the fact that we have neutral middleware in Arrow and drive from one of the two comparing Arrow data from the two clients.
> 
> But I am torn in a 'having tests is better than no tests' so I won't hold this up with requests for changes, but I also feel I cannot quite endorse it right so I may just recuse myself from this one. I am sure we will come to a fine decision as a team.

I fully understand your concerns. The reasons I did decide not to go with `rpy2` are three:
1. It introduces a new dependency (even if test-only)
2. Some of the conversions that those library generally operate might conceal or introduce new bugs. I feel like asserting in the target language directly is the most powerful option, as it bypasses any middleware.
3. In my experience (even if I never used `rpy2` and only marginally `reticulate`), those libraries aren't straightforward to set up and they come with their hurdles. I am pretty confident that using it would have taken me more time than setting up the Python classes I did here.

That said, I am more than open to try the alternative approach, maybe in a separate PR, and see how it looks. Most of the work here was me learning the SOMA R API and setting up the GHA workflow, which won't require any changes (except maybe to add another dependency).

.github/workflows/r-python-interop-testing.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Copy libtiledbsoma
+        ## we build 'as in an R package' with all sources within apis/r/
+        run: cd apis/r && ./copy_source.sh
+
+      - name: Install R-tiledbsoma
+        run: |
+          cd apis/r
+          R CMD build --no-build-vignettes --no-manual .
+          FILE=$(ls -1t *.tar.gz | head -n 1)
+          R CMD INSTALL $FILE
```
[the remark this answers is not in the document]
ebezzi: I think compiling from source gives the extra safety here: if an issue is found that causes a test to fail, the fix can be done and diagnosed within the PR. That said, we can reconsider if/when we decide to move this test suite to a nightly run. 

The `r-ci.sh` change would be useful, but only if you have time!

apis/system/tests/test_dataframe_write_python_read_r.py
```diff
@@ -0,0 +1,56 @@
...
+
+        df = rb.to_pandas()  # Alternatively, we could re-open the dataframe
+        return df
+
+    # Prepares an R script with the dependencies and loads the data.frame in `df`.
+    # Future assertions can be done by appending to this script
+    def base_R_script(self):
+        return f"""
+        library("tiledbsoma")
+        library("testthat")
```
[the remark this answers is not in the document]
ebezzi: `stopifnot()` is exactly what I was looking for when I found `testthat` - thanks!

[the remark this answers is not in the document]
eddelbuettel: A somewhat more recent (one or two major releases) changes in R was to extend `stopifnot` with a clear message. We use that a lot across the packages as it is like a nice informative assert on entry.  I.e. picking the first match from tiledb-r I see

```r
tiledb_version <- function(compact = FALSE) {
    stopifnot(`Argument 'compact' must be logical` = is.logical(compact))
```

and the only change I'd make today is reflecting that I now know that standard " are fine, no need for backticks.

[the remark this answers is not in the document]
ebezzi: Ah, that's great, I will certainly add it. I will also make sure that the R output gets captured correctly in the test runner. I am now working on the tests on `DenseNDArray` so I will incorporate the change in there. Thanks again!

Issue #1111, opened 2023-03-15
[r] Non-iterated `soma_reader` should not return partial results


mlin:
In `soma_reader()` we've been commanded to read the entire array at once. Currently we take one sip of the straw, and if that's only a partial read, we print a warning but proceed:

https://github.com/single-cell-data/TileDB-SOMA/blob/123cb4370057eb47c5657ca2f846336a86e6032e/apis/r/src/rinterface.cpp#L144-L152

I submit that it's the wrong behavior to return a partial array here, regardless of printing a warning (which might get lost in a long program output). We should do whatever's necessary to read the whole array; if we run out of memory, that's on the caller for having commanded the non-iterated read.

johnkerl:
@eddelbuettel you may have already done some work on this? Sorry for the delay in task assignment ...

eddelbuettel:
I have not and it is not entirely clear what can be done because SOMA(Array)Reader is not fully fleshed out here.

Core resizes and retries. SOMA(Array)Reader does not.  I am not sure what else can be done here _with the current C++ stack_.  We could rewrite our entries in terms of the iterated reader and iterated as needed -- that is about all I can think of.

eddelbuettel:
The old / basic / first pass iterator I wrote does that (modulo not having native iterators ...).  See eg the old test in `test_SOMAReader-Iterated.R`.  Oh, those test files are more candidates for the grande olde renaming part III....

johnkerl:
In Python, Pandas and Arrows have a `.concat` -- is there anything similar we can do in R with `res1` and `res2`?

johnkerl:
@eddelbuettel sorry I didn't read carefully enough.

What I meant was is there some `concat` something for the elements of `rl`. Maybe `rbind` is that thing?

johnkerl:
While in Python we _syntactically_ utter `foo.read().tables().concat()`, I am not asking here for that _syntax_. Only for some _semantic_ way (with R-appropriate syntax) to concatenate the read-batches.

eddelbuettel:
Yes rbind is that thing.  The canonical (old school) use is   `do.call()`  to sweep and operation over a list so here `do.call(rbind, result_list)`.   There various improvements, I happen to like  `dt <- rbindlist(list_of_dts)` which scans the list of `data.table` objects first, computes total sizes, allocates once and injects. Faster and more efficient than the piece-by-piece approach.

johnkerl:
[...]

PR #1146, opened 2023-03-20
[r] Add Seurat -> SOMA Ingestor

### LINKED ISSUE #942: [r] Implement `from_seurat`
Context: #839 

Requires #938 as a prerequisite

mojaveazure:
Add support for creating SOMA experiments from `Seurat` objects; all done using SeuratObject as a suggested dependency

New function:
- `write_soma()`: generic function that takes an object `x`, a URI, and configuration and generates a complete SOMA experiment from it

Implemented SOMA methods:
- `write_soma.Seurat()`: create a SOMA experiment from a `Seurat` object
- `write_soma.Assay()`, `write_soma.DimReduc()`, `write_soma.Graph()`: helpers to write Seurat sub-objects to a SOMA experiment
- `write_soma.data.frame()`: helper to write a `data.frame` into a SOMA experiment
- `write_soma.matrix()`, `write_soma.Matrix()`, `write_soma.TsparseMatrix()`: helpers to write various matrix formats into a SOMA experiment

Future work (tracking in #1192):

resolves #942


apis/r/R/utils-uris.R
```diff
@@ -2,6 +2,75 @@ is_remote_uri <- function(x) {
...
+    user = tools::R_user_dir(package = pkg, which = 'data'),
+    rappdirs = rappdirs::user_data_dir(
+      appname = pkg,
+      appauthor = 'tiledb',
+      version = as.character(tiledb::tiledb_version(TRUE)),
+      roaming = getOption('tiledbsoma.io.user_dir.roaming', default = FALSE)
+    )
+  ))
+}
+
```
eddelbuettel, line 73: This is fine, but to me a needless overcomplication to me. Just stick with R >= 4.0 and assume access to base R user directory functionality.  Saves everybody one more dependency _which is now imposed on newer R versions too_.

apis/r/R/utils.R
```diff
@@ -71,6 +76,81 @@ as_arrow_table <- function(arrlst) {
...
+#' @keywords internal
+#'
+#' @noRd
+#'
+r_version <- function(repr = c('v', 'c')) {
+  repr <- repr[1L]
+  repr <- match.arg(arg = repr)
+  version <- paste(R.Version()[c('major', 'minor')], collapse = '.')
+  return(switch(EXPR = repr, v = package_version(version), version))
+}
```
eddelbuettel, line 152: I don't think you need this either:

```r
> getRversion() >= "4.0.0"
[1] TRUE
> getRversion() < "4.3.0"
[1] TRUE
> 
```

The S3 class `package_version` with its <, <=, ==, >, >= operators is a marvel.

eddelbuettel (approved):
Looks like a huge step in the right direction.

I have some small niggles on style, give them a minute of though.  Not vetoing anything over them though.

eddelbuettel (approved):
Looks god.  The padding step is to go from the "tighter" Seurat / sparse representation to something faithful to SOMA and its shape parameters?

apis/r/R/utils-uris.R
```diff
@@ -31,7 +31,7 @@ user_dir <- function(type = getOption('tiledbsoma.io.user_dir')) {
     'working',
     'tempdir',
     ifelse(
-      test = requireNamespace('tools', quietly = TRUE) && r_version() > '4.0.0',
+      test = requireNamespace('tools', quietly = TRUE) && getRversion() > '4.0.0',
```
eddelbuettel, line 34: :+1:

eddelbuettel (approved):

mojaveazure:
> Looks good. The padding step is to go from the "tighter" Seurat / sparse representation to something faithful to SOMA and its shape parameters?

Yes. In Seurat, not all matrices (namely `scale.data`) have to have the same `var` whereas SOMA requires each `X` layer to have the same `var` and `obs`; `pad_matrix()` allows us to expand those matrices to the full domain `[var, obs]`

eddelbuettel (approved):
:+1:

pablo-gar (approved):
I did some testing for `write_soma.Seurat()`

I created 5 different `SOMAExperiement` with different datasets from [Tabula Muris Senis](https://cellxgene.cziscience.com/collections/0b9d8a04-bb9d-44da-aa27-705bb65b54eb)

And then performed equality checks between the following SOMA parts and their corresponding equivalents from the Seurat object:

- `obs`
- `var`
- `obsm["X_pca"]`
- `obsm["X_tsne"]`
- `obsm["X_umap"]`
- `X["counts"]` (only the first 100 cells due to #1111)
- `X["data"]`  (only the first 100 cells due to #1111)

All data was created properly into the `SOMAExperiement` and all equality checks passed!

-------------

R code used for this

```r
library("tiledbsoma")
library("Seurat")

stopifnot_df_equal <- function(df1, df2) {
  
  df1 <- as.data.frame(df1)
  df2 <- as.data.frame(df2)
  
  stopifnot(nrow(df1) == nrow(df2))

  shared_cols <- intersect(colnames(df1), colnames(df2))
  
  df1 <- df1[,shared_cols]
  df2 <- df2[,shared_cols]
  
  bool_matrix <- df1 != df2
  bool_matrix[is.na(bool_matrix)] <- F 
  
  stopifnot(sum(colSums(bool_matrix)) == 0) 
  
}

# Get Seurat objects 
seurat_object <- readRDS("local.rds") # file from datasets here https://cellxgene.cziscience.com/collections/0b9d8a04-bb9d-44da-aa27-705bb65b54eb
seurat_obs <- seurat_object@meta.data
seurat_var <- seurat_object@assays$RNA@meta.features
seurat_pca <- seurat_object@reductions$pca@cell.embeddings
seurat_umap <- seurat_object@reductions$umap@cell.embeddings
seurat_tsne <- seurat_object@reductions$tsne@cell.embeddings
seurat_x_counts <- seurat_object@assays$RNA@counts
seurat_x_data <- seurat_object@assays$RNA@data

# Make soma object
unlink("from_seurat_soma", recursive = T)
write_soma(seurat_object, "from_seurat_soma")

# Get soma objects
soma_seurat <- SOMAExperimentOpen("from_seurat_soma/")

soma_obs <- as.data.frame(soma_seurat$get("obs")$read())
soma_var <- as.data.frame(soma_seurat$get("ms")$get("RNA")$get("var")$read())
soma_pca <- soma_seurat$get("ms")$get("RNA")$get("obsm")$get("X_pca")$read_dense_matrix()
soma_umap <- soma_seurat$get("ms")$get("RNA")$get("obsm")$get("X_umap")$read_dense_matrix()
soma_tsne <- soma_seurat$get("ms")$get("RNA")$get("obsm")$get("X_tsne")$read_dense_matrix()
soma_x_counts <- soma_seurat$get("ms")$get("RNA")$get("X")$get("counts")$read_sparse_matrix()
soma_x_data <- soma_seurat$get("ms")$get("RNA")$get("X")$get("data")$read_sparse_matrix()

# Perform equality comparisons
stopifnot_df_equal(seurat_obs, soma_obs)
stopifnot_df_equal(seurat_var, soma_var)
stopifnot(seurat_pca == soma_pca)
stopifnot(seurat_umap == soma_umap)
stopifnot(seurat_tsne == soma_tsne)

for (i in 1:100) {
  stopifnot(seurat_x_counts[,i] == soma_x_counts[i,])
  stopifnot(seurat_x_data[,i] == soma_x_data[i,])
}
```

eddelbuettel:
Nice -- regarding #1111 please see #1194 and this idiom it uses (but which is already supported, you should be able to use it now):

```r
## pass with budget of 45mb
ctx <- tiledbsoma::SOMATileDBContext$new(c(soma.init_buffer_bytes="45000000"))
sdf2 <- SOMADataFrameOpen(uri, tiledbsoma_ctx = ctx)
expect_silent(sdf2$read())
```

where `uri` is the path to the data set that fails with the smaller default budget.

aaronwolen (changes requested):
This is great and fills a gaping hole in our functionality. I made a few comments inline but there's a larger issue I think we should address: I'm just realizing we're currently missing R6 methods for adding new dataframes and ndarrays to a SOMACollection as described in the [spec](https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md#somacollection-1) (e.g., `add_new_dataframe()`). This is functionality that you've already implemented in this PR so let's use that code to add these missing R6 methods, which `write_soma()` and friends can invoke.

aaronwolen (changes requested):
This is great and fills a gaping hole in our functionality. I made a few comments inline but there's a larger issue I think we should address: I'm just realizing we're currently missing R6 methods for adding new dataframes and ndarrays to a SOMACollection as described in the [spec](https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md#somacollection-1) (e.g., `add_new_dataframe()`). This is functionality that you've already implemented in this PR so let's use that code to add these missing R6 methods, which `write_soma()` and friends can invoke.

apis/r/R/utils-uris.R
aaronwolen: I agree with @eddelbuettel. I'm not sure this is needed.

apis/r/R/write_soma.R
```diff
@@ -0,0 +1,375 @@
...
+#' The array type for each column will be determined by
+#' \code{\link[arrow:infer_type]{arrow::infer_type}()}
+#'
+#' @method write_soma data.frame
+#' @export
+#'
+write_soma.data.frame <- function(
+  x,
+  uri,
+  soma,
```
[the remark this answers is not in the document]
mojaveazure: Changed to `soma_parent`

apis/r/R/write_seurat.R
```diff
@@ -0,0 +1,547 @@
...
+#' the \code{\link[tiledbsoma:SOMAExperiment]{experiment}} level
+#'
+#' @inherit write_soma_seurat_sub_objects sections
+#'
+#' @method write_soma Seurat
+#' @export
+#'
+write_soma.Seurat <- function(
+  x,
+  uri = NULL,
```
aaronwolen, line 380: We require an explicit URI everywhere else in SOMA and I don't think we should make an exception here. I def appreciate you're trying to make it easier for users but auto-selecting a location based on an option would be a new feature and something we should discuss as a group, or in a separate PR.

apis/r/R/write_seurat.R
```diff
@@ -0,0 +1,547 @@
...
+    "'uri' must be a single character value" = is.null(uri) ||
+      is_scalar_character(uri)
+  )
+  uri <- uri %||% file_path(user_dir(), SeuratObject::Project(x))
+  if (!is_remote_uri(uri)) {
+    if (isTRUE(overwrite)) {
+      unlink(x = uri, recursive = TRUE, force = TRUE)
+    }
+    dir.create(dirname(uri), recursive = TRUE)
+  }
```
aaronwolen, line 397: This will silently skip the deletion step with remote URIs. I don't think we want to implement anything that results in different behavior for local vs remote locations. Adding an overwrite option is also something we should discuss to make sure we're all on the same page and we implement it in a consistent way across the codebase.

apis/r/R/write_seurat.R
johnkerl: 💯

apis/r/R/write_seurat.R
johnkerl: Correction:

> have differences _only when there are language-specific or toolchain-specific needs to do so_

apis/r/R/utils-uris.R
johnkerl: looks like the conversation around this was resolved but this function hasn't been removed yet. let's get that action item closed out please.

apis/r/R/write_seurat.R
```diff
@@ -0,0 +1,539 @@
...
+#' Expression matrices are transposed (cells as rows) prior to writing. All
+#' other slots, including results from extended assays (eg. \code{SCTAssay},
+#' \code{ChromatinAssay}) are lost
+#'
+#' @method write_soma Assay
+#' @export
+#'
+write_soma.Assay <- function(
+  x,
+  uri = NULL,
```
johnkerl, line 57: @aaronwolen @mojaveazure if I understand, this one is OK & necessary -- correct?

aaronwolen (changes requested):
Just noticed there was a force push mid-review so I'm going to submit now and take a look at the changes.

apis/r/R/write_seurat.R
```diff
@@ -0,0 +1,547 @@
...
+    if (isTRUE(overwrite)) {
+      unlink(x = uri, recursive = TRUE, force = TRUE)
+    }
+    dir.create(dirname(uri), recursive = TRUE)
+  }
+  experiment <- SOMAExperimentCreate(
+    uri = uri,
+    platform_config = platform_config,
+    tiledbsoma_ctx = tiledbsoma_ctx
+  )
```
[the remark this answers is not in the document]
mojaveazure: There's already an error further if something exists at `uri` provided by TileDB
```R
> uri <- file.path(tempdir(), 'test')
> dir.create(uri, recursive = TRUE)
> exp1 <- SOMAExperimentCreate(uri)
> exp2 <- SOMAExperimentCreate(uri)
Error: [TileDB::StorageManager] Error: Cannot create group; Group 'file:///tmp/Rtmp95o01s/test' already exists
```

aaronwolen (commented):
Again, great work here. There are a few items that need to be addressed  to improve consistency and/or compliance with the spec, then it should be good to go.

aaronwolen (changes requested):
Again, great work here. There are a few items that need to be addressed  to improve consistency and/or compliance with the spec, then it should be good to go.

aaronwolen (commented):
Again, great work here. There are a few items that need to be addressed  to improve consistency and/or compliance with the spec, then it should be good to go.

apis/r/R/write_soma.R
```diff
@@ -0,0 +1,420 @@
...
+#'
+#' @noRd
+#'
+.df_index <- function(
+    x,
+    alt = 'rownames',
+    axis = 'obs',
+    prefix = 'tiledbsoma',
+    ...
+) {
```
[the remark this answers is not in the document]
mojaveazure: This function should also be applicable for eventual SCE &rarr; SOMA to format `colData`/`rowData` for SOMA ingestion

apis/r/R/write_soma.R
```diff
@@ -0,0 +1,420 @@
...
+  # Create a proper URI
+  uri <- .check_soma_uri(
+    uri = uri,
+    soma_parent = soma_parent,
+    absolute = absolute
+  )
+  # Transpose the matrix
+  if (isTRUE(transpose)) {
+    x <- t(x)
+  }
```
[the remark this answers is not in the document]
mojaveazure: Probably not, but it doesn't hurt. In the Seurat world, transposition is often done within the function and triggered by an argument. In the Bioc world, it seems to be that transposition is done with explicit calls to `t`; this function allows both, especially by leaving `transpose = FALSE` by default

aaronwolen (approved):
🚀 

Last thing can you bump the version and update NEWS?

PR #1149, opened 2023-03-21
[r] [python] support bytes metadata field + R SOMA_ENCODING_VERSION

ebezzi:
**Issue and/or context:**
The R-Python interop testing found two issues:
1. Metadata fields (`SOMA_OBJECT_TYPE_METADATA_KEY`, `SOMA_ENCODING_VERSION`) are written using bytes by the R API
2. The `SOMA_ENCODING_VERSION` does not match

**Changes:**
1. Decode the metadata fields to UTF-8 strings if they're bytes
2. Bump R's `SOMA_ENCODING_VERSION` to 1

**Notes for Reviewer:**
Note that R is not currently doing the same metadata checks that Python is doing w.r.t. metadata. Worth adding them ?


eddelbuettel (approved):

johnkerl (approved):
🚢

Issue #390, opened 2022-10-10
[python] Add more user-level control for verbosity in `tiledbsoma.io`


bkmartinjr:
There is code in the core modules that are very verbose with their logging. I suspect the idea is to keep notebook users informed of import/export progress.  But many (most?) use cases are embedding the library into other apps & packages, which do not want logging around progress updates, etc.

Two suggestions:
* make this level of human-centered logging a local or global config (eg `from_anndata(verbose=True)`, so that the level of non-error-related verbosity can be reduced
* move this kind of logging out of the core modules (ie, only do it in the `io` package)

johnkerl:
@bkmartinjr i very, very, very strongly want _minimal_ (more-than-nothing) ingest-logging _on by default_ in notebooks -- i made some effort to reduce the verbosity from what it was (`info` level is now shorter than `debug` -- thanks for the feedback that this is still "spammy") -- silent-by-default is just plain bad UX for operations which can take several minutes -- we need to be showing 'liveness by default' within notebook environments

another option might be to check some notebook-related environment variable at startup

johnkerl:
note that for the immediate term one can do `tiledbsoma.logging.warn()`

bkmartinjr:
is there a way to auto-magically detect "notebook" vs "not notebook"?  I will give it some thought, but I am not sure I have one. The possible alternative is an explicit config, _on_ by default, which an app (who wants less logging) can set to "off".

johnkerl:
looks like we both had messages in flight at the same time -- `tiledbsoma.logging.warn()

`

johnkerl:
@bkmartinjr is this resolved by #416?

bkmartinjr:
>  is this resolved by #416?

No, I left the logging intact when I refactored the code.  We need to have a design discussion about this at some point before 1.0 (but I don't think it is urgent).

johnkerl:
Cool @bkmartinjr -- I changed the title to _ IO module has spammy logging_

PR #1150, opened 2023-03-21
[python] Set log level depending on in-notebook or not, and use `stdout`

johnkerl:
**Issue and/or context:**

User feedback after https://github.com/single-cell-data/TileDB-SOMA/releases/tag/1.0.0

Also involves #390.

**Changes:**

* Log to stdout rather than stderr. This makes notebook cells not be pink.
* Log by default at a higher verbosity level for notebook environments.

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/logging.py
+++ b/apis/python/src/tiledbsoma/logging.py
@@ -4,9 +4,8 @@
 # Licensed under the MIT License.
 
 import logging
+import sys
 from typing import Optional
-
-logger = logging.getLogger("tiledbsoma")
 
 
 def warning() -> None:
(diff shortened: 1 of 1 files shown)


Shelnutt2 (approved):

johnkerl:
We may also want (on this PR):

* logger to `stderr` for non-notebooks
* logger to `stdout` for notebooks

bkmartinjr (commented):
I still believe this is an anti-pattern.  See the Python docs for an example of why: https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library.

Quoting:

> When developing a library which uses logging, you should take care to document how the library uses logging - for example, the names of loggers used. Some consideration also needs to be given to its logging configuration. If the using application does not use logging, and library code makes logging calls, then (as described in the previous section) events of severity WARNING and greater will be printed to sys.stderr. This is regarded as the best default behaviour.

*and*

> Note It is strongly advised that you do not add any handlers other than [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

PR #1156, opened 2023-03-23
[R] Fix reading very sparse assays

mojaveazure:
`matrix::sparseMatrix` drops unused rows/columns at the end of the matrix. This causes issues when reading in an `X` layer to a Seurat assay 

As a fix, explicitly set the matrix shape to `c(query$n_obs, query$n_vars)` when reading

### DIFF
--- a/apis/r/R/SOMAExperimentAxisQuery.R
+++ b/apis/r/R/SOMAExperimentAxisQuery.R
@@ -779,6 +779,7 @@ SOMAExperimentAxisQuery <- R6::R6Class(
           i = self$indexer$by_obs(obs)$as_vector() + 1L,
           j = self$indexer$by_var(var)$as_vector() + 1L,
           x = self$X(lyr)$GetColumnByName('soma_data')$as_vector(),
+          dims = c(self$n_obs, self$n_vars),
           repr = switch(EXPR = repr, D = 'T', repr)
         )
         mat <- Matrix::t(mat)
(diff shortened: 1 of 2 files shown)


apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -779,6 +779,7 @@ SOMAExperimentAxisQuery <- R6::R6Class(
           i = self$indexer$by_obs(obs)$as_vector() + 1L,
           j = self$indexer$by_var(var)$as_vector() + 1L,
           x = self$X(lyr)$GetColumnByName('soma_data')$as_vector(),
+          dims = c(self$n_obs, self$n_vars),
```
eddelbuettel, line 782: Nice. I think we landed a similar change for NDArray earlier.

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -779,6 +779,7 @@ SOMAExperimentAxisQuery <- R6::R6Class(
           i = self$indexer$by_obs(obs)$as_vector() + 1L,
```
aaronwolen, line 779: Looking at this again, is this internal `as_matrix()` function redundant with `private$as_matrix()`?

mojaveazure: Yes? I had built `private$.as_matrix()` to try to share code between the assay, dimensional reduction, and graph methods. However, DRs are dense and graphs have a shape of `[obs, obs]` instead of `[obs, var]` so sharing was moot. In the future, `private$.as_matrix()` might be useful for converting to v5 assays but as of now it should be removed in favor of the internal `as_matrix()`

(I forgot about it in the original PR and didn't want to expand the scope of this one, but can if you want)

aaronwolen (approved):
Thanks!

mlin (approved):
Confirming, I installed from this branch and it fixed the immediate issue I'd hit using `to_seurat()` in a Cell Census vignette. Thanks!

PR #1157, opened 2023-03-23
[python] Ingest `uns` values from AnnData files

thetorpedodog:
This ingests `uns` data into a similar structure to what v0.1.x does: a nested collection of groups, TileDB objects, and metadata.

- Scalar values are ingested as metadata entries.
- Pandas dataframes are ingested as SOMA dataframes, using the same logic as non-uns dataframes.
- Numpy ndarrays are ingested as SOMA dense ndarrays.

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -78,12 +78,13 @@ def from_h5ad(
     use_relative_uri: Optional[bool] = None,
     X_kind: Union[Type[SparseNDArray], Type[DenseNDArray]] = SparseNDArray,
 ) -> str:
-    """Reads an ``.h5ad`` file and writes it to a tiledbsoma Experiment.
+    """Reads an ``.h5ad`` file and writes it to an :class:`Experiment`.
 
-    Measurement data is stored in a Measurement in the experiment's ``ms`` field,
-    with the key provided by ``measurement_name``. Data elements are available
-    at the standard fields (``var``, ``X``, etc.). ``uns`` unstructured data is
-    partially supported, and is available at the measurement's ``uns`` key
+    Measurement data is stored in a :class:`Measurement` in the experiment's
+    ``ms`` field, with the key provided by ``measurement_name``. Data elements
+    are available at the standard fields (``var``, ``X``, etc.). Unstructured
+    data from ``uns`` is partially supported (structured arrays and non-numeric
+    NDArrays are skipped), and is available at the measurement's ``uns`` key
     (i.e., at ``your_experiment.ms[measurement_name]["uns"]``).
 
     Args:
@@ -158,12 +159,13 @@ def from_anndata(
     use_relative_uri: Optional[bool] = None,
     X_kind: Union[Type[SparseNDArray], Type[DenseNDArray]] = SparseNDArray,
 ) -> str:
-    """Writes an anndata object to a tiledbsoma Experiment.
+    """Writes an anndata object to a :class:`Experiment`.
 
-    Measurement data is stored in a Measurement in the experiment's ``ms`` field,
-    with the key provided by ``measurement_name``. Data elements are available
-    at the standard fields (``var``, ``X``, etc.). ``uns`` unstructured data is
-    partially supported, and is available at the measurement's ``uns`` key
+    Measurement data is stored in a :class:`Measurement` in the experiment's
+    ``ms`` field, with the key provided by ``measurement_name``. Data elements
+    are available at the standard fields (``var``, ``X``, etc.). Unstructured
+    data from ``uns`` is partially supported (structured arrays and non-numeric
+    NDArrays are skipped), and is available at the measurement's ``uns`` key
     (i.e., at ``your_experiment.ms[measurement_name]["uns"]``).
 
     Args:
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -78,20 +78,36 @@ def from_h5ad(
     use_relative_uri: Optional[bool] = None,
     X_kind: Union[Type[SparseNDArray], Type[DenseNDArray]] = SparseNDArray,
 ) -> str:
-    """Reads an ``.h5ad`` file and writes to a TileDB group structure.
+    """Reads an ``.h5ad`` file and writes it to a tiledbsoma Experiment.
 
-    Returns a URI for the created experiment.
+    Measurement data is stored in a Measurement in the experiment's ``ms`` field,
+    with the key provided by ``measurement_name``. Data elements are available
+    at the standard fields (``var``, ``X``, etc.). ``uns`` unstructured data is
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1006,6 +1017,116 @@ def _chunk_is_contained_in_axis(
...
+) -> None:
+    with _create_or_open_coll(
+        Collection, _util.uri_joinpath(parent.uri, parent_key), ingest_mode
+    ) as coll:
+        _maybe_set(parent, parent_key, coll, use_relative_uri=use_relative_uri)
+        coll.metadata["soma_tiledbsoma:type"] = "uns"
+        for key, value in dct.items():
+            if key == "rank_genes_groups":
+                # This is a structured array, which we do not support.
+                logging.log_io(None, f"skipping uns structured array {key!r}")
```
johnkerl, line 1053: this is worth a message at the lower-verbosity level -- only on exceptional occasions will people be using the debug level

johnkerl (commented):
I'm trying this now with some saved-off `.h5ad` files.

More info soon.

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: @thetorpedodog please update & i'll accept :)

johnkerl: unless this is stale but this PR has only a single commit so i'm not sure how it could have changed since i previously reviewed, unless you force-pushed which i also don't see 🤔

johnkerl (commented):
Another thought -- it's a bit weird that people can say `exp.ms["RNA"]["var"]` or `exp.ms["RNA"].var` but only `exp.ms["RNA"]["uns"]` and not `exp.ms["RNA"].uns`

Not saying we need to make the last one possible,but, maybe make it clear somehow/somewhere that they can't

thetorpedodog:
Made some changes in this respect; please take another look.

thetorpedodog:
> Another thought -- it's a bit weird that people can say `exp.ms["RNA"]["var"]` or `exp.ms["RNA"].var` but only `exp.ms["RNA"]["uns"]` and not `exp.ms["RNA"].uns`
> 
> Not saying we need to make the last one possible,but, maybe make it clear somehow/somewhere that they can't

This could be done with adding a somacore [`_mixin.item`](https://github.com/single-cell-data/SOMA/blob/b0f63f6fe0c87132ce3bfe9eff3e347850da7f31/python-spec/src/somacore/_mixin.py#L15) field to tiledbsoma’s Measurement class, but I would hold off on embedding something like that in the API until we are more certain about how much support and compatibility with `uns` we want.

johnkerl:
> > Another thought -- it's a bit weird that people can say `exp.ms["RNA"]["var"]` or `exp.ms["RNA"].var` but only `exp.ms["RNA"]["uns"]` and not `exp.ms["RNA"].uns`
> > 
> > Not saying we need to make the last one possible,but, maybe make it clear somehow/somewhere that they can't
> 
> This could be done with adding a somacore [`_mixin.item`](https://github.com/single-cell-data/SOMA/blob/b0f63f6fe0c87132ce3bfe9eff3e347850da7f31/python-spec/src/somacore/_mixin.py#L15) field to tiledbsoma’s Measurement class, but I would hold off on embedding something like that in the API until we are more certain about how much support and compatibility with `uns` we want.

Again I'm not asking to expose `.uns` as a property. I'm asking for clear user-level documentation -- somewhere -- show them they need to do `["uns"]`.

thetorpedodog:
> Again I'm not asking to expose `.uns` as a property. I'm asking for clear user-level documentation -- somewhere -- showing users they need to do `["uns"]`.

Added to the ingestor docstrings, which seems like the only sensible place to put it.

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -77,20 +78,36 @@ def from_h5ad(
     use_relative_uri: Optional[bool] = None,
     X_kind: Union[Type[SparseNDArray], Type[DenseNDArray]] = SparseNDArray,
 ) -> str:
-    """Reads an ``.h5ad`` file and writes to a TileDB group structure.
+    """Reads an ``.h5ad`` file and writes it to a tiledbsoma Experiment.
```
johnkerl, line 81:
```suggestion
    """Reads an ``.h5ad`` file and writes it to a tiledbsoma :class:`Experiment`.
```

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -77,20 +78,36 @@ def from_h5ad(
...
 
-    Returns a URI for the created experiment.
+    Measurement data is stored in a Measurement in the experiment's ``ms`` field,
```
johnkerl, line 83:
```suggestion
    Measurement data is stored in a :class:`Measurement` in the experiment's ``ms`` field,
```

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -77,20 +78,36 @@ def from_h5ad(
...
+    with the key provided by ``measurement_name``. Data elements are available
+    at the standard fields (``var``, ``X``, etc.). ``uns`` unstructured data is
+    partially supported, and is available at the measurement's ``uns`` key
```
johnkerl, line 86: In what sense is our support "partial"? This sounds scary and vague ...

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -141,20 +158,36 @@ def from_anndata(
     use_relative_uri: Optional[bool] = None,
     X_kind: Union[Type[SparseNDArray], Type[DenseNDArray]] = SparseNDArray,
 ) -> str:
-    """Top-level writer method for creating a TileDB group for an :class:`Experiment` object.
+    """Writes an anndata object to a tiledbsoma Experiment.
```
johnkerl, line 161:
```suggestion
    """Writes an anndata object to a tiledbsoma :class:`Experiment`.
```

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -141,20 +158,36 @@ def from_anndata(
...
 
-    Returns a URI for the created experiment.
+    Measurement data is stored in a Measurement in the experiment's ``ms`` field,
```
johnkerl, line 163:
```suggestion
    Measurement data is stored in a :class:`Measurement` in the experiment's ``ms`` field,
```

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: Now that we have #1161, let's not go backward

johnkerl (changes requested):

thetorpedodog:
Updated.

johnkerl (approved):

johnkerl:
Thanks @thetorpedodog -- this is huge!! :)

PR #1173, opened 2023-03-27
Filenames `soma_reader` -> `soma_array_reader`

johnkerl:
**Issue and/or context:**

Part two of #1147, which renamed `SOMAReader` -> `SOMAArrayReader` within source code. This PR does similarly with filenames.

Leaving this in draft a while while I sort out some things: in particular R still has `soma_reader` as a function name, not touched by #1147.


libtiledbsoma/src/soma_array_reader.cc
```diff
@@ -125,7 +125,8 @@ void SOMAArrayReader::reset(
     batch_size_ = batch_size;
 
     result_order_ = "auto";
-    if (result_order != "auto") {  // default "auto" is set in soma_reader.h
+    if (result_order !=
+        "auto") {  // default "auto" is set in soma_array_reader.h
```
eddelbuettel, line 129: (I understand it's all automatic formatting and what not but I really do think line 128 was more readable than the entirely equiivalent pair of 128+129 which blends it into the indented block.  Anyway, that ship has sailed.)

eddelbuettel (approved):
:ship:

nguyenv (approved):

libtiledbsoma/src/soma_array_reader.cc
johnkerl: @eddelbuettel it has, and the PR CI failed until/unless I ran `make format`.

PR #1176, opened 2023-03-27
[r] Also show osVersion when displaying package versions

eddelbuettel:
**Issue and/or context:**

Small 'quality of life' improvement also showing OS version

**Changes:**

Adds a line to the helper function:

```r
> tiledbsoma::show_package_versions()
tiledbsoma:    0.0.0.9014
tiledb-r:      0.19.0.1
tiledb core:   2.16.0
libtiledbsoma: libtiledbsoma=d5492c54-modified;libtiledb=2.16.0
R:             R version 4.2.3 (2023-03-15)
OS:            Ubuntu 22.10
> 
```

**Notes for Reviewer:**

closes #1148
[SC 26873](https://app.shortcut.com/tiledb-inc/story/26873/also-show-utils-osversion-when-showing)

PR #1177, opened 2023-03-27
[python] Include OS info in `show_package_versions`

johnkerl:
Python counterpart to #1176

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -249,6 +249,7 @@ setuptools.setup(
     install_requires=[
         "anndata",
         "attrs>=22.2",
+        "distro",
         # Pinning numba & its particular numpy constraints:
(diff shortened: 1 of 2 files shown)


eddelbuettel (approved):
Looks good, and reviewable at my Python level.

This sits on top of `uname` so you get the kernel _but not the OS release_ for which R puts extra grease in as discussed in the thread of the matching PR. 

Per https://note.nkmk.me/en/python-platform-system-release-version/ the 'distro' package may help. Might be overkill.

johnkerl:
@eddelbuettel using `os` might be underkill :) 

I think `distro` is exactly what we want -- thank you! :)

eddelbuettel (approved):
Nicer!

PR #1186, opened 2023-03-28
[python] Don't use `distro` in `show_package_versions`

johnkerl:
Conda issue: https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/9#issuecomment-1486852905

Reverts https://github.com/single-cell-data/TileDB-SOMA/pull/1177/commits/70cce388ae5dd81d5d371b265e5c5f4573569c45

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -249,7 +249,6 @@ setuptools.setup(
     install_requires=[
         "anndata",
         "attrs>=22.2",
-        "distro",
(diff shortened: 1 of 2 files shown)


Shelnutt2 (approved):
Thanks! I've got CI running for conda package: https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=31210&view=results . Once that is confirmed I'll test it downstream and we can merge.

PR #1189, opened 2023-03-28
Add `--tiledb` Argument For `setup.py`

nguyenv:
**Issue and/or context:**

`setup.py` could not find TileDB headers and libraries if installed in custom location.

**Changes:**

Add `--tiledb` argument to `setup.py`, so it can be added to include and lib paths when building Pybind11 bindings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ build ?= Release
 .PHONY: install
 install: clean
 	@./scripts/bld --prefix=${prefix} --tiledb=${tiledb} --build=${build}
-	@pip install -v -e apis/python
+	@cd apis/python && python setup.py install --tiledb=${tiledb}
 
 .PHONY: r-build
 r-build: clean
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -48,6 +48,9 @@ for arg in args:
     if (start, eq) == ("--libtiledbsoma", "="):
         libtiledbsoma_dir = pathlib.Path(last)
         sys.argv.remove(arg)
+    if arg.find("--tiledb") == 0:
+        tiledb_dir = pathlib.Path(arg.split("=")[1])
+        sys.argv.remove(arg)
 
 if libtiledbsoma_dir is None:
     scripts_dir = this_dir / "dist_links" / "scripts"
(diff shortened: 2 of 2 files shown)


johnkerl (approved):
🚢 
Thanks @nguyenv !!!!! :)

johnkerl:
@nguyenv Please update `apis/python/README.md` to suggest

```
TILEDB_PATH=/usr/local python -m pip install -v -e .
```

if

```
python -m pip install -v -e .
```

doesn't work and if there is a local build of TileDB Embedded in `/usr/local`.

apis/python/README.md
```diff
@@ -31,6 +31,7 @@ $ python -m pip install --upgrade tiledbsoma
 * `cd` into your checkout and then `cd apis/python`
 * `python -m pip install .`
 * Or, if you wish to modify the code and run it, `python -m pip install -v -e .`
+* If the TileDB and TileDB-SOMA libraries are locally installed to a custom directory, such as `/usr/local`, set the path with environmental variables `TILEDB_PATH` and `TILEDBSOMA_PATH`, `TILEDB_PATH=/usr/local python -m pip install -v -e .`
```
johnkerl, line 34:
```suggestion
* If the TileDB and TileDB-SOMA libraries are locally installed to a custom directory, such as `/usr/local`, set the path with environment variables `TILEDB_PATH` and `TILEDBSOMA_PATH`, `TILEDB_PATH=/usr/local python -m pip install -v -e .`
```

bekadavis9 (approved):
Fixed my build! Thanks @nguyenv !!!

eddelbuettel:
Could `pkg-config` report those paths to the user?

jdblischak:
Note that we'll also have to update the conda recipe when this update is released, since [`build-tiledbsoma-py.sh`](https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/dab34042bfdf7c7ababdf2bbbcebe6757263a35a/recipe/build-tiledbsoma-py.sh) currently relies on the `--libtiledbsoma` flag

Shelnutt2:
> Note that we'll also have to update the conda recipe when this update is released, since [`build-tiledbsoma-py.sh`](https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/dab34042bfdf7c7ababdf2bbbcebe6757263a35a/recipe/build-tiledbsoma-py.sh) currently relies on the `--libtiledbsoma` flag

@nguyenv is there a reason we don't want to support flags and env variables? I think TileDB-Py support both?

PR #1190, opened 2023-03-28
[python] Include PyPI wheel-publish in release CI

### LINKED ISSUE #706: [ci/python] Publish to PyPI on tags of main branch
It turns out from #700 #701 #704 yesterday that the wheel-publish process for tags on `main` is more manual than I had realized.

Current manual procedure:

See also:
https://github.com/single-cell-data/TileDB-SOMA/wiki#branches-and-releases

On conversation with @mlin it turns out we need to apply this recipe:
https://cibuildwheel.readthedocs.io/en/stable/deliver-to-pypi/#automatic-method

This issue tracks this work remaining to be done.

johnkerl:
Resolves #706

### DIFF
--- a/.github/workflows/python-packaging.yml
+++ b/.github/workflows/python-packaging.yml
@@ -25,13 +25,13 @@ jobs:
     - name: Build sdist
       run: python setup.py sdist
       working-directory: ./apis/python
-    - name: Upload sdist artifact
+    - name: Upload sdist artifact # upload to GitHub Actions storage
       uses: actions/upload-artifact@v3
       with:
         name: sdist
         path: apis/python/dist/
 
-  # This step builds wheels and uploads them to GitHub action storage.
+  # This step builds wheels and uploads them to GitHub Actions storage.
   # See also https://github.com/single-cell-data/TileDB-SOMA/issues/700.
   # See also https://github.com/single-cell-data/TileDB-SOMA/wiki/PyPI-packaging-WIP
   # for important transitional context.
@@ -68,7 +68,7 @@ jobs:
         #   run CMake build or not. Otherwise it'll keep reusing the library file built in the
         #   first iteration of cibuildwheel's outer loop, resulting in wheels with the library
         #   built for the wrong python version.
-    - name: Upload wheels-${{ matrix.platform }}
+    - name: Upload wheels-${{ matrix.platform }} # upload to GitHub Actions storage
       uses: actions/upload-artifact@v3
       with:
         name: wheels-${{ matrix.platform }}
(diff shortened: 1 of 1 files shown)


.github/workflows/python-packaging.yml
```diff
@@ -25,13 +25,13 @@ jobs:
     - name: Build sdist
       run: python setup.py sdist
       working-directory: ./apis/python
-    - name: Upload sdist artifact
+    - name: Upload sdist artifact # upload to GitHub Actions storage
```
thetorpedodog, line 28: any reason not to make this part of the step name, like "Upload sdist to GitHub Actions storage"?

.github/workflows/python-packaging.yml
```diff
@@ -122,42 +122,27 @@ jobs:
...
-    #  with:
-    #    user: __token__
-    #    password: ${{ secrets.PYPI_TOKEN }}
-    #    packages_dir: sdist, dist-wheel
-    #    verbose: true
+    # Reasoning:
+    # * We can mark a GitHub release as pre-release and not-latest while wheels are building
+    # * Then let this CI complete
+    # * Then mark the GitHub release as release and latest, so anyone installing from PyPI will get
+    #   usable wheels.
```
thetorpedodog, line 138: I don’t entirely follow the reasoning behind these steps for a couple reasons:

- The way I understand the prerelease flag is that it’s a characteristic of the version (i.e., tag), not a state. For example, v1.0.0 is a release, but 1.0.1b1 is a pre-release.
- The state of the tag on GitHub has no effect on what PyPI does; once artifacts for a tag are uploaded, it becomes a release, and is the "latest" release if its version is the highest one (which is basically how GitHub’s automatic "latest release" system also works).

thetorpedodog (commented):
The mechanism looks fine here, but I have some questions about the procedure…

.github/workflows/python-packaging.yml
johnkerl: > once artifacts for a tag are uploaded, it becomes a release

if this is how it works then my worries are alleviated

i want to avoid the (admittely small) time window in which verson n+1 exists but doesn't have wheels and users would get a bad install

thetorpedodog: Provided you upload the sdist and the wheels at the same time, that should not be an issue.

.github/workflows/python-packaging.yml
```diff
@@ -68,7 +68,7 @@ jobs:
         #   run CMake build or not. Otherwise it'll keep reusing the library file built in the
         #   first iteration of cibuildwheel's outer loop, resulting in wheels with the library
         #   built for the wrong python version.
-    - name: Upload wheels-${{ matrix.platform }}
+    - name: Upload wheels-${{ matrix.platform }} # upload to GitHub Actions storage
```
thetorpedodog, line 71: ibid.

thetorpedodog (approved):
Looks good to me; you might want to get mlin’s approval too, since he’s a bit more familiar with the tiledbsoma-specific release process stuff than I am.

johnkerl:
@mlin ping 🙏

johnkerl:
@mlin it's been a few days & I want to merge this on @thetorpedodog 's accept.

Please let me know if you have concerns & we can do a follow-on PR.

PR #1194, opened 2023-03-28
[r] Error on incomplete (non-iterated) reads

eddelbuettel:
**Issue and/or context:**

Reads that end in status INCOMPLETE should error as desired by #1111. 

**Changes:**

The response to INCOMPLETE is set to error with an explicit message about the memory budget option. 

Tests have been added to assert failure of a known-too-large data set under the 16mb default, and success with about three times that.

We also nudged a few loglevel messages to debug because some other earlier change now makes the default info, and added a comment when the data set package is referenced.

**Notes for Reviewer:**

Still open whether to turn iterated reads off.  This could be added.

#1111
[SC 26948](https://app.shortcut.com/tiledb-inc/story/26948/error-out-on-incomplete-queries)


aaronwolen (approved):
This looks good to me. Maybe double check with @mlin before merging since he raised the issue.

mlin (approved):
I'm OK with this as a strict improvement over returning an incomplete result. If I followed the thread on #1111 correctly, then it will still be possible to revisit this and do the iteration & `rbindlist()` automatically, in the future IF the error message proves to be a source of complaints.

PR #1195, opened 2023-03-29
[r] Fix query performance [WIP / not ready for review / please do not review]

johnkerl:
Some queries which specify both obs and var filters, even on small data, are found to be slower in tiledbsoma-r than tiledbsoma-py. Debugging through core reveals that the core strategy selector was choosing the legacy reader for X queries as initiated by tiledbsoma-r, vs choosing the sparse global-order reader as initiated by tiledbsoma-py. This in turn was caused by selection of row-major layout for the query; we should be using auto.

### DIFF
--- a/apis/r/R/utils-tiledb.R
+++ b/apis/r/R/utils-tiledb.R
@@ -7,7 +7,7 @@ tiledb_zstd_filter <- function(level = 3L) {
 }
 
 match_query_layout <- function(layout) {
-  layouts <- c("ROW_MAJOR", "COL_MAJOR", "GLOBAL_ORDER", "UNORDERED")
+  layouts <- c("ROW_MAJOR", "COL_MAJOR", "GLOBAL_ORDER", "UNORDERED", "auto") # "auto" is used by libtiledbsoma
   match.arg(layout, choices = layouts, several.ok = FALSE)
 }
 
--- a/apis/r/R/SOMASparseNdArray.R
+++ b/apis/r/R/SOMASparseNdArray.R
@@ -94,7 +94,7 @@ SOMASparseNDArray <- R6::R6Class(
     #' @return An [`arrow::Table`].
     read_arrow_table = function(
       coords = NULL,
-      result_order = "ROW_MAJOR",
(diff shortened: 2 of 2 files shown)


apis/r/R/utils-tiledb.R
```diff
@@ -7,7 +7,7 @@ tiledb_zstd_filter <- function(level = 3L) {
 }
 
 match_query_layout <- function(layout) {
-  layouts <- c("ROW_MAJOR", "COL_MAJOR", "GLOBAL_ORDER", "UNORDERED")
+  layouts <- c("ROW_MAJOR", "COL_MAJOR", "GLOBAL_ORDER", "UNORDERED", "auto") # "auto" is used by libtiledbsoma
```
aaronwolen, line 10: Good catch! Can you also remove GLOBAL_ORDER and UNORDERED, since those aren't valid values for soma_array_reader?

aaronwolen (approved):

eddelbuettel (commented):
I am puzzled how this could possibly have broken CI.

johnkerl:
there was one spot where I had `"auto"` that should have remained `"UNORDERED"` -- `libtiledbsoma` accepts `"auto"` not `"UNORDERED"`, tiledb-r the other way around

also, there is some CI slowdown i haven't figured out -- jobs "queued" but
https://github.com/organizations/single-cell-data/settings/actions/hosted-runners
showing nothing running ... 👀

johnkerl:
> there was one spot where I had "auto" that should have remained "UNORDERED" -- libtiledbsoma accepts "auto" not "UNORDERED", tiledb-r the other way around

... fixed by #1196

eddelbuettel:
Yeah I missed that.  Sometimes helpful to run tests locally first.

johnkerl:
@eddelbuettel indeed, my miss.

Another confusing issue in the mix was GitHub Actionsn were broken for a while earlier; including (among other things) cancelled jobs showing up as green. I failed to adequately deconvolve multiple issues going on simultaneously before merging #1195.

PR #1196, opened 2023-03-29
[r] Update result_order docs and defaults

aaronwolen:
**Issue and/or context:** #1195

**Changes:** A follow-up to #1195 that updates the documentation for `result_order` to remove invalid options: global_order and unordered. I've also set the result_order default to "auto" for all reader functions.

**Notes for Reviewer:** GitHub Actions is currently suffering an outage so no CI atm. However, I did validate this locally and confirmed it addresses the minor issue introduced in #1195.

PR #1197, opened 2023-03-29
[r] Revert #1195

johnkerl:
Need to re-think #1195 after some CI fails.


johnkerl:
Resolved by #1196

eddelbuettel:
I got held up by something else but in the meantime also confirmed that #1195 broke things when testing locally.

Issue #1198, opened 2023-03-30
[python] Update set-timestamp logic, pending a future TileDB-Py release


johnkerl:
Please see `TODO` comment at [type up later once I save it].

See also #1169.

jp-dark:
Closing issue as stale.

Issue #1169, opened 2023-03-25
[Bug][python] Excessive allocation of `tiledb.Ctx` objects causes resource starvation & fatal error


bkmartinjr:
## Summary

Opening a SOMA Collection in `tiledbsoma` creates a new `tiledb.Ctx` object, which in turn creates a threadpool. In some common and acceptable SOMA code patterns (e.g.,, using `Collection.__getitem__` to access a sub-collection), the implementation will generate a very large number of unreferenced (not garbage collected) `tiledb.Ctx` objects, which can lead to memory map resource starvation errors on Linux (due to the Linux kernel limit on mapped segments), which in turn leads to a failure to allocate a thread pool.

Work-around: embed `gc.collect()` calls directly in end-user code, in any location where there are a large number of Collection opens. In my opinion, this work-around should not be necessary -- it would be much better if the temporary `Ctx` objects were not created in the first place, as they serve no purpose other than to pass a timestamp in the `tiledb.Group` open.

Test case and analysis below. The test case is somewhat contrived to make the occasional error happen quickly (by tweaking the GC thresholds).

## Versions
```
tiledbsoma.__version__        1.1.1
TileDB-Py tiledb.version()    (0, 21, 1)
TileDB core version           2.15.0
libtiledbsoma version()       libtiledbsoma=;libtiledb=2.15.0
python version                3.9.16.final.0
```
Note: also fails on tiledbsoma==1.0.0.  The core issue has been present since the timestamp support was introduced, pre-1.0.0.

System: Ubuntu 20.04.6 LTS
Kernel vm.max_map_count = 65530

## Additional Detail

I have Census-related scripts (using tiledbsoma) that periodically fail with:

> tiledb.cc.TileDBError: [TileDB::C++API] Error: Failed to create context

Or more rarely, they fail with an unexplained SEGV which is likely related. These failures occur without memory pressure or other obvious system issues. The error occurred much more frequently on Linux EC2 instances with very high core counts (eg., r6id.32xlarge), and more rarely on smaller boxes.

Setting `"config.logging_level": "5"` in the TileDB context produced a more nuanced error:

> [2023-03-25 17:11:31.045] [Process: 896131] [error] [1679764279772244317-Global] C API: TileDB Internal, std::exception; Error initializing thread pool of concurrency level 0; Resource temporarily unavailable

[...]

PR #1199, opened 2023-03-30
[python] Avoid context leak on group-set-timestamp operation; use TileDB Embedded 2.15.1

johnkerl:
**Issue and/or context:** #1169. See also a follow-on tracker which is #1198.

See also #1222

### DIFF
--- a/apis/python/src/tiledbsoma/_tdb_handles.py
+++ b/apis/python/src/tiledbsoma/_tdb_handles.py
@@ -220,24 +220,11 @@ class GroupWrapper(Wrapper[tiledb.Group]):
         context: SOMATileDBContext,
         timestamp: int,
     ) -> tiledb.Group:
-        # * We want to do open-group-at-timestamp.
-        # * Timestamps are read from tiledb config.
-        # * We can't set-config on an already opened group.
-        # * We could create a new ctx, with the current ctx's config plus the
-        #   new timestamp config, but that causes a thread-pool leak. See also
-        #   https://github.com/single-cell-data/TileDB-SOMA/issues/1169.
-        # * For now, we do this close, set-config, and re-open workaround.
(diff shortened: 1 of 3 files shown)


apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -220,9 +220,24 @@ def _opener(
...
+        # * Timestamps are read from tiledb config.
+        # * We can't set-config on an already opened group.
+        # * We could create a new ctx, with the current ctx's config plus the
+        #   new timestamp config, but that causes a thread-pool leak. See also
+        #   https://github.com/single-cell-data/TileDB-SOMA/issues/1169.
+        # * For now, we do this close, set-config, and re-open workaround.
+        # * An upcoming TileDB-Py release will remove this need for us.
+        # TODO: make that happen.
+        # Tracking issue: https://github.com/single-cell-data/TileDB-SOMA/issues/1198
+        G = tiledb.Group(uri, mode)
```
Shelnutt2, line 233: I suggest we wait for the next TileDB-Py version where we just take the config on start, and avoid having to open and close.

nguyenv: FYI TileDB-Py 0.21.2 was released this morning and includes the addition of a `Group` constructor that takes in a `Config` argument.

johnkerl: @nguyenv on it 😎

johnkerl:
This is blocked until TileDB-Py 0.21.2 reaches PyPI:
https://pypi.org/project/tiledb/#history

johnkerl:
@nguyenv good to go?

nguyenv (approved):
Yes looks good to me.

Issue #1182, opened 2023-03-27
[c++] `tiledb.group_create` via `SOMAGroupWriter`


johnkerl:
This is one `tiledb-py` entrypoint which slices out very cleanly from the `tiledbsoma` implementation. Good/easy/quick for some initial dev work.

Parent: #922

johnkerl:
Blocked pending resolution of #1237

johnkerl:
@nguyenv AFAICT you are taking this over? I'm happy to drive this through to completion but I'll follow your sequencing as it develops.

johnkerl:
Closed after discussion with @nguyenv

Issue #1122, opened 2023-03-15
[c++] Implement group member-names-to-URIs map in C++


johnkerl:
Split-out from https://github.com/single-cell-data/TileDB-SOMA/issues/1122

This is listed separate from https://github.com/single-cell-data/TileDB-SOMA/issues/1248 because a member-names-to-URI mapping function does not exist in `tiledb::Group` and will be added for TileDB-SOMA.

johnkerl:
Blocked pending resolution of #1237

PR #1202, opened 2023-03-31
[c++/python/r] Move group-create to `libtiledbsoma`

johnkerl:
**Issue and/or context:**

#1182

**Changes:**

Replace `tiledb_group_create` (from TileDB-Py and TileDB-R) with `libtliedbsoma::SOMAGroupWriter::create`.

**Notes for Reviewer:**

The group-create is a small, isolable change that doesn't touch the layered complexities of `_tdb_handles.py`. This is a good place to set a pattern for the second `libtiledbsoma` class after the venerable `SOMAArrayReader`. More simply put, this is a quick and easy checkbox we can check off on #922.

Note that `SOMAGroupReader` is underway on #1122.


apis/python/tests/test_dataframe_index_columns.py
```diff
@@ -1812,70 +1812,100 @@ def test_types_write_errors(
             ["int32"],
             None,
             [pa.array([3202, 3203])],
-            RuntimeError,  # Static type (INT64) does not match expected type (INT32)
+            (
+                tiledb.cc.TileDBError,
```
johnkerl, line 1816: not caused by this PR, but necessary today in my sandbox (either on `main` or this branch)

eddelbuettel (commented):
(Concerning the R and C++ parts)  Seems fine but I cannot help but wonder if it is a longer-time winning strategy to duplicate what seems like identical code in two place.  But that decision was made above my paygrade so ...

Will hold on approving until I seen green everywhere.

johnkerl:
Thanks for the update @Shelnutt2 ! Moving to draft.

johnkerl:
> (Concerning the R and C++ parts) Seems fine but I cannot help but wonder if it is a longer-time winning strategy to duplicate what seems like identical code in two places. But that decision was made above my paygrade so ...
> 
> Will hold on approving until I seen green everywhere.

@eddelbuettel I put the `SOMAGroupWriter` logic in a common place, then did the Python bindings from the pybind11 file and the R bindings from the Rcpp file -- I think these are necessary ...

johnkerl:
Closing on discussion with @nguyenv

PR #1206, opened 2023-03-31
[c++] Fix build on MacOS 13.3

johnkerl:
**Issue and/or context:**

The third-party `catch2` package has

```
/Users/johnkerl/git/single-cell-data/TileDB-SOMA/build/libtiledbsoma/_deps/catch2-src/src/catch2/catch_session.cpp:64:25: warning: variable 'reporterIdx' set but not used [-Wunused-but-set-variable]
            std::size_t reporterIdx = 0;
```

which is not new,  but, newly this errors out in MacOS 13.3 (which yours truly installed just today).

**Changes:**

Since we don't control this third-party library but we do control how we compile it, apply the correct flags.


eddelbuettel (approved):

bekadavis9:
A bit late here but thanks for the fix!

Issue #1193, opened 2023-03-28
[Bug][python] missing version in `show_package_versions()` output


bkmartinjr:
**Describe the bug**
In release tiledbsoma==1.2.1, the `show_package_versions()` return value has an empty (non-existent) version number for libtiledbsoma:

```
In [2]: import tiledbsoma
In [3]: tiledbsoma.show_package_versions()
tiledbsoma.__version__        1.2.1
TileDB-Py tiledb.version()    (0, 21, 1)
TileDB core version           2.15.0
libtiledbsoma version()       libtiledbsoma=;libtiledb=2.15.0
python version                3.9.16.final.0
OS version                    Linux 5.15.0-1033-aws
```

I expected it to have some sort of version for the SOMA C++ layer, e.g., `libtiledbsoma=??.??.??`

johnkerl:
@bkmartinjr weird. I see

```
>>> soma.show_package_versions()
tiledbsoma.__version__        1.1.1.post32.dev1588231737
TileDB-Py tiledb.version()    (0, 21, 0)
TileDB core version           2.15.0
libtiledbsoma version()       libtiledbsoma=5eaa7e39;libtiledb=2.15.0
python version                3.10.6.final.0
OS version                    Ubuntu 22.04
```

```
>>> soma.show_package_versions()
tiledbsoma.__version__        1.2.1.post4.dev1874900368
TileDB-Py tiledb.version()    (0, 21, 1)
TileDB core version           2.15.0
libtiledbsoma version()       libtiledbsoma=6fc0b590;libtiledb=2.15.0
python version                3.9.0.final.0
OS version                    Darwin 22.3.0
```

on Linux & Mac, respectively.

From
https://github.com/single-cell-data/TileDB-SOMA/blob/1.2.1/libtiledbsoma/src/version.cc#L49
that thing after the `libtiledbsoma=` has to be `VERSION` which is either `"dev"` or `BUILD_COMMIT_HASH`, so it must be the latter ... and then
https://github.com/single-cell-data/TileDB-SOMA/blob/1.2.1/libtiledbsoma/src/CMakeLists.txt#L18-L41

So did you get this via `pip install` or via local checkout?

johnkerl:
Never mind; I can repro by doing `pip install tiledbsoma` (from PyPI, which I normally do not, since as a developer my normal M.O. is editable install from source) and I get

```
>>> soma.show_package_versions()
tiledbsoma.__version__        1.2.1
[...]

PR #1207, opened 2023-03-31
[c++] Omit no-longer-needed libtiledbsoma commit-hash info from `show_package_versions`

johnkerl:
As in #1193. The commit-hash information is (a) not always available; (b) never useful to the end user; (c) completely superseded now that we are (as we have been for some months now) doing tagged releases on `main`.

### DIFF
--- a/libtiledbsoma/src/version.cc
+++ b/libtiledbsoma/src/version.cc
@@ -34,19 +34,12 @@
 #include <tiledb/tiledb>
 #include "tiledbsoma/logger_public.h"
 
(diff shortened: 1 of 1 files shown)


nguyenv (approved):

johnkerl:
Thanks @nguyenv ! :)

eddelbuettel (approved):

eddelbuettel:
I liked it in the sense that commits _are_ more granular than tags, but I concur that if it is flaky under some builds it is not worth it.  Better to have clean information reliably.

Issue #1171, opened 2023-03-26
[r] improve usability of `SOMAAxisQuery` coords


mlin:
The following works:

```{r}
expt <- tiledbsoma::SOMAExperimentOpen(
  "s3://cellxgene-data-public/cell-census/2023-03-16/soma/census_data/mus_musculus",
  tiledbsoma_ctx = tiledbsoma::SOMATileDBContext$new(c("vfs.s3.region" = "us-west-2"))
)
expt_query <- tiledbsoma::SOMAExperimentAxisQuery$new(
  expt, "RNA",
  obs_query = tiledbsoma::SOMAAxisQuery$new(coords = list(soma_joinid = bit64::as.integer64(0:1000)))
)
print(expt_query$n_obs)
```

But there are a couple pitfalls setting `coords` in the `SOMAAxisQuery`:

1. If we supply `coords = list(soma_joinid = 0:1000)` (without passing through `bit64::as.integer64`) then the query successfully returns 0 rows.
2. If we forget to put the list wrapper, `coords = 0:1000` (with/without bit64) then we get a difficult-to-understand error message `Error in eval(ei, envir) : Expecting a string vector: [type=NULL; required=STRSXP].`  The error message here should be more helpful. (Should we accept a bare vector and interpret it as `soma_joinid`? Perhaps not, because making the developer write out `soma_joinid` serves as a reminder it's zero-based.)

eddelbuettel:
A good milestone that it works.  That said:
- point 1 is something @aaronwolen and I discussed. If and when we have the schema information and int comes in but int64 is needed then we should just upcast seamlessly.
- the list wrapper is a bit of pain too. I presume at least when we only have one dim we can similarly "listify" for the user.  In cases where there are more than dim it is difficult to do without lists.

mlin:
@eddelbuettel thanks for the quick feedback. Re the list wrapper on point 2 -- expanding on my final parenthetical comment -- indeed my first attempt at this query had simply `coords = 1:1000`, trying to use one-based indexing as seems natural in R. But then, being forced to write out `soma_joinid = ` reminded me that the SOMA spec defines that to be zero-based. So there's some small value in requiring the explicit field name. I'm sure the zero vs one-based indexing is going to come up again and lead to a larger discussion / audit. For now I think it suffices to continue erroring out, just with a more-helpful message.

mlin:
[...]

PR #1209, opened 2023-04-01
[r] Upcast axis query coords to integer64

eddelbuettel:
**Issue and/or context:**

In #1171 @mlin points out some usabilit shortcomings with an axis query.  This PR addresses one of the two concerns and lifts the requirement for explicit `as.integer64()` casts.  Changing the naming is a litte tricker as can be confronted with one or two dimension data sets and use a list with names if case fewer than all dimensions are constrained in the query.

**Changes:**

Integer vectors are now converted to `integer64` as needed.

**Notes for Reviewer:**

#1171
[SC 27100](https://app.shortcut.com/tiledb-inc/story/27100/r-improve-usability-of-somaaxisquery-coords)

### DIFF
--- a/apis/r/R/SOMADataFrame.R
+++ b/apis/r/R/SOMADataFrame.R
@@ -174,7 +174,7 @@ SOMADataFrame <- R6::R6Class(
               is.null(column_names) || all(column_names %in% c(self$dimnames(), self$attrnames()))
       )
 
-      coords <- validate_read_coords(coords, dimnames = self$dimnames())
+      coords <- validate_read_coords(coords, dimnames = self$dimnames(), int64cast = FALSE)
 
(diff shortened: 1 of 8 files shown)


eddelbuettel:
Reverting to draft -- turns out this bites in another code path where the column in fact 32 bit int.

eddelbuettel:
@mlin @aaronwolen @johnkerl @mojaveazure :  This is ready for review.

eddelbuettel:
Thanks for the quick approval @johnkerl -- I may wait for @mlin to ponder it and should probably throw in a new test or two.

johnkerl:
@eddelbuettel 💯

mlin (approved):
🚢

Issue #1089, opened 2023-03-10
[docs/python] Links missing in docs


ebezzi:
The docsite should have links to all Python objects, including:
1. Internal objects
2. Objects from `somacore` and `TileDB`
3. Common libraries: scipy, numpy, Arrow

Verify that intersphinx is set up correctly to pick those up, and add explicit `:ref:` when necessary.

johnkerl:
This is complete except for `intersphinx`. Note that the Arrow et al. links _do_ work correctly -- they are working without the assistance of `intersphinx`.

We will be moving our doc infra to Quarto soon (#1055) and `intersphinx` may not survive the move, so I am leaving that bit as unnecessary and soon to be obsoleted.

PR #1210, opened 2023-04-03
[docs/python] More doc links/styling

johnkerl:
Iterative work on https://github.com/single-cell-data/TileDB-SOMA/issues/1089


apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -105,16 +105,16 @@ def read(
         partitions: Optional[options.ReadPartitions] = None,
         platform_config: Optional[PlatformConfig] = None,
     ) -> "SparseNDArrayRead":
-        """Reads a user-defined slice of the SparseNDArray.
+        """Reads a user-defined slice of the :class:`SparseNDArray`.
```
thetorpedodog, line 108: given that we’re in the SparseNDArray class itself I’m not sure the cross-reference is that useful here.

thetorpedodog (approved):
one suggestion to take or leave; otherwise all fine

PR #1212, opened 2023-04-03
[R] [WIP] Port ephemeral classes to R

### LINKED ISSUE #1022: [r] Port ephemeral classes from Python to R
Split-out from #839.

TBD how much of this, and how, goes over idiomatically to Python once #736 is merged.

In any event #736 is a prerequisite.

mojaveazure:
Port ephemeral collection classes to R

Implemented:
- `EphemeralCollectionBase`: base ephemeral collection class
- `Collection`: ephemeral equivalent of `SOMACollection`

resolves #1022


apis/r/NAMESPACE
```diff
@@ -5,7 +5,9 @@ S3method("[[<-",MappingBase)
 S3method(as.list,MappingBase)
 S3method(length,MappingBase)
 S3method(names,MappingBase)
+export(Collection)
```
eddelbuettel, line 8: I have been pondering this yesterday and still think 'Collection' may be a little too generic and hence _possibly_ at the risk of a NAMESPACE clash.  That said I don't yet know or think of another package having such an identifier, and maybe being uppercase will save us.

apis/r/NAMESPACE
```diff
@@ -36,6 +37,9 @@ export(ScalarMap)
 export(TileDBArray)
 export(TileDBGroup)
 export(TileDBObject)
+export(VirtualCollection)
```
johnkerl, line 40: tiledbsoma-py and tiledbsoma-r aren't identical (different languages after all) but i don't see any value-add in calling these ephemeral in python and virtual in r ...

apis/r/R/ephemeral.R
```diff
@@ -0,0 +1,450 @@
+#' Ephemeral Collection Base
+#'
+#' @description Base class for ephemeral collections
```
johnkerl, line 3: please add a few words around what an ephemeral collection _is_: query results; in RAM; not directly associated with soma/seurat/bioc storage formats

apis/r/R/ephemeral.R
```diff
@@ -0,0 +1,450 @@
...
+#'
+#' @keywords internal
+#'
+#' @export
+#'
+EphemeralCollectionBase <- R6::R6Class(
+  classname = 'EphemeralCollectionBase',
+  inherit = SOMACollectionBase,
+  public = list(
+    # Overwrite TileDBObject methods
```
johnkerl, line 13: Override

apis/r/R/ephemeral.R
```diff
@@ -0,0 +1,450 @@
...
+    #' @description Create an ephemeral collection
+    #'
+    #' @template param-dots-ignored
+    #'
+    #'
+    initialize = function(...) {
+      if (rlang::dots_n(...)) {
```
johnkerl, line 20: tiledb and czi both are _highly_ polyglot companies -- any cleverness whatsoever (be it Python or R or anything else) -- i would ask for a quick comment on what this does

something like `os.path_exists(foo)` is clear without a comment; `if (rlange::dots_n(...))` is (to me) not

apis/r/R/ephemeral.R
```diff
@@ -0,0 +1,450 @@
...
+            warning(conditionMessage(e), call. = FALSE, immediate. = TRUE)
+          }
+        )
+      }
+      private$.data <- list()
+    },
+    #' @description \Sexpr[results=rd]{tiledbsoma:::rd_ephemeral_desc()}
+    #'
+    #' @return Returns \code{FALSE} as ephemeral collections do not
+    #' exist on-disk
```
johnkerl, line 33: ✨

apis/r/R/ephemeral.R
```diff
@@ -0,0 +1,450 @@
...
+          edited = ' and cannot be edited',
+          custom = paste0(' ', trimws(msg))
+        ),
+        call. = FALSE
+      )
+    }
+  )
+)
+
+#' Virtual Collections
```
johnkerl, line 319: as above -- i don't like virtual in r vs ephemeral in python -- without any (apparent) value-add in having different names

johnkerl (changes requested):
some name-changes

i especially like the unit tests! :)

aaronwolen (approved):
Looks great.

johnkerl (approved):
🚢

PR #1214, opened 2023-04-04
[r] Return lists of Arrow 'Table' objects in `read(iterated=TRUE)`

eddelbuettel:
**Issue and/or context:**

The common `read()` accessor for data frames and (dense and sparse) arrays has a boolean toggle `iterated`. When true, one-shot reads are attempted (subject to the memory budget constraints).  

When false, we had previously pushed the responsibility of iterating to the user.  This PR suggests to change that and to return a list of Arrow Table objects to the user.

A list is _the_ fundamental 'large' data structure in R. It is common and well understood; users can then proceed to iterated over the chunks as needed.

**Changes:**

A rather small code change for each of the three accessors along with a simplification in the unit tests as the test no longer needs to loop.

**Notes for Reviewer:**

[SC 27280](https://app.shortcut.com/tiledb-inc/story/27280/r-provide-iterated-reads-returning-list-of-arrow-tables)

### DIFF
--- a/apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
+++ b/apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
@@ -87,33 +87,19 @@ test_that("Iterated Interface from SOMA Classes", {
                       dense = SOMADenseNDArray$new(uri, internal_use_only = "allowed_use"))
         expect_true(inherits(sdf, "SOMAArrayBase"))
 
-        switch(tc,
-               data.frame = sdf$read(iterated = TRUE),
-               sparse = sdf$read_arrow_table(iterated = TRUE),
-               dense = sdf$read_arrow_table(iterated = TRUE))
-
-        expect_false(sdf$read_complete())
-        dat <- sdf$read_next()
-        n <- dat$num_rows
-        expect_true(n > 0)
-
-        expect_false(sdf$read_complete())
-        dat <- sdf$read_next()
-        n <- n + dat$num_rows
-        expect_true(n > 0)
-
-        expect_false(sdf$read_complete())
-        dat <- sdf$read_next()
-        n <- n + dat$num_rows
-        expect_true(n > 0)
-
-        expect_false(sdf$read_complete())
-        dat <- sdf$read_next()
-        n <- n + dat$num_rows
-        expect_true(n > 0)
-
-        expect_equal(n, 4848644)
+        rl <- switch(tc,
+                     data.frame = sdf$read(iterated = TRUE),
+                     sparse = sdf$read_arrow_table(iterated = TRUE),
+                     dense = sdf$read_arrow_table(iterated = TRUE))
+        expect_true(is.list(rl))
         expect_true(sdf$read_complete())
+        n <- length(rl)
+        expect_true(n > 0)
+
+        dat <- do.call(rbind, rl)
+        expect_true(inherits(dat, "Table"))
+        expect_equal(dat$num_columns, 3)
+        expect_equal(dat$num_rows, 4848644)
 
         rm(sdf)
     }
--- a/apis/r/R/SOMADataFrame.R
+++ b/apis/r/R/SOMADataFrame.R
@@ -192,23 +192,28 @@ SOMADataFrame <- R6::R6Class(
       }
 
       if (isFALSE(iterated)) {
+          cfg <- as.character(tiledb::config(self$tiledbsoma_ctx$get_tiledb_context()))
           rl <- soma_array_reader(uri = uri,
-                            colnames = column_names,   # NULL is dealt with by soma_array_reader()
-                            qc = value_filter,         # idem
-                            dim_points = coords,       # idem
-                            loglevel = log_level,      # idem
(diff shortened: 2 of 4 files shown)


apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
```diff
@@ -87,33 +87,19 @@ test_that("Iterated Interface from SOMA Classes", {
...
-
-        expect_false(sdf$read_complete())
-        dat <- sdf$read_next()
-        n <- n + dat$num_rows
+        rl <- switch(tc,
+                     data.frame = sdf$read(iterated = TRUE),
+                     sparse = sdf$read_arrow_table(iterated = TRUE),
+                     dense = sdf$read_arrow_table(iterated = TRUE))
+        expect_true(is.list(rl))
+        expect_true(sdf$read_complete())
```
johnkerl, line 95: i think this is not good -- my understanding is that for bigger-than-RAM data, the user must be able to get one chunk, process it, ask for the next chunk, and so on

having a list of _all_ chunks in RAM all at once, as I understand it, removes that feature

cc @mlin @atolopko-czi

johnkerl (changes requested):
And, my apologies if we should have gotten on the same page sooner, before coding on this PR ...

apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
atolopko-czi: In Python API, users can explicitly read all data into memory with an explicit `read().concat()...` call chain. So it's seems reasonable to support similar behavior in R. However, I would think the _default behavior_ should be _iterated_, since iterating over larger-than-core data is a key design tenet of TileDB-SOMA.

If this method is going to support both modes (iteration and all-at-once), we need a test for each mode, right?

eddelbuettel: Thanks for the feedback!  I reply in turn.

@johnkerl "It's complicated".  What you describe has been available since the fall -- but it is a much less idiomatic approach.  What we have here is _much_ cleaner for a typical R user.  First, we return a list (the one data structure made for this that every R user is familiar and can feed into `*apply`, or `purrr`, or ....  Second, because the chunks are Arrow Table objects they are _not yet materialized_ yet the user can do so at every loop step.  I really do think this gives the best control given "the circumstances": a user cannot control machine ram.

@atolopko-czi We have supported and tested iterated and non-iterated reads since the fall, please see test files (now pointing to `main` ie not the changed iterated test from this PR):  [test-SOMArrayReader-Basics.R](https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/r/tests/testthat/test-SOMAArrayReader-Basics.R) and [test-SOMAArrayReader-Iterated.R](https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R).

That said, I am of course as alwaus open to any and all suggestions.  This is merely a PR, but I do think it answers the need _as I would see it as an R user_.  Happy to hear from others.

atolopko-czi: Ah, I no longer have concerns now that I understand `Arrow Table objects they are not yet materialized` and agree that R idioms should dictate how the "sequence" (to choose a general term) of tables is returned. Ty!

eddelbuettel: I probably should have included an example....  So here it goes.

<details>

```r
> library(tiledbsoma)
> tdir <- tempfile()
> tgzfile <- system.file("raw-data", "soco-pbmc3k.tar.gz", package="pbmc3k.tiledb")
> untar(tarfile = tgzfile, exdir = tdir)
> uri <- file.path(tdir, "soco", "pbmc3k_processed", "ms", "RNA", "X", "data")
> sdf <- SOMADataFrameOpen(uri)
> rl <- sdf$read(iterated=TRUE)
> rl
[[1]]
Table
1213080 rows x 3 columns
$soma_dim_0 <int64 not null>
$soma_dim_1 <int64 not null>
$soma_data <float not null>

[[2]]
Table
1211242 rows x 3 columns
$soma_dim_0 <int64 not null>
$soma_dim_1 <int64 not null>
$soma_data <float not null>

[[3]]
Table
1213080 rows x 3 columns
$soma_dim_0 <int64 not null>
$soma_dim_1 <int64 not null>
$soma_data <float not null>

[[4]]
Table
1211242 rows x 3 columns
$soma_dim_0 <int64 not null>
$soma_dim_1 <int64 not null>
$soma_data <float not null>

>
> # and if one wants to (and can) stack
> bigtab <- do.call(rbind, rl)
> bigtab
Table
4848644 rows x 3 columns
$soma_dim_0 <int64 not null>
$soma_dim_1 <int64 not null>
$soma_data <float not null>
> 
```
</details>

johnkerl:
> Arrow Table objects they are not yet materialized

This is the key insight, and given this, my concerns are allayed.

johnkerl (approved):

Issue #1124, opened 2023-03-15
[python] Investigate TileDB-SOMA python package compatibility against pandas 2.0


maniarathi:
Investigate TileDB-SOMA Python package's compatibility with pandas [2.0.0RC0](https://pypi.org/project/pandas/2.0.0rc0/) to understand what changes, if any, will be required in order to support pandas 2.0.0. 

To be assigned to behnam pending GitHub -- leaving with @maniarathi for now

cc @nguyenv as pre-heads-up.

May need sync with `pyarrow`'s `to_pandas` etc

beroy:
https://github.com/single-cell-data/TileDB-SOMA/pull/1217

beroy:
The main finding here is using panda 2.0 will not much affect the current TileDB-SOMA code. However, given that the main use case of panda 2.0 in TileDB-SOMA happens through arrow and current build of parrow is not using panda 2.0 it can affect TileDB-SOMA when parrow uses panda 2.0 in the future. One possible next step is to build parrow with panda 2.0 and try that on TileDB-SOMA.

PR #1217, opened 2023-04-04
[python] Switching to pandas2 in `setup.py`

beroy:
For #1124 

Note that this change by itself may not be sufficient as most of our usecases is in the form of arrow table.concact.to_pandas() I tried to use `pd.set_option("mode.dtype_backend", "pyarrow")` but was not able to get the same effect and als to_pandas(...) does not have the dtype argument like some of the pandas methods.

**Issue and/or context:1124

**Changes:Updating pandas version

**Notes for Reviewer:**

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -264,7 +264,7 @@ setuptools.setup(
         # longer need to support the old pip solver (default on ubuntu 20.04).
         "numba==0.56.4",
         "numpy>=1.18,<1.24",
-        "pandas",
+        "pandas==2.0.0",
         "pyarrow>=9.0.0",
         "scanpy>=1.9.2",
         "scipy",


atolopko-czi (changes requested):
Let's try the official `2.0.0` version instead of the `rc0`. Looks like that's what causing the build failure as well.

beroy:
Too strange! I don't get that error locally.

atolopko-czi:
So when pandas2 is an explicit requirement of TileDB-SOMA, is pyarrow's `to_pandas()` returning a PyArrow-backed Pandas DataFrame? E.g., when executing `a_soma_dataframe.read().concat().to_pandas().  Since this "zero-copy" DataFrame instantiation is one of main benefits of the Pandas2 upgrade (https://datapythonista.me/blog/pandas-20-and-the-arrow-revolution-part-i), it's worth understanding what's happening in practice.

I think we'll also need to upgrade SOMA (the core repo) to use pandas2.

After doing a local install of TileDB-SOMA (via make install), what version of pandas does `pip list` report?

beroy:
Same error with 2.0.0

beroy:
```pip list
packaging           23.0
pandas              2.0.0
patsy               0.5.3
Pillow              9.4.0
pip                 22.3.1
pipdeptree          2.7.0
pluggy              1.0.0
pyarrow             11.0.0
pybind11-global     2.10.4
```

beroy:
Should we try to build pyarrow with panda2?

beroy:
The main finding here is using panda 2.0 will not much affect the current TileDB-SOMA code. However, given that the main use case of panda 2.0 in TileDB-SOMA happens through arrow and current build of arrow is not using panda 2.0 it can affect TileDB-SOMA when arrow uses panda 2.0 in the future. One possible next step is to build arrow with panda 2.0 and try that on TileDB-SOMA.

PR #1218, opened 2023-04-05
[r] Rename NdArray file names to match NDArray class names

johnkerl:
Quick/minor quality-of-life mod.


mojaveazure (approved):

eddelbuettel (changes requested):
If we have time for things like this please also bulk rename test file to _either_ use an underscore _or_ a hyphen.  I like hyphens better, you pick.

johnkerl:
> If we have time for things like this please also bulk rename test file to either use an underscore or a hyphen. I like hyphens better, you pick.

@eddelbuettel 💯 !! :) 

I did this before on https://github.com/single-cell-data/TileDB-SOMA/pull/553 and am happy to do it again.

eddelbuettel (approved):
:shipit:

PR #1216, opened 2023-04-04
[r] Port `TileDBCreateOptions` from Python to R

johnkerl:
**Issue and/or context:** #1183 

**Changes:**

**Notes for Reviewer:**

PR #1219, opened 2023-04-05
[r] Add `pbmc3k.tiledb` info to `apis/r/README.md`

johnkerl:
Found during work on #1216.


eddelbuettel (changes requested):
Sure but as the data package has not dependencies we do not need to point to CRAN.  Shorter, simpler, more concise without it.

eddelbuettel (approved):

nguyenv (approved):

Issue #1200, opened 2023-03-30
[python] `io.add_matrix_to_collection` always writes sparse arrays


mojaveazure:
`io.add_matrix_to_collection` [always writes a sparse array](https://github.com/single-cell-data/TileDB-SOMA/blob/ad00c8ed58f78ddc795b6ae5478947eb9cb512f0/apis/python/src/tiledbsoma/io/ingest.py#L656-L666), even when given a dense array. This poses an issue when someone uses it to add an array to `obsm` or `varm`, which the [spec defines as collections of dense arrays](https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md#somaexperiment-and-somameasurement). Moreover, the [R API expects](https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/r/R/SOMAExperimentAxisQuery.R#L571-L575) these to be dense arrays and treats them as such, which causes errors when coming from Python into R

johnkerl:
This is now a duplicate of #1245.

PR #1225, opened 2023-04-07
[r] Seurat outgestors expect sparse arrays in `obsm` and `varm`

mojaveazure:
Following discussion about #1200, `query$to_seurat()` and `query$to_seurat_reduction()` now expect arrays in `obsm` and `varm` to be sparse. Warnings are thrown if arrays are dense, but will continue to read them.

See also #1245.


eddelbuettel (approved):
This looks good to me!

johnkerl (approved):

PR #1224, opened 2023-04-06
[python] Wire `context` through `tiledbsoma.io.from_anndata`

johnkerl:
**Issue and/or context:**

Solves #1223.

**Changes:**

Wire `context` all the way from `tiledbsoma.io.from_anndata` to group/array I/O methods.

**Notes for Reviewer:**

I don't have ideas for unit-testing strategy.

What I did to test locally:

```
unset AWS_ACCESS_KEY_ID
unset AWS_SECRET_ACCESS_KEY
unset AWS_DEFAULT_REGION
```

Then edit `cfg.txt` and put in

```
vfs.s3.aws_access_key_id ***
vfs.s3.aws_secret_access_key ****
vfs.s3.region us-east-1
```

Then

```
ingestor --tiledb-config cfg.txt /path/to/local.h5ad s3://bucket/path
```

Issue #1223, opened 2023-04-06
[Bug][Python] `tiledbsoma.io.from_anndata` needs to handle `ctx` and `platform_config` equally


johnkerl:
**Describe the bug**

Corner-case bug:

* For anyone not using `tiledbsoma.io`, no problem
* For anyone using `tiledbsoma.io` not writing to S3 (e.g. local disk or TileDB Cloud), no problem
* For anyone using `tiledbsoma.io`, and writing to S3, and S3 creds available via `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`, no problem
* The bug is for anyone using `tiledbsoma.io`, writing to S3, without S3 creds via the environment, and only within the `ctx` parameter to `tiledbsoma.io.from_anndata` -- this `ctx` needs to get wired all the way down to array/group I/O

**To Reproduce**

```
./devtools/ingestor /path/to/some.h5ad s3://bucket/path/to/experiment
```

while not having the above AWS environment variablesset.

**Versions (please complete the following information):**
 - TileDB-SOMA version: 1.2.1
 - Language and language version (e.g. Python 3.8, R 4.2.2): Python 3.9 & 3.10 (but any)
 - OS (e.g. MacOS, Ubuntu Linux): MacOS & Linux

PR #1227, opened 2023-04-07
[python] Support CLI config arguments in `devtools/ingestor`

johnkerl:
**Issue and/or context:** This was useful on #1223; also very useful on some parameter-tuning work I'm doing.

**Changes:** Allows `devtools/ingestor` to get saved-off `tiledb.Config` and `PlatformConfig` from files at the command line.

**Notes for the reviewer:** This PR wires the `context` from the CLI to `tiledbsoma.io`. #1224 wires it from there.

Sample invocation:

```
ingestor --platform-config-file=pc.json --tiledb-config-file cfg.txt ~/s/a/pbmc-small.h5ad ~/s/v/pbmc-small
```

### DIFF
--- a/apis/python/devtools/ingestor
+++ b/apis/python/devtools/ingestor
@@ -11,11 +11,13 @@ A simple driver for ingestion of anndata to a TileDB Experiment.
 """
 
 import argparse
+import json
 import os
 import sys
 from typing import Optional
 
 import tiledb
+from somacore import options
 
 import tiledbsoma
 import tiledbsoma._exception
@@ -90,8 +92,33 @@ select `relative=True`. (This is the default.)
         nargs=1,
     )
     parser.add_argument(
+        # Given a tiledb.Config object `cfg`, you can do `cfg.save("mycfg.txt")`.
+        # It'll look like this:
+        #
+        # ...
+        # config.logging_level 0
+        # filestore.buffer_size 104857600
+        # rest.curl.buffer_size 524288
+        # ...
+        #
+        # Then you can edit it if you like. Then you can have this script read it
+        # using `--tiledb-config mycfg.txt`.
         "--tiledb-config",
         help="TileDB config data from `config.save()`",
+        type=str,
+        default=None,
+    )
+    parser.add_argument(
+        # Given a PlatformConfig object -- just a nested dict -- named `platform_config` you can do
+        #
+        #   import json
+        #   with open("pc.json", "w") as F:
+        #     json.dump(platform_config, F)
+        #
+        # Then you can edit it if you like. Then you can have this script read it
+        # using `--platform-config pc.json`.
+        "--platform-config",
+        help="Platform-config data as JSON",
         type=str,
         default=None,
     )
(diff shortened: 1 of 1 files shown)


apis/python/devtools/ingestor
```diff
@@ -90,11 +92,36 @@ select `relative=True`. (This is the default.)
...
+        #   with open("pc.json", "w") as F:
+        #     json.dump(platform_config, F)
+        #
+        # Then you can edit it if you like. Then you can have this script read it
+        # using `--platform-config pc.json`.
+        "--platform-config",
+        help="Platform-config data as JSON",
+        type=str,
+        default=None,
+    )
```
atolopko-czi, line 124: would indicate specifically that this expects a file and not a json string either via option name `--platform-config-file` or in `help` text.  Same for `--tiledb-config`, if not too late to change that.  Or maybe they are made to accept either file or json string?

apis/python/devtools/ingestor
```diff
@@ -90,11 +92,36 @@ select `relative=True`. (This is the default.)
...
+        # It'll look like this:
+        #
+        # ...
+        # config.logging_level 0
+        # filestore.buffer_size 104857600
+        # rest.curl.buffer_size 524288
+        # ...
+        #
+        # Then you can edit it if you like. Then you can have this script read it
+        # using `--tiledb-config mycfg.txt`.
```
thetorpedodog, line 105: A version of this text could appear in the `help` text itself.

atolopko-czi (commented):
Not sure if we have unit tests yet for the ingestor, but if so, can the new option be tested?

apis/python/devtools/ingestor
```diff
@@ -90,11 +92,36 @@ select `relative=True`. (This is the default.)
...
+        #
+        #   import json
+        #   with open("pc.json", "w") as F:
+        #     json.dump(platform_config, F)
+        #
+        # Then you can edit it if you like. Then you can have this script read it
+        # using `--platform-config pc.json`.
+        "--platform-config",
+        help="Platform-config data as JSON",
+        type=str,
```
thetorpedodog, line 122: For things intended to be opened as files, consider using https://docs.python.org/3/library/argparse.html#argparse.FileType

apis/python/devtools/ingestor
johnkerl: @atolopko-czi this takes a file -- i'll clarify

and we can also take it from a json string, with another CLI argument if folks like

johnkerl:
> Not sure if we have unit tests yet for the ingestor, but if so, can the new option be tested?

this is why this is in `devtools` ...

ebezzi (approved):

thetorpedodog:
> > Not sure if we have unit tests yet for the ingestor, but if so, can the new option be tested?
> 
> This is why this is in `devtools` ... we don't unit-test this but it's used interactively

Is this intended to be used by outside users or strictly by us?

johnkerl:
> Is this intended to be used by outside users or strictly by us?

Just us devs

PR #1228, opened 2023-04-07
[r] Enable parallelized reads for experiment queries

### LINKED ISSUE #943: [r] Use Futures library to accelerate `ExperimentQuery`
Context: #839

mojaveazure:
Add support for parallelized reads in experiment queries. Parallelization is controlle by new `n_cores` parameter in `query$read()`; parallelization is not available on Windows due to paralleization package

Direct dependencies added:
- parallel (included in standard R installations)

Questions for reviewers:
- Is `parallel::mc*` good enough or should we use a more feature-full implementation (eg. `future.apply::future_*apply()`, `pbapply::pb*apply()`)

resolves #943


apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -123,8 +123,10 @@ SOMAExperimentAxisQuery <- R6::R6Class(
     #' @param X_layers The name(s) of the `X` layer(s) to read and return.
     #' @param obs_column_names,var_column_names Specify which column names in
     #' `var` and `obs` dataframes to read and return.
+    #' @param n_cores Number of cores to use for parallelized reads; disabled
+    #' on Windows
     read = function(
-      X_layers = NULL, obs_column_names = NULL, var_column_names = NULL) {
+      X_layers = NULL, obs_column_names = NULL, var_column_names = NULL, n_cores = 1L) {
```
eddelbuettel, line 129: There is an _existing_ option R use for this, we use this in tiledb-R as well (mostly to throttle _down_ at CRAN at their request, please see https://github.com/TileDB-Inc/TileDB-R/blob/master/R/Config.R#L226-L234  

So maybe using `n_cores=getOption("Ncpus", 2L)` (or 1) is what we want here?

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -134,14 +136,33 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
-        assert_subset(var_column_names, self$var_df$colnames(), "column")
+        assert_subset(var_column_names, self$var_df$colnames(), "column"),
+        "'n_cores' must be a single positive integer value" = rlang::is_integerish(n_cores, 1L) &&
+          n_cores > 0L
       )
 
+      if (Sys.info()['sysname'] == 'Windows') {
+        if (n_cores != 1L) {
+          warning(
+            "Parallelization is not supported on Windows",
```
eddelbuettel, line 147: It's been years for me since I mucked with Windows but as I recall even package `parallel` supports a version of parallelism via `mclapply` -- see the help.  There are some limits because `fork` isn't there.

But the bigger point may be that this is a non-issue for us:  `tiledbsoma` is a non-Windows package right now.  I think we should keep it simpler and remove all Windows concerns because 

https://github.com/single-cell-data/TileDB-SOMA/blob/797ac6720e4fdbf257e29d337cde851806a732e2/apis/r/DESCRIPTION#L67

eddelbuettel (changes requested):
This is awesome -- I particularly like that you stuck with base R which I happen to have good experience with (even from before `parallel` was created and we mostly relied on `snow` (from which is took parts, `snow` still exists) and `multicore` (which it subsumed). 

I know there are alternatives, and they have their fans and deservedly so but sticking with something simpler first makes sense to me.

Lastly, I think we should throw out the stubs for Windows as this is (currently, at least) a non-Windows package and we have more fundamental architectural issues to address before it becomes one.

apis/r/R/SOMAExperimentAxisQuery.R
mojaveazure: My philosophy is to default to no parallelization (1), and require that users who want to parallelization should enable it explicitly

apis/r/R/SOMAExperimentAxisQuery.R
eddelbuettel: I misremembered: Windows 'tolerates' `mclapply` but falls back to serial `lapply` use essentially already giving us the fallback made explicit here.  

So for simplicitly (and while there is no Windows build) I still think we can remove the special case made here that will simply not be triggered until we have a Windows build.

apis/r/R/SOMAExperimentAxisQuery.R
eddelbuettel: CRAN allows two in all settings.  Many users want more.

But TileDB Core uses all the cores it can get (unless the user overrides).  So does TileDB-R.

eddelbuettel:
Package `parallel` is in Imports but not (yet?) in NAMESPACE.  That would lead to a nag.

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -150,19 +176,23 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
           x_layer
         },
-        layer_name = X_layers
+        layer_name = X_layers,
+        mc.cores = n_cores
       )
 
       # TODO: parallelize with futures
       obs_ft <- self$obs(obs_column_names)
       var_ft <- self$var(var_column_names)
```
aaronwolen, line 185: We should also try performing these 2 axis queries in parallel.

johnkerl:
Per @mojaveazure the hoped-for performance improvement here was not obtained.

PR #1252, opened 2023-04-12
[python] Pin anndata to < 0.9 since it doesn't support Python 3.7 like we do

johnkerl:
Context: https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/11#issuecomment-1505662937

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -209,9 +209,6 @@ if os.name == "posix" and sys.platform != "darwin":
 # be at top level, outside any if-block
 # https://github.com/pypa/cibuildwheel/blob/7c4bbf8cb31d856a0fe547faf8edf165cd48ce74/cibuildwheel/projectfiles.py#L41-L46
 
-# Needed for Python 3.7 which anndata 0.9 doesn't support but we do
-anndata_version = "anndata < 0.9" if sys.version_info < (3, 8, 0) else "anndata"
-
 setuptools.setup(
     name="tiledbsoma",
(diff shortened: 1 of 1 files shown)


jdblischak:
Is it possible to pin anndata < 0.9 only for Python 3.7 builds? My update to the conda recipe allows users to install the latest anndata if they are using Python >=3.8. It would be nice to be able to regularly test that pytiledbsoma is both forwards and backwards compatible with anndata

johnkerl:
@jdblischak good call!! :) 

I think this will work: https://github.com/single-cell-data/TileDB-SOMA/pull/1252/commits/0ae2b545c20176b3eb4d23d897f7943601840ae7

johnkerl:
@seth ping

apis/python/setup.py
```diff
@@ -209,6 +209,9 @@ def run(self):
 # be at top level, outside any if-block
 # https://github.com/pypa/cibuildwheel/blob/7c4bbf8cb31d856a0fe547faf8edf165cd48ce74/cibuildwheel/projectfiles.py#L41-L46
 
+# Needed for Python 3.7 which anndata 0.9 doesn't support but we do
+anndata_version = "anndata < 0.9" if sys.version_info < (3, 8, 0) else "anndata"
```
thetorpedodog, line 213: You can use platform-specific dependency specification and avoid this manual version detection:

https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#platform-specific-dependencies

thetorpedodog (approved):

jdblischak (approved):

Issue #1242, opened 2023-04-10
Develop a new release procedure


johnkerl:
Status quo:

https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases

Needs to be updated to include:

* Conda green-CI test instructions before tagging a release
* Whether to use an RC
* Other items discussed 1-1-1 between @aaronwolen @sethshelnutt @johnkerl, to be typed up
* All info inclusive/understandable/executable by all project partners

johnkerl:
* https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases
* https://github.com/single-cell-data/TileDB-SOMA/wiki/PyPI-packaging-notes

PR #1253, opened 2023-04-13
[python] Flexible `--libtiledbsoma` specification in `setup.py`

johnkerl:
As discussed at https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/13.

Context: https://github.com/single-cell-data/TileDB-SOMA/issues/1242


jdblischak (approved):
Should we also add an equivalent flag for the path to libtiledb, eg `--libtiledb`? If there are use cases where these paths need to be separately specified by env vars, presumably the same level of granularity would be useful for the cli args

johnkerl:
@jdblischak sounds great to me! I'll defer to @nguyenv for final accept.

PR #1254, opened 2023-04-14
[r] Upcast SOMADataFrame coords to int64

aaronwolen:
**Issue and/or context:** #1171, #1209

**Changes:** This PR is a follow-up to #1209 and adds support for automatically upcasting integer coordinates passed to `SOMADataFrame$read()` to int64. Now it's possible to slice with just:

```r
experiment$obs$read(coords = 0:9)
```

**Notes for Reviewer:** The internal `validate_read_coords()` function gains a `schema` argument, which is used to determine whether casting is necessary. This can be extended in the future to improve coordinate validation for other data types. Added a new suite of tests to validate behavior for different cases.


eddelbuettel (approved):
Nice.  As Dave Letterman would ask on TV:  "Does it float?" (when launching something from the roof ...)

I got myself tied into some knots when I tried to cover the upcasting as the possible set of argument types are somewhat general -- and it looks like I missed one.  

But big thumbs up for smoothing the UI some more.   From (more natural) `int` to the (internally used) `integer64` is a win.

PR #1260, opened 2023-04-18
[python] Write TileDB Cloud–specific metadata to Experiments

thetorpedodog:
TileDB Cloud uses the `dataset_type` key to provide richer information about datasets. They need to be set on the root Experiment.

### DIFF
--- a/apis/python/src/tiledbsoma/_experiment.py
+++ b/apis/python/src/tiledbsoma/_experiment.py
@@ -6,11 +6,14 @@
 """Implementation of a SOMA Experiment.
 """
 
+from typing import Any
+
 from somacore import experiment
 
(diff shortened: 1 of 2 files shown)


johnkerl (approved):
This feels liike something should opt-in, and customizable, e.g. via Platform Config.

OTOH saying `dataset_type = "soma"` seems like a very generic and innocent thing to do ...

thetorpedodog:
> This feels liike something should opt-in, and customizable, e.g. via Platform Config.
> 
> OTOH saying `dataset_type = "soma"` seems like a very generic and innocent thing to do ...

I can see adding an opt-out in the future, but for now having it automatically do the “right thing” seems good enough.

Issue #1192, opened 2023-03-28
[r] Seurat <> SOMA Interop Future Work


mojaveazure:
Inital Seurat <> SOMA interop added in #1119  (#941) and #1146 (#942). This issue is tracking additional goals desired for Seurat <> SOMA interop

SOMA &rarr; Seurat goals:
- [ ] allow `SOMAExperiment` &rarr; `Seurat` conversion instead of `SOMAExperiment` &rarr; query &rarr; `Seurat`
- [ ] allow appending data from SOMA to existing `Seurat` object
- [ ] support for multimodal loading
- [ ] support for extended assays (eg. `SCTAssay`, `ChromatinAssay`)
- [ ] support for spatially-resolved datasets

Seurat &rarr; SOMA goals:
- [ ] allow ingesting of extended assays (eg. `SCTAssay`, `ChromatinAssay`)
- [ ] allow ingesting of spatially resolved datasets

johnkerl:
❤️ thanks @mojaveazure !! :)

PR #1261, opened 2023-04-18
[r] [WIP] Port `to_seurat()` to `SOMAExperiment`

mojaveazure:
Allow creating `Seurat` object directly from SOMA experiments; this includes support reading in multimodal SOMA objects into a multimodal `Seurat` object

Implemented SOMA methods:
- `SOMAExperiment$to_seurat()`: allows loading in entire `Seurat` objects, including partially suppressing layers in every `ms/X`, dimensional reduction information, and nearest-neighbor graphs

Future work (tracked in #1192):

### DIFF
--- a/apis/r/R/SOMAExperiment.R
+++ b/apis/r/R/SOMAExperiment.R
@@ -37,7 +37,7 @@ SOMAExperiment <- R6::R6Class(
     #'   ADT = c("ensembl_id")
     #' )
     #' }
-    #' By default, loads in entire feature-level meta data for all measurements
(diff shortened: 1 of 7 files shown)


aaronwolen (changes requested):
This is looking great. I left a few comments but one issue that needs to be addressed making the conversion via an empty query because it will always perform 2 reads from obs and var (once just to retrieve the joinids and then again to fetch the data). I love the effort to reuse code but in this case you might need to factor out the common components of SOMAExperiment$to_seurat and SOMAExperimentAxisQuery$to_seurat rather than trying to use the latter to implement the former.

Issue #1184, opened 2023-03-27
[r/python] Create a roundtrip AnnData -> SOMA -> Seurat test


ebezzi:
Create a roundtrip test that:

- [X] Starts from an AnnData, converts it to SOMA, and then to Seurat
- [x] Starts from a Seurat, converts it to SOMA, and then to AnnData

These tests should:
1. Use assertions to make sure that the final file is equivalent to the original (that should include, at minimum, `X`, `raw.X`, `var` and `obs`).
2. Use a real dataset (pbmc3k suggested)

This work is pending the completion of the Seurat conversion work.

PR #1263, opened 2023-04-19
[r][python] Create an AnnData -> SOMA -> Seurat test 

ebezzi:
**Issue and/or context:**
#1184 

**Changes:**
Adds an AnnData -> SOMA -> Seurat using the existing python/R framework

**Notes for Reviewer:**
1. It is possible to add more assertions if desired (e.g. on the X layer)
2. It is possible to replicate the tests on different artifacts. This one uses a stripped down version of pbmc3k that was already in the repo
3. Right now, the R side requires a query since `to_seurat` is only available at a query level. Eventually it will be available on `SOMAExperiment` (#1261) and this test should be updated to be simpler.
4. Right now we can only do one "trip" of the test, since Seurat -> SOMA isn't in main yet (it's [here](https://github.com/single-cell-data/TileDB-SOMA/pull/1146) though). Once this is merged we can add the other test.

### DIFF
--- a/apis/system/tests/test_write_anndata_read_seurat.py
+++ b/apis/system/tests/test_write_anndata_read_seurat.py
@@ -1,0 +1,66 @@
+import anndata as ad
+import pytest
+
+import tiledbsoma as soma
+
+from .common import TestWritePythonReadR, embed_python_list_into_R_code
+
+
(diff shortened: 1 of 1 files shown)


johnkerl (approved):
LGTM! :)

PR #1265, opened 2023-04-19
[python] Explicit casting of Numpy nulls to Arrow nulls

mojaveazure:
Cast Numpy nulls (eg. NaN) to an arrow null when writing data frames. Necessary when obs/var in AnnData/h5ad contain nulled columns

Reprex:
```python
import os
import shutil
import tiledbsoma
import tiledbsoma.io
import tempfile
import pandas as pd
import scanpy as sc
import numpy as np

SOMA_PATH = os.path.join(tempfile.gettempdir(), "soma-exp")
tiledbsoma.logging.debug()

adata = sc.datasets.pbmc3k()

# Create a pandas series with datatype categorical containing all Nan values
adata.obs["empty"] = pd.Categorical(
    [np.NaN] * adata.n_obs, dtype=pd.CategoricalDtype(categories=[], ordered=False)
)

if os.path.exists(SOMA_PATH):
    print(f"Removing existing SOMA: {SOMA_PATH}")
    shutil.rmtree(SOMA_PATH)

tiledbsoma.io.from_anndata(
    experiment_uri=SOMA_PATH, anndata=adata, measurement_name="RNA"
)
```

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -711,7 +711,7 @@ def _build_tiledb_schema(
     dom = tiledb.Domain(dims, ctx=context.tiledb_ctx)
 
     attrs = []
-    metadata = schema.metadata or dict()
+    metadata = schema.metadata or {}
     for attr_name in schema.names:
         if attr_name in index_column_names:
             continue
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -720,7 +720,7 @@ def _build_tiledb_schema(
             dtype=_arrow_types.tiledb_type_from_arrow_type(
                 schema.field(attr_name).type
             ),
-            nullable=metadata.get(bytes(attr_name, "utf-8")) == b"nullable",
+            nullable=metadata.get(attr_name.encode("utf-8")) == b"nullable",
             filters=tiledb_create_options.attr_filters(attr_name, ["ZstdFilter"]),
             ctx=context.tiledb_ctx,
         )
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -711,6 +711,7 @@ def _build_tiledb_schema(
     dom = tiledb.Domain(dims, ctx=context.tiledb_ctx)
 
     attrs = []
+    metadata = schema.metadata or dict()
     for attr_name in schema.names:
         if attr_name in index_column_names:
             continue
@@ -719,6 +720,7 @@ def _build_tiledb_schema(
             dtype=_arrow_types.tiledb_type_from_arrow_type(
                 schema.field(attr_name).type
             ),
+            nullable=metadata.get(bytes(attr_name, 'utf-8')) == b'nullable',
             filters=tiledb_create_options.attr_filters(attr_name, ["ZstdFilter"]),
             ctx=context.tiledb_ctx,
         )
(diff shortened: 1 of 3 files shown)


johnkerl (commented):
LGTM but I also want @thetorpedodog 's input

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -711,6 +711,7 @@ def _build_tiledb_schema(
     dom = tiledb.Domain(dims, ctx=context.tiledb_ctx)
 
     attrs = []
+    metadata = schema.metadata or dict()
```
thetorpedodog, line 714: prefer `{}` in place of `dict()`

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -719,6 +720,7 @@ def _build_tiledb_schema(
             dtype=_arrow_types.tiledb_type_from_arrow_type(
                 schema.field(attr_name).type
             ),
+            nullable=metadata.get(bytes(attr_name, "utf-8")) == b"nullable",
```
thetorpedodog, line 723: This can also be written as `attr_name.encode("utf-8")`, which (imo) is a little easier to understand

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -572,10 +572,18 @@ def _write_dataframe(
     df.set_index(SOMA_JOINID, inplace=True)
 
     # Categoricals are not yet well supported, so we must flatten
+    null_fields = set()
     for k in df:
         if df[k].dtype == "category":
             df[k] = df[k].astype(df[k].cat.categories.dtype)
+        elif df[k].isnull().all():
+            df[k] = pa.nulls(df.shape[0], pa.infer_type(df[k]))
+            null_fields.add(k)
```
thetorpedodog, line 581: Can these two both be true? i.e., a category might be all-nulls?

(also, should this be if *any* entry is null?)

apis/python/tests/test_basic_anndata_io.py
```diff
@@ -417,3 +418,21 @@ def test_export_anndata(adata):
         assert readback.obsp[key].shape == adata.obsp[key].shape
     for key in adata.varp.keys():
         assert readback.varp[key].shape == adata.varp[key].shape
+
+
+def test_null_obs(adata):
+    tempdir = tempfile.TemporaryDirectory()
```
thetorpedodog, line 424: While the pattern you're using here is the one from the existing tests, there are two different ways this could be slightly improved:

`tempfile.TemporaryDirectory()` works as a context manager, yielding the path of the temporary directory, so you can do:

```python
with tempfile.TemporaryDirectory() as output_path:
    ...  # do stuff with output_path
```

BUT in the specific case of a unit test, pytest offers a "fixture" which it magically provides based on parameter name (like `adata` here) called `tmp_path`, which will be a `pathlib.Path` that it automatically cleans up:

https://github.com/single-cell-data/TileDB-SOMA/blob/a78b1e6be9fe5c51d8fcfd317b1e05649424cd12/apis/python/tests/test_collection.py#L154

The other tests should probably be updated to match (but not right this minute).

thetorpedodog (approved):
I have a few minor suggestions and one question. Looks good overall!

apis/python/src/tiledbsoma/io/ingest.py
mojaveazure: Yes and yes, thanks for catching!

apis/python/tests/test_basic_anndata_io.py
mojaveazure: Good to know; I've replaced `tempdir` with the fixture for this particular test

PR #1268, opened 2023-04-20
Unwrap numpy scalar values when importing `uns` data.

thetorpedodog:
Entries in `uns` may be read as numpy scalars rather than native Python types (e.g. `numpy.float64` rather than `float`). We only accept normal Python values in metadata, so unwrap those numpy scalars.


aaronwolen (approved):
Thank you!

mojaveazure (approved):

PR #1272, opened 2023-04-21
[r] Refine group member caching behavior

aaronwolen:
**Changes:** This refines member cache handling in the `TileDBGroup` class.

**Notes for Reviewer:** When adding a new member, the original URI is now preserved in the member cache.

### DIFF
--- a/apis/r/R/TileDBGroup.R
+++ b/apis/r/R/TileDBGroup.R
@@ -104,13 +104,16 @@ TileDBGroup <- R6::R6Class(
       stopifnot(is_scalar_character(name))
 
       private$open("WRITE")
+      on.exit(self$close())
       tiledb::tiledb_group_remove_member(
         grp = self$object,
         uri = name
       )
-      self$close()
-      # TODO: Avoid closing/re-opening the group to update the cache
-      private$update_member_cache()
+
+      # Drop member if cache has been initialized
+      if (is.list(private$member_cache)) {
+        private$member_cache[[name]] <- NULL
+      }
     },
 
     #' @description Length in the number of members. (lifecycle: experimental)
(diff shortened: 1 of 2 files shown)


Shelnutt2 (approved):

mojaveazure (approved):

apis/r/R/TileDBGroup.R
```diff
@@ -64,6 +64,23 @@ TileDBGroup <- R6::R6Class(
       )
       # TODO: Avoid closing/re-opening the group to update the cache
       self$close()
+
+      # We manually add the new member to member_cache in order to preserve the
+      # original URI. Otherwise TileDB Cloud creation URIs are retrieved from
+      # using tiledb_group_member() in the form tiledb://namespace/uuid. In this
```
johnkerl, line 70: I would say a "creation URI" would be of the form `tiledb://namespace/s3://bucketname/path`, whereas a "post-creation URI" would be of the for you indicate here, namely, `tiledb://namespace/uuid`.

I don't think "creation URI" or "post-creation URI" are particularly standard terms, but, I don't think `tiledb://namespace/uuid` is something we should be calling a "creation URI".

I'd also at a clarifying bit to your comment, namely, _why_ it's not possible to append new children: please point out explicitly on the comment that this is where we _do_ need URIs of the form `tiledb://namespace/s3://bucketname/path`.

apis/r/R/TileDBGroup.R
```diff
@@ -64,6 +64,23 @@ TileDBGroup <- R6::R6Class(
...
+      # form it's not possible to append new children, which is necessary during
+      # ingestion.
+      if (is.null(private$member_cache)) private$member_cache <- list()
+      private$member_cache[[name]] <- list(
+        type = tiledb::tiledb_object_type(object$uri),
+        uri = object$uri,
+        name = name
+      )
+
+      # We still need to update member_cache to pick-up existing members.
```
johnkerl, line 80: pick-up -> pick up

johnkerl (commented):
Sorry, I didn't see this until after merge.

Issue #1271, opened 2023-04-21
[r] [bug] `SOMAExperimentAxisQuery` `$X()` `$obs()` and `$var()` should  have an argument `iterated`


pablo-gar:
**Describe the bug**
[r] [bug] [`SOMAExperimentAxisQuery`](https://github.com/single-cell-data/TileDB-SOMA/blob/02e0b2f9d7f728bfef7ee8d2e2df326f53c871f8/apis/r/R/SOMAExperimentAxisQuery.R#L28) `$X()` `$obs()` and `$var()` should  have an argument `iterated` hat gets passed to `read_arrow_table`

Otherwise there are no  low-latency reading capabilities o those methods from an Experiment Query.

johnkerl:
Duplicate of #1285

PR #1273, opened 2023-04-21
[r] Refactor readers for `X`, `obs`, and `var` of `SOMAExperimentAxisQuery`

pablo-gar:
**Issue and/or context:**

#1271

### DIFF
--- a/apis/r/R/SOMAExperimentAxisQuery.R
+++ b/apis/r/R/SOMAExperimentAxisQuery.R
@@ -63,23 +63,39 @@ SOMAExperimentAxisQuery <- R6::R6Class(
 
     #' @description Retrieve obs [`arrow::Table`]
     #' @param column_names A character vector of column names to retrieve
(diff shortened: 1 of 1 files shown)


pablo-gar:
Converted to draft -- testing `iterator = TRUE` for any of its uses, my current environment hangs for my >10 min when calling within or outside the `ExperimentAxisQuery`

pablo-gar:
Behavior for hanging for a long time when `iterator=TRUE` and large SOMA objects is not an outcome of this PR, any improvement to this have to be done at the level of `SOMADataFrame$read` and `SOMASparseNDArray$read`.

eddelbuettel (commented):
Looks good per se as the changes but could you add a unit test for a large enough object (or by dialing default buffer size down) showing that one (or more) of these accessors do iterate correctly?

pablo-gar:
superseded by #1274

PR #1274, opened 2023-04-24
[r] Add iterator classes

pablo-gar:
## Summary

address #1285 and #1348

* Implements read iterators in a similar way as python API.
* Removes iteration capabilities from SOMADenseNDArray. 
* Expands matrix zero-based shim introduced by @mlin. It now supports arithmetic operations and `Matrix::DenseMatrix` 
* SOMADataFrame and SOMASparseNDArray use correspoding read iterators.
   * Final shape of exported sparse matrices is the same as `SOMASparseNDArray$shape()`
* Refractors code to accommodate  the changes above. 
* Updates docs.
* Updates tests.

## Notes for reviewers

* @eddelbuettel @aaronwolen Please review changes to:
   * `SOMADataFrame`
   * `SOMASparseNDArray`
   * `SOMADenseNDArray`
   * `SOMAArrayBase`
   * `ReadIter`
   * `SparseReadIter`
   * `TableReadIter`
   * `utils-readerTransformers`

* @mlin  please review changes to:
   * `SOMASparseNDArray` and in particular `$read_sparse_matrix_zero_based`.
   * `utils-readerTransformers` and in particular `arrow_table_to_sparse`
   * `utils-matrixZeroBasedView`

## Usage example (May 24)

```r
library(tiledbsoma)

ctx <- tiledbsoma::SOMATileDBContext$new(
   config = c("soma.init_buffer_bytes" = "300")
)
 
human_experiment = load_dataset("soma-exp-pbmc-small", tiledbsoma_ctx = ctx)

###
# X reading (SOMASparseNDArray)

x_sparse <- human_experiment$ms$get("RNA")$X$get("data")
    
# Arrow
sparse_arrow <- x_sparse$read()$tables()

sparse_arrow$read_next()
# Table
# 37 rows x 3 columns
# $soma_dim_0 <int64 not null>
# $soma_dim_1 <int64 not null>
# $soma_data <double not null>

sparse_arrow$concat()
# Table
# 4419 rows x 3 columns
# $soma_dim_0 <int64 not null>
# $soma_dim_1 <int64 not null>
# $soma_data <double not null>

# Sparse zero_based view
sparse <- x_sparse$read()$sparse_matrix(zero_based = T)

sparse$read_next()
# Non-mutable 0-based 'view' class for matrices.
#  To get 1-based matrix use `x$one_based_matrix` or `as.one.based(x)`

sparse$concat()
# Non-mutable 0-based 'view' class for matrices.
#  To get 1-based matrix use `x$one_based_matrix` or `as.one.based(x)`

###
# obs reading (SOMADataFrame)

ctx <- tiledbsoma::SOMATileDBContext$new(
   config = c("soma.init_buffer_bytes" = "30000000")
)
 
human_experiment = SOMAExperimentOpen("./apis/python/notebooks/data/sparse/pbmc3k/", tiledbsoma_ctx =ctx)

# Full data frame
df <- human_experiment$obs$read()
df$read_next()
# Table
# 2638 rows x 6 columns
# $soma_joinid <int64 not null>
# $obs_id <large_string not null>
# $n_genes <int64 not null>
# $percent_mito <float not null>
# $n_counts <float not null>
# $louvain <large_string not null>

df$read_next()
# NULL
# Warning message:
# In df$read_next() : Iteration complete, returning NULL

# Iterated 
ctx <- tiledbsoma::SOMATileDBContext$new(
   config = c("soma.init_buffer_bytes" = "300")
)
 
human_experiment = SOMAExperimentOpen("./apis/python/notebooks/data/sparse/pbmc3k/", tiledbsoma_ctx =ctx)
iterator <- human_experiment$obs$read(iterated = TRUE)
iterator$read_next()
# Table
# 18 rows x 6 columns
# $soma_joinid <int64 not null>
# $obs_id <large_string not null>
# $n_genes <int64 not null>
# $percent_mito <float not null>
# $n_counts <float not null>
# $louvain <large_string not null>

iterator$concat() 
# Table
# 2620 rows x 6 columns
# $soma_joinid <int64 not null>
# $obs_id <large_string not null>
# $n_genes <int64 not null>
# $percent_mito <float not null>
# $n_counts <float not null>
# $louvain <large_string not null>
```


mlin:
@pablo-gar can you talk me through the need for arithmetic ops on matrixZeroBasedView?

I had a rationale for supporting only essential operations on it, namely that this forces user to explicitly use `as.one.based()` to do anything else with it, thus minimizing opportunities to implicitly mix zero- and one-based matrices in expressions. 

I'm open to the possibilities that (i) the zero-based arithmetic ops are essential, or (ii) I should give up on that rationale -- just would like to hear more to either end.

eddelbuettel:
@pablo-gar:  On 'git checkout pablo-gar/r_read_iterators; ./cleanup; R CMD INSTALL .` I hit an immediate speed bump

```sh
installing to /usr/local/lib/R/site-library/00LOCK-r/00new/tiledbsoma/libs
** R
** inst
** byte-compile and prepare package for lazy loading
** help
Error : EphemeralCollection.Rd:7: undefined exports: DenseReadIter
ERROR: installing Rd objects failed for package ‘tiledbsoma’
* removing ‘/usr/local/lib/R/site-library/tiledbsoma’
* restoring previous ‘/usr/local/lib/R/site-library/tiledbsoma’
Warning message:
In install.packages(pkgs = f, lib = lib, repos = if (isMatchingFile(f)) NULL else repos) :
  installation of package ‘.’ had non-zero exit status
edd@rob:~/git/tiledb-soma/apis/r(pablo-gar/r_read_iterators)$ 
```

eddelbuettel:
Easy enough to fix via

```sh
edd@rob:~/git/tiledb-soma/apis/r(pablo-gar/r_read_iterators)$ git diff
diff --git a/apis/r/NAMESPACE b/apis/r/NAMESPACE
index 76adcc03..9dfd9b77 100644
--- a/apis/r/NAMESPACE
+++ b/apis/r/NAMESPACE
@@ -22,7 +22,7 @@ S3method(write_soma,TsparseMatrix)
 S3method(write_soma,data.frame)
 S3method(write_soma,matrix)
 export(ConfigList)
-export(DenseReadIter)
+#export(DenseReadIter)
 export(EphemeralCollection)
 export(EphemeralExperiment)
 export(EphemeralMeasurement)
edd@rob:~/git/tiledb-soma/apis/r(pablo-gar/r_read_iterators)$ 
```

eddelbuettel (commented):
It's a large PR, and it currently borks tests so no way to approve it yet -- but I like it!

I am wondering if we should simplify and reduce its scope a little and take e.g. the changes to @mlin 's zero-based view out into a different PR ?

The addition of the iterator looks good (in principle) to me and does deliver what it set out to do.

apis/r/R/SparseReadIter.R
```diff
@@ -0,0 +1,80 @@
...
+        # Initiate super class
+          super$initialize (uri = uri, config = config, colnames = colnames, qc = qc,
+                            dim_points = dim_points, loglevel = loglevel)
+    
+          private$repr <- repr
+          
+          # Get max soma dims for indeces via tiledb
+          tiledb_array <- tiledb::tiledb_array(uri)
+          tiledb::tiledb_array_open(tiledb_array, type = "READ")
+          max_soma_dim_0 <- as.integer(max(tiledb::tiledb_array_get_non_empty_domain_from_index(tiledb_array, 1)))
```
pablo-gar, line 43: assert this is a 32-byte AND 

get  max_soma_dom_0 - min_soma_dim_0 as the worst-case cardinality

pablo-gar:
> Did you forget to add a file? I see this in your TODO list so maybe your use in documentation 'got ahead' ?

Yes apologies! It was the SOMADenseArray iterator that I did not include in the commit, which we will not implement in any case. Will fix!

pablo-gar:
> @pablo-gar can you talk me through the need for arithmetic ops on matrixZeroBasedView?
> 
> I had a rationale for supporting only essential operations on it, namely that this forces user to explicitly use `as.one.based()` to do anything else with it, thus minimizing opportunities to implicitly mix zero- and one-based matrices in expressions.
> 
> I'm open to the possibilities that (i) the zero-based arithmetic ops are essential, or (ii) I should give up on that rationale -- just would like to hear more to either end.


Not all of them are essential, only addition is essential to be able to seemingly concatenate chunks of the iterator. 

A user can quickly do `chunk_1 + chunk_2` and in fact that's what the `$concat()` method does:

 https://github.com/single-cell-data/TileDB-SOMA/blob/4534226c3b6c1dc885928f1224bab0da7a4b6d60/apis/r/R/SparseReadIter.R#L62

I can reduce the scope to only support iteration, but I think it leads to a poor user experience to not allow for the others given the relatively easy lift.

apis/r/R/SparseReadIter.R
pablo-gar: obsolete comment

apis/r/R/SOMASparseNDArray.R
```diff
@@ -179,51 +168,53 @@ SOMASparseNDArray <- R6::R6Class(
...
       log_level = "warn"
     ) {
-      private$check_open_for_read()
 
-      repr <- match.arg(repr)
       dims <- self$dimensions()
       attr <- self$attributes()
+      shape <- self$shape()
+      
+      stopifnot("'repr' must be a sinlge character string" = length(repr) == 1 | mode(repr) == "character",
```
eddelbuettel, line 185: Typo:   'single'

apis/r/R/ReadIter.R
```diff
@@ -0,0 +1,84 @@
...
+          } else {
+              rl <- sr_next(private$soma_reader_pointer)
+              return(private$soma_reader_transform(rl))
+          }
+      }
+    },
+    
+    #' @description  Concatenate remainder of iterator
+    # to be refined in derived classes
+    concat = function() {
```
johnkerl, line 67: Sorry for the very belated review -- @aaronwolen asked me to help review, and I'm happy to -- I believe my review is speaking for the both of us

There are definitely multiple ways to go here. However, given that on the Python side we have a single `read`, options of `.tables`, etc., and `concat` -- e.g. `sdf.read().tables().concat()` -- we should do the same here. It won't be ideal for everyone but it will work, it will parallel the Python implementation, and we can add some keystroke-saving syntactic-sugar functions _on top of these_ later if we want.

Let's get this `concat` going, and get rid of the `iterated` argument to `read`. Then `read` will always return an iterator (low complexity), and `tables()` et al. will transform an iterator of one type to an iterator of another, and `concat` will have the single job of doing concatenation.

Note that `apis/r/R/utils-readerTransformers.R` on this PR already has reader-transformers, so (I think and hope) it should be easy to connect them as methods on the iterator objects.

Likewise, `concat()` exists in non-stub form elsewhere on this PR and should be able to be connected as a method on the iterator objects.

If `read_next` remains, it needs to be remain as a helper method, not as public API.

apis/r/R/SOMADataFrame.R
```diff
@@ -178,7 +178,7 @@ SOMADataFrame <- R6::R6Class(
     #' @param iterated Option boolean indicated whether data is read in call (when
     #' `FALSE`, the default value) or in several iterated steps.
     #' @param log_level Optional logging level with default value of `"warn"`.
-    #' @return An [`arrow::Table`].
+    #' @return arrow::\link[arrow]{Table} or \link{TableReadIter}
```
johnkerl, line 181: As above: let's simply have this always iterate. Then `concat` is an easy tack-on to the return value from this.

apis/r/R/SOMADenseNDArray.R
```diff
@@ -114,79 +114,43 @@ SOMADenseNDArray <- R6::R6Class(
...
-                        (is.null(names(coords)) && length(coords) == length(self$dimnames())) ||
-                        (!is.null(names(coords)) && all(names(coords) %in% self$dimnames()))
-                    )
-
-          ## if unnamed (and test for length has passed in previous statement) set names
-          if (is.null(names(coords))) names(coords) <- self$dimnames()
-
-          ## convert integer to integer64 to match dimension type
-          coords <- lapply(coords, function(x) if (inherits(x, "integer")) bit64::as.integer64(x) else x)
+        coords <- private$convert_coords(coords)
```
johnkerl, line 131: I always love a helper method 😎

apis/r/R/SOMASparseNDArray.R
```diff
@@ -114,7 +114,7 @@ SOMASparseNDArray <- R6::R6Class(
     #' @param iterated Option boolean indicated whether data is read in call (when
     #' `FALSE`, the default value) or in several iterated steps.
     #' @param log_level Optional logging level with default value of `"warn"`.
-    #' @return An [`arrow::Table`].
+    #' @return arrow::\link[arrow]{Table} or \link{TableReadIter}
```
johnkerl, line 117: As above: by group discussion and decision some months ago we decided to replace `read_foo()` with separate `read()` chained with `foo()`. See #656 and #663 for context.

This needs to be just `read`, always returning Arrow table, and its return value needs conversation methods to convert to other formats.

johnkerl (changes requested):

apis/r/R/ReadIter.R
pablo-gar: Agreed with everything and I will try to get this done tonight.

> If read_next remains, it needs to be remain as a helper method, not as public API.

@johnkerl I'm not sure that the suggestion here is? Unfortunately in R there's no generic `next()` function, so I'm not sure how we could give users the ability to iterate themselves unless we:

1. make `read_next` a public method
2. OR we create function `next()` that calls internally `read_next`

The second one is a little less intuitive imo.

mojaveazure: > read will always return an iterator

I'm not sure I understand the benefit to this. While R does have [package that implements iterators](https://cran.r-project.org/package=iterators), it isn't widely used (or at all in the single-cell world). Because iterators don't exist in R, all this does is add complexity to R users. While I understand the desire to make the R API to act like the Python API, we do need to keep in mind that R is not Python and there are things that Python handles natively that just don't work as well in R

pablo-gar: I think the key is here:

> we can add some keystroke-saving syntactic-sugar functions on top of these later if we want

and imo it's not a "if we want" but more a "we need to". I think we can keep the low-level functionality in the R6 level as proposed and then we provide wrapper functions like `as_arrow_table(soma_obj)`[ see here](https://arrow.apache.org/docs/dev/r/reference/as_arrow_table.html), `as(soma_obj, "dgCMatrix")`,  `as.data.frame(soma_obj)`, etc

pablo-gar: @johnkerl I have addressed your comment with the exception of 

> If read_next remains, it needs to be remain as a helper method, not as public API.

johnkerl: @pablo-gar looking this morning -- thank you! :)

johnkerl: @pablo-gar sorry to confuse.

This looks good to me -- thank you!! :)

johnkerl (approved):

apis/r/R/ReadIter.R
pablo-gar: Awesome! I will update the docs and merge main into the PR branch

PR #1275, opened 2023-04-24
[r/ci] Update R MacOS CI, given R 4.3.0

johnkerl:
**Issue and/or context:**

We found on another R package that `R-latest.pkg` doesn't exist. Update here thanks to @eddelbuettel .


eddelbuettel (approved):
Looks good to me

PR #1279, opened 2023-04-24
[r] Add SOMA Experiment Query -> SingleCellExperiment outgestor

### LINKED ISSUE #1276: [r] Support Bioconductor query/outgest

mojaveazure:
Add support for creating `SingleCellExperiment` objects from SOMA experiment queries; all done using SingleCellExperiment as a suggested depdendency

Implemented SOMA methods:
- `SOMAExperimentAxisQuery$to_single_cell_experiment()`: allows loading in entire `SingleCellExperiment` objects

Future work:
- Add `SingleCellExperiment` -> SOMA ingestor via `write_soma()`
- Port `$to_single_cell_experiment()` to the `SOMAExperiment` level to allow loading of alternate experiments

Notes:
~Will not merge until support for core 2.16 is merged in~ Resolved

resolves #1276


apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -705,6 +705,259 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+    #'
+    to_single_cell_experiment = function(
+      X_layers = NULL,
+      obs_index = NULL,
+      var_index = NULL,
+      obs_column_names = NULL,
+      var_column_names = NULL,
+      obsm_layers = NULL,
+      obsp_layers = NULL,
+      varp_layers = NULL
```
johnkerl, line 731: Is there _never_ `varm`??

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -705,6 +705,259 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+          is.character(x)
+        )
+        if (is.null(names(x))) {
+          return(stats::setNames(nm = x))
+        }
+        unnamed <- !nzchar(names(x))
+        names(x)[unnamed] <- x[unnamed]
+        return(x)
+      }
+      # Because SingleCellExperiment is broken
```
johnkerl, line 767: I know you know what this means

Please add a sentence or two for the benefit of the rest of us :)

apis/r/R/utils-sce.R
```diff
@@ -0,0 +1,43 @@
+.anndata_to_sce_rd <- function(x) {
```
johnkerl, line 1: Let's please rename to `utils-single-cell-experiment.R`

johnkerl (commented):
LGTM (modulo comments below), although I'm not R-tastic enough to accept. (I will say unit-test coverage looks quite good :) )

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -764,7 +764,12 @@ SOMAExperimentAxisQuery <- R6::R6Class(
         names(x)[unnamed] <- x[unnamed]
         return(x)
       }
-      # Because SingleCellExperiment is broken
+      # Re-implement SingleCellExperiment's internal `.mat2hits` function as
+      # theirs doesn't recognize the difference between `Matrix` and `matrix`
+      # without {Matrix} being attached
+      # Also correct cases where `S4Vectors::SelfHits(from =` and
+      # `S4Vectors::SelfHits(to =` are not integers by explicitly casting to
+      # integer
```
johnkerl, line 772: ❤️

apis/r/R/SOMAExperimentAxisQuery.R
johnkerl: OK let's at least put a code comment saying it can't exist, and its omission is by thoughtful design.

We live in a polyglot world, and this is a polyglot project, and anyone with less R/Seurat expertise may be unnecessarily confused/concerned.

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -705,6 +705,264 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+        FUN = function(layer, var_ids = row.names(var), obs_ids = row.names(obs)) {
+          mat <- private$.as_matrix(self$X(layer), transpose = TRUE)
+          dimnames(mat) <- list(var_ids, obs_ids)
+          return(mat)
+        }
+      )
+      names(layers) <- names(X_layers)
+      # Load in reduced dimensions
+      ms_obsm <- tryCatch(expr = self$ms$obsm$names(), error = null)
+      skip_rd <- isFALSE(obsm_layers) || rlang::is_na(obsm_layers)
```
johnkerl, line 807: Please spell out whatever `rd` is standing for

You do save yourself a few keystrokes typing it out here, as a one-time cost, but the added clarity to every reader forever after will be a net time savings

johnkerl: ... looks like it's `skip_reduced_dims`

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -705,6 +705,264 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+        }
+        skip_rd <- TRUE
+      }
+      if (skip_rd) {
+        reduced_dims <- list()
+      } else {
+        if (isTRUE(obsm_layers)) {
+          obsm_layers <- NULL
+        }
+        obsm_layers <- obsm_layers %||% stats::setNames(
```
johnkerl, line 824: I know my feedback is delayed and I should have asked sooner but this is a wonderful opportunity for some simple little helper methods -- it would be easier to review, for sure, but also easier on future maintainers

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -705,6 +705,264 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+        if (!skip_cp && !is.null(obsp_layers)) {
+          warning(
+            "colPairs requested but none were found, skipping",
+            call. = FALSE,
+            immediate. = TRUE
+          )
+        }
+        skip_cp <- TRUE
+      }
+      if (skip_cp) {
```
johnkerl, line 864: `skip_col_pairs` please

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -705,6 +705,264 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+        if (!skip_rp && !is.null(varp_layers)) {
+          warning(
+            "rowPairs requested but none were found, skipping",
+            call. = FALSE,
+            immediate. = TRUE
+          )
+        }
+        skip_rp <- TRUE
+      }
+      if (skip_rp) {
```
johnkerl, line 896: `skip_row_pairs` please

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -705,6 +705,264 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+        SummarizedExperiment::rowData(sce) <- as(var, 'DataFrame')
+      }
+      if (ncol(obs)) {
+        SummarizedExperiment::colData(sce) <- as(obs, 'DataFrame')
+      }
+      # Validate and return
+      methods::validObject(sce)
+      return(sce)
+    },
+    #' @description Convenience method for \code{$to_single_cell_experiment()}
```
johnkerl, line 933: as above: let's make more of these convenience methods

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -800,6 +1058,112 @@ SOMAExperimentAxisQuery <- R6::R6Class(
       }
       return(mat)
     },
+    .load_df = function(df = 'obs', index = NULL, columns = NULL) {
```
johnkerl, line 1061: `df` -> `df_name` -- `df` makes it sound like it _is_ a dataframe (it isn't) leading to a little unnecessary cognitive hiccup

johnkerl: Also, passing along a pro-tip from @eddelbuettel you can have `df_name = c('obs', 'var')` -- and  then `df_name <- match.arg(df_name)` -- for added clarity to code-users as well as code-readers/code-maintainers

johnkerl: likewise `columns` sounds like they are actual columns, leading to another cognitive hiccup when it turns out to be not the case: please `column_names`

apis/r/R/roxygen.R
```diff
@@ -169,3 +169,133 @@ rd_soma_field <- function(field) {
     )
   ))
 }
+
+rd_outgest_index <- function(type = 'v3', axis = 'obs') {
```
johnkerl, line 173: Please spell out whatever rd is standing for

You do save yourself a few keystrokes typing it out here, as a one-time cost, but the added clarity to every reader forever after will be a net time savings

apis/r/R/roxygen.R
```diff
@@ -169,3 +169,133 @@ rd_soma_field <- function(field) {
...
+    label,
+    ' names; uses \\code{paste0("',
+    label,
+    '", ',
+    axis,
+    '_joinids())} by default'
+  ))
+}
+
+rd_outgest_mdnames <- function(type = 'v3', axis = 'obs') {
```
johnkerl, line 192: Similarly please spell out `md`

johnkerl (changes requested):
Several changes requested, mainly around maintainability

After this I'll run this code on some various data files & may have more to say from that point of view later

Sorry again for the belated review

apis/r/R/roxygen.R
```diff
@@ -169,3 +169,133 @@ rd_soma_field <- function(field) {
     )
   ))
 }
+
```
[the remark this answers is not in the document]
mojaveazure: It contains functions for use in documentation only; each function returns an Rd-formatted string and is designed to be called in `\Sexpr{}` in Roxygen comment blocks

apis/r/R/roxygen.R
mojaveazure: Rd stands for "R documentation", but is referred to as "Rd" in all manuals (such as ["Writing R Extensions"](https://cran.r-project.org/doc/manuals/R-exts.html#Rd-format) and ["Parsing Rd Files"](https://developer.r-project.org/parseRd.pdf)) and help pages (such as [`tools::parse_Rd()`](https://stat.ethz.ch/R-manual/R-devel/library/tools/html/parse_Rd.html)) within R itself

johnkerl: Then please add a top-of-function comment that this is an R-documentation generator.

This is quite obvious to the R-jedi (you being one) but this is a polyglottal project, with polyglottal developers as well as polyglottal users. Please just add an explanatory comment for the rest of us.

apis/r/R/roxygen.R
```diff
@@ -169,3 +169,133 @@ rd_soma_field <- function(field) {
...
+        switch(EXPR = axis, obs = 'cell', var = 'feature'),
+        '-level meta data'
+      ),
+      sce = switch(EXPR = axis, obs = '\\code{colData}', var = '\\code{rowData}')
+    ),
+    '; by default, loads all columns'
+  ))
+}
+
+rd_outgest_mlayers <- function(type = 'v3', axis = 'obsm') {
```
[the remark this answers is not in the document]
mojaveazure: `m` comes from `obsm` and `varm`; it could be `rd_outgest_m_layers`, but the `m` is a SOMA-specific suffix

johnkerl (commented):
Some testing observations.

```
experiment <- SOMAExperimentOpen('/var/s/v/ileum')
experiment <- SOMAExperimentOpen('tiledb://johnkerl-tiledb/ileum')
```

both work well -- which is grand.

```
query <- SOMAExperimentAxisQuery$new(
  experiment = experiment,
  measurement_name = "RNA"
)

> length(query$obs_joinids())
[1] 32458
> length(query$var_joinids())
[1] 22068

obj <- query$to_sce()
```

results in

```
Error: Read of 'file:///var/s/v/ileum/ms/RNA/X/data' is incomplete.
Consider increasing the memory allocation via the configuration
option 'soma.init_buffer_bytes', or using iterated partial reads.
```

Trying this

```
platform_config <- PlatformConfig$new()
platform_config$set('tiledb', 'create', 'soma.init_buffer_bytes', 32*1024**3)
experiment <- SOMAExperimentOpen('tiledb://johnkerl-tiledb/ileum', platform_config=platform_config)
```

results in the same memory issue.

I thought to try a particular cell-type (using [this](https://github.com/single-cell-data/TileDB-SOMA/blob/1.2.4/apis/python/devtools/peek-exp.py#L22-L30)) with this output

```
cell_type
enteroendocrine cell of small intestine            18
paneth cell of epithelium of small intestine       34
transit amplifying cell of small intestine         53
smooth muscle fiber of ileum                       54
mast cell                                          92
endothelial cell of lymphatic vessel               96
pericyte cell                                     121
glial cell                                        175
ileal goblet cell                                 230
progenitor cell                                   382
endothelial cell                                  565
fibroblast                                        571
enterocyte of epithelium proper of ileum          809
innate lymphoid cell                             1382
mononuclear phagocyte                            1635
B cell                                           3183
plasma cell                                      3898
native cell                                      4203
alpha-beta T cell                               14957
```

and I found that `B cell` didn't do well even with a larger memory allocation:

```
query <- SOMAExperimentAxisQuery$new(
    experiment = experiment,
    measurement_name = "RNA",
    obs_query = SOMAAxisQuery$new(value_filter = 'cell_type == "B cell"')
)
obj <- query$to_sce()
```

```
Error: Read of 'file:///var/s/v/ileum/ms/RNA/X/data' is incomplete.
Consider increasing the memory allocation via the configuration
option 'soma.init_buffer_bytes', or using iterated partial reads.
```

 It worked fine for `mast cell` (which as we see above has 92 `obs` rows):

```
query <- SOMAExperimentAxisQuery$new(
    experiment = experiment,
    measurement_name = "RNA",
    obs_query = SOMAAxisQuery$new(value_filter = 'cell_type == "mast cell"')
)
obj <- query$to_sce()
```

```
> obj
class: SingleCellExperiment
dim: 22068 92
metadata(0):
assays(1): data
rownames(22068): var0 var1 ... var22066 var22067
rowData names(16): gene_ids mito ... feature_name feature_reference
colnames(92): obs275 obs531 ... obs32244 obs32339
colData names(33): obs_id n_counts ... ethnicity development_stage
reducedDimNames(1): UMAP_HM
mainExpName: RNA
altExpNames(0):
```

I think our inability to get the query result back even with 32GB buffer size, for such a small number of cells (B cell) is worrisome.

I'm not convinced that the 32GB buffer size is getting everywhere it needs to get.

I'd be OK with that getting analyzed on this PR, or, on a follow-up PR -- either way.

And I offer to debug it with you.

johnkerl:
Concretely:

* Let's make sure (on this PR or a subsequent) the larger buffer size is being wired into everywhere it needs to go
* I'd love it if we could leverage the iterated-read on `to_single_cell_experiment` but I acknowledge SCE might not be able to _accept_ chunks of X

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -171,22 +171,14 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
-    #' cell embeddings; pass \code{FALSE} to suppress loading in any
-    #' dimensional reductions; by default, loads all dimensional
-    #' reduction information
-    #' @param varm_layers Named vector of arrays in \code{varm} to load in as
-    #' the feature loadings; names must be names of array in \code{obsm} (eg.
-    #' \code{varm_layers = c(X_pca = 'PCs')}); will try to determine
-    #' \code{varm_layers} from \code{obsm_layers}
-    #' @param obsp_layers Names of arrays in \code{obsp} to load in as
-    #' \code{\link[SeuratObject]{Graph}s}; by default, loads all graphs
+    #' @param X_layers \Sexpr[results=rd]{tiledbsoma:::rd_outgest_xlayers()}
```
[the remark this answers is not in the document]
mojaveazure: This change makes use of shared documentation generators introduced in this PR

johnkerl:
> I think our inability to get the query result back even with 32GB buffer size, for such a small number of cells (B cell) is worrisome.
> 
> I'm not convinced that the 32GB buffer size is getting everywhere it needs to get.
> 
> I'd be OK with that getting analyzed on this PR, or, on a follow-up PR -- either way.
> 
> And I offer to debug it with you.

I verified by `printf` in `libtiledbsoma/src/utils/column_buffer.cc` that indeed we are seeing the default 16MB buffer size even when I set

```

```
platform_config <- PlatformConfig$new()
platform_config$set('tiledb', 'create', 'soma.init_buffer_bytes', 32*1024**3)

experiment <- SOMAExperimentOpen('tiledb://johnkerl-tiledb/ileum', platform_config=platform_config)

query <- SOMAExperimentAxisQuery$new(
    experiment = experiment,
    measurement_name = "RNA",
    obs_query = SOMAAxisQuery$new(value_filter = 'cell_type == "B cell"')
)
obj <- query$to_sce()
```

I'll debug a bit more.

**Review by aaronwolen — CHANGES_REQUESTED** [review]
I haven't been able to fully review but wanted to post a couple of comments I had.

File: apis/r/R/SOMAExperimentAxisQuery.R [hunk 16]
```diff
@@ -326,35 +326,39 @@ SOMAExperimentAxisQuery <- R6::R6Class(
       # Use joinids if the dimension names are empty
       dim_names <- Map("%||%", dim_names, coords)
 
+      # dims <- switch(
+      #   EXPR = collection,
```
johnkerl, line 330: please GC

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -326,35 +326,39 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+      #   vapply_int(dim_names, length)
+      # )
+
       Matrix::sparseMatrix(
         i = mat_coords$i$as_vector(),
         j = mat_coords$j$as_vector(),
         x = tbl$soma_data$as_vector(),
         index1 = FALSE,
         dims = vapply_int(dim_names, length),
+        # dims = vapply_int(dim_names, function(x) max(range(x))) + 1L,
```
johnkerl, line 345: please GC

apis/r/R/SOMAExperimentAxisQuery.R
johnkerl: @mojaveazure I maintain my ask

apis/r/R/SOMAExperimentAxisQuery.R
johnkerl: i retract this

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -993,6 +1143,253 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+      obj <- if (isFALSE(column_names) || rlang::is_na(column_names)) {
+        as.data.frame(matrix(nrow = length(ids), ncol = 0L))
+      } else {
+        as.data.frame(soma_reader(column_names)$to_data_frame())
+      }
+      row.names(obj) <- ids
+      return(obj)
+    },
+    .load_m_axis = function(layer, m_axis = c('obsm', 'varm'), type = "Embeddings") {
+      # browser()
```
johnkerl, line 1186: please GC

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -993,6 +1143,253 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+        },
+        varm = {
+          soma_collection <- self$ms$varm
+          soma_joinids <- self$var_joinids()
+        }
+      )
+      soma_axis <- soma_collection$get(layer)
+      mat <- if (inherits(soma_axis, 'SOMASparseNDArray')) {
+        as.matrix(self$to_sparse_matrix(collection = m_axis, layer_name = layer))
+        # as.matrix(soma_axis$read(unname(coords))$sparse_matrix()$concat())
```
johnkerl, line 1207: please GC

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -993,6 +1143,253 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+    .load_p_axis = function(layer, p_axis = c('obsp', 'varp'), repr = c('C', 'T', 'R', 'D')) {
+      stopifnot(
+        is_scalar_character(layer),
+        is.character(p_axis),
+        is.character(repr)
+        # is_scalar_character(repr)
+      )
+      p_axis <- match.arg(p_axis)
+      repr <- match.arg(repr)
+      # switch(
```
johnkerl, line 1247: please GC

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -994,6 +1132,233 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+          soma_df <- self$var_df
+          soma_reader <- self$var
+          soma_joinids <- self$var_joinids()
+        }
+      )
+      ids <- if (is.null(index)) {
+        paste0(df_name, soma_joinids$as_vector())
+      } else {
+        index <- match.arg(arg = index, choices = soma_df$attrnames())
+        soma_reader(index)$concat()$GetColumnByName(index)$as_vector()
```
aaronwolen, line 1160: Looks like we're performing 2 separate reads here: once to retrieve ids from the designated index column and the once to read in the rest of the data. Can we consolidate to avoid the extra overhead?

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -994,6 +1132,233 @@ SOMAExperimentAxisQuery <- R6::R6Class(
       }
       return(mat)
     },
+    # Helper methods to load aspects of a measurement for the ecosystems
```
aaronwolen, line 1135: A little more documentation here would benefit other devs because there some nuances here:

- column_names is polymorphic
- a data.frame with 0 columns is returned if column_names = FALSE
- all columns are retrieved if column_names = TRUE

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -994,6 +1132,233 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+      ids <- if (is.null(index)) {
+        paste0(df_name, soma_joinids$as_vector())
+      } else {
+        index <- match.arg(arg = index, choices = soma_df$attrnames())
+        soma_reader(index)$concat()$GetColumnByName(index)$as_vector()
+      }
+      if (isTRUE(column_names)) {
+        column_names <- NULL
+      }
+      column_names <- column_names %||% setdiff(x = soma_df$attrnames(), y = index)
```
aaronwolen, line 1165: Nit: I'm all for explicitly naming arguments when dealing with functions that have many or non-obvious parameters, but for something like `setdiff()` I personally find it diminishes readability.

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -994,6 +1132,233 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+    .load_p_axis = function(layer, p_axis = c('obsp', 'varp'), repr = c('C', 'T', 'R', 'D')) {
+      stopifnot(
+        is_scalar_character(layer),
+        is.character(p_axis),
+        is.character(repr)
+      )
+      p_axis <- match.arg(p_axis)
+      repr <- match.arg(repr)
+      mat <- self$to_sparse_matrix(collection = p_axis, layer_name = layer)
+      return(switch(
```
aaronwolen, line 1233: Let's add a log statement prior to converting [CR]sparseMatrix to make it clear where the time is being spent.

apis/r/tests/testthat/test-SOMAExperiment-query-matrix-outgest.R
```diff
@@ -164,7 +164,10 @@ test_that("matrix outgest assertions", {
   # joinds are used as default dimnames if no obs/var_index is provided
   expect_identical(
     dimnames(query$to_sparse_matrix("X", "counts")),
-    list(as.character(query$obs_joinids()), as.character(query$var_joinids()))
+    lapply(
+      list(query$obs_joinids(), query$var_joinids()),
+      function(x) as.character(x$as_vector())
```
aaronwolen, line 169: Why do we need both `$as_vector()` and `as.character()`?

apis/r/tests/testthat/test-SCEOutgest.R
```diff
@@ -0,0 +1,424 @@
+test_that("Load SCE object from ExperimentQuery mechanics", {
```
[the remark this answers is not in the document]
aaronwolen: Totally agree! 

My only suggestion is that we wrap all of the code used to create these test experiments into a helper function since we're performing lagely the same steps in test-SeuratOutgest-object, test-SeuratOutgest-reduction, and now test-SCEOutgest.

Or perhaps we update the existing helper, `create_and_populate_experiment()`, to optionally populate the other slots.

apis/r/R/utils-single-cell-experiment.R
```diff
@@ -0,0 +1,78 @@
+.anndata_to_sce_reduced_dim <- function(x) {
```
aaronwolen, line 1: Since we'll likely add support for exporting to other bioc classes in the future I suggest we rename utils-single-cell-experiment.R to utils-bioc.R, as a common place for all bioconductor-related utilities.

aaronwolen (changes requested):
I ran some tests on a few different datasets and this works great! It's very close but I made several comments that should be addressed before merging. 

Re the documentation generator functions, I'm personally on the fence. On one hand, I totally appreciate the desire to reduce content duplication across the package. On the other hand, I'm a little concerned the added complexity will make it difficult for others to contribute to the docs. It's a tricky balance to strike and I'm not sure what the right call is here but I'd be interested to what others think (@eddelbuettel, @johnkerl, @pablo-gar).

apis/r/R/SOMAExperimentAxisQuery.R
mojaveazure: There's a comment in the code highlighting the ommision of `varm_layers` from the parameter list (and has been since https://github.com/single-cell-data/TileDB-SOMA/pull/1279/commits/961aad3fca749b0a0b25f9d0b84beb240285a023)
image

apis/r/tests/testthat/test-SOMAExperiment-query-matrix-outgest.R
mojaveazure: DelayedArray (dependency of SingleCellExperiment) defines `as.character.Array()`; this conflicts with casting of `arrow::Array` objects to character. Arrow only defines `as.character.ArrowDatum()` as its the parent class of `arrow::Array`; R's method resolution selects `DelayedArray:::as.character.Array()` over `arrow:::as.character.ArrowDatum()` as "Array" is a better match for "Array" than "ArrowDatum"

aaronwolen: Nice catch!

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -994,6 +1132,233 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+              "for",
+              sQuote(layer),
+              "are encoded as dense instead of sparse; all arrays should be saved as",
+              sQuote('SOMASparseNDArrays')
+            )),
+            collapse = '\n'
+          ),
+          call. = FALSE,
+          immediate. = TRUE
+        )
```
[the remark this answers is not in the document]
mojaveazure: Rebased and reworked `.load_m_axis()` to make use of expanded `to_sparse_matrix()` and `warn_if_dense()`

pablo-gar:
@mojaveazure I'd be happy to test in the [Census acceptance tests](https://github.com/chanzuckerberg/cellxgene-census/blob/main/api/r/cellxgene.census/tests/testthat/acceptance-tests.R), let me know when it's ready for that type of testing and I will go ahead and implement the tests on our end.

apis/r/R/utils-single-cell-experiment.R
```diff
@@ -0,0 +1,78 @@
...
+    base::which
+  }
+  i <- f(mat != 0, arr.ind = TRUE)
+  return(S4Vectors::SelfHits(
+    from = as.integer(i[, 1L]),
+    to = as.integer(i[, 2L]),
+    nnode = nrow(mat),
+    x = mat[i]
+  ))
+}
```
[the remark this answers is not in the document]
mojaveazure: Reworked both into an expanded `check_package()` (updated the existing function with a `version` argument)

apis/r/tests/testthat/test-SCEOutgest.R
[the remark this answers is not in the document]
mojaveazure: `create_and_populate_experiment()` has been expanded to allow creation of `obsm`, `varm`, `obsp`, and `varp` layers; the updated functionality has been used for both Seurat and SCE tests

mojaveazure:
@pablo-gar Awesome, that'd be great! I feel that this is ready for acceptance testing as it's passing all our internal tests and larger datasets to help stress-test would be fantastic. As this is R, it has all the same memory constraints as the Seurat outgestion, but functionally it should be ready to go

johnkerl (approved):
Accepting (but do not merge until @aaronwolen also accepts)

The issue with

```
platform_config <- PlatformConfig$new()
platform_config$set('tiledb', 'create', 'soma.init_buffer_bytes', 32*1024**3)
```

as noted above at https://github.com/single-cell-data/TileDB-SOMA/pull/1279#pullrequestreview-1442645344 -- I've since found (as narrated on #1423) is not with this PR but ratther with the config-setter -- the right way is not set-after-create, but rather set-on-create -- something like:

```
ctx <- tiledbsoma::SOMATileDBContext$new(c(soma.init_buffer_bytes="2000000000"))
```

aaronwolen (approved):
One more [comment](https://github.com/single-cell-data/TileDB-SOMA/pull/1279#discussion_r1263924578) but otherwise I think this is ready to go!

pablo-gar:
Two things: good news and potentially some areas to improve scalability.

## Good news

Stress tests for the outgestor in the context of [Census](https://github.com/chanzuckerberg/cellxgene-census) indicate that:

1. All of the expected queries work.
2. The running times for large queries is in the same order as magnitude as `to_seurat`: 0.8 - 1.5x  SCE/Seurat.

See [interim results for the acceptances tests](https://github.com/chanzuckerberg/cellxgene-census/tree/5ae392a0cb8a9b035ec2c8f56a11bac98d35d4bd/api/r/cellxgene.census/tests) and a summary in this [spreadsheet](https://docs.google.com/spreadsheets/d/1dlp6ExRLUsMz7U020VvXjMiNEhtIFZmXsFVIwWj_yzk/edit#gid=1030770423)


## Opportunity to improve -- loading `X` with with nnz > 2^31-1

For very large queries that have `X` matrices with nnz > 2^31-1 I believe it will be possible to create `SingleCellExperiment` objects (albeit with enough memory).

For these cases this seem to be occurring (tested on a 5M cell x 60K gene query): 

1. `X` matrix is successfully created in `dgTMatrix` format via `tables()$read()$concat()`... kudos to Matrix package.
2. However the matrix is attempted to be converted to a `dgCMatrix` sometime during the outgestion process.
3. This leads to the following error from Matrix package

```
Error: Error in `asMethod(object)`: unable to coerce from TsparseMatrix to [CR]sparseMatrixwhen length of 'i' slot exceeds 2^31-1
```

**SingleCellExperiment is capable of handling `dgTMatrix` for the expression matrices, thus I believe we should at the very least provide the user with the opportunity to stay in that format to allow for loading of these `X` matrices with nnz > 2^31-1** 


@aaronwolen @mojaveazure thoughts?

aaronwolen:
@pablo-gar, this is an excellent find! @mojaveazure, given this info I think it makes sense to default to dgT/COO unless the user explicitly asks for something else—and that option is something we could potentially tackle in a separate PR.

Issue #1277, opened 2023-04-24
[r] Support R Conda installs


johnkerl:
cc @aaronwolen @Shelnutt2 @eddelbuettel 

* [X] Remove the use of the special build of C++, i.e. remove the use of the cmake "build for R" option and remove the use of the `rcppspdlog` header. R has to use the regular SOMA C++ library.
* [X] Link against the shared library not static: #1293.
* [X] Remove the use of the bundled `.tar.gz` of the source code. No need for this, i.e TileDB-R doesn't bundle a `.tar.gz` of TileDB core C++ code.
* [ ] Add to the feedstock after the above is done.

Shelnutt2:
> 2. Sure when one exists it will be easy to adapt the link process. That is actually independent of Conda use.

Not quite independent because we must use the common shared library built by conda (libtiledbsoma). So the R package for SOMA needs to be able to support linking against the shared library we have in the env. This is is what this is referring to.

eddelbuettel:
Someone would need to provide a shared library the R package can build against.  Currently the build instructions (eg [here](https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/README.md#developer-makefile) have no such shared library.

eddelbuettel:
To witness, when I allow `bld` (which is still outside `api/r`) to populate files below /usr/local the build (after copying over headers, cmake files, pkgconfig files, ...) eventually fails on a missing link for `fmt::format()`:

```sh
Install the project...
-- Install configuration: "Release"
-- Up-to-date: /usr/local/include/span
-- Up-to-date: /usr/local/include/span/span.hpp
-- Up-to-date: /usr/local/include/thread_pool
-- Up-to-date: /usr/local/include/thread_pool/thread_pool.h
-- Up-to-date: /usr/local/include/thread_pool/producer_consumer_queue.h
-- Up-to-date: /usr/local/include/thread_pool/status.h
-- Up-to-date: /usr/local/include/tiledbsoma
-- Up-to-date: /usr/local/include/tiledbsoma/array_buffers.h
-- Up-to-date: /usr/local/include/tiledbsoma/tiledbsoma
-- Up-to-date: /usr/local/include/tiledbsoma/managed_query.h
-- Up-to-date: /usr/local/include/tiledbsoma/util.h
-- Up-to-date: /usr/local/include/tiledbsoma/logger_public.h
[...]

PR #1280, opened 2023-04-25
[r] [WIP] build against shared library libtiledbsoma

eddelbuettel:
**Issue and/or context:**

This PR enables building against a system-level `libtiledbsoma`.  It is at this point a trial balloon as we likely need to add other fallback options in case the system library is not present.  But the current (draft) proposal works locally under Linux and should work in CI.

Tracking issues: #1277 , #1293

**Changes:**

Instruct build for use `libfmt` header-only, remove all special R tags in cmake build instructions, install to `/usr/local` (in CI) and call `ldconfig` in CI.  Then instruct `pkg-config` to look for both `libtiledbsoma` and `libtiledb`.   We have a fallback artifact for the latter, the former can be built from repo sources where but _could not_ be built from a standalone tar.gz not containing the library sources.

**Notes for Reviewer:**

This is a draft at this point.


eddelbuettel:
So there is apparently some more work to do at the cmake level as it assume there was a shared library of libtiledb to also install along with libtiledbsome:

```sh
-- Install configuration: "Release"
-- Installing: /usr/local/lib/libtiledb.so.2.15
CMake Error at src/cmake_install.cmake:46 (file):
  file INSTALL cannot copy file
  "/home/runner/work/TileDB-SOMA/TileDB-SOMA/build/externals/install/lib/libtiledb.so.2.15"
  to "/usr/local/lib/libtiledb.so.2.15": Success.
Call Stack (most recent call first):
  cmake_install.cmake:47 (include)
  ```

johnkerl:
@eddelbuettel now I'm getting an R CI fail on Ubuntu.

Can you take a look?

eddelbuettel:
#1302 took care of this.

PR #1284, opened 2023-04-25
[python] Auto-consolidate fragment metadata and commits

johnkerl:
**Issue and/or context:**

Note that fragment metadata and commits are easy to consolidate and we should always do it immediately post-write. This is not to be confused with consolidating array data, which is opt-in.

Tracked in SC-25195.

**Changes:** Leverages TileDB Cloud 2.15.


johnkerl:
@Shelnutt2 @aaronwolen ping

johnkerl:
@thetorpedodog thanks, must have missed a `git add`.

thetorpedodog (approved):

PR #1290, opened 2023-04-26
Early version of the generic profiler. This profiler:

beroy:
1) Uses time tool to extract perf and memory usage 2) Uses SqllitDict to keep track of profile k/v data. 3) Designed to allow custom profiler extract stack info in the future

A simple test for the prfile:
`python main.py python tests/objects.py`

**Issue and/or context:**

**Changes:**

**Notes for Reviewer:**


dsadgat:
@beroy did you mean to add this to the data viz sprint, or some other sprint? Thanks!

beroy:
@dsadgat I don't think. @atolopko-czi  what do you suggest?

atolopko-czi:
> @dsadgat I don't think. @atolopko-czi what do you suggest?

right, not part of data viz sprint (already removed, seemingly)

profiler/data.py
```diff
@@ -0,0 +1,137 @@
...
+        )
+
+
+def extract_key_from_filename(filename):
+    match = re.match("(.+)\.run", filename)
+    assert match
+    return match.groups()[0]
+
+
+class ProfilerDict:
```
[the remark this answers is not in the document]
atolopko-czi: Any reason why `ProfilerDict` and `ProfileDB` can't be combined into one class? They seem to both be implementing the file-based db. 

For accommodate future, alternate implementions, I would recommend adding an abstract class that defines the interface for any database implementation we might want to support (file-based, SQLlite, etc.), as `class ProfilerDB(ABC): ...` And then merge these two classes into a single class as `class FileSystemProfilerDB(ProfilerDB): ...`.  See https://docs.python.org/3/library/abc.html and https://realpython.com/lessons/abstract-classes/.

profiler/data.py
```diff
@@ -0,0 +1,137 @@
...
+        print(f"db size = {len(self.profile_dict)}")
+        for key, value in self.profile_dict.items():
+            print(f"db {key} = {value}")
+
+    def addProfileData(self, data: ProfileData):
+        self.profile_dict[f"{data.now}"] = data
+
+    def extractProfileData(self, process: str, metric: str):
+        result = {}
+        for key, value in self.profile_dict.items():
```
atolopko-czi, line 130: Since we want to find profiler runs by the "process" value, the db should use the "process" value as the key, rather than a timestamp. Then it can find all runs for a given process efficiently without iterating through (worst case) all of the runs.

In fact, it seems like the file-based DB implementation could use a dir structure with process subdirs as the first "key", and timestamps file names as the second "key".

profiler/data.py
```diff
@@ -0,0 +1,137 @@
...
+    def print(self):
+        print("Printing DB")
+        print(f"db size = {len(self.profile_dict)}")
+        for key, value in self.profile_dict.items():
+            print(f"db {key} = {value}")
+
+    def addProfileData(self, data: ProfileData):
+        self.profile_dict[f"{data.now}"] = data
+
+    def extractProfileData(self, process: str, metric: str):
```
atolopko-czi, line 128: Recommend just returning the a list of `ProfileData` objects. The caller can then extract either a single metric or multiple metrics, as needed. This makes the method more symmetric with the `addProfileData` method.  

Would then rename this to indicate that it's returning _all_ profile runs for a given process, e.g. `find_all_by_process`.  Could also have a `find(process, timestamp)` method to find a specific run, but maybe a case of YAGNI.

profiler/main.py
```diff
@@ -0,0 +1,160 @@
...
+        p2out=prof2,
+        rt=real_time,
+        ut=usr_time,
+        st=sys_time,
+        max_set_size=max_resident_set_size,
+        page_reclaims=page_reclaims,
+        page_faults=page_faults,
+        cycles_elapsed=cycles_elapsed,
+        peak_memory=peak_memory_footprint,
+        other_stats=tiledb_stats,
```
atolopko-czi, line 71: would either name this `tiledb_stats` or have `other_stats` be a dict of arbitrary stats, with `tiledb` as the sole key value for now.  Until this profiler tool is moved out of TileDB-SOMA, it seems fine to just rename it to `tiledb_stats` for now.

profiler/data.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+class ProfileData:
+    """This class represent the dat stored per run"""
+
+    date = str(date.today())
+    now = str(datetime.now())
+    idx = now.find(".")
+    if idx != -1:
+        now = now[: idx - 1]
```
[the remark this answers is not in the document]
atolopko-czi: all of these vars should be set in __init__, otherwise they are class variables and won't change for new instances of ProfileData. This is just a latent bug for now, but worth fixing.

Might also have the caller just pass in the timestamp, to make this "struct"-like class a bit dumber. :)

profiler/data.py
```diff
@@ -0,0 +1,137 @@
...
+        self.cycles_elapsed = cycles_elapsed
+        self.process = process
+        self.p1out = p1out
+        self.p2out = p2out
+        self.other_stats = other_stats
+
+    def __repr__(self):
+        return (
+            f"{self.__class__} process {self.process} rt {self.rt} ut {self.ut} st {self.st} "
+            f"max mem {self.max_set_size} p1out {self.p1out} p2out {self.p2out} other_stats {self.other_stats} "
```
atolopko-czi, line 55: should display _all_ attributes; should get that for free if converting this to use `attrs` library.

profiler/data.py
atolopko-czi, line 1: recommend adding a simple cli command to query the db. or just simple docs to show an example of how to query using a python shell:

```
import data

db = data.ProfileDB()
db.extractProfileData('python tests/objects.py', 'peak_memory')
```

atolopko-czi (changes requested):
It's a good start! I added a lot of suggestions, many of which are at the API and db design levels, while some are smaller and pedantic (e.g. related to Python idiomatic usage). Worth picking off the larger design issues before focusing on smaller comments. Happy to discuss live next week, if that's helpful. 

One other broad ask is that we add doc strings and Python type hinting to all of the code. And a README to explain usage (after addressing design changes).

profiler/data.py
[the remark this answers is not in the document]
ebezzi: +1 for Andrew's suggestion. Also consider using `datetime.utcnow()` to avoid potential issues with timezones.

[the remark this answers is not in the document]
beroy: Thanks. These all make sense.

profiler/data.py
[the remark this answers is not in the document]
beroy: I had a similar idea but wanted any new implementation that implement a persistent Dict class (like SQLiteDict) can be used instead of `ProfilerDict`. As you can see ProfileDB does not really care how `ProfilerDict` is implemented. To be more clear I could call ProfileDict `FileSystemProfilerDict`. LMK what you think @atolopko-czi

profiler/data.py
```diff
@@ -0,0 +1,137 @@
...
+        return result
+
+    def __iter__(self):
+        for item in self.items():
+            yield item
+
+    def __setitem__(self, key, value):
+        filename = f"{self.path}/{key}.run"
+        with open(filename, "wb") as f:
+            pickle.dump(value, f)
```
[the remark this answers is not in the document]
beroy: I think I tried json first but it complained about complexity of the objects. Will try again.

profiler/main.py
```diff
@@ -0,0 +1,160 @@
...
+        "--prof2_output",
+        required=False,
+        help="The flamegraph output produced by prof2",
+    )
+    args = parser.parse_args()
+
+    print(f"Process to be run: {args.process}")
+    """Running the main process using time -v to get detailed memory and time"""
+    p = subprocess.Popen(
+        ["/usr/bin/time", "-al"] + args.process, stdout=PIPE, stderr=PIPE
```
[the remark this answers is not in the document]
beroy: I agree. That can be a problem. I'll add that as a future task if that's ok.

profiler/main.py
```diff
@@ -0,0 +1,160 @@
...
+            p1 = subprocess.Popen(command.split())
+            # potential example of a custom profiler here is:
+            # p1 = subprocess.Popen(["py-spy", "record", "-o", "profile.svg", "--", "python", "tests/objects.py"])
+            print(f"Running second profiler {command.split()}")
+        else:
+            print(
+                f"Second profiler {args.prof1} missing output flamegraph file location"
+            )
+
+    if args.prof2 is not None:
```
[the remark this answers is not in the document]
beroy: Good question. I just started by two at first. I will make it two and store the results as a list

profiler/tests/heavy.py
```diff
@@ -0,0 +1,43 @@
...
+it = 0
+while t:
+    print(f"Iteration {it} {t}")
+    try:
+        t = next(itr.tables())
+    except StopIteration:
+        break
+    if it == 5:
+        break
+    it = it + 1
```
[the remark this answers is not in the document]
beroy: Sorry this part of the code wasn't supposed to be in.

profiler/tests/objects.py
```diff
@@ -0,0 +1,52 @@
+import time
+
+import tiledbsoma
+
+time.sleep(1000)
```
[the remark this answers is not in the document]
beroy: You are right.

profiler/main.py
```diff
@@ -0,0 +1,169 @@
...
+
+import tiledbsoma
+from data import FileBasedProfileDB, ProfileData, ProfileDB
+
+TILEDB_STATS_FILE_PATH = "./tiledb_stats.txt"
+
+
+def parse_time_output(
+    output: str, process: str, prof1: Optional[str], prof2: Optional[str]
+):
```
atolopko-czi, line 19:
```suggestion
) -> ProfileData:
```

profiler/main.py
```diff
@@ -0,0 +1,169 @@
...
+import somacore
+
+import tiledbsoma
+from data import FileBasedProfileDB, ProfileData, ProfileDB
+
+TILEDB_STATS_FILE_PATH = "./tiledb_stats.txt"
+
+
+def parse_time_output(
+    output: str, process: str, prof1: Optional[str], prof2: Optional[str]
```
atolopko-czi, line 18: nit: reverse order of `output` and `process`

profiler/main.py
```diff
@@ -0,0 +1,169 @@
...
+            peak_memory_footprint = int(peak_memory_footprint_match.groups()[0])
+            print(f"peak_memory_footprint = {peak_memory_footprint}")
+    tiledb_stats: str = None
+    if os.path.isfile(TILEDB_STATS_FILE_PATH):
+        with open(TILEDB_STATS_FILE_PATH, "r") as f:
+            print("TileDB stats found")
+            tiledb_stats = f.read()
+    custom_out = [prof1, prof2]
+    data: ProfileData = ProfileData(
+        process=" ".join(process),
```
atolopko-czi, line 65: what is the purpose of the `join` call?

profiler/main.py
```diff
@@ -0,0 +1,169 @@
...
+    if p1 is not None:
+        p1.wait()
+    if p2 is not None:
+        p2.wait()
+
+    o: str = stderr.decode("utf-8")
+    print(f"The benchmarked process output:\n {o}")
+    # Parse the generated output from the time utility
+    data: ProfileData = parse_time_output(
+        stdout.decode("utf-8"), args.process, args.prof1_output, args.prof2_output
```
atolopko-czi, line 158: nit: it's odd to pass in the `prof{1,2}_output` to a method called `parse...`.  maybe rename method to `build_profile_data`? or add in the extra output items after this call?

profiler/main.py
```diff
@@ -0,0 +1,169 @@
...
+        if args.prof1_output is not None:
+            command = args.prof2.replace("PID", str(p.pid))
+            p2 = subprocess.Popen(command.split())
+            print(f"Running third profiler {command}")
+        else:
+            print(
+                f"Third profiler {args.prof2} missing output flamegraph file location"
+            )
+
+    stderr, stdout = p.communicate()
```
atolopko-czi, line 148:
```suggestion
    stdout, stderr = p.communicate()
```
and fix usages below

profiler/tests/cell_label_harmonization_scanvi_pan_tissue.py
```diff
@@ -0,0 +1,121 @@
+import cellxgene_census
+import numpy as np
+import pandas as pd
+
+
+def get_census_data(
```
atolopko-czi, line 6: It doesn't seem right to add a Census-centric test to this repo. I would remove this. (Census could run such a test on its own, using this profiler.)

profiler/tests/heavy.py
```diff
@@ -0,0 +1,39 @@
...
+
+import cellxgene_census
+import tiledb
+
+""" This is heavy read and process SOMA script sample developed for early testing of the profiler."""
+
+tiledb.stats_enable()
+tiledb.stats_reset()
+
+census = cellxgene_census.open_soma(census_version="2023-05-15")
```
atolopko-czi, line 12: Again, wouldn't depend upon the census for tests specific to this repo. We do need a large soma experiment to run tests over, but I would make that entirely independent of the census. Even if it meant copying a census experiment to a new s3 location "owned" by this repo/team.  But it would be better still to be able to generate a "representative" experiment programmatically with dummy data.  In particular, we could have various SOMA experiments that use different schema settings, which might provide a more controlled way test performance.

Bottom line is that I think we need to think about the performance tests as a separate profiling-related task, and independent of developing the profiler itself. We could keep the `objects.py` test here for now.

atolopko-czi (changes requested):
Thanks for all the changes--they look good. I've added some more minor comments/suggestions. 

Some larger considerations that we should discuss and resolve:

* We should remove the census-related profiler tests and figure out how best to test TileDB-SOMA with independent test data and/or move those census-related tests to the census project (see my comments for a bit more on this).
* I think each run (ProfileData) should record additional context about the instance it's running on, such as: `uname -a` output, memory info (physical/swap), cpu count. Recording the python version would be good too, although not all processes are Python-based. So perhaps the generic profiler could run a "capture context" script to capture various system/env details, including all of the above suggestions, and store in a flexible manner.  It would also be useful to record the git commit sha1, which could also be determined by the "capture context" script.
* As an expansion of the profiler_pyplot, it could also output a Pandas DataFrame for a given process, with runs as rows and metrics as columns.  The DataFrame could be obtained as an object, if called in code, or output to stdout, if called as a script. That could be a nice way to see all the metrics at once, which may be easier than individual graphs for single metrics (though I love the graphs too).
*  It would be good to have the code in this directory be checked by `pre-commit`, running ruff, mypy, etc.
* We should get more eyes on this PR, particularly from the TileDB folks. If nothing else, agreement on whether `profiler` dir is good place to put this (works for me!).

profiler/data.py
```diff
@@ -0,0 +1,114 @@
...
+    max_set_size: int = attrib()
+    page_reclaims: int = attrib()
+    page_faults: int = attrib()
+    cycles_elapsed: int = attrib()
+    peak_memory: int = attrib()
+    tiledb_stats = attrib()
+    date: str = attrib()
+    now: str = attrib()
+    somacore_version: str = attrib()
+    tiledbsoma_version: str = attrib()
```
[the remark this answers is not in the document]
beroy: I actually tried doing that but get an error when passing named argument (process in this case).

profiler/data.py
```diff
@@ -0,0 +1,114 @@
...
+        self.path = PROFILE_PATH
+
+    def __str__(self):
+        result = ""
+        if os.path.exists(self.path):
+            for process in os.listdir(self.path):
+                result += f"{process}: "
+                for run in os.listdir(f"{self.path}/{process}"):
+                    result += f"{run} "
+            return result
```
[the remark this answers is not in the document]
beroy: Good idea

profiler/main.py
beroy: Process here is a list of strings. The type hint is wrong. :(

profiler/profiler_plot.py
```diff
@@ -0,0 +1,41 @@
...
+    parser.add_argument(
+        "process",
+        nargs="+",
+        help="The main process name (and optional arguments) to be plotted",
+    )
+    parser.add_argument(
+        "-m",
+        "--metric",
+        required=False,
+        help="the metric to be plotted",
```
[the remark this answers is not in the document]
beroy: In README.md. Right?

profiler/tests/heavy.py
atolopko-czi: Looks like the wrong test file was deleted, right?

profiler/profiler_plot.py
```diff
@@ -0,0 +1,164 @@
...
+    context = data.context
+    if metric in context.keys():
+        return context[metric]
+    else:
+        raise Exception(f"context does not have the following metric {metric}")
+    return None
+
+
+def create_pandas_df(data: List[ProfileData]) -> pd.DataFrame:
+    """Create panda datat frame for all the runs of a given process
```
atolopko-czi, line 61:
```suggestion
    """Create pandas dataframe for all the runs of a given process
```

profiler/profiler_plot.py
```diff
@@ -0,0 +1,164 @@
...
+        nargs="+",
+        help="The main process name (and optional arguments) to be plotted",
+    )
+    parser.add_argument(
+        "-pd",
+        "--pandas",
+        required=False,
+        help="create pandas dataframe and prints it",
+        action="store_true",
+    )
```
atolopko-czi, line 100: Thanks for adding! the underlying method is probably more useful than the CLI option, since I'm assuming if we want to work dataframes we'd do that from a script.  As is, Pandas default column and row truncation will hide most of the output to stdout. But fine to keep this!

atolopko-czi, line 100: Thanks for adding! the underlying method is probably more useful than the CLI option, since I'm assuming if we want to work dataframes we'd do that from a script.  As is, Pandas default column and row truncation will hide most of the output to stdout. But fine to keep this!

profiler/profiler_plot.py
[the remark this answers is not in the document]
atolopko-czi: was thinking directly in the help output here, but yes README too!

profiler/README.md
```diff
@@ -0,0 +1,74 @@
+This tool is profiler for [SOMA](https://github.com/single-cell-data/SOMA/tree/main) workloads collecting both performance and memory usage across runs and alos can be used for detecting performance or memory hot spots.
```
atolopko-czi, line 1:
```suggestion
This tool is a profiler for [SOMA](https://github.com/single-cell-data/SOMA/tree/main) workloads collecting both performance and memory usage across runs and can also be used for detecting performance or memory hot spots.
```

profiler/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+tmp = sys.stdout
+my_result = StringIO()
+sys.stdout = my_result
+tiledb.stats_dump()
+tiledb.stats_disable()
+sys.stdout = tmp
+
+with open("tiledb_stats.txt", "w") as f:
+    f.write(my_result.getvalue())
```
atolopko-czi, line 71: as future work, we could provide a convenience function that encapsulates the above, for profiling scripts to reuse.

atolopko-czi (approved):
Thanks for all the changes! Approving, but see the few minor comments I've made, and a [PR](https://github.com/single-cell-data/TileDB-SOMA/pull/1473) I added with some fixes to merge in, if you approve.

I'm still thinking that storing the data as json instead of pickling could be good since it would support evolution of the data (adding new metrics columns). But we can consider that as future work.

Finally, I would just check with the greater TileDB-SOMA team that everyone is okay with merging this to the repo, as opposed to e.g. placing it in a new repo entirely.

johnkerl:
@atolopko-czi @beroy 

> Finally, I would just check with the greater TileDB-SOMA team that everyone is okay with merging this to the repo, as opposed to e.g. placing it in a new repo entirely.

I'm OK with it being in the `profiler` directory -- this is sufficient decomposition IMO.

dsadgat:
@atolopko-czi this ticket seems impossible to remove from our sprint. Is there an ETA on completion?

atolopko-czi:
> this ticket seems impossible to remove from our sprint. Is there an ETA on completion?

@dsadgat It's near the finish line.

profiler/profiler_plot.py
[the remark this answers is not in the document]
beroy: I'll go with profile.py

PR #1296, opened 2023-04-26
[python] Require TileDB-Py >= 0.21.1

johnkerl:
**Issue and/or context:** Normally in our `setup.py` we pin like `tiledb==0.21.*`. However, I found the following today:

* I have an EC2 instance I normally leave powered off. It had a recent `tiledbsoma` version, and therewith, `tiledb-py 0.21.1` 
* Today I did `pip install tiledbsoma`
* This upgraded `tiledbsoma` to 1.2.2 (recent) but it left `tiledb-py` as-is, at 0.21.1.
* However we _truly require features_ from 0.21.2 -- in particular the new `config` argument to `Group`
* I had to manually `pip install -U tiledb`. Our users shouldn't have to.

(Side note: our `tiledbsoma`-to-core dependency is fine: https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/cmake/Modules/FindTileDB_EP.cmake)

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -289,7 +289,7 @@ setuptools.setup(
         "scanpy>=1.9.2",
         "scipy",
         "somacore==1.0.2",
-        "tiledb==0.21.*",
+        "tiledb>=0.21.2",
         "typing-extensions",  # Note "-" even though `import typing_extensions`
     ],
     extras_require={


aaronwolen (approved):

thetorpedodog:
I was going to ask if you wanted to pin to 2.15.<b>2</b>+, but that is in fact what the code already does, just the commit message title refers to 2.15.1.

johnkerl:
Good catch & thanks @thetorpedodog 

I had in my head 2.15.0 -> 2.15.1 and also 0.21.1 -> 0.21.2 and end up splicing the two chaotically 😬

Issue #1294, opened 2023-04-26
[Bug] `AttributeError` in `tiledbsoma.io.from_anndata()`


maarten-devries:
**Describe the bug**

I tried to use `tiledbsoma.io.from_anndata()`. I get an `AttributeError: 'Ctx' object has no attribute '_open_timestamp_ms'`. 

**To Reproduce**
```
import tiledb
import tiledbsoma as soma
import tiledbsoma.io
import scanpy as sc

config["rest.token"] = os.environ['TILEDB_API_KEY']
ctx = tiledb.Ctx(config)

adata = sc.datasets.pbmc3k_processed() 
experiment_uri = "tiledb://OUR_S3_URL" 
tiledbsoma.io.from_anndata(experiment_uri, adata, measurement_name="RNA", context=ctx, platform_config=config)  


```

**Versions (please complete the following information):**
 - TileDB-SOMA version: 1.2.2
 - Language and language version (e.g. Python 3.8, R 4.2.2): Python 3.9.16
 - OS (e.g. MacOS, Ubuntu Linux): Ubuntu Linux
 - Note: you can use `tiledbsoma.show_package_versions()` (Python) or `tiledbsoma::show_package_versions()` (R)

**Additional context**

maarten-devries:
Perhaps I am passing in the wrong `context` here - please give me a minute while I figure that out.

johnkerl:
@maarten-devries no worries, I have a repro I can work with

maarten-devries:
Changing `ctx = tiledb.Ctx(config)`
to 
`tiledb_ctx = tiledb.cloud.Ctx()
ctx = tiledbsoma.SOMATileDBContext(tiledb_ctx=tiledb_ctx)`

resolves at least this issue.

PR #1297, opened 2023-04-26
[python] Add a run-time type check on contexts provided to `tiledbsoma.io`

johnkerl:
**Issue and/or context:** #1294 is too easy a mistake to make.


atolopko-czi (approved):
Looks reasonable, so approving, but wondering whether this is worth pushing further down into the bowels of TileDB-SOMA? Say, into the various open() calls where this arg ends up.

johnkerl:
@atolopko-czi quite right tbh

PR #1299, opened 2023-04-26
Tighten TileDB pypi dependency to be a 0.21.x version.

thetorpedodog:
It was brought up to me that using `tiledb>=0.21.2` could result in v0.22.0 matching, which is not what we want (this would cause a core version mismatch, which is Bad™). Instead, we should pin to `~=0.21.2`, which means "any v0.21.x version where x >= 2".


johnkerl (approved):
✨

PR #1303, opened 2023-04-27
[c++/r/python] Update to core 2.15.2, TileDB-Py 0.21.3, TileDB-R 0.19.1

johnkerl:
**Issue and/or context:** As these packages are now released.

**Changes:**

**Notes for Reviewer:** See also https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/17

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -289,7 +289,7 @@ setuptools.setup(
         "scanpy>=1.9.2",
         "scipy",
         "somacore==1.0.2",
-        "tiledb~=0.21.2",
+        "tiledb~=0.21.3",
         "typing-extensions",  # Note "-" even though `import typing_extensions`
     ],
     extras_require={
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -38,7 +38,7 @@ Imports:
     Matrix,
     stats,
     bit64,
-    tiledb (>= 0.19.0),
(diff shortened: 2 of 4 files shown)


johnkerl:
In the R CI fail at
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/4821095098/jobs/8586704033?pr=1303
I do not understand

> https://github.com/single-cell-data/TileDB-SOMA/actions/runs/4821095098/jobs/8586704033?pr=1303

as I thought we had TileDB-R 0.19.1 available on CRAN
https://cran.r-project.org/web/packages/tiledb/index.html

johnkerl:
I notice that TileDB-Py 0.21.3 has not yet arrived in Anaconda:
https://anaconda.org/tiledb/tiledb-py/labels

eddelbuettel (approved):
:+1:

eddelbuettel:
> Package required and available but unsuitable version: ‘tiledb’

So sorry about that.  I had hoped an installation yesterday would be enough to have `tiledb` covered today by PPM which goes into r2u for binaries but no so such luck.  We should therefore probably add the line

```
 Rscript -e "install.packages('tiledb', repos = c('https://eddelbuettel.r-universe.dev', 'https://cloud.r-project.org'))" 
```

as a shell command in the yaml.  And maybe we even want to keep it to always test with upcoming tiledb.

PR #1305, opened 2023-04-28
[r] Seurat <> SOMA Integration Improvements

mojaveazure:
- Improved messaging for `to_seurat()`

  - Silence unnecessary warnings for default arguments
  - Silence duplicate warnings emitted by `to_seurat()` and downstream methods
- Improve messaging about ingesting extended assays
- `write_soma.Seurat()` now returns the URI instead of the experiment to be more inline with Python

### DIFF
--- a/apis/r/R/SOMAExperimentAxisQuery.R
+++ b/apis/r/R/SOMAExperimentAxisQuery.R
@@ -278,7 +278,7 @@ SOMAExperimentAxisQuery <- R6::R6Class(
       ms_embed <- tryCatch(expr = self$ms$obsm$names(), error = null)
       skip_reducs <- isFALSE(obsm_layers) || rlang::is_na(obsm_layers)
       if (is.null(ms_embed)) {
-        if (!skip_reducs) {
+        if (!(skip_reducs || is.null(obsm_layers))) {
           warning("No reductions found", call. = FALSE, immediate. = TRUE)
         }
         skip_reducs <- TRUE
--- a/apis/r/R/SOMAExperimentAxisQuery.R
+++ b/apis/r/R/SOMAExperimentAxisQuery.R
@@ -3,21 +3,6 @@
 #'
 NULL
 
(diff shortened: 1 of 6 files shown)


eddelbuettel:
Quick thought:  Can you show 'before' vs 'after' examples?

mojaveazure:
Right, sorry. Currently outgestion with `query$to_seurat()` emits warnings using default values:
```R
> exp <- SOMAExperimentOpen(
+   "s3://tiledb-paulhoffman/pbmc3k-raw", 
+   tiledbsoma_ctx = SOMATileDBContext$new(c(soma.init_buffer_bytes = "45000000"))
+ )
> query <- SOMAExperimentAxisQuery$new(exp, "RNA")
> obj <- query$to_seurat(c(data = "data"))
Warning: No reductions found
Warning: No graphs found in 'obsp'
```
After this patch, outgestion with `query$to_seurat()` no longer emits these warnings when using the default values

Messaging about extended assays during ingestion from Seurat currently is as follows
```R
> library(Seurat)
> library(SeuratData)
> pbmc3k <- UpdateSeuratObject(LoadData("pbmc3k"))
> pbmc3k <- SCTransform(pbmc3k)
> class(pbmc3k[["RNA"]])
[1] "Assay"
attr(,"package")
[1] "SeuratObject"
> class(pbmc3k[["SCT"]])
[1] "SCTAssay"
attr(,"package")
[1] "Seurat"
> write_soma(pbmc3k, file.path(tempdir(), "pbmc3k-sct"))
Warning: Extended assays (eg. SCTAssay) cannot be written
to SOMAs at this time
```
Which isn't entirely correct. We can write out the standard parts of a Seurat `Assay`, just not the extended bits. After this patch, the warning is now:
```R
Warning: Extended assays (eg. SCTAssay) are not fully
supported; core Assay data has been written but
additional slots have been skipped
```

aaronwolen (approved):

Issue #1232, opened 2023-04-08
[r] Discussion: R one-based indexing vs. `soma_joinid`


mlin:
This ticket is for capturing discussions & decisions!

----

In the R ecosystem all vector and array indexing is one-based, which users/devs are highly accustomed to.

The [SOMA abstract specification](https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md) defines both dense & sparse arrays to have zero-based offset indexing. Although we need to reinterpret that slightly for R, that's not inherently problematic because it's entirely natural to treat R array offsets/positions as one-based instead.

Bigger papercuts arise from `soma_joinid`, which we define as follows:

> Every `SOMADataFrame` must contain a column called `soma_joinid`, of type `int64` and domain `[0, 2^63-1]`. The `soma_joinid` column contains a unique value for each row in the `SOMADataFrame`, and intended to act as a joint key for other objects, such as `SOMASparseNDArray`.

Note that `soma_joinid` *can* take the value 0, but *needn't* -- in principle they can be any unique numbers throughout the specified domain (although in current practice they do usually equal zero-based row numbers in the `obs`/`var` dataframes).

This will cause friction any time we have an NDArray dimension indexed by a `soma_joinid`, including all `X` matrices and the Census' feature-dataset presence matrix. Because `soma_joinid` can take the value zero, it seems we need to add one to it in order to populate a one-based dimension of an R `sparseMatrix`:

https://github.com/single-cell-data/TileDB-SOMA/blob/4e009065cdd94bb2803eac6e441e5613f268ee0a/apis/r/R/SOMASparseNDArray.R#L174-L175

Yet **this is in some sense inappropriate**, because `soma_joinid` is defined as an actual ID, not an offset/row number.

Here's an example [R vignette](http://gistpreview.github.io/?e55025fb1fae42116e41f78baf26a462) / [upstream Python notebook](https://github.com/chanzuckerberg/cellxgene-census/blob/main/api/python/notebooks/api_demo/census_dataset_presence.ipynb); where, because the R `sparseMatrix` dimensions are effectively `soma_joinid+1`, to access it by `soma_joinid` found from other dataframes, we need to explicitly add one to those IDs.

----

A few strawperson ideas for discussion:

1. Implicitly add one to `soma_joinid` *everywhere*, including dataframes as well as `sparseMatrix` dimensions. 
    * Reservations about implicitly changing an *ID*, not an offset/row number.
[...]

PR #1313, opened 2023-04-30
[r] SOMASparseNDArray: sparseMatrix zero-based shim to facilitate soma_joinid lookup

mlin:
Extended discussion: #1232

To load a one-based `Matrix::sparseMatrix` with the contents of a zero-based `SOMASparseNDArray`, we add one to the offsets on each dimension. However, these dimensions are usually populated with `soma_joinid` intended to match the `soma_joinid` column in the `obs` & `var` data frames, and shifting them by one makes this join operation very error-prone. 

Here we introduce a minimal shim for `sparseMatrix` that provides matrix access with zero-based indexes. To make this explicit for the user, we rename `SOMASparseNDArray$read_sparse_matrix()` to `SOMASparseNDArray$read_sparse_matrix_zero_based()`. The shim supports only minimal access operations, which is intentional to prevent "mixing" it with conventional one-based objects. If needed, the fully-featured `sparseMatrix` is recovered by calling `as.one.based()` on the shim. Thus, the default behavior is to match `soma_joinid` but an advanced user can explicitly change to one-based indexing for further use in R.

This is a refinement of the first attempt #1306 which was far more complex and error-prone, attempting a subclass with selected methods overridden instead of the distinct wrapper shim shown here.


mlin:
@eddelbuettel Updated NAMESPACE and added the Rd's for the new methods. Perhaps we leave updating the existing Rd's to a followup since, as you noted, it's a lot of unrelated changes.

apis/r/NAMESPACE
```diff
@@ -1,8 +1,12 @@
 # Generated by roxygen2: do not edit by hand
 
+S3method("[",matrixZeroBasedView)
+S3method("[<-",matrixZeroBasedView)
```
eddelbuettel, line 4: Thanks for adding it.

eddelbuettel (approved):
I think this is good to go.

mlin:
@eddelbuettel thanks!

@mojaveazure @aaronwolen double-checking just since this was such a tricky topic -- any last comments/concerns?

PR #1316, opened 2023-05-01
[r/python] Seurat to AnnData test

ebezzi:
**Issue and/or context:**
#1184 

**Changes:**
Adds a Seurat -> SOMA -> AnnData test

**Notes for Reviewer:**
Relatively simple test that starts from a pre-generated RDS artifact, converts it to SOMA and then to AnnData. Assertions are done in R since we can't read the original Seurat. To do them in Python, we could use Reticulate to read it, but it's probably better to keep it simple.


johnkerl (approved):

apis/system/tests/test_write_seurat_read_anndata.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @pytest.fixture(scope="class")
+    def anndata_from_seurat(self):
+        base_script = f"""
+        library("tiledbsoma")
+        library("SeuratObject")
+
+        seuratObject <- readRDS("apis/system/tests/data/pbmc-small.rds")
+        write_soma(seuratObject, "{self.uri}")
+        """
```
mojaveazure, line 26: Why are we bundling `pbmc_small` in this repo instead of using the copy provided by SeuratObject?

```R
library(SeuratObject)

data("pbmc_small")
write_soma(pbmc_small, "{self.uri}")
```

mojaveazure (changes requested):

apis/system/tests/test_write_seurat_read_anndata.py
ebezzi: Nice! I didn't know it existed. Changed.

mojaveazure (approved):

Issue #1317, opened 2023-05-01
[c++] Installer is installing `.cc` files


johnkerl:
Following #1300 we have non-install of non-public headers (which is good).

However, we're getting installs of `.cc` files:

```
$ ./scripts/bld --prefix=testing
```

```
$ find testing -type f | sort
testing/bin/tdbsoma
testing/include/cli/arrow_adapter.h
testing/include/cli/carrow.h
testing/include/cli/cli.cc
testing/include/soma/managed_query.cc
testing/include/soma/managed_query.h
testing/include/soma/soma_array.cc
testing/include/soma/soma_array.h
testing/include/span/span.hpp
testing/include/tiledbsoma/logger_public.h
testing/include/tiledbsoma/tiledbsoma
testing/include/tiledbsoma/tiledbsoma_export.h
testing/include/utils/array_buffers.h
testing/include/utils/column_buffer.cc
testing/include/utils/column_buffer.h
testing/include/utils/common.h
testing/include/utils/logger.cc
testing/include/utils/logger.h
testing/include/utils/stats.cc
testing/include/utils/stats.h
testing/include/utils/util.cc
testing/include/utils/util.h
testing/include/utils/version.cc
testing/include/utils/version.h
testing/lib/cmake/tiledbsoma/TileDBSomaConfig.cmake
testing/lib/cmake/tiledbsoma/TileDBSomaTargets-release.cmake
testing/lib/cmake/tiledbsoma/TileDBSomaTargets.cmake
testing/lib/libtiledbsoma.dylib
testing/lib/pkgconfig/tiledbsoma.pc
```

PR #1301, opened 2023-04-27
[c++] Reorganize directory to separate public and private headers

nguyenv:
**Issue and/or context:**

#1300 

This is one of the several changes made in #1259 that will be broken off into smaller PRs.

This PR sets up the new directory structure such that, in the future, all public headers are contained in `src/tiledbsoma`. Although for now, we make all headers public as to not break the Python and R APIs. This allows us to merge in code with smaller PRs instead of all at once. After the Python and R bindings are refactored using only the public C++ API, the headers not residing in `src/tiledbsoma` will be made internal only.

PR #1319, opened 2023-05-01
[c++] Do not install source files

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/1317

The source files are getting installed alongside the header files as of #1301.

**Changes:**

Explicitly list out header files to install instead of using wildcard matching.

### DIFF
--- a/libtiledbsoma/src/CMakeLists.txt
+++ b/libtiledbsoma/src/CMakeLists.txt
@@ -174,6 +174,10 @@ install(DIRECTORY
   ${CMAKE_CURRENT_SOURCE_DIR}/utils
   ${CMAKE_CURRENT_SOURCE_DIR}/external/include/span
   TYPE INCLUDE
+  FILES_MATCHING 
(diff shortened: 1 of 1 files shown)


libtiledbsoma/src/CMakeLists.txt
```diff
@@ -174,6 +174,10 @@ install(DIRECTORY
   ${CMAKE_CURRENT_SOURCE_DIR}/utils
   ${CMAKE_CURRENT_SOURCE_DIR}/external/include/span
   TYPE INCLUDE
+  FILES_MATCHING
```
Shelnutt2, line 177: It would be better to not install the directories but to only set the install targets to the specific files we want to include. This is how TileDB handles cmake installs for headers.

johnkerl (approved):
LGTM and it looks like @Shelnutt2 's concerns are addressed :)

Issue #1248, opened 2023-04-11
[c++] Add member functions to `SOMAGroup`


nguyenv:
Split-off from https://github.com/single-cell-data/TileDB-SOMA/issues/1240

This includes the implementing following functions from `tiledb::Group`:

-  `member`
-  `has_member`
-  `add_member`
-  `remove_member`
-  `member_count`

nguyenv:
#1339

Issue #1247, opened 2023-04-11
[c++] Add metadata functions to `SOMAGroup`


nguyenv:
Split-off from https://github.com/single-cell-data/TileDB-SOMA/issues/1240

This includes the implementing following functions from `tiledb::Group`:

- `get_metadata`
- `has_metadata`
- `metadata_num`
- `set_metadata`
- `delete_metadata`

nguyenv:
#1339

PR #1339, opened 2023-05-05
[c++] Addition of `SOMAGroup`

nguyenv:
**Issue and/or context:**
https://github.com/single-cell-data/TileDB-SOMA/issues/1247
https://github.com/single-cell-data/TileDB-SOMA/issues/1248
https://github.com/single-cell-data/TileDB-SOMA/issues/1122

**Changes:**

Addition of internal `SOMAGroup` class that will be used to build `SOMACollection`, `SOMAExperiment`, and `SOMAMeasurement`.

### DIFF
--- a/libtiledbsoma/src/soma/soma_group.h
+++ b/libtiledbsoma/src/soma/soma_group.h
@@ -34,7 +34,7 @@
 #define SOMA_GROUP
 
 #include <future>
-#include <stdexcept> 
+#include <stdexcept>
 #include <tiledb/tiledb>
 #include <tiledb/tiledb_experimental>
 
--- a/libtiledbsoma/src/soma/soma_group.h
+++ b/libtiledbsoma/src/soma/soma_group.h
@@ -1,0 +1,285 @@
+/**
+ * @file   soma_group.h
+ *
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2023 TileDB, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @section DESCRIPTION
+ *
+ *   This declares the SOMAGroup class.
+ */
+
+#ifndef SOMA_GROUP
+#define SOMA_GROUP
+
+#include <future>
+#include <stdexcept>  // for windows: error C2039: 'runtime_error': is not a member of 'std'
+#include <tiledb/tiledb>
+#include <tiledb/tiledb_experimental>
+
+#include "../utils/common.h"
+
+namespace tiledbsoma {
+using namespace tiledb;
+
+class SOMAGroup {
+   public:
+    //===================================================================
+    //= public static
+    //===================================================================
+
+    /**
+     * @brief Open a group at the specified URI and return SOMAGroup
+     * object.
+     *
+     * @param mode TILEDB_READ or TILEDB_WRITE
+     * @param uri URI of the group
+     * @param name Name of the group
+     * @param platform_config Config parameter dictionary
+     * @param timestamp Timestamp
+     * @return std::unique_ptr<SOMAGroup> SOMAGroup
+     */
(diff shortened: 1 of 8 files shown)


johnkerl (commented):
Whether on this PR or on a subsequent we will need an additional indrection layer.

Context:

* The [SOMA spec](https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md#object-lifecycle) requires that:
  * Whether a SOMA array is opened for read or write, while `read` is only allowed in `read` mode and `write` is only allowed in `write` mode, in either mode one must be able to access the array schema and array metadata
  * Likewise, whether a SOMA group is opened for read or write, while adding/removing members is only allowed in `write` mode, in either mode one must be able to access the (current) member list and metadata
* For arrays: TileDB-Py, TileDB-R, and TileDB Core support this
  * Example: `tiledb.open(/ileum/obs', 'r').schema` works and so does `tiledb.open(/ileum/obs', 'w').schema`
* For groups: TileDB-Py, TileDB-R, and TileDB core do _not_ support this
  * `tiledb.Group('s/v/ileum', 'w')['obs']` and `tiledb.Group('s/v/ileum', 'w').meta` do not return results
  * See also SC-28610
* Since core doesn't let us read members-list or metadata when the group is opened for write, we have to do it ourselves in-package. The current TileDB-SOMA-Py uses an indirection layer:
  * https://github.com/single-cell-data/TileDB-SOMA/blob/1.2.3/apis/python/src/tiledbsoma/_tdb_handles.py#L85-L89
  * https://github.com/single-cell-data/TileDB-SOMA/blob/1.2.3/apis/python/src/tiledbsoma/_collection.py#L137
  * https://github.com/single-cell-data/TileDB-SOMA/blob/1.2.3/apis/python/src/tiledbsoma/_tdb_handles.py#L229-L233
  * @thetorpedodog has more context here, having created the TileDB-SOMA-Py group I/O indirection layer

That needs to be done for C++ as well: implementation as well as unit-test cases to validate it.

If that work isn't done on this PR (which is OK with me) then file a tracking task to make sure it does get done.

nguyenv:
I see. Thank for for the feedback. I will revert back to draft and make changes to this PR.

nguyenv:
I worked on the above yesterday, but realize now I'm going to need to open a subsequent PR further off in the future to do the above. 

Right now the metadata value is return as a `const void*`. This gets destroyed when we close the group. So when the group is opened in write mode, the pointer does not exist, and we cannot get the value. In the future, the metadata values will be returned as an `ArrowTable`. I think once we do that, there won't be a problem regardless of whether it's in read or write mode because we'll be able to store the value instead of the pointer.

johnkerl:
> I worked on the above yesterday, but realize now I'm going to need to open a subsequent PR further off in the future to do the above.

All good and thanks @nguyenv ! Just please be sure this to-do, which you've identified here as a comment on a PR, is tracked in GitHub Issues.

nguyenv:
Listed as #1351

libtiledbsoma/src/soma/soma_group.h
```diff
@@ -0,0 +1,285 @@
...
+     * @param relative is the URI relative to the SOMAGroup location
+     * @param name of member
+     */
+    void add_member(
+        const std::string& uri, bool relative, const std::string& name);
+
+    /**
+     * Get the number of members in the SOMAGroup.
+     */
+    uint64_t get_length() const;
```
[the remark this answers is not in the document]
nguyenv: https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md#somacollection-1

I have this as length because the spec specifies a `get length` operation for `SOMACollection`, but I can change this to count if that name is better.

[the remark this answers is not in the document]
nguyenv: (`SOMACollection` is going to be implemented by `SOMAGroup`.)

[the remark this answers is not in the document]
eddelbuettel: Spec wins over Dirk, clearly.

libtiledbsoma/test/unit_soma_group.cc
```diff
@@ -0,0 +1,226 @@
...
+    REQUIRE(sg->has_member("subarray") == true);
+    sg->close();
+}
+
+TEST_CASE("SOMAGroup: metadata") {
+    auto ctx = std::make_shared<Context>();
+
+    std::string uri = "mem://unit-test-group";
+    Group::create(*ctx, uri);
+    auto sr = SOMAGroup::open(TILEDB_WRITE, ctx, uri, "metadata", 1);
```
beroy, line 196: What does sr stand for?

libtiledbsoma/test/unit_soma_group.cc
```diff
@@ -0,0 +1,226 @@
...
+
+    std::string uri_main_group = "mem://main-group";
+    Group::create(*ctx, uri_main_group);
+
+    std::string uri_sub_group = "mem://sub-group";
+    Group::create(*ctx, uri_sub_group);
+
+    auto [uri_sub_array, expected_nnz] = create_array("mem://sub-array", *ctx);
+
+    auto sg = SOMAGroup::open(TILEDB_WRITE, ctx, uri_main_group, "metadata", 1);
```
beroy, line 164: What does sg stand for?

libtiledbsoma/src/utils/common.h
```diff
@@ -39,6 +39,10 @@
 
 namespace tiledbsoma {
 
+using MetadataValue =
+    std::tuple<std::string, tiledb_datatype_t, uint32_t, const void*>;
+enum MetadataInfo { key = 0, dtype, num, value };
```
beroy, line 44: Given that `MetaDataInfo` is only used in test, wouldn't it make sense to move it there?

nguyenv:
Oh I'm sorry @beroy - I did not see your comments until just now. I will read them now and make a subsequent PR with necessary changes.

libtiledbsoma/test/unit_soma_group.cc
nguyenv: This should be changed. It use to stand for `SOMAReader` but now we change that to `SOMAGroup` and `SOMAArray`.

libtiledbsoma/test/unit_soma_group.cc
nguyenv: `SOMAGroup`. I will rename to something more clear.

libtiledbsoma/src/utils/common.h
nguyenv: I am planning on using this in code outside of the unit tests soon.

Issue #1346, opened 2023-05-08
[r] Implement stateful open and close


johnkerl:
All the items marked "1325" on #1241 

* [X] `mode()` accessor exists
  * [X] Returns `"READ"` after array/group opened for read
  * [X] Returns `"WRITE"` after array/group opened for write
  * [X] Returns `"WRITE"` calling `create`
  * [X] Returns `"CLOSED"` after array/group is closed
* [X] For each object type: if opened for reading: `write` raises an error
* For each object type: if opened for writing:
  * [X] `read()` raises an error
  * [X] for all types: metadata can be inspected
  * [X] for arrays: schema, and data derived from the schema, can be inspected
  * [X] for collections: members of the collection are accessible
* After `close`:
  * [X] `read` raises an error
  * [X] `write` raises an error
* [X] The `open` methods return an object opened for read (default, or `mode="r"`) or write (`mode="w"`)
* [X] Whether mode is read or write, `open` raises an error if the object does not exist
* [X] The `create` methods return an object opened in write mode
* [X] When we add a new SOMA object to a collection, that child object is implicitly opened in write mode
* [X] When we get a SOMA object from a collection, that child object is implicitly opened in read mode
* [X] When the collection is closed, all such implicitly opened child objects are also to be closed

Follow-up on #1378

mlin:
@johnkerl a few important requirements for collections:

* when we add a new SOMA object to a collection, that child object is implicitly opened in write mode
* when we get a SOMA object from a collection, that child object is implicitly opened in read mode
* when the collection is closed, all such implicitly opened child objects are also to be closed

johnkerl:
The only remaining issue is tracked on #1378; closing this issue.

PR #1359, opened 2023-05-10
[r] Re-order methods in `ephemeral.R`

johnkerl:
This is a line-count-reducing PR which will help simplify PR #1325 for issue #1346.

It groups methods in `R/ephemeral.R` in what feels like a more intuitive order.

This is not strictly necessary of course; but on #1325 I found this a friction-reducer while reasoning about behavior of the ephemeral classes.


aaronwolen (approved):
I like this, thanks!

Issue #1348, opened 2023-05-09
[Bug] [r] `SOMASparseNDArray$read_sparse_matrix_zero_based()` raises error when calling from a sparse matrix created by python's `from_anndata`


pablo-gar:
**Describe the bug**

Hey @mlin doing some work with the R iterators I think I have uncovered an unexpected issue for `SOMASparseNDArray$read_sparse_matrix_zero_based()`

I can see that in your tests you used a mock `SOMASparseNDArray` with a shape c(10,10)

https://github.com/single-cell-data/TileDB-SOMA/blob/d8a6fc9982a359ec02d5b39da3207a3185556ece/apis/r/tests/testthat/test-SOMASparseNDArray.R#L4

However, when building a `SOMAExperiment` from python using the `from_anndata` method we end up with `SOMASparseNDArray`  that is of shape `9223372036854773760` by `9223372036854773760`, because of this 
https://github.com/single-cell-data/TileDB-SOMA/issues/1327#issuecomment-1533476037

Then if I try to read the `X` sparse matrix of the `SOMAExperiment`  created with `from_anndata` you hit a numeric overflow issue because the way we use shape to create the `Matrix::SparseMatrix` .

The actual error is from this
https://github.com/single-cell-data/TileDB-SOMA/blob/d8a6fc9982a359ec02d5b39da3207a3185556ece/apis/r/R/SOMASparseNDArray.R#L203

Which triggers this
https://github.com/cran/Matrix/blob/d35b9ca2e877a6fc55dde6ad7c9ccfc0b35624d9/R/sparseMatrix.R#L426

**To Reproduce**

```r
> library("tiledbsoma")
> human_experiment = SOMAExperimentOpen("./apis/python/notebooks/data/sparse/")
> x_sparse <- human_experiment$ms$get("RNA")$X$get("data")
> x_sparse$read_sparse_matrix_zero_based(coords = list(1:10, 1:10))
Error in Matrix::sparseMatrix(i = 1 + as.numeric(tbl$GetColumnByName("soma_dim_0")),  : 
  invalid 'dims'
In addition: Warning message:
In as.integer.integer64(self$shape()) : NAs produced by integer overflow
```

mlin:
@pablo-gar @eddelbuettel would a reasonable solution here be to use `dims = as.integer(min(self$shape(), .Machine$integer.max))`?

eddelbuettel:
[...]

PR #1367, opened 2023-05-11
[r] [RFC] Add `SOMASparseNDArray$read_spam_matrix()`

mojaveazure:
Add support for reading spam matrices from sparse arrays. This allows users to read in data exceding the capacity of Matrix-based sparse matrices (`nnz > .Machine$integer.max || dim(mat) > .Machine$integer.max`)

Implemented SOMA methods:

- `SOMASparseNDArray$read_spam_matrix()`: allows reading in an array as a spam matrix (both 32- and 64-bit)

Future work:

- determine if this is desired functionality
- we may want to combine all sparse-matrix readers into one `read_sparse_matrix()` or `read_matrix()` method with options to determine how to realize the matrix

Note for reviewers:

I don't know if this is functionality we want to support. I'm more than happy to say that we don't want to support spam, in which case we can close and move on

partially addresses #1348

### DIFF
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -63,7 +63,9 @@ Suggests:
     testthat (>= 3.0.0),
     pbmc3k.tiledb,
     SeuratObject (>= 4.1.0),
-    datasets
+    datasets,
+    spam,
+    spam64
 VignetteBuilder: knitr
 Config/testthat/edition: 2
 OS_type: unix
--- a/apis/r/R/SOMASparseNDArray.R
+++ b/apis/r/R/SOMASparseNDArray.R
@@ -214,6 +214,81 @@ SOMASparseNDArray <- R6::R6Class(
       }
     },
 
+    #' @description Read as a spam-based sparse matrix (lifecycle: experimental)
+    #' @param coords ...
+    #' @template param-result-order
+    #' @param log_level ...
+    #'
+    read_spam_matrix = function(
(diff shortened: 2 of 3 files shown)


eddelbuettel (approved):
:+1:

I think that is a great and reasonably "free" option for us on capital-l "Large" data

_Edit: Didn't see the test breakage.  Odd.  Any pointers on this @mojaveazure ?

eddelbuettel:
Looks like we get other repurcussions from being a spam object and not inheriting from Matrix?

mojaveazure:
Yeah; that and `t,Matrix()` is an S4 method using an implicit S4 generic defined by Matrix, not an S3 method on `base::t()`

```R
> base::t
function (x) 
UseMethod("t")
<bytecode: 0x560e860bc968>
<environment: namespace:base>
> Matrix::t
standardGeneric for "t" defined from package "base"

function (x) 
standardGeneric("t")
<environment: 0x560e860e89c8>
Methods may be defined for arguments: x
Use  showMethods(t)  for currently available ones.
```
S4 generics/methods get weird with implicit declaration and overriding S3 generics

pablo-gar:
look great! 

Can we hold on on merging this PR? I introduced some design decisions on #1274. For example the addition of `SOMASparseNDarray$convert_coords()` and `utils-readTransformers.R` where we should put the arrow-to-spam conversion. And we would also likely want have an iterated version of `read_spam_matrix`

PR #1370, opened 2023-05-12
[python] Add run-time type checks to `PlatformConfig` on ingest/create

johnkerl:
**Issue and/or context:**

Resolves a UX issue encountered yesterday.

Note that Python doesn't do run-time type-checking.

We do have types enabled, and we do have CI for `ruff` and `mypy`, and we do have `typeguard` enabled in our CI which _does_ enforce run-time type checking at CI time -- for all code within this package.

However, `typeguard` is not used at runtime outside of CI, so we must add manual checks at the user interface where arguments are coming from out-of-package. This PR is one in a series of several PRs wherein we manually add a type-check to a crucial argument.

### DIFF
--- a/apis/python/src/tiledbsoma/_util.py
+++ b/apis/python/src/tiledbsoma/_util.py
@@ -14,6 +14,7 @@ import somacore
 from somacore import options
 
 from ._types import OpenTimestamp, Slice, is_slice_of
+from .options._tiledb_create_options import TileDBCreateOptions
 
 
 def get_start_stamp() -> float:
@@ -260,3 +261,12 @@ def ms_to_datetime(millis: int) -> datetime.datetime:
     secs, millis = divmod(millis, 1000)
     dt = datetime.datetime.fromtimestamp(secs, tz=datetime.timezone.utc)
     return dt.replace(microsecond=millis * 1000)
+
+
+def validate_platform_config(platform_config: Any) -> None:
+    """Offers a run-time type check on this ingestion/creation argument, which is easy to miskey."""
+    if platform_config is not None:
+        if isinstance(platform_config, TileDBCreateOptions):
+            raise TypeError(
+                """platform_config argument is of type TileDBCreateOptions -- please wrap it in `{"tiledb": {"create": argument}}`."""
+            )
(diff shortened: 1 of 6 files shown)


apis/python/src/tiledbsoma/_util.py
```diff
@@ -260,3 +261,12 @@ def ms_to_datetime(millis: int) -> datetime.datetime:
     secs, millis = divmod(millis, 1000)
     dt = datetime.datetime.fromtimestamp(secs, tz=datetime.timezone.utc)
     return dt.replace(microsecond=millis * 1000)
+
+
+def validate_platform_config(platform_config: Any) -> None:
+    """Offers a run-time type check on this ingestion/creation argument, which is easy to miskey."""
+    if platform_config is not None:
+        if isinstance(platform_config, TileDBCreateOptions):
```
thetorpedodog, line 269: At runtime, `TileDBCreateOptions` is just `dict`. You probably want to check for the presence of certain `TileDBCreateOptions` keys at the root level.

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -186,6 +196,8 @@ def from_anndata(
               writing any data to the array. Useful to prepare for appending
               multiple H5AD files to a single SOMA.
 
+        XXX use_relative_uri
```
thetorpedodog, line 199: XXX here

atolopko-czi (approved):
LGTM

apis/python/src/tiledbsoma/_util.py
johnkerl: @thetorpedodog `TileDBCreateOptions` is a `class`:
https://github.com/single-cell-data/TileDB-SOMA/blob/1.2.3/apis/python/src/tiledbsoma/options/_tiledb_create_options.py#L44

`PlatformConfig` is a `dict`:
https://github.com/single-cell-data/SOMA/blob/python-v1.0.0/python-spec/src/somacore/options.py#L93

thetorpedodog: Whoops, I was confusing it with one of the other types we have. I don’t entirely follow what the intent is here, because a user shouldn’t be creating a CreateOptions themselves in the normal course of business (it’s not on any publicly exposed APIs).

PR #1371, opened 2023-05-12
[python] Update `platform_config` usage advice

johnkerl:
**Issue and/or context:** Follow-on to #1370, with improved usage advice -- tracking issue #1394

**Changes:** Use the intended syntax.

**Notes for Reviewer:** Where is a good reference link we can point users to that has a list of available keys? The only list I'm aware of is in [TileDBCreateOptions](https://github.com/single-cell-data/TileDB-SOMA/blob/1.2.3/apis/python/src/tiledbsoma/options/_tiledb_create_options.py) method names, but this is not exported API.

### DIFF
--- a/apis/python/src/tiledbsoma/_collection.py
+++ b/apis/python/src/tiledbsoma/_collection.py
@@ -96,9 +96,8 @@ class CollectionBase(
             uri:
                 The location to create this SOMA collection at.
             platform_config:
-                Platform-specific options used to create this array,
-                provided via ``{"tiledb": {"create": ...}}`` nested keys,
-                where the ``...`` is of type ``tiledbsoma.options._tiledb_create_options.TileDBCreateOptions``.
+                Platform-specific options used to create this collection, provided in the form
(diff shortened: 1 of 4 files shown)


thetorpedodog:
There is documentation in this README file. It’s not perfect but it is a start: https://github.com/single-cell-data/TileDB-SOMA/tree/main/apis/python#platform_config-format

atolopko-czi:
> **Notes for Reviewer:** Where is a good reference link we can point users to that has a list of available keys? The only list I'm aware of is in [TileDBCreateOptions](https://github.com/single-cell-data/TileDB-SOMA/blob/1.2.3/apis/python/src/tiledbsoma/options/_tiledb_create_options.py) method names, but this is not exported API.

Maybe document in the [README](https://github.com/single-cell-data/TileDB-SOMA/blob/1.2.3/apis/python/README.md#platform_config-format)?

atolopko-czi (approved):
LGTM, but I might consider using a generic/dummy name for the innermost key like, `some_create_option`

PR #1372, opened 2023-05-12
[r] Provide reference to `libtiledbsoma` source via soft-link

eddelbuettel:
**Issue and/or context:**

The R package in `apis/r` needs the `libtiledbsoma` library.  Current package practice, when a build is required as no system installation is found is to bring the source in 'externally'.   Using a soft-link back to the library source was considered at some point but discarded as it seemed unclear how `git`, `R` (esp when build the 'release' tarball) and different file systems would react.

It turns out that those fears were unfounded. Using a soft-link (even to a directory) works just fine in recent `git`, and `R CMD build` transparently follows the link and includes the sources as a backup.  This also works at r-universe per initial tests.

This is an improved approach as it ties the _usage_ of the library closer to the _provision_ and minimises risk of slippage.

**Changes:**

As before, when a system library is seen no build is needed.  If a build is needed, it is made with the source which are either included in the tar.gz or accessible when the repo is checked (as in the case of CI at GH Actions).  The download step is skipped which allowed for a minor refactoring of the `configure` script which is again the only place from which
`src/Makevars` is constructed.   An extra wrinkle of testing for an 'out of $PATH' `cmake` was added as needed at CRAN in case we ever upload (and/or this setup serves as model for other uses).

**Notes for Reviewer:**

[SC 28874](https://app.shortcut.com/tiledb-inc/story/28874/examine-viability-of-libtiledbsoma-soft-link-into-apis-r-src-to-avoid-source-download)


apis/r/tools/get_tarball.R
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env Rscript
+
+## version pinning info
+tiledb_core_version <- "2.15.2"
```
johnkerl, line 4: to minimize keystroking on future dependency-upgrade PRs, @gspowley wrote a very nice `scripts/update-tiledb-version.py`:

it looks like that script might not update this current script as-is -- ?

updating that script to handle this file might be done on this PR, or on a follow-up PR

apis/r/src/libtiledbsoma
```diff
@@ -0,0 +1 @@
+../../../libtiledbsoma/
```
johnkerl, line 1: ✨

apis/r/configure
```diff
@@ -26,11 +26,30 @@ if [ $? -eq 0 ]; then
             src/Makevars.in > src/Makevars
 
         echo "** updated src/Makevars for system library via pkg-config"
+
+        ## we can exit now as we have a working setup, re
```
johnkerl, line 30: missing end of sentence?

johnkerl (commented):
LGTM -- very elegant solution!

I left one comment about our dependency-updater script

Let me try this locally on my Mac laptop (outside of CI) before approving

Thanks again!

apis/r/tools/get_tarball.R
eddelbuettel: I wondered about that too but I thought the above was better because we only to change _once_ for version and sha1 rather than three times (with largely redundant long URLs):

https://github.com/single-cell-data/TileDB-SOMA/blob/326d5eac52ba6fd65ea2fda62faa8e7cf5a6c1a7/apis/r/tools/get_tarballs.R#L8-L19

Easy enough to revert if that is what we prefer.  In tiledb-r I actually update two-line text 'key: value' text file that is then read (via a [[DCF]()](https://www.debian.org/doc/debian-policy/ch-controlfields.html) reader R has in base so no need to worry about jaml or json...):

https://github.com/TileDB-Inc/TileDB-R/blob/870f9343bd4fab056e8e46ee83b23acfb47b423d/tools/tiledbVersion.txt#L1-L2

apis/r/configure
eddelbuettel: Thanks, fixed now.

eddelbuettel:
> Let me try this locally on my Mac laptop (outside of CI) before approving

Let me know tomorrow or Sunday how it went. My score is pretty bright green at this point:
- linux w/ system lib (locally)
- linux w/o system lib in container (locally)
- linux w/o system lib at CI
- linux w/o system lib at r-universe
- linux w/o system via via remotes::github_install() in container (locally)
- macos w/o system lib at CI
- macos w/o system lib at r-universe

johnkerl (approved):
Looks good on personal non-CI mac which is the last bit of ☘️ IMHO :D

PR #1377, opened 2023-05-15
[python] Address leading-underscore issue for `uns` data

johnkerl:
**Issue and/or context:**

Found during testing. Issue is that TileDB Cloud disallows leading underscores, but we sometimes have things like `uns/_scvi`. Failure:

```
error creating the group, please check your inputs and try again: name must begin with an unaccented alphanumeric character
```

**Changes:**

Prepend a leading `u`, e.g. `uns/u_scvi`.

**Notes for Reviewer:**

This is still in draft until I get some unit-test cases up that test the affected case. (Already verified ad-hoc on laptop.)

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -10,6 +10,7 @@ other formats. Currently only ``.h5ad`` (`AnnData <https://anndata.readthedocs.i
 """
 
 import math
+import re
 import time
 from typing import (
     Any,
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1223,7 +1231,9 @@ def _ingest_uns_dict(
                     context=context,
                     ingest_mode=ingest_mode,
                 ) as df:
-                    _maybe_set(coll, key, df, use_relative_uri=use_relative_uri)
+                    _maybe_set(
+                        coll, key, df, use_relative_uri=use_relative_uri
+                    )
```
thetorpedodog, line 1236: spurious format change?

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1194,6 +1195,13 @@ def _ingest_uns_dict(
         _maybe_set(parent, parent_key, coll, use_relative_uri=use_relative_uri)
         coll.metadata["soma_tiledbsoma_type"] = "uns"
         for key, value in dct.items():
+
+            # Things like uns/_scvi cannot be uploaded to TileDB Cloud since
+            # "name must begin with an unaccented alphanumeric character" --
+            # so turn that into "u_scvi".
+            if not re.match('^[a-zA-Z].*$', key):
```
thetorpedodog, line 1202:
```python
import string
# ...
# for bonus points maybe pull this concatenation to the module level
if key[0] not in string.ascii_letters + string.digits:
  key = "u" + key
```

Also: Should this only be done for group metadata entries? Like, would it make sense if `_neighbors` was instead just `"bob"` that it should be `_neighbors: bob` in the metadata?

thetorpedodog (commented):
unfortunate that we don’t really have a better way to deal with this than putting something in front of it, but that's unstructured data for you.

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: @thetorpedodog idk; no longer there after a fresh lint run

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: @thetorpedodog re the former: done

Re the latter: I don't know but I have no signal from "data in the wild" that we should do so.

thetorpedodog (approved):

johnkerl:
Names like `_scvi` are functional on TileDB Cloud.

PR #1382, opened 2023-05-16
[r] Remove two debug statements

johnkerl:
**Issue and/or context:**

Remnants from #1325

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/r/R/TileDBGroup.R
+++ b/apis/r/R/TileDBGroup.R
@@ -75,10 +75,8 @@ TileDBGroup <- R6::R6Class(
         for (member in private$.member_cache) {
           if (!is.null(member$object)) {
             if (member$object$is_open()) {
-              spdl::debug("Openchi {} '{}'", member$object$class(), member$object$uri)
               member$object$close()
             } else {
-              spdl::debug("Clsdchi {} '{}'", member$object$class(), member$object$uri)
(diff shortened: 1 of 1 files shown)


aaronwolen (approved):

apis/r/R/TileDBGroup.R
```diff
@@ -75,10 +75,8 @@ TileDBGroup <- R6::R6Class(
         for (member in private$.member_cache) {
           if (!is.null(member$object)) {
             if (member$object$is_open()) {
-              spdl::debug("Openchi {} '{}'", member$object$class(), member$object$uri)
               member$object$close()
             } else {
-              spdl::debug("Clsdchi {} '{}'", member$object$class(), member$object$uri)
```
eddelbuettel, line 81: _Technically_  you could also keep them, or downgrade them to `spdl::trace(...)` as they only trigger when the level is set...

johnkerl: @eddelbuettel agreed but (a) they're redundant; (b) I really really did not intend to keep them

eddelbuettel: And they do slow things down even if at marginal levels so removal is good too. It's all good :grinning:

Issue #1378, opened 2023-05-16
[python] Code-review follow-up


johnkerl:
I don't understand what behavior would lead to this case—this seems like it would violate one of the preconditions that I would expect. How do we arrive here?

_Originally posted by @thetorpedodog in https://github.com/single-cell-data/TileDB-SOMA/pull/1325#discussion_r1194163445_

Parent: #1346

johnkerl:
@thetorpedodog here is the logic in question:
https://github.com/single-cell-data/TileDB-SOMA/blob/2fe22f86827f8dd2bae4b9001946d59dd3766fb0/apis/r/R/TileDBGroup.R#L164-L175

Here is a repro script

```
library(tiledbsoma)

uri <- '/var/s/v/ileum'
experiment <- SOMAExperimentOpen(uri)

cat("EXP          MODE:", experiment$mode(), "\n")
cat("EXP/MS       MODE:", experiment$ms$mode(), "\n")

exp_ms_rna <- experiment$ms$get('RNA')
cat("EXP/MS/RNA   MODE:", exp_ms_rna$mode(), "\n")

exp_ms_rna <- experiment$ms$get('RNA', 'WRITE')
cat("EXP/MS/RNA+W MODE:", exp_ms_rna$mode(), "\n")

experiment$close()
```

With the logic in question commented out we get

```
EXP          MODE: READ
EXP/MS       MODE: READ
EXP/MS/RNA   MODE: READ
EXP/MS/RNA+W MODE: READ
```

With the logic in question not commented out, i.e. as-is in the snippet linked to above, we get

```
EXP          MODE: READ
EXP/MS       MODE: READ
EXP/MS/RNA   MODE: READ
EXP/MS/RNA+W MODE: WRITE
```

Thoughts @thetorpedodog @mlin ? Have I mis-thought the API somehow with regard to implicit opens?

thetorpedodog:
In the Python API, there is no way to open a sub-element in a different mode than the parent. This was how I had imagined the API contract in general; children are always in the same mode etc. as the parents. (Likewise, once you open a child once, getting it again will always return the same object.)

Is this present to support an existing use case?

johnkerl:
@thetorpedodog the use-case is here -- before and after #1325:

[...]

PR #1383, opened 2023-05-16
[r] Update mode semantics for implicit open

johnkerl:
Addresses #1378:

* Remove `mode` optarg on `collection$get`
* Have opened children inherit parent's mode


aaronwolen (approved):
Thanks!

PR #204, opened 2022-06-28
Make a simple default display for SOMA objects in notebooks

johnkerl:
## Before

```
>>> soma
<tiledbsc.soma.SOMA object at 0x7f98d0b33b80>
```

## After

```
>>> soma
Name:    pbmc-small
URI:     tiledb-data/pbmc-small
(n_obs, n_var): (80, 20)
X:       'data'
obs:     'orig.ident', 'nCount_RNA', 'nFeature_RNA', 'RNA_snn_res.0.8', 'letter.idents', 'groups', 'RNA_snn_res.1'
var:     'vst.mean', 'vst.variance', 'vst.variance.expected', 'vst.variance.standardized', 'vst.variable'
obsm:    'X_tsne', 'X_pca'
varm:    'PCs'
obsp:    'distances'
varp:
raw/X:   'data'
raw/var: 'vst.mean', 'vst.variance', 'vst.variance.expected', 'vst.variance.standardized', 'vst.variable'
uns:     neighbors
```

PR #1385, opened 2023-05-17
[python] MacOS arm64 wheel-builds for PyPI

johnkerl:
**Issue and/or context:** #204

**Changes and notes:**

* As noted in the code comments, using today's GitHub Actions the best we can do for MacOS arm64 is to _cross-compile_ for arm64 while actually running on x86_64.
  * This is tracked here: https://github.com/github/roadmap/issues/528
  * TL;DR maybe 2023Q4
* This also means we can _build_ the arm64 wheels in CI but we cannot _smoke-test them_ -- since we do not have arm64 hardware to run the smoke test on
  * Consolation prize:
    * Get the artifacts from the CI job
    * Download and unzip `wheels-macos-arm64.zip`
    * `pip install tiledbsoma-1.2.3.post52.dev48715262-cp311-cp311-macosx_11_0_arm64.whl` (or whatever your OS version)
    * In Python, `import tiledbsoma` and `tiledbsoma.show_package_versions()`
    * Run unit tests via `python -m pytest apis/python/tests`
* For dev work on this PR:
  * I temporarily commented out the on-failure ticket-cutter; I'll restore that.
  * Likewise, I had it build wheels on PR push; I'll restore that.
  * Do note re the latter that this _does_ succeed wheel-build CI: see https://github.com/single-cell-data/TileDB-SOMA/actions/runs/5017091359
* 1,000 thanks to @ihnorton for technical help


johnkerl:
@beroy @atolopko-czi @ihnorton ping 🙏

ebezzi (approved):

atolopko-czi (approved):
If it works, then it LGTM. 😄 Changes look sane.

PR #1388, opened 2023-05-17
[WIP] PyPi publish workflow

### LINKED ISSUE #1330: [bug/python] Autotwine to PyPI is not working correctly on releases
As seen, for most recent example, on [1.2.3](https://github.com/single-cell-data/TileDB-SOMA/releases/tag/1.2.3), it does not run to completion:

* https://github.com/single-cell-data/TileDB-SOMA/actions/runs/4876455117

Also note we have a nightly for pre-release check; this is having an internal fail but it isn't being flagged as a job fail, which is two things to debug and fix:

* https://github.com/single-cell-data/TileDB-SOMA/actions/runs/4870503867 is green
* https://github.com/single-cell-data/TileDB-SOMA/actions/runs/4870503867/jobs/8686743842 subcomponent, if you expand it, actually did not succeed

bkmartinjr:
WORK IN PROGRESS

Changes to python-packaging.yml:
* Update the PyPi publication workflows dependency to match released version of  `pypa/gh-action-pypi-publish`
* Fix whitespace formatting issue in YML

Also, TEMPORARY/WIP changes for debugging, which will be reverted:
* disable PyPi and create-issue-on-error steps
* disable scheduled 
* run on this branch, not on `main`

Fixes #1330

### DIFF
--- a/.github/workflows/python-packaging.yml
+++ b/.github/workflows/python-packaging.yml
@@ -18,18 +18,18 @@ jobs:
     name: Build source distribution
     runs-on: ubuntu-20.04
     steps:
-    - name: Checkout TileDB-SOMA
-      uses: actions/checkout@v3
-      with:
-        fetch-depth: 0  # ensure we get all tags to inform package version determination
-    - name: Build sdist
-      run: python setup.py sdist
-      working-directory: ./apis/python
-    - name: Upload sdist artifact to GitHub Actions storage
-      uses: actions/upload-artifact@v3
-      with:
-        name: sdist
-        path: apis/python/dist/
+      - name: Checkout TileDB-SOMA
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # ensure we get all tags to inform package version determination
+      - name: Build sdist
+        run: python setup.py sdist
+        working-directory: ./apis/python
+      - name: Upload sdist artifact to GitHub Actions storage
+        uses: actions/upload-artifact@v3
+        with:
+          name: sdist
+          path: apis/python/dist/
 
   # This step builds wheels and uploads them to GitHub Actions storage.
   # See also https://github.com/single-cell-data/TileDB-SOMA/issues/700.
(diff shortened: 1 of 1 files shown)


.github/workflows/python-packaging.yml
```diff
@@ -1,35 +1,30 @@
 name: TileDB-SOMA python sdist & wheels
 
 on:
-  # I.e. on request
+  # trigger via user request, e.g.,
+  #     gh workflow run python-packaging.yml --ref branch
+  # Specify the branch or other ref as required.
```
thetorpedodog, line 6: Could you add a note at the top that `workflow_dispatch` pushes to test pypi rather than real pypi?

thetorpedodog (approved):
Looks good overall; you might want to have mlin look it over just to make sure I haven't missed anything.

.github/workflows/python-packaging.yml
bkmartinjr: will do

bkmartinjr:
> Looks good overall; you might want to have mlin look it over just to make sure I haven't missed anything.

I plan to wait for review by _all_ reviewers on this PR!

.github/workflows/python-packaging.yml
```diff
@@ -1,35 +1,37 @@
...
-  # pull_request:
-  #   branches: [main]
-  # paths-ignore:
-  #   - 'apis/r/**'
+
+  # Trigger publication to PyPi via new release event.
   release:
     types: [published]
-  schedule:
-    - cron: "42 9 * * *"
```
mlin, line 14: Thanks for getting to this, looks great, my only question is whether to keep the nightly build? It was a way of warning us about package-breaking changes that inadvertently make it onto main (since the packaging workflow doesn't otherwise trigger automatically + routinely). Which has happened a number of times.

bkmartinjr: There is a discussion in slack about this point.

TL;DR:
* PyPi/TestPyPi won't accept the same version number twice
* to do a nightly publication, we would need to only do it when there has been a commit, i.e., a new version number will be generated.
* So as originally written, the workflow would fail on any day when there was not a commit to main
* I volunteered to add this gate (ie. skip if no commit in the past 24 hours) if it was desired, and as an interim, removed the scheduled run.

So LMK if this is worth adding.  I'll probably do it as a separate PR, as it makes the TestPyPi action logic more complex.

thetorpedodog: Since we don't care about version ordering on testpypi, we could do something like this:

1. Generate a fake version number using the commit hash (e.g. 0.0.[convert 32 bits of the commit hash to an integer])
2. See if that release already exists in testpypi.
3. If not, make a release to testpypi.

bkmartinjr: That would require a different build process, right?  The version is encoded in the whl/sdist, which is shared by both workflows.  Are you aware of a means of revising the version during the twine/upload process?

thetorpedodog: If it's a real workflow, use the normal "real version" logic; if it's a fake workflow (nightly / manual), override the entire thing with a fake version.

bkmartinjr: I"m not following - I'll need to touch base offline to get further details, if we want to proceed down this path.  Let me know if that is the case and I"ll hit you in another venue.

johnkerl (approved):
🚢 

Thank you!!!

.github/workflows/python-packaging.yml
mlin: If auto-pushing to TestPyPI causes complications, I think it'd be fine to *not* push to it at all for the nightly build. I believe this would be a condition something like `if: ${{ github.event_name != 'schedule' }}`

There is still some value in frequently exercising the preceding parts of this workflow that build the wheels and run basic checks on them. (Yet it's not suitable to do so for every push to main, since it uses a lot of CI resources to fill in the platform/pyversion matrix)

thetorpedodog: to be clear, my suggestions here were not something to do on this change, but thoughts for the future if we ever want to go in that direction.

bkmartinjr: OK, I am now understanding!

I'll make an amendment to the PR along the lines Mike has suggested and re-tag you for review.  I.e., execute the workflow on a cron, _excluding_ TestPyPi/PyPi push.

Thanks!

bkmartinjr: OK, PTAL as of `9bee92c`:
- added the scheduled run
- TestPyPi is conditional upon user request event
- PyPi is conditional on release event
- modified the error handler (bug file) to be dependent on all branches so it will catch failure in any invocation, including the scheduled event

bkmartinjr: I also revalidated the TestPyPi functionality.  [The run is here](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/5015669731).

mlin (approved):

thetorpedodog (approved):
Still approved from last time.

PR #1395, opened 2023-05-19
[python] Improve creation-URI construction

johnkerl:
**Issue and/or context:** Improves the construction of creation URIs for resume-mode.


johnkerl:
Dup of #1396

PR #1397, opened 2023-05-22
[c++] Build and upload library artifacts for Mac (arm, x86) and Linux

eddelbuettel:
**NOTE: Do not merge yet.** The PR has been updated and now works off a _release_ but requires a PAT to upload. We may want to take care of the PAT first.

See also #1454 which depends on this.

**Issue and/or context:**

Deployment benefits from having pre-made artifacts which lowers the installation 'cost' for users and smoothes deployments.  Similar to how TileDB Core has binaries for the key architecture and OS combinations, SOMA would benefit from having this.

**Changes:**

GitHub Actions are set up to create and upload artifacts for macOS (both old x86_64 and new arm64) and linux.  

**Notes for Reviewer:**

This should still be considered draft as it will likely benefit from refinements such as tagging to actual release version, switching the linux container to a manylinux one for more portable use and so on. 

[SC 29249](https://app.shortcut.com/tiledb-inc/story/29249/c-create-libtiledbsoma-build-artifacts) / #1401

### DIFF
--- a/.github/workflows/library-artifacts.yaml
+++ b/.github/workflows/library-artifacts.yaml
@@ -7,10 +7,11 @@ on:
   #    - ".pre-commit-config.yaml"
   #push:
   #  branches: [main, master]
+  workflow_dispatch:
   release:
     types:
       - created
-
+      
 env:
   ## for workflows triggered by release this is the release tag created (cf https://docs.github.com/en/actions/learn-github-actions/contexts)
   ## note that this is filtered below and then used via 'steps.versubstr.outputs.substring'
--- a/.github/workflows/library-artifacts.yaml
+++ b/.github/workflows/library-artifacts.yaml
@@ -11,7 +11,7 @@ on:
   release:
     types:
       - created
-      
+
 env:
   ## for workflows triggered by release this is the release tag created (cf https://docs.github.com/en/actions/learn-github-actions/contexts)
   ## note that this is filtered below and then used via 'steps.versubstr.outputs.substring'
@@ -44,7 +44,7 @@ jobs:
             ARTIFACT_OS: 'macos'
             ARTIFACT_ARCH: 'arm64'
           - environ: "linux"
-            imageName: "ubuntu-latest"
+            imageName: "ubuntu-20.04"
             CMAKE_OSX_ARCHITECTURES: ""
             MACOSX_DEPLOYMENT_TARGET: ""
             ARTIFACT_OS: 'linux'
--- a/.github/workflows/library-artifacts.yaml
+++ b/.github/workflows/library-artifacts.yaml
@@ -1,0 +1,104 @@
+name: artifacts
+
+on:
+  pull_request:
+    paths-ignore:
+      - "apis/**"
+      - ".pre-commit-config.yaml"
+  push:
+    branches: [main, master]
+  release:
+
+env:
+  TILEDBSOMA_SHA: "dev"
+  TILEDBSOMA_BUILD: "${{ github.workspace }}/tiledb_build"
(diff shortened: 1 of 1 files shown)


eddelbuettel:
Good point, @johnkerl -- so I just poked @Shelnutt2 for review too.

I'll try to work on 'using these' artifacts side for the R package later today.

.github/workflows/library-artifacts.yaml
```diff
@@ -0,0 +1,105 @@
...
+  pull_request:
+    paths-ignore:
+      - "apis/**"
+      - ".pre-commit-config.yaml"
+  push:
+    branches: [main, master]
+  release:
+
+env:
+  TILEDBSOMA_VER: "1.2.4"
```
johnkerl, line 13: @eddelbuettel -- an idea from @Shelnutt2 (conversationally) -- we can get these from github variables
https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
`GITHUB_REF` and `GITHUB_SHA`

johnkerl (changes requested):

.github/workflows/library-artifacts.yaml
eddelbuettel: Oh yes sure -- he and I chatted today too. But this all can rest til 1.0R is shipped and then we clean
- where/how to get release and sha from
- how tickle (releases only ?)
- properly sequence (as the library needs to be a pre-requirement for the R package)

I think it can rest til then.

(And note that when I built / tested this I had no release tags to work off so it's all "in flux" in that way too.)

eddelbuettel: So "your requesting change" is not all that meaningful.  Or feel free to self-assign.  I did what I could here.

The remainder has external dependencies exogeneous to this PR.

eddelbuettel: That has been done ... but note that the Action now _depends_ on two things

- triggered by a release (i.e. when a release is made and tagged, the action runs and provides the artifacts)
- this _requires_ a PAT for access, we may have to add one / align the name we refer to

johnkerl: @eddelbuettel re the first: I think we can add this

```
on:
  #pull_request:
  #  paths-ignore:
  #    - "apis/**"
  #    - ".pre-commit-config.yaml"
  #push:
  #  branches: [main, master]
  release:
    types:
      - created
  workflow_dispatch: # add this
```

Then it will be manually launchable in the GitHub web UI

johnkerl: and re the PAT -- IIUC we can re-use the same one from other actions ...

eddelbuettel: Good one re 'workflow_dispatch' -- I had not used that, but it makes sense.  The main one here is to turn _off_ all the usual action on pushes, merges, ... 

Also re the PAT I had not even check but I wired one up in my test repo so I would only need to carry its identifier over into this.

eddelbuettel: @johnkerl Done for the workflow_dispatch.  I see no existing PAT so we'd have to get one in.

eddelbuettel:
We can close this / replace this it with something else in due course.  I have a working alternate.  We should probably keep this one open til we have the issue addressed, be it via this PR or another one.

johnkerl:
> I have a working alternate

@eddelbuettel hyperlink please? Thanks!

eddelbuettel:
That was a comment made six weeks ago.

johnkerl:
@eddelbuettel I know, and am doing bookkeeping, and tying up some loose ends I missed before.

If there is/was a working alternate, please let me know what/whre it is.

Thanks!

Issue #1394, opened 2023-05-19
[r] Let `PlatformConfig` be `TileDBCreateOptions` at top level


johnkerl:
User feedback:

* The `create` methods take an optional `PlatformConfig`
  * This should be of the form `{"tiledb": {"create": {key: value, ...}}}`
* Issues:
  * We lacked type-checking so if the argument supplied by the user was `{key: value, ...}` or a `TileDBCreateOptions` then it was silently ignored -- at minimum we need a run-time type-check. Fixed by #1370 and #1371.

Solution:

* We should allow `TileDBCreateOptions` at the base level.
  * Bonus: we get type-checking on the kwarg names that we don't get with bare dicts
  * Bonus: we get real-time menu/discovery of available create options in our on-line help via simply doc-linking to `TileDBCreateOptions`
 
Tasking:

* [x] #1370 done
* [x] #1371 done
* [x]  https://github.com/single-cell-data/SOMA/pull/165 done
* [x]  https://github.com/single-cell-data/SOMA/pull/166 done
* [ ] Expose `TileDBCreateOptions` at top level in this repo
  * [X] In Python #1394
  * [ ] In R
* [ ] Let the `create` methods in this repo take a union of `PlatformConfig` and `TileDBCreateOptions`
  * [X] In Python #1408 and #1430
  * [ ] In R
* [ ] Update the online docs to link to `TileDBCreateOptions`
  * [X] In Python #1394
  * [ ] In R

johnkerl:
@mojaveazure please work with @thetorpedodog on implementing the R side of this.

jp-dark:
Closing issue as stale.

PR #1408, opened 2023-05-23
Make TileDBCreateOptions suitable for human consumption.

thetorpedodog:
This updates the TileDBCreateOptions type to have attributes for each of the settings it provides.  This means that a user can construct one using normal-looking Python syntax, without having to build their own JSON-like structure to be re-parsed.  TileDBCreateOptions is now exported from tiledbsoma.options.

It also enables passing a TileDBCreateOptions directly as the platform_config value in creation operations, implementing the pattern that was recently legalized by the spec change.

---

Python half of #1394.


apis/python/src/tiledbsoma/_common_nd_array.py
```diff
@@ -88,7 +88,6 @@ def create(
         # the upper limit is exactly 2**63-1.
 
         context = _validate_soma_tiledb_context(context)
-        _util.validate_platform_config(platform_config)
```
johnkerl, line 91: I'm confused

What I want is to have -- and to verify by unit-test cases -- that:

* If they call `create` with `platform_config={"tiledb": {"create": {key: value, ...}}}` it works
* If they call `create` with `platform_config = TileDBCreateOptions(key=value)` it works
* If they call `create` with _anything else_ we raise an exception right there at runtime

Maybe I'm missing something but I don't see how the third bullet's functionality is isn't lost by this PR

johnkerl (changes requested):

apis/python/src/tiledbsoma/options/_tiledb_create_options.py
```diff
@@ -1,170 +1,290 @@
...
+        return _filters_from(self.attrs, name, default)
+
+
+_T = TypeVar("_T")
+
+
+def _dig_platform_config(
+    input: object, typ: Type[_T], full_path: Tuple[str, ...]
+) -> Union[Dict[str, object], _T]:
+    """Looks for an object of the given type along the path of a"""
```
johnkerl, line 212: Incomplete sentence?

thetorpedodog: Oops!

johnkerl (approved):

apis/python/src/tiledbsoma/options/_tiledb_create_options.py
```diff
@@ -1,170 +1,301 @@
...
+        if isinstance(current, typ):
+            return current
+        if not isinstance(current, dict):
+            return {}
+        key, path = path[0], path[1:]
+        try:
+            current = current[key]
+        except KeyError:
+            return {}
+    if not isinstance(current, (typ, dict)):
```
atolopko-czi, line 236: Will this `isinstance(current, typ)` ever be true here? If not, can just check for the `dict` at this point.

apis/python/src/tiledbsoma/options/_tiledb_create_options.py
```diff
@@ -1,170 +1,301 @@
...
+def _dig_platform_config(
+    input: object, typ: Type[_T], full_path: Tuple[str, ...]
+) -> Union[Dict[str, object], _T]:
+    """Looks for an object of the given type in dictionaries.
 
-    def __iter__(self) -> Iterator[str]:
-        return iter(self._config)
+    This is used to extract a valid object out of `platform_config`. If an
+    object matching `typ` is found, it is returned directly; otherwise we
+    descend in the next dictionary key specified by `full_path`.
```
atolopko-czi, line 222: Maybe add to the docstring that a `dict` is also allowed as the leaf element of the path.

apis/python/tests/test_platform_config.py
```diff
@@ -76,34 +75,78 @@ def test_platform_config(adata):
...
             "tiledb": {
-                "create": TileDBCreateOptions({"dims": {"soma_dim_0": {"tile": 6}}})
+                "create": tco.TileDBCreateOptions(dims={"soma_dim_0": {"tile": 6}})
             }
         }
     )
     assert tdb_create_options.dim_tile("soma_dim_0") == 6
 
 
+def test__from_platform_config__admits_create_options():
```
atolopko-czi, line 121:
```suggestion
def test__from_platform_config__admits_create_options_at_top_level():
```

apis/python/tests/test_platform_config.py
```diff
@@ -76,34 +75,78 @@ def test_platform_config(adata):
...
-def test__from_platform_config__admits_tiledb_create_options_object():
-    tdb_create_options = TileDBCreateOptions.from_platform_config(
+def test__from_platform_config__admits_create_options_in_shallow_dict():
+    tdb_create_options = tco.TileDBCreateOptions.from_platform_config(
+        {"tiledb": tco.TileDBCreateOptions(dims={"soma_dim_0": {"tile": 6}})}
+    )
+    assert tdb_create_options.dim_tile("soma_dim_0") == 6
+
+
+def test__from_platform_config__admits_create_options_in_dict():
```
atolopko-czi, line 110:
```suggestion
def test__from_platform_config__admits_create_options_in_fully_specified_dict():
```
not sure of my wording choice, but suggest something similar to help reader

atolopko-czi (approved):
A couple minor comments, but LGTM as is.

Look forward to using the simpler config design!

apis/python/tests/test_platform_config.py
thetorpedodog: updated these and other names.

Issue #1403, opened 2023-05-22
[r] make `matrixZeroBasedView` an R6 class and encapsulate all of its functionalities as R6 methods


pablo-gar:
As discussed in our spec meeting we would like to isolate access patterns of the `matrixZeroBasedView` into low-level methods. 

This means to stop support for `[]` and all other generics (e.g. arithmetic operations). It was proposed to follow the same design that Arrow does:

1. Create R6 class
2. Keep all 0-based operations on the "low-level" R6 methods
3. (In the future) If desired, create generics that support more seamlessly R-like usage e.g. `[]`.

pablo-gar:
Done #1411

PR #1411, opened 2023-05-24
[r] Refactor `matrixZeroBasedView`

pablo-gar:
addresses #1403


mlin:
@pablo-gar implementation as R6 class looks good here. I wonder will switching to R6 will make it harder if we do come back to adding the `[` syntax sugar in the future? The `matrixZeroBasedView` originally was in S4 following the examples I could find for overriding `[`. I'm OK with switching to R6 overall, just thought this point was worth checking.

apis/r/R/utils-matrixZeroBasedView.R
```diff
@@ -1,109 +1,94 @@
...
+  
+    #' @description Initialize (lifecycle: experimental)
+    #' @param x \link{matrix} or Matrix::\link[Matrix]{sparseMatrix} or Matrix::\link[Matrix]{denseMatrix}
+    initialize = function(x) {
+      if (!inherits(x, "matrix") && !inherits(x, "sparseMatrix") && !inherits(one_based_matrix, "denseMatrix")) {
+        stop("Matrix object must inherit class matrix or Matrix::sparseMatrix or Matrix:denseMatrix")
+      }
+      if (length(dim(x)) != 2) {
+        stop("Only two-dimensional matrices are supported")
+      }
```
eddelbuettel, line 20: Just as an FYI we have in many other places switched to using the more compact `stopifnot` where the named argument _is_ the error message.  So this would be (untested, and I sometimes get the boolean inversion wrong on the first try)

```r
stopifnot("Matrix object must inherit class matrix or Matrix::sparseMatrix or Matrix:denseMatrix" =
      inherits(x, "matrix") || inherits(x, "sparseMatrix") || inherits(one_based_matrix, "denseMatrix",
      "Only two-dimensional matrices are supported" = length(dim(x)) == 2)
```

apis/r/R/utils-matrixZeroBasedView.R
```diff
@@ -1,109 +1,94 @@
...
+      private$one_based_matrix
+    },
+    
+    #' @description Perform arithmetic sum between this link{matrixZeroBasedView}
+    #' and another link{matrixZeroBasedView}.
+    #' @param x the link{matrixZeroBasedView} to sum.
+    #' @return The result of the sum as a \link{matrixZeroBasedView}.
+    sum = function(x) {
+      if (!inherits(x, "matrixZeroBasedView")) {
+        "Only arithmetic sum with another 'matrixZeroBasedView` is supported"
```
eddelbuettel, line 75: Looks like this misses the `stop()` ?

eddelbuettel (approved):
Looks good to me with a narrower NAMESPACE and possible cleaner usage idioms in the tests.  Overall I have no strong feelings and think I could live with it either way -- as discussed this is a hard problem.

I will approve now but you should probably wait to hear from @mlin before merging.

A few minor stylistic suggestions sprinkled in.  Feel free to adapt or ignore, nothing earth-shattering here.

PR #1414, opened 2023-05-24
Restore `SOMASparseNDArray$read_sparse_matrix()`

mojaveazure:
Restore reading of one-based sparse matrices. Zero-based sparse matrices cause issues with the downstream ecosystems (eg. #1279). This PR brings back support for reading one-based matrices while retaining ability to read zero-based sparse matrices. Code is shared between both readers to reduce duplication.

Also makes uses of `Matrix::sparseMatrix(index1 = FALSE)` to reduce the amount of coercion from zero-based in SOMA &rarr; 1-based in call to `Matrix::sparseMatrix()` &rarr; 0-based internally within Matrix

### DIFF
--- a/apis/r/R/SOMASparseNDArray.R
+++ b/apis/r/R/SOMASparseNDArray.R
@@ -188,7 +188,6 @@ SOMASparseNDArray <- R6::R6Class(
       log_level = "warn"
     ) {
       private$check_open_for_read()
-      repr <- repr[1L]
       repr <- match.arg(repr)
       dims <- self$dimensions()
       attr <- self$attributes()
--- a/apis/r/R/SOMASparseNDArray.R
+++ b/apis/r/R/SOMASparseNDArray.R
@@ -174,7 +174,59 @@ SOMASparseNDArray <- R6::R6Class(
     #' @param coords Optional `list` of integer vectors, one for each dimension, with a
     #' length equal to the number of values to read. If `NULL`, all values are
     #' read. List elements can be named when specifying a subset of dimensions.
-    #' @template param-result-order
+    #' @param repr Optional one-character code for sparse matrix representation type
+    #' @param iterated Option boolean indicated whether data is read in call (when
+    #' `FALSE`, the default value) or in several iterated steps.
+    #' @param log_level Optional logging level with default value of `"warn"`.
+    #' @return A sparse matrix whose exact representation
+    #' is determined by \code{repr}.
+    read_sparse_matrix = function(
+      coords = NULL,
+      result_order = "auto",
+      repr = c("C", "T", "R"),
+      iterated = FALSE,
+      log_level = "warn"
+    ) {
+      private$check_open_for_read()
+      repr <- repr[1L]
+      repr <- match.arg(repr)
+      dims <- self$dimensions()
+      attr <- self$attributes()
+      stopifnot("Array must have two dimensions" = length(dims) == 2,
+                "Array must contain columns 'soma_dim_0' and 'soma_dim_1'" =
+                  all.equal(c("soma_dim_0", "soma_dim_1"), names(dims)),
+                "Array must contain column 'soma_data'" = all.equal("soma_data", names(attr)))
+
+      if (isFALSE(iterated)) {
+        tbl <- self$read_arrow_table(coords = coords, result_order = result_order, log_level = log_level)
+        mat <- Matrix::sparseMatrix(
+          i = as.numeric(tbl$GetColumnByName("soma_dim_0")),
+          j = as.numeric(tbl$GetColumnByName("soma_dim_1")),
+          x = as.numeric(tbl$GetColumnByName("soma_data")),
+          index1 = FALSE,
+          dims = as.integer(self$shape()),
+          repr = repr
+        )
+        return(mat)
+      } else {
+        ## should we error if this isn't null?
+        if (!is.null(private$soma_reader_pointer)) {
+          warning("pointer not null, skipping")
+        } else {
+          private$soma_reader_setup()
+          private$sparse_repr <- repr
+          if (rlang::is_na(private$read_next)) {
+            private$zero_based <- FALSE
+          }
+        }
+        invisible(NULL)
+      }
+    },
+
+    #' @description Read as a zero-indexed sparse matrix (lifecycle: experimental)
+    #' @param coords Optional `list` of integer vectors, one for each dimension, with a
+    #' length equal to the number of values to read. If `NULL`, all values are
+    #' read. List elements can be named when specifying a subset of dimensions.
     #' @param repr Optional one-character code for sparse matrix representation type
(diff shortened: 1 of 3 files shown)


apis/r/R/SOMASparseNDArray.R
```diff
@@ -174,57 +174,108 @@ SOMASparseNDArray <- R6::R6Class(
...
+    read_sparse_matrix = function(
       coords = NULL,
       result_order = "auto",
       repr = c("C", "T", "R"),
       iterated = FALSE,
       log_level = "warn"
     ) {
       private$check_open_for_read()
-
+      repr <- repr[1L]
```
eddelbuettel, line 191: I don't think you need that given the next line.  `match.arg()` will not let you pass a vector through.

eddelbuettel (commented):
I haven't looked in full detail but it looks generally ok.  One worry is whether this will bite with eg #1411 ?

apis/r/R/SOMASparseNDArray.R
mojaveazure: I've had `match.arg()` fail when passing default vectors, eg

```R
f <- function(repr = c("C", "T", "R")) {
  match.arg(repr)
}

g <- function(repr = c("C", "T", "R")) {
  f(x)
}

g()
```

but can't seem to reproduce it; I'll remove the line

eddelbuettel: g() is 'outside the spec' methinks.  match.arg() does some black art _with the signature of its enclosing function_.  You break that in the example, you get to keep the pieces.

At least that's how it reads to me.

eddelbuettel: Department of _Couldn't be quicker and dirtier if we tried_ calling on line 2:

```r
> f <- function(repr = c("C", "T", "R")) {
  match.arg(repr)
}
> f(c("T", "R", "R"))
Error in match.arg(repr) : 'arg' must be of length 1
> 
```

You had the right idea there but is already baked in.  And by default more explicit (--> actual error).

mojaveazure:
Good question; haven't seen #1411 so I'm tagging @pablo-gar to ensure he sees this. From a cursory glance, it seems like the only things changing are the use of `matrixZeroBasedView$new()` and no-longer needing the complicated zero-to-one based conversions in the ecosystem methods

apis/r/R/SOMASparseNDArray.R
```diff
@@ -188,7 +188,6 @@ SOMASparseNDArray <- R6::R6Class(
       log_level = "warn"
     ) {
       private$check_open_for_read()
-      repr <- repr[1L]
```
eddelbuettel: Thanks!

pablo-gar:
@mojaveazure What's the problem of reading a zero-based matrix and then converting to one-based? 

Like here:
https://github.com/single-cell-data/TileDB-SOMA/blob/99e50947a5d2c38c6a5c07eaa40ed4da64c8b8eb/apis/r/R/SOMAExperimentAxisQuery.R#L707

I think that gives you effectively the same result as `read_sparse_matrix()`

mojaveazure:
@pablo-gar I believe it does give the same result, but in a far less convoluted way. What `as.one.based(self$ms$obsp$get(obsp_layer)$read_sparse_matrix_zero_based(repr = 'C'))` is doing is
1. reading in coordinates as zero-based (SOMA)
2. converting to one-based for creation with `Matrix::sparseMatrix()`
3. converting to zero-based for internal usage within Matrix
4. wrapping the matrix in the zero-based view
5. extracting the matrix from the zero-based view

In perusing #1411, step 5 seems to get a lot more complicated https://github.com/single-cell-data/TileDB-SOMA/blob/f0b9d51fc47dba8f4da115536f6fb3c87b3a7276/apis/r/R/SOMAExperimentAxisQuery.R#L592-L596

This PR instead simplifies this process to
1. reading in coordinates as zero-based
2. creating with zero-based coordinates using `Matrix::sparseMatrix(index1 = FALSE)`
3. Letting Matrix handle 1-based for the ecosystems and preserving the `matrixZeroBasedView` wrapper

I've argued [before](https://github.com/single-cell-data/TileDB-SOMA/issues/1232#issuecomment-1523975809) (and [again](https://github.com/single-cell-data/TileDB-SOMA/issues/1232#issuecomment-1524672937)) that the vast majority of users won't need zero-based matrices, so providing easy access for one-based matrices makes life easier. Moreover, because the zero-based view is a wrapper around one-based matrices, this simply exposes the first part of that process

pablo-gar:
If this is a UX wrapper I'm not opposed nor in favor but I would strongly suggest we don't duplicate code and create risks for bugs and/or added dev work in case we ever modify either of sparse options. 

So I propose we either implement these function this way:

```r
read_sparse_matrix = function(...) {
   mat <- self$read_sparse_matrix_zero_based(...)
   mat$get_one_based_matrix()
}
```

OR the converse:

```r
read_sparse_matrix_zero_based = function(...) {
   mat <- self$read_sparse_matrix()
   matrixZeroBasedView$new(mat)
}
```

mojaveazure:
@pablo-gar This PR does just that: `read_sparse_matrix_zero_based()` wraps `read_sparse_matrix()` just as the `matrixZeroBasedView()` wraps sparse matrices https://github.com/single-cell-data/TileDB-SOMA/blob/06e0105d32cd5100fa64865705aefee235d4efcd/apis/r/R/SOMASparseNDArray.R#L238-L266

mojaveazure:
@pablo-gar Whoops, didn't see that your edit saying you already saw that, my bad

pablo-gar:
Np! I'm processing as I type! I should process and then type after.

The other part that I'm trying to grasp is how this will collide with #1274 but I think that should be an "easy fix" on that other PR

pablo-gar:
@mojaveazure also see that we may get rid altogether of the ability to do "full" reads

https://github.com/single-cell-data/TileDB-SOMA/pull/1274#discussion_r1204641834

mojaveazure:
@pablo-gar Hmm, interesting. I'm not sure how well that will work in the R-world as R doesn't have a useful version of iterator and both Seurat and SingleCellExperiment need the full matrices (at least for now, there are ways around that but it'll involve a lot more work)

pablo-gar:
@mojaveazure yup it'll be a long call like in Python

```r
full_mat <- soma_sparse$read()$sparse_matrix()$concat()
```

pablo-gar (approved):

mlin:
Belated- I'm also neutral on this, but having merged it, shouldn't we also update [SOMAExperimentAxisQuery.R](https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/r/R/SOMAExperimentAxisQuery.R) to use it instead of `read_sparse_matrix_zero_based()`?

mojaveazure:
@mlin Yes; that will be brought about in future PRs, namely #1279 for SingleCellExperiment and #1261 for Seurat

PR #1415, opened 2023-05-24
[r/docs] Fine-tune article display in pkgdown setup

### LINKED ISSUE #1376: [r] Apply logical ordering in the `pkgdown` article-list
At https://single-cell-data.github.io/TileDB-SOMA/:

* Reference entries should be alphabetical -- this is automatic
* Articles are currently listed sorted (non-case-folded) by HTML file name

The former is OK; the latter is a bit non-intuitive. It's putting "SOMA Objects Overview" 3rd in the list when it should be first.

I believe we can use a `_pkgdown.yml` [like this one](https://github.com/TileDB-Inc/TileDB-Cloud-R/blob/master/_pkgdown.yml) to specify the sort ordering.

cc @eddelbuettel @aaronwolen @mojaveazure

eddelbuettel:
Fixes #1376

**Issue and/or context:**

As discussed in #1376, the order of articles ("vignettes") presented is rule-based (by filename) and not ideal.

**Changes:**

We impose an explicit ordering and segmentation in the navigation bar:

image

We also update the basic rendering to bootstrap5 which uses crisper CSS all-around.  The logic in the GitHub Action should be unaffected the basic `build_site()` is embedded in the 'update and push to gh-pages' step.

**Notes for Reviewer:**

#1376
[SC 29399](https://app.shortcut.com/tiledb-inc/story/29399/fine-tune-pkdown-listing-of-vignettes_

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -1,2 +1,0 @@
-template:
-  bootstrap: 5
--- a/apis/r/_pkgdown.yml
+++ b/apis/r/_pkgdown.yml
@@ -1,0 +1,25 @@
+template:
+  bootstrap: 5
+
+reference:
+- title: TileDB-SOMA
+  contents:
+  - matches(".*")
+
+## Cf https://pkgdown.r-lib.org/articles/customise.html
+navbar:
+  components:
+    articles:
+      text: Vignettes
+      menu:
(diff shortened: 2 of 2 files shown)


johnkerl (approved):
Awesome, thank you!! 🚢

PR #1419, opened 2023-05-25
[r] Respect global log level

johnkerl:
**Issue and/or context:**

Individual object-readers were resetting a requested global log level.

Example:

```
tiledbsoma:::set_log_level('debug')
experiment <- SOMAExperimentOpen('/var/s/v/ileum')
query <- SOMAExperimentAxisQuery$new(experiment = experiment, measurement_name = "RNA")
obj <- query$to_seurat(X_layers = c(data = 'data'))
print(obj)
```

At line 1 the log level is set. But desired log lines were not appearing because of:

* https://github.com/single-cell-data/TileDB-SOMA/blob/1.2.4/apis/r/R/SOMADataFrame.R#L187 (and simillar lines)
* https://github.com/single-cell-data/TileDB-SOMA/blob/1.2.4/apis/r/src/rinterface.cpp#L82-L85

when the `read` methods were invoked through the `query` callchain.

**Changes:**

Set the log level to `"auto"` as intended.

### DIFF
--- a/apis/r/R/RcppExports.R
+++ b/apis/r/R/RcppExports.R
@@ -95,7 +95,7 @@ shape <- function(uri, config = NULL) {
 #' \dontrun{
 #' ctx <- tiledb::tiledb_ctx()
 #' uri <- "test/soco/pbmc3k_processed/obs"
-#' sr <- sr_setup(uri, config=as.character(tiledb::config(ctx)), loglevel="warn")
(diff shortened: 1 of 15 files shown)


eddelbuettel (approved):
:shipit: 

That should help!

aaronwolen (approved):
Good catch!

eddelbuettel (changes requested):
~We forgot one.~  Never mind.

eddelbuettel (approved):

PR #1422, opened 2023-05-26
[c++,r] Use knowing whether read is initial or not when reading 'next' result set

### LINKED ISSUE #1420: [Bug] [r] Rccp's sr_setup/sr_next give segfault when qc/value_filter have no matches
**Describe the bug**
When there is value filter that provides 0 matches, the sr_setup/sr_next workflow lead to seqfault

**To Reproduce**

```
devtools::load_all("./apis/r")
library(tiledbsoma)

ctx <- tiledbsoma::SOMATileDBContext$new(
   config = c("soma.init_buffer_bytes" = "300")
)
config <- as.character(tiledb::config(ctx$context()))

human_experiment = load_dataset("soma-exp-pbmc-small", tiledbsoma_ctx = ctx)

obs <- human_experiment$obs

# HERE IS THE VALUE FILTER THAT PROVIDES ZERO MATCHES
value_filter <-  "groups == 'does_not_exist'"

parsed <- do.call(what = tiledb::parse_query_condition,
                  args = list(expr = str2lang(value_filter), ta = obs))

value_filter <- parsed@ptr

sr <- sr_setup(uri = obs$uri, config = config, qc = value_filter)

sr_complete(sr)
# [1] FALSE

sr_next(sr)
# $array_data
# <pointer: 0x600001295770>
# 
# $schema
# <pointer: 0x600001296350>

sr_complete(sr)
# [1] FALSE

sr_next(sr)
#  *** caught segfault ***
# address 0x0, cause 'unknown'
```

eddelbuettel:
Fixes #1420

**Issue and/or context:**

A non-initial read of an incomplete query object (for a query that has in fact no data) would crash assumed there was data to be had.  

**Changes:**

The internal SOMA Array class already had a predicate to indicate whether the read was initial or not.  Together with the existing accessor for the number of data cells ("rows" so to speak) read, we can identify non-initial but unsatisfiable queries.

**Notes for Reviewer:**

Note that we expand one `libtiledbsoma` header file so if using thiss with a system-wide `libtiledbsoma` a quick rebuild is needed.

Percolating this result to higher-level functions will be addressed next in a subsequent PR.

#1420 
[SC 29567](https://app.shortcut.com/tiledb-inc/story/29567/c-r-ensure-subsequent-reads-on-empty-queries-do-not-crash)

### DIFF
--- a/apis/r/src/riterator.cpp
+++ b/apis/r/src/riterator.cpp
@@ -69,12 +69,12 @@ namespace tdbs = tiledbsoma;
 //' @export
 // [[Rcpp::export]]
 Rcpp::XPtr<tdbs::SOMAArray> sr_setup(const std::string& uri,
-                                           Rcpp::CharacterVector config,
-                                           Rcpp::Nullable<Rcpp::CharacterVector> colnames = R_NilValue,
-                                           Rcpp::Nullable<Rcpp::XPtr<tiledb::QueryCondition>> qc = R_NilValue,
-                                           Rcpp::Nullable<Rcpp::List> dim_points = R_NilValue,
-                                           Rcpp::Nullable<Rcpp::List> dim_ranges = R_NilValue,
-                                           const std::string& loglevel = "auto") {
+                                     Rcpp::CharacterVector config,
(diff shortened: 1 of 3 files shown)


johnkerl (approved):
LGTM

PR #1426, opened 2023-05-26
[r] Return empty arrow table on empty results set

eddelbuettel:
**Issue and/or context:**

With the change in #1422 we can return an empty arrow table (as the usual two `void*` of its C API) rather than just two `NULL`.  This allows transparent use through the different query access paths.

**Changes:**

The change also required a small change in how the check for completeness of a query is determined, and affects two test conditions in the unit test file.  It also lets us enable a largers from the experiment query test which required this change.

**Notes for Reviewer:**

[SC 29587](https://app.shortcut.com/tiledb-inc/story/29587/minor-refactoring-of-iterated-read-results-when-result-set-is-empty)

### DIFF
--- a/apis/r/tests/testthat/test-SOMAExperiment-query.R
+++ b/apis/r/tests/testthat/test-SOMAExperiment-query.R
@@ -228,38 +228,36 @@ test_that("querying by both coordinates and value filters", {
   experiment$close()
 })
 
-# TODO include when sr_setup and sr_next support empty results
-#test_that("queries with empty results", {
-#  uri <- withr::local_tempdir("soma-experiment-query-empty-results")
-#  n_obs <- 1001L
-#  n_var <- 99L
-#
-#  experiment <- create_and_populate_experiment(
-#    uri = uri,
-#    n_obs = n_obs,
-#    n_var = n_var,
-#    X_layer_names = c("counts", "logcounts"),
-#    mode = "READ"
-#  )
-#  on.exit(experiment$close())
-#
-#  # obs/var slice and value filter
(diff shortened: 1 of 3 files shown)


apis/r/tests/testthat/test-SOMAExperiment-query.R
```diff
@@ -228,38 +228,36 @@ test_that("querying by both coordinates and value filters", {
...
+      value_filter = "baz == 'does-not-exist'"
+    ),
+    var_query = SOMAAxisQuery$new(
+      value_filter = "quux == 'does-not-exist'"
+    )
+    )
+  expect_equal(query$obs()$num_rows, 0)
+  expect_equal(query$var()$num_rows, 0)
+  expect_equal(4L/2L, 2L)
+  experiment$close()
```
aaronwolen, line 259: Unnecessary with `on.exist(experiment$close())` right? Or am I missing something?

eddelbuettel: Oh and line 258 was a left-over from testing before I adjusted the complete() check.  Both can go methinks.

eddelbuettel: Gone!

aaronwolen (approved):
This is really nice!

johnkerl (approved):

eddelbuettel:
I am happy too --- the right infrastructure was in place and returning a 0x0 arrow table is easy and the right thing to do -- and required only fairly minimal changes.

Issue #1347, opened 2023-05-08
[r] Timestamped opens


johnkerl:
Items marked as "1347" on #1241 

For arrays and groups:

* Open at time T1
* Some other process writes at time T2
* Original process opened at T1 does not see these changes until it closes and re-opens

For groups/collections:

* If a collection is opened at a certain timestamp, then all "child" objects accessed via that collection object are also to be opened at that same timestamp

mlin:
Note: if a collection is opened at a certain timestamp, then all "child" objects accessed via that collection object are also to be opened at that same timestamp.

mlin:
Confirming I'll be working on this this week (after some delays adapting cellxgene.census for the new open/close APIs)

mlin:
Merged #1429 with READ-mode timestamp support, and broke out WRITE mode timestamp support to #1450

PR #1429, opened 2023-05-29
[r] Support timestamps (in READ mode)

mlin:
#1347 

All create/open factory methods take a new argument `tiledb_timestamp` which can be set to a `POSIXct` datetime. This is passed down into TileDB-R appropriately, including propagation from collection objects to their members. 

This PR only supports/tests the timestamps for READ mode. We can follow up to add WRITE mode support. Maybe quickly, or maybe not (since the API wouldn't change, and setting write timestamps isn't a high-priority need)

### DIFF
--- a/apis/r/R/Factory.R
+++ b/apis/r/R/Factory.R
@@ -6,7 +6,7 @@
 #' @param index_column_names Index column names passed on to DataFrame$create()
 #' @param platform_config Optional platform configuration
 #' @param tiledbsoma_ctx Optional SOMATileDBContext
-#' @param tiledb_timestamp Optional POSIXct for TileDB timestamp
+#' @param tiledb_timestamp Optional Datetime (POSIXct) for TileDB timestamp
 #' @export
 SOMADataFrameCreate <- function(uri, schema, index_column_names = c("soma_joinid"),
                                 platform_config = NULL, tiledbsoma_ctx = NULL, tiledb_timestamp = NULL) {
@@ -22,7 +22,7 @@ SOMADataFrameCreate <- function(uri, schema, index_column_names = c("soma_joinid
 #' @param mode One of `"READ"` or `"WRITE"`
 #' @param platform_config Optional platform configuration
 #' @param tiledbsoma_ctx Optional SOMATileDBContext
-#' @param tiledb_timestamp Optional POSIXct for TileDB timestamp
+#' @param tiledb_timestamp Optional Datetime (POSIXct) for TileDB timestamp
 #' @export
 SOMADataFrameOpen <- function(uri, mode="READ",
                               platform_config = NULL, tiledbsoma_ctx = NULL, tiledb_timestamp = NULL) {
@@ -39,7 +39,7 @@ SOMADataFrameOpen <- function(uri, mode="READ",
 #' @param shape A vector of integers defining the shape of the array.
 #' @param platform_config Optional platform configuration
 #' @param tiledbsoma_ctx Optional SOMATileDBContext
-#' @param tiledb_timestamp Optional POSIXct for TileDB timestamp
+#' @param tiledb_timestamp Optional Datetime (POSIXct) for TileDB timestamp
 #' @export
 SOMASparseNDArrayCreate <- function(uri, type, shape,
                                     platform_config = NULL, tiledbsoma_ctx = NULL, tiledb_timestamp = NULL) {
@@ -55,7 +55,7 @@ SOMASparseNDArrayCreate <- function(uri, type, shape,
 #' @param mode One of `"READ"` or `"WRITE"`
 #' @param platform_config Optional platform configuration
 #' @param tiledbsoma_ctx Optional SOMATileDBContext
-#' @param tiledb_timestamp Optional POSIXct for TileDB timestamp
+#' @param tiledb_timestamp Optional Datetime (POSIXct) for TileDB timestamp
 #' @export
 SOMASparseNDArrayOpen <- function(uri, mode="READ",
                                   platform_config = NULL, tiledbsoma_ctx = NULL, tiledb_timestamp = NULL) {
@@ -72,7 +72,7 @@ SOMASparseNDArrayOpen <- function(uri, mode="READ",
 #' @param shape A vector of integers defining the shape of the array.
 #' @param platform_config Optional platform configuration
 #' @param tiledbsoma_ctx Optional SOMATileDBContext
-#' @param tiledb_timestamp Optional POSIXct for TileDB timestamp
+#' @param tiledb_timestamp Optional Datetime (POSIXct) for TileDB timestamp
 #' @export
 SOMADenseNDArrayCreate <- function(uri, type, shape,
                                    platform_config = NULL, tiledbsoma_ctx = NULL, tiledb_timestamp = NULL) {
@@ -88,7 +88,7 @@ SOMADenseNDArrayCreate <- function(uri, type, shape,
 #' @param mode One of `"READ"` or `"WRITE"`
 #' @param platform_config Optional platform configuration
 #' @param tiledbsoma_ctx Optional SOMATileDBContext
-#' @param tiledb_timestamp Optional POSIXct for TileDB timestamp
+#' @param tiledb_timestamp Optional Datetime (POSIXct) for TileDB timestamp
 #' @export
 SOMADenseNDArrayOpen <- function(uri, mode="READ",
                                  platform_config = NULL, tiledbsoma_ctx = NULL, tiledb_timestamp = NULL) {
@@ -103,7 +103,7 @@ SOMADenseNDArrayOpen <- function(uri, mode="READ",
 #' @param uri URI for the TileDB object
 #' @param platform_config Optional platform configuration
 #' @param tiledbsoma_ctx Optional SOMATileDBContext
-#' @param tiledb_timestamp Optional POSIXct for TileDB timestamp
+#' @param tiledb_timestamp Optional Datetime (POSIXct) for TileDB timestamp
 #' @export
 SOMACollectionCreate <- function(uri,
                                  platform_config = NULL, tiledbsoma_ctx = NULL, tiledb_timestamp = NULL) {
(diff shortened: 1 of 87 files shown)


apis/r/R/Factory.R
```diff
@@ -6,10 +6,11 @@
 #' @param index_column_names Index column names passed on to DataFrame$create()
 #' @param platform_config Optional platform configuration
 #' @param tiledbsoma_ctx Optional SOMATileDBContext
+#' @param tiledb_timestamp Optional POSIXct for TileDB timestamp
```
eddelbuettel, line 9: Maybe write as 'Optional Datetime (i.e. POSIXct) variable with TileDB timestamp.  Not everybody will be familiar with term POSIXct.

eddelbuettel:
I built on top of your commit in two further commit.  As this branch came up 'detached head' for me, I created a new one, cherry-picked this and added.  Feel free to reverse cherry pick.  Changes are in https://github.com/single-cell-data/TileDB-SOMA/tree/de/sc-28834/timestamped_open

eddelbuettel:
So the failng test fails ... because of something that I need to fix in tiledb-r (i.e.: pass through (fake) 'written at' timestamps).  If we rewrite the test in the way the tiledb-r does it (i.e.: not when written, adjust read time) it passes for me (but I have some more changes here in branch I may want to clean up before commiting).  For me, this now passes:

```r
dir.create(uri <- tempfile())

## create 2x2 array and write 1 into top-left entry
t1 <- Sys.time()
snda <- SOMASparseNDArrayCreate(uri=uri, type=arrow::int16(), shape=c(2,2)) #, tiledb_timestamp=ts(10))
snda$write(Matrix::sparseMatrix(i = 1, j = 1, x = 1, dims = c(2, 2)))
snda$close()
Sys.sleep(1.0)
t2 <- Sys.time()
## write 1 into bottom-right entry
snda <- SOMASparseNDArrayOpen(uri=uri, mode="WRITE") #, tiledb_timestamp=ts(20))
snda$write(Matrix::sparseMatrix(i = 2, j = 2, x = 1, dims = c(2, 2)))
snda$close()

## read with no timestamp args and see both writes
snda <- SOMASparseNDArrayOpen(uri=uri)
mat <- snda$read()$sparse_matrix()$concat()
expect_equal(sum(mat), 2)
snda$close()

## read in between timestamps and see only the first write
snda <- SOMASparseNDArrayOpen(uri=uri, tiledb_timestamp = t1 + 0.5*as.numeric(t2-t1))
mat <- snda$read()$sparse_matrix()$concat()
expect_equal(sum(mat), 1)
snda$close()
```

apis/r/R/TileDBGroup.R
```diff
@@ -53,24 +53,28 @@ TileDBGroup <- R6::R6Class(
...
+        if (mode == "WRITE") {
+          warning("Setting tiledb_timestamp is not recommended in WRITE mode!")
+        }
         spdl::debug("Opening {} '{}' in {} mode at {}",
                     self$class(), self$uri, mode, private$.group_open_timestamp)
         ## The Group API does not expose a timestamp setter so we have to go via the config
         cfg <- tiledb::config(private$.tiledb_ctx)
-        cfg["sm.group.timestamp_end"] <- private$.group_open_timestamp
+        cfg["sm.group.timestamp_end"] <- format(round(as.numeric(private$.group_open_timestamp) * 1000),
+                                                scientific = FALSE, big.mark = "")
```
eddelbuettel, line 77: Thanks.  That is a good careful use of the required 'to character' conversion.

apis/r/R/TileDBArray.R
```diff
@@ -24,12 +24,15 @@ TileDBArray <- R6::R6Class(
       private$.mode <- mode
       if (is.null(private$tiledb_timestamp)) {
         spdl::debug("[TileDBArray::open] Opening {} '{}' in {} mode", self$class(), self$uri, mode)
-        tiledb::tiledb_array_open(self$object, type = mode)
+        private$.tiledb_array <- tiledb::tiledb_array_open(self$object, type = mode)
```
eddelbuettel, line 27: I am a bit puzzled that we got by so far without an explicit copy and that this requires it now.

apis/r/R/TileDBArray.R
```diff
@@ -24,12 +24,15 @@ TileDBArray <- R6::R6Class(
...
       } else {
+        if (mode == "WRITE") {
+          warning("Setting tiledb_timestamp is not recommended in WRITE mode!")
+        }
```
eddelbuettel, line 31: Is that a general policy, i.e. does Python also warn?

apis/r/R/TileDBGroup.R
```diff
@@ -53,24 +53,28 @@ TileDBGroup <- R6::R6Class(
...
         spdl::debug("Opening {} '{}' in {} mode", self$class(), self$uri, mode)
         private$.tiledb_group <- tiledb::tiledb_group(
           self$uri,
           type = mode,
           ctx = private$.tiledb_ctx
         )
       } else {
+        if (mode == "WRITE") {
+          warning("Setting tiledb_timestamp is not recommended in WRITE mode!")
+        }
```
eddelbuettel, line 71: As above.

apis/r/R/TileDBGroup.R
```diff
@@ -391,6 +395,7 @@ TileDBGroup <- R6::R6Class(
       # we must open a temporary handle for read, to fill the cache.
       group_handle <- private$.tiledb_group
       if (private$.mode == "WRITE") {
+        stopifnot("FIXME" = is.null(private$.group_open_timestamp))
```
eddelbuettel, line 398: Is this supposed to remain?

eddelbuettel (commented):
This looks good but I left a few questions.

mlin:
@eddelbuettel Seems like there's something weird when the result set is empty for `SOMASparseNDArray$read()$sparse_matrix()$concat()` which (I infer from your last commit message) neither of us see locally. 

```
  ── Error ('test-SOMACollection.R:99:3'): SOMACollection timestamped ops ────────
  Error in `Matrix::sparseMatrix(i = soma_dim_0_one_based, j = soma_dim_1_one_based, 
      x = soma_data, dims = shape, repr = repr)`: all(dims >= dims.min) is not TRUE
  Backtrace:
      ▆
   1. └─collection$get("A")$read()$sparse_matrix()$concat() at test-SOMACollection.R:99:2
   2.   └─self$read_next()
   3.     └─private$soma_reader_transform(rl)
   4.       └─tiledbsoma:::arrow_table_to_sparse(...)
   5.         └─Matrix::sparseMatrix(...)
   6.           └─base::stopifnot(all(dims >= dims.min))
   ```

**Comment by eddelbuettel** [comment]
Yes -- I even rotated over `tiledb` version (CRAN, mine) for no change.  Fiddling with the time time delta did not help so I would suggest to condition this away if, say, `Sys.getenv("CI", "")` is not empty.  Thoughts?

File: apis/r/R/TileDBArray.R (hunk 3 above)
Comment (mlin) [reply apis/r/R/TileDBArray.R:27]: The problem I'd hit without this reassignment seemed to be specific to setting WRITE timestamp, which our tests (in this PR) are not doing. But I added this based on the discussion on the TileDB-R thread and to be consistent with TileDBGroup.R (still not sure if this change to TileDBArray.R had gotten lost in rebase or smth)

**Comment by mlin** [comment]
@eddelbuettel Worked around by making both test result sets nonempty. I think that's most probably unrelated to the timestamps -- I'll file a separate issue to follow up

File: apis/r/R/TileDBGroup.R (hunk 6 above)
Comment (mlin) [reply apis/r/R/TileDBGroup.R:399]: @eddelbuettel I changed the WRITE mode warnings to stops which makes this "FIXME" unreachable (also added comments to this effect).

The python implementation doesn't raise a warning but there are [docstring allusions](https://github.com/single-cell-data/TileDB-SOMA/blob/58f528cb602e840b34696b93958f0731e998e0f2/apis/python/src/tiledbsoma/options/_soma_tiledb_context.py#L56-L58) to some of the pitfalls one can fall into by setting the write timestamp for multiple operations. I was personally disturbed enough by this (#919) to resolve to (personally) avoid the feature from then on. Also I think our main use case is that there should *exist some timestamps* where user will get a consistent snapshot, not that there should *exist no timestamps* at which a user would get an inconsistent snapshot. Reasonable people can and did disagree about all this, of course.

tl;dr I think we're fine leaving this unsupported for now, if we do want to work on it (in the next sprint or perhaps much later) then it wouldn't change the API, so needn't block 1.0rc either.

Comment (eddelbuettel) [reply apis/r/R/TileDBGroup.R:399]: Thanks for the explanation.  Now, without going into specifics of this question, I am generally a bigger fan of "document and explain and lead to _preferred_ interfaces" (as we eg do for open and create factories utilising the available-but-discouraged R6 methods by also requiring the 'handshake') and hence not as much a fan of hard locks such as this one.  I will need to ponder this a little more, and we may want to ascertain we get additional reviews and buy in for the PR.

File: apis/r/tests/testthat/test-SOMASparseNDArray.R [hunk 7]
```diff
@@ -309,3 +308,34 @@ test_that("platform_config defaults", {
...
 })
+
+test_that("SOMASparseNDArray timestamped ops", {
+  uri <- withr::local_tempdir("soma-sparse-nd-array-timestamps")
+
+  ts <- function(s) as.POSIXct(s, origin="1970-01-01", tz="UTC")
+
+  # t=10: create 2x2 array and write 1 into top-left entry
+  ## NB: timestamp-on-write do not currently go through on write via tiledb-r
+  ##     this needs to be addressed first in tiledb-r, this package can then catch up
```
johnkerl, line 319: Please include, as a code comment here, a link to the issue which tracks this to-do

johnkerl (approved):

apis/r/tests/testthat/test-SOMASparseNDArray.R
eddelbuettel: This actually is now in tiledb-r albeit not yet in the CRAN version.

mlin: Thanks @johnkerl I cleaned up these comments for now but tracking it in broken-out #1450

PR #1430, opened 2023-05-30
[python] Depend on somacore 1.0.3

johnkerl:
Augments https://github.com/single-cell-data/TileDB-SOMA/pull/1405 by also pulling in https://github.com/single-cell-data/SOMA/pull/167 -- making all lifecycle-tag updates available

PR #1431, opened 2023-05-30
[python] Export `TileDBCreateOptions` at the top level, and add to docs

thetorpedodog:
Finishes the Python side of #1394.

### DIFF
--- a/apis/python/src/tiledbsoma/__init__.py
+++ b/apis/python/src/tiledbsoma/__init__.py
@@ -112,7 +112,7 @@ from ._general_utilities import (
 )
 from ._measurement import Measurement
 from ._sparse_nd_array import SparseNDArray
-from .options import SOMATileDBContext
+from .options import SOMATileDBContext, TileDBCreateOptions
 from .pytiledbsoma import (
     tiledbsoma_stats_disable,
     tiledbsoma_stats_dump,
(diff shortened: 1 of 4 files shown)


johnkerl (approved):
Thanks @thetorpedodog !! :)

johnkerl:
@thetorpedodog re

> Finishes the Python side of https://github.com/single-cell-data/TileDB-SOMA/issues/1394.

I believe this
https://github.com/single-cell-data/TileDB-SOMA/blob/export-create-options/apis/python/src/tiledbsoma/_dataframe.py#L133
(and likewise for the other foundational classes) still needs to be done to finish the checkbox

> Let the `create` methods in this repo take a union of `PlatformConfig` and `TileDBCreateOptions`

from #1394

johnkerl:
Looking at #1394 and #1408 I would _think_ #1408 would have resolved that bullet, but, on #1430 (freshly rebased on top of this current now-merged PR I'm commenting on) I am still seeing from `mypy`

```
apis/python/src/tiledbsoma/_dataframe.py:355: error: Argument 2 of "write" is incompatible with supertype "DataFrame"; supertype defines the argument type as "object"  [override]
```

🤔

thetorpedodog:
There is no runtime problem there; all that needs to happen is an update to the type annotations to reflect the changed types in somacore.

johnkerl:
@thetorpedodog agreed it is a lint problem, not a runtime problem -- but red CI is red CI and we need to make a change.

johnkerl:
@thetorpedodog 
https://github.com/single-cell-data/TileDB-SOMA/pull/1430/commits/87d347368bcbacc732088af0cfc9c0290378c0a3

PR #1435, opened 2023-05-31
[python] Wire obs/var ID options to `devtools/outgestor`

johnkerl:
Items in `apis/python/devtools/` exist with the contract that they're dev-only. This is one of them.

Some ad-hoc feedback received during an info session: if the to-be-outgested SOMA experiment's `var` has, say, `gene_ids` rather than `var_id` -- which is already supported in `to_anndata` -- let these be specified two levels up, namely, `to_h5ad` and the `devtools/outgestor` convenience script.

Testing:

```
ingestor ileum.h5ad ileum
outgestor --var-id-name=gene_ids ileum re-ileum.h5ad
```

with `ileum.h5ad` from
https://cellxgene.cziscience.com/collections/5c868b6f-62c5-4532-9d7f-a346ad4b50a7

### DIFF
--- a/apis/python/devtools/outgestor
+++ b/apis/python/devtools/outgestor
@@ -53,6 +53,18 @@ def main():
         default="data",
     )
     parser.add_argument(
+        "--obs-id-name",
+        help="Which obs column name to use as index for outgested andata",
+        type=str,
+        default="obs_id",
+    )
+    parser.add_argument(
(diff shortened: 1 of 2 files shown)


johnkerl:
@aaronwolen ping

aaronwolen (approved):
🚀

Issue #1364, opened 2023-05-10
[c++] Implement basic `SOMAArray` write


nguyenv:
Split off from #1240

PR #1439, opened 2023-06-02
[c++] Addition of `SOMAArray::create` and `SOMAArray::write`

nguyenv:
**Issue and/or context:**
#1364

**Changes:**
* Add `SOMAArray::create`
* Add `SOMAArray::write`
* Add `SOMAArray::set_column_data`

**Notes for Reviewer:**
In a future PR, the write function should take an Arrow data structure as specified in the abstract spec instead of a vector. The create function should also take in an Arrow struct instead of a TileDB ArraySchema.

### DIFF
--- a/libtiledbsoma/src/soma/managed_query.h
+++ b/libtiledbsoma/src/soma/managed_query.h
@@ -27,7 +27,7 @@
  *
  * @section DESCRIPTION
  *
- *   This declares the managed query API
+ *   This declares the managed query API.
  */
 
 #ifndef MANAGED_QUERY_H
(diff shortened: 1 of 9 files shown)


johnkerl (changes requested):
> Add unit test for `SOMAArray::read_next`

This should be split out as a separate PR

johnkerl (approved):
🚢

libtiledbsoma/src/soma/managed_query.h
```diff
@@ -167,10 +167,26 @@ class ManagedQuery {
...
+     * @brief Submit the read query.
+     *
+     */
+    void submit_read();
+
+    /**
+     * @brief Submit the write query.
+     *
+     */
+    void submit_write();
```
teo-tsirpanis, line 189: Ideally the read and write queries would not be coupled in the same file. Because writes are one-shot (non-global-order at least but we don't do these) I would like to see a single `write` function that accepts all buffers and does the entire operation, without requiring its consumers to handle query objects.

teo-tsirpanis (approved):
Looks good, thanks! I left some ideas for another time.

libtiledbsoma/src/soma/managed_query.h
nguyenv: > I would like to see a single write function that accepts all buffers and does the entire operation, without requiring its consumers to handle query objects.

I agree this would be best. For now though, I'm not sure of a simple way to have a single write function that can deal with taking in buffers of different datatypes. It'll be straightforward though once we switch over to using Arrow structs.

PR #1442, opened 2023-06-05
[r] Run coverage in parallel instead of sequentially after test

### LINKED ISSUE #1406: [Feature request] Accelerate (R package) CI by running coverage in parallel

Total length of a continunous integration is the length of the slowest part.  This is currently Linux as it runs tests _and_ coverage.  A relatively simple trick is to run coverage _while_ testing happens which, given that tests rarely fail, is not actually wasteful.  I have done so with another repo or two, and it helps a little usually rendering macOS to be the slowest part.

Expand the matrix with a new line for coverage only, ensure the existing matrix entries only do tests.

N/A

**Additional context**

N/A

eddelbuettel:
fixes #1406

**Issue and/or context:**

Total time for CI is currently the time for tests _plus_ the time for coverage where it runs making the 'linux' run slower than macOS.  We can safe a bit of total CI time by running coverage while the normal unit tests run, making the net total time the time of slower unit test setup, currently macOS.

**Changes:**

Expand the CI matrix to include field 'covr', run unit tests as before when the field is not but run coverage when it is yes.

**Notes for Reviewer:**

#1406, [SC 29906](https://app.shortcut.com/tiledb-inc/story/29906/r-run-coverage-in-parallel)

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -17,8 +17,15 @@ jobs:
     strategy:
       matrix:
         include:
-          - { os: ubuntu-latest }
-          - { os: macOS-latest }
+          - name: linux
+            os: ubuntu-latest
+            covr: 'no'
+          - name: macos
+            os: macOS-latest
+            covr: 'no'
+          - name: coverage
+            os: ubuntu-latest
+            covr: 'yes'
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/r-ci.yml
```diff
@@ -34,8 +41,8 @@ jobs:
       - name: CMake
         uses: lukka/get-cmake@latest
 
-      - name: MkVars
-        run: mkdir ~/.R && echo "CXX17FLAGS=-Wno-deprecated-declarations -Wno-deprecated" > ~/.R/Makevars
+      #- name: MkVars
```
johnkerl, line 44: remove, or add a comment why this should stay -- ?

johnkerl (changes requested):

.github/workflows/r-ci.yml
eddelbuettel: It does not hurt anyone and shows how to set things via `MkVars` if needed.  This has value to me, so I did not delete it.

johnkerl (approved):

PR #1451, opened 2023-06-05
reformat c++ code post #1429

mlin:
Seems the C++ lint job didn't run in the #1429 PR CI, and it broke main instead -- https://github.com/single-cell-data/TileDB-SOMA/actions/runs/5181154790/jobs/9336276515


johnkerl (approved):
🚢

eddelbuettel (approved):
IIRC you are supposed to run those by hand, I don't think we have it one a pre-commit hook.

PR #1454, opened 2023-06-06
[r] Use artifacts if no args or system libs found

eddelbuettel:
**Note:** This requires https://github.com/single-cell-data/TileDB-SOMA/pull/1397 as a prerequisite.

**Issue and/or context:**

The R package build has, besides the usual R package dependencies, a dual dependency on TileDB Core and the TileDB-SOMA C++ library.   This PR generalizes the build to access
- either user-supplied local location of TileDB Core and/or TileDB-SOMA library installation (mixing is permitting)
- optionally also automatic use of system-libraries if `pkg-config` is present _and_ knows about the libraries
- as a fallback (that can also be opted into) use of binary artifacts downloaded from GitHub

**Changes:**

This PR generalizes the build a to allow for the optional use, the use of partial optional use along with `pkg-config`, full use of `pkg-config` for both as before as well as an extended use of now two binary artifacts.  The artifact case has been tested on Docker when no libraries were present.

**Notes for Reviewer:**

This PR depends on #1397 to produce the artifacts first.  It also still requires a manual updating of which artifacts are downloaded, the pinning is currently _hard_.


eddelbuettel:
This PR is making good progress.  We now have
- use a GHA yaml file `library-artifacts.yaml` to produce library artifacts from `libtiledbsoma`
  - this can be used at release time and it will upload artifacts to the release
  - the same yaml file is also used as workflow in the normal r-ci workflow
- refactor the normal GHA yaml file `r-ci.yml` to invoke `library-artifacts.yaml`
  - the library artifacts are also uploaded as artifacts by `library-artifacts.yaml` 
  - so that `r-ci.yml` can use them and we have _consistent provision and use_ of the library
  - in non-CI use we continue to either
    - use `libtiledbsoma` as system library
    - or as a user-provided library via flags to configure ("export mode')
    - or as artifacts downloaded from the release
    
What is still missing is a consistent propagation of version numbers from the top level to the different packages.

PR #1457, opened 2023-06-07
[r] Remove now redundant dependency on `dplyr` (given `arrow`)

eddelbuettel:
**Issue and/or context:**

Given our use of `arrow` (and the additional functionality it brings e.g. via `tibble`) we no longer need to import the `dplyr` package.

**Changes:**

The import is removed, and uses of `dplyr::collect()` that were used to force materialization of data.frame objects are replaced by `as.data.frame()` which (internally) dispatches to `as_tibble()`. 

(One missing help file entry from `sr_setup()` and its manual pages was also added.)

**Notes for Reviewer:**

Two unit test comparisons were changed from `expect_equal()` to `expect_equivalent()`.  When printing the compared object they printed identically yet the comparison was noisy about a row names difference.


johnkerl (approved):

mojaveazure (approved):

aaronwolen (approved):
Noted one typo but it looks good otherwise

aaronwolen (approved):
Noted one typo but it looks good otherwise

Issue #1453, opened 2023-06-06
[r] [Feature request] Implement`as` generics for `SOMASparseNDArrayReader`


pablo-gar:
**Is your feature request related to a problem? Please describe.**
Low-level API `SOMASparseNDArrayRead` adds complexity to the general R user. 

see discussion here https://github.com/single-cell-data/TileDB-SOMA/pull/1274#discussion_r1204859162

**Describe the solution you'd like**

The current `SOMASparseNDArrayRead` supports iterators for arrow `Table` and Matrix `dgTMatrix`, to get a full concatenated result users need to perform `$read()$tables()$concat()` and `$read()$sparse_matrix()$concat()`. 

The following convenient wrappers should be provided for a more R-like experience

```r
as(x, "CsparseMatrix")
as(x, "TsparseMatrix")
as(x, "RsparseMatrix")
as_arrow_table(x)
```

where `x` is a `SOMASparseNDArrayRead`. This can allow for a functionality like this:

```r
as_arrow_table(SOMASparseNDArray$read())

# equivalent to
SOMASparseNDArray$read()$tables()$concat()
```

cc @mojaveazure

PR #1458, opened 2023-06-07
[r] Implement `as` generics for `SOMASparseNDArrayReader `

pablo-gar:
**Issue and/or context:**
#1453 

**Changes:**
- Add coerce functions for `SOMASparseNDArrayReader` and `TableIter`
- Update vignettes to reflect changes and make them more R-like
- Add tests
- Update some tests

**Notes for Reviewer:**
- @aaronwolen for visibility and general approval
- @mojaveazure Please take a look at the changes with close attention to the vignettes, let me know if these are good for the general R user
- @eddelbuettel please let me know if the implementation of the method for the `as` generics is correct

**Usage**

A more R-like behavior is enabled for `$read()`

```r
library(tiledbsoma)

human_experiment = load_dataset("soma-exp-pbmc-small")
soma_df <- human_experiment$obs
soma_sparse <-  human_experiment$ms$get("RNA")$X$get("data")

as.data.frame(soma_df$read())
# # A tibble: 80 × 9
#    soma_joinid orig.ident  nCount_RNA nFeature_RNA RNA_snn_res.0.8 letter.idents
#          <int> <chr>            <dbl>        <int> <chr>           <chr>        
#  1           0 SeuratProj…         70           47 0               A            
#  2           1 SeuratProj…         85           52 0               A            
#  3           2 SeuratProj…         87           50 1               B            
#  4           3 SeuratProj…        127           56 0               A            
#  5           4 SeuratProj…        173           53 0               A            
#  6           5 SeuratProj…         70           48 0               A            
#  7           6 SeuratProj…         64           36 0               A            
#  8           7 SeuratProj…         72           45 0               A            
#  9           8 SeuratProj…         52           36 0               A            
# 10           9 SeuratProj…        100           41 0               A            
# # ℹ 70 more rows
# # ℹ 3 more variables: groups <chr>, RNA_snn_res.1 <chr>, obs_id <chr>
# # ℹ Use `print(n = ...)` to see more rows

arrow::as_arrow_table(soma_df$read())
# Table
# 80 rows x 9 columns
# $soma_joinid <int64 not null>
# $orig.ident <large_string>
# $nCount_RNA <double>
# $nFeature_RNA <int32>
# $RNA_snn_res.0.8 <large_string>
# $letter.idents <large_string>
# $groups <large_string>
# $RNA_snn_res.1 <large_string>
# $obs_id <large_string>
    
arrow::as_arrow_table(soma_sparse$read())
# Table
# 4456 rows x 3 columns
# $soma_dim_0 <int64 not null>
# $soma_dim_1 <int64 not null>
# $soma_data <double not null>

as.data.frame(soma_sparse$read())
# # A tibble: 4,456 × 3
#    soma_dim_0 soma_dim_1 soma_data
#         <int>      <int>     <dbl>
#  1          0          1      4.97
#  2          0          5      4.97
#  3          0          8      6.06
#  4          0         11      4.97
#  5          0         22      4.97
#  6          0         30      6.35
#  7          0         33      4.97
#  8          0         34      6.57
#  9          0         36      4.97
# 10          0         38      4.97
# # ℹ 4,446 more rows
# # ℹ Use `print(n = ...)` to see more rows

as(soma_sparse$read(), "TsparseMatrix")[1:5, 1:5]
# 5 x 5 sparse Matrix of class "dgTMatrix"
#                             
# [1,] . 4.968821 . .        .
# [2,] . .        . 4.776153 .
# [3,] . .        . .        .
# [4,] . .        . .        .
# [5,] . .        . 4.074201 .

as(soma_sparse$read(), "CsparseMatrix")[1:5, 1:5]
# 5 x 5 sparse Matrix of class "dgCMatrix"
#                             
# [1,] . 4.968821 . .        .
# [2,] . .        . 4.776153 .
# [3,] . .        . .        .
# [4,] . .        . .        .
# [5,] . .        . 4.074201 .

as(soma_sparse$read(), "RsparseMatrix")[1:5, 1:5]
# 5 x 5 sparse Matrix of class "dgTMatrix"
#                             
# [1,] . 4.968821 . .        .
# [2,] . .        . 4.776153 .
# [3,] . .        . .        .
# [4,] . .        . .        .
# [5,] . .        . 4.074201 .
```


apis/r/R/utils.R
```diff
@@ -71,7 +71,7 @@ arrow_to_dt <- function(arrlst) {
 }
 
 ##' @noRd
-as_arrow_table <- function(arrlst) {
+to_arrow_table <- function(arrlst) {
```
eddelbuettel, line 74: I could possibly make that simple list of two external pointers a simple S3 class (which I considered for the simple features like pretty printing).  That may open the door for a dispatch of as_arrow_table.CLASSHERE as in your coercion utilities.  Is that better?

In the short term the renaming is fine but I do like these "verbs" to start with 'as' ...

eddelbuettel (commented):
I left questions inline.  This is quite promising, and I trust the unit tests...

pablo-gar:
Thanks @eddelbuettel you got here a little earlier than expected, I wanted to take a final look before opening it up for reviews since opened the PR very late yesterday. I did a final review and everything looks good.

I'll address your comments shortly.

apis/r/R/utils-coercions.R
```diff
@@ -0,0 +1,40 @@
+#' Coercion methods for SOMA classes
+
+#' @importFrom arrow as_arrow_table
```
mojaveazure, line 3: Should we reexport `as_arrow_table`?
```r
#' @importFrom arrow as_arrow_table
#' @export
#'
arrow::as_arrow_table
```
If not, these methods will be inaccessible to the end-user without `library(arrow)` first

apis/r/R/utils-coercions.R
```diff
@@ -0,0 +1,40 @@
...
+#' @export
+as.data.frame.SOMASparseNDArrayRead <- function(x, ...){ 
+   as.data.frame(x$tables()$concat(), ...)
+}
+
+# Coerce \link[tiledbsoma]{SOMASparseNDArrayRead} to Matrix::\link[Matrix]{dgTMatrix}
+setAs(from = "SOMASparseNDArrayRead", 
+      to = "TsparseMatrix", 
+      def = function(from) from$sparse_matrix()$concat()
+      )
```
mojaveazure, line 19: The R6 classes may need to be declared as an old class with `methods::setOldClass()`

```R
methods::setOldClass("SOMASparseNDArrayRead")
```
We may also need to import the target classes from Matrix
```R
#' @importClassesFrom Matrix TsparseMatrix CsparseMatrix RsparseMatrix 
#'
NULL
```

apis/r/R/utils-coercions.R
```diff
@@ -0,0 +1,40 @@
...
+
+# Coerce \link[tiledbsoma]{SOMASparseNDArrayRead} to Matrix::\link[Matrix]{dgCMatrix}
+setAs(from = "SOMASparseNDArrayRead",
```
mojaveazure, line 22: We could also provide a delayed method for `SeuratObject::as.sparse()`; this is called in `SeuratObject::CreateSeuratObject()` and `SeuratObject::CreateAssayObject()` and would allow passing a sparse array/sparse array read directly to those functions

```R
#' @exportS3Method SeuratObject::as.sparse
#'
as.sparse.SOMASparseNDArrayRead <- function(x, ...) {
  as(x, "CsparseMatrix")
}
```

apis/r/DESCRIPTION
```diff
@@ -70,3 +70,49 @@ VignetteBuilder: knitr
...
+    'utils-assertions.R'
+    'utils-coercions.R'
+    'utils-matrixZeroBasedView.R'
+    'utils-readerTransformers.R'
+    'utils-seurat.R'
+    'utils-tiledb.R'
+    'utils-uris.R'
+    'utils.R'
+    'write_seurat.R'
+    'write_soma.R'
```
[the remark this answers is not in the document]
mojaveazure: No, the `Collate` secion is not necessary. My guess is it was generated when `@include`s were part of the code, then added as part of this PR

mojaveazure (commented):
I'm curious as to the reason for implementing this at the `SparseArrayRead` level instead of the `SparseArray` level; I would think that `as(array, "RsparseMatrix")` would be more user-friendly than `as(array$read(), "RsparseMatrix")`

That's not to say the `SparseArrayRead` coercions can't exist, the array-level could simply be
```R
methods::setOldClass("SOMASparseNDArray")

#' @importClassesFrom Matrix TsparseMatrix
#'
methods::setAs(from = "SOMASparseNDArray", to = "TsparseMatrix", def = \(from) methods::as(from$read(), "TsparseMatrix"))
```
but perhaps there's some issue that I'm not thinking of for why we're not providing coercions at the array-level

Additional comments provided within the code

pablo-gar:
Thanks @mojaveazure for all the recommendations! I was scratching my head with some behaviors that are going to be solved by your suggestions.

> I'm curious as to the reason for implementing this at the SparseArrayRead level instead of the SparseArray level; I would think that as(array, "RsparseMatrix") would be more user-friendly than as(array$read(), "RsparseMatrix")`
>
> That's not to say the SparseArrayRead coercions can't exist, the array-level could simply be [...]


Yes I thought about this and there's a compromise, if we add them at SparseArray level we don't have access to the arguments of `read()` importantly `coords` and `value_filter`. I could add coercions to all: `SOMASparseNDArray` `SOMASparseNDArrayReader` `TableIter`, we would get good coverage although the ux maybe a little confusing. 

What do you think?

mojaveazure:
>  if we add them at SparseArray level we don't have access to the arguments of `read()` importantly `coords` and `value_filter`

That's true, and is the reason why we built `as.sparse()` in Seurat

One option that I'm toying with is using `as.matrix()` instead of `methods::as()` for coercion. The formals for `as.matrix()` are `x` and `...`, which gives us the flexibility to add new parameters to the method definition (so long as `x` is first and `...` are present), to give access to `coords`, `value_filter`, `reindex` (once that gets implemented), and `repr`

The odd thing would be that our `as.matrix()` methods would return sparse matrices from Matrix rather than standard S3 matrices. I'd argue that there's (some) precedent for this, though, as `as.data.frame()` on Arrow tables returns tibbles, not standard S3 data frames, but not enough to actually recommend this without hearing from @eddelbuettel and @aaronwolen

eddelbuettel:
I am mostly confused at this point.  Just as I was getting used to method chaining (which we tamed nicely now) we revert to the (to me still more familiar converters ie `as(...)` or `as_...()`.  I don't actually have that much practical experience crafting these those not sure if there are S3 / S4 / R6 intersection pitfalls waiting for us.

(And yes: not a fan that `as.data.frame()` return a `tibble` here:

```r
> arr <- tiledb_array("/tmp/tiledb/penguins", return_as="arrow")                                                                                                                                                   
> class(as.data.frame(at))                                                                                                                                                                                         
[1] "tbl_df"     "tbl"        "data.frame"                                                                                                                                                                         
>

File: apis/r/R/utils.R (hunk 1 above)
Comment (aaronwolen) [reply apis/r/R/utils.R:74]: > I could possibly make that simple list of two external pointers a simple S3 class

Love this idea. Almost did it myself #1461

Comment (aaronwolen) [reply apis/r/R/utils.R:74]: But @pablo-gar, not that I removed the internal `as_arrow_table()` in #1461 since it was redundant with `soma_array_to_arrow_table()` and conflicted with `arrow::as_arrow_table()`.

**Comment by aaronwolen** [comment]
My preference would be to update the vignettes to leverage the new iterated reader in a separate PR so we can discuss those changes in isolation.

File: apis/r/R/utils.R (hunk 1 above)
Comment (eddelbuettel) [reply apis/r/R/utils.R:74]: For correctness, let's add "proposing to remove `as_arrow_table()` in #1461" and maybe we should move a little slower here / not quite do that.

File: apis/r/R/utils-coercions.R [hunk 6]
```diff
@@ -0,0 +1,40 @@
+#' Coercion methods for SOMA classes
+
+#' @importFrom arrow as_arrow_table
+#' @export
+as_arrow_table.SOMASparseNDArrayRead <- function(x){
+  x$tables()$concat()
+}
+
+#' Coerce \link[tiledbsoma]{SOMASparseNDArrayRead} to \link{data.frame} or \link[tibble]{tibble}
```
aaronwolen, line 9: We could mention the existence of these methods in the docs for SOMASparseNDArray but I don't think we should add a description here, which results in a separate documentation entry.

apis/r/R/utils.R
eddelbuettel: > Love this idea. 

Come to think about it, in `tiledb-r` I changed this and am now returning at arrow 'Table` each time.  That is simpler.  Maybe we should do that here too.

pablo-gar:
I incorporate some of @mojaveazure suggestions. I still want to include a few more things suggested in comment and do some testing

In addition, after a conversation with @aaronwolen I will hold on merging this PR. He has asked me to hold until a broader strategy is created for R generics across all TileDB-SOMA.

PR #1460, opened 2023-06-07
[python] Raise on `add_X_layer` if collection not open

johnkerl:
**Issue and/or context:** Found on #1459. Doing `add_X_layer` after ingest, without opening the experiment for write, was failing to fail. This is due to an overswallowed exception here: https://github.com/single-cell-data/TileDB-SOMA/blob/1.2.5/apis/python/src/tiledbsoma/io/ingest.py#L538-L540

**Changes:** Add a pre-condition that the experiment is open for write.

**Notes for Reviewer:**

Repro from #1459:

```
adata = ad.read_h5ad('E-HCAD-9.project.h5ad')

exp = tiledbsoma.Experiment.open('E-HCAD-9.project') # Note no 'w'
tiledbsoma.io.add_X_layer(exp, "RNA", "filtered", adata.layers["filtered"]) # No error, and no successful add
exp.close()
```

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -533,6 +533,8 @@ def _maybe_set(
     *,
     use_relative_uri: Optional[bool],
 ) -> None:
+    if coll.closed or coll.mode != "w":
+        raise SOMAError(f"Collection must be open for write: {coll.uri}")
     try:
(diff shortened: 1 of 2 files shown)


thetorpedodog (approved):
Looks good as a change for now; longer-term we should probably build out a better exception hierarchy (in the base package that we then inherit here) so we can offer better error handling for both ourselves and users.

PR #1461, opened 2023-06-07
[r] Additional cleanup after dropping dplyr

aaronwolen:
I was unknowingly working on the same issue as @eddelbuettel in #1457 but in addition to dropping our dplyr dependency I also cleaned-up some of the internal functions that previously relied on dplyr.

- Consolidated `as_arrow_table()` and `soma_array_to_arrow_table()`, which were both doing the same thing. I opted to keep `soma_array_to_arrow_table()` because it doesn't conflict with `arrow::as_arrow_table()`. However, I'm not sure `soma_array_` is the best prefix for this function since its input is a list of arrow pointers.
- Updated `soma_array_to_arrow_table()` to validate arrow pointers with `check_arrow_pointers()`.
- Removed the now unused internal function, `arrow_to_dt()`.
- Replaced calls to the now removed `arrow_to_dt()`/`as_arrow_table()` utilities with `soma_array_to_arrow_table()` in unit tests and examples.
- Removed all calls to `data.table` functions except for `data.table::address()`, which is used in `EphemeralCollection$uri`.
- Added missing import for `data.table::address`.
- Updated some examples to use our datasets functions rather than a hard-coded path, which makes them runnable now.


apis/r/NAMESPACE
```diff
@@ -81,7 +81,6 @@ importFrom(arrow,RecordBatch)
 importFrom(arrow,concat_arrays)
 importFrom(bit64,as.integer64)
 importFrom(bit64,lim.integer64)
-importFrom(data.table,data.table)
```
eddelbuettel, line 84: This only removes an import.  Below you define a new one.  It is not yet here -- did you miss a roxygenize run?

apis/r/R/RcppExports.R
```diff
@@ -96,22 +94,16 @@ shape <- function(uri, config = NULL) {
...
-#'     sr |>
-#'         sr_next() |>
-#'         as_arrow_table() |>
-#'         collect() |>
-#'         as.data.frame() |>
-#'         data.table() -> D
-#'     rl <- rbind(rl, D)
+#'   dat <- sr_next(sr)
+#'   rb <- arrow::RecordBatch$import_from_c(dat$array_data, dat$schema)
+#'   rl <- rbind(rl, as.data.frame(rb))
```
eddelbuettel, line 104: We could also rbind with the as.data.frame.  I think we do that in another spot.  Works either way.

apis/r/R/utils-arrow.R
```diff
@@ -178,3 +178,10 @@ arrow_schema_from_tiledb_schema <- function(x) {
   )
   arrow::schema(fields)
 }
+
+#' Validate external pointer to ArrowArray
+#' @noRd
+check_arrow_pointers <- function(arrlst) {
+    stopifnot("First argument must be an external pointer to ArrowArray" = check_arrow_array_tag(arrlst[[1]]),
+              "Second argument must be an external pointer to ArrowSchema" = check_arrow_schema_tag(arrlst[[2]]))
+}
```
eddelbuettel, line 187: I am not the hugest fan of moving things from one 'utils' file to another.  I could see consolidating _all_ utils* file into one.   Different personal preferences I suppose.

apis/r/NAMESPACE
aaronwolen: I marked as draft b/c I'm still working through some rebase issues but I'll regenerate the docs once I'm sure I didn't leave anything behind.

apis/r/man/soma_array_reader.Rd
```diff
@@ -54,9 +54,7 @@ not iterate; if your data is large than the initial read size consider the \code
 functions.
 }
 \examples{
-\dontrun{
```
eddelbuettel, line 57: I think we are better off keeping these.  Our test coverage is good, this just slows things down and if we ever go to CRAN we have to inject throttles to 'Ncpu' cores in all examples...

eddelbuettel (commented):
I left a few comments in you may want to look at before I approve.

apis/r/R/ephemeral.R
```diff
@@ -3,6 +3,7 @@
 #' @description \Sexpr[results=rd]{tiledbsoma:::rd_ephemeral_cls("collection", TRUE)}
 #'
 #' @keywords internal
+#' @importFrom data.table address
```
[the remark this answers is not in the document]
aaronwolen: Still in draft mode, docs aren't updated yet.

apis/r/R/utils-arrow.R
aaronwolen: We already have category-specific utils files so I just moved it to the relevant location to make it easier to find.

apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
```diff
@@ -13,52 +12,56 @@ test_that("Iterated Interface from SOMAArrayReader", {
     uri <- file.path(tdir, "soco", "pbmc3k_processed", "ms", "RNA", "X", "data")
     expect_true(dir.exists(uri))
 
-    ctx <- tiledb_ctx()
-    sr <- sr_setup(uri, config=as.character(config(ctx)), loglevel="warn")
+    ctx <- tiledb::tiledb_ctx()
```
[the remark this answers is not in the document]
aaronwolen: Good call, will do.

apis/r/R/RcppExports.R
aaronwolen: I think it's already using rbind with as.data.frame:

```
rbind(rl, as.data.frame(rb))
```

did you mean _without_ as.data.frame? Or am I misunderstanding?

apis/r/man/soma_array_reader.Rd
aaronwolen: Ah okay, I'll revert. I just wanted to verify the examples worked.

johnkerl (approved):
LGTM, looks like all concerns have been addressed

apis/r/R/RcppExports.R
eddelbuettel: Outside of this example we don't really deal with the pair of pointers any more.  Maybe

```r
at <- soma_array_to_arrow_table( sr_next( sr ) )
```

is better here too?

PR #1468, opened 2023-06-09
[python] Skip ingest of zero-cell uns arrays

### LINKED ISSUE #1466: [bug/python] Zero-length `uns` array in `.h5ad` raises an error
In a particular `.h5ad` input from a customer, with a zero-length array in an `uns` slot, the ingest fails with

```
SOMADenseNDArray shape must be a non-zero-length tuple of positive ints
```

johnkerl:
**Issue and/or context:** Resolves #1466

**Changes:** Don't ingest uningestibles

**Notes for Reviewer:** @aaronwolen @mojaveazure looking for compbio feedback on this approach


aaronwolen (approved):
Thank you!

PR #1474, opened 2023-06-12
[python] Simplify internal ingestion API via `IngestionParams`

johnkerl:
**Issue and/or context:** Currently, the user-level `ingest_mode` string in `tiledbsoma.io` is re-used for internal decision-making, via things like `if ingest_mode == "resume": ...`. This is already a little confusing, but it will be worse with append-mode (#1278).

**Changes:** On this PR we simplify by separating the single, easily keyable user-level mode string from a set of orthogonal, directly iffable boolean flags.

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -66,6 +66,35 @@ Matrix = Union[DenseMatrix, SparseMatrix]
 _NDArr = TypeVar("_NDArr", bound=NDArray)
 _TDBO = TypeVar("_TDBO", bound=TileDBObject[RawHandle])
 
+# ----------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


johnkerl:
@atolopko-czi @aaronwolen @mojaveazure ping 🙏

atolopko-czi (approved):
Looks reasonable to me!

Can `IngestMode` type be removed now? 

I might have even kept the name as `IngestMode` (or `IngestionMode`) in place of the new name, `IngestionParams`, since the caller that is _using_ this class care more about the "Mode", and doesn't care about the actual settings (params).

I suppose you could also introduce constants for each of the `IngestionParam` values, which would it allow to be used as `IngestionParams.WRITE`, e.g., or `IngestionMode.WRITE`.

PR #1476, opened 2023-06-14
[c++] `SOMAArray::write` Modified To Take in `ArrayBuffers`

nguyenv:
**Issue and/or context:**

Separated out from PR https://github.com/single-cell-data/TileDB-SOMA/pull/1449.

**Changes:**

Previously, `SOMAArray` write buffers were set with `set_column_buffer`. The `write` function now takes in an `ArrayBuffers` data structure directly.

**Notes for Reviewer:**

N/A

### DIFF
--- a/libtiledbsoma/src/soma/managed_query.h
+++ b/libtiledbsoma/src/soma/managed_query.h
@@ -169,11 +169,15 @@ class ManagedQuery {
     /**
      * @brief Set column data for write query.
      *
-     * @param column_buffer Column data
+     * @param column_name Column name
+     * @param buff Buffer array pointer with elements of the column type.
+     * @param nelements Number of array elements in buffer
      */
-    template <typename T>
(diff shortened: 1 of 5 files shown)


eddelbuettel (approved):
LGTM -- while you're at it maybe increment the copyright years to 2023 too.

libtiledbsoma/src/soma/managed_query.h
```diff
@@ -5,7 +5,7 @@
  *
  * The MIT License
  *
- * @copyright Copyright (c) 2022 TileDB, Inc.
+ * @copyright Copyright (c) 2023 TileDB, Inc.
```
eddelbuettel, line 8: I don't mean to a pedant but if 2022 was already present I think we usuallt do 2022-2023.  Not a replacement.

eddelbuettel (changes requested):

eddelbuettel (approved):

PR #1478, opened 2023-06-15
[r] Add `to_sparse_matrix()` method to `SOMAExperimentAxisQuery`

aaronwolen:
**Changes:** This PR adds a new method to `SOMAExperimentAxisQuery` that allows users to load any layer from the queried experiment as a sparse matrix with _named_ dimensions (i.e., row and column names are populated).

As noted in the docs, dimension names default to joinids but may come from any obs or var column (depending on the collection type) containing values that uniquely identify each record.

Quick example:

```r
library(tiledbsoma)
experiment <- load_dataset("soma-exp-pbmc-small")

query <- SOMAExperimentAxisQuery$new(
  experiment = experiment,
  measurement_name = "RNA",
  obs_query = SOMAAxisQuery$new(coords = 0:9),
  var_query = SOMAAxisQuery$new(coords = 0:9)
)

query$to_sparse_matrix(
  collection = "X",
  layer = "counts",
  obs_index = "obs_id",
  var_index = "var_id"
)

## 10 x 10 sparse Matrix of class "dgTMatrix"
##   [[ suppressing 10 column names ‘MS4A1’, ‘CD79B’, ‘CD79A’ ... ]]
##
## ATGCCAGAACGACT . 1 . . . 1 . .  3 .
## CATGGCCTGTGCAT . . . 1 . . . .  7 .
## GAACCTGATGAACC . . . . . . . . 11 .
## TGACTGGATTCTCA . . . . . . . . 13 .
## AGTCAGACTGCACA . . . 1 . . . .  3 .
## TCTGATACACGTGT . . . 1 . . . .  4 .
## TGGTATCTAAACAG . . . . . . . .  6 .
## GCAGCTCTGTTTCT . . . 1 . . . .  4 .
## GATATAACACGCAT . . . . . . . .  2 .
## AATGTTGACAGTCA . 1 . . . . . . 21 .
```

More usage examples are available in the new unit tests and updated soma-experiment-queries vignette.

[sc-30317]

**Notes for Reviewer:** This was a feature request from a user who wanted an easier way to create named matrices without having to go through Seurat. IMO this was a great suggestion and I think it will be useful for many users.

The current implementation does not support specifying a sparse encoding via the typical `repr` argument because converting from a dgT (coo) matrix to dgC (csc) can be expensive and I don't want our method to silently bear that cost when a) the operation can be easily performed by the user after the fact and b) a fast c++ implementation of this operation is on the horizon.

### DIFF
--- a/apis/r/R/SOMAExperimentAxisQuery.R
+++ b/apis/r/R/SOMAExperimentAxisQuery.R
@@ -197,6 +197,7 @@ SOMAExperimentAxisQuery <- R6::R6Class(
 
         "Must specify a single layer name" =
           is.null(layer_name) || is_scalar_character(layer_name),
+        assert_subset(layer_name, self$ms[[collection]]$names(), "layer"),
 
         "Must specify a single obs index" =
           is.null(obs_index) || is_scalar_character(obs_index),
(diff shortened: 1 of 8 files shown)


eddelbuettel (approved):
:+1:

ckrilow:
This is great, fantastic work everyone.

mojaveazure (approved):

apis/r/tests/testthat/test-SOMAExperiment-query-matrix-outgest.R
```diff
@@ -0,0 +1,209 @@
+test_that("matrix outgest with all results", {
+  pbmc_small <- get_data("pbmc_small", package = "SeuratObject")
```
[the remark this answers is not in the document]
mojaveazure: It's true that it's a rare occurance, but I've had it happen. It's more important with Bioc packages as there are times when CRAN rebuilds their Bioc collection at the time they're running a check for your package, which'll cause failures, but it's rare. I've just gotten in the habit of checking for suggested packages rather than assuming they're there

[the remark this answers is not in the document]
eddelbuettel: You are supposed to do that.

CRAN was historically a little lax on checking it, but the language in _Writing R Extensions_ has gotten tighter.  It is still not fully enforced.  (IIRC it's an opt-in check when running `R CMD check`.)

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -169,6 +169,161 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+          soma_dim_0 = self$obs_joinids(),
+          soma_dim_1 = self$obs_joinids()
+        ),
+        varp = list(
+          soma_dim_0 = self$var_joinids(),
+          soma_dim_1 = self$var_joinids()
+        )
+      )
+
+      # TODO: Coords must be vectors until read() supports arrow arrays
```
pablo-gar, line 269: Small comment: this one is a constant pain point throughout many use cases, we should make it a high priority for post 1.0. In this case support for arrow arrays would be great, but ultimately a lightweight representation of ranges would be ideal

pablo-gar (approved):
Very clean and great addition!

apis/r/R/SOMAExperimentAxisQuery.R
johnkerl: @pablo-gar please file a GitHub issue to track -- thanks!

aaronwolen: https://github.com/single-cell-data/TileDB-SOMA/issues/1479

Issue #1489, opened 2023-06-21
[ci/python] Run full CI on merges to `release-i.j` branches


johnkerl:
## Problem 1

Today we had a CI fail (on MacOS and Python 3.7 only) once 1.2.7 was tagged:
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/5336828062/jobs/9672650908

This may be because GithHub just updated the runner images yesterday: https://github.com/actions/runner-images/pull/7752

Regardless, this needs to get caught pre-tag not post-tag.

We run "full" CI on merges to `main`, and on releases (post-tag):
https://github.com/single-cell-data/TileDB-SOMA/blob/main/.github/workflows/python-ci-full.yml#L6-L10

That was mostly a good decision -- "full CI" is distinct from "minimal CI" in that the latter runs on every PR and we were running into resource-starvation issues which were keeping people from doing dev work: https://github.com/single-cell-data/TileDB-SOMA/issues/693

We need to modify `python-ci-full.yml` to run not only on merges to `main` and on releases (post-tag) but also on merges to any `release-i.j` branch.

## Problem 2

Issue here: https://github.com/actions/setup-python/issues/682

affects us here:
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/5336828062/jobs/9672070106#step:8:40

That was today. I went back and re-ran a successful CI run from last week, and on that re-run today, it failed as above:
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/5269863811/jobs/9673256042

Per @ihnorton : we can use https://github.com/PySport/kloppy/commit/16079581a6a03b46fada13324d52b8da9f8550b1 to get our `.github/worksflows` YAML using non-broken runneers

johnkerl:
@mlin @thetorpedodog I tried this out on
https://github.com/single-cell-data/TileDB-SOMA/pull/1492
which was a PR merged to `release-1.2` but at
https://github.com/single-cell-data/TileDB-SOMA/actions/workflows/python-ci-full.yml
I did _not_ get a full CI run when #1492 was merged to `release-1.2`

Do you have any advice?

thetorpedodog:
> I notice that for core they use no quotes on `release-*`: https://github.com/TileDB-Inc/TileDB/blob/2.15.4/.github/workflows/full-ci.yml#L6

[...]

PR #1490, opened 2023-06-21
[ci/python] Run full CI on merges to `release-i.j` branches

johnkerl:
As in issue #1489

### DIFF
--- a/.github/workflows/python-ci-full.yml
+++ b/.github/workflows/python-ci-full.yml
@@ -5,7 +5,9 @@ name: TileDB-SOMA Python CI (Full)
 # releases.
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+      - 'release-**'
   release:
     types: [published]
   # You can also invoke this workflow manually from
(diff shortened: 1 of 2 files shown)


.github/workflows/python-ci-full.yml
```diff
@@ -5,7 +5,9 @@ name: TileDB-SOMA Python CI (Full)
 # releases.
 on:
   push:
-    branches: [main]
+    branches:
+      - main
+      - 'release-**'
```
thetorpedodog, line 10: Is two stars here intentional?

thetorpedodog (approved):

.github/workflows/python-ci-full.yml
johnkerl: @thetorpedodog just googlecopypasta ...

johnkerl: @thetorpedodog the thing I googlecopypastaed from may have been wanting to follow branches that have slashes in them -- I think I saw `release/**` -- since we don't use slashes in our release-branch names, I think a single star would suffice

PR #1491, opened 2023-06-21
[python] Trivial commit to test #1490 (1 of 2)

johnkerl:
If we did #1490 right then full CI will run when this PR is merged to `main` (1 of 2). See also #1492.


johnkerl:
No need; any subsequent PR to `main` will test this out.

PR #1496, opened 2023-06-21
[ci/python] Second try for #1490

johnkerl:
Attempts to address this problem: https://github.com/single-cell-data/TileDB-SOMA/issues/1489#issuecomment-1601824893


johnkerl:
Fixed by #1497 instead

PR #1498, opened 2023-06-22
[WIP] [python] Append-mode sketching

johnkerl:
This is a draft PR, with no reviewers, explicitly marked as not ready for review.

Please do not review this PR.

It is posted here solely for bus-factor risk-reduction, temporarily cloud-hosting some incomplete prototyping work in support of an upcoming second round of design discussion.

Context: #1278 

Split-out PRs from this raw material:

* #1554 
* #1556 
* A TBD third PR will do the actual ingest part

PR #1500, opened 2023-06-22
[python] Minor function-extract in ingestion logic

johnkerl:
Minor opportunity found on PR #1498 for issue #1278

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -641,27 +641,11 @@ def _create_or_open_coll(
     return cls.create(uri, context=context)
 
 
-def _write_dataframe(
-    df_uri: str,
-    df: pd.DataFrame,
-    id_column_name: Optional[str],
-    *,
-    ingestion_params: IngestionParams,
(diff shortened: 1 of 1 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -682,6 +666,30 @@ def _write_dataframe(
...
+    id_column_name: Optional[str],
+    *,
+    ingestion_params: IngestionParams,
+    platform_config: Optional[PlatformConfig] = None,
+    context: Optional[SOMATileDBContext] = None,
+) -> DataFrame:
+    s = _util.get_start_stamp()
+    logging.log_io(None, f"START  WRITING {df_uri}")
+
+    df[SOMA_JOINID] = np.asarray(range(len(df)), dtype=np.int64)
```
thetorpedodog, line 684: This can be `np.arange(len(df), dtype=np.int64)`.

thetorpedodog (approved):
one optional suggestion for an improvement along the way. you can take it, or leave this change to be purely a function extraction; either is good

PR #1501, opened 2023-06-22
[r] Adjust `SOMAExperimentAxisQuery$to_sparse_matrix()` to handle gapped matrices

mojaveazure:
Change `SOMAExperimentAxisQuery$to_sparse_matrix()` to utilize soma join IDs for dims instead of labels. This better handles gaps (all implicitly stored) in the arrays, which can happen by storage or by query

For obsm/varm, utilize span of `soma_dim_1` for columns (`dim()[2L]`) as there is no join ID for this axis

### DIFF
--- a/apis/r/R/SOMAExperimentAxisQuery.R
+++ b/apis/r/R/SOMAExperimentAxisQuery.R
@@ -319,30 +319,12 @@ SOMAExperimentAxisQuery <- R6::R6Class(
       # Use joinids if the dimension names are empty
       dim_names <- Map("%||%", dim_names, coords)
 
-      dims <- switch(
-        EXPR = collection,
-        X = list(
-          self$indexer$by_obs(coords$soma_dim_0),
-          self$indexer$by_var(coords$soma_dim_1)
-        ),
-        obsm = list(
-          self$indexer$by_obs(coords$soma_dim_0),
-          mat_coords$j
-        ),
-        varm = list(
-          self$indexer$by_var(coords$soma_dim_0),
-          mat_coords$j
-        ),
-        obsp = lapply(coords, self$indexer$by_obs),
-        varp = lapply(coords, self$indexer$by_var)
-      )
-
       Matrix::sparseMatrix(
         i = mat_coords$i$as_vector(),
         j = mat_coords$j$as_vector(),
         x = tbl$soma_data$as_vector(),
         index1 = FALSE,
-        dims = vapply_int(dims, function(x) max(range(x$as_vector()))) + 1L,
+        dims = vapply_int(dim_names, length),
         dimnames = dim_names,
         repr = "T"
       )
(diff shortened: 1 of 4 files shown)


johnkerl (commented):
LGTM but I defer to @aaronwolen and @pablo-gar for final app-R-oval :)

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -304,21 +304,45 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
+          self$indexer$by_obs(coords$soma_dim_0),
+          mat_coords$j
+        ),
+        varm = list(
+          self$indexer$by_var(coords$soma_dim_0),
+          mat_coords$j
+        ),
+        obsp = lapply(coords, self$indexer$by_obs),
+        varp = lapply(coords, self$indexer$by_var)
+      )
```
aaronwolen, line 338: I don't understand why we need to add this extra switch statement or change the way `dims` are calculated below. The tests all still passed after removing lines 322–338 and restoring `vapply_int(dim_names, length)`.  Maybe I'm missing something?

apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -304,21 +304,45 @@ SOMAExperimentAxisQuery <- R6::R6Class(
...
         X = list(obs_labels, var_labels),
-        obsm = list(obs_labels, unique(tbl$soma_dim_1)$as_vector()),
-        varm = list(var_labels, unique(tbl$soma_dim_1)$as_vector()),
+        obsm = {
+          soma_dim_1 <- range(tbl$soma_dim_1$as_vector())
+          list(obs_labels, seq(min(soma_dim_1), max(soma_dim_1)))
+        },
+        varm = {
+          soma_dim_1 <- range(tbl$soma_dim_1$as_vector())
+          list(var_labels, seq(min(soma_dim_1), max(soma_dim_1)))
```
aaronwolen, line 313: So the concrete use case this accommodates is something like a matrix of PCs with one or more empty columns?

aaronwolen (changes requested):
Posted a couple of questions to discuss.

apis/r/R/SOMAExperimentAxisQuery.R
mojaveazure: Whoops, my bad. You're right, this is unnecessary and I've removed it

apis/r/R/SOMAExperimentAxisQuery.R
mojaveazure: Yes, or more concretely, a query where the selection of cells all have a PC/UMAP/embedding value of zero

aaronwolen (approved):
Thanks for making those changes, @mojaveazure. My last ask before merging is to update the method docs to explain this behavior and the motivation for it.

Issue #1487, opened 2023-06-21
[r] Improve handling of arrays with with domains greater than 2^31 - 1


aaronwolen:

johnkerl:
For reference: experimental branch https://github.com/single-cell-data/TileDB-SOMA/tree/aaronwolen/improve-handling-of-64bit-arrays

Issue #1445, opened 2023-06-05
[r/python] Parent/tracking issue for shape


johnkerl:
## Factors

**There are two possible meanings of shape:**

* Bounding box for the data that are currently in a sparse array
* Capacity or domain, i.e. the bounds outside of which no data are allowed to be written

There exist software systems where "shape" takes the former meaning, and software systems where "shape" takes the latter meaning. For SOMA, shape takes the latter meaning.

People need to be able to access both of these values for all the arrays within a SOMA experiment. For systems in which "shape" takes the former meaning, accessors might be called "shape" and "maxshape", respectively. For SOMA, where "shape" takes the latter meaning, accessors are non-empty domain and "shape", respectively.

For the TileDB implementation of SOMA, note that domains are fixed at array-creation time; an array cannot have its domain enlarged post-create.

**Non-append and append modes:**

There is an important single-cell use-case in which users have some amount of data in a SOMA experiment, then add or append more data later -- for example, appending the results of the next day's laboratory run. Users don't know, and can't know, the final dimensions of their data at array-creation time; there must be room for growth.

There is another use-case, the "immutable" mode, in which some number of inputs are used to create a timestamped SOMA experiment. When later data arrive, and entirely new experiment is written. This design pattern is used, for example, by the CZ CELLxGENE Discover Census.

**R 31-bit issues:**

There are container classes in which the shape-as-capacity of a sparse array -- even one with a very small number of occupied datapoints -- must not exceed 2**31-1. This is an important interoperability issue for the room-for-growth use-case: experiments written by Python, with room-for-growth shape, should still be readable by R software without signed-int32-overflow issues.

Even when array shapes do not exceed 2**31-1, R software can abort with cholmod errors trying to read a sparse array (even with the number of occupied cells is quite small).

## Things to be done

* [x] Documentation for the R and Python packages need to be clear, for each foundational class, that "shape" means _the domain outside of which no data can be written_: #1446.
* [ ] The sparse n-d array class, in R and Python both, needs support for non-empty-domain read access.
  * We will use TileDB array metadata to track this.
  * @mojaveazure has an open design document on this
[...]

PR #1504, opened 2023-06-23
[r] Improve handling of arrays with with domains greater than 2^31 - 1

aaronwolen:
**Issue and/or context:** #1487

**Changes:** This makes a few adjustment to improve our handling of arrays with domains greater than 2^31-1.

- `SOMASparseNDArray$read()` no longer warns if `nnz() > .Machine$integer.max`, since these arrays can be read without issue using `SOMASparseNDArray$read()$tables()`
- `SOMASparseNDArrayRead$sparse_matrix()` truncates `shape` to `2^31 - 1` if `shape > 2^31-1` and lets the user know via a warning. I decided to add this logic here rather than deeper in the stack so that it executes only once when reading sparse matrices, rather than once per iteration.
- Fixed `SparseReaditer`'s `shape` assertion
- `arrow_table_to_sparse` adds a final check to ensure that `shape` is `<= 2^31 - 1` and all coordinates are within `[0, 2^31 - 1)` (since `shape` is 1-based and `coords` are 0-based)
- `arrow_table_to_sparse` was also slightly simplified to work directly with 0-based coords and leverage `Matrix::sparseMatrix()`'s `index1` argument
- `SOMAExperimentAxisQuery$to_seurat_graph()` was refactored to leverage `$to_sparse_matrix()` and avoid the costly creation of a `dgCMatrix` with domains `.Machine$integer.max`

### DIFF
--- a/apis/r/R/SOMASparseNDArrayRead.R
+++ b/apis/r/R/SOMASparseNDArrayRead.R
@@ -24,7 +24,19 @@ SOMASparseNDArrayRead <- R6::R6Class(
     #' @return \link{SparseReadIter}
     sparse_matrix = function(zero_based=FALSE) {
       #TODO implement zero_based argument, currently doesn't do anything
-      SparseReadIter$new(sr = private$sr, shape = private$shape, zero_based=zero_based)
+
+      shape <- private$shape
+      if (any(shape >= .Machine$integer.max)) {
+        warning(
+          "Array domain exceeds '.Machine$integer.max'.\n",
+          "  - Result will only include coordinates within [0, 2^31].\n",
+          "  - The full range of coordinates can be obtained with $tables().",
+          call. = FALSE
+        )
+        shape <- pmin(shape, .Machine$integer.max)
+      }
+
+      SparseReadIter$new(private$sr, shape = shape, zero_based = zero_based)
     },
 
     #' @description Read as a arrow::\link[arrow]{Table} (lifecycle: experimental).
--- a/apis/r/R/SOMASparseNDArrayRead.R
+++ b/apis/r/R/SOMASparseNDArrayRead.R
@@ -25,10 +25,11 @@ SOMASparseNDArrayRead <- R6::R6Class(
     sparse_matrix = function(zero_based=FALSE) {
       #TODO implement zero_based argument, currently doesn't do anything
 
+
       if (any(private$shape >= .Machine$integer.max)) {
         warning(
-          "Array domain exceeds '.Machine$integer.max'.\n",
-          "  - Result will only include coordinates within [0, 2^31].\n",
+          "Array's 0-based domain exceeds '.Machine$integer.max'.\n",
+          "  - Result will only include coordinates within [0, 2^31 - 1).\n",
           "  - The full range of coordinates can be obtained with $tables().",
           call. = FALSE,
           immediate. = TRUE
--- a/apis/r/R/SOMASparseNDArrayRead.R
+++ b/apis/r/R/SOMASparseNDArrayRead.R
@@ -25,18 +25,17 @@ SOMASparseNDArrayRead <- R6::R6Class(
     sparse_matrix = function(zero_based=FALSE) {
       #TODO implement zero_based argument, currently doesn't do anything
 
-      shape <- private$shape
-      if (any(shape >= .Machine$integer.max)) {
+      if (any(private$shape >= .Machine$integer.max)) {
         warning(
           "Array domain exceeds '.Machine$integer.max'.\n",
           "  - Result will only include coordinates within [0, 2^31].\n",
           "  - The full range of coordinates can be obtained with $tables().",
           call. = FALSE
         )
-        shape <- pmin(shape, .Machine$integer.max)
+        private$shape <- pmin(private$shape, .Machine$integer.max - 1L)
       }
 
-      SparseReadIter$new(private$sr, shape = shape, zero_based = zero_based)
+      SparseReadIter$new(private$sr, private$shape, zero_based = zero_based)
     },
 
     #' @description Read as a arrow::\link[arrow]{Table} (lifecycle: experimental).
--- a/apis/r/R/SOMASparseNDArrayRead.R
+++ b/apis/r/R/SOMASparseNDArrayRead.R
@@ -30,7 +30,8 @@ SOMASparseNDArrayRead <- R6::R6Class(
           "Array domain exceeds '.Machine$integer.max'.\n",
           "  - Result will only include coordinates within [0, 2^31].\n",
           "  - The full range of coordinates can be obtained with $tables().",
-          call. = FALSE
+          call. = FALSE,
+          immediate. = TRUE
         )
         private$shape <- pmin(private$shape, .Machine$integer.max - 1L)
       }
--- a/apis/r/R/SOMASparseNDArrayRead.R
+++ b/apis/r/R/SOMASparseNDArrayRead.R
@@ -26,15 +26,15 @@ SOMASparseNDArrayRead <- R6::R6Class(
       #TODO implement zero_based argument, currently doesn't do anything
 
 
-      if (any(private$shape >= .Machine$integer.max)) {
+      if (any(private$shape > .Machine$integer.max)) {
         warning(
-          "Array's 0-based domain exceeds '.Machine$integer.max'.\n",
+          "Array's shape exceeds '.Machine$integer.max'.\n",
           "  - Result will only include coordinates within [0, 2^31 - 1).\n",
           "  - The full range of coordinates can be obtained with $tables().",
           call. = FALSE,
           immediate. = TRUE
         )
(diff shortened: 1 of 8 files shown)


johnkerl (approved):
🚢

apis/r/man/ConfigList.Rd
```diff
@@ -10,7 +10,7 @@ Essentially, serves as a nested map where the inner map is a
 \code{\{<param>: \link[tiledbsoma:ScalarMap]{\{<key>: <value>\}}\}}
 }
 \section{Super class}{
-\code{\link[tiledbsoma:MappingBase]{tiledbsoma::MappingBase}} -> \code{ConfigList}
+\code{tiledbsoma::MappingBase} -> \code{ConfigList}
```
eddelbuettel, line 13: I am pretty sure we playing commit ping-pong here:

```
43d09be3c (Dirk Eddelbuettel 2023-06-21 21:22:59 -0500  13) \code{\link[tiledbsoma:MappingBase]{tiledbsoma::MappingBase}} -> \code{ConfigList}
```

I am using roxygen2 7.2.3.  What else could be different?  I am not setting an environment variables or magic.

apis/r/tests/testthat/helper-test-soma-objects.R
```diff
@@ -137,3 +137,51 @@ create_and_populate_experiment <- function(
...
+# values at `(0,0)`, `(2^31 - 2, 2^31 - 2)` and `(2^31 - 1, 2^31 - 1)`. This is
+# intended to test R's ability to read from arrays created with tiledbsoma-py
+# before the default domain was changed to `[0, 2^31)`.
+#
+#  Row/Column:   0      ...    2147483646 |    2147483647
+#  0             1      ...    0          |    0
+#  ...           ...    ...    ...        |    ...
+#  2147483646    0      ...    2          |    0
+#  ---------------------------------------|---------------
+#  2147483647    0      ...    0          |    3
```
eddelbuettel, line 151: Very nice

eddelbuettel (commented):
One question on Rd files -- something seems different between our machines. Ideas?

apis/r/R/utils-readerTransformers.R
```diff
@@ -33,25 +33,33 @@ arrow_table_to_sparse <- function(tbl, repr = c("C", "T", "R"), shape = NULL, ze
...
-  if(any(shape > .Machine$integer.max)) {
-      stop("The shape of the array is larger than supported by Matrix::sparseMatrix",
-           call. = FALSE)
+  if (any(shape > .Machine$integer.max)) {
+    stop("'shape' must not exceed '.Machine$integer.max'.", call. = FALSE)
+  }
+
+  exceedsInt32Limit <- (
+    tbl$soma_dim_0 >= .Machine$integer.max | tbl$soma_dim_1 >= .Machine$integer.max
+  )
```
mojaveazure, line 46: One thing this PR changes is the removal of `nnz <= .Machine$integer.max`, which is important as sparse matrices cannot have more than 2^31-1 non-zero values. Can we add a check here for that (`length(tbl$soma_data) > .Machine$integer.max` I think)

apis/r/man/ConfigList.Rd
mojaveazure: Roxygen2 results in different behavior depending on how it's run (eg. RStudio/R console vs `Rscript -e`). I find running it with `Rscript -e` works best, especially with regards to auto-linking docs

eddelbuettel: Same, albeit via `r`.  I always use my `roxy.r` wrapper, with option `-f` for full:

```r
## loop over all argument, with fallback of the current directory, and
## call roxygenize() on the given directory with roclets="rd" set
sapply(ifelse(length(argv) > 0, argv, "."),
       FUN = roxygenize,
       roclets = if (tolower(opt$roclets) == "null") NULL else strsplit(opt$roclets,",")[[1]])
```

Do we know / can we know what RStudio, or VSCode, or ... do differently?  Can we settle on a harmonizing step that unifies this?   I have not strong view in the matter and mostly deploy `roxygen2` so that it stays out of my hair (i.e. my personal default use it version 6.0.1 with just `namespace,rd` as roclets as strictly no fancy pants).

apis/r/R/utils-readerTransformers.R
aaronwolen: Good catch. I removed the assertion from `SOMASparseNDArray$read()` because it was blocking `SOMASparseNDArray$read()$tables()`, which is perfectly acceptable.

apis/r/man/ConfigList.Rd
aaronwolen: I reverted since my changes don't require any edits to the docs.

eddelbuettel (approved):
I would need a wee bit more time for a full review but this looks ready to sail. :shipit:

mojaveazure (approved):
LGTM

apis/r/R/SOMASparseNDArrayRead.R
```diff
@@ -24,7 +24,20 @@ SOMASparseNDArrayRead <- R6::R6Class(
     #' @return \link{SparseReadIter}
     sparse_matrix = function(zero_based=FALSE) {
       #TODO implement zero_based argument, currently doesn't do anything
-      SparseReadIter$new(sr = private$sr, shape = private$shape, zero_based=zero_based)
+
+
+      if (any(private$shape > .Machine$integer.max)) {
+        warning(
```
pablo-gar, line 30: Should we just error out here?  A warning and modifying the expected output is error-prone if the user is not paying close attention

apis/r/R/utils-readerTransformers.R
```diff
@@ -33,25 +33,36 @@ arrow_table_to_sparse <- function(tbl, repr = c("C", "T", "R"), shape = NULL, ze
...
-                              j = soma_dim_1_one_based,
-                              x = soma_data,
-                              dims = shape, repr = repr)
-  if(zero_based) {
+  mat <- Matrix::sparseMatrix(i = tbl$soma_dim_0$as_vector(),
+                              j = tbl$soma_dim_1$as_vector(),
+                              x = tbl$soma_data$as_vector(),
+                              dims = shape,
+                              repr = repr,
+                              index1 = FALSE)
```
pablo-gar, line 64: nice! Mike and I missed this arg in the past

pablo-gar (approved):
Looks good to me with a big asterisk that truncating shapes and throwing a warning is prone to ghost errors for the user.

apis/r/R/utils-readerTransformers.R
eddelbuettel: I recall all of use discussing it at some point a few weeks back.  No panacea, but it shows the Matrix authors know about issues / a need for zero base.

apis/r/R/SOMASparseNDArrayRead.R
aaronwolen: I struggled with this question but if we error out here we preclude the creation of sparse matrices from arrays with domains >= 2^31-1, even when all non-zero elements are an within R-friendly range of `[0, 2^31-1)`. To me this seemed like a reasonable trade-off: we can still create sparse matrices from these arrays, but we warn the user the resulting matrix's shape will be truncated to `c(2^31-1, 2^31-1)`. In the unlikely scenario that the array (& query) contains non-zero elements at indices >= 2^31-1 (or more than than 2^32-1 values), then we error out and suggest they access the data as an arrow table.

What do you think?

johnkerl: Is there a performant way to find the max _used index_ and check if _that_ is > 2^31-1?

aaronwolen: Currently we check the coordinates of the arrow table before attempting to construct a sparse matrix. We could potentially use non-empty domain beforehand but we'll have an easier option in place after we starting the bounding box metadata.

johnkerl: @aaronwolen this is awesome!!  Please either create a tracking task, or, make a note on #1445 so we remember

pablo-gar: > Thanks, @pablo-gar. I struggled with this question but if we error out here we preclude the creation of sparse matrices from arrays with domains >= 2^31-1, even when all non-zero elements are an within R-friendly range of [0, 2^31-1). 

Fair. 

>  In the unlikely scenario that the array (& query) contains non-zero elements at indices >= 2^31-1 (or more than than 2^32-1 values), then we error out and suggest they access the data as an arrow table.

And great solution

> > is there a performant way in R to find the max used index and check if that is > 2^31-1?
> We could potentially use non-empty domain. 

[I tried this already](https://github.com/single-cell-data/TileDB-SOMA/pull/1274#discussion_r1192742716) and decided not to go with it and to make the shape representation a bona-fide representation of the sparse shape and for parity with Python. I agree with @aaronwolen that once the bounding-box design is finalized we can revisit (and revisit in Python as well)

Issue #1335, opened 2023-05-04
[ci] Clone/adapt backport infra from TileDB core, TileDB-VCF, etc.


johnkerl:
This will help automate cherry-picking of PRs from `main` into `i.j.k++` micro releases.

See also #1242 -- we will need more doc updates as well.

Mirrors Shortcut story 28545

To do before closing this issue:

* PR #1506 
* Test-drive this
* Document at https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases

johnkerl:
Latest fail
https://github.com/single-cell-data/TileDB-SOMA/pull/1510#issuecomment-1610106718
👀

johnkerl:
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/5393905508/jobs/9794302903#step:2:31

```
Your repository allows merge commits and rebase merging.
 However, Backport only supports rebased and merged pull requests with a single commit and squashed and merged pull requests.
 Consider only allowing squash merging.
 See https://help.github.com/en/github/administering-a-repository/about-merge-methods-on-github for more information.
Backporting 3abdf709f4e915d9e23e22142e9[6](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/5393905508/jobs/9794302903#step:2:7)86298cbc8548 from #1510
/usr/bin/git clone ***github.com/single-cell-data/TileDB-SOMA.git
Cloning into 'TileDB-SOMA'...
/usr/bin/git config --global user.email github-actions[bot]@users.noreply.github.com
/usr/bin/git config --global user.name github-actions[bot]
Backporting to release-1.2 on backport-1510-to-release-1.2
  /usr/bin/git switch release-1.2
  Switched to a new branch 'release-1.2'
  branch 'release-1.2' set up to track 'origin/release-1.2'.
  /usr/bin/git switch --create backport-1510-to-release-1.2
  Switched to a new branch 'backport-1510-to-release-1.2'
  /usr/bin/git cherry-pick 3abdf[7](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/5393905508/jobs/9794302903#step:2:8)09f4e915d9e23e22142e96[8](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/5393905508/jobs/9794302903#step:2:9)62[9](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/5393905508/jobs/9794302903#step:2:10)8cbc8548
[...]

PR #1506, opened 2023-06-26
[ci] Initial experiments with backport-bot

johnkerl:
**Issue and/or context:** #1335

**Changes:**

**Notes for Reviewer:**

Not sure what else is needed to complete this -- things I've found:

* https://github.com/TileDB-Inc/TileDB/pull/1669/files
* https://github.com/tibdex/backport/blob/main/README.md
* https://github.com/marketplace/actions/backporting

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,20 @@
+name: Backport
+on:
+  pull_request:
+    types:
+      - closed
+      - labeled
+    paths-ignore:
+      - '_quarto.yml'
+      - 'quarto-materials/*'
+      - '**/.md'
+
+jobs:
+  backport:
+    runs-on: ubuntu-18.04
+    name: Backport
+    steps:
+      - name: Backport
+        uses: tibdex/backport@v1
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/backport.yml
```diff
@@ -0,0 +1,20 @@
...
+      - closed
+      - labeled
+    paths-ignore:
+      - '_quarto.yml'
+      - 'quarto-materials/*'
+      - '**/.md'
+
+jobs:
+  backport:
+    runs-on: ubuntu-18.04
```
eddelbuettel, line 14: Consider 20.04 or 22.04?

The suggested-to-be-included https://github.com/tibdex/backport/blob/main/.github/workflows/backport.yml has ubuntu-latest.

.github/workflows/backport.yml
```diff
@@ -0,0 +1,20 @@
...
+    name: Backport
+    steps:
+      - name: Backport
+        uses: tibdex/backport@v1
```
eddelbuettel, line 18: Current version seems to be 2.0.3 per https://github.com/tibdex/backport.  But if I read this right (haven't spent a lot of time on it) they suggest to copy their yml file (see prev comment)

.github/workflows/backport.yml
johnkerl: Yes @eddelbuettel 100% copy from core

Will update ...

johnkerl: ... or perhaps leave as-is for compatibility with core

If this works out of the box, great. If it doesn't work out of the box (while core's does work) the first thing I'd want to do is remove all differerences between our config and core's ... so, perhaps best to leave them the same fro the get-go.

eddelbuettel: Can 100% be argued both ways because what Core has clearly works yet may need an update....  

I am completely in support for "headache avoidance" so carbon-copy has a lot of merit.  

As, at least to me, has my hint to not encode ubuntu-18.04 in the year 2023 because they will withdraw for it at some point.  So a classic 'lose-lose', eventually.   But a win in the short-term.

eddelbuettel (approved):
Approve with caveats as mention.  You could try replacing `ubuntu-18.04` with `ubuntu-latest`.  That should not affect the yaml magic.  Or so one hopes...

johnkerl:
Testing now with this very PR, and branch `release-1.2`

johnkerl:
Action https://github.com/single-cell-data/TileDB-SOMA/actions/runs/5390788414/jobs/9786667557 exists ...

eddelbuettel:
I think I ran into this 'endless queue' with a repo too.  See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners which lists no ubuntu-18.04 runners.   If it never picks up it may be worth changing to a runner they support.  (And yes I understand that Core's still work but who knows -- maybe they treat 'first runs' differently from 'grandfathered existing runs'.  Just a guess....)

Issue #1404, opened 2023-05-22
[python] Increase re-use fraction in `uns` ingest


johnkerl:
The goal of `ingest_mode="resume"` is, given explicit user contract that the same input file is being ingested (after, say, a network event), to not re-write parts of the experiment that had already succeeded, and only go one to do the as-yet-undone parts.

This has generally been implemented, however, for `uns` objects we are creating a redundant layer.

This is a minor optimization for ingestion time, storage space, and query time for `uns`.

PR #1509, opened 2023-06-26
[python] Don't re-write `uns` NDArrays when in resume mode

johnkerl:
**Issue and/or context:** Addresses #1404.

**Changes:** As seen on this PR, `uns` arrays boil down to metadata, `soma.DataFrame`, or `soma.DenseNDArray`. Writing the first is as quick as checking it, and doesn't result in fragment layers; writing the second goes through `_write_dataframe` which already has a resume-mode check; writing the third goes through `_ingest_uns_ndarray` which -- as of this PR -- does the same right thing that `_write_dataframe` already does.

**Notes for Reviewer:**

Testing:

* All unit tests
* Also:

```
devtools/ingestor --ingest-mode=resume ~/s/a/kaminski.h5ad kaminski
```

Run that twice and verify `uns` arrays are `Skipped`.

(Data available in `s3://tiledb-johnkerl/s/a/` for TileDB folks.)

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -1384,6 +1384,7 @@ def _ingest_uns_node(
             platform_config,
             context=context,
             use_relative_uri=use_relative_uri,
+            ingestion_params=ingestion_params,
         )
         return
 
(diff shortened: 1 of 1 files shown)


johnkerl:
@thetorpedodog @aaronwolen @mojaveazure ping 🙏

thetorpedodog (approved):

johnkerl:
Thanks @thetorpedodog !

PR #1510, opened 2023-06-27
[ci] Backport Ubuntu-image update

johnkerl:
**Issue and/or context:** #1335

**Changes:** See PR diff

**Notes for Reviewer:**

Following on @eddelbuettel 's comments on #1506 -- in my tiptoe into backport-bot for `TileDB-SOMA`, I thought to imitate core as closely as possible (since known-good), then tweak versions after. However we found that `ubuntu-18.04` actually resulted in launch failures so let's try `ubuntu-20.04`.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   backport:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     name: Backport
     steps:
       - name: Backport
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   backport:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     name: Backport
     steps:
       - name: Backport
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -11,7 +11,7 @@ on:
 
 jobs:
   backport:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
eddelbuettel, line 14: Upstream uses `ubuntu-latest`. 

Ubuntu 22.04 is current.

So why 20.04?

eddelbuettel (approved):
I still think ubuntu-latest is a saner choice but 22.04 is defensible if one thinks stability matters.  For something as small as an action, I think I learn towards automatically rolling.

.github/workflows/backport.yml
```diff
@@ -11,7 +11,7 @@ on:
 
 jobs:
   backport:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
```
eddelbuettel, line 14: Yay!

johnkerl:
Testing now with this very PR, and branch `release-1.2`

Issue #866, opened 2023-02-03
[tracking] Add support for enumerated types AKA categoricals AKA factors


johnkerl:
Many systems such as AnnData, Pandas, Arrow, and the R language itself support categoricals.

* Simple example: `red, yelllow, green`
* Users get to specify sort order -- if these were strings they'd be `green, red, yellow`
* Storage backends can map these to encoded integers with a lookup mapping if desired
* Similarly, these lend themselves well to dictionary encoding in backend storage

Status quo in TileDB-SOMA has been that these are "decategoricalized" or "flattened" to strings (or ints, etc.)

* Sort order becomes lexical/numeric
* Conversation from AnnData to SOMA back to AnnData is less faithful
* See also #106

Evaluation plan:

* Core 2.17 upcoming release will have support for this @davisp
  * https://github.com/TileDB-Inc/TileDB/pull/4051
* Pre-release evaluation in TileDB-Py @nguyenv 
  * https://github.com/TileDB-Inc/TileDB-Py/pull/1790
* Pre-release evaluation in TileDB-R @eddelbuettel 
  * https://github.com/TileDB-Inc/TileDB-R/pull/562
* Pre-release evaluation in TileDB-SOMA-Py @nguyenv 
  * https://github.com/single-cell-data/TileDB-SOMA/pull/1511 -- as of 2023-07-05 not ready for eval
* Pre-release evaluation in TileDB-SOMA-R @eddelbuettel 
  * https://github.com/single-cell-data/TileDB-SOMA/pull/1559 -- as of 2023-07-27 still WIP awaiting 2.17
  * @eddelbuettel has split out tracking issue #1558 for R
* Pre-release evaluation in CELLxGENE Discover Census Python @bkmartinjr 
* Pre-release evaluation in CELLxGENE Discover Census R @bkmartinjr

johnkerl:
Only remaining issue is #1710 which has its own tracking; closing this parent/tracker issue.

PR #1511, opened 2023-06-28
[python] Support `Enumeration` in Python API

nguyenv:
**Issue and/or context:** #866

**Changes:**

Reads are done in C++. Writes are done through TileDB-Py (https://github.com/TileDB-Inc/TileDB-Py/pull/1790).

**Notes for Reviewer:**

Depends on https://github.com/single-cell-data/TileDB-SOMA/pull/1519.


johnkerl:
[sc-30316]

nguyenv:
This PR is now at a point where it can be evaluated for feedback.

johnkerl:
@nguyenv those are tests I created -- I'll help take a look

johnkerl (approved):

johnkerl:
This has the same interop fail as #1519 which I am debugging. Otherwise looks good.

PR #1513, opened 2023-06-30
[r] Add bounding box to arrays

mojaveazure:
Add support for encoding and retrieving bounding box information from arrays. The bounding box stored in array-level metadata and created during write

Bounding box metadata is structured as `{<DIMENSION_NAME>_DOMAIN: [<lower_bound>, <upper_bound>]}` (for example, `{soma_dim_0_DOMAIN: [0, 79]}`). The bounds are 0-based, inclusive, and stored as 64-bit integers

New SOMA Methods
  - `TileDBArray$non_empty_domain()`: retrieves the maximal value of the non-empty domain via `tiledb::tiledb_array_get_non_empty_domain_from_name()`
    - parameter `index1`: return the one-based index for the non-empty domain; defaults to `FALSE`
  - `TileDBArray$used_shape()`: retrieves the bounding box for array-level meta data
    - parameter `simplify`: return only the maximal bounds of the used shape (essentially assumes the lower bound is zero); defaults to `FALSE`
    - parameter `index`: return the one-based index for the non-empty domain; defaults to `FALSE`
 
Modified SOMA Methods
  - `SOMASparseNDArray$write()`: now takes a parameter `bbox` for an optional bounding box. By default, is `NULL` and bounding box is taken from `values`; encodes bounding box in array meta data as 64-bit integers

Context: #1445

### DIFF
--- a/apis/r/R/SOMASparseNDArray.R
+++ b/apis/r/R/SOMASparseNDArray.R
@@ -165,7 +165,7 @@ SOMASparseNDArray <- R6::R6Class(
         }
         bbox[[x]] <- xrange
       }
-      names(bbox) <- paste0(names(bbox), '_DOMAIN')
+      names(bbox) <- paste0(names(bbox), '_domain')
       self$set_metadata(bbox)
       private$.write_coo_dataframe(coo)
     },
--- a/apis/r/R/SOMASparseNDArray.R
+++ b/apis/r/R/SOMASparseNDArray.R
@@ -63,9 +63,13 @@ SOMASparseNDArray <- R6::R6Class(
     #' [`TsparseMatrix`][`Matrix::TsparseMatrix-class`]. Character dimension
     #' names are ignored because `SOMANDArray`'s use integer indexing.
     #'
-    write = function(values) {
+    write = function(values, bbox = NULL) {
       stopifnot(
-        "'values' must be a matrix" = is_matrix(values)
+        "'values' must be a matrix" = is_matrix(values),
+        is.null(bbox) || length(bbox) == length(dim(values)),
+        is.null(bbox) ||
+          (is_integerish(bbox) || bit64::is.integer64(bbox)) ||
+          (is.list(bbox) && vapply_lgl(bbox, function(x, n) length(x) == n, n = length(dim(values))))
       )
       # coerce to a TsparseMatrix, which uses 0-based COO indexing
       values <- as(values, Class = "TsparseMatrix")
@@ -74,7 +78,76 @@ SOMASparseNDArray <- R6::R6Class(
         j = bit64::as.integer64(values@j),
         x = values@x
       )
-      colnames(coo) <- c(self$dimnames(), self$attrnames())
+      dnames <- self$dimnames()
+      colnames(coo) <- c(dnames, self$attrnames())
+      ranges <- sapply(
+        X = dnames,
+        FUN = function(x) {
+          return(range(coo[[x]]))
+        },
+        simplify = FALSE,
+        USE.NAMES = TRUE
+      )
+      bbox <- bbox %||% ranges
+      if (is.null(names(bbox))) {
+        names(bbox) <- dnames
+      }
+      if (!is_named(bbox, allow_empty = FALSE)) {
+        idx <- which(!nzchar(names(bbox)))
+        names(bbox)[idx] <- dnames[idx]
+      }
+      bbox <- bbox[dnames]
+      if (length(bbox) != length(dnames)) {
+        stop("dnames")
+      }
+      if (is_integerish(bbox) || bit64::is.integer64(bbox)) {
+        bbox <- sapply(
+          X = names(bbox),
+          FUN = function(x) {
+            return(sort(c(min(ranges[[x]]), bbox[[x]])))
+          },
+          simplify = FALSE
+        )
+      }
+      for (x in dnames) {
+        xrange <- bbox[[x]]
+        if (any(is.na(xrange))) {
+          stop(
+            "Ranges in the bounding box must be finite (offending: ",
+            sQuote(x),
+            ")",
+            call. = FALSE
+          )
+        }
+        if (!(is_integerish(xrange) || bit64::is.integer64(xrange))) {
+          stop(
+            "Ranges in the bounding box must be integers",
+            call. = FALSE
+          )
+        }
+        xrange <- sort(bit64::as.integer64(xrange))
+        if (length(xrange) != 2L) {
+          stop(
+            "Ranges in the bounding box must be two 64-bit integers",
+            call. = FALSE
+          )
+        }
+        if (xrange[1L] < 0 || xrange[1L] > min(ranges[[x]])) {
+          stop(
+            "Ranges in the bounding box must be positive and less than the lowest value being added",
+            call. = FALSE
+          )
+        }
+        if (xrange[2L] < min(ranges[[x]])) {
+          stop(
+            "Ranges in the bounding box must be greater than the largest value being added",
+            call. = FALSE
+          )
+        }
+        bbox[[x]] <- xrange
+      }
+      names(bbox) <- paste0(names(bbox), '_DOMAIN')
+      self$set_metadata(bbox)
       private$.write_coo_dataframe(coo)
     },
 
--- a/apis/r/R/SOMASparseNDArray.R
+++ b/apis/r/R/SOMASparseNDArray.R
@@ -95,6 +95,10 @@ SOMASparseNDArray <- R6::R6Class(
         names(bbox) <- dnames
       }
       if (!is_named(bbox, allow_empty = FALSE)) {
+        # Determine which indexes of `bbox` are unnamed (incl empty strings "")
+        # Python equivalent:
+        # bbox = pandas.Series([[0, 99], [0, 299]], index=["soma_dim_0", ""])
+        # [i for i, key in enumerate(bbox.keys()) if not len(key)]
         idx <- which(!nzchar(names(bbox)))
         names(bbox)[idx] <- dnames[idx]
       }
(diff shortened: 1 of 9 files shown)


apis/r/R/SOMASparseNDArray.R
```diff
@@ -63,9 +63,13 @@ SOMASparseNDArray <- R6::R6Class(
...
     #'
-    write = function(values) {
+    write = function(values, bbox = NULL) {
       stopifnot(
-        "'values' must be a matrix" = is_matrix(values)
+        "'values' must be a matrix" = is_matrix(values),
+        is.null(bbox) || length(bbox) == length(dim(values)),
+        is.null(bbox) ||
+          (is_integerish(bbox) || bit64::is.integer64(bbox)) ||
+          (is.list(bbox) && vapply_lgl(bbox, function(x, n) length(x) == n, n = length(dim(values))))
```
johnkerl, line 74: split this up into multiple `stopifnot`. the "must be a matrix" is just one error message covering a multitude of possible errors.

i think we need three stopifnots:

* is a matrix
* null or lengths match
* integer types; dimensions

apis/r/R/SOMASparseNDArray.R
```diff
@@ -74,7 +78,76 @@ SOMASparseNDArray <- R6::R6Class(
...
+        },
+        simplify = FALSE,
+        USE.NAMES = TRUE
+      )
+      bbox <- bbox %||% ranges
+      if (is.null(names(bbox))) {
+        names(bbox) <- dnames
+      }
+      if (!is_named(bbox, allow_empty = FALSE)) {
+        idx <- which(!nzchar(names(bbox)))
```
johnkerl, line 98: R jedi know what this means. the rest of us don't. and yes we can fire up an R session and do `?nzchar` in order to read the code. but please meet the rest of us halfway at least and put in some comments about what each bit accomplishes. you will be doing the world a kindness and a service.

apis/r/R/SOMASparseNDArray.R
```diff
@@ -74,7 +78,76 @@ SOMASparseNDArray <- R6::R6Class(
...
+        names(bbox)[idx] <- dnames[idx]
+      }
+      bbox <- bbox[dnames]
+      if (length(bbox) != length(dnames)) {
+        stop("dnames")
```
johnkerl, line 103: can be more informative to the person who will read the error message

apis/r/R/SOMASparseNDArray.R
```diff
@@ -74,7 +78,76 @@ SOMASparseNDArray <- R6::R6Class(
...
+            return(sort(c(min(ranges[[x]]), bbox[[x]])))
+          },
+          simplify = FALSE
+        )
+      }
+      for (x in dnames) {
+        xrange <- bbox[[x]]
+        if (any(is.na(xrange))) {
+          stop(
+            "Ranges in the bounding box must be finite (offending: ",
```
johnkerl, line 118: i love the `offending: ...` bit 😎 -- very helpful

apis/r/R/SOMASparseNDArray.R
```diff
@@ -74,7 +78,76 @@ SOMASparseNDArray <- R6::R6Class(
...
+        if (!(is_integerish(xrange) || bit64::is.integer64(xrange))) {
+          stop(
+            "Ranges in the bounding box must be integers",
+            call. = FALSE
+          )
+        }
+        xrange <- sort(bit64::as.integer64(xrange))
+        if (length(xrange) != 2L) {
+          stop(
+            "Ranges in the bounding box must be two 64-bit integers",
```
johnkerl, line 133: must _already_ be, or must be _capable of being converted to_?

apis/r/R/SOMASparseNDArray.R
```diff
@@ -74,7 +78,76 @@ SOMASparseNDArray <- R6::R6Class(
...
+            call. = FALSE
+          )
+        }
+        if (xrange[1L] < 0 || xrange[1L] > min(ranges[[x]])) {
+          stop(
+            "Ranges in the bounding box must be positive and less than the lowest value being added",
```
johnkerl, line 139: you have `< 0` but `positive`

there are negative, zero, and positive

so i think you should be saying non-negative if you're testing not < 0

apis/r/R/TileDBArray.R
```diff
@@ -140,6 +140,71 @@ TileDBArray <- R6::R6Class(
       ))
     },
 
+    used_shape = function(simplify = FALSE, index1 = FALSE) {
```
johnkerl, line 150: needs a docstring -- this is user-facing API

johnkerl (changes requested):

johnkerl:
also: on this PR or a subsequent, some vignette material showing the value add / "why would I use this" for growable soma experiments

note: that will be more useful once append mode even _exists_ in R which it doesn't yet https://github.com/single-cell-data/TileDB-SOMA/issues/1630

apis/r/R/SOMASparseNDArray.R
mojaveazure: This is not the case. The `'values' must be a matrix` message occurs only if `is_matrix(values)` returns `FALSE`; the other conditions don't (yet) have an informative message
```
>  uri <- withr::local_tempdir("sparse-ndarray-bbox")
>  nrows <- 100L
>  ncols <- 500L
>  ndarray <- SOMASparseNDArrayCreate(uri, type = arrow::int32(), shape = c(nrows, ncols))
>  mat <- create_sparse_matrix_with_int_dims(nrows, ncols)
>  ndarray$write(mat, bbox = TRUE) # violation of the second test
Error in ndarray$write(mat, TRUE) :
is.null(bbox) || length(bbox) == length(dim(values)) is not TRUE
```
Informative error messages are needed, and I will add them, but it is already "multiple `stopifnots`"

apis/r/R/TileDBArray.R
mojaveazure: Added in 02fe817

apis/r/R/SOMASparseNDArray.R
mojaveazure: Added in 5ef5a2d

apis/r/R/TileDBArray.R
```diff
@@ -140,6 +140,84 @@ TileDBArray <- R6::R6Class(
       ))
     },
 
+    #' @description Retrieve the range of indexes for a dimension that were
+    #'  explicitly written.
+    #' @param simplify Return a vector of [`bit64:integer64`]s containing only
+    #' the upper bounds.
+    #' @param index1 Return the used shape with 1-based indices.
```
johnkerl, line 147:
```suggestion
    #' @param index1 Return the used shape with 1-based indices (0-based indices are returned by default)
```

apis/r/R/SOMASparseNDArray.R
mojaveazure: Added in 9b3ea56

apis/r/R/SOMASparseNDArray.R
mojaveazure: Addressed in 9b3ea56

apis/r/R/SOMASparseNDArray.R
mojaveazure: Informative error messages added in 9b3ea56

apis/r/R/SOMASparseNDArray.R
mojaveazure: Capable of being converted to; addressed in f38cc1c

johnkerl (approved):
This looks good to me:

```
> devtools::install_github('satijalab/seurat-data')

> library(SeuratData)

> InstallData("pbmc3k")

> data('pbmc3k')

> pbmc3k
An object of class Seurat
13714 features across 2700 samples within 1 assay
Active assay: RNA (13714 features, 0 variable features)
```

```
> write_soma(pbmc3k, 'pbmc3k-tiledb')
[1] "pbmc3k-tiledb"
```

```
$ show-object-metadata.py pbmc3k-tiledb/ms/RNA/X/counts
[array] pbmc3k-tiledb/ms/RNA/X/counts
- soma_dim_0_DOMAIN: (0, 2699)
- soma_dim_1_DOMAIN: (0, 13713)
- soma_encoding_version: b'1'
- soma_object_type: b'SOMASparseNDArray'
``` 

```
> exp <- SOMAExperimentOpen('pbmc3k-tiledb')

> exp$obs$used_shape()
[2023-08-29 14:22:18.754] [default] [Process: 80022] [warning] The following dimensions have no bounding box, non-empty domain used
instead: ‘soma_joinid’
$soma_joinid
integer64
[1] 0    2699

> exp$ms$get("RNA")$X$get("counts")$used_shape()
$soma_dim_0
integer64
[1] 0    2699

$soma_dim_1
integer64
[1] 0     13713
```

johnkerl (changes requested):
@mojaveazure this is awesome!

My only ask is, I don't like the capitalization of `soma_dim_0_DOMAIN` -- let's make it simply `soma_dim_0_domain`

johnkerl (approved):

johnkerl (changes requested):
Sorry for the late update @mojaveazure but the spec
https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md
says that we can only put primitive types into metadata values. (I was reminded of this while drafting a Python port of this PR.) So we need `soma_dim_0_lo` `soma_dim_0_hi` etc as separate ints.

johnkerl (commented):
@mojaveazure I already take that back

https://github.com/single-cell-data/TileDB-SOMA/blob/1.4.3/apis/python/src/tiledbsoma/_tdb_handles.py#L363-L365
has a bug -- it is missing a `return` before the `raise`

johnkerl (approved):

PR #1514, opened 2023-06-30
[python] Ingest uns string arrays as single-attr SOMADataFrame

### LINKED ISSUE #1400: [python] Increase ingestion fraction in `uns` (arrays of string)
There are some spots with `(uns ndarray): unsupported dtype dtype('O') (object)` that were handled in `main-old`, needing to be forward-ported to `main`.

johnkerl:
**Issue and/or context:** Resolves #1400

**Changes:** As discussed on #1400

**Notes for Reviewer:**

```
>>> import tiledbsoma

>>> exp = tiledbsoma.Experiment.open("./pbmc3k")


>>> exp.ms["RNA"]["uns"]["louvain_colors"].read().concat()
pyarrow.Table
soma_joinid: int64
values: large_string
----
soma_joinid: [[0,1,2,3,4,5,6,7]]
values: [["#1f77b4","#ff7f0e","#2ca02c","#d62728","#9467bd","#8c564b","#e377c2","#bcbd22"]]

>>> exp.ms["RNA"]["uns"]["louvain_colors"].read().concat().to_pandas()
   soma_joinid   values
0            0  #1f77b4
1            1  #ff7f0e
2            2  #2ca02c
3            3  #d62728
4            4  #9467bd
5            5  #8c564b
6            6  #e377c2
7            7  #bcbd22
```

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -678,7 +678,7 @@ def _write_dataframe(
     platform_config: Optional[PlatformConfig] = None,
     context: Optional[SOMATileDBContext] = None,
 ) -> DataFrame:
-    s = _util.get_start_stamp()
+    _util.get_start_stamp()
     logging.log_io(None, f"START  WRITING {df_uri}")
 
     df[SOMA_JOINID] = np.arange(len(df), dtype=np.int64)
@@ -687,6 +687,26 @@ def _write_dataframe(
     if id_column_name is not None:
         df.rename(columns={"index": id_column_name}, inplace=True)
     df.set_index(SOMA_JOINID, inplace=True)
+
+    return _write_dataframe_impl(
+        df,
+        df_uri,
+        ingestion_params=ingestion_params,
+        platform_config=platform_config,
+        context=context,
+    )
+
+
+def _write_dataframe_impl(
+    df: pd.DataFrame,
+    df_uri: str,
+    *,
+    ingestion_params: IngestionParams,
+    platform_config: Optional[PlatformConfig] = None,
+    context: Optional[SOMATileDBContext] = None,
+) -> DataFrame:
+    s = _util.get_start_stamp()
+    logging.log_io(None, f"START  WRITING {df_uri}")
 
     arrow_table = _df_to_arrow(df)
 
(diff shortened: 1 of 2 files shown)


johnkerl:
@thetorpedodog @mojaveazure ping

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1377,15 +1394,29 @@ def _ingest_uns_node(
...
-            coll,
-            key,
-            value,
-            platform_config,
-            context=context,
-            use_relative_uri=use_relative_uri,
-            ingestion_params=ingestion_params,
-        )
+
+        if value.dtype.char == "U" or value.dtype.char == "O":
```
thetorpedodog, line 1398: nit: `if value.dtype.char in "UO"` (or `("U", "O")` to be a bit more explicit)

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1377,15 +1394,29 @@ def _ingest_uns_node(
...
+            # In the wild it's quite common to see arrays of strings in uns data.
```
thetorpedodog, line 1399: Is it necessary to check if all the entries in an `O` are strings (or None), or is this an assumption we can make?

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1394,6 +1425,52 @@ def _ingest_uns_node(
...
+            f" (uns object): string-array is not one-dimensional"
+        )
+        logging.log_io(msg, msg)
+        return
+
+    n = len(value)
+    df_uri = _util.uri_joinpath(coll.uri, key)
+    df = pd.DataFrame(
+        data={
+            "soma_joinid": np.asarray(range(n), dtype=np.int64),
```
thetorpedodog, line 1458: np.arange here

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1394,6 +1425,52 @@ def _ingest_uns_node(
...
+            "values": [str(e) for e in value],
```
thetorpedodog, line 1459: If this is nullable, this would turn `None` into `"None"`. Is this what we want? (If the dataframe is non-nullable, `str(e) if e else ""` might be a better formulation, so that Nones become the empty string.)

thetorpedodog (commented):
The code overall looks good; I have some minor questions about things which might be fine in practice.

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: @thetorpedodog this is an assumption we can make

johnkerl:
@thetorpedodog ready for round two 🙏

thetorpedodog (approved):

PR #1516, opened 2023-07-03
Usability improvements to profiler

atolopko-czi:
**Issue and/or context:**

While attempting to use the profiler for CELLxGENE Census PyTorch DataExperimentPipe benchmarking, I ended up making numerous improvements to the profiler.

**Changes:**

Usage changes:
- To ensure availability and consistency of parsed metrics, require use of gnu-time specifically.
- Add option to specify the path to gnu-time command
- Make profiler into runnable module by adding `__main__.py` and `__init___.py` (run as `python -m profiler`)
- Rename `profiler_plot` module to `report` (it can output both json and plots)
- Rename 'process' arg to 'command' for the profiled command
- In report.py module, rename `--pandas` option to `--json` and output as json, using Pandas 'columns' format
- updated README

Database changes:
- Rename db default output dir to 'profiling_db'
- Output db record data in json format, instead of as pickled data (human-readable and more easily processed)
- The db record key now uses cmd hash instead of full string (full cmd can be too long for a filename)
- Store timestamp instead of utcnow() string, renamed from `datetime1
- Remove 'now' from ProfileData (redundant with 'timestamp')
- Save full profiling command  in db, stored as command.txt file for human readability now that hash of cmd string is used in the key

Major code changes:
- Update profiler parsing for gnu-time output
- Refactor context_generator to be a standard method call, rather than as a separate process
- Use gitpython library instead of git process call 

Misc:
- add requirements to `setup.py`

**Notes for Reviewer:**


profiler/data.py
```diff
@@ -1,51 +1,63 @@
...
+    file_system_inputs: int
+    file_system_outputs: int
+    socket_messages_sent: int
+    socket_messages_received: int
+    signals_delivered: int
+    page_size_bytes: int
+    exit_status: int
+    custom_out: List[Optional[str]]
+
+    command_key: str = attr.field()
```
beroy, line 48: What does command_kay do and what exactly does this decorator do?

profiler/data.py
```diff
@@ -1,51 +1,63 @@
...
+    def _command_key_factory(self):
+        return _command_key(self.command)
+
+
+DEFAULT_PROFILE_DB_PATH = "./profiling_db"
+
+
+def _command_key(command: str) -> str:
+    """Remove space characters from profileDB command keys."""
+    return hashlib.md5(command.encode("utf-8")).hexdigest()
```
beroy, line 60: Is the goal here to convert commandline to hash and name the files as hash?

profiler/data.py
```diff
@@ -70,45 +82,50 @@ def close(self):
...
-                    + str(len(os.listdir(f"{self.path}/{process}")))
-                    + "\n"
+                    f"[{command_hash.split('/')[-1]}] \"{command}\": {n_runs} runs\n"
                 )
             return result
         return ""
 
-    def find(self, process) -> List[ProfileData]:
-        key = improve_profileDB_key(process)
+    def find(self, command) -> List[ProfileData]:
```
beroy, line 106: Is command the process plus its arguments?

profiler/profiler.py
```diff
@@ -151,15 +149,17 @@ def main():
...
-    args = parser.parse_args()
+    args = parser.parse_args(sys.argv[1:])
 
-    print(f"Process to be run: {args.process}")
-    # Running the main process using time -v to get detailed memory and time"""
+    print(f"Command to be run: {args.command}", file=stderr)
+    # Run the command, using `time -v` to get detailed memory and time"""
     p = subprocess.Popen(
-        ["/usr/bin/time", "-al"] + args.process, stdout=PIPE, stderr=PIPE
+        [args.gtime_cmd, "--format", GNU_TIME_FORMAT] + args.command.split(" "),
```
beroy, line 157: Where is gtime command set and wondering why gnu time is a `/usr/bin/time` choice than `/usr/bin/time`

profiler/profiler.py
```diff
@@ -3,129 +3,127 @@
...
-    context_process = subprocess.Popen(
-        ["python", "context_generator.py"], stdout=PIPE, stderr=PIPE
-    )
-    stdout, stderr = context_process.communicate()
-    stats = json.loads(stdout.decode("utf-8"))
-    return stats
+from .context_generator import host_context
+from .data import FileBasedProfileDB, ProfileData, ProfileDB
+
+GNU_TIME_FORMAT = (
```
beroy, line 19: Why do we need both `GNU_TIME_FORMAT` and `GNU_TIME_FORMAT`?

profiler/data.py
atolopko-czi: this is a hash of the command, which allows storing the profiling data under a name that has a bounded length; the commands I was using to profile ended up being longer than allowed by the filesystem.

profiler/data.py
atolopko-czi: yup, exactly!

profiler/data.py
atolopko-czi: yes, it's the full command, which includes the args

profiler/profiler.py
atolopko-czi: Now that it's a cmd-line arg, it's up to the user to specify the path to gnu-time if it's not the default value (/usr/bin/time). The Click option for this param defines the default value.

profiler/profiler.py
atolopko-czi: Assuming you mean why do we need both `GNU_TIME_FORMAT` and `GNU_TIME_OUTPUT_REGEXP`. The former is for parsing the output, and the latter is for defining the output. We needed a custom output format to get elapsed (wall clock) time stat as a seconds value (not HH:MM:SS), which simplified things. The elapsed time format differs if the value exceeds an hour, which I found out the hard way. :) So I chose to just output as seconds for simplicity.

Now that we can control the output format, I could have moved to a simpler-to-parse format, but stuck with nearly the same format for now.

PR #1517, opened 2023-07-05
Remove GitHub Python 3.7 build workaround.

thetorpedodog:
Since GitHub appears to have pushed a working version of Python 3.7.17, this workaround is no longer needed.

Rev e66553c6fa76265941546f8474fadeb85ccc784c is a change where I had the build run.


johnkerl (commented):
🚢

PR #1519, opened 2023-07-05
[c++] Support `Enumeration` in C++ codebase

nguyenv:
**Issue and/or context:**

#866

As per discussion with @eddelbuettel, separating the C++ code out from #1511 so that it can be utilized in both the Python and R APIs.

**Changes:**

* Addition of `SOMAArray::get_attr_to_enum_mapping`, `SOMAArray::get_enum` and `SOMAArray::get_enum_label_on_attr`
* Attach an enumeration/dictionary to the `ColumnBuffer`, if applicable; this is used when converting from `ArrayBuffers` to Arrow Tables in the Python and R APIs


libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -432,6 +432,15 @@ std::vector<int64_t> SOMAArray::shape() {
...
 
+Enumeration SOMAArray::get_enum(std::string name) {
+    return ArrayExperimental::get_enumeration(*ctx_, *arr_, name);
+}
+
+std::optional<std::string> SOMAArray::get_enum_label_on_attr(std::string name) {
+    auto attr = arr_->schema().attribute(name);
+    return AttributeExperimental::get_enumeration_name(*ctx_, attr);
+}
+
```
eddelbuettel, line 443: These two are good.  

I keep wondering if we also need
- a 'get me enumerations' given the array, ie something that loops over the attributes and returns them all
- the name mapping is still confusing because attribute name is not necessarily the same as enumeration, but if they differ how do we discover the mapping?
- the second can also we wrapped as a simple `bool` predicate checking whether given a name the return is null or not.

libtiledbsoma/src/utils/column_buffer.cc
```diff
@@ -42,26 +42,31 @@ using namespace tiledb;
...
     std::shared_ptr<Array> array, std::string_view name) {
-    return ColumnBuffer::create(array->schema(), name);
-}
-
-std::shared_ptr<ColumnBuffer> ColumnBuffer::create(
-    ArraySchema schema, std::string_view name) {
+    auto schema = array->schema();
     auto name_str = std::string(name);  // string for TileDB API
 
     if (schema.has_attribute(name_str)) {
```
eddelbuettel, line 48: I so far have also limited myself to attributes but I think the core implementation is more powerful and we could have enumerations as dimensions (!!) given that we already have both int and char dims anyway.

eddelbuettel (commented):
Thanks for starting this.  I may try to give it a whirl building with it next.

libtiledbsoma/src/utils/column_buffer.cc
johnkerl: let's double-check with @davisp -- ?

davisp: Big nopes to enumeration support for dimensions. The primitives are close, but there are a whole bunch of non-obvious behavior issues we’d run into due to dimensions requiring a defined sort and a bunch of other things I don’t know enough about to even begin listing.

Which is to say that its probably not technically infeasible, but the consensus at the design stage was call dimension support out of scope for the MVP/V1 implementation.

eddelbuettel: Thank you so much for clarifying.  I have been running so far with 'attributes only' too but must have hallucinated myself into thinking you had mused about dims too.  I probably poorly inferred from query conditions.   All good then.

davisp: No worries! This is absolutely one of those situations where, given the primitives, it seems like adding the functionality would be straightforward, but then when we start working through the edge cases there are a whole bunch of open questions on what the behavior should be with no clear answers.

The current implementation ignores dimensions purely out of expediency to get something usable first which both makes sure that its usable as designed and should also help shake out anything that was overlooked. I’d wager a small amount of money that enumerations for dimensions will exist eventually, its just a matter of making sure our behaviors are all intentional for long term support.

eddelbuettel: Man you make me look good because I did use the lawyer-advised (kidding here) conditional above:

> and we could have enumerations as dimensions

libtiledbsoma/src/soma/soma_array.cc
nguyenv: I think a function that returns the mapping of attributes with its associated enumeration would be great to add. I'll work on that now.

libtiledbsoma/src/utils/column_buffer.cc
```diff
@@ -56,7 +56,7 @@ std::shared_ptr<ColumnBuffer> ColumnBuffer::create(
         if (enum_name.has_value()) {
             enumeration = std::make_optional<Enumeration>(
                 ArrayExperimental::get_enumeration(
-                    schema.context(), *array, name_str));
+                    schema.context(), *array, *enum_name));
```
eddelbuettel, line 59: Nice.  That may have been what I ran into earlier flipping between my branch and your branch and trying an example with 'multiple column using one enumeration'.  Will try this tomorrow.

eddelbuettel: I did indeed do the trick.

nguyenv:
Should we also merge the C++ changes from #1674 into this branch?

johnkerl:
@nguyenv sure? @eddelbuettel what do you think?

eddelbuettel:
@johnkerl @nguyenv I think we will have some truly awful rebases coming up as a few files, notably in libtiledbsoma, were changed concurrently in different branches.  The good news is that things work now, so if we're careful they will still once we're done aligning.

nguyenv:
Yeah I am anxious about all merge conflicts that are about to happen and wondering why the best way to go about this is.

johnkerl:
@nguyenv @eddelbuettel I take that back -- preferring to keep this C++ PR foundational, merge it without 2.17.0-rc0 dependencies, and go from there.

johnkerl (approved):

johnkerl:
> Yeah I am anxious about all merge conflicts that are about to happen and wondering why the best way to go about this is.

One at a time. This PR is foundational. #1511 next.

Issue #1518, opened 2023-07-05
[r] Slow export of obsm/varm arrays to Seurat


johnkerl:
As discussed 1-1 with @aaronwolen in Slack today -- @aaronwolen please fill out details

aaronwolen:
As part of #1504 `SOMAExperimentAxisQuery$to_seurat_graph()` was updated to leverage the newer `to_sparse_matrix()` method and avoid creating CsparseMatrix objects with large dimensions, which is computationally expensive.

These changed need to be carried through to `to_seurat_reduction()` to avoid the same problem, which @johnkerl encountered in CI [here](https://github.com/single-cell-data/TileDB-SOMA/pull/1440#issuecomment-1614953938).

aaronwolen:
Fixed in #1521

PR #1521, opened 2023-07-06
[r] Optimize export of obsm/varm arrays to Seurat

aaronwolen:
**Issue and/or context:** #1518

**Changes:** In #1504 `SOMAExperimentAxisQuery$to_seurat_graph()` was updated to leverage the newer `to_sparse_matrix()`, which  avoided the expensive creation of CsparseMatrix objects with dimensions equal to the underlying array's domain. This carries over the same change to `SOMAExperimentAxisQuery$to_seurat_graph()`.

This also addresses the CI issues described [here](https://github.com/single-cell-data/TileDB-SOMA/pull/1440#issuecomment-1614953938) and blocking #1440.

**Notes for Reviewer:** I'm on the fence about whether or not we need to throw a warning every time a user attempts to call `to_seurat_reduction()` on a dense ND array. It should be an incredibly rare occurrence and I'm not sure what value the warning provides. Let me know if you have thoughts.

### DIFF
--- a/apis/r/R/utils-assertions.R
+++ b/apis/r/R/utils-assertions.R
@@ -208,7 +208,7 @@ warn_if_dense <- function(collection_name, layer) {
       "  - all '%s' arrays should be saved as 'SOMASparseNDArrays'\n",
       collection_name
     )
-    warning(msg1, msg2, call. = FALSE, .immediate = TRUE)
+    warning(msg1, msg2, call. = FALSE, immediate. = TRUE)
   }
   invisible(is_sparse)
 }
--- a/apis/r/R/utils-assertions.R
+++ b/apis/r/R/utils-assertions.R
@@ -185,3 +185,30 @@ recursively_make_integer64 <- function(x) {
     }
     x
 }
+
+#' Warn if using a SOMADenseNDArray
+#' @param collection_name name of the SOMA collection
+#' @param layer a SOMASparseNDArray or SOMADenseNDArray
+#' @return Invisible logical indicating if the layer is sparse
+#' @references https://github.com/single-cell-data/TileDB-SOMA/issues/1245
+#' @noRd
+
+warn_if_dense <- function(collection_name, layer) {
+  stopifnot(
+    is_scalar_character(collection_name),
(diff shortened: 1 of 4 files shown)


apis/r/R/utils-assertions.R
```diff
@@ -185,3 +185,30 @@ recursively_make_integer64 <- function(x) {
...
+  if (!is_sparse) {
+    msg1 <- sprintf(
+      "Values for '%s' are encoded as dense instead of sparse\n",
+      basename(layer$uri)
+    )
+    msg2 <- sprintf(
+      "  - all '%s' arrays should be saved as 'SOMASparseNDArrays'\n",
+      collection_name
+    )
+    warning(msg1, msg2, call. = FALSE, .immediate = TRUE)
```
mojaveazure, line 211: The warning exists to inform the user that their SOMA is techinically valid, but uses data structures that are not preferred. I belive this information is useful, but you're right that doing it every time is a bit much. We could set it to once-per-session using R options

```suggestion
    if (isTRUE(getOption("tiledbsoma.warnings.m_axis_dense", TRUE))) {
      warning(msg1, msg2, call. = FALSE, immediate.= TRUE)
      options(tiledbsoma.warnings.m_axis_dense = FALSE)
    }
```

mojaveazure, line 211: This should be [`immediate.`, not `.immediate`](https://stat.ethz.ch/R-manual/R-devel/library/base/html/warning.html)
```suggestion
    warning(msg1, msg2, call. = FALSE, immediate. = TRUE)
```

mojaveazure (approved):

apis/r/R/utils-assertions.R
aaronwolen: Doh!

mojaveazure: (note that this will likely require changes to the tests)

aaronwolen: I like the idea of once per-session warnings. There are a few places in the package we could apply this to so let's punt for now and make a separate PR that implements it package-wide.

PR #1523, opened 2023-07-07
[r] Add SOMANDArrayBase class

aaronwolen:
**Issue and/or context:** This adds a new class `SOMANDArrayBase` which centralizes the common functionality between `SOMASparseNDArray` and `SOMADDenseNDArray`. Beyond just reducing code duplication, it also more closely matches the Python implementation for easier maintenance/updates across both packages.

**Notes for Reviewer:** Since we now depend on `rlang` I replaced a couple of our assertion functions with `rlang`'s


apis/r/R/SOMADenseNDArray.R
```diff
@@ -128,7 +45,7 @@ SOMADenseNDArray <- R6::R6Class(
       result_order <- map_query_layout(match_query_layout(result_order))
 
       if (!is.null(coords)) {
-        coords <- private$convert_coords(coords)
+        coords <- private$.convert_coords(coords)
```
eddelbuettel, line 48: (Mere comment and not a request for change:)  Declaring a method `private` is pretty expressive in my reading.  I noticed that we starting sprinkling leading dots onto them two.  On the margin I find this makes it less readable.

aaronwolen: I really don't have a strong opinion on this, other than we should be consistent. I will say I find the `.` prefix helpful for quickly spotting private methods when scrolling through a large class with lots of methods and properties but there are other signals too.

eddelbuettel: Jump down half a page to where the 'privates' are declared and note that some have dots, and others don't. Why?

aaronwolen: Different PRs from different people at different times. I added dots to the the private methods/fields relevant to this PR but left others as they are. I'm all for a separate PR that addresses this inconsistency in one fell swoop.

johnkerl: I'll happily dot-prefix private methods

eddelbuettel (approved):

Issue #1494, opened 2023-06-21
[Bug] [r] For `SOMASparseNDArray` in `sr_setup` specifying coordinates (dim_points) beyond domain returns full array instead of empty array or error


pablo-gar:
**Describe the bug**
In `sr_setup` specifying coordinates (dim_points) beyond domain returns full array instead of empty array. See example below.

**To Reproduce**
The sparse matrix has a shape of 80 by 230. Using coords to read the first 5 rows returns intended results. Using coords to read 5 rows beyond domain (>80) returns the full array. 

```r
library(tiledbsoma)

ctx <- tiledbsoma::SOMATileDBContext$new()
config <- as.character(tiledb::config(ctx$context()))

human_experiment = load_dataset("soma-exp-pbmc-small", tiledbsoma_ctx = ctx)

X <- human_experiment$ms$get("RNA")$X$get("data")
X$shape()
# integer64
# [1] 80  230

#----------------
# First 5 rows and  first 5 columns (works as intended)
coords <- list(soma_dim_0=bit64::as.integer64(0:5),
               soma_dim_1=bit64::as.integer64(0:5))

sr <- sr_setup(uri = X$uri, config = config, dim_points = coords)     
chunk <- sr_next(sr)
                   
arrow::as_arrow_table(                   
    arrow::RecordBatch$import_from_c(chunk$array_data, chunk$schema)    
)

# Table
# 5 rows x 3 columns
# $soma_dim_0 <int64 not null>
# $soma_dim_1 <int64 not null>
# $soma_data <double not null>

#----------------
# Go beyond the limts in rows (incorrectly returns full array)
coords <- list(soma_dim_0=bit64::as.integer64(81:86),
               soma_dim_1=bit64::as.integer64(0:5))

sr <- sr_setup(uri = X$uri, config = config, dim_points = coords)     
chunk <- sr_next(sr)
                   
arrow::as_arrow_table(                   
    arrow::RecordBatch$import_from_c(chunk$array_data, chunk$schema)    
)

# Table
# 141 rows x 3 columns
# $soma_dim_0 <int64 not null>
# $soma_dim_1 <int64 not null>
# $soma_data <double not null>
```

**Versions (please complete the following information):**
```r
R version 4.3.0 (2023-04-21)
Platform: x86_64-apple-darwin20 (64-bit)
Running under: macOS Ventura 13.4

Matrix products: default
[...]

PR #1526, opened 2023-07-10
[r] Check dimension domain when setting dim points and ranges

eddelbuettel:
**Issue and/or context:**

In issue #1494 it has been highlighted that out-of-domain 'dim points' (and 'dim ranges') are not checked in the iterated reader and that an error should be triggered for unsuitable values.

**Changes:**

This PR adds checks that requested dimensions points (and ranges) conform to the domain of the respective dimension.

**Notes for Reviewer:**

#1494


johnkerl (approved):
🚀

aaronwolen (approved):
Thanks!

pablo-gar (approved):
Can't comment on the cpp changes. But the tests address the original issue.

PR #1533, opened 2023-07-17
[c++/python] Fix new pybind11 capsule build error

johnkerl:
**Issue and/or context:**

Addresses recent CI fails as of #1530 #1531 #1532, starting July 15.

Also note `pybind11` 2.11.0 dropped two days ago (July 14) https://pypi.org/project/pybind11/#files.

I'm still not entirely clear here, as I have `pybind11` 2.10.0 -- not 2.11.0 -- in my EC2 Linux sandbox where this problem also now appears. That, combined with the fact that the build failure is in C++.

**Changes:** As recommended by @ihnorton who has pointed out a similar change for https://github.com/TileDB-Inc/TileDB-Py

**Notes for Reviewer:**

### DIFF
--- a/apis/python/pyproject.toml
+++ b/apis/python/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-    "pybind11[global]==2.10.0",
+    "pybind11[global]>=2.10.0",
     "setuptools>=65.5.1",
     "wheel>=0.37.1",
     "cmake>=3.21",
--- a/apis/python/pyproject.toml
+++ b/apis/python/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-    "pybind11[global]>=2.10.0",
+    "pybind11[global]==2.10.0",
     "setuptools>=65.5.1",
     "wheel>=0.37.1",
     "cmake>=3.21",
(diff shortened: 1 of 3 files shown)


johnkerl:
Note: this is failing to build for me in my Ubuntu sandbox, on the exact same checkout...

https://gist.github.com/johnkerl/6485c23fc37bc378ca1b40e9f5ccec82

nguyenv:
I am also having issues installing locally on Ubuntu even with the version downgrade though I am not sure why. What worked for me was applying this patch:

```
diff --git a/apis/python/src/tiledbsoma/query_condition.cc b/apis/python/src/tiledbsoma/query_condition.cc
index 1330681d..b08dcaca 100644
--- a/apis/python/src/tiledbsoma/query_condition.cc
+++ b/apis/python/src/tiledbsoma/query_condition.cc
@@ -100,7 +100,7 @@ public:

   shared_ptr<QueryCondition> ptr() { return qc_; }

-  py::capsule __capsule__() { return py::capsule(&qc_, "qc", nullptr); }
+  py::capsule __capsule__() { return py::capsule(&qc_, "qc"); }

   PyQueryCondition
   combine(PyQueryCondition rhs,
```

I don't think we need the nullptr here. The third argument is for an optional destructor.

nguyenv (approved):

PR #1554, opened 2023-07-27
[python] Append-mode pre-check logic

johnkerl:
**Issue and/or context:**

Tracking issue: #1278 

**Changes:**

* Create the `Signature` class which will be used for append-mode pre-check on an upcoming PR
* Unit-test that

**Notes for Reviewer:**

* This is a split-out PR -- created for mercifulness to reviewers -- as a line-count reduction from raw material on #1498.
* Question for @aaronwolen @mojaveazure : If an input file has raw and non-raw data do we want them both to be ingested? Or just one or the other? This is actually a design question -- which I'll also put in our tracking doc.

### DIFF
--- a/apis/python/src/tiledbsoma/_arrow_types.py
+++ b/apis/python/src/tiledbsoma/_arrow_types.py
@@ -30,6 +30,7 @@ from typing import Any, Dict, Union
 
 import numpy as np
 import numpy.typing as npt
+import pandas as pd
 import pyarrow as pa
 import tiledb
 
@@ -159,3 +160,31 @@ def tiledb_schema_to_arrow(tdb_schema: tiledb.ArraySchema) -> pa.Schema:
         arrow_schema_dict[name] = arrow_type_from_tiledb_dtype(attr.dtype, attr.isascii)
 
     return pa.schema(arrow_schema_dict)
+
+
+def df_to_arrow(df: pd.DataFrame) -> pa.Table:
+    """
+    Categoricals are not yet well supported, so we must flatten.
+    Also replace Numpy/Pandas-style nulls with Arrow-style nulls.
+    """
+    null_fields = set()
+    for k in df:
+        if df[k].dtype == "category":
+            df[k] = df[k].astype(df[k].cat.categories.dtype)
+        if df[k].isnull().any():
+            if df[k].isnull().all():
+                df[k] = pa.nulls(df.shape[0], pa.infer_type(df[k]))
+            else:
+                df[k].where(
+                    df[k].notnull(),
+                    pd.Series(pa.nulls(df[k].isnull().sum(), pa.infer_type(df[k]))),
+                    inplace=True,
+                )
+            null_fields.add(k)
+    arrow_table = pa.Table.from_pandas(df)
+    if null_fields:
+        md = arrow_table.schema.metadata
+        md.update(dict.fromkeys(null_fields, "nullable"))
+        arrow_table = arrow_table.replace_schema_metadata(md)
+
+    return arrow_table
--- a/apis/python/src/tiledbsoma/_arrow_types.py
+++ b/apis/python/src/tiledbsoma/_arrow_types.py
@@ -168,6 +168,7 @@ def df_to_arrow(df: pd.DataFrame) -> pa.Table:
     Also replace Numpy/Pandas-style nulls with Arrow-style nulls.
     """
     null_fields = set()
+    # Not for name, col in df.items() since we need df[k] on the left-hand sides
     for k in df:
         if df[k].dtype == "category":
             df[k] = df[k].astype(df[k].cat.categories.dtype)
--- a/apis/python/src/tiledbsoma/io/registration/signatures.py
+++ b/apis/python/src/tiledbsoma/io/registration/signatures.py
@@ -41,8 +41,7 @@ def _string_dict_from_arrow_schema(schema: pa.Schema) -> Dict[str, str]:
     # pre-check an AnnData/H5AD input to see if it's appendable to an existing SOMA experiment, we
     # must not punish the AnnData/H5AD input for it not having a soma_joinid column in its obs and
     # var.
-    if "soma_joinid" in retval:
-        del retval["soma_joinid"]
+    retval.pop("soma_joinid", None)
     return retval
 
 
--- a/apis/python/src/tiledbsoma/io/registration/signatures.py
+++ b/apis/python/src/tiledbsoma/io/registration/signatures.py
@@ -110,7 +110,7 @@ class Signature:
     varm_dtypes: Dict[str, str]
 
     @classmethod
-    def fromAnnData(
+    def from_anndata(
         cls,
         adata: ad.AnnData,
         *,
@@ -175,7 +175,7 @@ class Signature:
         )
 
     @classmethod
-    def fromH5AD(
+    def from_h5ad(
         cls,
         h5ad_file_name: str,
         *,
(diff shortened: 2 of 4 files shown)


apis/python/src/tiledbsoma/io/registration/signatures.py
```diff
@@ -0,0 +1,305 @@
...
+        retval = str(t)
+        # As noted in the Signature class, we pre-check logic from the ingestor.
+        # As detailed elsewhere, Arrow string and large_string must map to TileDB
+        # string, which is large-only. Thus string and large_string form an equivalence
+        # class. Similarly for Arrow binary and large_binary.
+        if retval == "large_string":
+            return "string"
+        if retval == "large_binary":
+            return "binary"
+        return retval
```
thetorpedodog, line 31: Optional: This could be implemented with a little less repetition with a dictionary:

```python
_TYPE_MAP = {
  "large_string": "string",
  "large_binary": "binary",
}

def _stringify_type(t: pa.DataType) -> str:
  str_t = str(t)
  return _TYPE_MAP.get(str_t, str_t)
```

I would also recommend pulling the `stringify_type` function to the root level; since it doesn’t use anything from its closure it doesn’t really gain anything from being in one.

apis/python/src/tiledbsoma/io/registration/signatures.py
```diff
@@ -0,0 +1,305 @@
...
+    # TODO include 'raw' in X_dtypes or no? Different for AnnData and for SOMA. When in doubt,
+    # lean SOMA.
+    X_dtypes: Dict[str, str]
+    raw_X_dtype: Optional[str]
+
+    obsm_dtypes: Dict[str, str]
+    varm_dtypes: Dict[str, str]
+
+    @classmethod
+    def fromAnnData(
```
thetorpedodog, line 108: snake case `from_anndata` etc.?

apis/python/src/tiledbsoma/io/registration/signatures.py
```diff
@@ -0,0 +1,305 @@
...
+                var_schema=var_schema,
+                X_dtypes=X_dtypes,
+                raw_X_dtype=raw_X_dtype,
+                raw_var_schema=raw_var_schema,
+                obsm_dtypes=obsm_dtypes,
+                varm_dtypes=varm_dtypes,
+            )
+
+    @classmethod
+    def compatible(cls, signatures: Dict[str, Self]) -> Tuple[bool, Optional[str]]:
```
thetorpedodog, line 245: This seems more like you would want to handle it as a `check_compatibility` function that maybe raised an exception of some kind. The signature as-is has the risk of being misunderstood; a person could write

```python
if compatible(some_sigs):
  do_dangerous_thing(...)
```

and since a 2-tuple is non-empty, it would always evaluate as truthy and `do_dangerous_thing`

apis/python/src/tiledbsoma/io/registration/signatures.py
```diff
@@ -0,0 +1,305 @@
...
+        """
+        Determines if two signatures from SOMA experiment or AnnData/H5AD will be safe from
```
thetorpedodog, line 247: any number of signatures

apis/python/src/tiledbsoma/io/registration/signatures.py
```diff
@@ -0,0 +1,305 @@
...
+            return False
+        if self.raw_var_schema != other.raw_var_schema:
+            return False
+
+        if self.obsm_dtypes != other.obsm_dtypes:
+            return False
+        if self.varm_dtypes != other.varm_dtypes:
+            return False
+
+        return True
```
thetorpedodog, line 288: Can this be replaced with `self == other`?

apis/python/src/tiledbsoma/io/registration/signatures.py
```diff
@@ -0,0 +1,305 @@
...
+    def fromJSON(cls, s: str) -> Self:
+        dikt = json.loads(s)
+        return cls(
+            dikt["obs_schema"],
+            dikt["var_schema"],
+            dikt["raw_var_schema"],
+            dikt["X_dtypes"],
+            dikt["raw_X_dtype"],
+            dikt["obsm_dtypes"],
+            dikt["varm_dtypes"],
```
thetorpedodog, line 304: You can save some typing with `cls(**dikt)`. Also maybe worth declaring the class as `kw_only=True`, because it seems like there’s no obviously correct order for these.

apis/python/src/tiledbsoma/_arrow_types.py
```diff
@@ -159,3 +160,31 @@ def tiledb_schema_to_arrow(tdb_schema: tiledb.ArraySchema) -> pa.Schema:
...
     return pa.schema(arrow_schema_dict)
+
+
+def df_to_arrow(df: pd.DataFrame) -> pa.Table:
+    """
+    Categoricals are not yet well supported, so we must flatten.
+    Also replace Numpy/Pandas-style nulls with Arrow-style nulls.
+    """
+    null_fields = set()
+    for k in df:
```
thetorpedodog, line 171: I see this is just moved code, so you don’t need to make this change if you don’t want to, but:

Dataframes implement the `dict`-like `.items` method, so this can be `for name, col in df.items()` to make some usage of `df[k]` a bit more readable

thetorpedodog (commented):
Overall the structure of this code is good; I do have some suggestions and a few concerns about the API right now, though.

apis/python/src/tiledbsoma/_arrow_types.py
johnkerl: @thetorpedodog no b/c I want `df[k]` as left-hand sides to mutate

johnkerl:
@thetorpedodog ready for round two

apis/python/src/tiledbsoma/io/registration/signatures.py
```diff
@@ -0,0 +1,308 @@
...
+    distributed-computing/distributed-logging environment.
+
+    As noted in the Signature class, we pre-check logic from the ingestor.  As detailed elsewhere,
+    Arrow string and large_string must map to TileDB string, which is large-only. Thus string and
+    large_string form an equivalence class. Similarly for Arrow binary and large_binary.
+    """
+    _EQUIVALENCES = {
+        "large_string": "string",
+        "large_binary": "binary",
+    }
```
thetorpedodog, line 27: this can either be a module-level constant or make it a regular variable name like `equivalences`

apis/python/src/tiledbsoma/io/registration/signatures.py
```diff
@@ -0,0 +1,308 @@
...
+            return False
+
+        if self.obsm_dtypes != other.obsm_dtypes:
+            return False
+        if self.varm_dtypes != other.varm_dtypes:
+            return False
+
+        return True
+
+    def toJSON(self) -> str:
```
thetorpedodog, line 299: nit: `to_json`/`from_json`

apis/python/src/tiledbsoma/io/registration/signatures.py
```diff
@@ -0,0 +1,308 @@
...
+        """Presents a signature as JSON which is suitable for distributed logging."""
+        return json.dumps(
+            self, default=lambda o: attrs.asdict(o), sort_keys=True, indent=4
```
thetorpedodog, line 302: This seems like it could be `json.dumps(attrs.asdict(self), ...)` rather than setting `default=`. Am I missing something?

(If there is a reason to set `default`, then it can be shortened to `default=attrs.asdict`, since that’s all the lambda does.)

apis/python/src/tiledbsoma/io/registration/signatures.py
```diff
@@ -0,0 +1,308 @@
...
+        df.rename(columns={"index": default_index_name}, inplace=True)
+    else:
+        df.reset_index(inplace=True)
+
+    arrow_table = df_to_arrow(df)
+    arrow_schema = arrow_table.schema.remove_metadata()
+    return _string_dict_from_arrow_schema(arrow_schema)
+
+
+@attrs.define
```
thetorpedodog, line 77: I would [use `kw_only` here](https://www.attrs.org/en/stable/api-attr.html#attr.s:~:text=kw_only%20,__init__) to avoid accidentally passing things in as positional arguments. 

```python
@attrs.define(kw_only=True)
class Signature:
  ...
```

thetorpedodog (approved):
Other than the few comments below, looks good to me style-wise. The logic seems reasonable but I’m not *super* familiar; you may wish to get another reviewer to give that a once-over (but not mandatory).

apis/python/src/tiledbsoma/io/registration/signatures.py
johnkerl: It is necessary to exist; I get errors without. I'll simplify the syntax as you suggest.

johnkerl:
@aaronwolen @mojaveazure any final concerns before I merge? This is of course a multi-PR (at least 3-PR) WIP and I'm happy to take feedback before or after the merge ...

johnkerl:
Merging -- happy to take belated feedback as a follow-on PR

PR #1556, opened 2023-07-27
[python] Append-mode label/ID-mapping logic

johnkerl:
**Issue and/or context:**

Tracking issue: #1278 

**Changes:**

* Create the classes `ExperimentAmbientLabelMapping`, `AxisAmbientLabelMapping`, `ExperimentIDMapping`, and `AxisIDMapping`
* None of these will be exposed to the user; they are internal-only.

**Notes for Reviewer:**

* This is a split-out PR -- created for mercifulness to reviewers -- as a line-count reduction from raw material on #1498. See also #1554.
* See also #1554 which is another, orthogonal component.
* Actual use by the ingestor logic will be on a subsequent PR.

### DIFF
--- a/apis/python/src/tiledbsoma/io/registration/__init__.py
+++ b/apis/python/src/tiledbsoma/io/registration/__init__.py
@@ -16,7 +16,7 @@ Essential ideas:
   we will expect total input to have data for 300 cells. In particular, as multiple inputs
   are appended to a SOMA experiment, ``obs`` will grow taller.
 
-- The input ``var`` must also havesome (user-specified) column containing string identifiers.
+- The input ``var`` must also have some (user-specified) column containing string identifiers.
   Nominally these are Ensembl IDs like ENSG00000142208, or HGNC IDs like AKT1. Nominally
   these will be the same for all input files, although it's to be expected that one input file
   may have data for some infrequently expressed genes that don't appear in other input files.
--- a/apis/python/src/tiledbsoma/io/registration/__init__.py
+++ b/apis/python/src/tiledbsoma/io/registration/__init__.py
@@ -1,4 +1,102 @@
-"""TODO: docstring"""
+"""
+Support for soma_joinid remapping for append-mode ingestion.
+
+This is an internal-use class; none of it is user-facing API.
+
+The SOMA experiment ``obs``, ``var``, ``X``, etc. are indexed by soma_joinid.  Input AnnData/H5AD
+``obs`` and ``var`` are indexed by an index column; ``X` et al. are indexed zero-up by row numbers
+aligning with ``obs`` and ``var``. This raises the issue: if we have a SOMA experiment created
+by ingesting two or more AnnData/H5AD files, how do we compute soma_joinid values for the full experiment?
+
+Essential ideas:
+
+- The input ``obs`` must have some (user-specified) column which contains values unique across all
+  AnnData/H5AD inputs. Nominally this will be a cell barcode, and nominally this will be different
+  for all input files. If one input has data for 100 cells and another has data for 200 cells,
+  we will expect total input to have data for 300 cells. In particular, as multiple inputs
+  are appended to a SOMA experiment, ``obs`` will grow taller.
+
+- The input ``var`` must also havesome (user-specified) column containing string identifiers.
+  Nominally these are Ensembl IDs like ENSG00000142208, or HGNC IDs like AKT1. Nominally
+  these will be the same for all input files, although it's to be expected that one input file
+  may have data for some infrequently expressed genes that don't appear in other input files.
+  In particular, as multiple inputs are appended to a SOMA experiment, ``var`` may gain
+  a few rows here and there.
+
+- Putting the last two together means that ``X`` (which is sparse) will mainly gain new
+  rows (cells) for each input file, perhaps with some columns (genes) that didn't appear for
+  previous input files.
+
+The purpose of the registration mappings is to simply track a mapping from user-specified
+``obs`` and ``var`` ID-column values to soma_joinid values.
+
+Append-mode ingestion has a _registration pass_, which must be sequential and must use all input
+files to compute join-ID mappings, and then an _ingestion pass_, which can be parallelized across
+input files.
+
+There are two kinds of mappings: _ambient mappings_ which contain string-to-int join-ID mappings
+for _all_ inputs, and ID mappings which contain int-to-int offset-to-join-ID mappings for each _single_ input.
+
+Example:
+
+- Input 1 has obs IDs ``["AAAT", "ACTG", "AGAG"]`` numbered 0, 1, 2 within input 1.
+- Input 1 has var IDs ``["AKT1", "APOE", "ESR1", "TP53", "VEGFA"]`` numbered 0, 1, 2, 3, 4 within input1.
+- Input 2 has obs IDs ``["CAAT", "CCTG", "CGAG"]`` numbered 0, 1, 2 within input 2.
+- Input 2 has var IDs ``["APOE", "EGFR", "TP53", "VEGFA"]`` numbered 0, 1, 2, 3 within input 2.
+
+Then registration produces ambient mappings like this:
+
+- For ``obs``, ``AxisAmbientLabelMapping`` of
+
+    AAAT:0
+    ACTG:1
+    AGAG:2
+    CAAT:3
+    CCTG:4
+    CGAG:5
+
+- For ``var``, ``AxisAmbientLabelMapping`` of
+
+    AKT1:0
+    APOE:1
+    ESR1:2
+    TP53:3
+    VEGFA:4
+    EGFR:5
+
+- ``ExperimentAmbientLabelMapping`` containing these two axes.
+
+This registration data is passed to the ingestor for each input file. Within the ingestion logic
+itself and without user intervention, the ingestor selects out the int-to-int mappings from AnnData 0-up
+offsets to registered SOMA join IDs like this:
+
+- For input 1's ``obs``, ``AxisIDMapping`` of
+
+    0:0 (for AAAT)
+    1:1 (for ACTG)
+    2:2 (for AGAG)
+
+- For input 1's ``var``, ``AxisIDMapping`` of
+
+    0:0 (for AKT1)
+    1:1 (for APOE)
+    2:2 (for ESR1)
+    3:3 (for TP53)
+    4:4 (for VEGFA)
+
+- For input 2's ``obs``, ``AxisIDMapping`` of
+
+    3:3 (for CAAT)
+    4:4 (for CCTG)
+    5:5 (for CGAG)
+
+- For input 2's ``var``, ``AxisIDMapping`` of
+
+    0:1 (for APOE)
+    1:5 (for EGFR)
+    2:3 (for TP53)
+    3:4 (for VEGFA)
+"""
 
 from .ambient_label_mappings import (
     AxisAmbientLabelMapping,
--- a/apis/python/src/tiledbsoma/io/registration/__init__.py
+++ b/apis/python/src/tiledbsoma/io/registration/__init__.py
@@ -1,0 +1,15 @@
+"""TODO: docstring"""
+
+from .ambient_label_mappings import (
+    AxisAmbientLabelMapping,
+    ExperimentAmbientLabelMapping,
+)
+from .id_mappings import AxisIDMapping, ExperimentIDMapping, get_dataframe_values
+
+__all__ = (
+    "AxisIDMapping",
+    "AxisAmbientLabelMapping",
+    "ExperimentIDMapping",
+    "ExperimentAmbientLabelMapping",
+    "get_dataframe_values",
+)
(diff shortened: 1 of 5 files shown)


apis/python/src/tiledbsoma/io/registration/__init__.py
```diff
@@ -0,0 +1,113 @@
+"""
+Support for soma_joinid remapping for append-mode ingestion.
+
+This is an internal-use class; none of it is user-facing API.
```
thetorpedodog, line 4: An internal-use module? If so, might be worth naming it `tiledbsoma.io._registration`.

apis/python/src/tiledbsoma/io/registration/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+
+Essential ideas:
+
+- The input ``obs`` must have some (user-specified) column which contains values unique across all
+  AnnData/H5AD inputs. Nominally this will be a cell barcode, and nominally this will be different
+  for all input files. If one input has data for 100 cells and another has data for 200 cells,
+  we will expect total input to have data for 300 cells. In particular, as multiple inputs
+  are appended to a SOMA experiment, ``obs`` will grow taller.
+
+- The input ``var`` must also havesome (user-specified) column containing string identifiers.
```
thetorpedodog, line 19: "have some"

apis/python/src/tiledbsoma/io/registration/ambient_label_mappings.py
```diff
@@ -0,0 +1,405 @@
...
+        isolation. This is used when a user is ingesting a single AnnData/H5AD to a single SOMA
+        experiment, not in append mode, but allowing us to still have the bulk of the ingestor code
+        to be non-duplicated between non-append mode and append mode.
+        """
+        tiledbsoma.logging.logger.info("Registration: registering AnnData dataframe.")
+
+        if index_field_name is None:
+            index_field_name = df.index.name
+        if index_field_name is None:
+            index_field_name = "index"
```
thetorpedodog, line 74: suggestion: use `or`s to shorten this:

`index_field_name = index_field_name or df.index.name or "index"`

apis/python/src/tiledbsoma/io/registration/id_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+import anndata as ad
+import attrs
+import pandas as pd
+from typing_extensions import Self
+
+import tiledbsoma
+import tiledbsoma.logging
+
+
+@attrs.define(kw_only=True)
```
thetorpedodog, line 12: Since this is a one-property class `kw_only` seems excessive to me here, but it’s fine to keep it this way if you prefer.

apis/python/src/tiledbsoma/io/registration/id_mappings.py
```diff
@@ -0,0 +1,81 @@
...
+class AxisIDMapping:
+    """
+    For a single to-be-appended AnnData/H5AD input in SOMA multi-file append-mode ingestion, this
+    class tracks the mapping of input-data ``obs`` or ``var`` 0-up offsets to SOMA join ID values
+    for the destination SOMA experiment.
+
+    See module-level comments for more information.
+    """
+
+    data: List[int]
```
thetorpedodog, line 22: Do we expect to be mutating this object during the course of an ingestion (i.e., changing the contents of this list, as opposed to creating a new `AxisIDMapping` with different contents)? If not, it might be worth making this a `Tuple[int, ...]` to avoid accidental spooky action at a distance (e.g. somebody passes `whatever.data` into a method that mutates it).

apis/python/tests/test_registration_mappings.py
```diff
@@ -0,0 +1,314 @@
...
+import numpy as np
+import pandas as pd
+import pytest
+
+import tiledbsoma.io
+import tiledbsoma.io.registration as registration
+
+"""
+Test join-id registrations for ingesting multiple AnnData objects into a single SOMA Experiment.
+"""
```
thetorpedodog, line 15: nit: module docstrings need to go above all other statements, including imports

thetorpedodog (commented):
This was actually quite a bit easier to get through than I expected! Overall very well written and documented. I have a few minor suggestions but otherwise no concerns from me.

apis/python/src/tiledbsoma/io/registration/id_mappings.py
johnkerl: @thetorpedodog it has one property _today_ but it may have two _tomorrow_ and your code-review guidance on #1554 was excellent. Let us not in our over-eager parsimony deprive future maintainers of _this_ file from excellent advice implement on _that_ file. Let us emplace good design patterns from the get-go. Not to mention simple consistency -- _all_ the `@attrs` classes in this module will be `kw_only`, not a mix of some-yes-some-no.

apis/python/src/tiledbsoma/io/registration/__init__.py
johnkerl: @thetorpedodog as noted in this PR's description field as well as #1278 this PR is one in a series. There is user-facing (simple, but user-facing) API content in `tiledbsoma.io.registration` coming up.

johnkerl:
@thetorpedodog ready for round two! :)

johnkerl:
@thetorpedodog ping 🙏

apis/python/src/tiledbsoma/io/registration/__init__.py
thetorpedodog: OK, but this is a module, not a class. And given that this will be exporting public-facing APIs, and it *is* exporting several classes/functions, it is kind of weird to have "none of it is user-facing API" in the docstring (is it temporary in case we do a release with this included but not finished?). Whatever the goal is is fine, just the docstring and the code itself are sending mixed messages.

apis/python/src/tiledbsoma/io/registration/ambient_label_mappings.py
```diff
@@ -0,0 +1,402 @@
...
+        )
+
+    @classmethod
+    def from_isolated_soma_experiment(
+        cls,
+        experiment_uri: Optional[str] = None,
+        *,
+        obs_field_name: str = "obs_id",
+        var_field_name: str = "var_id",
+        # XXX CTX/CFG
```
thetorpedodog, line 204: note-to-self comment from development?

apis/python/src/tiledbsoma/io/registration/ambient_label_mappings.py
```diff
@@ -0,0 +1,402 @@
...
+            )
+
+        else:
+            tiledbsoma.logging.logger.info(
+                f"Registration: starting with experiment {experiment_uri}"
+            )
+
+            with tiledbsoma.Experiment.open(experiment_uri) as exp:
+                for batch in exp.obs.read(column_names=["soma_joinid", obs_field_name]):
+                    for i in range(len(batch[0])):
```
thetorpedodog, line 227: could this be `zip(batch[0], batch[1])`?

apis/python/src/tiledbsoma/io/registration/ambient_label_mappings.py
```diff
@@ -0,0 +1,402 @@
...
+        *,
+        measurement_name: str,
+        obs_field_name: str = "obs_id",
+        var_field_name: str = "var_id",
+    ) -> Self:
+        """Extends registration data to one more AnnData input."""
+        tiledbsoma.logging.logger.info("Registration: registering AnnData object.")
+
+        obs_next_soma_joinid = previous.obs_axis.get_next_start_soma_joinid()
+        obs_map = previous.obs_axis.data
```
thetorpedodog, line 271: It looks like you might want to do

```python
obs_map = dict(previous.obs_axis.data)
```

because right now the below is going to end up reaching in and modifying `previous.obs_axis`’s `data` field. (Likewise with other dicts in this factory.)

apis/python/src/tiledbsoma/io/registration/ambient_label_mappings.py
```diff
@@ -0,0 +1,402 @@
...
+                measurement_name=measurement_name,
+            )
+
+        tiledbsoma.logging.logger.info("Registration: complete.")
+        return registration_data
+
+    def show(self) -> None:
+        print(f"obs:{len(self.obs_axis.data)}")
+        for k, v in self.var_axes.items():
+            print(f"{k}/var:{len(v.data)}")
```
thetorpedodog, line 387: Would this make sense as a `__str__` or `__repr__` method that instead returns the contents that you’re printing here?

apis/python/src/tiledbsoma/io/registration/id_mappings.py
```diff
@@ -0,0 +1,82 @@
...
+    """Extracts the label values (e.g. cell barcode, gene symbol) from an AnnData/H5AD
+    ``obs`` or ``var`` dataframe."""
+    if field_name in df:
+        return [str(e) for e in df[field_name]]
+    if field_name == df.index.name:
+        return list(df.index)
+    if df.index.name is None:
+        return list(df.index)
+    if df.index.name == "index":
+        return list(df.index)
```
thetorpedodog, line 81: maybe `if df.index.name in (field_name, "index", None)` to reduce repetition.

thetorpedodog (commented):
A few notes I think I might have missed before. Sorry for losing track of this!

apis/python/src/tiledbsoma/io/registration/__init__.py
johnkerl: @thetorpedodog and I want to defer the leading-underscore, exports, etc -- visibility questions -- until a follow-on PR wherein there _is_ user-facing content (there isn't any yet) and it will be clearer what needs to happen where.

I appreciate and value your feedback. I want to action this particular feedback-point on a follow-on PR.

apis/python/src/tiledbsoma/io/registration/ambient_label_mappings.py
johnkerl: > OK, but this is a module, not a class. And given that this will be exporting public-facing APIs, and it is exporting several classes/functions, it is kind of weird to have "none of it is user-facing API" in the docstring (is it temporary in case we do a release with this included but not finished?). Whatever the goal is is fine, just the docstring and the code itself are sending mixed messages.

@thetorpedodog and I want to defer the leading-underscore, exports, etc -- visibility questions -- until a follow-on PR wherein there _is_ user-facing content (there isn't any yet) and it will be clearer what needs to happen where.

I appreciate and value your feedback. I want to action this particular feedback-point on a follow-on PR.

johnkerl:
@thetorpedodog ready for round three!

apis/python/src/tiledbsoma/io/registration/ambient_label_mappings.py
```diff
@@ -0,0 +1,402 @@
...
+        if experiment_uri is not None and tiledbsoma.Experiment.exists(experiment_uri):
+            registration_data = cls.from_isolated_soma_experiment(
+                experiment_uri,
+                obs_field_name=obs_field_name,
+                var_field_name=var_field_name,
+                context=context,
+            )
+        else:
+            registration_data = cls(
+                obs_axis=AxisAmbientLabelMapping(data={}, field_name=obs_field_name),
```
thetorpedodog, line 364: Optional suggestion: if passing `data={}` in is a thing you do routinely, you could make `data: Dict[...] = attrs.field(factory=dict)` so that by default you get a new empty dict as the value and you don’t need to explicitly provide it if empty.

thetorpedodog (approved):
A couple small, optional suggestions, but this looks good to me overall. May still want to have a domain expert check it over but it looks like it does everything it should and stylistically it looks great.

apis/python/src/tiledbsoma/io/registration/ambient_label_mappings.py
johnkerl: 🔥

johnkerl:
@mojaveazure @aaronwolen thoughts? Again this is just a scaffolding PR, ingest-per-se coming as a follow-up ...

apis/python/src/tiledbsoma/io/registration/ambient_label_mappings.py
```diff
@@ -0,0 +1,403 @@
...
+    class tracks the mapping of input-data ``obs`` or ``var`` ID-column name (barcode ID, gene
+    symbol) to SOMA join IDs for SOMA experiment ``obs`` or ``var``.
+
+    See module-level comments for more information.
+    """
+
+    data: Dict[str, int]
+    field_name: str
+
+    def get_next_start_soma_joinid(self) -> int:
```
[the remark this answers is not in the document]
johnkerl: Hmm, @thetorpedodog I get that

There's a lot of `get_` in SOMA so I'll leave this

But thank you! :)

apis/python/src/tiledbsoma/io/registration/ambient_label_mappings.py
johnkerl: @thetorpedodog thanks!! Awesome to know.

In this case it's highly localized, not a pattern -- so I think I'll leave it as-is

johnkerl:
Happy to take further belated feedback on other PRs on #1278x

PR #1557, opened 2023-07-27
Replace dataclasses with attrs.

thetorpedodog:
This replaces the one (two) remaining use(s) of dataclasses with attrs. Elsewhere we use attrs since it has some minor benefits over dataclasses so this brings consistency. The API for users is completely compatible.

### DIFF
--- a/apis/python/src/tiledbsoma/_query_condition.py
+++ b/apis/python/src/tiledbsoma/_query_condition.py
@@ -7,9 +7,9 @@
 filtering query results on attribute values.
 """
 import ast
-from dataclasses import dataclass, field
 from typing import Any, Callable, List, Tuple, Union
 
+import attrs
 import numpy as np
 import tiledb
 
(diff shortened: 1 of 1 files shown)


johnkerl (approved):
Po-tay-to po-tah-to to me, but, why not :)

johnkerl:
> Elsewhere we use attrs since it has some minor benefits over dataclasses

@thetorpedodog for the record please take a quick moment to summarize those benefits.

thetorpedodog:
> > Elsewhere we use attrs since it has some minor benefits over dataclasses
> 
> @thetorpedodog for the record please take a quick moment to summarize those benefits.

Specifically, [aliases, validators, and converters](https://attrs.org/en/stable/why.html#:~:text=validators%2C%20converters%2C%20equality%20customization%2C%20or%20extensibility%20in%20general), which we use extensively.

Issue #1567, opened 2023-08-02
[r] Prepare for TileDB-SOMA R GA


johnkerl:
## ​​Prerequsites

- [X] ​​Obtain CZI sign-off on proceeding with R general availability release

## ​​Dependencies

* ​​TileDB core 2.16.1
  - [X] ​​Release tag and GitHub artifacts https://github.com/TileDB-Inc/TileDB/releases/tag/2.16.1 — note 2.16.1 has a bugfix which is a release-blocker for TileDB-SOMA R GA
* ​​TileDB-Py 0.22.1
  - [X] ​​Release tag and GitHub artifacts https://github.com/TileDB-Inc/TileDB-Py/releases/tag/0.22.1
  - [X] ​​Artifacts on PyPI [https://pypi.org/project/tiledb/](https://pypi.org/project/tiledb/#history)[#history](https://pypi.org/project/tiledb/#history)
* ​​TileDB-R 0.20.2
  - [X] Release tag and GitHub artifacts https://github.com/TileDB-Inc/TileDB-R/releases/tag/0.20.2
  - [X] Artifacts available on r-universe https://tiledb-inc.r-universe.dev/tiledb -- artifacts not yet available on CRAN as we are one week into a two-week CRAN organizational break that ends August 7th

## ​​Release Preparation

- [X] ​​Merge https://github.com/single-cell-data/TileDB-SOMA/tree/aaronwolen/docs-cleanup to reduce cross-section of exported namespace
- [X] Temporarily change r-universe build cadence to release-only so users can obtain 0.22.1 builds from there until the update is available on CRAN https://github.com/TileDB-Inc/tiledb-inc.r-universe.dev/pull/2
- [X] ​​Depend on  2.16.1, 0.22.1, and 0.20.2 in Python, R, and C++ dependency files: https://github.com/single-cell-data/TileDB-SOMA/pull/1561
- [X] ​​Update CI to install tiledb-r​ from R-universe due to the CRAN hiatus noted above
  - [X] Step 1: https://github.com/TileDB-Inc/tiledb-inc.r-universe.dev/pull/2
  - [X] Step 2: https://github.com/single-cell-data/TileDB-SOMA/pull/1570
- [X] ​​Regenerate docs
- [X] Verify no errors in `R CMD check` locally, and submit to rhub
    - Verifies that builds succeed in different CRAN-like environments — this is an efficient way to test/safety-check the package on a variety environments
[...]

PR #1570, opened 2023-08-02
[r] Temporarily install tiledb-r from r-universe

aaronwolen:
**Issue and/or context:** #1567

**Changes:** This bumps the minimum required version of tiledb-r to 0.20.2. Also modifies the R API's CI to install tiledb-r from r-universe. tiledbsoma 1.4.0 now requires tiledb-r >= 0.20.2 (built against TileDB 2.16.1). We won't be able to upload tiledb-r 0.20.2 to CRAN until they return from break on Aug 7. In the meantime, we can install tiledb-r builds from R-universe, which already has 0.20.2 available.

Thanks to @eddelbuettel for the assist.

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -37,11 +37,13 @@ jobs:
       - name: Bootstrap
         run: cd apis/r && tools/r-ci.sh bootstrap
 
-      - name: Dependencies
-        run: cd apis/r && tools/r-ci.sh install_all
-
       - name: Temporarily install tiledb-r 0.20.2 from testing
         run: sudo Rscript -e 'install.packages("tiledb", repos = c("https://cloud.r-project.org", "https://tiledb-inc.r-universe.dev"))'
+        env:
+          USE_BPSM: FALSE
+
+      - name: Dependencies
+        run: cd apis/r && tools/r-ci.sh install_all
 
       - name: CMake
         uses: lukka/get-cmake@latest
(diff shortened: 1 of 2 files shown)


.github/workflows/r-ci.yml
```diff
@@ -37,11 +37,13 @@ jobs:
       - name: Bootstrap
         run: cd apis/r && tools/r-ci.sh bootstrap
 
-      - name: Dependencies
-        run: cd apis/r && tools/r-ci.sh install_all
-
       - name: Temporarily install tiledb-r 0.20.2 from testing
         run: sudo Rscript -e 'install.packages("tiledb", repos = c("https://cloud.r-project.org", "https://tiledb-inc.r-universe.dev"))'
+        env:
+          USE_BPSM: FALSE
```
eddelbuettel, line 43: That will let you do it in one step but will bring all dependencies of tiledb in 'raw' and not binaries so it will be slower.

aaronwolen: Successful on [macOS](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/5743324145/job/15567348850?pr=1570#step:4:56) but not [ubuntu](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/5743324145/job/15567348680?pr=1570#step:4:118).

eddelbuettel: I may have something better.  Give it a few more minutes to finish in the other repo.

eddelbuettel: See your slack DMs.

eddelbuettel (approved):
Well well well "given that I wrote it" easy for me to say "works for me" 

So big fat :shipit:

PR #1571, opened 2023-08-02
[r] Experiments for r-universe dependencies on Mac

johnkerl:
**Issue and/or context:**

* Issue #1567 
* Attempting to fix https://github.com/r-universe/tiledb-inc/actions/runs/5741370604

**Changes:** As suggested by @eddelbuettel 

**Notes for Reviewer:**

### DIFF
--- a/apis/r/configure
+++ b/apis/r/configure
@@ -10,6 +10,9 @@ if test -z "${R_HOME}"; then
     exit 1
 fi
 
+# Temporary
+${R_HOME}/bin/Rscript tools/get_tarball.R -e 'if (packageVersion("tiledb") != "0.20.2") install.packages("tiledb", repos = c("https://tiledb-inc.r-universe.dev", "https://cloud.r-project.org"))'
+
 ## Check for pkg-config and use it to inquire about tiledb and tiledbsoma build options
 pkg-config --version >/dev/null 2>&1
 if [ $? -eq 0 ]; then
--- a/apis/r/configure
+++ b/apis/r/configure
@@ -11,7 +11,7 @@ if test -z "${R_HOME}"; then
 fi
 
 # Temporary
-${R_HOME}/bin/Rscript tools/get_tarball.R -e 'if (packageVersion("tiledb") != "0.20.2") install.packages("tiledb", repos = c("https://tiledb-inc.r-universe.dev", "https://cloud.r-project.org"))'
+${R_HOME}/bin/Rscript -e 'if (packageVersion("tiledb") != "0.20.2") install.packages("tiledb", repos = c("https://tiledb-inc.r-universe.dev", "https://cloud.r-project.org"))'
 
 ## Check for pkg-config and use it to inquire about tiledb and tiledbsoma build options
 pkg-config --version >/dev/null 2>&1


apis/r/configure
```diff
@@ -10,6 +10,9 @@ if test -z "${R_HOME}"; then
     exit 1
 fi
 
+# Temporary
+${R_HOME}/bin/Rscript tools/get_tarball.R -e 'if (packageVersion("tiledb") != "0.20.2") install.packages("tiledb", repos = c("https://tiledb-inc.r-universe.dev", "https://cloud.r-project.org"))'
```
eddelbuettel, line 14: Take out `tools/get_tarball.R ` here please.

eddelbuettel:
I just realized I can test this in a Docker container and did.  That is of course the Linux case and there again we do get tricked by `bspm` (used by r2u) which under the chosen 'fast' parameterization only looks at best available r2u binary which is 0.20.1 so it does nothing. If we add the extra processing of turning `bspm` off (as we do in CI) then it works and we get 0.20.2 (but see below).

Now, as this (temp) change is only needed for macOS where `bspm` is not active we do not need this.  I think the PR is good.

(Now, on Linux, and for reasons I do not fully understand, R also yelled at me that I used a hardcode path [in essence it checks whether installations are relocatable].  I know that check, it just has never happened to me with tiledb-r on Linux so this is strage. I added a new argument `INSTALL_args="--no-staged-install"` to the `install.packages()` call.  That got me 0.20.2 as desired.)  (Minutes later:  Oh I see. It appears to be a side effect of using cmake during the build. We avoid this otherwise with system libraries and/or artifacts.)

In the container I just got:

```r
> library(tiledb)
TileDB R 0.20.2 with TileDB Embedded 2.16.1 on Ubuntu 22.04.2 LTS.
See https://tiledb.com for more information about TileDB.
> library(tiledbsoma)
> show_package_versions()
tiledbsoma:    0.99.3
tiledb-r:      0.20.2
tiledb core:   2.16.1
libtiledbsoma: libtiledb=2.16.1
R:             R version 4.3.1 (2023-06-16)
OS:            Ubuntu 22.04.2 LTS
> 
```

However, I am having a current side-effect that I do not understand.  I can run `./configure`.  It gets me tiledb 0.20.2 in `/usr/local/lib/R/site-library`.  But if I run `R CMD INSTALL` to follow-up with `tiledbsoma` then the just installed `tiledb` dispappears from `/usr/local/lib/R/site-library`.   That us beyond puzzling.

Now, we are doing this PR chiefly to get macOS build adjusted.  So maybe it is worth testing it on a mac?   My expandeded (with debug print)  snippet in `configure` is now 

```sh
# Temporary
${R_HOME}/bin/Rscript -e 'print(packageVersion("tiledb"))'
${R_HOME}/bin/Rscript -e 'if (packageVersion("tiledb") != "0.20.2") { bspm::disable(); install.packages("tiledb", repos = c("https://tiledb-inc.r-universe.dev", "https://cloud.r-project.org"), INSTALL_opts="--no
-staged-install") }'
${R_HOME}/bin/Rscript -e 'print(packageVersion("tiledb"))'
```

and as the `bspm::disable()` is not needed on macOS we could try

```sh
# Temporary
${R_HOME}/bin/Rscript -e 'print(packageVersion("tiledb"))'
${R_HOME}/bin/Rscript -e 'if (packageVersion("tiledb") != "0.20.2") install.packages("tiledb", repos = c("https://tiledb-inc.r-universe.dev", "https://cloud.r-project.org"), INSTALL_opts="--no-staged-install")'
${R_HOME}/bin/Rscript -e 'print(packageVersion("tiledb"))'
```

Now, I just tried this at the (arm-based) [macOS builder of the R Project](https://mac.r-project.org/macbuilder/submit.html) but sadly it still fails because it appears to check before building and fails. Hard to debug remotely.

Maybe you can try on your macOS machine.

eddelbuettel:
We don't.  The else is an empty null op because macOS does not use `bspm`.

johnkerl:
Sorry @eddelbuettel I already deleted what I wrote as it was in error. Sorry to confuse.

eddelbuettel:
@johnkerl Most excellent as that is what it should do.  


(And could you please also delete the comment-on-the-comment from now 40 mins ago?  Much obliged.)

eddelbuettel (approved):
As suggested, it seems like it is worth a try.   :shipit:

PR #1572, opened 2023-08-03
[r] WIP branch

aaronwolen:
**Issue and/or context:**

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -37,12 +37,6 @@ jobs:
       - name: Bootstrap
         run: cd apis/r && tools/r-ci.sh bootstrap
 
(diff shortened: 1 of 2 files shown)


eddelbuettel:
I was worried this would clash with r2u/bspm and its preference if binary over source but it found the need for source on Ubuntu too.   Did you by chance set the `bspm` variable via `options(bspm.version.check=FALSE)` or does it happen to do the right thing here for us?  If so three cheers for finding an alternative!

aaronwolen:
> Did you by chance set the bspm variable via options(bspm.version.check=FALSE) or does it happen to do the right thing here for us? 

I did not. I wanted to try the simplest approach first and then cross my fingers.

aaronwolen:
Closing in favor of #1571

Issue #1456, opened 2023-06-06
[Bug] [c++/python] Querying DataFrame with pyarrow array coords in multiprocessing returns invalid result


atolopko-czi:
**Describe the bug**

If the `DataFrame.read()` method is called in multiple child processes, where the `coords` arg for each child process is a disjoint _slice_ of a `pyarrow.Array` object, then the result of the `read() ` in each child process is the same, returning the first process's slice of the array in all cases. This does _not_ occur if the slice is first converted to a Python list.

**To Reproduce**

```
import multiprocessing as mp
import os
from concurrent.futures import ProcessPoolExecutor
from typing import Union, List

import pyarrow
import tiledbsoma
from tiledbsoma import Experiment

import cellxgene_census


def read_obs(ids: Union[List[int], pyarrow.Array], test_case_name: str):
    log_prefix = f"[{test_case_name} {os.getpid()}]"

    with cellxgene_census.open_soma() as census:
        exp: Experiment = census['census_data']['homo_sapiens']
        obs_ids_out = exp.obs.read(column_names=['soma_joinid'],
                                   coords=[ids]).concat().to_pandas()['soma_joinid'].to_list()

        obs_ids_in = ids.tolist() if isinstance(ids, pyarrow.Array) else ids
        is_valid = set(obs_ids_out) == set(obs_ids_in)
        print(f"{log_prefix} {'VALID' if is_valid else 'INVALID'} result for test case '{test_case_name}'")
        print(f"{log_prefix} {obs_ids_in=}")
        print(f"{log_prefix} {obs_ids_out=}")


def run_mp(id_slices: List[Union[List[int], pyarrow.Array]], test_case_name: str):
    with ProcessPoolExecutor() as executor:
        for id_slice in id_slices:
            executor.submit(read_obs, id_slice, test_case_name)


if __name__ == '__main__':
    tiledbsoma.show_package_versions()

    num_processes = 3
    num_ids_per_process = 10
    all_ids = range(num_processes * num_ids_per_process)

    mp.set_start_method("spawn")

    all_ids_pyarrow = pyarrow.array(all_ids)
    slices = [all_ids_pyarrow[i * num_ids_per_process:i * num_ids_per_process + num_ids_per_process]
              for i in range(num_processes)]
    run_mp(slices, "pyarrow")

    all_ids_list = list(all_ids)
    slices = [all_ids_list[i * num_ids_per_process:i * num_ids_per_process + num_ids_per_process]
[...]

PR #1573, opened 2023-08-03
[python] Fix For #1456: Invalid Result With PyArrow Array Coords

nguyenv:
**Issue and/or context:**

#1456

The PyArrow `_export_to_c` function sets a pointer to the beginning of the given Arrow array. However, input data may not always point to the start. The coordinates listed in the bug report uses disjointed slices that are offset.

**Changes:**

Directly convert the input Arrow array into a Python list using PyArrow's `Array.tolist`. Then convert the Python list into a C++ vector using Pybind11's `cast`. Pass this into `SOMAArray::set_dim_points`.

The option for passing partitioning indexes to `set_dim_points_arrow` has been removed for simplicity as this is not used anywhere in TileDB-SOMA-Py.

### DIFF
--- a/apis/python/tests/test_experiment_query.py
+++ b/apis/python/tests/test_experiment_query.py
@@ -675,21 +675,26 @@ def test_experiment_query_column_names(soma_experiment):
         assert set(ad.obs.keys()) == {"soma_joinid", "label"}
         assert set(ad.var.keys()) == {"soma_joinid", "label"}
 
+
 @pytest.mark.parametrize("n_obs,n_vars", [(1001, 99)])
-def test_experiment_query_mp_disjoint_arrow_coords(soma_experiment):  
+def test_experiment_query_mp_disjoint_arrow_coords(soma_experiment):
     """
-    Verify Pyarrow join ids that are offset are correctly handled. 
-    """   
-    all_ids = pa.array(range(30))
-    slices = [all_ids[i * 10:i * 10 + 10]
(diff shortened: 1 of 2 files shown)


thetorpedodog (commented):
This looks good code-wise, but could you add a regression test for this change?

apis/python/tests/test_experiment_query.py
```diff
@@ -675,6 +675,21 @@ def test_experiment_query_column_names(soma_experiment):
...
         assert set(ad.var.keys()) == {"soma_joinid", "label"}
 
+@pytest.mark.parametrize("n_obs,n_vars", [(1001, 99)])
+def test_experiment_query_mp_disjoint_arrow_coords(soma_experiment):  
+    """
+    Verify Pyarrow join ids that are offset are correctly handled. 
+    """   
+    all_ids = pa.array(range(30))
+    slices = [all_ids[i * 10:i * 10 + 10]
+              for i in range(3)]
```
thetorpedodog, line 685: It might be easier to understand this by just writing

```
slices = [
    pa.array(range(10)),
    pa.array(range(10, 20)),
    pa.array(range(20, 30)),
]
```

thetorpedodog (approved):
Running `black` (or using `pre-commit`, which will run it for you) will get the formatting consistent and eliminate that ❌ in the lint check. Other than that (and this one suggestion), this looks good.

Issue #1575, opened 2023-08-04
[c++] `SOMAObject` `open` Functions Should Take Start and End Timestamps


nguyenv:
Split-off from https://github.com/single-cell-data/TileDB-SOMA/issues/1240

PR #1576, opened 2023-08-04
[c++] `SOMACollection` Takes Start and End Timestamp

nguyenv:
**Issue and/or context:**

#1575

**Changes:**

* `SOMAGroup` derived objects (ie. `SOMA{Collection,Measurement,Experiment}`) now take in both the start and end timestamps which is consistent with `SOMAArray` derived objects
* Set the `name` (label used by `ManagedQuery` when debugging) of the `SOMAArray` to the URI

### DIFF
--- a/libtiledbsoma/src/soma/soma_array.h
+++ b/libtiledbsoma/src/soma/soma_array.h
@@ -77,7 +77,7 @@ class SOMAArray {
      * @param batch_size Read batch size
      * @param result_order Read result order: automatic (default), rowmajor, or
      * colmajor
-     * @param timestamp Pair indicating timestamp start and end
+     * @param timestamp Optional pair indicating timestamp start and end
      * @return std::unique_ptr<SOMAArray> SOMAArray
      */
     static std::unique_ptr<SOMAArray> open(
@@ -102,7 +102,7 @@ class SOMAArray {
      * @param batch_size Read batch size
      * @param result_order Read result order: automatic (default), rowmajor, or
      * colmajor
-     * @param timestamp Pair indicating timestamp start and end
+     * @param timestamp Optional pair indicating timestamp start and end
      * @return std::unique_ptr<SOMAArray> SOMAArray
      */
     static std::unique_ptr<SOMAArray> open(
--- a/libtiledbsoma/src/soma/soma_array.h
+++ b/libtiledbsoma/src/soma/soma_array.h
@@ -77,6 +77,7 @@ class SOMAArray {
      * @param batch_size Read batch size
      * @param result_order Read result order: automatic (default), rowmajor, or
      * colmajor
+     * @param timestamp Pair indicating timestamp start and end
      * @return std::unique_ptr<SOMAArray> SOMAArray
      */
     static std::unique_ptr<SOMAArray> open(
@@ -101,6 +102,7 @@ class SOMAArray {
      * @param batch_size Read batch size
      * @param result_order Read result order: automatic (default), rowmajor, or
      * colmajor
+     * @param timestamp Pair indicating timestamp start and end
      * @return std::unique_ptr<SOMAArray> SOMAArray
      */
     static std::unique_ptr<SOMAArray> open(
--- a/libtiledbsoma/src/soma/soma_collection.h
+++ b/libtiledbsoma/src/soma/soma_collection.h
@@ -81,12 +81,14 @@ class SOMACollection : public SOMAObject {
      * @param uri URI of the array
      * @param mode read or write
      * @param platform_config Config parameter dictionary
+     * @param timestamp Pair indicating timestamp start and end
      * @return std::shared_ptr<SOMACollection> SOMACollection
      */
     static std::unique_ptr<SOMACollection> open(
         std::string_view uri,
         OpenMode mode,
-        std::map<std::string, std::string> platform_config = {});
+        std::map<std::string, std::string> platform_config = {},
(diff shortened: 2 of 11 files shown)


teo-tsirpanis (approved):

libtiledbsoma/src/soma/soma_dense_ndarray.cc
```diff
@@ -98,7 +98,7 @@ SOMADenseNDArray::SOMADenseNDArray(
     array_ = std::make_shared<SOMAArray>(
         mode,
         uri,
-        "unnamed",  // name
+        uri,  // label used when debugging
```
eddelbuettel, line 101: This value displayed when we do spdlog which may unduly enlarge the logging line. I am wondering if 'somewhere' (here, or on the loggiing side?) we should maybe break this into tokens (ie from, say, s3://some/user/bucket to just bucket) to display something more concise ?  Else we can of course change the default logging format and drop this 'name' field..  Or make it a parameter.  Or keep the old constant.

libtiledbsoma/src/soma/soma_array.h
```diff
@@ -77,6 +77,7 @@ class SOMAArray {
      * @param batch_size Read batch size
      * @param result_order Read result order: automatic (default), rowmajor, or
      * colmajor
+     * @param timestamp Pair indicating timestamp start and end
```
eddelbuettel, line 80: Should we say 'optional'?  Don't we usually signal when a parameter is not required?

libtiledbsoma/src/soma/soma_array.h
```diff
@@ -101,6 +102,7 @@ class SOMAArray {
...
```
eddelbuettel, line 105: Optional?

libtiledbsoma/src/soma/soma_collection.h
```diff
@@ -81,12 +81,14 @@ class SOMACollection : public SOMAObject {
      * @param uri URI of the array
      * @param mode read or write
      * @param platform_config Config parameter dictionary
+     * @param timestamp Pair indicating timestamp start and end
```
eddelbuettel, line 84: Optional?

libtiledbsoma/src/soma/soma_collection.h
```diff
@@ -111,12 +117,13 @@ class SOMACollection : public SOMAObject {
      * @param uri URI of the array
      * @param ctx TileDB context
      * @param key key of the array
+     * @param timestamp Pair indicating timestamp start and end
```
eddelbuettel, line 120: Optional?

libtiledbsoma/src/soma/soma_group.h
```diff
@@ -70,15 +70,15 @@ class SOMAGroup {
      * @param uri URI of the group
      * @param name Name of the group
      * @param platform_config Config parameter dictionary
-     * @param timestamp Timestamp
+     * @param timestamp Pair indicating timestamp start and end
```
eddelbuettel, line 73: Optional?

libtiledbsoma/src/soma/soma_group.h
```diff
@@ -88,15 +88,15 @@ class SOMAGroup {
      * @param ctx TileDB context
      * @param uri URI of the group
      * @param name Name of the group
-     * @param timestamp Timestamp
+     * @param timestamp Pair indicating timestamp start and end
```
eddelbuettel, line 91: Optional?

libtiledbsoma/src/soma/soma_group.h
```diff
@@ -109,14 +109,14 @@ class SOMAGroup {
      * @param uri URI of the group
      * @param name Name of the group
      * @param ctx TileDB context
-     * @param timestamp Timestamp
+     * @param timestamp Pair indicating timestamp start and end
```
eddelbuettel, line 112: Optional?

libtiledbsoma/src/soma/soma_group.h
```diff
@@ -127,9 +127,11 @@ class SOMAGroup {
      * Open the SOMAGroup object.
      *
      * @param mode read or write
-     * @param timestamp Timestamp
+     * @param timestamp Pair indicating timestamp start and end
```
eddelbuettel, line 130: Optional?

eddelbuettel (commented):
This looks like a clean extension.

One question regarding the change from "unnamed" to a potentially much longer uri hiding much of the debugging line info, and one question whether we should flag the timestamp as an optional parameter.

libtiledbsoma/src/soma/soma_dense_ndarray.cc
nguyenv: Makes sense - I will tokenize the filepath and grab just the filename then.

eddelbuettel: That would be a nice net improvement. Screenshot to make the color stand out -- we would then see the key part of URI:

image

libtiledbsoma/src/soma/soma_dense_ndarray.cc
```diff
@@ -95,10 +97,11 @@ SOMADenseNDArray::SOMADenseNDArray(
     std::vector<std::string> column_names,
     ResultOrder result_order,
     std::optional<std::pair<uint64_t, uint64_t>> timestamp) {
+    std::string array_name = std::filesystem::path(uri).filename();
```
eddelbuettel, line 100: Very nice!   The filesystem library is lovely.

eddelbuettel (approved):

PR #1578, opened 2023-08-04
Depend on core 2.16.2, TileDB-Py 0.22.2, and TileDB-R 0.20.3

johnkerl:
**Issue and/or context:** #1567

**Changes:** Micro-bump core and R/Python layers

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -291,7 +291,7 @@ setuptools.setup(
         "scanpy>=1.9.2",
         "scipy",
         "somacore==1.0.4",
-        "tiledb~=0.22.1",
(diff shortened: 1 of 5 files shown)


eddelbuettel (changes requested):
Please add the change in apis/r/configure to fetch 0.20.3 as needed.

johnkerl:
> Please add the change in apis/r/configure to fetch 0.20.3 as needed.
Thanks @eddelbuettel I thought I had, let me double-check

aaronwolen (approved):

eddelbuettel (approved):
Thanks for adding it, when I checked earlier file `configure` was not yet in the set.

PR #1580, opened 2023-08-05
[r] Include system prerequisites in install notes

johnkerl:
Testing installs of 1.4.0 on fresh EC2 instances, and Docker images.

With `cmake` and `git`, I can successfully install with R 4.1 or 4.3.

### DIFF
--- a/apis/r/README.md
+++ b/apis/r/README.md
@@ -14,18 +14,15 @@ install.packages('tiledbsoma', repos = c('https://tiledb-inc.r-universe.dev',
                                           'https://cloud.r-project.org'))
  ```
 
-Installing on Unix-like platforms without binaries requires `cmake` and `git`.
+Installing from source on Unix-like platforms requires `cmake` and `git`.
 
 Alternatively, tiledbsoma can be installed directly from [Conda](https://anaconda.org/tiledb/r-tiledbsoma), which serves binaries for multiple architectures.
 
- ```bash
- mamba install -c conda-forge -c tiledb r-tiledbsoma
- ```
+```bash
+mamba install -c conda-forge -c tiledb r-tiledbsoma
+```
 
 *Note, we're using `mamba` here as a drop-in replacement for `conda` to accelerate the install process.*
-
-The r-universe repo serves macOS binaries and the source package for other Unix-like platforms. The
-conda channel serves binaries for multiple architectures.
 
 ## From source
 
--- a/apis/r/README.md
+++ b/apis/r/README.md
@@ -6,20 +6,26 @@ This is the R implementation of the [SOMA API specification](https://github.com/
 
 ## Release packages
 
-TileDB-SOMA releases are available on R-universe and [Conda](https://anaconda.org/tiledb/r-tiledbsoma), and can be installed directly from R or `mamba` as indicated below.
+TileDB-SOMA releases can be installed from R-universe, which serves macOS binaries and the source
+package for other Unix-like platforms.
 
 ```r
 install.packages('tiledbsoma', repos = c('https://tiledb-inc.r-universe.dev',
-                                         'https://cloud.r-project.org'))
-```
+                                          'https://cloud.r-project.org'))
+ ```
 
-System prerequisites include `cmake` and `git`.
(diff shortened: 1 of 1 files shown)


eddelbuettel:
On the margin I don't think it is needed (beyond what it is in SystemRequirements: already).  We should rather focus on providing artifacts which make the builds both faster and more fail-safe.  

Most users will take macOS binaries anyway.

johnkerl:
@eddelbuettel this is based on simply installing on vanilla Linux hosts and observing that our current install instructions fail unless preceded by installs of `cmake` and `git`. I believe it benefits the out-of-the-box UX for our upcoming GA to simply state this fact clearly, rather than letting users discover the dependencies via a failed installation.

eddelbuettel:
@johnkerl I understand what you wrote and I still stand by my previous message. People who have to rely on cmake and git are people we have otherwise failed and this has been in some of the install notices anyway.  But whatever.

eddelbuettel:
@johnkerl Also the script in https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/r/tools/check_cmake_and_git.R also warns and errors, but sadly not messages do not bubble up.

But my deeper point is that tapering over our somewhat complex installation with messages that users tend not to read is one approach but possibly not the most effective.  So I reiterate the need to focus on _artifacts_,  They really help (for the platforms supported) and users see things "just work".  Let's focus on that, whether or not we a line to docs, and let's ensure we generate a preferable experience that works.

johnkerl:
@eddelbuettel thank you

We will not the binary-artifact setup done today or tomorrow -- as much as it does need to be done, there is more work.

Having a simple, proactive message is a clear UX win for users today, for the setup we have today.

eddelbuettel:
@johnkerl Yes which is why the message is in the file I listed.  

https://github.com/single-cell-data/TileDB-SOMA/blob/a380da222a8b20d48258f2686dd9d058462fd176/apis/r/tools/check_cmake_and_git.R#L3-L14

Going forward we need to ensure to provide the best possible automated installation experience.  Post-hoc adding a line in a README covers our bases but does not meaningfully improve the user experience.

aaronwolen (changes requested):
This is good info that we should add but I think the current presentation has the potential to confuse users. As is, it looks like the system deps are required for conda. How about a slight reordering?

---

TileDB-SOMA releases can be installed from R-universe, which serves macOS binaries and the source package for other Unix-like platforms.

```r
install.packages('tiledbsoma', repos = c('https://tiledb-inc.r-universe.dev',
                                          'https://cloud.r-project.org'))
 ```

Installing on Unix-like platforms without binaries requires `cmake` and `git`.

Alternatively, tiledbsoma can be installed directly from [Conda](https://anaconda.org/tiledb/r-tiledbsoma), which serves binaries for multiple architectures.

 ```bash
 mamba install -c conda-forge -c tiledb r-tiledbsoma
 ```

*Note, we're using `mamba` here as a drop-in replacement for `conda` to accelerate the install process.*

**Comment by johnkerl** [comment]
@aaronwolen requested changes made -- please take a look 🙏

**Review by aaronwolen — APPROVED** [review]
Noted a bit of extra text at the end that should be removed.

File: apis/r/README.md [hunk 1]
```diff
@@ -6,18 +6,26 @@ This is the R implementation of the [SOMA API specification](https://github.com/
...
-TileDB-SOMA releases are available on R-universe and [Conda](https://anaconda.org/tiledb/r-tiledbsoma), and can be installed directly from R or `mamba` as indicated below.
+TileDB-SOMA releases can be installed from R-universe, which serves macOS binaries and the source
+package for other Unix-like platforms.
 
 ```r
 install.packages('tiledbsoma', repos = c('https://tiledb-inc.r-universe.dev',
-                                         'https://cloud.r-project.org'))
-```
+                                          'https://cloud.r-project.org'))
+ ```
```
eddelbuettel, line 15: Tiny bits of accidental whitespace on lines 14 and 15?

johnkerl:
Canceling CIs to free up resources since this is a readme-only PR

PR #1581, opened 2023-08-07
[r] Remove workaround for tiledb-r 0.20.3 while CRAN was on break

eddelbuettel:
**Issue and/or context:**

Versions 0.20.2 and 0.20.3 (current) could not be uploaded to CRAN while CRAN was on break, but provided an update to TileDB Embedded that was needed.  To obtain it, installation relied on a temporary installation via r-universe.  With version 0.20.3 now on CRAN this is no longer needed.

**Changes:**

Removal of a temporary accomation.

**Notes for Reviewer:**

[SC-32602](https://app.shortcut.com/tiledb-inc/story/32602/remove-temporary-cran-accomodation)

### DIFF
--- a/apis/r/configure
+++ b/apis/r/configure
@@ -10,9 +10,6 @@ if test -z "${R_HOME}"; then
     exit 1
 fi
 
-# Temporary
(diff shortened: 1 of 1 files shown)


eddelbuettel:
Thanks @mojaveazure !

Issue #1577, opened 2023-08-04
[c++] `SOMAObject`s need metadata accessible in both read and write mode


nguyenv:
Split-off from https://github.com/single-cell-data/TileDB-SOMA/issues/1240

PR #1584, opened 2023-08-08
[c++] Support reading `SOMAObject` metadata in read and write mode

nguyenv:
**Issue and/or context:**

- #1577
- https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md#somametadatamapping

**Changes:**

* Cache metadata values  when opening the array or group so that it can be accessed in either read or write mode If the `SOMAArray` or `SOMAGroup` is already open in read mode, fill the cache by using `metadata_num` and `get_metadata_from_index`. If it is open in write, open a temp handle in read mode to fill the metadata mapping and then close the temp handle.
* Add metadata functions and accompanying unit tests for `SOMAObject`s
* `get_metadata` checks if key exists and returns `std::nullopt` if DNE

**Notes for Reviewer:**

N/A


eddelbuettel (approved):
LGTM

PR #1592, opened 2023-08-14
[python] Make `tiledbsoma.io._create_or_open_coll` compatible with 1.2.7

johnkerl:
Relates to #1474 and #1548.

PR #1606, opened 2023-08-17
[python] Update to `_create_or_open_coll` for 1.2.7 compatibility

johnkerl:
Testing on #1592 did not cover enough cases.


johnkerl:
[sc-33078]

aaronwolen (approved):

thetorpedodog:
This seems like it’s probably fine, but what is the motivation here?

johnkerl:
@thetorpedodog chatting you on Slack

johnkerl:
[sc-33078]

PR #1627, opened 2023-08-18
[c++] Support reading member-to-uri mapping in read and write mode

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/1623

**Changes:**

Extend caching introduced in #1584 to also store `SOMAGroup` member-to-uri mapping so that this information can also be accessed in both read and write mode.

**Notes for Reviewer:**

N/A


eddelbuettel (approved):
:+1:

PR #1631, opened 2023-08-24
[c++] `SOMAObject` Metadata Methods Should Be Virtual

nguyenv:
**Issue and/or context:**

Follow up PR for https://github.com/single-cell-data/TileDB-SOMA/pull/1584#discussion_r1298880070

**Changes:**

* Correct `SOMAObject` metadata methods to be virtual
* Update documentation for `get_metadata`

### DIFF
--- a/libtiledbsoma/src/soma/soma_collection.h
+++ b/libtiledbsoma/src/soma/soma_collection.h
@@ -367,12 +367,18 @@ class SOMACollection : public SOMAObject {
      * @return MetadataValue (std::tuple<std::string, tiledb_datatype_t,
      * uint32_t, const void*>)
      */
-    std::map<std::string, MetadataValue> get_metadata() {
-        return group_->get_metadata();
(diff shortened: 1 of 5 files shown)


bekadavis9 (approved):

eddelbuettel (approved):
Cannot help but wonder if implementing it in the parent class (as it was) is simpler / easier to maintain than repeating it across the derived classes.  Oh well.

nguyenv:
> Cannot help but wonder if implementing it in the parent class (as it was) is simpler / easier to maintain than repeating it across the derived classes. Oh well.

Ah trust me I would definitely prefer this too. I spent over two days trying to make this work, but it involved doing some really hacky stuff like inheriting from `std::enable_shared_from_this` (which I also learned does not play well with Pybind11) and was still segfaulting by the end of it. I ended up scraping that work and doing this instead.

PR #1639, opened 2023-08-29
[python] Check that `update_obs`/`update_var` do not update row counts

johnkerl:
These should be used to mutate columns only.

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -1124,6 +1124,7 @@ def update_obs(
     _update_dataframe(
         exp.obs,
         new_data,
+        "update_obs",
         context=context,
         platform_config=platform_config,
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1206,6 +1209,26 @@ def _update_dataframe(
...
     )
 
+    with DataFrame.open(
+        sdf.uri, mode="r", context=context, platform_config=platform_config
+    ) as sdf_r:
+
+        # Until we someday support deletes, this is the correct check on the existing,
+        # contiguous soma join IDs compared to the new contiguous ones about to be created.
+        old_jids = sorted(
+            [
```
thetorpedodog, line 1219: no need for the outer `[]` here; you can pass a generator into `sorted`.

thetorpedodog (approved):

Issue #1591, opened 2023-08-14
[r] `update_obs` and `update_var` for R


johnkerl:
Port #1562 and #1588 from R

aaronwolen:
Done in #1643

PR #1643, opened 2023-08-30
[r] Support updating obs and var

aaronwolen:
**Issue and/or context:** #1591 

**Changes:**  Adds `update_obs` and `update_var` methods to the `SOMAExperiment` class. These methods accept either an arrow table, arrow record batch, or a `data.frame` as input and update the existing `SOMADataFrame` on disk. The core logic for these methods is implemented in `SOMADataFrame$update`.

Other changes:

- `SOMADataFrame` attribute filters now default to ZSTD compression if no filter is specified in the `platform_config`, previously we were defaulting to no compression.

- Added new internal utility to create tiledb attributes from arrow fields (`tiledb_attr_from_arrow_field()`), which is now used by `SOMADataFrame$create` and `SOMADataFrame$update`

- `TileDBArray` class now has a private method for reopening an array in a different mode

**Notes for Reviewer:**

### DIFF
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -6,7 +6,7 @@ Description: Interface for working with 'TileDB'-based Stack of Matrices,
     like those commonly used for single cell data analysis. It is documented at
     <https://github.com/single-cell-data>; a formal specification available is at
     <https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md>.
-Version: 1.4.3
+Version: 1.4.3.1
 Authors@R: c(
     person(given = "Aaron",
            family = "Wolen",
--- a/apis/r/R/SOMADataFrame.R
+++ b/apis/r/R/SOMADataFrame.R
@@ -306,7 +306,7 @@ SOMADataFrame <- R6::R6Class(
           is_arrow_schema(schema),
         is.character(index_column_names) && length(index_column_names) > 0,
         "All 'index_column_names' must be defined in the 'schema'" =
-          assert_subset(index_column_names, schema$names, type = "field"),
+          assert_subset(index_column_names, schema$names, "indexed field"),
         "Column names must not start with reserved prefix 'soma_'" =
           all(!startsWith(setdiff(schema$names, "soma_joinid"), "soma_"))
       )
--- a/apis/r/R/SOMADataFrame.R
+++ b/apis/r/R/SOMADataFrame.R
@@ -221,13 +221,16 @@ SOMADataFrame <- R6::R6Class(
     },
 
     #' @description Update (lifecycle: experimental)
-    #' @param values An [`arrow::Table`] or [`arrow::RecordBatch`].
+    #' @param values A `data.frame`, [`arrow::Table`], or
+    #' [`arrow::RecordBatch`].
     update = function(values) {
       private$check_open_for_write()
       stopifnot(
-        "'values' must be an Arrow Table or RecordBatch" =
-          (is_arrow_table(values) || is_arrow_record_batch(values))
+        "'values' must be a data.frame, Arrow Table or RecordBatch" =
+          is.data.frame(values) || is_arrow_table(values) || is_arrow_record_batch(values)
       )
+
+      if (is.data.frame(values)) values <- arrow::as_arrow_table(values)
 
       # Retrieve existing soma_joinids from array to:
       # - validate number of rows in values matches number of rows in array
--- a/apis/r/R/SOMADataFrame.R
+++ b/apis/r/R/SOMADataFrame.R
@@ -230,7 +230,10 @@ SOMADataFrame <- R6::R6Class(
     #'   `values` will be added
     #' - any columns present in both will be left alone, with the exception that
     #'   if `values` has a different type for the column, the entire update
-    #'   will because attribute types cannot be changed
+    #'   will fail because attribute types cannot be changed
+    #'
+    #' Furthermore, `values` must contain the same number of rows as the current
+    #' `SOMADataFrame`.
     #'
     #' @param values A `data.frame`, [`arrow::Table`], or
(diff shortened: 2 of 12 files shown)


apis/r/DESCRIPTION
```diff
@@ -4,7 +4,7 @@ Title: TileDB SOMA
 Description: Interface for working with 'TileDB'-based Stack of Matrices,
     Annotated ('SOMA'): an open data model for representing annotated matrices,
     like those commonly used for single cell data analysis.
-Version: 1.4.4
```
aaronwolen, line 7: Appending a dev number to the previous released version (1.4.3) since we haven't released 1.4.4 yet.

apis/r/R/SOMADataFrame.R
```diff
@@ -225,7 +218,118 @@ SOMADataFrame <- R6::R6Class(
...
+              !row_index_name %in% colnames(values)
+          )
+          values[[row_index_name]] <- rownames(values)
+        }
+        values <- arrow::as_arrow_table(values)
+      }
+
+      # Retrieve existing soma_joinids from array to:
+      # - validate number of rows in values matches number of rows in array
+      # - add original soma_joinids to values if not present
```
johnkerl, line 265: You might consider adding a comment to the effect that this suffices _only_ until the day we start implementing row-deletes

apis/r/R/SOMAExperiment.R
```diff
@@ -27,6 +32,30 @@ SOMAExperiment <- R6::R6Class(
         obs_query = obs_query,
         var_query = var_query
       )
+    },
+
+    #' @description Update the obs [`SOMADataFrame`] to add or remove columns.
```
johnkerl, line 37: Please add a comment saying the row count mustn't change & we'll fail it if they try

apis/r/R/SOMAExperiment.R
```diff
@@ -27,6 +32,30 @@ SOMAExperiment <- R6::R6Class(
...
+    #' See [`SOMADataFrame$update()`][`SOMADataFrame`] for details.
+    #' @param values A `data.frame`, [`arrow::Table`], or
+    #' [`arrow::RecordBatch`].
+    update_obs = function(values, row_index_name = NULL) {
+      self$obs$update(values, row_index_name)
+    },
+
+    #' @description Update the var `SOMADataFrame` to add or remove columns.
```
johnkerl, line 45: Please add a comment saying the row count mustn't change & we'll fail it if they try

apis/r/R/TileDBArray.R
```diff
@@ -15,11 +15,14 @@ TileDBArray <- R6::R6Class(
...
     #' @return The object, invisibly
-    open = function(mode=c("READ", "WRITE"), internal_use_only = NULL) {
+    open = function(mode = c("READ", "WRITE"), internal_use_only = NULL) {
       mode <- match.arg(mode)
       if (is.null(internal_use_only) || internal_use_only != "allowed_use") {
-        stop(paste("Use of the open() method is for internal use only. Consider using a",
-                   "factory method as e.g. 'SOMADataFrameOpen()'."), call. = FALSE)
+        stop(
+          "Use of the open() method is for internal use only.\n",
+          "  - Consider using a factory method as e.g. 'SOMADataFrameOpen()'.",
```
johnkerl, line 23:
```suggestion
          "Consider using a factory method as e.g. 'SOMADataFrameOpen()'.",
```

johnkerl (changes requested):

apis/r/R/SOMAExperiment.R
aaronwolen: Addressed in 99f611bc102c1337f2a3ae43ad1aee509f6b1bde

apis/r/R/SOMAExperiment.R
aaronwolen: Addressed in 99f611bc102c1337f2a3ae43ad1aee509f6b1bde

apis/r/R/TileDBArray.R
aaronwolen: This is intentional and we use this format in other places:

```
Primary message.
  - hint/suggestion 1
  - hint/suggestion 2
```

apis/r/DESCRIPTION
eddelbuettel: FWIW I 'fixed' this too by reverting main to 1.4.3 in one of the PRs from yesterday.

apis/r/R/SOMADataFrame.R
```diff
@@ -225,7 +218,121 @@ SOMADataFrame <- R6::R6Class(
...
+    #' @details
+    #' Update the existing `SOMADataFrame` to add or remove columns based on the
+    #' input:
+    #' - columns present in the current the `SOMADataFrame` but absent from the
+    #'   new `values` will be dropped
+    #' - columns absent in current `SOMADataFrame` but present in the new
+    #'   `values` will be added
+    #' - any columns present in both will be left alone, with the exception that
+    #'   if `values` has a different type for the column, the entire update
+    #'   will fail because attribute types cannot be changed
```
johnkerl, line 233:
```suggestion
    #'   will fail because attribute types cannot be changed.
```

apis/r/R/SOMADataFrame.R
johnkerl: IDK what if there were 10, 0-9, and they deleted number 2, so the old obs has 0-1 and 3-9. But the new obs has 0-4 and 6-9. (Artificial example I know, and we can discuss in the future since it's a moot point for this PR.)

aaronwolen: I'm pulling the old values from the array, not generating new ones. So if joinid 2 had been deleted at some point, I'd retrieve values [0,1,3–9] from the array and insert that vector into a new `soma_joinid` column in values.

PR #1644, opened 2023-08-30
[r] Adjust bounding box to be spec compliant

mojaveazure:
Flatten out the bounding box metadata to be single integer64 values instead of arrays of two integer64 values. Instead of just `<dimname>_domain`, bounding boxes are now stored as `<dimname>_domain_lo` and `<dimname>_domain_hi`

`SOMASparseArray$write()` is now modified to flatten the bounding box before writing and `TileDBArray$used_shape()` is modified to read four metadata values instead of two

The inputs and to `SOMASparseNDArray$write(bbox =)` and `TileDBArray$used_shape()` are unchanged

Context: issue #1445; follow-on to PR #1513.

### DIFF
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -4,7 +4,7 @@ Title: TileDB SOMA
 Description: Interface for working with 'TileDB'-based Stack of Matrices,
     Annotated ('SOMA'): an open data model for representing annotated matrices,
     like those commonly used for single cell data analysis.
-Version: 1.4.3
+Version: 1.4.4
 Authors@R: c(
     person(given = "Aaron",
            family = "Wolen",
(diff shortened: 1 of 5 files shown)


johnkerl (changes requested):
let's please replace `_lo` and `_hi` with `_lower` and `_upper`, which sounds a bit less informal

johnkerl (approved):
🚢

apis/r/DESCRIPTION
```diff
@@ -4,7 +4,7 @@ Title: TileDB SOMA
 Description: Interface for working with 'TileDB'-based Stack of Matrices,
     Annotated ('SOMA'): an open data model for representing annotated matrices,
     like those commonly used for single cell data analysis.
-Version: 1.4.3
+Version: 1.4.4
```
aaronwolen, line 7: @mojaveazure Normally we'd bump the dev version here (`<major>.<minor>.<patch>.<dev>`) and save patch bumps for a release.

PR #1645, opened 2023-08-31
Don't accept lists for Python metadata entries, but on purpose.

thetorpedodog:
We (the royal we) accidentally forgot a return statement in the metadata type-checking function. However, this turned out to be the right thing to do anyway, so ¯\_(°_o)_/¯.

This change removes the superfluous if block entirely and tightens up the types that we do accept to match the SOMA specification:

- primitives
- strings
- nothing else

### DIFF
--- a/apis/python/src/tiledbsoma/_tdb_handles.py
+++ b/apis/python/src/tiledbsoma/_tdb_handles.py
@@ -24,7 +24,6 @@ from typing import (
 )
 
 import attrs
-import numpy as np
 import tiledb
 from somacore import options
 from typing_extensions import Literal, Self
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -353,14 +352,12 @@ def __repr__(self) -> str:
 def _check_metadata_type(key: str, obj: object) -> None:
     """Pre-checks that a metadata entry can be stored in an array.
 
-    These checks are reproduced from the TileDB Python metadata-setting methods.
-    We have to do this since we don't commit our changes until the very end.
+    These checks are reproduced from the TileDB Python metadata-setting methods,
+    but are slightly more restrictive than what TileDB allows in general.
```
johnkerl, line 356:
```suggestion
    but are slightly more restrictive than what TileDB allows in general. In particular, TileDB disallows arrays
    of string within metadata keys while allowing arrays of int. The SOMA spec, however, disallows arrays
    of any kind. Please also see
    https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/python/devtools/ingestor#L42-L46
```

johnkerl (changes requested):

apis/python/src/tiledbsoma/_tdb_handles.py
thetorpedodog: went with slightly different wording but the same general info

johnkerl (approved):
🚢 🎉

PR #1636, opened 2023-08-28
[r] Add a finalizer for the SOMAArray instance and assure valgrind tests pass

eddelbuettel:
**Issue and/or context:**

The SOMAArray instance used by the iterated reader contains a shared pointer to a Context instance.  This PR adds a finalizer to ensure this is properly unwould.  Unit tests are refactored to test under `valgrind` and an actions has been added.

**Changes:**

Addiition of a finalizer function, refactored tests and added valgrind action

**Notes for Reviewer:**

[SC 33353](https://app.shortcut.com/tiledb-inc/story/33353/added-finalizers-as-well-test-restructuring-for-valgrind-compliance) and the notion doc referenced therein

PR #1657, opened 2023-09-01
[r] Refactor iterated reader lifetime

eddelbuettel:
**Issue and/or context:**

PR #1636 has been seen "to work in theory yet not in practice" so the lifetime management of the iterated reader and its SOMArray object has been refactored.

**Changes:**

A struct is wrapped around the shared pointer which allows for R lifetime management (of an external pointer to the struct) to coexist with the C++ shared_pointer (inside the struct).  No more explicit finalizer.

**Notes for Reviewer:**

[SC 33659](https://app.shortcut.com/tiledb-inc/story/33659/r-refactor-iterated-reader-lifetime)

### DIFF
--- a/apis/r/R/RcppExports.R
+++ b/apis/r/R/RcppExports.R
@@ -91,10 +91,6 @@ sr_next <- function(sr) {
     .Call(`_tiledbsoma_sr_next`, sr)
 }
 
-sr_finalize <- function(sr) {
-    invisible(.Call(`_tiledbsoma_sr_finalize`, sr))
-}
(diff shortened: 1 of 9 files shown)


johnkerl (approved):
Works on my Mac with `export CI=true` and `testthat::test_local("tests/testthat")` 🚀

eddelbuettel:
Yes, it is good to 'see' the summaries per test file so `cd tests; CI=true R -q --no-save < testthat.R` is handy too:

image

Issue #1550, opened 2023-07-22
[python] High default thread allocation


bkmartinjr:
I noticed that some of our Census code had very high thread counts.  Roughly 650 threads per process in normal usage.  We use a lot of worker processes, so the thread counts were getting very high on high-CPU count boxes (18,000+ threads for just one application on a 128 CPU box).

Digging into it, it appears that tiledbsoma allocates a large number of threads at the time of package import, even in the case where the user does not utilize the default TileDB context.  The TileDB-Py package on the other hand has a much lower thread allocation for the same operations.

Both packages allocate the same number of threads when creating a context.  However, `import tiledbsoma` also creates approximately 4 threads per CPU (before any use of the package), and they appear to be retained even when not used.  Because the Census package _alwasys_ creates a context (for AWS config of the bucket region) we end up with a large number of unused/overhead threads.

Test code below.  Running it on an EC2 instance which use use for our Census builds yields:
* at import of tiledbsoma - 393 threads allocated per process
* after open of an Array, the number increases to 649
Whereas using TileDB-Py, the numbers are roughly 1/2 of that.

Example run on a high-CPU-count box, with essentially equivalent operations performed using the two packages:
```
$ python tmp/threads.py tiledb
---TileDB-Py---
at the beginning: pid=2159529, cpus=128 number of tasks=1
after import tiledb: pid=2159529, cpus=128 number of tasks=73
after open array with tiledb: pid=2159529, cpus=128 number of tasks=330
after array close and gc: pid=2159529, cpus=128 number of tasks=74

$ python tmp/threads.py tiledbsoma
---TileDB-SOMA---
at the beginning: pid=2159914, cpus=128 number of tasks=1
after import tiledbsoma: pid=2159914, cpus=128 number of tasks=393
after create context and open collection: pid=2159914, cpus=128 number of tasks=649
after del variables and gc: pid=2159914, cpus=128 number of tasks=393
```

Numbers are lower on boxes with lower number of CPUs as you would expect.

Code (may only work on Linux, as it uses `/proc/`):
```
import gc
import os
import sys


def print_my_thread_count(msg=""):
    pid = os.getpid()
    tasks = os.listdir(f"/proc/{pid}/task")
    print(f"{msg}: pid={pid}, cpus={os.cpu_count()} number of tasks={len(tasks)}")


TDB_CONFIG = {
    "vfs.s3.region": "us-west-2",
    "vfs.s3.no_sign_request": "true",
}
[...]

PR #1661, opened 2023-09-05
Lazily initialize a default global TileDB context.

thetorpedodog:
We use the basic `functools.cache` decorator to ensure one default Ctx is created.

This avoids immediately constructing a Ctx upon importing tiledbsoma, and means that if a user only ever passes in their own context into SOMATileDBContext constructors, the default is never initialized.

Helps address #1550. This change is convenient because it imposes no user API changes; everything Just Works.


bkmartinjr (approved):
Oh yea!  This is great!

PR #1664, opened 2023-09-07
Initialize TileDB Contexts extremely lazily.

thetorpedodog:
To allow users to mess with their TileDB configs and modify them without actually instantiating an actual TileDB Context (which entails allocation of significant resources), we lazily wait until the caller actually requests the Context to create it just in time. This ensures that we don't waste resources that are not easy to reclaim (particularly the shared global context, which is never GCable).

Further addresses #1550. It may fix it; I leave it to @bkmartinjr et al to see if it completely addresses their issues once this change is deployed.

### DIFF
--- a/apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
+++ b/apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
@@ -201,19 +201,25 @@ class SOMATileDBContext:
         Args:
             tiledb_config:
                 A dictionary of parameters for `tiledb.Config() <https://tiledb-inc-tiledb.readthedocs-hosted.com/projects/tiledb-py/en/stable/python-api.html#config>`_.
+                To remove a parameter from the existing config, provide ``None``
+                as the value.
             tiledb_ctx:
                 A TileDB Context to replace the current context with.
             timestamp:
                 A timestamp to replace the current timestamp with.
                 Explicitly passing ``None`` will remove the timestamp.
+                For details, see the description of ``timestamp``
+                in :meth:`__init__`.
 
         Lifecycle:
             Experimental.
 
         Examples:
-            >>> context.replace(timestamp=0)
-
-            >>> context.replace(tiledb_config={"vfs.s3.region": "us-east-2"})
+            >>> context.replace(timestamp=1_512_658_800_000)  # UNIX millis
+            >>> new_region_context = context.replace(
+            ...     tiledb_config={"vfs.s3.region": "us-east-2"})
+            >>> back_to_default_context = new_region_context.replace(
+            ...     tiledb_config={"vfs.s3.region": None})
         """
         with self._lock:
             if tiledb_config is not None:
@@ -222,10 +228,9 @@ class SOMATileDBContext:
                         "Either tiledb_config or tiledb_ctx may be provided"
                         " to replace(), but not both."
                     )
-                new_config = dict()
                 new_config = self._internal_tiledb_config()
                 new_config.update(tiledb_config)
-                tiledb_config = new_config
+                tiledb_config = {k: v for (k, v) in new_config.items() if v is not None}
             if timestamp is _SENTINEL:
                 # Keep the existing timestamp if not overridden.
                 timestamp = self._timestamp_ms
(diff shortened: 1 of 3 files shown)


apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -101,11 +215,23 @@ def replace(
...
-            changes["tiledb_ctx"] = tiledb.Ctx(config=new_config)
-        return attrs.evolve(self, **changes)
+        with self._lock:
+            if tiledb_config is not None:
+                if tiledb_ctx:
+                    raise ValueError(
+                        "Either tiledb_config or tiledb_ctx may be provided"
+                        " to replace(), but not both."
+                    )
+                new_config = dict()
```
bkmartinjr, line 225: typo?  I think you want to delete this line in favor of the next line

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -101,11 +215,23 @@ def replace(
...
+                new_config = self._internal_tiledb_config()
+                new_config.update(tiledb_config)
```
bkmartinjr, line 227: I'm not sure if this is intentional, but....
* if I pass `tiledb_config` to `__init__`, the user-provided config entirely replaces the default config
* if I pass `tiledb_config` to `replace`, it simply updates the existing config

I can see cases for both, but I am wondering if they should not be consistent?

Also, how do I remove a config option using `replace`?

bkmartinjr (commented):
this is looking good.  Comments:
1. the `tiledb_config` handling - i.e., does it have _update_ or _replace_ behavior - seem different between the constructor and the replace function.  Replace seems to do an _update_.   Given that the use can always access the current config (via `.tiledb_config`), I wonder if both `__init__` and `replace` should not have "replace" semantics?  This would involve only a change to the `replace` method as I understand the code.
2. I believe this will resolve most of my thread pool over-allocation issues, at least those that are solvable at the Python level (we still have the issue of duplication of DB instances to be fixed in libtiledbsoma).
3. Is there an equivalent issue to resolve in the R lib?

Leaving this as a comment until question (1) is resolved.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
thetorpedodog: Whoops! This is a remnant of a previous version (where `_internal_tiledb_config` was not guaranteed to return a fresh dict).

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
thetorpedodog: In both cases this would be equivalent, right?

- When passing the config to `__init__`, there is (almost) no existing config to replace. It will overwrite any entry of the `_default_config` that overlaps (technically, the one that overlaps).
- When passing it to `replace`, it does the same thing, but the base configuration is that of the existing context object.

> Also, how do I remove a config option using `replace`?

This wasn’t supported in the existing API, but passing `None` as a value seems like as good an idea as any.

thetorpedodog (commented):
I did the think where I saw the comments, read them, intended to respond, and then apparently decided in my brain that I had already done so and here we are now.

johnkerl (approved):

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
bkmartinjr: > When passing the config to __init__, there is (almost) no existing config to replace. It will overwrite any entry of the _default_config that overlaps (technically, the one that overlaps).

With `__init__`, this approach will not work for any key in the default config, but not specified by the user.

I had assumed that the default config was "best practices", and that we would want to retain this config for any context where the user does not specifically overwrite it. Consider a case where the user is only creating a context because they want to set the AWS S3 region - in this case, I argue they want to inherit the default configuration and simply update (add to it, i.e., add the S3 region config param).

If this isn't true, then I'm unclear why we ever have a default configuration in the code...

So, assuming we have defaults for a good reason, then I think we need to follow one of the following options:

a) Keep code as is, but add public access to the default config, so that I can do:

      > ctx = SOMATileDBContext(SOMATileDBContext.default_config.update({'vfs.s3.region': 'us-west-1'}))

b) Or, have `__init__` do the above by default (i.e., use the merge of the default config and user-provided config), and accept keys with a `None` value when I want to nuke them.

In either case, supporting the "delete `None` valued keys" seems smart.

bkmartinjr (approved):

PR #1668, opened 2023-09-11
[c++] Don't use Python to generate C++ test inputs

johnkerl:
**Issue and/or context:** Found in #1511 / #866 et al. while attempting pre-tag/pre-merge tests with core 2.17.0-rc0. Here we use `tiledbsoma-py` to generate test data for `libtiledbsoma` code. That's currently failing due to `tiledbsoma-py` in the generator step being built with core 2.16 rather than 2.17. But this highlights a bigger issue: let's not be circular in our test paths. Rather, let's validate C++ code using C++ and shell tools; then move up the testing stack to Python and R.

**Changes:** Simply can the generated inputs into `test/soco.tgz`. As an added bonus we get some backward-compatilbility test here.

**Notes for Reviewer:**

### DIFF
--- a/.github/workflows/python-ci-single.yml
+++ b/.github/workflows/python-ci-single.yml
@@ -133,6 +133,12 @@ jobs:
         ls -l test
         echo ls -l test/soco
         ls -l test/soco
+        echo ls -l /home/runner/work/TileDB-SOMA/TileDB-SOMA/libtiledbsoma/test/../../test/soco
+        ls -l /home/runner/work/TileDB-SOMA/TileDB-SOMA/libtiledbsoma/test/../../test/soco
+        echo ls -l /home/runner/work/TileDB-SOMA/TileDB-SOMA/libtiledbsoma/test/../../test/soco/pbmc3k_processed
+        ls -l /home/runner/work/TileDB-SOMA/TileDB-SOMA/libtiledbsoma/test/../../test/soco/pbmc3k_processed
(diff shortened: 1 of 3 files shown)


.github/workflows/python-ci-single.yml
```diff
@@ -103,20 +103,29 @@ jobs:
...
-    - name: Generate test data
+    - name: Obtain test data
       shell: bash
       run: |
-        mkdir -p test/soco
-        ./apis/python/devtools/ingestor --debug --soco -o test/soco -n data/pbmc3k_processed.h5ad data/10x-pbmc-multiome-v1.0/subset_100_100.h5ad
+        rm -rf test/soco
+        mkdir -p test
+        cd test
+        tar zxf soco.tgz
```
eddelbuettel, line 112: cd'ing into a director you just made and then expanding a file you expect in the directory you just made may not work, or I am behind on what types of wheaties I shoud have for breakfast.

johnkerl: @eddelbuettel it does indeed work (witness CI passes)

The `mkdir -p` is unnecessary as `test/soco.tgz` is under source control

eddelbuettel: I see, so `mkdir -p test` tricked me as it is "empty" as `test/soco.tgz` exists at the same time.  Still reads weird.

But that is not real worry then.  Thanks for the cluebat.

PR #1676, opened 2023-09-14
[python] Update default-filter-list handling in unit tests

johnkerl:
For #866 / core 2.17

### DIFF
--- a/apis/python/tests/test_platform_config.py
+++ b/apis/python/tests/test_platform_config.py
@@ -66,7 +66,11 @@ def test_platform_config(adata):
             assert x_arr.attr("soma_data").filters == [tiledb.NoOpFilter()]
             assert x_arr.dim("soma_dim_0").tile == 6
             assert x_arr.dim("soma_dim_0").filters == [tiledb.ZstdFilter(level=2)]
-            assert x_arr.dim("soma_dim_1").filters == []
+            # As of 2.17.0 this is the default when empty filter-list, or none at all,
+            # is requested. Those who want truly no filtering can request a no-op filter.
+            assert list(x_arr.dim("soma_dim_1").filters) == [
+                tiledb.ZstdFilter(level=-1)
+            ]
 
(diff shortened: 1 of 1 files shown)


apis/python/tests/test_platform_config.py
```diff
@@ -66,7 +66,11 @@ def test_platform_config(adata):
             assert x_arr.attr("soma_data").filters == [tiledb.NoOpFilter()]
             assert x_arr.dim("soma_dim_0").tile == 6
             assert x_arr.dim("soma_dim_0").filters == [tiledb.ZstdFilter(level=2)]
-            assert x_arr.dim("soma_dim_1").filters == []
+            # As of 2.17.0 this is the default when empty filter-list, or none at all,
+            # is requested. Those who want truly no filtering can request a no-op filter.
+            assert list(x_arr.dim("soma_dim_1").filters) == [
+                tiledb.ZstdFilter(level=-1)
+            ]
```
nguyenv, line 73: I don't think casting to `list` is necessary.

```suggestion
            assert x_arr.dim("soma_dim_1").filters == [tiledb.ZstdFilter(level=-1)]
```

nguyenv (approved):

apis/python/tests/test_platform_config.py
johnkerl: @nguyenv it is indeed -- without that, it's `list` on one side and `FilterList` on the other. Without the cast, the test fails.

nguyenv: We should be able to handle checking equality between `FilterList` and any container type.

```
>>> import tiledb
>>> tiledb.FilterList([]) == []
True
>>> tiledb.FilterList([tiledb.ZstdFilter()]) == [tiledb.ZstdFilter()]
True
>>> tiledb.FilterList([tiledb.ZstdFilter()]) == (tiledb.ZstdFilter(),)
True
```

PR #1559, opened 2023-07-28
[r] Add enumeration support 

eddelbuettel:
**Issue and/or context:**

As described in #1558 and #866, adding enumeration support is desirable once we have TileDB Embedded 2.17 available

**Changes:**

This PR supports reading of columns with enumerations (aka dictionaries aka factor variable) directly via Arrow. Preliminary write support is also available (but still goes through the `tiledb` R package for writes).

**Notes for Reviewer:**

~This PR is now work-in-progress and not ready for a merge while we await TileDB 2.17.~  The branch and PR are ready but should only be merged once prequisites are been merged.  It likely needs #1519 (C++ side) and #1663 (CI support).

CI is turned off as the TileDB default build is still without support for enumerations.

PR #1677, opened 2023-09-14
[c++] Depend on core 2.17.0

johnkerl:
**Issue and/or context:** For #866. The "real" PRs are #1519, #1511, and #1559 which are at the moment depending on core 2.17.0-rc0. On this PR, if CI passes, I'll attempt to first merge solely the core-dependency change -- and then the big three PRs will be able to be rebased on top of this once it's merged.

**Changes:**

* Core 2.17.0
* #1676 redux 
* TileDB-Py 0.23.0
* TileDB-R 0.21.0

**Notes for Reviewer:**

### DIFF
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -42,7 +42,7 @@ Imports:
     Matrix,
     stats,
     bit64,
-    tiledb (>= 0.20.3),
+    tiledb (>= 0.21.0),
     arrow,
     utils,
     fs,
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -291,7 +291,7 @@ setuptools.setup(
         "scanpy>=1.9.2",
         "scipy",
         "somacore==1.0.4",
-        "tiledb~=0.22.2",
+        "tiledb~=0.23.0",
         "typing-extensions",  # Note "-" even though `import typing_extensions`
     ],
     extras_require={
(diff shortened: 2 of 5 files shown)


apis/r/DESCRIPTION
```diff
@@ -42,7 +42,7 @@ Imports:
     Matrix,
     stats,
     bit64,
-    tiledb (>= 0.20.3),
+    tiledb (>= 0.21.0),
```
eddelbuettel, line 45: As that does not exist yet in repo, it would break CI etc.

johnkerl: yup, ready and waiting for it to exist

johnkerl: indeed! :)

eddelbuettel: We could consider not imposing the constraint as it will break.

We can ensure in CI etc that 0.21.0 is present.  Fundamentally there is a difference between supplying newer version in the helper scripts such as `get_tarball.R` and imposing them to be present.

eddelbuettel (approved):
This is internally consistent in ensuring "2.17.0 and its resulting API packages shall be used".

But the practical problem is that we don't have them yet as binaries.  Not at Conda (feedstock issue), not at r2u.   We do have them at r-universe and as source at CRAN.  Close but is it close enough?

johnkerl:
Conda is for later in our release procedure:
https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases

Green CI matters for this PR.

This Python/R interop which has been solidly quietly green for months is now vexing me.

johnkerl:
I'll merge this PR as it fixes some issues. But Python/R interop remains to be solved.

johnkerl:
I'll debug the interop first thing tomorrow.

PR #1678, opened 2023-09-14
[python] 1676 redux

johnkerl:
This is the same as #1676, but applied to `main` rather than stacked atop #1519.

Issue: #866


johnkerl:
Essentially a dup of #1677. Closing as such.

PR #1679, opened 2023-09-15
Interop canary [WIP]

johnkerl:
**Issue and/or context:**

**Changes:**

**Notes for Reviewer:**

PR #1680, opened 2023-09-15
[c++] Removed unused-variable warning

johnkerl:
A nit, but one easy to neaten


johnkerl:
R/Python interop CI on #1679 which is CI-green and about to be merged

PR #1682, opened 2023-09-15
[ci] Have clang-format show version

johnkerl:
Internal info only


eddelbuettel (approved):
Not sure I am a fan of this.  I would run the CI script in Docker under the same container. 

But if you must ...

PR #1683, opened 2023-09-15
Update `write_soma.data.frame()` to support categoricals

mojaveazure:
Allow `write_soma.data.frame()` to write factors as categoricals. `write_soma()` will no longer cast factors to characters prior to writing


eddelbuettel (approved):
:+1:

johnkerl (approved):

PR #1685, opened 2023-09-15
[ci] Debug MacOS `cmake` error

johnkerl:
Debugging https://github.com/single-cell-data/TileDB-SOMA/actions/runs/6203245198/job/16843554651 which is post-merge CI fail

Also note that some time ago we found hard (non-scalable) limits on simultaneous MacOS CI and so limited that pre-commit (as a calculated tradeoff), as at the time we were choked on CI. See also #693 and #750.

### DIFF
--- a/.github/workflows/python-ci-minimal.yml
+++ b/.github/workflows/python-ci-minimal.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        os: [ubuntu-22.04]
+        os: [ubuntu-22.04, macos-12]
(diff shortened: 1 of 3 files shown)


johnkerl:
Comparing previous success
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/6200120705/job/16834110163
and recent fail
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/6203434073/job/16844023067?pr=1685
I had thought perhaps cmake versions were the issue. But the old/green gets 3.27 no probem.

johnkerl:
Superseded by #1688

PR #1686, opened 2023-09-16
[r] Update `write_soma.data.frame()` to support categoricals

mojaveazure:
Allow `write_soma.data.frame()` to write factors as categoricals. `write_soma()` will no longer cast factors to characters prior to writing
Redo of #1683 since it got removed from #1559

### DIFF
--- a/apis/r/R/write_soma.R
+++ b/apis/r/R/write_soma.R
@@ -93,6 +93,7 @@ write_soma.data.frame <- function(
   relative = TRUE
 ) {
   stopifnot(
+    "'x' must be named" = is_named(x, allow_empty = FALSE),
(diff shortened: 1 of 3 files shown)


eddelbuettel (approved):
Should we add a simple test for round-turn?

aaronwolen (approved):
Thank you! I'll second Dirk's request for a unit test but otherwise looks good to go.

johnkerl (approved):

johnkerl (changes requested):
Let's get the unit-test case in

johnkerl (approved):
🚢 
Thanks @mojaveazure !!

PR #1687, opened 2023-09-16
Pass user configuration/timestamp to query condition init

Shelnutt2:
The query condition init needs the array schema and the enumeration values. Fetching the enumeration values requires opening the array. This was previously changed but the opening of the array did not properly consider the the user configurations or timestamps. This change passes the user config and timestamps to the array open in python inside the query condition.

The overall design needs significant improvement. Opening the array inside the query condition is likely to cause other issues as now there are additional context created and additional performance impacts of opening the array and closing it after enumerations are fetched.

It is left to a followup pull request to change the design to a sustainable one that users enumerations from C++, in a similar way to the previous passing of array schemas which avoided multiple array opens.

### DIFF
--- a/apis/python/src/tiledbsoma/_query_condition.py
+++ b/apis/python/src/tiledbsoma/_query_condition.py
@@ -7,7 +7,7 @@
 filtering query results on attribute values.
 """
 import ast
-from typing import Any, Callable, List, Tuple, Union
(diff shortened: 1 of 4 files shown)


eddelbuettel:
I would _love_ for us to do this in a way that is not specific to Python.

ihnorton (approved):

johnkerl (approved):

PR #1690, opened 2023-09-16
[python] Consolidation and vacuuming are now platform configuration options

Shelnutt2:
**Issue and/or context:**

Consolidation and vacuuming need to be controllable by the user, and should default at the low level to off. In higher level APIs we need to follow up with automatic handling.

**Changes:**

Commit and fragment_metadata consolidation and vacuuming can improve the opening and query performance of SOMA experiments. Vacuuming requires slight coordination though and should not happen by default. Instead a platform config allows the user to control these operations based. This will be expanded to defaults for top-level `io` packages where its more likely a user is doing a one-shot ingestion and will want automatic handling.

A new platform config, `consolidate_and_vacuum` has been added which is a boolean to handle this behavior.

**Notes for Reviewer:**

### DIFF
--- a/.github/workflows/python-ci-minimal.yml
+++ b/.github/workflows/python-ci-minimal.yml
@@ -29,6 +29,7 @@ jobs:
       python_version: ${{ matrix.python-version }}
       cc: ${{ matrix.cc }}
       cxx: ${{ matrix.cxx }}
+      is_mac: ${{ contains(matrix.os, 'macos') }}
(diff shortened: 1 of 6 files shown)


bkmartinjr (approved):
LGTM.  May need a unit test, at your discretion.

johnkerl:
CI fails are orthogonal and are tracked by the still-open https://github.com/single-cell-data/TileDB-SOMA/pull/1692. Merging this PR and monitoring its post-merge `python-ci-full`.

PR #1691, opened 2023-09-17
[python] Categorical-write corner-case

johnkerl:
**Issue and/or context:** As discussed with @bkmartinjr. Parent context: #866.

**Changes:** When schema indicates non-categorical, permit auto-flattening write of categorical dataframe data in order to smooth the transition from tiledbsoma-without-categories to tiledbsoma-with-categories.

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -419,12 +419,9 @@ class DataFrame(TileDBArray, somacore.DataFrame):
                 else:
                     # Schema is non-categorical but the user is writing categorical.
                     # Simply decategoricalize for them.
-                    pvalues = []
-                    for chunk in col.chunks:
-                        pvalues += [
-                            chunk.dictionary[i.as_py()].as_py() for i in chunk.indices
-                        ]
-                    cols_map[name] = pd.Series(pvalues)
+                    cols_map[name] = pa.chunked_array(
+                        [chunk.dictionary_decode() for chunk in col.chunks]
+                    )
             else:
                 cols_map[name] = col.to_pandas()
 
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -412,9 +412,22 @@ class DataFrame(TileDBArray, somacore.DataFrame):
             n = len(col)
             cols_map = dim_cols_map if name in dim_names_set else attr_cols_map
             if pa.types.is_dictionary(col.type) and col.num_chunks != 0:
-                cols_map[name] = col.chunk(0).indices.to_pandas()
+                attr = self._handle.schema.attr(name)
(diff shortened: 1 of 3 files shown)


johnkerl:
CI fail
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/6214041643/job/16865628212?pr=1691
```
__ZN10tiledbsoma8LOG_INFOERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
```
demangles to
```
_tiledbsoma::LOG_INFO(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)
```

That is not an error to be ignored, of course, but it is also orthogonal to the changes on this PR.

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -412,9 +412,22 @@ def write(
...
+                if attr.enum_label is not None:
+                    # Normal case: writing categorical data to categorical schema.
+                    cols_map[name] = col.chunk(0).indices.to_pandas()
+                else:
+                    # Schema is non-categorical but the user is writing categorical.
+                    # Simply decategoricalize for them.
+                    pvalues = []
+                    for chunk in col.chunks:
+                        pvalues += [
+                            chunk.dictionary[i.as_py()].as_py() for i in chunk.indices
```
bkmartinjr, line 425: Apologies if I'm misreading this, but this appears to be iterating over _every_ element in the column in a Python loop.  That is going to be unbelievably slow and use a great deal of memory for anything other than very small Arrow arrays.

There are built-ins for doing this.  I believe you want (please confirm with tests) to replace the entire thing with:

> cols_map[name] = pa.chunked_array([chunk.dictionary_decode() for chunk in col.chunks])

apis/python/tests/test_dataframe.py
```diff
@@ -848,6 +849,9 @@ def test_write_categorical_types(tmp_path):
                 "B": pd.Categorical(
                     ["a", "b", "a", "b"], ordered=False, categories=["b", "a"]
                 ),
+                "C": pd.Categorical(
+                    ["a", "b", "a", "b"], ordered=False, categories=["a", "b"]
+                ),
```
bkmartinjr, line 854: this needs to also test categoricals that contain primitives other than string. Commonly see int and bool in real-world data.

bkmartinjr (changes requested):
Suggested an improved approach to flattening the dictionary array, which should use less memory and be dramatically faster.

johnkerl:
Orthogonal/pre-existing CI issue as noted here:
https://github.com/single-cell-data/TileDB-SOMA/pull/1691#issuecomment-1722487891

Unit tests passing on MacOS laptop as well as on an Ubuntu EC2 instance.

bkmartinjr (approved):
LGTM

PR #1692, opened 2023-09-17
[ci] CI canary [WIP/RFC]

johnkerl:
Internal-only/debug-only canary to examine CI issues arising simultaneously with #1691, while apparently not caused by #1691.

Our post-merge CI has fail-fast set to false -- run all tests to completion -- while our pre-merge CI has fail-fast set to true. This is, generally, a good idea as a large chunk of our CI compute resources are absorbed by pre-merge CI. However, on #1691 I want to see if the Python 3.7 fail there is related _only_ to 3.7. As is on #1691, with fail-fast set to true, the 3.7 fail is cancelling signal from the other CI jobs.

Parent context: #866.

### DIFF
--- a/.github/workflows/python-ci-minimal.yml
+++ b/.github/workflows/python-ci-minimal.yml
@@ -15,7 +15,8 @@ on:
 jobs:
   build:
     strategy:
-      fail-fast: true
+      #fail-fast: true
+      fail-fast: false
       matrix:
         os: [ubuntu-22.04, macos-12]
         python-version: ['3.10', '3.7']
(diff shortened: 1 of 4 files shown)


johnkerl:
This is progress: https://github.com/single-cell-data/TileDB-SOMA/pull/1692/commits/9a08fcc4b99a69168593b5e1a19d60631d783293

Before: an error involving

```
__ZN10tiledbsoma8LOG_INFOERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
```

which demangles to
```
_tiledbsoma::LOG_INFO(
    std::__cxx11::basic_string<
        char,
        std::char_traits<char>,
        std::allocator<char>
    > const&
)
```

-- note the `const` _after_ the string bit, and right before the `&`. We do have another error; but I believe this has fixed one of the two issues.

cc @ihnorton @Shelnutt2

Issue #1703, opened 2023-09-19
[python] SparseNDArray read returns incorrect result when one dimension is empty-indexed


bkmartinjr:
The SparseNDArray.read method returns incorrect query results when the `coords` provided are a unique combination of:
* an empty array for dim_0
* and a non-empty array for dim_1

Concrete test case below.  All three cases _should_ identical results (namely, an empty pa.Table). The third case, which should be identical, instead returns _all_ data in the SparseNDArray. Bug was found in the 1.5 pre-release (2.17), but it may exist in previous versions as well.

Bug was found via incorrect ExperimentAxisQuery.to_anndata results, but I narrowed it down to the SparseNDArray.read behavior;

```python
import tiledbsoma as soma
import pyarrow as pa


def main():
    ctx = soma.SOMATileDBContext(
        tiledb_config={
            "vfs.s3.region": "us-west-2",
            "soma.init_buffer_bytes": 4 * 1024**3,
        }
    )

    with soma.open(
        "s3://cellxgene-data-public/cell-census/2023-09-14/soma/", context=ctx
    ) as census:
        mouse_experiment: soma.Experiment = census["census_data"]["mus_musculus"]
        X = mouse_experiment.ms["RNA"].X["raw"]

        # Case 1 - works correctly, in that it returns an empty result
        var_joinids = slice(None)
        for tbl in X.read((pa.array([], type=pa.int64()), var_joinids)).tables():
            print("Case 1:", len(tbl))

        # Case 2 - works correctly, in that it returns an empty result
        var_joinids = None
        for tbl in X.read((pa.array([], type=pa.int64()), var_joinids)).tables():
            print("Case 2:", len(tbl))

        # Case 3 - FAILS, as it returns _all_ data in the sparse array
        var_joinids = (
            mouse_experiment.ms["RNA"]
            .var.read(
                (slice(None),),
                column_names=["soma_joinid"],
            )
            .concat()
            .column("soma_joinid")
            .combine_chunks()
        )
        for tbl in X.read((pa.array([], type=pa.int64()), var_joinids)).tables():
            print("Case 3:", len(tbl))


if __name__ == "__main__":
    main()
```

Running it shows that the third case is incorrect (the first two return zero-length results, but the third case does not):
```
python tmp/query_fail.py 
Case 1: 0
Case 2: 0
Case 3: 536870912
Case 3: 536870912
<snip, for brevity>
Case 3: 536870912
Case 3: 536870912
```

[...]

PR #1706, opened 2023-09-20
[c++] Fix for empty-indexed multi-dimensional array

nguyenv:
**Issue and/or context:**
https://github.com/single-cell-data/TileDB-SOMA/issues/1703

**Changes:**
Modify `subarray_range_empty_` to be a mapping of dimension name to whether the query range is empty. `is_empty_query()` returns true if any of the ranges are empty.


johnkerl (approved):
🚀

eddelbuettel (approved):
:+1:

johnkerl:
Analysis here: https://github.com/single-cell-data/TileDB-SOMA/issues/1703#issuecomment-1729612897

PR #1707, opened 2023-09-21
[python] Support enumerations in `update_obs`/`update_var`

johnkerl:
**Issue and/or context:** Validates `update_obs`/`update_var` when the newly added column is an enumeration: tracking issue #1710, with grandparent #866.

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/io/_registration/signatures.py
+++ b/apis/python/src/tiledbsoma/io/_registration/signatures.py
@@ -72,8 +72,9 @@ def _string_dict_from_pandas_dataframe(
         df.reset_index(inplace=True)
         if default_index_name in df:
             if "index" in df:
-                # Avoid warning:
+                # Avoid the warning:
                 # "A value is trying to be set on a copy of a slice from a DataFrame"
+                # which would occur if we did:
                 # df.drop(columns=["index"], inplace=True)
                 df = df.drop(columns=["index"])
         else:
--- a/apis/python/src/tiledbsoma/io/_registration/signatures.py
+++ b/apis/python/src/tiledbsoma/io/_registration/signatures.py
@@ -72,7 +72,10 @@ def _string_dict_from_pandas_dataframe(
         df.reset_index(inplace=True)
         if default_index_name in df:
             if "index" in df:
-                df.drop(columns=["index"], inplace=True)
+                # Avoid warning:
+                # "A value is trying to be set on a copy of a slice from a DataFrame"
+                # df.drop(columns=["index"], inplace=True)
+                df = df.drop(columns=["index"])
         else:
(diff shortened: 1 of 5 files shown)


johnkerl:
This is having a weird CI fail
which I cannot repro on my own Mac; also I've manually triggered full CI here
https://github.com/single-cell-data/TileDB-SOMA/actions/workflows/python-ci-full.yml

cc @ihnorton I _thought_ we resolved all mac-CI issues a couple weeks ago but I may have missed something ... 👀 🤔

apis/python/src/tiledbsoma/io/_registration/signatures.py
```diff
@@ -72,7 +72,10 @@ def _string_dict_from_pandas_dataframe(
         df.reset_index(inplace=True)
         if default_index_name in df:
             if "index" in df:
-                df.drop(columns=["index"], inplace=True)
+                # Avoid warning:
+                # "A value is trying to be set on a copy of a slice from a DataFrame"
+                # df.drop(columns=["index"], inplace=True)
```
thetorpedodog, line 77: To make sure I understand this comment: we're *not* doing `inplace=True` to avoid this warning? It might be good to phrase this as something like "We don't do this in-place to avoid the 'A value is trying ..."

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1513,12 +1513,22 @@ def _update_dataframe(
         # schema-creation logic.
         atype = arrow_schema.field(add_key).type
         dtype = tiledb_type_from_arrow_type(atype)
+
+        enum_label = None
```
thetorpedodog, line 1517: You might want to label this as `enum_label: Optional[???] = None`

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1513,12 +1513,22 @@ def _update_dataframe(
...
+        if pa.types.is_dictionary(arrow_table.schema.field(add_key).type):
+            enum_label = add_key
+            ordered = atype.ordered
+            dt = cast(pd.CategoricalDtype, new_data[add_key].dtype)
+            values = dt.categories
+            se.add_enumeration(tiledb.Enumeration(add_key, ordered, list(values)))
```
thetorpedodog, line 1523: Passing the things to `tiledb.Enumeration` as kwargs might make this easier to follow (particularly the `ordered` boolean parameter), and reduce the desire to give these temporary names.

thetorpedodog (commented):
A couple minor code style points but otherwise looks mostly good. Not familiar enough with the way enumerations work to judge that part but provided somebody can OK that, I'm happy.

aaronwolen (approved):

Issue #1714, opened 2023-09-25
Replace SOMA C++ API TileDB Args with `ArrowSchema` or `ArrowArray`


nguyenv:

johnkerl:
Dup of #1713

PR #1640, opened 2023-08-29
[c++/python][wip] Bind Updated C++ API in pybind11

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/1714 [to be split off]
#1715

**Changes:**

- `pytiledbsoma.cc` now contains init hooks into each class binding
- `SOMAArray` binding now moved into `soma_array.cc` -> however, binding `SOMAArray`, `SOMAGroup`, and `SOMAObject` may not be necessary as they are internal classes in the C++ API
- [ ] TODO potentially move stats and enum bindings into separate modules
- [ ] TODO documentation
- [ ] TODO unit tests

**Notes for Reviewer:**

WIP

PR #1716, opened 2023-09-26
[c++] Take In Arrow Structs For `create` Methods

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/1714

Split off from https://github.com/single-cell-data/TileDB-SOMA/pull/1640.

**Changes:**

* Add converter for ArrowSchema to TileDB ArraySchema
* Add `create` static method that takes in Arrow structs for DataFrame, NDSparseArray, and NDDenseArray

**Notes for Reviewer:**

- [ ] Handle enumerations
    - [ ]  Pass argument ArrowArray* where each child array holds the enumation label and index values
    - [ ]  Pass argument map<string, string> to link attribute name to enumeration label
- [x] Set nullable attributes
- [ ] Set filters
- [ ] Switch out raw pointers with pass-by-reference or `unique_ptr`
- [ ] Add unit tests

### DIFF
--- a/libtiledbsoma/src/soma/soma_dataframe.cc
+++ b/libtiledbsoma/src/soma/soma_dataframe.cc
@@ -46,13 +46,14 @@ using namespace tiledb;
 
 std::unique_ptr<SOMADataFrame> SOMADataFrame::create(
     std::string_view uri,
-    ArrowSchema* schema,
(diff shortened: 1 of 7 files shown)


nguyenv:
Replaced with #2367

PR #1720, opened 2023-09-26
[r] Update `SOMADataFrame$create()` to automatically mark `ordered()`

mojaveazure:
Add an check in `SOMADataFrame$create()` to detect if a column in the schema is ordered. If so, automatically mark the enumerations as such.

Note: this only happens if
- there are enumerations for the column
- the column in the arrow schema is marked as ordered
- there is no attribute `'ordered'` on the enumerations for the column


johnkerl (approved):

eddelbuettel (approved):
Thanks for this nice addition to snooth out another corner.

johnkerl:
Backport CI is not triggering; I will manually create a backport PR for `release-1.5`

PR #1723, opened 2023-09-27
[python] update to numba 0.58

mlin:
The newer version lets us clean up some existing dependency complications -- especially a constraint against newer versions of numpy, which was getting more problematic as time goes on.


mlin:
wheels test: https://github.com/single-cell-data/TileDB-SOMA/actions/runs/6324289205/job/17174223306

thetorpedodog (approved):
we like a simpler setup

johnkerl (approved):

Issue #1614, opened 2023-08-17
C++ optimization for re-indexing and CSR conversion [parent]


beroy:
Wrapping up multiple parallel C++ kernels and integrate them inside both critical python and R components int the SOMA stack. They include perfect hashing (parallel khash) and COO -> CSR conversion kernel

Issue #1611, opened 2023-08-17
[python] Refactoring `ExperimentAxisQuery` to use the C++ CSR and re-indexer


beroy:
Currently ExperimentAxisQuery is the only class in SOMA that does provide its own implementation. We need to refactor the functionality between SOMA and TileDB-SOMA such that we can use C++ re-indexer and CSR for `by_vars` and `by_obs` and `csr_iter` functions of that class.

Context: #1614

johnkerl:
@beroy is this completed?

Issue #1610, opened 2023-08-17
[r] Add R support for re-indexer


beroy:
[parent #1614]
Understand Rcpp API and provide the right connector for C++ re-indexer in R.

johnkerl:
[sc-36686]

johnkerl:
[sc-40890]

Issue #1590, opened 2023-08-14
Add parallel re-indexer implemented in C++ and provide API to be used in SOMA Axis Query for both R and Python


beroy:
**Is your feature request related to a problem? Please describe.**
The re-indexing is needed across vars and obs coordinates is a needed feature which currently uses [Pandas reindexing](https://github.com/single-cell-data/SOMA/blob/53a772e45e2915ed31bf79f2b504d6e4068730b4/python-spec/src/somacore/query/query.py#L578-L592). An example of the main case can be found [here](https://github.com/chanzuckerberg/cellxgene-census/blob/main/api/python/cellxgene_census/src/cellxgene_census/experimental/util/_csr_iter.py
 )

Unfortunately the Pandas re-indexing mechanism is not parallelizable. 

**Describe the solution you'd like**
Based on our research the best khash is the most performant hash with some extra cost on the used memory. Additionally, Pandas use khash internally (although single threaded). Our PoC shows that a thread pool implemented in C++ can make khash speed scale linearly by up to 8 or 16 threads. Using threadpool inside tileDB core is not a choice they are not open to public. So the solution is to add the threadpool to tileDB-SOMA and then provide the right cython API that can be used by SOMA Axis Query in both python and R.

**Describe alternatives you've considered**
We look into a bunch of other hashing mechanism including BT hashing and the conclusion was khash is the best choice.

bkmartinjr:
> Unfortunately the Pandas re-indexing mechanism is not parallelizable.

It is certainly feasible to externally partition (multi-threaded) calls to Pandas.Index.get_indexer (although this is not explicitly supported by Pandas, it works fine in the current implementation).  The larger motivation is that we want a common, high-performance implementation to be accessible to both R and Python TileDB-SOMA implementations, and there is an opportunity to also reduce over-parallelization by having a single TileDB-SOMA thread pool which is used for other work (e.g., the upcoming CSR/CSC performance effort).

PR #1728, opened 2023-09-27
[c++] Reindexer overrides and fast COO/CSR

beroy:
**Issue and/or context:**

#1611 
#1614 
#1590 
#718 
#1610

**Changes:**

- libtiledbsoma
- python apis
- AxisQuery and some other python code

**Notes for Reviewer:**

Implement a re-indexer similar to Pandas reindexer
Provide _pybind11_ for the re-indexer (currently only numpy and soon Arrow interface) for python

- _khash.h_ and _khashl.h_ are [klib](https://github.com/attractivechaos/klib) library  

Expose __AxisQuery__ and fast_csr to use the re-indexer to speedup COO to CSR conversion

For the time being switching between the two indexer (IntIndexer and pandas indexer) is possible using two variables (named use_int_indexer) in _query and _fast_csr_. As a part of this PR review, I'd like to figure out the best way to have this new feature  easily enabled or disabled.


thetorpedodog:
Reading this now, I have a good idea of how we can add an abstraction in somacore to make this change much less invasive and more straightforward. I have some more work to do here, but this is the start: https://github.com/single-cell-data/SOMA/compare/main...build-index-abstraction

bkmartinjr:
> this is the start

I like @thetorpedodog's starting refinement.

Questions before we finalize it:

1. if the indexer needs access to shared state, e.g., a thread pool or context, how is that passed to `_build_index`? Short term, I don't think we need it, but longer term is seems like it might be useful.
2. do we like adopting the Pandas API `get_indexer` API, or do we want to tweak that?  (the current proposal seems OK to me)

thetorpedodog:
> 1. if the indexer needs access to shared state, e.g., a thread pool or context, how is that passed to `_build_index`? Short term, I don't think we need it, but longer term is seems like it might be useful.

I’ve been working on some more plumbing on this, and (as of the current version) whatever thing you pass to `index_factory` will be called for all indices in the current AxisQuery. I am happy to hear more ideas.

> 2. do we like adopting the Pandas API `get_indexer` API, or do we want to tweak that?  (the current proposal seems OK to me)

I would suggest mirroring the `get_indexer` API, as reflected in the [current version of the `IndexLike` protocol](https://github.com/single-cell-data/SOMA/blob/1d88f6b5350ec762e028d043d2294df1d1b7a688/python-spec/src/somacore/query/types.py#L10-L26).

(Also, naming is hard. There’s `AxisIndexer`, `pd.Index`, `some_pd_index.get_indexer`, …)

bkmartinjr:
@thetorpedodog - TY.

On (1) - factory, seems good. I can't immediately think of use cases that would be unsupported.

On (2) - naming,  I'm fine with the current proposal. I've always found `get_indexer` slightly confusing, but I don't have a better suggestion (to my eye, it returns the _result_ of indexing, not an indexer).

beroy:
@thetorpedodog and I had a chat about the planning and next steps. With this structure in order to pass the. factory down to soma core, the only function that needs to be `overriden is somacore.Experiment.axis_query` (via `tiledbsoma.Experiment.axis_query` where the index factor is passed to query) which is an very small change.

The plan moving forward is for me to keep my PR on top of his branch (https://github.com/single-cell-data/SOMA/compare/main...build-index-abstraction). Finally, when his PR is in and after then next somacore release, my PR can finally land.

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -0,0 +1,202 @@
...
+            std::chrono::system_clock::now().time_since_epoch())
+            .count();
+    return microseconds_since_epoch;
+}
+
+static std::mutex log_mutex;
+
+static bool no_logging_s = true;
+
+static void log(std::string message, int ln, std::string severity = "INFO") {
```
Shelnutt2, line 49: Please replace log usage with the existing logger, https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/src/utils/logger.h.

I.e: https://github.com/single-cell-data/TileDB-SOMA/blob/af2ead33888de1405be5cec0c91784b78880eec4/libtiledbsoma/src/soma/managed_query.cc#L116

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -0,0 +1,202 @@
...
+
+uint64_t current_time() {
+    uint64_t microseconds_since_epoch =
+        std::chrono::duration_cast<std::chrono::microseconds>(
+            std::chrono::system_clock::now().time_since_epoch())
+            .count();
+    return microseconds_since_epoch;
+}
+
+static std::mutex log_mutex;
```
Shelnutt2, line 45: If you replace the log function with the existing logger this mutex can be removed and we can avoid single threading on logging.

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -0,0 +1,202 @@
...
+
+static bool no_logging_s = true;
```
Shelnutt2, line 47: Lets remove this and use the existing log infrastructure.

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -0,0 +1,202 @@
...
+ * @section DESCRIPTION
+ *
+ * This file defines the a pybind11 api into SOMA C++ library.
+ */
+
+#include "reindexer.h"
+#include <unistd.h>
+#include "khash.h"
+
+uint64_t current_time() {
```
Shelnutt2, line 37: I believe this can be removed when switching to the exiting logger.

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -0,0 +1,202 @@
...
+        k = kh_put(
+            m64,
+            hash_,
+            keys[i],
+            &ret);  // hash_wrapper<IntType>::put(hash_, key, &ret);
+        assert(k != kh_end(hash_));
+        kh_val(hash_, k) = counter;
+        counter++;
+    }
+    size_ = size;
```
Shelnutt2, line 79: In order to comply with C++ Core Guidelines, class level attributes should be set in the constructor and not inside the function. Lets avoid setting size, thread_count and declaring the thread_pool in this function.

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    return results;
+}
+
+void IntIndexer::lookup(int64_t* keys, int64_t* results, int size) {
+    log("Lookup with threadcount " + std::to_string(thread_count_) +
+            " on data size of " + std::to_string(size),
+        __LINE__);
+    if (thread_count_ == 1) {
```
Shelnutt2, line 117: Can we avoid the optimization of the single thread. Keeping the code with a single code path improve readability, maintainability and test-ability.

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -0,0 +1,202 @@
...
+        tasks.emplace_back(std::move(task));
+        log("Task for processing  = " + std::to_string(start) + " " +
+                std::to_string(end) + " inserted in the queue",
+            __LINE__);
+    }
+    tiledb_thread_pool_->wait_all(tasks);
+}
+
+int64_t* IntIndexer::serial_lookup(const int64_t* keys, int size) {
+    int64_t* results = new int64_t[size];
```
Shelnutt2, line 177: Can we switch this to return a `std::vector` instead of a raw array?

libtiledbsoma/src/reindexer/reindexer.h
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * Lookup for the single lookup thread situation
+     * @param keys array of keys to lookup
+     * @param size // Number of key array
+     * @return and array of looked up value (same size as keys)
+     */
+    int64_t* serial_lookup(const int64_t* keys, int size);
+    /* The same lookup function but for vector of keys */
+    std::vector<int64_t> serial_lookup(const std::vector<int64_t>& keys);
+    void lookup(int64_t* keys, int64_t* results, int size);
```
Shelnutt2, line 66: Please add doxygen style comments for all functions and class members.

libtiledbsoma/src/reindexer/reindexer.h
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+#pragma once
+#include <assert.h>
+#include <thread_pool/thread_pool.h>
+#include <unistd.h>
+#include "khash.h"
+
+// A macro controlling when the deletion of hash tables happen
+#define CMAKE_PYTHON_BUILD
```
Shelnutt2, line 40: Can you explain more why we need to define this inline?

libtiledbsoma/src/reindexer/reindexer.h
```diff
@@ -0,0 +1,81 @@
...
+
+KHASH_MAP_INIT_INT64(m64, int64_t)
```
Shelnutt2, line 42: Please add comments explaining this is a typedef.

libtiledbsoma/src/CMakeLists.txt
```diff
@@ -197,6 +198,13 @@ install(FILES
   DESTINATION "include/tiledbsoma"
 )
 
+install(FILES
```
Shelnutt2, line 201: Why are we installing the threadpool headers?

libtiledbsoma/src/CMakeLists.txt
```diff
@@ -208,6 +216,12 @@ install(FILES
   DESTINATION "include/tiledbsoma/utils"
 )
 
+install(FILES
```
Shelnutt2, line 219: Why does the reindexer and khash headers need to be installed? If we want to expose the reindexer there should be a clean public facing API so the internal class can stay abstracted from direct interaction.

libtiledbsoma/src/CMakeLists.txt
```diff
@@ -162,7 +163,7 @@ endif()
 # Install header files 
 # target_sources FILE_SET is preferred with cmake>=3.23
 # TODO Uncomment after finishing Python and R bindings
-# set(TILEDB_SOMA_PUBLIC_HEADERS
+set(TILEDB_SOMA_PUBLIC_HEADERS
```
Shelnutt2, line 166: Is this suppose to be uncommented?

Shelnutt2 (commented):
Some initial comments, feedback and suggestions.

libtiledbsoma/src/reindexer/reindexer.cc
beroy: Sure I will replace the logger

libtiledbsoma/src/reindexer/reindexer.cc
beroy: The only reason for having this path is for potential testing.

libtiledbsoma/src/CMakeLists.txt
beroy: Because I'm using the threadpool

libtiledbsoma/src/CMakeLists.txt
beroy: Fixed

libtiledbsoma/src/reindexer/reindexer.cc
beroy: That's good point. I'll switch the logger.

libtiledbsoma/src/reindexer/reindexer.h
beroy: Sorry part of debugging. Not used any more.

nguyenv:
As discussed with @beroy offline, the `IntIndex` code should reside in `apis/python/src/tiledbsoma` alongside the other Pybind11 specific classes and should not be a part of the C++ API. The problem is that `reindexer.h` depends on `<thread_pool/thread_pool.h>`  which is code internal to `libtiledbsoma`. That means when building the Pybind11 code, `reindexer.h` does not have access to `thread_pool.h` unless we install `thread_pool.h` somwhere -- which is not an ideal solution either. Opening discussion to any ideas of how we can resolve this.

beroy:
@nguyenv I'm ok with this solution as long as we have general agreement. @johnkerl @thetorpedodog

nguyenv:
My preference actually would be to place the `thread_pool.cc`, `thread_pool.h`, and `khash.h` code in `api/python/tiledbsoma/src`. Then in `setup.py` add `thread_pool.cc` as a source file into `Pybind11Extension` (along with `reindexer.cc`). It's still not ideal because we now carry two copies of the same code in the repo. But it avoids installing internal headers and code that is meant to be Python-only.

eddelbuettel:
I presume R would have to do the same making it three copies of internal files.

nguyenv:
Yikes that's too many copies 😬

beroy:
Yes, I was gonna make the same comment api is not the rest place!

libtiledbsoma/src/reindexer/reindexer.h
```diff
@@ -0,0 +1,85 @@
...
+    int thread_count_;
+    int size_;
+    /*
+     * The created 64bit hash table
+     */
+    khash_t(m64) * hash_;
+    /*
+     * TileDB threadpool
+     */
+    std::shared_ptr<tiledbsoma::ThreadPool> tiledb_thread_pool_ = nullptr;
```
nguyenv, line 82: We could probably avoid needing to install the `thread_pool.h` and `khash.h` headers altogether by forward declaring `ThreadPool` and `khash_t(m64)` since these are both pointers and then move the includes into the cc. This way we can keep the threadpool and khash code private.

eddelbuettel:
There is a design bug here in if one queries with _fewer keys_ than there were default threads the logic goes haywire.

libtiledbsoma/src/reindexer/reindexer.h
```diff
@@ -0,0 +1,96 @@
...
+     * Used for parallel lookup using khash
+     * @param keys array of keys to lookup
+     * @param result array for lookup results
+     * @param size // Number of key array
+     * @return and array of looked up value (same size as keys)
+     */
+    void lookup(const int64_t* keys, int64_t* results, int size);
+    void lookup(
+        const std::vector<int64_t>& keys,
+        std::vector<int64_t> results) {
```
eddelbuettel, line 68: This is missing an `&` for results.  As it stands the modified vector does not come out.

apis/python/src/tiledbsoma/pytiledbsoma.cc
```diff
@@ -50,659 +50,727 @@ using namespace py::literals;
...
+                   } else if (
+                       !strcmp(arrow_schema.format, "U") ||
+                       !strcmp(arrow_schema.format, "Z")) {
+                       reader.set_dim_points(
+                           dim, coords.cast<std::vector<std::string>>());
+                   } else {
+                       throw TileDBSOMAError(fmt::format(
+                           "[pytiledbsoma] set_dim_points: type={} not "
+                           "supported",
+                           arrow_schema.format));
```
nguyenv, line 480: We need to remove `fmt` formatting from the Pybind11 code. Just doing basic string concatenation here is fine.

beroy:
@thetorpedodog this PR is now synced your SOMA [branch](https://github.com/single-cell-data/SOMA/tree/build-index-abstraction)

eddelbuettel:
I rebased it once more as we also had corrections to the CI system last evening.  The branch and PR are now current to the main branch and should not be affected by other issues.  Let's see how CI shakes out.

apis/python/src/tiledbsoma/IntIndexer.py
thetorpedodog, line 1: It looks like `indexer_map_locations` could go in the experiment.py file. There doesn’t seem to be any need for the `IntIndexer` class, since it is just one static method that isn’t referenced anywhere.

Since `indexer_map_locations` isn’t ever going to be directly used by callers (and will only be internally referred to in the experiment.py file), it can be named `_indexer_map_locations`.

apis/python/src/tiledbsoma/pytiledbsoma.cc
```diff
@@ -693,5 +694,69 @@ PYBIND11_MODULE(pytiledbsoma, m) {
...
+                size_t length = buffer.shape[0];
+                indexer.map_locations(keys.data(), keys.size(), num_threads);
+            })
+        .def(
+            "map_locations",
+            [](IntIndexer& indexer,
+               std::vector<int64_t> keys,
+               int num_threads) {
+                indexer.map_locations(keys.data(), keys.size(), num_threads);
+            })
```
thetorpedodog, line 724: It looks like this is only ever going to be used like

```python
indexer = clib.IntIndexer()
indexer.map_locations(...)
# pass indexer off to be used elsewhere
```

In this case, I would just make it so you pass the relevant data directly into the initializer, so there’s no way to end up with an invalid data structure:

```python
indexer = clib.IntIndexer(...)
# pass it off to be used elsewhere
```

libtiledbsoma/test/test_indexer.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+# 1d array to list
+def indexer_test(keys: np.array, lookups: np.array):
+    reindexing_exception: bool = False
+    try:
+        indexer = IntIndexer.map_locations(keys)
+        results = indexer.get_indexer(lookups)
+    except pd.errors.InvalidIndexError:
+        reindexing_exception = True
```
thetorpedodog, line 16: Instead of testing the invalid-index and valid cases in the same test, it might be easier to follow if one function tests the good cases, and another tests the bad cases.

apis/python/setup.py
thetorpedodog, line 1: For the purposes of verifying things over the course of the review process, could you update the dependency on `somacore` to temporarily be

`git+https://github.com/single-cell-data/SOMA.git@41b8ab465c6d05284ff56dc972560d3d04f7704a`

?

This will be a temporary measure which will only be used in this branch. Once we release somacore, you can then change the dependency to the new somacore version.

apis/python/src/tiledbsoma/IntIndexer.py
thetorpedodog, line 1: I don't think this file or the IntIndexer class need to exist at all; there is no reason for an external user to ever call it.

apis/python/src/tiledbsoma/_experiment.py
```diff
@@ -5,16 +5,40 @@
...
+            "Reindexing only valid with uniquely valued Index objects"
+        )
+    context: SOMATileDBContext = _validate_soma_tiledb_context(
+        SOMATileDBContext(tiledb.default_ctx())
+    )
+    compute_concurrency: int = 5
+    if context._tiledb_ctx:
+        compute_concurrency = int(
+            int(context._tiledb_ctx.config()["sm.compute_concurrency_level"]) / 2
+        )
```
thetorpedodog, line 37: I think I composed a reply on the last revision of this but GitHub ate it or I forgot to save or something.

This ends up constructing an entirely new TileDB Context, which is one thing we want to avoid. This is something we should get from the context that the Experiment already owns.

thetorpedodog:
After rebasing this, the new revision of somacore for the setup.py file is: https://github.com/single-cell-data/SOMA/commit/1d71cc1f54eb700f6215d1d82cfa74e0d01571d3

apis/python/src/tiledbsoma/_experiment.py
beroy: Fixed

apis/python/src/tiledbsoma/IntIndexer.py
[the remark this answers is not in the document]
beroy: The reason for implementing this API is its similarity to pandas indexer:
```
        indexer = IntIndexer.map_locations(keys)
        indexer.get_indexer(lookups)
```
@bkmartinjr and @beroy discussed this before and having this API opened up (to SOMA) can make it easy to replace pandas indexer usecases with this faster option.

apis/python/src/tiledbsoma/_experiment.py
```diff
@@ -5,16 +5,41 @@
...
 from ._collection import Collection, CollectionBase
 from ._dataframe import DataFrame
 from ._measurement import Measurement
 from ._tdb_handles import Wrapper
 from ._tiledb_object import AnyTileDBObject
+from .options import SOMATileDBContext
+from .options._soma_tiledb_context import _validate_soma_tiledb_context
+
+
+def _indexer_map_locations(keys: np.typing.NDArray[np.int64], context: SOMATileDBContext) -> clib.IntIndexer:
```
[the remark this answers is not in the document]
beroy: Referring to the API discussion, this change is only possible if we remove IntIndexer API.

apis/python/src/tiledbsoma/pytiledbsoma.cc
beroy: `pybind11` imports constructors from C++ directly. For example here I can use my map_locations inside the API's map_location but that's not possible for constructors. For example, in this case, for the non-copy constructor to work it must have `IntIndexer::IntIndexer(py::array_t<int64_t>& keys)`, int but the non-copy constructor of my C++ IntIndexer has the following signature `IntIndexer::IntIndexer(const int64_t* keys, int size, int threads)`. Opening up `py::*` types to C++ is something that I strongly tried to avoid.

apis/python/src/tiledbsoma/IntIndexer.py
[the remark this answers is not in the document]
bkmartinjr: I am a little confused what the core question is here, so I'll throw in a few comments.

1. Do we need to do reindexing on anything other than the obs/var dimensions of the ExperimentAxisQuery (e.g, somacore.AxisIndexer.by_obs / somacore.AxisIndexer.by_var)?
2. Do these use cases motivate a public API (such as is present in this file)?

On the first, there are _definitely_ use cases that require using the indexer directly - for example striding over high-dimensionality sparse matrices and using reindexing to reshape them (outside the context of any given query). This repo already has code that does this:  see _read_iters.py, BlockwiseReadIterBase:cpp_reindexer_overwrites(), which is currently using pandas.Index, but needs to be ported to the new khash code.

(side question: why has that port not been done as part of this PR?)

On the second: our cellxgene-census package would greatly benefit from a faster indexer in various computational methods. Use cases are similar - we are striding over large sparse arrays in a variety of ways, and want to reorganize or map between joinids and positional indices. I had always assumed we would use the `tiledbsoma` perfect hash function for this, rather than having to duplicate the implementation (and I don't want to use private interfaces for all the obvious reasons).

@thetorpedodog -can you elaborate on your comment that no external user would ever want to call this?  I know I definitely _would_ use it, and it would make a large positive impact.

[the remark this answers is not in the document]
thetorpedodog: I was under the impression that this interface was internal-only, since the only thing that used it at the moment is the axis query itself. In this case, we should expose the function directly somehow (is there a better place than putting it at the root of tiledbsoma?), rather than indirecting it through a static method on an otherwise-unused class.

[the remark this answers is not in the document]
bkmartinjr: I am very open to your suggestions on how to expose this.  But essentially what I'd like is a public/supported API that provides the ability to create the indexer object, and the get_indexer / lookup on that.  I..e, the equivalent of pandas.Index() and pandas.Index.get_indexer.

We may also want to consider  making this a tiledbsoma priviate (i.e., not somacore) API for the time being.  Perhaps in a `util` sub-module or something like that....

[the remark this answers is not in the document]
beroy: OK, sounds good. I'll create a submodule (`util`) and move it under that and update my unittests.

apis/python/src/tiledbsoma/_experiment.py
```diff
@@ -74,3 +98,26 @@ def _set_create_metadata(cls, handle: Wrapper[Any]) -> None:
...
+        Lifecycle: maturing
+        """
+        # mypy doesn't quite understand descriptors so it issues a spurious
+        # error here.
+        return query.ExperimentAxisQuery(  # type: ignore
+            self,
+            measurement_name,
+            obs_query=obs_query or query.AxisQuery(),
+            var_query=var_query or query.AxisQuery(),
+            index_factory=_indexer_map_locations,  # comment this line to disable thw C++ indexer
```
thetorpedodog, line 122: With [the context in the somacore review](https://github.com/single-cell-data/SOMA/pull/181#discussion_r1422604804) for why we want to have this still just take a `NDArray` as its only parameter, we can still pass the context through to the index builder:

```python
def indexer_map_locations(keys: ..., ctx: ...) -> clib.IntIndexer:
  ...

class Experiment(...):
  ...

  def axis_query(self, ...):
    return query.ExperimentAxisQuery(
        ...,
        index_factory=functools.partial(indexer_map_locations, context=self.context),
    )
```

This means that the `indexer_map_locations` function still knows about the `self.context`, but the somacore code doesn’t need to plumb it around just to pass back into our code.

apis/python/src/tiledbsoma/utils/IntIndexer.py
```diff
@@ -0,0 +1,21 @@
...
+"""External API"""
+
+
+class IntIndexer:
+    @staticmethod
+    def map_locations(
+        keys: np.typing.NDArray[np.int64],
+        context: Union[SOMATileDBContext, None] = None,
+    ) -> clib.IntIndexer:  # type ignore
+        return _indexer_map_locations(keys, context)
```
thetorpedodog, line 18: What I was trying to suggest in the last round of review is that there is no need for a `IntIndexer.map_locations` function; the `indexer_map_locations` function should be exported directly without wrapping it in a class.

Assuming we want to export this name on the `utils` namespace (which makes sense to me), I think the best way to accomplish this would be something like:

```python
# in tiledbsoma/utils.py

# This is a naming suggestion; 
def build_index(keys: ..., context: ...) -> clib.IntIndexer:
  # the contents of _indexer_map_locations above
```

```python
# in tiledbsoma/_experiment.py
from . import utils

class Experiment(...):
  def axis_query(...):
    return query.ExperimentAxisQuery(
        ...,
        index_factory=functools.partial(utils.build_index, ...),
    )
```

This will make the function available to end users, with its home in the `tiledbsoma.utils` namespace.

beroy: Yeah, from my perspective, a static class function and a global function sounds the same (me being naive). I like this model and if @bkmartinjr has no objection I'll go ahead with your suggestion @thetorpedodog.

beroy: Make the suggested changes. Plz lmk if it all looks good @thetorpedodog

libtiledbsoma/test/test_indexer.py
thetorpedodog: Since this function now only tests failing cases, can this be shortened to eliminate the extra logic? It seems like it should be able to be something like:

```
def indexer_test_fail(...):
    try:
        index = build_index(keys)
        index.get_indexer(lookups)
        raise AssertionError("should have failed")
    except ...:
        pass

    try:
        pd_index = pd.Index(keys)
        pd_index.get_indexer(lookups)
        raise AssertionError("should have failed")
    except ...:
        pass
```

apis/python/src/tiledbsoma/_experiment.py
```diff
@@ -74,3 +73,26 @@ def _set_create_metadata(cls, handle: Wrapper[Any]) -> None:
...
+        Lifecycle: maturing
+        """
+        # mypy doesn't quite understand descriptors so it issues a spurious
+        # error here.
+        return query.ExperimentAxisQuery(  # type: ignore
+            self,
+            measurement_name,
+            obs_query=obs_query or query.AxisQuery(),
+            var_query=var_query or query.AxisQuery(),
+            index_factory=build_index,  # comment this line to disable thw C++ indexer
```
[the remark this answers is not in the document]
beroy: Hi @thetorpedodog I managed to implemented all of your requested changes except for this one because it continuously fails this after that change:
```
$ python  libtiledbsoma/src/reindexer/example.py
99187
  File "/Users/brobatmili/projects/TileDB-SOMA/libtiledbsoma/src/reindexer/example.py", line 26, in <module>
    main()
  File "/Users/brobatmili/projects/TileDB-SOMA/libtiledbsoma/src/reindexer/example.py", line 21, in main
    query.to_anndata(X_name="raw")
  File "/Users/brobatmili/soma_index_abstraction/lib/python3.11/site-packages/somacore/query/query.py", line 294, in to_anndata
    return self._read(
           ^^^^^^^^^^^
  File "/Users/brobatmili/soma_index_abstraction/lib/python3.11/site-packages/somacore/query/query.py", line 401, in _read
    x_matrices = {
                 ^
  File "/Users/brobatmili/soma_index_abstraction/lib/python3.11/site-packages/somacore/query/query.py", line 402, in <dictcomp>
    _xname: _fast_csr.read_csr(
            ^^^^^^^^^^^^^^^^^^^
  File "/Users/brobatmili/soma_index_abstraction/lib/python3.11/site-packages/somacore/query/_fast_csr.py", line 29, in read_csr
    acc = _CSRAccumulator(
          ^^^^^^^^^^^^^^^^
  File "/Users/brobatmili/soma_index_abstraction/lib/python3.11/site-packages/somacore/query/_fast_csr.py", line 98, in __init__
    self.obs_indexer = index_factory(self.obs_joinids, context)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: build_index() got multiple values for argument 'context'
```
Appreciate any suggestions.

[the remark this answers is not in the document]
thetorpedodog: In this case you will want to go back to `git+https://github.com/single-cell-data/SOMA.git@1d71cc1f54eb700f6215d1d82cfa74e0d01571d3` as the dependency version for somacore.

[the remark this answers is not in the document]
beroy: It works. Thanks

apis/python/src/tiledbsoma/utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def build_index(keys: np.typing.NDArray[np.int64], context: SOMATileDBContext) -> clib.IntIndexer:
+    if len(np.unique(keys)) != len(keys):
+        raise pd.errors.InvalidIndexError(
+            "Reindexing only valid with uniquely valued Index objects"
+        )
+    # If we chose to keep `context` as optional, we would add this line here:
+    #     context = context or options.SOMATileDBContext()
+    tdb_concurrency = int(context.tiledb_ctx.config().get("sm.compute_concurrency_level", _validate_soma_tiledb_context(
```
thetorpedodog, line 17: The "default" parameter to `.get` should just be the default value we want (or, in this case, twice the number of threads we want since we divide it by two below). If there's no `sm.compute_concurrency_level` entry in the config, this is going to try to convert a SOMATileDBContext into an int, which will fail.

apis/python/src/tiledbsoma/_experiment.py
```diff
@@ -74,3 +75,26 @@ def _set_create_metadata(cls, handle: Wrapper[Any]) -> None:
...
+    def axis_query(  # type: ignore
+            self,
+            measurement_name: str,
+            *,
+            obs_query: Optional[query.AxisQuery] = None,
+            var_query: Optional[query.AxisQuery] = None,
+    ) -> query.ExperimentAxisQuery[Self]:
+        """Creates an axis query over this experiment.
+
+
```
thetorpedodog, line 89: extra space here

apis/python/src/tiledbsoma/utils.py
```diff
@@ -0,0 +1,22 @@
...
+from .options._soma_tiledb_context import _validate_soma_tiledb_context
+
+
+def build_index(keys: np.typing.NDArray[np.int64], context: SOMATileDBContext) -> clib.IntIndexer:
+    if len(np.unique(keys)) != len(keys):
+        raise pd.errors.InvalidIndexError(
+            "Reindexing only valid with uniquely valued Index objects"
+        )
+    # If we chose to keep `context` as optional, we would add this line here:
+    #     context = context or options.SOMATileDBContext()
```
thetorpedodog, line 16: don't forget to remove these explanatory comments which were only for the review process

apis/python/src/tiledbsoma/utils.py
```diff
@@ -0,0 +1,22 @@
...
+    tdb_concurrency = int(context.tiledb_ctx.config().get("sm.compute_concurrency_level", _validate_soma_tiledb_context(
+        SOMATileDBContext(tiledb.default_ctx()))))
+    thread_count = tdb_concurrency // 2  # // is the floor-division operator: https://docs.python.org/3/glossary.html#term-floor-division
```
thetorpedodog, line 19: likewise with this comment

apis/__init__.py
thetorpedodog, line 1: It looks like something keeps re-creating these files if you delete them. If you can figure out what that is and make it stop, that's great; otherwise check that you exclude them whenever you do a Git commit.

apis/python/src/tiledbsoma/utils.py
```diff
@@ -0,0 +1,22 @@
+import numpy as np
+import pandas as pd
+from tiledbsoma import pytiledbsoma as clib
+from typing import Any
+import tiledb
+from .options import SOMATileDBContext
+from .options._soma_tiledb_context import _validate_soma_tiledb_context
+
+
+def build_index(keys: np.typing.NDArray[np.int64], context: SOMATileDBContext) -> clib.IntIndexer:
```
thetorpedodog, line 10: Since this is kind of public-facing, it should probably include a docstring. This might be a good start:

```
"""Builds an indexer object compatible with :meth:`pd.Index.get_indexer`."""
```

thetorpedodog (commented):
The Python side of this is pretty close to done; I think with these few changes it will be completely ready and you should only need an OK on the C++ code.

apis/__init__.py
beroy: Sure
\

apis/python/setup.py
```diff
@@ -186,6 +187,7 @@ def run(self):
     "dist_links/libtiledbsoma/external/include",
     "../../build/externals/install/include",
     str(libtiledbsoma_dir / "include"),
+    str(libtiledbsoma_dir / "include/tiledbsoma/soma/"),
```
nguyenv, line 190: Why did this need to be added?

libtiledbsoma/src/reindexer/example.py
```diff
@@ -0,0 +1,25 @@
+import os
+from time import perf_counter
+
+import cellxgene_census
+
+import tiledbsoma as soma
+
+# soma.pytiledbsoma.config_logging("debug")
+print(os.getpid())
```
nguyenv, line 9: Remove?

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -0,0 +1,159 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @section DESCRIPTION
+ *
+ * This file defines the a pybind11 api into SOMA C++ library.
```
nguyenv, line 30:
```suggestion
 * This file defines the pybind11 api in the SOMA C++ library.
```

nguyenv: Is this only for Pybind11 btw? I thought I heard this was going into the R API as well.

libtiledbsoma/test/test_indexer.py
```diff
@@ -0,0 +1,193 @@
+import numpy as np
+import pandas as pd
+import tiledb
+from tiledbsoma.utils import build_index
+from tiledbsoma.options import SOMATileDBContext
+from tiledbsoma.options._soma_tiledb_context import _validate_soma_tiledb_context
+
+
+# tiledbsoma.pytiledbsoma.config_logging("debug")
```
nguyenv, line 9: Remove

libtiledbsoma/test/test_indexer.py
```diff
@@ -0,0 +1,193 @@
...
+        raise AssertionError("should have failed")
+    except pd.errors.InvalidIndexError:
+        pass
+
+    try:
+        pd_index = pd.Index(keys)
+        pd_index.get_indexer(lookups)
+        raise AssertionError("should have failed")
+    except pd.errors.InvalidIndexError:
+        pass
```
nguyenv, line 36: https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.raises

There is a way to test for expected exceptions in Pytest by using `with pytest.raises(...):`

apis/python/setup.py
```diff
@@ -293,9 +295,9 @@ def run(self):
         "pyarrow>=9.0.0; platform_system!='Darwin'",
         "scanpy>=1.9.2",
         "scipy",
-        "somacore==1.0.6",
+        # "somacore==1.0.4", # temporarily change for this PR! TODO will be reversed before landing
+        "somacore @ git+https://github.com/single-cell-data/SOMA.git@1d71cc1f54eb700f6215d1d82cfa74e0d01571d3",
         "tiledb~=0.24.0",
-        "typing-extensions",  # Note "-" even though `import typing_extensions`
```
nguyenv, line 298: Was this supposed to be removed?

nguyenv (commented):
Left some comments for minor changes.

libtiledbsoma/src/reindexer/reindexer.h
```diff
@@ -0,0 +1,95 @@
...
+
+   private:
+    /*
+     * The created 64bit hash table
+     */
+    kh_m64_s* hash_;
+    /*
+     * TileDB threadpool
+     */
+    std::shared_ptr<tiledbsoma::ThreadPool> tiledb_thread_pool_ = nullptr;
```
nguyenv, line 92: Should this be a `unique_ptr` or `shared_ptr`? Because you used `make_unique` to set this in `IntIndexer::map_locations`.

libtiledbsoma/test/test_indexer.py
thetorpedodog: Are these tests also run with pytest? I hadn’t realized that; I thought they were just executed as `python the-test-file.py`.

nguyenv: Yeah when you [run](https://github.com/single-cell-data/TileDB-SOMA/blob/main/Makefile#L35) `make test` it also runs all the Python unit tests in `libtiledbsoma/test`.

libtiledbsoma/src/reindexer/reindexer.h
beroy: It's to be shared_ptr because of the `ThreadPool` type forwarding requires sizeof working on T.

```
/unique_ptr.h:63:19: error: invalid application of 'sizeof' to an incomplete type 'tiledbsoma::ThreadPool'
          static_assert(sizeof(_Tp) >= 0, "cannot delete an incomplete type");
          ```

File: apis/python/setup.py (hunk 34 above)
Comment (beroy) [reply apis/python/setup.py:298]: Put it back in

File: libtiledbsoma/src/reindexer/reindexer.cc (hunk 31 above)
Comment (beroy) [reply libtiledbsoma/src/reindexer/reindexer.cc]: Yes, but R will directly use the C++ API

Comment (nguyenv) [reply libtiledbsoma/src/reindexer/reindexer.cc]: Should the reference to Pybind11 be removed from the description then?

File: libtiledbsoma/src/reindexer/reindexer.cc [hunk 36]
```diff
@@ -0,0 +1,159 @@
...
+    if (size < 10) {
+        threads = 1;
+    }
+    if (size < threads)
+        throw std::runtime_error(
+            "The number of keys " + std::to_string(size) +
+            " must be larger than the number of threads " +
+            std::to_string(threads) + " .");
+
+    LOG_DEBUG(fmt::format(
```
nguyenv, line 61: We might want to consider adding `LOG_DEBUG` lines for codecov to ignore in `pyproject.toml`. I'm not sure if these debugging lines are enabled when running CI. If they aren't then we could probably get rid of a lot of false positives this way.

nguyenv:
It looks like it is currently failling because it is using an older version of somacore

libtiledbsoma/src/reindexer/reindexer.cc
nguyenv: Since codecov is complaining, it looks like we should add unit tests that directly test the C++ reindexer code.

beroy: Yeah, I was actually thinking about that. I added that codecov support recently and it's biting me first. :)

beroy: @nguyenv done

beroy:
The issue with dependency and tests are fixed @nguyenv

nguyenv:
Looks like you need to run pre-commit for formatting

thetorpedodog (commented):
This looks good to me on the Python side; consider this a LGTM for when somebody fully approves on the C++ side. Given that it looks like it’s very close, I will merge in and release the somacore changes to go with it.

apis/python/src/tiledbsoma/_experiment.py
```diff
@@ -74,3 +75,24 @@ def _set_create_metadata(cls, handle: Wrapper[Any]) -> None:
         # TileDB Cloud UI to detect that they are SOMA datasets.
         handle.metadata["dataset_type"] = "soma"
         return super()._set_create_metadata(handle)
+
+    # in tiledbsoma
```
thetorpedodog, line 79: nit: remove inaccurate comment

libtiledbsoma/src/reindexer/example.py
```diff
@@ -0,0 +1,21 @@
...
+        with census["census_data"]["homo_sapiens"].axis_query(
+            measurement_name="RNA",
+            obs_query=soma.AxisQuery(value_filter="""tissue_general == 'eye'"""),
+        ) as query:
+            query.to_anndata(X_name="raw")
+    t2 = perf_counter()
+    print(f"End to end time {t2 - t1}")
+
+
+main()
```
thetorpedodog, line 21:
```python

if __name__ == "__main__":
  main()
```

apis/python/src/tiledbsoma/_experiment.py
```diff
@@ -43,7 +45,6 @@ class Experiment(  # type: ignore[misc]  # __eq__ false positive
             A collection of named measurements.
 
     Example:
-        >>> import tiledbsoma
```
thetorpedodog, line 46: unrelated removal?

thetorpedodog (commented):
a few small notes which I missed last time. might also be worth just doing a full rebase so that all the changes relative to main are accurately reflected, and so we're sure about testing. otherwise good python-wise!

libtiledbsoma/src/reindexer/reindexer.h
```diff
@@ -0,0 +1,99 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @section DESCRIPTION
+ *
+ * This file defines the a pybind11 api into SOMA C++ library.
+ */
+
+#pragma once
```
nguyenv, line 33: I don't have a strong opinion on using `#pragma once` vs `#ifndef` header guards in general, but it would be more consistent in this repo to use the latter.

libtiledbsoma/test/test_indexer.cc
```diff
@@ -0,0 +1,132 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @section DESCRIPTION
+ *
+ * This file manages unit tests for column buffers
```
nguyenv, line 30: nit: Update description

libtiledbsoma/test/test_indexer.cc
```diff
@@ -0,0 +1,132 @@
+/**
+ * @file   test_indexer.cc
+ *
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2022 TileDB, Inc.
```
nguyenv, line 8: nit: 2024

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -0,0 +1,163 @@
+/**
+ * @file   reindexer.cc
+ *
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2022 TileDB, Inc.
```
nguyenv, line 8: nit: 2024

libtiledbsoma/src/reindexer/reindexer.h
```diff
@@ -0,0 +1,99 @@
+/**
+ * @file   reindexer.h
+ *
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2022 TileDB, Inc.
```
nguyenv, line 8: nit: 2024

libtiledbsoma/test/test_indexer.cc
```diff
@@ -0,0 +1,132 @@
...
+    indexer.map_locations(keys);
+    auto* hash = kh_init(m64);
+    int ret;
+    khint64_t k;
+
+    for (size_t i = 0; i < keys.size(); i++) {
+        k = kh_put(m64, hash, keys[i], &ret);
+        assert(k != kh_end(hash));
+        kh_val(hash, k) = i;
+    }
```
nguyenv, line 66:
```suggestion
    for (auto key& : keys) {
        k = kh_put(m64, hash, key, &ret);
        assert(k != kh_end(hash));
        kh_val(hash, k) = i;
    }
```

libtiledbsoma/test/test_indexer.py
nguyenv: Oh wait I'm looking at this file again, and I see what you're saying now. Why isn't this written for Pytest?

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -0,0 +1,163 @@
...
+
+    LOG_DEBUG(fmt::format(
+        "[Re-indexer] Start of Map locations with {} keys and {} threads",
+        size,
+        threads));
+    hash_ = kh_init(m64);
+    kh_resize(m64, hash_, size * 1.25);
+    LOG_DEBUG(
+        fmt::format("[Re-indexer] Thread pool started and hash table created"));
+    int ret;
```
nguyenv, line 74: What is `ret` for these `kh_*` functions? They get passed in as an argument but are unused. Is it the return code? Do we need to check that?

apis/python/src/tiledbsoma/utils.py
```diff
@@ -0,0 +1,24 @@
...
+from .options import SOMATileDBContext
+
+"""Builds an indexer object compatible with :meth:`pd.Index.get_indexer`."""
+
+
+def build_index(
+    keys: np.typing.NDArray[np.int64], context: SOMATileDBContext
+) -> clib.IntIndexer:
+    if len(np.unique(keys)) != len(keys):
+        raise pd.errors.InvalidIndexError(
```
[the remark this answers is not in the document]
beroy: The behavior of the indexer is modeled after Pandas indexer. That's why we decided to use this.

.github/workflows/r-ci.yml
```diff
@@ -51,6 +51,9 @@ jobs:
       - name: Bootstrap
         run: cd apis/r && tools/r-ci.sh bootstrap
 
+      - name: Install tiledb
+        run: cd apis/r && Rscript -e 'remotes::install_version("tiledb", "0.22.0")'
```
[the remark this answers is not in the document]
beroy: Yes

libtiledbsoma/src/reindexer/reindexer.cc
beroy: Yes, that's the return code. In our case we just check the return object.

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -0,0 +1,163 @@
...
+    for (int i = 0; i < size; i++) {
+        k = kh_put(
+            m64,
+            hash_,
+            keys[i],
+            &ret);  // hash_wrapper<IntType>::put(hash_, key, &ret);
+        assert(k != kh_end(hash_));
+        kh_val(hash_, k) = counter;
+        counter++;
+    }
```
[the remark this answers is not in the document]
nguyenv: oic now. Maybe if it's possible the pointers could be replaced with `std::array` but not necessary.

libtiledbsoma/test/test_indexer.cc
nguyenv: Nevermind about this. I missed that this was a pointer.

beroy: Actually this one is not a pointer but i is used in the loop so writing this way is more intuitive.

libtiledbsoma/test/test_indexer.py
```diff
@@ -0,0 +1,191 @@
...
+    },
+]
+
+
+def test_indexer():
+    for data in test_data:
+        indexer_test(data["keys"], data["lookups"], not data["pass"])
+
+
+if __name__ == "__main__":
```
nguyenv, line 190: Since you fixed this to run with Pytest now, this main function is not necessary. I do think in a future PR we should parameterize these inputs and test for exceptions using `pytest.raises` instead.

nguyenv: You don't need to call test_indexer - pytest will pick up that unit test automatically

libtiledbsoma/test/test_indexer.cc
nguyenv:
```suggestion
 * This file manages unit tests for the reindexer
```

nguyenv (approved):
Thanks for adding this and making the changes. I have one last minor change request, but I think it's good to go.

PR #1731, opened 2023-09-28
Depend on core 2.17.1, TileDB-Py 0.23.1, TileDB-R 0.21.1

johnkerl:
#866

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -40,6 +40,16 @@ jobs:
 
       - name: Install BioConductor package SingleCellExperiment
         run: cd apis/r && tools/r-ci.sh install_bioc SingleCellExperiment
+
+      # Given propagation delay for binaries at https://r2u.stat.illinois.edu/ubuntu/pool/dists/jammy/main/
+      - name: Install r-universe build of tiledb-r (macOS)
+        if: ${{ matrix.os == 'macOS-latest' }}
+        run: cd apis/r && Rscript -e "install.packages('tiledb', repos = c('https://eddelbuettel.r-universe.dev', 'https://cloud.r-project.org'))"
+
+      # Given propagation delay for binaries at https://r2u.stat.illinois.edu/ubuntu/pool/dists/jammy/main/
+      - name: Install r-universe build of tiledb-r (linux)
+        if: ${{ matrix.os != 'macOS-latest' }}
+        run: cd apis/r && Rscript -e "options(bspm.version.check=TRUE); install.packages('tiledb', repos = c('https://eddelbuettel.r-universe.dev/bin/linux/jammy/4.3/', 'https://cloud.r-project.org'))"  
        
       - name: Dependencies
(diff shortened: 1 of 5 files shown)


johnkerl:
IIUC R CI is failing b/c for source we have 0.21.1
https://cran.r-project.org/web/packages/tiledb/index.html
but for binary we have only 0.21.0
https://r2u.stat.illinois.edu/ubuntu/pool/dists/jammy/main/
```
[r-cran-tiledb_0.21.0-1.ca2204.1_amd64.deb](https://r2u.stat.illinois.edu/ubuntu/pool/dists/jammy/main/r-cran-tiledb_0.21.0-1.ca2204.1_amd64.deb)          15-Sep-2023 15:17            20268468
```

eddelbuettel:
One way is to add this again:

```yaml
       - name: Install r-universe build of tiledb-r (macOS)
         if: ${{ matrix.os == 'macOS-latest' }}
         run: cd apis/r && Rscript -e "install.packages('tiledb', repos = c('https://eddelbuettel.r-universe.dev', 'https://cloud.r-project.org'))"

       - name: Install r-universe build of tiledb-r (linux)
         run: cd apis/r && Rscript -e "options(bspm.version.check=TRUE); install.packages('tiledb', repos = c('https://eddelbuettel.r-universe.dev/bin/linux/jammy/4.3/', 'https://cloud.r-project.org'))"  
```

PPM did not run yesterday (unusual), and has not run today (can happen) so new official r2u binary of tiledb(-r) 0.21.1 (or any other updated CRAN packages).  For tiledb-r the r-universe alternate makes a good fix.

eddelbuettel:
@johnkerl We could indeed.  I was careful not to push it too much, not being a company resource etc pp.  We would then always test with 'master branch of tiledb-r' which would not be a bad test.  All this could be part of a broader discussion about how we may want to rejig the building and testing of various intertwined components here.

(What we really want is a magic wand or a more fault tolerant routine here.  "Try r2u if it has the binary, else build from source (slower) or resort to r-universe" is one (not coded-up but plausible) approach.  I think right we can through the fix back it because it is not part of the released package but really only part of what happens in GitHub Actions.  "And what happens in GitHub Actions stays in GitHub Actions....".   Otherwise lame that r2u "prod" is delayed but such is life.)

johnkerl:
Thanks @eddelbuettel -- I think I get it now. (We've discussed such things before, but, non-R-expert that I am, it can take a while to "sink in".

> We would then always test with 'master branch of tiledb-r' which would not be a bad test

I think this is not a state we want to remain in TBH. We should be testing against tagged-and-released software.

This mod is only appropriate when we _are_ testing tagged-and-released software (b/c we know within our heads) and are _only_ working around a binary-artifact propagation delay which is (a) too long, and (b) completely out of our control.

I'll put reverting this mod on my to-do list for a subsequent PR.

Thanks again for explaining (again)! :)

eddelbuettel (approved):
:shipit:

PR #1738, opened 2023-09-29
[ci/r] Use TileDB-R 0.21.1 binaries

johnkerl:
#866

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -42,10 +42,15 @@ jobs:
         run: cd apis/r && tools/r-ci.sh install_bioc SingleCellExperiment
 
       # The next two stanzas are necessary given propagation delay for binaries at
-      #   https://r2u.stat.illinois.edu/ubuntu/pool/dists/jammy/main/
-      # which occurs when source for tiledb-r version-next is available but binaries are not yet.
-      # To check if binaries are available:
-      #   docker run --rm -ti rocker/r2u Rscript -e 'install.packages("tiledb")'
+      # https://r2u.stat.illinois.edu/ which may lack binaries (for short periods of usually a day)
+      # when sources have been updated. The default installation could switch to installation from
+      # sources --- but that would require installing all build dependencies. To see what the most
+      # recent binary is run e.g.  docker run --rm -ti rocker/r2u bash -c 'apt update -qq &&
+      # apt-cache show r-cran-tiledb' Using the r-universe builds (as below) is a suitable fallback
+      # as they update more frequently than CRAN.
+
+      # Uncomment these next two stanzas as needed whenever we've just released a new tiledb-r for
+      # which source is available but binaries are not yet:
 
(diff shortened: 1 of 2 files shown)


.github/workflows/r-ci.yml
```diff
@@ -41,15 +41,20 @@ jobs:
...
-      # Given propagation delay for binaries at https://r2u.stat.illinois.edu/ubuntu/pool/dists/jammy/main/
-      - name: Install r-universe build of tiledb-r (linux)
-        if: ${{ matrix.os != 'macOS-latest' }}
-        run: cd apis/r && Rscript -e "options(bspm.version.check=TRUE); install.packages('tiledb', repos = c('https://eddelbuettel.r-universe.dev/bin/linux/jammy/4.3/', 'https://cloud.r-project.org'))"  
+      # The next two stanzas are necessary given propagation delay for binaries at
+      #   https://r2u.stat.illinois.edu/ubuntu/pool/dists/jammy/main/
+      # which occurs when source for tiledb-r version-next is available but binaries are not yet.
+      # To check if binaries are available:
+      #   docker run --rm -ti rocker/r2u Rscript -e 'install.packages("tiledb")'
+
```
eddelbuettel, line 49: I just found the preceding block (just visually) very hard to read and mistook the comment (which may or may not belong here) as a code snippet, which it is not.  Maybe double `##` or more indentation may separate it visually.

On the margin, documentation like this should be in the wiki or discussion section or notion or dropbox.  

If you are sure documentation belongs here (no strongly feeling myself, but I would be remove it) may I suggest added  whitespace as well as expanded content.  A suggestion follows.

```
      #   The next two stanzas are necessary given propagation delay for binaries at
      #   https://r2u.stat.illinois.edu/ which may lack binaries (for short periods of
      #   usually a day) when sources have been updated. The default installation could
      #   switch to installation from sources---but that would require installing all
      #   build dependencies.  The see what the most recent binary is run e.g.
      #   docker run --rm -ti rocker/r2u bash -c 'apt update -qq && apt-cache show r-cran-tiledb'
      #   Using the r-universe builds (as below) is a suitable fallback as they update more
      #   frequently than CRAN
```

.github/workflows/r-ci.yml
```diff
@@ -41,15 +41,20 @@ jobs:
...
+      #- name: Install r-universe build of tiledb-r (macOS)
+      #  if: ${{ matrix.os == 'macOS-latest' }}
+      #  run: cd apis/r && Rscript -e "install.packages('tiledb', repos = c('https://eddelbuettel.r-universe.dev', 'https://cloud.r-project.org'))"
+
+      #   docker run --rm -ti rocker/r2u Rscript -e 'install.packages("tiledb")'
```
eddelbuettel, line 54: That line seems superfluous too.

eddelbuettel (approved):
Fine in principle. I would change the wording as suggested but do what you feel appropriate.

PR #1741, opened 2023-09-29
Update tools used in pre-commit testing.

thetorpedodog:
This updates us to the latest versions of Black and Ruff, which results in some minor non-behavioral changes to the code.


bkmartinjr (approved):
LGTM!

Issue #1739, opened 2023-09-29
[python/r] Enumerations: check categorical-at-schema vs flat-at-write case


johnkerl:
Given enumeration support (#866) we need to decide a policy for the case when a dataframe schema has a categorical column at create time but a flat-vector column is presented at write time.

* Probably throw
* Needs discussion
* Needs implementation and unit tests
* Needs docstring documetation

PR #1748, opened 2023-09-29
[python] Check for downgraded categorical write

johnkerl:
**Issue and/or context:** #1739

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -411,23 +411,27 @@ class DataFrame(TileDBArray, somacore.DataFrame):
             n = len(col)
 
             cols_map = dim_cols_map if name in dim_names_set else attr_cols_map
-            if pa.types.is_dictionary(col.type) and col.num_chunks != 0:
-                if name in dim_names_set:
-                    # Dims are never categorical. Decategoricalize for them.
-                    cols_map[name] = pa.chunked_array(
-                        [chunk.dictionary_decode() for chunk in col.chunks]
-                    )
-                else:
-                    attr = self._handle.schema.attr(name)
-                    if attr.enum_label is not None:
-                        # Normal case: writing categorical data to categorical schema.
-                        cols_map[name] = col.chunk(0).indices.to_pandas()
-                    else:
-                        # Schema is non-categorical but the user is writing categorical.
-                        # Simply decategoricalize for them.
+            if pa.types.is_dictionary(col.type):
+                if col.num_chunks != 0:
+                    if name in dim_names_set:
+                        # Dims are never categorical. Decategoricalize for them.
                         cols_map[name] = pa.chunked_array(
                             [chunk.dictionary_decode() for chunk in col.chunks]
                         )
+                    else:
+                        attr = self._handle.schema.attr(name)
+                        if attr.enum_label is not None:
+                            # Normal case: writing categorical data to categorical schema.
+                            cols_map[name] = col.chunk(0).indices.to_pandas()
+                        else:
+                            # Schema is non-categorical but the user is writing categorical.
+                            # Simply decategoricalize for them.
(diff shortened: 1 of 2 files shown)


apis/python/tests/test_dataframe.py
```diff
@@ -139,11 +139,17 @@ def test_dataframe_with_enumeration(tmp_path):
...
+            data["foo"] = ["a", "bb", "ccc", "bb", "a"]
+            data["bar"] = ["cat", "dog", "cat", "cat", "cat"]
+            with pytest.raises(ValueError):
+                sdf.write(pa.Table.from_pydict(data))
+        else:
+            data["foo"] = pd.Categorical(["a", "bb", "ccc", "bb", "a"])
+            data["bar"] = pd.Categorical(["cat", "dog", "cat", "cat", "cat"])
+            sdf.write(pa.Table.from_pydict(data))
+            assert sdf.enumeration("foo") == enums["enmr1"]
+            assert sdf.enumeration("bar") == enums["enmr2"]
```
nguyenv, line 152: I don't think this needs to be parameterized.

```suggestion
          data["foo"] = ["a", "bb", "ccc", "bb", "a"]
          data["bar"] = ["cat", "dog", "cat", "cat", "cat"]
          with pytest.raises(ValueError):
              sdf.write(pa.Table.from_pydict(data))

          data["foo"] = pd.Categorical(["a", "bb", "ccc", "bb", "a"])
          data["bar"] = pd.Categorical(["cat", "dog", "cat", "cat", "cat"])
          sdf.write(pa.Table.from_pydict(data))
          assert sdf.enumeration("foo") == enums["enmr1"]
          assert sdf.enumeration("bar") == enums["enmr2"]
```

johnkerl:
@nguyenv ready for round two!

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -388,6 +388,11 @@ def write(
                 containing all columns, including the index columns. The schema for the values must
                 match the schema for the :class:`DataFrame`.
 
+                If a column is of categorical type in the schema and a flattened/non-categorical
+                column is presented for data on write, a ``ValueError`` is raised.  If a column is
+                of non-categorical type in the schema and a categorical column is presented for data
+                on write, the data are auto-flattened on behalf of the user.
```
bkmartinjr, line 394: I don't think the term `auto-flattened` is a good way to express the operation (the term is used in Pandas to imply index flattening, etc).

Suggestion:

"..., the data are written as an array of category values, and the category type information is not saved."

Or something like that (my suggestion is also not quite right - hoping we can iterate)

apis/python/tests/test_dataframe.py
```diff
@@ -122,15 +122,15 @@ def test_dataframe_with_float_dim(tmp_path, arrow_schema):
     assert sdf.index_column_names == ("bar",)
 
 
-def test_dataframe_with_enumeration(tmp_path):
+@pytest.mark.parametrize("downgrade", [False, True])
```
bkmartinjr, line 125: what is the `downgrade` param?

bkmartinjr (changes requested):
Mostly nits.  
* Not sure why there are R CI changes included in this PR
* Docstring suggestion (imperfect suggestion)
* Nit on unit test
Small stuff!

apis/python/tests/test_dataframe.py
johnkerl: oops went away on previous code-review-feedback param -- will remove

apis/python/src/tiledbsoma/_dataframe.py
johnkerl: > hoping we can iterate

Of course!!

nguyenv (approved):

bkmartinjr (approved):

Issue #1746, opened 2023-09-29
Enumerations-API refinement


johnkerl:
As of 1.5.0rc0, `SOMADataFrame.create` takes three new arguments, and should not.

* `create` needs to take no new arguments
  * Those three values must be inferred from the passed-in Arrow schema
* `write` will present new levels

Core support:

* Permit zero levels at `create` (currently >= 1 is required)
* Offer a streamlined `extend_enumeration` API

Out of scope and separately tracked: bit-width mutation for when the level-count passes the 8-bit or 16-bit boundary: #1745

This is multi-developer work involving @davisp @nguyenv @eddelbuettel @johnkerl ; mainly this issue mirrors [sc-34800].

Additional reminder: R/Python interop cases within `apis/system/tests`.

johnkerl:
[sc-34800]

johnkerl:
Sub-issue: #1867

PR #1755, opened 2023-10-04
[python] Enumerations-API refinement

nguyenv:
**Issue and/or context:**

#1746

**Changes:**

- Remove enumeration related arguments from `DataFrame.create`
- Add enumeration values on first write
- Depends on https://github.com/TileDB-Inc/TileDB-Py/pull/1847; passing locally when built against this branch

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -414,19 +414,24 @@ class DataFrame(TileDBArray, somacore.DataFrame):
                 if attr.enum_label is not None and col.num_chunks != 0:
                     if not pa.types.is_dictionary(col_info.type):
                         raise ValueError(
-                            f"Expected dictionary type for enumerated attribute {name} but saw {col_info.type}"
+                            "Expected dictionary type for enumerated attribute "
+                            "{name} but saw {col_info.type}"
                         )
 
                     enmr = self._handle.enum(attr.name)
 
-                    # extend only with new values
-                    old_vals = set(enmr.values())
-                    new_vals = set(col.chunk(0).dictionary.tolist())
-                    update_vals = list(new_vals - old_vals)
+                    # get new enumeration values
+                    update_vals = []
+                    for new_val in col.chunk(0).dictionary.tolist():
+                        if new_val not in enmr.values():
+                            update_vals.append(new_val)
 
-                    se = tiledb.ArraySchemaEvolution(self.context.tiledb_ctx)
-                    se.extend_enumeration(enmr.extend(update_vals))
-                    se.array_evolve(uri=self.uri)
+                    # only extend if there are new values
+                    if update_vals:
+                        se = tiledb.ArraySchemaEvolution(self.context.tiledb_ctx)
+                        new_enmr = enmr.extend(update_vals)
+                        se.extend_enumeration(new_enmr)
+                        se.array_evolve(uri=self.uri)
 
             cols_map = dim_cols_map if name in dim_names_set else attr_cols_map
             if pa.types.is_dictionary(col.type):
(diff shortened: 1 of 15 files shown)


apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -122,7 +122,19 @@ class DataFrame(TileDBArray, somacore.DataFrame):
...
+        self,
+        handle: ArrayWrapper,
+        *,
+        _dont_call_this_use_create_or_open_instead: str = "unset",
+    ):
+        self._first_write = True
+        super().__init__(
+            handle,
+            _dont_call_this_use_create_or_open_instead="tiledbsoma-internal-code",
+        )
```
nguyenv, line 137: @thetorpedodog I wanted your feedback on this addition I just added since you designed these classes.

On the first write to the DataFrame, we check if the PyArrow schema contains any dictionaries and update the TileDB ArraySchema with the enumeration values from the PyArrow array. I am keeping track of this by checking `self._first_write` which I now initialize in the `DataFrame` constructor. Upon first write, `self._first_write` gets toggled to False. I'm wondering if what I added here looks OK or if there's a better way to do what I'm trying to do.

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -122,7 +122,19 @@ class DataFrame(TileDBArray, somacore.DataFrame):
         it must be ``None``.
     """
 
-    __slots__ = ()
+    __slots__ = "_first_write"
```
thetorpedodog, line 125: nit: `("_first_write",)`

(I have also done this exact thing myself, so)

apis/python/src/tiledbsoma/_dataframe.py
thetorpedodog: This seems like as good a way to go about this as I could think of. I might make a minor change to naming (maybe call it something like `_enumerations_written`, which goes from False to True?) but overall it makes sense and is reasonably simple.

I have one question which is more of a practical concern: how does this interact with it when there is existing data in the already-written part of the DataFrame? Do we just always assume that it existing data will have a compatible enum schema with new data? Since I am only passingly familiar with how enumerations work, I don’t have a strong opinion about this; basically provided this is something that you’ve thought about, whatever the answer is is fine by me.

nguyenv: > how does this interact with it when there is existing data in the already-written part of the DataFrame?

That's an excellent point. I think I'll need something that checks whether this is the first write to a newly created DataFrame.

nguyenv:
`catch_clara.cc` file failure in C++ build? Not related to changes in this PR. Will investigate tomorrow morning.

johnkerl:
@nguyenv this is failing CI because it's not on top of #1838

nguyenv:
There is actually two more corrections I need to do. I already solved one. If there are no new enumeration values, then do not evolve the schema (before, we were extending with an empty list). But there's one more issue having to do with Boolean enums.

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -776,22 +786,35 @@ def _build_tiledb_schema(
...
+    for pa_attr in schema:
+        attr_name = pa_attr.name
+
+        if attr_name in index_column_names:
+            continue
+
+        has_enum = pa.types.is_dictionary(pa_attr.type)
+
+        if has_enum:
+            pa.types.is_integer(pa_attr.type.value_type)
```
johnkerl, line 801: @nguyenv it looks like this had been

```
something = pa.types.is_integer(pa_attr.type.value_type)
```

but then `ruff` saw the `something` was unused & removed it

anyway, i don't know if you want to omit this line entirely, or assign & use the return value, but just heads-up either way (i just happened to notice this while reading)

nguyenv: Thanks for catching that. I have now removed it. 

I don't know how I didn't catch this last week, but I'm still having issues with extending Boolean types. What really weird is that if I use `int64` as the dictionary value in schema and write the values as `int64` instead of `bool`, it works. That doesn't make any sense though.

Issue #1742, opened 2023-09-29
[r/python] Improve query performance for `in` conditions having many terms


johnkerl:
Tracking for [sc-24310] and https://github.com/chanzuckerberg/cellxgene-census/issues/36

PR #1756, opened 2023-10-04
[python] Support new set-membership query condition

johnkerl:
**Issue and/or context:** This is an adaptation of PR https://github.com/TileDB-Inc/TileDB-Py/pull/1837 for issue https://github.com/single-cell-data/TileDB-SOMA/issues/1742

### DIFF
--- a/apis/python/src/tiledbsoma/_query_condition.py
+++ b/apis/python/src/tiledbsoma/_query_condition.py
@@ -240,8 +240,9 @@ class QueryConditionTree(ast.NodeVisitor):
             else:
                 dt = self.array.schema.attr_or_dim_dtype(variable)
 
-            # sdf.read(column_names=["foo"], value_filter='bar == 999') should result in bar being
-            # added to the column names. See also https://github.com/single-cell-data/TileDB-SOMA/issues/755
+            # sdf.read(column_names=["foo"], value_filter='bar == 999') should
+            # result in bar being added to the column names. See also
+            # https://github.com/single-cell-data/TileDB-SOMA/issues/755
             att = self.get_att_from_node(node.left)
             if att not in self.query_attrs:
                 self.query_attrs.append(att)
@@ -354,6 +355,9 @@ class QueryConditionTree(ast.NodeVisitor):
                 )
             raise tiledb.TileDBError(f"Attribute `{att}` not found in schema.")
 
+        # sdf.read(column_names=["foo"], value_filter='bar == 999') should
+        # result in bar being added to the column names. See also
+        # https://github.com/single-cell-data/TileDB-SOMA/issues/755
         if att not in self.query_attrs:
             self.query_attrs.append(att)
 
--- a/apis/python/src/tiledbsoma/_query_condition.py
+++ b/apis/python/src/tiledbsoma/_query_condition.py
@@ -434,10 +434,11 @@ class QueryConditionTree(ast.NodeVisitor):
 
         create_fn_name = f"create_{dtype}"
 
-        if not hasattr(clib.PyQueryCondition, create_fn_name):
+        try:
+            return getattr(clib.PyQueryCondition, create_fn_name)
+        except AttributeError as ae:
             raise tiledb.TileDBError(f"PyQueryCondition.{create_fn_name}() not found.")
 
-        return getattr(clib.PyQueryCondition, create_fn_name)
 
     def visit_BinOp(self, node: ast.BinOp) -> clib.PyQueryCondition:
         try:
--- a/apis/python/src/tiledbsoma/_query_condition.py
+++ b/apis/python/src/tiledbsoma/_query_condition.py
@@ -436,9 +436,8 @@ class QueryConditionTree(ast.NodeVisitor):
 
         try:
             return getattr(clib.PyQueryCondition, create_fn_name)
-        except AttributeError as ae:
+        except AttributeError:
             raise tiledb.TileDBError(f"PyQueryCondition.{create_fn_name}() not found.")
-
 
     def visit_BinOp(self, node: ast.BinOp) -> clib.PyQueryCondition:
         try:
--- a/apis/python/src/tiledbsoma/_query_condition.py
+++ b/apis/python/src/tiledbsoma/_query_condition.py
@@ -136,8 +136,9 @@ class QueryCondition:
         config: Optional[dict],
         timestamps: Optional[Tuple[OpenTimestamp, OpenTimestamp]],
     ):
+        ctx = tiledb.Ctx(config)
         qctree = QueryConditionTree(
-            tiledb.open(uri, ctx=tiledb.Ctx(config), timestamp=timestamps), query_attrs
+            ctx, tiledb.open(uri, ctx=ctx, timestamp=timestamps), query_attrs
         )
         self.c_obj = qctree.visit(self.tree.body)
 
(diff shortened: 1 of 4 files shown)


johnkerl:
The unit-test case is failing because of this, needing to be fixed:

Before:

```
>>> uri = 'tiledb://johnkerl-tiledb/oron-var'
>>> sdf = tiledbsoma.DataFrame.open(uri)
NameError: name 'tiledbsoma' is not defined
>>> import tiledbsoma
>>> sdf = tiledbsoma.DataFrame.open(uri)
>>> sdf.read(
...     column_names=['feature_name'],
...     value_filter='feature_id in ["ENSG00000148584"]',
... ).concat().to_pandas()
  feature_name
0         A1CF
```

After:

```
>>> uri = 'tiledb://johnkerl-tiledb/oron-var'
>>> sdf = tiledbsoma.DataFrame.open(uri)
NameError: name 'tiledbsoma' is not defined
>>> import tiledbsoma
>>> sdf = tiledbsoma.DataFrame.open(uri)
>>> sdf.read(
...     column_names=['feature_name'],
...     value_filter='feature_id in ["ENSG00000148584"]',
... ).concat().to_pandas()
  feature_name       feature_id
0         A1CF  ENSG00000148584
```

That is to say, we need to preserve the feature that the queried-for attribute name gets added to the list of output attrs.

johnkerl:
(BTW the aforementioned unit-test case has since been fixed)

apis/python/tests/test_experiment_query.py
```diff
@@ -182,6 +182,37 @@ def test_experiment_query_value_filter(soma_experiment):
         assert query.var().concat()["label"].to_pylist() == var_label_values
 
 
+@pytest.mark.parametrize("n_obs,n_vars", [(1001, 99)])
```
nguyenv, line 185:
```suggestion
```

Unused.

johnkerl: @nguyenv actually the pre-existing `in` case and my new `not in` case both _fail_ if this parametrization line is removed from their respective test functions

This is because the `soma_experiment` fixture implicitly relies on them as arguments

apis/python/src/tiledbsoma/_query_condition.py
```diff
@@ -136,8 +136,9 @@ def init_query_condition(
         config: Optional[dict],
         timestamps: Optional[Tuple[OpenTimestamp, OpenTimestamp]],
     ):
+        ctx = tiledb.Ctx(config)
         qctree = QueryConditionTree(
-            tiledb.open(uri, ctx=tiledb.Ctx(config), timestamp=timestamps), query_attrs
+            ctx, tiledb.open(uri, ctx=ctx, timestamp=timestamps), query_attrs
```
johnkerl, line 141: @nguyenv @thetorpedodog the `ctx` is essential for the call chain

However this seems to be a back-pedal on our progress with #1169 and in particular #1664

I note that `init_query_condition` is called only from here
https://github.com/single-cell-data/TileDB-SOMA/blob/1.5.0rc1/apis/python/src/tiledbsoma/pytiledbsoma.cc#L160
where there is no user-specified context available (this is called at module-load time)

Thoughts?

nguyenv: The `QueryCondition` binding code is in dire need of rewriting. It utilizes an old method we were using with Pybind11 where we did not bind from core directly but have this unnecessary `PyQueryCondition` class as an inbetween that gets initialized with `init_query_condition`. We now bind everything directly but have not gone back yet to update `QueryCondition`.

johnkerl: @thetorpedodog `init_query_condition` is called here
https://github.com/single-cell-data/TileDB-SOMA/blob/1.5.0rc1/apis/python/src/tiledbsoma/pytiledbsoma.cc#L160
and implemented here
https://github.com/single-cell-data/TileDB-SOMA/blob/1.5.0rc1/apis/python/src/tiledbsoma/query_condition.cc#L140

@thetorpedodog I don't truly understand `init_query_condition` either (yet, anyway)

@nguyenv in terms of merging this PR -- which as-is will cause a thread-count regression, defeating some work on #1664 / #1169 -- can we set up a call to develop a plan to design and accomplish the necessary refactor within a 1.5.0 timeframe?

eddelbuettel: The API had two `init()` calls for `QueryCondition()` in lines 96 to 169 (if you include comments) of [tiledb/sm/cpp_api/query_condition.h](https://github.com/TileDB-Inc/TileDB/blob/d497fffbaba54cf66eb366f696fc7b132b9954f5/tiledb/sm/cpp_api/query_condition.h#L96-L169), one of these take a `ctx`.

nguyenv: > in terms of merging this PR -- which as-is will cause a thread-count regression, defeating some work on https://github.com/single-cell-data/TileDB-SOMA/pull/1664 / https://github.com/single-cell-data/TileDB-SOMA/issues/1169 -- can we set up a call to develop a plan to design and accomplish the necessary refactor within a 1.5.0 timeframe?

Yes we can discuss this tomorrow. I do not think that updating the `QueryCondition` bindings will take more than one days worth of work.

johnkerl:
CI fail
```
E   ImportError: cannot import name 'SparseDataset' from 'anndata._core.sparse_dataset' (/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/anndata/_core/sparse_dataset.py)
```
as AnnData 0.10.0 dropped today:
https://anndata.readthedocs.io/en/latest/#version-0-10

johnkerl:
Before:
```
anndata._core.sparse_dataset.SparseDataset
```

After:
```
anndata._core.sparse_dataset.CSRDataset
anndata._core.sparse_dataset.CSRDataset
```

johnkerl:
This is stacked atop the blocking PR #1766 until that PR is merged, at which point this PR will be on `main`.

apis/python/src/tiledbsoma/_query_condition.py
johnkerl: #1775 has been merged in, resolving these issues

johnkerl:
Update: this PR was a very simple copy of https://github.com/TileDB-Inc/TileDB-Py/pull/1837

Along the way we ran into the following, which were fixed:

* #1775 
* #1784 

So this should now be a very simple approve.

nguyenv (approved):

apis/python/src/tiledbsoma/_query_condition.py
```diff
@@ -405,6 +420,20 @@ def init_pyqc(self, pyqc: clib.PyQueryCondition, dtype: str) -> Callable:
...
+                dtype = "int64"
+            elif np.issubdtype(dtype, bool):
+                dtype = "uint8"
+
+        create_fn_name = f"create_{dtype}"
+
+        try:
+            return getattr(clib.PyQueryCondition, create_fn_name)
+        except AttributeError:
+            raise tiledb.TileDBError(f"PyQueryCondition.{create_fn_name}() not found.")
```
thetorpedodog, line 435: you might want to do `from None` or `AttributeError as ae: ... from ae` here.

thetorpedodog (approved):
one minor thing that you might want to address before submitting but not a showstopper.

PR #1775, opened 2023-10-10
[python] Create separate `set_condition`

nguyenv:
**Issue and/or context:** 

Issue #1742. Necessary prerequisite for #1756.

**Changes:**

- Separate `set_condition` from `SOMAArray` constructor
- `init_query_condition` function no longer takes `config` argument; now replaced by `ArraySchema` and a mapping of enumeration labels to NumPy datatype
- Rename `query_condition.cc` module loader to `load_query_condition`
- `PyQueryCondition::combine` static method no longer takes a ctx argument. Instead it uses the global ctx created by the `PyQueryCondition` constructor

### DIFF
--- a/apis/python/src/tiledbsoma/_query_condition.py
+++ b/apis/python/src/tiledbsoma/_query_condition.py
@@ -131,12 +131,11 @@ class QueryCondition:
 
     def init_query_condition(
         self,
+        ctx: tiledb.Ctx,
         uri: str,
-        query_attrs: List[str],
-        config: Optional[dict],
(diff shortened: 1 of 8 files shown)


nguyenv:
The macOS R unit test failure is fixed by #1779. I attempted to rebase this branch against main earlier but pulled in a lot of changed files. It would be easier to merge this branch into #1756 as-is and then have #1756 rebase against main.

johnkerl (approved):
🚢 🎉

Issue #1765, opened 2023-10-09
[python] Temporarily pin to AnnData 0.9.*


johnkerl:
As of Friday all Python CI is failing like this:
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/6434852473/job/17474905281

```
E   ImportError: cannot import name 'SparseDataset' from 'anndata._core.sparse_dataset' (/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/anndata/_core/sparse_dataset.py)
```

This is a nuanced issue with a nuanced solution in discussion between @nguyenv , @thetorpedodog , and myself.

It will be dealt with soon, however, to unblock CI and high-pri PRs over the next day or two I will _temporarily_ pin to AnnData 0.9. This issue will be kept open for tracking until the issue is dealt with properly.

johnkerl:
Closing in favor of #1785 for longer-term tracking

PR #1784, opened 2023-10-11
[python] Use AnnData 0.10.1

johnkerl:
**Issue and/or context:** As in #1765. There is now a 0.10.1 which solves the problem (we were not the only ones to suffer). See also #1785 for longer-term follow-up (next year).

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -269,7 +269,8 @@ setuptools.setup(
     install_requires=[
         # Needed for Python 3.7 which anndata 0.9 doesn't support but we do
         "anndata < 0.9; python_version<'3.8'",
-        "anndata; python_version>='3.8'",
+        # Tracked in https://github.com/single-cell-data/TileDB-SOMA/issues/1785
+        "anndata != 0.10.0; python_version>='3.8'",
         "attrs>=22.2",
         "numba~=0.58.0; python_version>='3.8'",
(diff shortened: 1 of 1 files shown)


apis/python/setup.py
```diff
@@ -269,9 +269,7 @@ def run(self):
     install_requires=[
         # Needed for Python 3.7 which anndata 0.9 doesn't support but we do
         "anndata < 0.9; python_version<'3.8'",
-        # Temporary until we align with a compatibility issue in AnnData 0.10.0.
-        # This is tracked in https://github.com/single-cell-data/TileDB-SOMA/issues/1765.
-        "anndata < 0.10.0; python_version>='3.8'",
+        "anndata; python_version>='3.8'",
```
thetorpedodog, line 272: use `anndata != 0.10.0` to avoid hitting the incompatible version

thetorpedodog (approved):

bkmartinjr (commented):
is there a need to make the same change to the somacore dependencies?

johnkerl:
@bkmartinjr no

johnkerl:
Not affected AFAIK

thetorpedodog:
> Not affected AFAIK

This is correct; we don’t play with any anndata internals in somacore, so it has no restrictions in and of itself.

PR #1792, opened 2023-10-16
[python] Blockwise SciPy sparse and Arrow Table iterator

### LINKED ISSUE #1503: [Feature request] CSR/CSC iterator over `ExperimentAxisQuery.X`
It is a very common pattern to iterate of an X layer slice by rows, where the slice is specified by an ExperimentAxisQuery, and the consumer wants the slices in a CSR/CSC format for computation (e.g., scipy.csr_matrix). The current API only provides COO iteration and does not provide direct reindexing support of joinids.

Propose an iterator API that:
* reads slices of a given size
* reindexes the joinids
* returns for each step the joinids and reindexed sparse matrix

In Python type sigs, a row-based iterator  method on ExperimentAxisQuery might look like:

```python
_RT = Tuple[Tuple[npt.NDArray[np.int64], npt.NDArray[np.int64]], sparse.spmatrix]

def X_sparse_iter(
    self: soma.ExperimentAxisQuery,
    X_name: str = "raw",  # the X layer to read
    row_stride: int = 2**16,  # row stride for each step
    fmt: Literal["csr", "csc"] = "csr",  # the resulting sparse format
) -> Iterator[_RT]
```

Example usage:
```python

with experiment.axis_query(...) as query:
    for (obs_joinids, var_join_ids), X_chunk in query.X_sparse_iter(X_name="raw"):
        ...
```
In this case, obs_joinids[i] and var_joinids[j] corresponds to X[i,j].

I have a fully functional and working _prototype_ implementation available [here](https://github.com/chanzuckerberg/cellxgene-census/blob/main/api/python/cellxgene_census/src/cellxgene_census/experimental/util/_csr_iter.py).  Important:
* The prototype does not utilize the ExperimentAxisQuery fast csr conversion, which would be necessary work for a "real" implementation.
* The prototype does lazy multi-threaded pipelining of the steps, which is essential for performance on the queries we typical do on the Census. 

There is a notebook in the same directory that shows example usage.

bkmartinjr:
**Issue and/or context:**

Fixes #1503

**Changes:**

Added blockwise SciPy and Arrow Table iterators for the SparseNDMatrix class.  Also usable form from ExperimentAxisQuery.X() method.  Example usage:

```python
uri = "s3://cellxgene-data-public/cell-census/2023-10-09/soma/"
context = soma.options.SOMATileDBContext(
    tiledb_config={
        "soma.init_buffer_bytes": 1 * 1024**3,
        "vfs.s3.region": "us-west-2",
    }
)
with soma.open(uri, context=context) as census:
    exp = census["census_data"]["homo_sapiens"]
    with exp.axis_query(
        measurement_name="RNA",
        obs_query=soma.AxisQuery(value_filter="tissue in ['pancreas', 'liver']"),
    ) as query:
        for sp, (obs_joinids, var_joinids) in query.X("raw").blockwise(axis=0).scipy():
            print(repr(sp))
```

produces:
```text
<65536x60664 sparse matrix of type '<class 'numpy.float32'>'
        with 77510135 stored elements in Compressed Sparse Row format>
<65536x60664 sparse matrix of type '<class 'numpy.float32'>'
        with 72742745 stored elements in Compressed Sparse Row format>
<65536x60664 sparse matrix of type '<class 'numpy.float32'>'
        with 110708131 stored elements in Compressed Sparse Row format>
...
```

The API is implemented on SparseNDArray (not ExperimentAxisQuery), and can be used directly on the sparse array, e.g.,
```python
with soma.open(an_array_uri) as A:
    for csr_matrix, (obs_joinids, var_jonids) in A.read().blockwise(axis=0).scipy(compress=True):
        ...
```

In addition to the SciPy blockwise iterator, you can also blockwise iterate over Arrow Tables, with all the same controls over axis, reindexing, step size, etc.
```python
In [6]: for tbl, indices in query.X("raw").blockwise(axis=0).tables(): print(tbl)
pyarrow.Table
soma_dim_0: int64
soma_dim_1: int64
soma_data: float
----
soma_dim_0: [[0,0,0,0,0,...,64270,64271,64976,65177,65263]]
soma_dim_1: [[27,33,43,46,58,...,50593,49918,50593,50593,50593]]
soma_data: [[1,2,17,1,2,...,1,1,1,2,1],[1,3,1,1,2,...,1,1,1,1,1]]
pyarrow.Table
soma_dim_0: int64
soma_dim_1: int64
soma_data: float
----
soma_dim_0: [[0,0,0,0,0,...,50062,50128,50151,50164,50255]]
soma_dim_1: [[4,7,20,21,27,...,30934,30934,30934,30934,30934]]
soma_data: [[1,1,1,1,1,...,1,1,1,1,1]]
```

The docstrings have more detail, but there are several parameters that control what is emitted by the iterator:
* compress: bool - if False, create a COO, else create a compressed sparse matrix
* axis: int - if zero (0), step over the rows and generate a CSR or COO (dependent on compress param).  If one (1), generate CSC or COO.
* size: int - number of coordinates on the blockwise axis, per iterator step

For example, to produce a CSC while stepping column (var)-wise:
> A.read().blockwise(axis=1).scipy(compress=True)

Or a COO while stepping row (obs)-wise:
> A.read().blockwise(axis=0).scipy(compress=False)

**Notes for Reviewer:**

1. Minor:  I added a .flake8 config, matching our black config for those of us who run flake8 in our IDEs.
2. The API works equally well with ExperimentAxisQuery.X, but as this is just a wrapper on he SparseNDArray.read, I implemented the bulk of the tests against SparseNDArray directly.


apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -442,4 +464,134 @@ def tables(self) -> TableReadIter:
...
+                determines the return type. If compressed==True and axis==0, a CSR
+                matrix is returned. If compressed==True and axis==1, a CSC matrix
+                is returned.
+            step:
+                Number of rows or columns to return in each iterator step.
+            compress:
+                If True, a CSC or CSR matrix is returned, dependent on the value of the
+                `axis` argument. If False, a COO matrix is returned.
+            reindex_sparse_axis:
+                If False (default), the sparse axis will also be reindexed from soma_joinid
```
atarashansky, line 546: The default value for `reindex_sparse_axis` looks to be True.

I think you got the True/False reversed?
"If True (default), the sparse axis will also be reindexed from soma_joinid to zero-based indices. If False, the sparse axis values will remain joinids."

bkmartinjr: Yes, I inverted the condition in the docstring. Thanks for catching.

When `compress==True`:
* Axis being iterated over is the `compressed` axis
* Axis *not* being iterated over is the `sparse` axis
The reason for this nomenclature is that these formats only compress coordinates on one axis (rows for CSR, ...)

So:
* axis==0 implies compressed axis is obs/row
* axis==1 implies compressed axis is var/col

thetorpedodog:
Overall shape of the API looks good. I have some implementation-level suggestions but will save that for Actual Review Time.

bkmartinjr:
> it may be too easily missed that this is the way to page through complete rows/columns 

IMHO, there is no reason we can't add the same (non-ragged) approach for Table iterators.  I just felt that it was a separate feature request, so I deferred it for now.  My implementation would facilitate it easily, as there is a stand-alone iterator already implemented which provides exactly this functionality for 2D arrays - just needs to be extended to nD, which I believe is straight forward.

@aaronwolen @pablo-gar - do you guys want to see this enhancement in this PR?  I can add, or we can push to future.  It would likely look something like:
```python
# generate ragged (current implementation)
A.read().tables()

# generate row-major non-ragged
A. read().tables(axis=0, step=10000)

# generate col-major non-ragged
A.read().tables(axis=1, step=10000)
```

The same constraints on the compatibility of `result_order` and axis would exist as with the `scipy` iterator.

bkmartinjr:
Regarding:

> splitting the scipy() method into a few others with individually simpler signatures and more-explicit names like scipy_sparse_rows() and so on...

I prefer the single signature as the defaults are almost always sufficient (in my experience and past usage).  That said, I am very happy to change this if we have an alternative consensus.  The other option is to simply add aliases (sugar) which call the base (most flexible) variant.

pablo-gar:
## Potential issues

I've had some time to digest it. Current potential issues I see:

- As @mlin mention there may be confusion as to which is "the" way to row/column pagination. _Ideally_ all iterators should have the potential to have row/column pagination, and this is the first example of it. @bkmartinjr mentions doing it for `tables()` is trivial now.
- I think having the `axis` argument of `read().scipy()` to be some times incompatible on the `result_order` of `.read()` is not ideal. Imo there should be only one place that defines the functionality.

## Suggestion

I have a suggestion that I think address the two comments above and which I thin is future proof:

**1. Adding `step` argument to `read()` and control stride using `result_order`**

```python
    def read(
        self,
        coords: options.SparseNDCoords = (),
        step: int = None,
        *,
        result_order: options.ResultOrderStr = options.ResultOrder.AUTO,
        batch_size: options.BatchSize = _UNBATCHED,
        partitions: Optional[options.ReadPartitions] = None,
    ) -> "SparseNDArrayRead":
        """
        Reads a user-defined slice of the :class:`SparseNDArray`.
        Args:
           [...]
           step: 
              When specified and `result_order` is "ROW_MAJOR" or "COLUMN_MAJOR",
              the resulting read iterators will guarantee full row or column pagination, respectively.
              When `None`, the resulting iterators produce ragged results per chunk.  
        [...]
        """
```  

**2. Removing `step` and `axis` arguments from `.read().scipy()`**

- Always returns coo if `compress = False`
- Returns csr or csc if `compress = True` and depending in `result_order` of `.read()`

```
    def scipy(
        self,
        compress: bool = True,
        reindex_sparse_axis: bool = True,
    ) -> Iterator[
        Tuple[
            Tuple[npt.NDArray[np.int64], npt.NDArray[np.int64]],
            Union[
                scipy.sparse.csr_matrix,
                scipy.sparse.csc_matrix,
                scipy.sparse.coo_matrix,
            ],
        ]
    ]
```

**3. Implement row/column pagination for other iterators: `coos()`,  `dense_tensors()`, and `tables()`**

Implement now for `tables()` since it's an easily lift. Implement later for others if not straightforward, and raise error if `step` is not `None`  in `read()`

atolopko-czi:
I prefer the `step` arg in read() as well, since `step` is more about the mechanics of the read operations, rather than the data type that is being requested.  However, we already have the `batch_size` arg. Can we just use that now (it's been unimplemented)?

While the `compress` and `axis` args lead to a "cleaner", minimal set of format-related read methods (i.e., just coos, tables, scipy), it's common in other libraries to have explicit methods for each of csr, csc, and coo. As an API reader, I would be looking such method names, and it would take slightly longer to understand how `compress` and `axis` map to these types. So I would prefer just having explicit `scipy_{csr,csc,coo}` methods. While I don't feel too strongly about this, thought it worth mentioning.

bkmartinjr:
There is a lot of confusion here that is (likely) the result of thinking about 2D matrics. Sort order and a step axis are entirely different concepts:
* Result order is the sort order of points (for those result formats that are otherwise unsorted).  Row-order means sort by [dim0, dim1, ..., dimN], and column-order means sort by [dimN, dimN-1, ..., dim0].
* Axis, when specified, is the single-dimension slice (step) that the iterator will produce.  It does NOT imply any sort order (other than this step partition), _unless_ the particular iterator has other sorting semantics (as CSC/CSR do on their compressed dimension).

For 2D scipy, these concepts are quite close.  For nD > 2D, and for formats that do not imply any particular point order (e.g., Table), result_order and step axis control entirely different things.

bkmartinjr:
> it's common in other libraries to have explicit methods for each of csr, csc, and coo.

this is not true in numpy, which is arguably the default (standard) for nd stuff in Python.

atarashansky:
GIven that result order and iteration axis are different concepts, it seems appropriate to leave `axis` as a parameter in `scipy()`.

Is there a way to separate the functionalities of compress and axis? For example, can we allow users to request their column chunks be returned in CSR format? I envision scenarios where users might want to carry out row-wise operations on matrix column chunks. Currently, setting compress=True with axis=1 results in csc format. While users can convert this themselves using .tocsr(), providing a direct option might enhance usability. It would also make it easy to write sugary wrappers like `scipy_csr() scipy_coo() scipy_csc()`.

thetorpedodog:
To put it on the record, [Pablo’s proposal](https://github.com/single-cell-data/TileDB-SOMA/pull/1792#issuecomment-1767406305) also looks reasonable to me. I don’t have enough subject matter expertise to say whether one is better than the other so I defer to the judgement of the rest of the group.

bkmartinjr:
The concern I have with Pablo's method is that it doesn't obviously work with non-ragged Table iterators > 2D.

Here is another counter proposal, which uses the fluent style to segment the concepts a bit.  Basic idea:
* there are ragged iterators (our current sutff)
* there are _stepped_ iterators (our new stuff).
* 
Stepped iterators make one promise -- given a single dimension (0, 1, ....), they will ensure that each iterator step yields a complete read of the array on that _single_ dimension.

API sketch:
```python

# generate ragged reads (no changes here)
A.read().tables()   # COO Arrow Table
A.read().coos()  # COO SparseCOOTensor

# Generate stepped reads
A.read().stepped(axis=0, step=1000).tables()  # Tables with a complete dim0, no sort order for points
A.read(result_order="row").stepped(axis=0, step=1000).tables()  # Tables w/ full dim0, each step sorted [d0, d1, ..., dN]
A.read(result_order="col").stepped(axis=0, step=1000).tables()  # Tables w/ full dim0, each step sorted [dN, dN-1, ..., d0]

# ....coos() supports same result_order options as .tables()

A.read().stepped(axis=0, step=1000).scipy(compress=True)  # scipy.sparse.csr_matrix.
A.read().stepped(axis=0, step=1000).scipy(compress=False)  # scipy.sparse.coo_matrix

A.read().stepped(axis=1, step=1000).tables()  # Tables with complete dim1
A.read(result_order="row").stepped(axis=1, step=1000).tables()  # Tables w/ full dim1, each step sorted [d0, d1, ..., dN]
A.read(result_order="col").stepped(axis=1, step=1000).tables()  # Tables w/ full dim1, each step sorted [dN, dN-1, ..., d0]

# ... other result_order options for .tables() is as above

A.read().stepped(axis=1, step=1000).scipy(compress=True)  # scipy.sparse.csc_matrix
A.read().stepped(axis=1, step=1000).scipy(compress=False)  # scipy.sparse.coo_matrix
```

Notes:
* scipy iterators _ignore_ result_order and produce SciPy canonical matrix order

Options?
* could rename "axis" to "step_axis"
* could rename "step" to "step_size"

bkmartinjr:
> Is there a way to disentangle compress and axis? For example, can we allow users to request their column chunks be returned in CSR format? I envision scenarios where users might want to carry out row-wise operations on matrix column chunks.

Yes, but this would just recapitulate the SciPy API, to no benefit that I can see.  Why not just use the SciPy `.tocsr()` and `.tocsc()` methods?

aaronwolen:
@atolopko-czi raises a good point. Can we reuse the `batch_size` argument that's already defined in the spec (where the arg belongs is a separate question)?  I'm also wondering if we could modify this to honor the [`partitions` argument](https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md#somareadpartitions).

bkmartinjr:
@atolopko-czi @aaronwolen - I was under the impression that `batch_size` and `partitions` were reserved for future use in partitioning (for concurrent/distributed use cases).  Is your proposal to co-opt them for this use instead, or did I misunderstand their original intent?

bkmartinjr:
FYI, to address various comments about wanting the same non-ragged iteration over Arrow Tables, I just added a first cut at a stepped iterator for tables.  *CAVEAT:* I'm sure everyone will hate the API, in different ways, so consider that open for debate :-)   

Demo:
```python
In [1]: for i in exp.ms['RNA'].X['raw'].read(coords=(slice(0,1999),)).stepped_tables(axis=0, step=1000):
   ...:     print(i)
   ...: 
pyarrow.Table
soma_dim_0: int64
soma_dim_1: int64
soma_data: float
----
soma_dim_0: [[0,0,0,0,0,...,0,254,494,571,623]]
soma_dim_1: [[21,37,50,63,76,...,33071,33184,33168,33170,32975]]
soma_data: [[1,1,4,2,3,...,1,1,1,1,1]]
pyarrow.Table
soma_dim_0: int64
soma_dim_1: int64
soma_data: float
----
soma_dim_0: [[1000,1000,1000,1000,1000,...,1097,1452,1759,1923,1952]]
soma_dim_1: [[37,41,50,51,54,...,33233,33344,33492,33232,33248]]
soma_data: [[1,1,2,1,1,...,1,1,1,1,1]]

In [2]: for i in exp.ms['RNA'].X['raw'].read(coords=(slice(0,1999),)).stepped_tables(axis=1, step=20000):
    ...:     print(i)
    ...: 
pyarrow.Table
soma_dim_0: int64
soma_dim_1: int64
soma_data: float
----
soma_dim_0: [[0,0,0,0,0,...,1999,1999,1999,1999,1999]]
soma_dim_1: [[21,37,50,63,76,...,19956,19964,19966,19981,19986]]
soma_data: [[1,1,4,2,3,...,1,1,1,2,1]]
pyarrow.Table
soma_dim_0: int64
soma_dim_1: int64
soma_data: float
----
soma_dim_0: [[0,0,0,0,0,...,1097,1452,1759,1923,1952]]
soma_dim_1: [[20002,20006,20008,20011,20014,...,33233,33344,33492,33232,33248]]
soma_data: [[1,1,1,2,1,...,1,1,1,1,1]]
```

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -79,3 +217,292 @@ def _from_table(self, arrow_table: pa.Table) -> pa.SparseCOOTensor:
...
+            step_coords = [*coords]
+            step_coords[axis] = coord_chunk
+            array._set_reader_coords(sr, step_coords)
+            yield pa.concat_tables(_arrow_table_reader(sr)), tuple(
+                joinids[0:axis] + [pa.array(coord_chunk)] + joinids[axis + 1 : ndim]
+            )
+
+    # with reindexing
+    def _reindexed_table_reader() -> SingleAxisTableIterReturn:
+        _tbl_iter = _table_reader() if eager else _if_eager(_table_reader())
```
thetorpedodog, line 277: duplicate `if eager` checks here?

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -79,3 +217,292 @@ def _from_table(self, arrow_table: pa.Table) -> pa.SparseCOOTensor:
...
+                        col = pd.Index(coords[axis]).get_indexer(col.to_numpy())  # type: ignore[no-untyped-call]
+                    else:
+                        col = minor_axes_indexer[d].get_indexer(col.to_numpy())  # type: ignore[no-untyped-call]
+                pytbl[f"soma_dim_{d}"] = col
+            pytbl["soma_data"] = tbl.column("soma_data")
+            yield pa.Table.from_pydict(pytbl), coords
+
+    if eager and not pool:
+        with concurrent.futures.ThreadPoolExecutor() as _pool:
+            pool = _pool
```
thetorpedodog, line 293: Changing out name bindings to backdoor pass new values into the function(s) below is a little sketchy to me—I would find it clearer to explicitly pass this in as a parameter.

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -79,3 +217,292 @@ def _from_table(self, arrow_table: pa.Table) -> pa.SparseCOOTensor:
...
+
+    if axis in reindex_disable:
+        raise SOMAError(
+            "Unable to disable reindexing of coordinates on CSC/CSR major axis"
+        )
+
+    shape = cast(Tuple[int, int], tuple(sr.shape))
+
+    # normalize params tuple
+    coords = tuple(coords[i] if i < len(coords) else None for i in range(2))
```
thetorpedodog, line 334: Since this appears in a couple places, it might be worth pulling out into a little function, largely so it’s easier to understand at a glance (maybe `_pad_with_none(coords, 2)`)

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -79,3 +217,292 @@ def _from_table(self, arrow_table: pa.Table) -> pa.SparseCOOTensor:
...
+def _single_axis_blockwise_table_iter(
+    array: "SparseNDArray",
+    sr: clib.SOMAArray,
+    coords: options.SparseNDCoords,
+    axis: int,
+    size: int,
+    reindex_disable: List[int],
+    eager: bool = True,
+    pool: Optional[concurrent.futures.Executor] = None,
+) -> SingleAxisTableIterReturn:
```
thetorpedodog, line 234: This and the below seem to have pretty similar structure. Would it make sense to put that common structure into the BlockwiseReadIter classes?

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -5,18 +5,46 @@
...
+from . import _util
+from ._exception import SOMAError
 from ._types import NTuple
 
+if TYPE_CHECKING:
+    from . import SparseNDArray
+
+
+# Convenience types
+RT = TypeVar("RT")
```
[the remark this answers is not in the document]
bkmartinjr: done

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -36,7 +64,117 @@ def concat(self) -> pa.Table:
         return pa.concat_tables(self)
 
 
-RT = TypeVar("RT")
+class BlockwiseReadIterBase(somacore.ReadIter[RT], metaclass=abc.ABCMeta):
+    """Private implementation class"""
+
+    _reader: Iterator[RT]
```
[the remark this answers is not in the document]
bkmartinjr: the reason I did not do that is the reader required the normalized arguments, as produced by _blockwise_config(). I like the idea though, so let me chew on how to make this a bit more robust

apis/python/src/tiledbsoma/_read_iters.py
bkmartinjr: I'm not entirely clear on the suggestion. Are you suggesting that the BlockwiseReadIter hold onto the common params, I the generators (such as this function) simply accept a BlockwiseReadIter as an argument?

(that does seem cleaner - I just wanted to confirm the suggest intent)

apis/python/src/tiledbsoma/_read_iters.py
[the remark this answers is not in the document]
bkmartinjr: i'm going to fix this by adding a subclass responsibility (abstract) method to create the reader.  Feels much cleaner.  TY for the suggestion!

apis/python/src/tiledbsoma/_read_iters.py
bkmartinjr: resolved by explicitly passing pool to generators

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -36,7 +64,117 @@ def concat(self) -> pa.Table:
         return pa.concat_tables(self)
 
 
-RT = TypeVar("RT")
+class BlockwiseReadIterBase(somacore.ReadIter[RT], metaclass=abc.ABCMeta):
```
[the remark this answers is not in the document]
bkmartinjr: left in place due to addition of the sub-class responsibility around reader creation

apis/python/src/tiledbsoma/_read_iters.py
thetorpedodog: It seems like these could be a method on the respective BlockwiseReadIter subclasses, and the common parts within them (it looks like some of the internal functions are substantially similar?) could be pulled into the base class (to the extent possible).

apis/python/src/tiledbsoma/_read_iters.py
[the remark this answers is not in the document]
thetorpedodog: Haven't gone through a full read-through yet but that is also a good structure to use.

apis/python/src/tiledbsoma/_read_iters.py
bkmartinjr: Oh, I see - you want to move all this into the classes.  Let me give that some thought - it certainly seems possible, although I'm not sure it helps or hurts comprehensibility

bkmartinjr: OK, just did a major refactoring along these lines.  PTAL!

pablo-gar (approved):
LGTM from an API design standpoint, a technical review should come from someone else in the team.

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -5,59 +5,420 @@
...
+]
+
+IndicesType = Tuple[npt.NDArray[np.int64], npt.NDArray[np.int64]]
+IJDType = Tuple[
+    Tuple[npt.NDArray[np.int64], npt.NDArray[np.int64]], npt.NDArray[np.generic]
+]
+
 
 class TableReadIter(somacore.ReadIter[pa.Table]):
     """Iterator over `Arrow Table <https://arrow.apache.org/docs/python/generated/pyarrow.Table.html>`_ elements"""
```
atolopko-czi, line 62:
```suggestion
    """Iterator over `Arrow Table <https://arrow.apache.org/docs/python/generated/pyarrow.Table.html>` elements"""
```

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -5,59 +5,420 @@
...
+        """Sub-class responsibility"""
+        raise NotImplementedError()
+
+    def __next__(self) -> _RT:
+        return next(self._reader)
+
+    def concat(self) -> _RT:
+        """
+        Unimplemented as there is little utility beyond that offered by a ragged
+        read iterator concat, other than reindexing.
```
atolopko-czi, line 202: Why not implement anyway? Seems like it would be a less surprising outcome for the user. And isn't reindexing a valid reason to enable this?

apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -411,24 +431,173 @@ def coos(self, shape: Optional[NTuple] = None) -> SparseCOOTensorReadIter:
...
+          `[0, N)`, where `N` is the number of coordinates read for the block (controlled with the `size` argument).
+        * all other axes will be reindexed will be reindexed to `[0, N)`, where `N` is the number of points read
+          on that axis across all blocks.
+
+        Args:
+            axis:
+                Required. The axis across which to yield blocks, indicated as the dimension number, e.g.,
+                `axis=0` will step across `soma_dim_0` (the first dimension).
+            size:
+                Optional. Number of coordinates in each block yielded by the iterator.
```
atolopko-czi, line 479: I had to drill down a few levels to see what the default is. Worth documenting here?

apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -411,24 +431,173 @@ def coos(self, shape: Optional[NTuple] = None) -> SparseCOOTensorReadIter:
...
-    def tables(self) -> TableReadIter:
+        Blockwise iterators yield an array "block" in some user-specified format, as well as a
+        list of coordinates contained in the individual block.
+
+        All blockwise iterators will reindex coordinates (i.e., map them from soma_joinid to an integer
+        in the range [0, N)), unless reindexing is specifically disabled for that axis, using the
+        `reindex_disable` argument. When reindexing:
+        * the primary iterator axis coordinates, as indicated by the `axis` argument, will be reindexed into the range
+          `[0, N)`, where `N` is the number of coordinates read for the block (controlled with the `size` argument).
+        * all other axes will be reindexed will be reindexed to `[0, N)`, where `N` is the number of points read
```
atolopko-czi, line 471:
```suggestion
        * all other axes will be reindexed to `[0, N)`, where `N` is the number of points read
```

apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -411,24 +431,173 @@ def coos(self, shape: Optional[NTuple] = None) -> SparseCOOTensorReadIter:
...
+                Optional. Sequence of axis which will _not_ be reindexed. Defaults to None, indicating
+                all axes will be reindexed.
+            eager:
+                Optional. If `True`, the iteartor will read ahead (using multi-threading) to improve overall
+                performance when iterating over a large result. Setting this flag to `False` will reduce memory
+                consumption, at the cost of additional processing time.
+
+        Examples:
+
+            A simple example iterating over the first 10000 elements, into SciPy sparse matrices:
```
atolopko-czi, line 490:
```suggestion
            A simple example iterating over the first 10000 elements of the first dimension, into blocks of SciPy sparse matrices:
```

atolopko-czi (commented):
Initial pass of a review, "publishing" early comments for now, all minor.

apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -379,7 +388,26 @@ def _set_bounding_box_metadata(
...
+        Lifecycle:
+            Experimental.
+        """
+        self.sr = sr
+        self.shape = tuple(sr.shape)
+        self.array = array
+        self.coords = coords
+
+
+class SparseNDArrayRead(_SparseNDArrayReadBase):
```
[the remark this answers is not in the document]
bkmartinjr: I'll add a variation on your suggestion.

apis/python/src/tiledbsoma/_read_iters.py
bkmartinjr: This link syntax is  used pervasively, and has something to do with how they are published.  AFAIK, not a typo.

@johnkerl - can you confirm?

apis/python/src/tiledbsoma/_read_iters.py
bkmartinjr: My principle was to not add additional code/API that the user could replicate with a single line on their own (e.g., `pandas.Index(...).get_indexer(...)`.  So yeah, it is a valid case, but I don't see what value we are providing by making our own API contain more sugar....

Love to get other's thoughts on this

apis/python/src/tiledbsoma/_sparse_nd_array.py
bkmartinjr: as it is a heuristic, I was reticent to add a hard-coded default.  But I can certainly weasel-word that...

bkmartinjr: added, PTAL.

apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -411,24 +437,176 @@ def coos(self, shape: Optional[NTuple] = None) -> SparseCOOTensorReadIter:
...
-    def tables(self) -> TableReadIter:
+        Blockwise iterators yield an array "block" in some user-specified format, as well as a
+        list of coordinates contained in the individual block.
+
+        All blockwise iterators will reindex coordinates (i.e., map them from soma_joinid to an integer
+        in the range [0, N)), unless reindexing is specifically disabled for that axis, using the
+        `reindex_disable` argument. When reindexing:
+        * the primary iterator axis coordinates, as indicated by the `axis` argument, will be reindexed into the range
+          `[0, N)`, where `N` is the number of coordinates read for the block (controlled with the `size` argument).
+        * all other axes will be reindexed to `[0, N)`, where `N` is the number of points read
```
mlin, line 477:
```suggestion
        * all other axes will be reindexed to `[0, M)`, where `M` is the number of points read
```

mlin: Suggestion just to avoid confusion comparing with the prior bullet point

apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -411,24 +437,176 @@ def coos(self, shape: Optional[NTuple] = None) -> SparseCOOTensorReadIter:
...
+    def scipy(self, *, compress: bool = True) -> BlockwiseScipyReadIter:
+        """
+        Returns a blockwise iterator of
+        `SciPy sparse matrix` <https://docs.scipy.org/doc/scipy/reference/sparse.html>
+        over a 2D SparseNDArray.
+
+        Args:
+            compress:
+                If True, a CSC or CSR matrix is returned, dependent on the value of the
+                `axis` argument. If False, a COO matrix is returned.
```
mlin, line 593: Maybe add something like

> Note: implementation details of SciPy CSC and CSR compression effectively require reindexing of the major axis (columns and rows, respectively). Therefore, this method will throw an error if the major axis was included in the `reindex_disable` argument to `blockwise()`. Reindexing can be disabled for the minor axis.

mlin (approved):

apis/python/src/tiledbsoma/_read_iters.py
atolopko-czi: OK, maybe something to consider for a SOMA API V2, where "Full" result read methods could be exposed at the `SparseRead` level, rather than at `ReadIter`.

apis/python/src/tiledbsoma/_sparse_nd_array.py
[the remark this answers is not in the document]
atolopko-czi: The added docstring looks great!

apis/python/src/tiledbsoma/_sparse_nd_array.py
atolopko-czi: Looks good, ty!  

(It's too bad SOMA doesn't have a central list of config settings like [TileDB config settings](https://docs.tiledb.com/main/how-to/configuration#configuration-parameters), where default values can be published to docs from code-level constants.)

apis/python/src/tiledbsoma/_read_iters.py
johnkerl: @bkmartinjr 
https://tiledbsoma.readthedocs.io/en/latest/_autosummary/tiledbsoma.SparseNDArray.html#tiledbsoma.SparseNDArray

atolopko-czi:
While test driving the blockwise API, I came across a possible bug. Querying a slice along the primary axis (0) that does not start at first row, and then requesting a scipy COO (uncompressed) with axis 0 reindexing disabled, causes an exception to be thrown. Captured in the following unit test:

```
@pytest.mark.parametrize("density,shape", [(0.1, (100, 100))])
def test_blockwise_scipy_reindex_disable_coo(
    a_random_sparse_nd_array: str, shape: Tuple[int, ...]
) -> None:
    # Passes
    with soma.open(a_random_sparse_nd_array, mode="r") as A:
        next(A.read((slice(0, 10),)).blockwise(axis=0, reindex_disable=0).scipy(compress=False))
    # Fails!
    # ValueError: row index exceeds matrix dimensions
    # .../venvs/tiledbsoma/lib/python3.10/site-packages/scipy/sparse/_coo.py:291: ValueError
    with soma.open(a_random_sparse_nd_array, mode="r") as A:
        next(A.read((slice(1, 10),)).blockwise(axis=0, reindex_disable=0).scipy(compress=False))
```

apis/python/src/tiledbsoma/_read_iters.py
bkmartinjr: @johnkerl - not sure I understand the response.  I _believe_ you are saying the underscore is required (something to do with the rendering format is my guess). If you have a doc pointer for the rendering/formatting language, it would be appreciated.

apis/python/src/tiledbsoma/_read_iters.py
bkmartinjr: The other issue I forgot to mention -- `concat` is complicated when you consider reindexing on the primary (blockwise) axis.  Naively implemented, it would just create a bunch of duplicate coordinates (zero to N), so we probably would need to either disallow reindexing on the primary dimension when concat is called, _or_ redefine it as reindexing the full concatenated result.

For SciPy blockwise reads, there is already a trivial implementation that does the same, without the sharp edges, which is the `vstack()` method in the SciPy sparse package.  This allows nice patterns like:

```
with soma.open(a_random_sparse_nd_array, mode="r") as A:
    full_csr_matrix = sparse.vstack(sp for sp, _ in A.read(coords).blockwise(axis=0).scipy())
```

bkmartinjr:
> While test driving the blockwise API, I came across a possible bug

@atolopko-czi - definitely a bug.  A corner case I had missed.  Fix and unit tests will be pushed shortly.

johnkerl (approved):

atolopko-czi (changes requested):
LGTM! Code structure is solid and code is well tested. API is straight forward to use. A few final nits for your consideration. Approving optimistically, ahead of fix for https://github.com/single-cell-data/TileDB-SOMA/pull/1792#issuecomment-1790846386. Looks forward to using this!

atolopko-czi (approved):

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -5,59 +5,424 @@
...
+        """
+        raise NotImplementedError("Blockwise iterators do not support concat operation")
+
+    _EagerRT = TypeVar("_EagerRT")
+
+    def _if_eager(
```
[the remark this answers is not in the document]
bkmartinjr: lolz, that was what I originally called it.  Switched for clarity, I thought. Will switch back.

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -5,59 +5,424 @@
...
+        self,
+        array: "SparseNDArray",
+        sr: clib.SOMAArray,
+        coords: options.SparseNDCoords,
+        axis: Union[int, Sequence[int]],
+        *,
+        size: Optional[Union[int, Sequence[int]]] = None,
+        reindex_disable: Optional[Union[int, Sequence[int]]] = None,
+        eager: bool = True,
+        pool: Optional[ThreadPoolExecutor] = None,
```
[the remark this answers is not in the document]
bkmartinjr: Will rename.  However, please note that `pool` does *not* override `eager`.

.flake8
thetorpedodog, line 1: Does your editor have Ruff integration? It should obviate the need for flake8 specifically.

apis/python/src/tiledbsoma/_read_iters.py
thetorpedodog: The underscore should be there, yes.

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -5,59 +5,424 @@
...
+        """
+        Unimplemented as there is little utility beyond that offered by a ragged
+        read iterator concat, other than reindexing.
+        """
+        raise NotImplementedError("Blockwise iterators do not support concat operation")
+
+    _EagerRT = TypeVar("_EagerRT")
+
+    def _if_eager(
+        self, x: Iterator[_EagerRT], _pool: Optional[ThreadPoolExecutor] = None
```
thetorpedodog, line 209: I think the `pool` here is always going to be `self.pool`, right?

bkmartinjr:
@atolopko-czi - fix for the bug you found is in the branch.  See 93d3233

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -5,59 +5,424 @@
...
+        eager: bool = True,
+        pool: Optional[ThreadPoolExecutor] = None,
+    ):
+        super().__init__()
+
+        self.ndim = len(sr.shape)
+        self.array = array
+        self.sr = sr
+        self.eager = eager
+        self.pool = pool
```
thetorpedodog: Would you want to make these `_`-prefixed to make it clear that they should not be modified by outsiders? Up to you; the fact that they are in an internal class that a user will never have cause to instantiate might be enough.

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -5,59 +5,424 @@
...
+    def _table_reader(self) -> BlockwiseSingleAxisTableIter:
+        """Private. Blockwise table reader. Helper function for sub-class use"""
+        kwargs: Dict[str, object] = {"result_order": self.sr.result_order}
+        for coord_chunk in _coords_strider(
+            self.coords[self.major_axis],
+            self.sr.shape[self.major_axis],
+            self.size[0],
+        ):
+            self.sr.reset(**kwargs)
+            step_coords = [*self.coords]
```
thetorpedodog, line 223: `list(self.coords)`?

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -5,59 +5,424 @@
...
+            step_coords[self.major_axis] = coord_chunk
+            self.array._set_reader_coords(self.sr, step_coords)
+            yield pa.concat_tables(_arrow_table_reader(self.sr)), tuple(
+                self.joinids[0 : self.major_axis]
+                + [pa.array(coord_chunk)]
+                + self.joinids[self.major_axis + 1 : self.ndim]
+            )
```
thetorpedodog, line 230: I think just switching out one element of a list would make this easier to understand:

```python
joinids = list(self.joinids)
joinids[self.major_axis] = pa.array(coord_chunk)
yield ..., tuple(joinids)
```

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -5,59 +5,420 @@
...
+
+    def __init__(
+        self,
+        array: "SparseNDArray",
+        sr: clib.SOMAArray,
+        coords: options.SparseNDCoords,
+        axis: Union[int, Sequence[int]],
+        *,
+        size: Optional[Union[int, Sequence[int]]] = None,
+        reindex_disable: Optional[Union[int, Sequence[int]]] = None,
```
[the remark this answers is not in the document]
thetorpedodog: I'm not a *hard* no, but I worry that doing so would overload the meaning of this parameter a bit.

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -5,59 +5,424 @@
...
+        """Private. Compressed sparse variants"""
+        assert self.compress
+        assert self.major_axis not in self.reindex_disable
+        for ((i, j), d), indices in self._if_eager(
+            self._sorted_tbl_reader(_pool), _pool
+        ):
+            major_coords, minor_coords = (
+                indices[self.major_axis],
+                indices[self.minor_axis],
+            )
```
thetorpedodog, line 397: maybe just make these two assignments?

apis/python/src/tiledbsoma/_read_iters.py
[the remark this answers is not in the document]
bkmartinjr: nice idea, will add.  more sugar, more unit tests :-)

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -79,3 +444,62 @@ def _from_table(self, arrow_table: pa.Table) -> pa.SparseCOOTensor:
...
         ).T
         return pa.SparseCOOTensor.from_numpy(coo_data, coo_coords, shape=self.shape)
+
+
+def _arrow_table_reader(sr: clib.SOMAArray) -> Iterator[pa.Table]:
+    """Private. Simple Table iterator on any Array"""
+    tbl = sr.read_next()
+    while tbl is not None:
+        yield tbl
+        tbl = sr.read_next()
```
thetorpedodog, line 454: Idle thought: This seems like it could somehow be `iter(???, None)` but the wrapper code to turn it into that ends up being more than this anyway.

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -79,3 +444,62 @@ def _from_table(self, arrow_table: pa.Table) -> pa.SparseCOOTensor:
...
+    """
+    Private.
+
+    Iterate over major coordinates, in stride sized steps, materializing each step as an
+    ndarray of coordinate values. Will be sorted in ascending order.
+
+    NB: SOMA slices are _closed_ (i.e., inclusive of both range start and stop)
+    """
+    # normalize coord to either a slice or ndarray
+    if coords is None or (isinstance(coords, slice) and coords == slice(None)):
```
thetorpedodog, line 469: probably no need to `isinstance` here; `coords == slice(None)` is going to be False for non-slices (I would expect)

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -79,3 +444,62 @@ def _from_table(self, arrow_table: pa.Table) -> pa.SparseCOOTensor:
...
+        coords = slice(0, length - 1)
+    elif isinstance(coords, slice):
+        if coords == slice(None):
+            coords = slice(0, length - 1)
```
thetorpedodog, line 473: you already covered this

thetorpedodog (commented):
Overall I like this a lot! A few things which are mostly smaller concerns but it’s looking good.

.flake8
bkmartinjr: it does, but it doesn't seem to work well.  I'll remove the .flake8

apis/python/src/tiledbsoma/_read_iters.py
bkmartinjr: No.  I want the ability to override the self.pool with the context-managed pool, so it is definitely not the same.  self.pool is whatever the user provided.

apis/python/src/tiledbsoma/_read_iters.py
bkmartinjr: I had assumed that it was OK as it was an implementation class.  But I can do it if it help

Are you suggest it for the instance fields, or the arguments to \_\_init__?

apis/python/src/tiledbsoma/_read_iters.py
bkmartinjr: yeah, I have played with this a bit, and haven't found anything cleaner/terser.

apis/python/src/tiledbsoma/_read_iters.py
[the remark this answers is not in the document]
bkmartinjr: I'm ambivalent - but have it implemented.  Who wants to decide?

apis/python/src/tiledbsoma/_read_iters.py
bkmartinjr: Acutally, it is needed to handle the case where coords is an NDArray and will broadcast the equality test along the entire array.

apis/python/src/tiledbsoma/_read_iters.py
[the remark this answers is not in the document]
atolopko-czi: I suggested it because while test driving the API, I mistakenly assumed it was a bool flag. :) After RTFM, it made sense, but it seemed like a minor trap. If we want to keep typing as is, maybe consider renaming the param to something, e.g. `reindex_disable_on_axis`

apis/python/src/tiledbsoma/_read_iters.py
thetorpedodog: I see. Could you add a comment to this effect? (not back in the full review mode right now; if you already have that's good)

bkmartinjr: will do

aaronwolen (approved):
This is fantastic. Docs are comprehensive and super clear. Thanks!

apis/python/src/tiledbsoma/_read_iters.py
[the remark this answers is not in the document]
aaronwolen: @atolopko-czi's `reindex_disable_on_axis` suggestion seems like good middle ground. The arg name is a bit verbose but its intent is clear.

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -168,6 +170,7 @@ def _validate_args(
         if not all(0 <= d < ndim for d in axis):
             raise ValueError("blockwise `axis` value must be in range [0, ndim)")
         if not all(0 <= d < ndim for d in reindex_disable):
+            print(reindex_disable)
```
thetorpedodog, line 173: stray debug print

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -5,59 +5,424 @@
...
+        )
+
+        self.major_axis = self.axis[0]
+        self.coords = _pad_with_none(coords, self.ndim)
+
+        # materialize all indexing info.
+        self.joinids: List[pa.Array] = [
+            pa.array(
+                np.concatenate(
+                    list(
```
thetorpedodog, line 112: is this `list(` needed? I would expect numpy to take any iterable you give it (though could be wrong)

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -5,59 +5,424 @@
...
+class BlockwiseTableReadIter(BlockwiseReadIterBase[BlockwiseTableReadIterResult]):
+    """Blockwise iterator over `Arrow Table <https://arrow.apache.org/docs/python/generated/pyarrow.Table.html>`_ elements"""
+
+    def _create_reader(self) -> Iterator[BlockwiseTableReadIterResult]:
+        """Private. Blockwise Arrow Table iterator, restricted to a single axis"""
+        if self.eager and not self.eager_iterator_pool:
+            with ThreadPoolExecutor() as _pool:
+                yield from self._reindexed_table_reader(
+                    _pool
+                ) if self.axes_to_reindex else self._table_reader()
```
thetorpedodog, line 261: These indentations are extremely weird. Maybe

```python
yield from (
    self._reindexed_whatever(...)
    if ... else ...
)
```
if it will let you?

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -79,3 +444,63 @@ def _from_table(self, arrow_table: pa.Table) -> pa.SparseCOOTensor:
...
+        for i in range(0, len(coords), stride):
+            yield cast(npt.NDArray[np.int64], coords[i : i + stride])
+
+
+_ElemT = TypeVar("_ElemT")
+
+
+def _pad_with_none(
+    s: Sequence[_ElemT], to_length: int
+) -> Tuple[Union[_ElemT, None], ...]:
```
thetorpedodog, line 504: `Optional[_ElemT]`, maybe?

apis/python/src/tiledbsoma/_read_iters.py
thetorpedodog: Just the instance fields, but as I said, this isn’t critical.

thetorpedodog (approved):
a few minor things but overall very good

apis/python/src/tiledbsoma/_read_iters.py
bkmartinjr: your expectations is wrong - it takes a Sequence, not an interable.

demo:
```
In [4]: np.concatenate( (np.zeros((i,)) for i in range(4)) )
TypeError                                 Traceback (most recent call last)
Cell In [4], line 1
----> 1 np.concatenate( (np.zeros((i,)) for i in range(4)) )

File <__array_function__ internals>:200, in concatenate(*args, **kwargs)

TypeError: The first input argument needs to be a sequence

In [5]: np.concatenate( tuple(np.zeros((i,)) for i in range(4)) )
Out[5]: array([0., 0., 0., 0., 0., 0.])
```

apis/python/src/tiledbsoma/_read_iters.py
bkmartinjr: it was whatever ruff/black put in place.  I will try your proposal - it should work fine.

apis/python/src/tiledbsoma/_read_iters.py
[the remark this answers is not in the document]
pablo-gar: I see disabling re-indexing a bit more of an advanced functionality, and will also likely happen on one axis frequently. Let's not add more cognitive dissonance with `bool` option. Re-naming the parameter is a +1. 

Thus

- `reindex_disable_on_axis` should be the name
- `reindex_disable_on_axis` should not take bool.

This is one of these cases where we can easily add the sugar in the future if needed, whereas removing it in the future can get complicated.

johnkerl:
🚀

Issue #783, opened 2023-01-21
[python] Errors raised as `RuntimeError` should raise `SOMAError`


bkmartinjr:
Errors raised from inside SOMAReader (see example below) can't be caught as a `TileDBError` or `SOMAError`, as they are caught and rethrown as a `RuntimeError` inside the C++ bindings.

```
In [28]: sdf.schema
Out[28]: 
soma_joinid: int64
obs_id: large_string
n_genes: int64
percent_mito: float
n_counts: float
louvain: large_string
is_b_cell: bool

In [29]: sdf.read(value_filter="n_genes == b'abc'")
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In [29], line 1
----> 1 sdf.read(value_filter="n_genes == b'abc'")

File ~/TileDB-SOMA/apis/python/src/tiledbsoma/dataframe.py:250, in DataFrame.read(***failed resolving arguments***)
    247 if value_filter is not None:
    248     query_condition = QueryCondition(value_filter)
--> 250 sr = clib.SOMAReader(
    251     self._uri,
    252     name=self.__class__.__name__,
    253     schema=A.schema,  # query_condition needs this
    254     column_names=column_names,
    255     query_condition=query_condition,
    256     platform_config={} if self._ctx is None else self._ctx.config().dict(),
    257     result_order=result_order.value,
    258 )
    260 if coords is not None:
    261     if not isinstance(coords, (list, tuple)):

RuntimeError: TileDBError: Cannot cast `b'abc'` to int64.```


Ideally, catching these as `SOMAError` would be possible, or worst case, as `TileDBError`

**Comment by bkmartinjr** [comment]
Here is a second evaluation error which returns a different (still wrong IMHO) error.  Ideally it would raise a SOMAError, clearly indicating that the operation is unsupported.

There is code in query_condition.py that is trying to test for this (use of unsupported operator), but for some reason it is not catching the error at parse time.

```
In [37]: sdf.read(value_filter="n_genes > 1+1")
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In [37], line 1
----> 1 sdf.read(value_filter="n_genes > 1+1")

File ~/TileDB-SOMA/apis/python/src/tiledbsoma/dataframe.py:250, in DataFrame.read(***failed resolving arguments***)
    247 if value_filter is not None:
    248     query_condition = QueryCondition(value_filter)
[...]

PR #1793, opened 2023-10-16
[python] Dataframe read path

nguyenv:
**Issue and/or context:**

**Running Changes:**

* When opening a `DataFrame` in read-mode, use `DataFrameWrapper` which wraps around `clib.SOMADataFrame`. Otherwise, `DataFrame` should use the already existing write-path with `ArrayWrapper` which wraps around a TileDB-Py Array
* Corrections to C++ unit tests where `SOMADataFrame` needs to use sparse array
* `SOMADataFrame::count` needs to use `nnz`
* C++ API `create` needs to set `soma_object_type` and `exists` needs to check `soma_object_type`
* Get full nonempty domains for `SOMADataFrame`
* Add getters and methods to `DataFrameWrapper` and `ArrayWrapper` (i.e nonempty domain, attr names, dim names, etc)
* Support query conditions for `SOMADataFrame`
* `_query_condition.py` no longer has dependency on TileDB-Py and uses Pyarrow schema instead of TileDB ArraySchema and `SOMAError` instead of `TileDBError`
* Reorganize Pybind11 code to place util functions and `PyQueryCondition` (note: should refactor `QueryCondition` to completely remove `PyQueryCondition` usage) into `common.h`
* `SOMAArray` now resides in its own `soma_array.cc` file
* `pytiledbsoma.cc` is now the top-level file that contains stats bindings and loads modules from `query_condition.cc`, `soma_array.cc`, `soma_dataframe.cc`
* Support enumerations / dictionaries for `SOMADataFrame`
* Support all enumeration values in `ArrowAdapter::to_arrow`
* `read` Pybind11 function now uses correct Arrow schema to convert dictionary arrays when importing from C to Python
* `SOMAObject::schema` returns `ArrowSchema` which calls from `SOMAArray::arrow_schema`
* Create C++ factory `unique_ptr<SOMAObject> SOMAObject::open` and bind in Pybind11 as `clib.SOMAObject.open` which returns the correct Python SOMA class
* Fix for #783


apis/python/src/tiledbsoma/_collection.py
```diff
@@ -425,9 +427,16 @@ def __getitem__(self, key: str) -> CollectionElementType:
             raise KeyError(err_str) from None
         if entry.soma is None:
             from . import _factory  # Delayed binding to resolve circular import.
+            from ._tdb_handles import Wrapper
+
+            wrapper: type[Wrapper[Any | Any | Any]]
```
thetorpedodog, line 432: Looks like this `Any` is left over from type troubleshooting?

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -354,17 +354,22 @@ def read(
...
+        to_clib_result_order = {
+            options.ResultOrder.AUTO: clib.ResultOrder.automatic,
+            options.ResultOrder.ROW_MAJOR: clib.ResultOrder.rowmajor,
+            options.ResultOrder.COLUMN_MAJOR: clib.ResultOrder.colmajor,
+            "auto": clib.ResultOrder.automatic,
+            "row-major": clib.ResultOrder.rowmajor,
+            "column-major": clib.ResultOrder.colmajor,
+        }
+        if result_order not in to_clib_result_order:
+            raise ValueError(f"Invalid result_order: {result_order}")
```
thetorpedodog, line 366: see comment on other change

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -500,34 +505,41 @@ def _set_reader_coord(
...
-            dim.dtype == "str" or dim.dtype == "bytes"
+            pa.types.is_large_string(dim.type)
+            or pa.types.is_large_binary(dim.type)
+            or pa.types.is_string(dim.type)
+            or pa.types.is_binary(dim.type)
         ):
             _util.validate_slice(coord)
             # Figure out which one.
-            dim_type: Union[Type[str], Type[bytes]] = type(dim.domain[0])
+            dim_type: Union[Type[str], Type[bytes]] = type(self.domain[dim_idx][0])
```
thetorpedodog, line 515: Since we look at `self.domain[dim_idx]` a bunch in this code, would it make sense to pull it out near the top of the function and store it in a variable?

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -553,38 +565,34 @@ def _set_reader_coord_by_py_seq_or_np_array(
...
-
         elif (
-            dim.dtype == "datetime64[s]"
-            or dim.dtype == "datetime64[ms]"
-            or dim.dtype == "datetime64[us]"
-            or dim.dtype == "datetime64[ns]"
+            pa.types.is_large_string(dim.type)
+            or pa.types.is_large_binary(dim.type)
+            or pa.types.is_string(dim.type)
+            or pa.types.is_binary(dim.type)
```
thetorpedodog, line 592: maybe define an `is_pyarrow_anystr` (or a better name???) function for this? It appears a lot.

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -49,7 +53,10 @@ def open(
     if not obj_type:
         raise DoesNotExistError(f"{uri!r} does not exist")
     if obj_type == "array":
-        return ArrayWrapper.open(uri, mode, context, timestamp)
+        if mode == "r" and clib.SOMADataFrame.exists(uri):
+            return DataFrameWrapper.open(uri, mode, context, timestamp)
+        else:
+            return ArrayWrapper.open(uri, mode, context, timestamp)
```
thetorpedodog, line 59: As I’ve been reading this, I wonder if it would be possible for these C++-based wrappers to take in the TileDB Array object. This would mean that in this code we don’t have to repeatedly open and close these objects for the `.exists` calls—as it stands, every time we do an open, it will end up falling through every case, so it will look like:

 1. Figure out what type to open:

     1. `SOMADataFrame.exists` check
         1. Open array.
         3. Read metadata, compare with `SOMADataFrame`.
         4. Close array. (Does this one actually involve an RPC though?)
     1. `SOMASomeOtherType.exists` check
         1. Open array.
         2. Read metadata, compare with `SOMASomeOtherType`.
         3. Close array.
    
    and only after some number of those are done, do we actually open the array for reals:

 2. Open array *again*, and use that handle this time.

Even if this wasn’t improved to the "just one open" case, it would probably be best to make it so that we only open twice at most: once to figure out its type, then again to use it.

I would say **this** is the largest concern I have about anything in this change.

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -244,6 +268,110 @@ def _do_initial_reads(self, reader: tiledb.Group) -> None:
...
+                )
+            else:
+                if pa.types.is_large_string(dtype) or pa.types.is_string(dtype):
+                    dtype = np.dtype("U")
+                elif pa.types.is_large_binary(dtype) or pa.types.is_binary(dtype):
+                    dtype = np.dtype("S")
+                else:
+                    dtype = np.dtype(dtype.to_pandas_dtype())
+                result.append(self._handle.non_empty_domain(name, dtype))
+        return None if len(result) == 0 else tuple(result)
```
thetorpedodog, line 360: Is the only difference between `domain` and `non_empty_domain` whether it calls `self._handle.domain` vs. `self._handle.non_empty_domain`?

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -244,6 +268,110 @@ def _do_initial_reads(self, reader: tiledb.Group) -> None:
...
+
+    @property
+    def attr_names(self) -> Tuple[str, ...]:
+        result = []
+        for field in self.schema:
+            if field.name not in self._handle.index_column_names:
+                result.append(field.name)
+        return tuple(result)
```
thetorpedodog, line 368: `tuple(f.name for f in self.schema if f.name not in self._handle.index_column_names)`

apis/python/src/tiledbsoma/_collection.py
nguyenv: No if I don't put this here then I get an error with the type checker :/

apis/python/src/tiledbsoma/_tdb_handles.py
nguyenv: Thank you for your feedback on this. You're right that calling a series of `SOMAObject.exists` is going to unnecessarily open and close multiple arrays and groups when it could probably be avoided with another design. I'll come up with something else.

nguyenv: I've made changes to this design so that this should (eventually) only open the TileDB object once instead of using `exists`. 

In C++, I've added a static method `unique_ptr<SOMAObject> open(std::string)` which is a factory that returns a `unique_ptr` to a `SOMAObject` class (ie. `unique_ptr<SOMADataFrame>`). It calls `tiledb::Object` to check whether the URI is a TileDB array or group. Then opens either a `SOMAArray` or `SOMAGroup`. It reads the metadata to see the actual SOMA type. We then pass the already opened `SOMAArray` or `SOMAGroup` to the concrete SOMA class (eg. `make_unique<SOMADataFrame>(open_soma_array)`.

In Pybind11, we call `SOMAObject::open` and can use `soma_object.type()` to get the type of the SOMA object. We can then cast that to the correct Python binding (eg. `py::cast(dynamic_cast<SOMADataFrame&>(*soma_object)))`).

apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -344,7 +344,7 @@ def used_shape(self) -> Tuple[Tuple[int, int], ...]:
         # In the unlikely event that a previous data update succeeded but the
         # subsequent metadata update did not, take the union of the core non-empty domain
         # (which is done as part of the data update) and the metadata bounding box.
-        ned = self.non_empty_domain()
+        ned = self.non_empty_domain() or tuple()
```
thetorpedodog, line 347: just `or ()`

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -48,11 +55,33 @@ def open(
...
+    open_mode = clib.OpenMode.read if mode == "r" else clib.OpenMode.write
+    config = {k: str(v) for k, v in context.tiledb_config.items()}
+    timestamp_ms = context._open_timestamp_ms(timestamp)
+    try:
+        obj = clib.SOMAObject.open(uri, open_mode, config, (0, timestamp_ms))
+        if obj.type == "SOMADataFrame":
+            return _tdb_handles.DataFrameWrapper._from_soma_object(obj, context)
+        raise SOMAError(f"clib.SOMAObject {obj.type!r} not yet supported")
+    except SOMAError as err:
+        raise err
```
thetorpedodog, line 84: this outer try-except looks like it was leftover from something earlier.

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -48,11 +55,33 @@ def open(
...
+        return _get_wrapper(uri, mode, context, timestamp)
+    except SOMAError:
+        if obj_type == "array":
+            return ArrayWrapper.open(uri, mode, context, timestamp)
+        if obj_type == "group":
+            return GroupWrapper.open(uri, mode, context, timestamp)
+        raise SOMAError(f"{uri!r} has unknown storage type {obj_type!r}")
+
+
+def _get_wrapper(
```
thetorpedodog, line 69: `_open_with_cpp_wrapper`? `_open_cpp_wrapped`? some other name?

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -48,11 +55,33 @@ def open(
...
+    context: SOMATileDBContext,
+    timestamp: Optional[OpenTimestamp] = None,
+) -> "_tdb_handles.DataFrameWrapper":
+    open_mode = clib.OpenMode.read if mode == "r" else clib.OpenMode.write
+    config = {k: str(v) for k, v in context.tiledb_config.items()}
+    timestamp_ms = context._open_timestamp_ms(timestamp)
+    try:
+        obj = clib.SOMAObject.open(uri, open_mode, config, (0, timestamp_ms))
+        if obj.type == "SOMADataFrame":
+            return _tdb_handles.DataFrameWrapper._from_soma_object(obj, context)
```
thetorpedodog, line 81: this is in the `_tdb_handles` module so should this just be `DataFrameWrapper`?

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -194,20 +243,37 @@ def _opener(
...
-        """
-        Retrieves the non-empty domain for each dimension, namely the smallest
-        and largest indices in each dimension for which the array/dataframe has
-        data occupied.  This is nominally the same as the domain used at
-        creation time, but if for example only a portion of the available domain
-        has actually had data written, this function will return a tighter
-        range.
-        """
-        return self._handle.nonempty_domain()  # type: ignore
+    def non_empty_domain(self) -> Optional[Tuple[Tuple[Any, Any], ...]]:
```
thetorpedodog, line 246: I think you may have already responded to this and I missed it, but I had some previous discussion on the merits of `object` vs. `Any` either in this change or one of the similar changes which I cannot find at the moment. Just wanted to bring it up again in case not; if you already have that’s fine.

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -194,20 +243,37 @@ def _opener(
...
+        except tiledb.TileDBError as e:
+            raise SOMAError(e)
 
     def enum(self, label: str) -> tiledb.Enumeration:
         return self._handle.enum(label)
 
+    @property
+    def domain(self) -> Tuple[Tuple[Any, Any], ...]:
+        dom = self._handle.schema.domain
+        return tuple(dom.dim(i).domain for i in range(dom.ndim))
```
thetorpedodog, line 261: 🎵 dom dim i dom dom ndim

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -247,6 +313,104 @@ def _do_initial_reads(self, reader: tiledb.Group) -> None:
...
+        timestamp: int,
+    ) -> clib.SOMADataFrame:
+        open_mode = clib.OpenMode.read if mode == "r" else clib.OpenMode.write
+        return clib.SOMADataFrame.open(
+            uri,
+            open_mode,
+            {k: str(v) for k, v in context.tiledb_config.items()},
+            [],
+            clib.ResultOrder.automatic,
+            (0, timestamp),
```
thetorpedodog, line 334: named arguments for safety and readability?

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -247,6 +313,104 @@ def _do_initial_reads(self, reader: tiledb.Group) -> None:
...
+    def schema(self) -> pa.Schema:
+        return self._handle.schema
+
+    @property
+    def meta(self) -> "MetadataWrapper":
+        return MetadataWrapper(self, dict(self._handle.meta))
+
+    @property
+    def ndim(self) -> int:
+        return int(len(self._handle.index_column_names))
```
thetorpedodog, line 358: is there ever a case where this will not already be an int?

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -247,6 +313,104 @@ def _do_initial_reads(self, reader: tiledb.Group) -> None:
         }
 
 
+class DataFrameWrapper(Wrapper[clib.SOMADataFrame]):
```
thetorpedodog, line 316: since this is for reads only, would it make sense to call this a DataFrameReadWrapper?

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -247,6 +313,104 @@ def _do_initial_reads(self, reader: tiledb.Group) -> None:
...
+                result.append(domain(name, dtype))
+        return tuple(result)
+
+    @property
+    def domain(self) -> Tuple[Tuple[Any, Any], ...]:
+        return self._cast_domain(self._handle.domain)
+
+    def non_empty_domain(self) -> Optional[Tuple[Tuple[Any, Any], ...]]:
+        result = self._cast_domain(self._handle.non_empty_domain)
+        return None if len(result) == 0 else result
```
thetorpedodog, line 395: `return result or None`, since an empty tuple is falsey

apis/python/src/tiledbsoma/_util.py
```diff
@@ -260,3 +262,26 @@ def ms_to_datetime(millis: int) -> datetime.datetime:
...
+
+
+def to_clib_result_order(result_order: options.ResultOrderStr) -> clib.ResultOrder:
+    to_clib_result_order = {
+        options.ResultOrder.AUTO: clib.ResultOrder.automatic,
+        options.ResultOrder.ROW_MAJOR: clib.ResultOrder.rowmajor,
+        options.ResultOrder.COLUMN_MAJOR: clib.ResultOrder.colmajor,
+        "auto": clib.ResultOrder.automatic,
+        "row-major": clib.ResultOrder.rowmajor,
+        "column-major": clib.ResultOrder.colmajor,
```
thetorpedodog, line 274: you can avoid this duplication by using the fact that you can turn a string into the enum value:

```python
result_order = options.ResultOrder(result_order)
to_clib_result_order = {...}
try:
  return to_clib_result_order
except KeyError as ke:
  raise ValueError(...) from ke

**Review by thetorpedodog — COMMENTED** [review]
(the usual disclaimer that I’m mostly looking at the Python side for style etc. and didn’t look at the C++ code in detail)

**Comment by thetorpedodog** [comment]
> (the usual disclaimer that I’m mostly looking at the Python side for style etc. and didn’t look at the C++ code in detail)

I should add that I like the way the open-and-select-the-right-wrapper thing works now. It’s very well structured and understandable.

File: apis/python/src/tiledbsoma/_tdb_handles.py (hunk 16 above)
Comment (nguyenv) [reply apis/python/src/tiledbsoma/_tdb_handles.py:316]: I am going to change the write path to use `DataFrameWrapper` too in a future PR.

File: libtiledbsoma/test/unit_soma_dense_ndarray.cc [hunk 19]
```diff
@@ -87,9 +87,6 @@ TEST_CASE("SOMADenseNDArray: basic") {
     REQUIRE(soma_dense->ctx() == ctx);
     REQUIRE(soma_dense->type() == "SOMADenseNDArray");
     REQUIRE(soma_dense->is_sparse() == false);
-    auto schema = soma_dense->schema();
```
johnkerl, line 90: why the removes?

libtiledbsoma/test/unit_soma_sparse_ndarray.cc
```diff
@@ -87,9 +87,6 @@ TEST_CASE("SOMASparseNDArray: basic") {
     REQUIRE(soma_sparse->ctx() == ctx);
     REQUIRE(soma_sparse->type() == "SOMASparseNDArray");
     REQUIRE(soma_sparse->is_sparse() == true);
-    auto schema = soma_sparse->schema();
```
johnkerl, line 90: ditto

libtiledbsoma/test/unit_soma_dataframe.cc
```diff
@@ -59,7 +59,7 @@ const std::string src_path = TILEDBSOMA_SOURCE_ROOT;
 namespace {
 ArraySchema create_schema(Context& ctx, bool allow_duplicates = false) {
     // Create schema
-    ArraySchema schema(ctx, TILEDB_DENSE);
+    ArraySchema schema(ctx, TILEDB_SPARSE);
```
johnkerl, line 62: ✅

libtiledbsoma/test/unit_soma_array.cc
```diff
@@ -127,10 +127,6 @@ std::tuple<std::vector<int64_t>, std::vector<int>> write_array(
             ResultOrder::automatic,
             std::pair<uint64_t, uint64_t>(timestamp + i, timestamp + i));
 
-        if (LOG_DEBUG_ENABLED()) {
```
johnkerl, line 130: why remove?

libtiledbsoma/src/soma/soma_sparse_ndarray.h
```diff
@@ -200,11 +199,11 @@ class SOMASparseNDArray : public SOMAObject {
     const std::string uri() const;
 
     /**
-     * Return data schema, in the form of a TileDB ArraySchema.
+     * Return the data schema, in the form of a ArrowSchema.
```
johnkerl, line 202:
```suggestion
     * Return the data schema, in the form of an ArrowSchema.
```

libtiledbsoma/src/soma/soma_dense_ndarray.h
```diff
@@ -200,11 +199,11 @@ class SOMADenseNDArray : public SOMAObject {
     const std::string uri() const;
 
     /**
-     * Return data schema, in the form of a TileDB ArraySchema.
+     * Return the data schema, in the form of a ArrowSchema.
```
johnkerl, line 202:
```suggestion
     * Return the data schema, in the form of an ArrowSchema.
```

apis/python/src/tiledbsoma/soma_array.cc
```diff
@@ -0,0 +1,506 @@
...
+        // and string indices. In a subsequent version we are now supporting
+        // various NumPy/PyArrow types including float32, float64, int8, uint16,
+        // etc. It is an unfortunate fact that pybind11 does _not_ successfully
+        // disambiguate between float32 and float64, or between int8 and int64,
+        // etc. given that we ask it to disambiguate using not just types but
+        // std::vector of types or std::vector of std::pair of types.
+        // Experiments have shown that when both float32 and float64 are
+        // implemented with overloaded names to be differentiated solely by
+        // type, pybind11 uses the _first found_. Therefore it is necessary for
+        // us to no longer use common overloaded names.
```
johnkerl, line 388: Thanks for leaving this in

That was the result of a few long days of hard debugging ... very well-earned wisdom. :)

apis/python/tests/test_platform_config.py
```diff
@@ -70,9 +70,20 @@ def test_platform_config(adata):
...
+            # the ArraySchema metadata very similar to how Pandas stores information
+            # within pa.Schema.pandas_metadata. This could hold not only which
+            # filters have been applied to the column, but other info that cannot
+            # be "directly" stored in the ArrowSchema such as whether the column
+            # is a TileDB attribute or dimension, whether this represent a dense
+            # or sparse array, etc. This may be as easy as simply copying the
+            # platform_config by calling pa.Schema.with_metadata(platform_config).
+            # var_df = exp.ms["RNA"].var
+            # var_arr = var_df._handle.reader
+            # assert var_arr.dim("soma_joinid").filters == [tiledb.ZstdFilter(level=1)]
```
johnkerl, line 86: :spar

libtiledbsoma/src/soma/soma_array.h
```diff
@@ -185,6 +186,19 @@ class SOMAArray {
      */
     std::shared_ptr<Context> ctx();
 
+    std::optional<std::string> type() {
```
johnkerl, line 189:
```suggestion
    std::optional<std::string> soma_object_type() {
```

The word `type` is too overloaded already

libtiledbsoma/src/soma/soma_array.h
```diff
@@ -472,14 +491,24 @@ class SOMAArray {
     uint64_t nnz();
 
     /**
-     * @brief Get the schema of the array.
+     * @brief Get the TileDB ArraySchema. This should eventually
+     * be removed in lieu of arrow_schema below.
      *
      * @return std::shared_ptr<ArraySchema> Schema
      */
     std::shared_ptr<ArraySchema> schema() const {
```
johnkerl, line 499:
```suggestion
    std::shared_ptr<ArraySchema> tiledb_schema() const {
```

libtiledbsoma/src/soma/soma_dataframe.cc
```diff
@@ -105,19 +105,37 @@ SOMADataFrame::SOMADataFrame(
...
     OpenMode mode, std::optional<std::pair<uint64_t, uint64_t>> timestamp) {
     array_->open(mode, timestamp);
-    array_->reset();
 }
 
 void SOMADataFrame::close() {
     array_->close();
 }
 
+bool SOMADataFrame::exists(std::string_view uri) {
```
johnkerl, line 119: this seems like the kind of thing that could be pushed down more into C++ (not on this PR, which is big enough already) :)

johnkerl (changes requested):
Nothing huge to sugggest -- some spelling changes etc

apis/python/src/tiledbsoma/soma_array.cc
```diff
@@ -0,0 +1,506 @@
...
+                           // Update the column_names list if it was not empty,
+                           // otherwise continue selecting all columns with an
+                           // empty column_names list
+                           if (!column_names.empty()) {
+                               column_names = new_column_names;
+                           }
+                       } catch (const std::exception& e) {
+                           TPY_ERROR_LOC(e.what());
+                       }   
+                       qc = py_query_condition.attr("c_obj")
```
ihnorton, line 151: What is the lifetime here?

apis/python/src/tiledbsoma/soma_array.cc
```diff
@@ -0,0 +1,506 @@
...
+        .def(
+            "reset",
+            [](SOMAArray& reader,
+               std::optional<std::vector<std::string>> column_names_in,
+               std::string_view batch_size,
+               ResultOrder result_order) {
+                // Handle optional args
+                std::vector<std::string> column_names;
+                if (column_names_in) {
+                    column_names = *column_names_in;
```
ihnorton, line 178: Do we know that this is always valid? We'll get a runtime_error if the optional is empty.

apis/python/src/tiledbsoma/soma_array.cc
```diff
@@ -0,0 +1,506 @@
...
+        // disambiguate between float32 and float64, or between int8 and int64,
+        // etc. given that we ask it to disambiguate using not just types but
+        // std::vector of types or std::vector of std::pair of types.
+        // Experiments have shown that when both float32 and float64 are
+        // implemented with overloaded names to be differentiated solely by
+        // type, pybind11 uses the _first found_. Therefore it is necessary for
+        // us to no longer use common overloaded names.
+
+        .def(
+            "set_dim_points_string_or_bytes",
```
ihnorton, line 312: Should these be `_` prefixed to indicate they are internal?

libtiledbsoma/test/unit_soma_dense_ndarray.cc
nguyenv: The `schema` method now returns an `ArrowSchema` instead of a TileDB `ArraySchema`. I can change modify the test to reflect this instead of just removing it altogether.

apis/python/src/tiledbsoma/soma_array.cc
nguyenv: I think this is checking if the optional is empty or not.

apis/python/src/tiledbsoma/soma_array.cc
nguyenv: No we've been using `SOMAArray.set_dim...` in this code for a while now so I don't want to change the syntax now.

apis/python/src/tiledbsoma/soma_array.cc
nguyenv: For the lifetime of the `ManagedQuery` / TileDB `Query` associated with the `SOMAArray`. We set the query condition with `SOMAArray::set_condition` -> `ManagedQuery::set_condition` -> `tiledb::Query::set_condition`. Everytime we run a new query, we construct a new `ManagedQuery` object which in turn constructs a new `Query` object.

apis/python/src/tiledbsoma/soma_array.cc
```diff
@@ -0,0 +1,506 @@
...
+                       qc = py_query_condition.attr("c_obj")
+                                   .cast<PyQueryCondition>()
+                                   .ptr()
+                                   .get();
+                   }   
+                   reader.reset(column_names);
+
+                    // Release python GIL after we're done accessing python
+                   // objects
+                   py::gil_scoped_release release;
```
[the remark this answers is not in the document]
nguyenv: No the call below sets the query condition to the `SOMAArray` which is not very expensive. Is it faster to not call this here?

thetorpedodog (commented):
Apologies if some of these are out of date; I forgot to hit "send" on a review long ago.

eddelbuettel (approved):
This is a rather large PR and it touches a substantial amount of Python code (that I am not the best person to review for) as well as C++ code in libtlledbsoma. The latter seems fine but we may want to wait for two or three approvals here given the size and scope.

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -521,37 +517,43 @@ def _set_reader_coord(
...
+            dim_type: Union[Type[str], Type[bytes]] = type(domain[0])
             # A ``None`` or empty start is always equivalent to empty str/bytes.
             start = coord.start or dim_type()
             if coord.stop is None:
                 # There's no way to specify "to infinity" for strings.
                 # We have to get the nonempty domain and use that as the end.
-                _, stop = self._handle.reader.nonempty_domain()[dim_idx]
+                ned = self._handle.non_empty_domain()
+                if ned is None:
+                    raise ValueError(
```
johnkerl, line 537: Worth adding a test for, if this is not a false positive

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -574,41 +576,18 @@ def _set_reader_coord_by_py_seq_or_np_array(
...
-
-        elif dim.dtype == np.float64:
-            sr.set_dim_points_float64(dim.name, coord)
-        elif dim.dtype == np.float32:
-            sr.set_dim_points_float32(dim.name, coord)
-
-        elif dim.dtype == "str" or dim.dtype == "bytes":
-            sr.set_dim_points_string_or_bytes(dim.name, coord)
+        try:
+            set_dim_points = getattr(sr, f"set_dim_points_{dim.type}")
```
johnkerl, line 580: Nice neaten :)

apis/python/src/tiledbsoma/_query_condition.py
```diff
@@ -130,15 +131,17 @@ def __attrs_post_init__(self):
...
+        schema: pa.Schema,
         query_attrs: Optional[List[str]],
     ):
-        qctree = QueryConditionTree(schema, enum_to_dtype, query_attrs)
-        self.c_obj = qctree.visit(self.tree.body)
+        try:
+            qctree = QueryConditionTree(schema, query_attrs)
+            self.c_obj = qctree.visit(self.tree.body)
+        except Exception as pex:
+            raise SOMAError(pex)
```
johnkerl, line 141: Ditto re above: worth adding a test for if this isn't a false positive by the coverage tool

apis/python/tests/test_experiment_query.py
```diff
@@ -628,23 +628,23 @@ def test_experiment_query_to_anndata_obsm_varm(soma_experiment):
...
-        )
-
-        assert set(ad.varp.keys()) == {"bar"}
-        varp = ad.varp["bar"]
-        assert isinstance(varp, np.ndarray)
-        assert varp.shape == (query.n_vars, query.n_vars)
-        assert np.array_equal(
-            query.varp("bar").coos().concat().to_scipy().todense(), varp
-        )
+        query.to_anndata("raw", obsp_layers=["foo"], varp_layers=["bar"])
```
johnkerl, line 631: Please uncomment/restore, or, remove entirely (and say why)

johnkerl (approved):

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -341,25 +334,28 @@ def read(
...
+        if self._handle._handle.timestamp is not None:
+            ts = (0, self._handle._handle.timestamp)
+
+        sr = clib.SOMADataFrame.open(
+            self._handle._handle.uri,
+            clib.OpenMode.read,
+            platform_config or {},
+            column_names or [],
+            _util.to_clib_result_order(result_order),
+            ts,
```
thetorpedodog, line 351: one last bump on the suggestion to pass these as named arguments; particularly since we don’t have static typechecking on this, there is no obvious order for these

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -574,41 +576,18 @@ def _set_reader_coord_by_py_seq_or_np_array(
...
+            return True
 
-        elif (
-            dim.dtype == "datetime64[s]"
-            or dim.dtype == "datetime64[ms]"
-            or dim.dtype == "datetime64[us]"
-            or dim.dtype == "datetime64[ns]"
-        ):
+        if _util.pa_types_is_string_or_bytes(dim.type):
+            sr.set_dim_points_string_or_bytes(dim.name, coord)
```
thetorpedodog, line 589: would recommend the structure

```python
if condition:
  do_thing()
  return True
if other_condition():
  ...
  return True
# fell through everything
raise ValueError(...)
```

so that the reader doesn’t have to go past all the if-elses to find the return (and the `raise` doesn’t need to live in its own `else` either).

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -194,16 +243,33 @@ def _opener(
...
-        creation time, but if for example only a portion of the available domain
-        has actually had data written, this function will return a tighter
-        range.
-        """
-        return self._handle.nonempty_domain()  # type: ignore
+    def non_empty_domain(self) -> Optional[Tuple[Tuple[object, object], ...]]:
+        try:
+            ned = self._handle.nonempty_domain()
+            if ned is None:
+                return None
```
thetorpedodog, line 250: Would it make sense to return `()` here and keep this function non-Optional? If there is a reason to return `None` specifically, that is also fine.

apis/python/src/tiledbsoma/_tiledb_array.py
```diff
@@ -65,9 +65,14 @@ def schema(self) -> pa.Schema:
         Lifecycle:
             Experimental.
         """
-        return tiledb_schema_to_arrow(self._tiledb_array_schema(), self.uri, self._ctx)
+        if isinstance(self._tiledb_array_schema(), tiledb.ArraySchema):
+            return tiledb_schema_to_arrow(
+                self._tiledb_array_schema(), self.uri, self._ctx
+            )
+        else:
```
thetorpedodog, line 72: nit: no else needed after a return

nguyenv: Changed it. Is it better for readability or something to not have the else here?

thetorpedodog: yeah, it reduces redundancy and drops an indent level, as well as making it obvious that you cannot get out of the if-block

thetorpedodog (approved):

nguyenv:
@beroy I have now merged your changes into my PR. The only change I did was move your reindexer unit tests to reside within the `apis/python/tests` directory. This is a part of changes within this PR where I've moved all Python unit tests from `libtilesoma/tests` into the Python API directory as to not have them split between two directories. Because this PR is already large enough as is, I have decided to leave the reindexer bindings within `pytiledbsoma.cc` for now but will be opening a subsequent PR where this gets separated out into its own `reindexer.cc`.

PR #1802, opened 2023-10-18
[r] Bump to 1.4.4 in `apis/r/DESCRIPTON` on `release-1.4`

johnkerl:
Following
https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases


eddelbuettel (approved):
The change is of course miniscule and fine.  The real effect will be the tagging as a release as only it will trigger new r-universe builds.

Issue #1769, opened 2023-10-09
[python] `tiledbsoma.io.update_uns` phase one


johnkerl:
This user-requested feature will be a companion of the existing

* `tiledbsoma.io.update_obs`
* `tiledbsoma.io.update_var`
* `tilebsoma.io.add_X_layer`
* `tilebsoma.io.add_matrix_to_collection`

[[sc-35349]](https://app.shortcut.com/tiledb-inc/story/35349/python-tiledbsoma-io-update-uns)

Only adding non-existing items from new `adata.uns` to existing `exp.uns`. This is the first step in the parent issue #2707.

johnkerl:
[sc-35349]

PR #1810, opened 2023-10-19
[python] Support default-coord reads of corner-written dense ND arrays

johnkerl:
**Issue and/or context:** In support of PR #1768 with issue #1769.

Needles to thread for `update_uns`:

The existing `tiledbsoma.io.update_obs`, `tiledbsoma.io.update_var`, `tiledbsoma.io.add_X_layer` methods all take as argument a `tiledbsoma.Experiment` which is already open for writing. Goal: do the same for `tiledbsoma.io.update_uns`.

Common use-case:
* Ingest data from AnnData to SOMA
* Outgest to AnnData; run ScanPy, curation, etc
* Write the updates back to SOMA

Technical factors:
* Whereas all the `X`, `obsm`, `varm`, `obsp`, `varp` are `SparseNDArray`, `uns` can contain `DenseNDArray` which are ingested from `numpy.ndarray`
* If the after-uns has `numpy.ndarray` that the before-uns didn't (new data on update), there is no problem
* If the after-uns has `numpy.ndarray` of the same dimensions as the before-uns, there is no problem
* Schema evolution does not support modification of dense domains
* We can try delete-and-re-add:
  * `tiledb.Array.delete_array` of the array accompanied by tiledb-group operations to delete the key/URI pair and re-add the key/URI pair
  * Note that for local-disk and S3, the new URI will be the same as the old URI; for cloud the new and old URIs will always be different URIs
  * Problem with that: we cannot delete-and-re-add at the same timestamp
  * If we do `del collection[key]` and `collection[key] = new_dense_nd_array` then we get `tiledbsoma._exception.SOMAError: replacing key 'a' is unsupported` which is a check that must remain (see links below)
* Also we cannot do: del collection[key], close the collection, and re-open at the same timestamp because of
  * https://github.com/single-cell-data/TileDB-SOMA/blob/1.5.0rc2/apis/python/src/tiledbsoma/_collection.py#L567-L570
  * https://github.com/single-cell-data/TileDB-SOMA/pull/957
  * https://github.com/single-cell-data/TileDB-SOMA/issues/919
* We could have `update_uns` do two opens-for-write: one to remove any shape-changing dense arrays, and one shifted by a millisecond to do the rest of the work
  * However this would mean `tiledbsoma.io.update_uns` would take a URI as first argument, different from all the other updaters in `tiledbsoma.io` which take an `Experiment` opened for write as first argument

Good feedback from @isaiah and @seth: simply have `tiledbsoma.io`'s `uns`-ingestion logic write bigger-than-necessary arrays for dense ND arrays -- as we already do for sparse.

This is compliant with the [SOMA specification](https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md#indexing-and-slicing).

**Changes:** As above.

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/_dense_nd_array.py
+++ b/apis/python/src/tiledbsoma/_dense_nd_array.py
@@ -11,6 +11,7 @@ from typing import Optional, Tuple
 
 import pyarrow as pa
 import somacore
+import tiledb
 from somacore import options
 from typing_extensions import Self
 
@@ -71,6 +72,18 @@ class DenseNDArray(NDArray, somacore.DenseNDArray):
 
     __slots__ = ()
 
+    def non_empty_domain(self) -> Tuple[Tuple[int, int], ...]:
+        """
+        Retrieves the non-empty domain for each dimension, namely the smallest
+        and largest indices in each dimension for which the sparse array has
+        data occupied.  This is nominally the same as the domain used at
+        creation time, but if for example only a corner of the available domain
+        has actually had data written, this function will return a tighter
+        range.
+        """
+        with tiledb.open(self.uri, ctx=self.context.tiledb_ctx) as A:
+            return A.nonempty_domain()  # type: ignore
+
     def read(
         self,
         coords: options.DenseNDCoords = (),
@@ -108,8 +121,23 @@ class DenseNDArray(NDArray, somacore.DenseNDArray):
         self._check_open_read()
         result_order = somacore.ResultOrder(result_order)
 
-        schema = self._handle.schema
-        target_shape = dense_indices_to_shape(coords, schema.shape, result_order)
+        # The dense_indices_to_shape includes, as one of its roles, how to handle default
+        # coordinates -- e.g. `dnda.read()`. The default for a DenseNDArray should be "all the data"
+        # -- but what is that? If the schema shape matches the non-empty domain -- e.g. at create,
+        # shape was 100x200, and at write, 100x200 cells were written, those are both the same. But
+        # if the array was written with room for growth -- e.g. created with shape
+        # 1,000,000x1,000,000 but only 100x200 cells were written -- then we need the non-empty
+        # domain.
+        #
+        # The non-empty domain is the corret choice in either case.
+        #
+        # The only exception is if the array has been created but no data have been written at
+        # all, in which case the best we can do is use the schema shape.
+        data_shape = self._handle.schema.shape
+        ned = self.non_empty_domain()
+        if ned is not None:
+            data_shape = tuple(slot[1] + 1 for slot in ned)
+        target_shape = dense_indices_to_shape(coords, data_shape, result_order)
 
         sr = self._soma_reader(result_order=result_order)
 
--- a/apis/python/src/tiledbsoma/_dense_nd_array.py
+++ b/apis/python/src/tiledbsoma/_dense_nd_array.py
@@ -11,7 +11,6 @@ from typing import Optional, Tuple
 
 import pyarrow as pa
 import somacore
-import tiledb
 from somacore import options
 from typing_extensions import Self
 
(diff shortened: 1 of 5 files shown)


apis/python/src/tiledbsoma/_dense_nd_array.py
```diff
@@ -71,6 +72,18 @@ class DenseNDArray(NDArray, somacore.DenseNDArray):
...
+        Retrieves the non-empty domain for each dimension, namely the smallest
+        and largest indices in each dimension for which the sparse array has
+        data occupied.  This is nominally the same as the domain used at
+        creation time, but if for example only a corner of the available domain
+        has actually had data written, this function will return a tighter
+        range.
+        """
+        with tiledb.open(self.uri, ctx=self.context.tiledb_ctx) as A:
+            return A.nonempty_domain()  # type: ignore
+
```
nguyenv, line 86: What I recommend instead is to remove this, and add a `TileDBArray._tiledb_nonempty_domain()` function into [_tiledb_array.py](https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/python/src/tiledbsoma/_tiledb_array.py#L45). This uses the already opened handle and does not introduce a new `DenseNDArray.non_empty_domain`.

On this line, call `ned = self._handle._tiledb_nonempty_domain()` which then just keep getting the NED internal.
https://github.com/single-cell-data/TileDB-SOMA/pull/1810/files#diff-1c8b39ff256e708f483b5e9ebdb3c0dd07f14a78cd487cbf1e8d7c49373d9997R137

nguyenv: Also minor nit, but I think we favor naming as `nonempty_domain` over `non_empty_domain`.

johnkerl: @nguyenv it's already released for sparse (i'm adding it for dense) and i don't want a backward-incompatible name change

thanks though!

nguyenv: OK nevermind about the name change. But I think the fact that both `SparseNDArray` and `DenseNDArray` use this is even more reason to add this into `TileDBArray` and then in both classes use

```
    def non_empty_domain(self) -> Tuple[Tuple[int, int], ...]:
        """
        Retrieves the non-empty domain for each dimension, namely the smallest and
        largest indices in each dimension for which the sparse array has data occupied.
        This is nominally the same as ``used_shape``, but if for example the
        leading/trailing rows/columns of the sparse array are entirely unoccupied, this
        function will return a tighter range.
        """
        return self._handle._tiledb_nonempty_domain()
```

johnkerl: just pushed a commit w/ that change! :)

johnkerl:
```
>>> exp.obs._handle. <---- tab-complete
exp.obs._handle.close(       exp.obs._handle.metadata     exp.obs._handle.reader       exp.obs._handle.uri
exp.obs._handle.closed       exp.obs._handle.mode         exp.obs._handle.schema       exp.obs._handle.writer
exp.obs._handle.context      exp.obs._handle.open(        exp.obs._handle.timestamp_ms

>>> type(exp.obs._handle)
<class 'tiledbsoma._tdb_handles.ArrayWrapper'>

>>> exp.obs._handle._
exp.obs._handle._abc_impl          exp.obs._handle._flush_hack(       exp.obs._handle._opener(
exp.obs._handle._check_open(       exp.obs._handle._handle
exp.obs._handle._do_initial_reads( exp.obs._handle._is_protocol
```

so we would need a follow-on once we've gotten `nonempty_domain` wired in from C++

nguyenv (approved):
Thanks!

PR #1814, opened 2023-10-20
[python/c++] Expose `uri`, `result_order`, and `column_names` for `SOMAArray`

nguyenv:
**Issue and/or context:**

[sc-35643]

**Changes:**

- Add `sr` (`SOMAArray`) Python bindings for `uri`, `result_order`, and `column_names`
- Add C++ `SOMAArray::result_order` which calls newly added `ManagedQuery::query_layout`
- Replace `SOMAArray::result_order_` with call to `ManagedQuery` to get `query_layout`. If the result order was set to `ResultOrder::automatic`, just returning back `ResultOrder::automatic` is not very useful information. The user probably wants to know if the layout is row-major or col-major

**Notes for Reviewer:**

A `schema` getter was also requested but is not possible yet in this PR but [will be available](https://github.com/single-cell-data/TileDB-SOMA/pull/1793) soon.

Currently `SOMAArray::schema` returns a `tiledb::ArraySchema`. In TileDB-Py, we do have a registered binding from `tiledb::ArraySchema` to `tiledb.ArraySchema`. This does not get picked up in TileDB-SOMA. In TileDB-SOMA, we "could" bind `tiledb::ArraySchema` to `tiledbsoma.ArraySchema` but TileDB-Py's `tiledb.ArraySchema` and TileDB-SOMA's `tiledbsoma.ArraySchema` are not compatible despite binding the same underlying C++ class. We would also have to separately bind all classes used by the `ArraySchema` such as `tiledb::Attribute`, `tiledb::Dimension`, `tiledb::Filter`, etc. which is massively inconvenient.

The refactored read path no longer uses the `tiledb::ArraySchema` and instead returns an `ArrowSchema`. From the C++ side we take the [TileDB ArraySchema](https://github.com/single-cell-data/TileDB-SOMA/pull/1793/files#diff-5783c742aaee44efc3a6c4be58bd289681d5ebaf421be55de8e642d1f1053c5dR474-R481) and converts this to an [ArrowSchema](https://github.com/single-cell-data/TileDB-SOMA/pull/1793/files#diff-e419a62446641fa84b7b42a6ab37141a624cb49a0784add4f49823cd1abef3d5R232-R295). In the bindings, the C-struct `ArrowSchema` is [converted](https://github.com/single-cell-data/TileDB-SOMA/pull/1793/files#diff-1eece2b39a5f3e4367ab323e3ca6d10cdd83de2b41daa25a652392907a4a0de6R104-R108) to a `pa.Schema`.

### DIFF
--- a/libtiledbsoma/test/test_soma_array.py
+++ b/libtiledbsoma/test/test_soma_array.py
@@ -38,6 +38,7 @@ def test_soma_array_basic_getters():
     assert sr.result_order == result_order
     assert sr.column_names == column_names
 
+
 def test_soma_array_obs():
     """Read all values from obs array into an arrow table."""
 
--- a/libtiledbsoma/test/test_soma_array.py
+++ b/libtiledbsoma/test/test_soma_array.py
@@ -15,6 +15,29 @@ if VERBOSE:
     clib.config_logging("debug")
 
 
+def test_soma_array_basic_getters():
+    name = "obs"
+    uri = os.path.join(SOMA_URI, name)
+    column_names = [
+        "soma_joinid",
+        "obs_id",
+        "n_genes",
+        "percent_mito",
+        "n_counts",
+        "louvain",
+        "is_b_cell",
+    ]
+    result_order = clib.ResultOrder.colmajor
+
+    sr = clib.SOMAArray(
+        uri=uri, name=name, column_names=column_names, result_order=result_order
+    )
+
+    assert sr.uri == uri
+    assert sr.nnz() == 2638
+    assert sr.result_order == result_order
+    assert sr.column_names == column_names
+
 def test_soma_array_obs():
     """Read all values from obs array into an arrow table."""
 
@@ -223,16 +246,6 @@ def test_soma_array_column_names():
     # test that all results are present in the arrow table (no incomplete queries)
     assert sr.results_complete()
     assert arrow_table.num_columns == 2
-
-
-def test_nnz():
-    name = "obs"
-    uri = os.path.join(SOMA_URI, name)
-    sr = clib.SOMAArray(uri)
-
-    total_cell_count = sr.nnz()
-
-    assert total_cell_count == 2638
 
 
 def test_soma_array_reset():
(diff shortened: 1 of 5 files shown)


libtiledbsoma/test/test_soma_array.py
```diff
@@ -15,6 +15,35 @@
...
+    result_order = clib.ResultOrder.colmajor
+
+    sr = clib.SOMAArray(
+        uri=uri, name=name, column_names=column_names, result_order=result_order
+    )
+
+    assert sr.uri == uri
+    assert sr.nnz() == 2638
+
+    # Note that column_names is set AFTER read_next even though it is an arugment
```
bkmartinjr, line 39: Would be nice-to-have if this prop was supported even before doing a read_next.  I.e., given a SOMARead, you can always get the column names regardless of read state.

I did a quick code read, and it looks like reset is called during the SOMAArray constructor.  This initial reset() call doesn't include the column names, so the ManagedQuery.  If the initial reset included the column names, would this issue be resolved?

bkmartinjr (changes requested):
My only comment is that having the query column_names state be dependent on doing a read is not great, as it means that any users of this interface have to track additional state about the object in order to use this prop.

nguyenv:
OK I'll change it to make the info available before reads too.

bkmartinjr:
> Replace SOMAArray::result_order_ with call to ManagedQuery to get query_layout. If the result order was set to ResultOrder::automatic, just returning back ResultOrder::automatic is not very useful information. The user probably wants to know if the layout is row-major or col-major


I think it is also useful to know what the user asked for, i.e.,. will sorting be applied?  Currently there is a large performance penalty if result order is either row or col, but not if it is auto due to different readers being used in the core DB.  This is independent of layout.

I'd rather have layout available as part of the schema definition, and not mixed together with query definition.

nguyenv:
Upon looking at the code again, it seems I confused myself too while making the changes. I reverted `result_order_` back to how it was and am now just returning that.

bkmartinjr (approved):
LGTM!

PR #1817, opened 2023-10-23
[python] `NDArray` read path

nguyenv:
**Issue and/or context:**

To be merged on top of #1793.

**Changes:**

- When opening a `SparseNDArray` in read-mode, use `SparseNDArrayWrapper` which wraps around `clib.SOMASparseNDArray`
- When opening a `DenseNDArray` in read-mode, use `DenseNDArrayWrapper` which wraps around `clib.SOMADenseNDArray`
- Completely remove `SOMAArray`
- When the R API replaces `SOMAArray` with the new `SOMADataFrame`, `SOMASparseNDArray`, and `SOMADenseNDArray` classes, we can reorganize the C++ install headers so that they no longer install internal headers


apis/python/src/tiledbsoma/_collection.py
```diff
@@ -429,15 +429,20 @@ def __getitem__(self, key: str) -> CollectionElementType:
             from . import _factory  # Delayed binding to resolve circular import.
             from ._tdb_handles import Wrapper
 
+            uri = entry.entry.uri
             wrapper: type[Wrapper[Any | Any | Any]]
-            if self.mode == "r" and clib.SOMADataFrame.exists(entry.entry.uri):
+            if self.mode == "r" and clib.SOMADataFrame.exists(uri):
```
thetorpedodog, line 434: I probably should have noticed this earlier, but: we probably need to be passing the context we have to this function, right? It might contain user credentials or other important configuration (e.g. endpoint locations)

apis/python/src/tiledbsoma/_common_nd_array.py
```diff
@@ -105,7 +106,8 @@ def shape(self) -> Tuple[int, ...]:
         Lifecycle:
             Experimental.
         """
-        return cast(Tuple[int, ...], tuple(self._soma_reader().shape))
+        handle: SOMAArray = self._handle
```
thetorpedodog, line 109: This comment is only tangentially related to this line, so:

Given that `_handle` is now something different for all of the TileDB Array–based types, it’s probably worth it to pull that generic specialization out of `TileDBArray` and put it into each of the concrete ones (so this becomes a `TileDBArray[SOMAArray]`, or however the wrapper type generic stuff works because I kind of forget, etc.).

This isn’t *critical*, so if it ends up being a mess I wouldn’t worry about it but if it is reasonably straightforward it’s worth considering.

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -270,18 +270,16 @@ def count(self) -> int:
...
-        """Doc place holder.
-
-        Returns:
-            Tuple[Any, ...]: _description_
-        """
-        return tuple(self._soma_reader().get_enum(name))
-
-    def column_to_enumeration(self, name: str) -> str:
-        return str(self._soma_reader().get_enum_label_on_attr(name))
+    def enumeration(self, name: str) -> Optional[Tuple[Any, ...]]:
```
thetorpedodog, line 276: To avoid use of `Any`, it may be better to call this `Tuple[object, ...]`. Doing so would have the advantage that the caller would need to either check or assert that it gets the returned type it wants, which is safer. But also it has the disadvantage that the caller would need to either check or assert that it gets the returned type it wants, which is kind of annoying.

apis/python/src/tiledbsoma/_factory.py
```diff
@@ -154,12 +174,21 @@ def _reify_handle(hdl: _Wrapper) -> "_tiledb_object.TileDBObject[_Wrapper]":
...
+        or (
+            typename == "SOMADenseNDArray"
+            and isinstance(hdl, _tdb_handles.DenseNDArrayWrapper)
+        )
     ):
         raise SOMAError(
             f"cannot open {hdl.uri!r}: a {type(hdl._handle)}"
             f" cannot be converted to a {typename}"
         )
+    print(typename, cls, type(hdl))
```
thetorpedodog, line 191: stray debug print

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -270,18 +270,16 @@ def count(self) -> int:
             Experimental.
         """
         self._check_open_read()
-        return cast(int, self._soma_reader().nnz())
+        handle: clib.DataFrameWrapper = self._handle
+        return cast(int, handle.count)
```
thetorpedodog, line 274: Could some of these be avoided by making a `libtiledbsoma.pyi` file? That would also give us some tooling help in editors. Up to you, and even if you do want to, it’s not something that needs to happen in this change.

apis/python/src/tiledbsoma/_dense_nd_array.py
```diff
@@ -120,13 +123,26 @@ def read(
...
         target_shape = dense_indices_to_shape(coords, data_shape, result_order)
 
-        sr = self._soma_reader(result_order=result_order)
+        to_clib_result_order = {
+            options.ResultOrder.AUTO: clib.ResultOrder.automatic,
+            options.ResultOrder.ROW_MAJOR: clib.ResultOrder.rowmajor,
+            options.ResultOrder.COLUMN_MAJOR: clib.ResultOrder.colmajor,
+            "auto": clib.ResultOrder.automatic,
+            "row-major": clib.ResultOrder.rowmajor,
+            "column-major": clib.ResultOrder.colmajor,
```
thetorpedodog, line 139: you can avoid this by writing `result_order = options.ResultOrder(result_order)` above, which will handle the enumification directly:

```
>>> import enum
>>> class X(enum.Enum):
...     ONE = 1
...     TWO = 2
...     THREE = 3
...     
... 
>>> X(1)
<X.ONE: 1>
>>> X(X.ONE)
<X.ONE: 1>
```

This would also obviate the `if not in` below.

apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -144,10 +145,22 @@ def read(
...
+        to_clib_result_order = {
+            options.ResultOrder.AUTO: clib.ResultOrder.automatic,
+            options.ResultOrder.ROW_MAJOR: clib.ResultOrder.rowmajor,
+            options.ResultOrder.COLUMN_MAJOR: clib.ResultOrder.colmajor,
+            "auto": clib.ResultOrder.automatic,
+            "row-major": clib.ResultOrder.rowmajor,
+            "column-major": clib.ResultOrder.colmajor,
+        }
+        if result_order not in to_clib_result_order:
+            raise ValueError(f"Invalid result_order: {result_order}")
```
thetorpedodog, line 157: to go with the above it looks like this could be pulled into a function.

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -369,6 +366,88 @@ def dim_names(self) -> Tuple[str, ...]:
...
+        timestamp: int,
+    ) -> clib.SOMADataFrame:
+        open_mode = clib.OpenMode.read if mode == "r" else clib.OpenMode.write
+        return clib.SOMADataFrame.open(
+            uri,
+            open_mode,
+            {k: str(v) for k, v in context.tiledb_config.items()},
+            [],
+            clib.ResultOrder.automatic,
+            (0, timestamp),
```
thetorpedodog, line 387: Could you add names to these arguments? The floating `[]` makes me nervous because I have no idea what it belongs to.

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -265,26 +280,8 @@ def _do_initial_reads(self, reader: tiledb.Group) -> None:
...
-        return clib.SOMADataFrame.open(
-            uri,
-            open_mode,
-            {k: str(v) for k, v in context.tiledb_config.items()},
-            [],
-            clib.ResultOrder.automatic,
-            (0, timestamp),
-        )
+class SOMAArrayWrapper(Wrapper[SOMAArray]):
+    """Wrapper for Array-derived SOMAObject classes."""
```
thetorpedodog, line 284: If you add `ARRAY_IMPL: Type[SOMAArray]` here…

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -369,6 +366,88 @@ def dim_names(self) -> Tuple[str, ...]:
...
+    ) -> clib.SOMADataFrame:
+        open_mode = clib.OpenMode.read if mode == "r" else clib.OpenMode.write
+        return clib.SOMADataFrame.open(
+            uri,
+            open_mode,
+            {k: str(v) for k, v in context.tiledb_config.items()},
+            [],
+            clib.ResultOrder.automatic,
+            (0, timestamp),
+        )
```
thetorpedodog, line 388: …and move this to `SOMAArrayWrapper`, replacing `clib.SOMADataFrame` with `cls.ARRAY_IMPL`, and set `ARRAY_IMPL = SOMADataFrame` here…

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -369,6 +366,88 @@ def dim_names(self) -> Tuple[str, ...]:
...
+    ) -> clib.SOMASparseNDArray:
+        open_mode = clib.OpenMode.read if mode == "r" else clib.OpenMode.write
+        return clib.SOMASparseNDArray.open(
+            uri,
+            open_mode,
+            {k: str(v) for k, v in context.tiledb_config.items()},
+            [],
+            clib.ResultOrder.automatic,
+            (0, timestamp),
+        )
```
thetorpedodog, line 414: …then we can eliminate all this duplicated code (by similarly setting an `ARRAY_IMPL` here and below).

thetorpedodog (commented):
With the caveat that I have not yet exensively looked at the C++ side of things, here is a pass on the Python side. Sorry for missing this yesterday!

apis/python/src/tiledbsoma/_tdb_handles.py
beroy: @nguyenv the PR also does a major refactoring on the pybind11 files which is not mentioned in the description! Would that stay as a part of this PR?

nguyenv: https://github.com/single-cell-data/TileDB-SOMA/pull/1793

It is in the description for the PR above which is higher prescendent than this PR. I should mark this PR as draft for now as I'm dealing with segfault issues arising from the newly blockchain iterator.

nguyenv: I will definitely help you with reorganizing the reindexer code if that is your main concern with the Pybind11 refactoring.

nguyenv:
Closing as these changes have now all been separated into more digestable PRs and are either already merged or ready to be reviewed.

https://github.com/single-cell-data/TileDB-SOMA/pull/2124
https://github.com/single-cell-data/TileDB-SOMA/pull/2126
https://github.com/single-cell-data/TileDB-SOMA/pull/2129
https://github.com/single-cell-data/TileDB-SOMA/pull/2132
https://github.com/single-cell-data/TileDB-SOMA/pull/2133

PR #1835, opened 2023-10-28
[r] Update `SOMADataFrame` creation and writes

eddelbuettel:
**Issue and/or context:**

TileDB Core 2.17.3 provides updated semantics for Enumerations (aka `factor` variables).  The TileDB R package (in upcoming release 0.21.2 or the current snapshot 0.21.1.10) connects this to R, and this PR lets the TileDB SOMA package take advantage of it.

**Changes:**

Data frame objects are created with 'empty' enumerations, i.e. without factor levels. These are added as needed during writes.

**Notes for Reviewer:**

~The PR is still in _draft_ form as it needed to 'park' four test files affected by an unrelated update to `SeuratObject`.~

The CI setup is once again modified to let is use 'newer than CRAN' package versions of TileDB R via r-universe.

We can lift the 'draft' status when either or both of these constraints can be lifted.

[SC 35945](https://app.shortcut.com/tiledb-inc/story/35945/tiledb-r-incorporate-core-enumeration-prs-for-zero-levels-at-create-and-level-extension-at-write)


eddelbuettel:
CI fails because I overlooked that TileDB R 0.2.1.10 does not automatically bring in TileDB Core 2.17.3.

apis/r/R/SOMADataFrame.R
```diff
@@ -123,10 +114,21 @@ SOMADataFrame <- R6::R6Class(
...
+        ## enumerations = if (any(!sapply(levels, is.null))) levels else NULL
         )
 
+      for (field_name in attr_column_names) {
+          fieldtype <- schema$GetFieldByName(field_name)$type
+          if (is(fieldtype, "DictionaryType")) {
+              tiledb::tiledb_array_schema_set_enumeration_empty(schema = tdb_schema,
+                                                                attr = tdb_attrs[[field_name]],
+                                                                enum_name = field_name,
+                                                                type_str = "UTF8",
```
johnkerl, line 127: while UTF-8 strings are indeed the nominal (and by far the most common) value type, they are quite general. we don't have existing unit-test coverage for this in tiledbsoma-r to date, but, i think this should also be `tiledb_type_from_arrow_type` here on the arrow-schema's value type -- so `arrow::string` would map to '"UTF8"`, and `arrow::float64` to `"FLOAT64"`, etc.

eddelbuettel: I am sorry but are asking for a function such as https://github.com/single-cell-data/TileDB-SOMA/blob/501ec8465b3e9bc1383c7e20cfc14d12bfddf00a/apis/r/R/utils-arrow.R#L43-L91 ?

johnkerl: @eddelbuettel yes

Arrow dictionary types have an index type -- often `int8` -- and a value type -- very often `string` -- but both can vary and both need to be looked up, and `tiledb_type_from_arrow_type` seems to be the best way to do that

johnkerl: @eddelbuettel IIUC we're lacking support, and unit-test, for non-string enumeration value types

eddelbuettel: @johnkerl Do you have an example data set with such factors?

johnkerl:
```
schema <- arrow::schema(
    arrow::field("soma_joinid", arrow::int64(), nullable = FALSE),
    arrow::field("int", arrow::int32(), nullable = FALSE),
    arrow::field("string", arrow::string(), nullable = FALSE),
    arrow::field(
        "enum",
        arrow::dictionary(index_type = arrow::int8(), value_type = arrow::type_name_goes_here(), ordered = FALSE), # <---
        nullable=FALSE
    )
)
cat("\n")
cat("SCHEMA\n")
print(schema)

sdf <- tiledbsoma::SOMADataFrameCreate(uri, schema = schema)

data <- arrow::arrow_table(
    soma_joinid = bit64::as.integer64(0L:4L),
    int = 100:104L,
    string = c("red", "yellow", "green", "red", "green"),
    enum = factor(c("red", "yellow", "green", "red", "green"),levels=c("red", "yellow", "green"))
)
cat("\n")
cat("DATA\n")
print(data)

sdf$write(data)

sdf$close()
```

eddelbuettel: Booleans are really integers underneath so even when I start from `factor(c(FALSE, TRUE, TRUE, FALSE))` as soon as you look at it how it realised it is `integer` inside:

```r
> factor(c(TRUE, FALSE, FALSE, TRUE))     # we can create from `logical` and it 'looks ok'
[1] TRUE  FALSE FALSE TRUE 
Levels: FALSE TRUE
> levels(factor(c(TRUE, FALSE, FALSE, TRUE)))
[1] "FALSE" "TRUE" 
> as.logical(factor(c(TRUE, FALSE, FALSE, TRUE)))
[1]  TRUE FALSE FALSE  TRUE
> str(factor(c(TRUE, FALSE, FALSE, TRUE)))    # but it actually drops the type info
 Factor w/ 2 levels "FALSE","TRUE": 2 1 1 2
>

Comment (eddelbuettel) [reply apis/r/R/SOMADataFrame.R:127]: The key may is in the `factor(c(..., levels=c(...)))` you omitted.  I am not sure we can actually create anything here.  I already demonstrated that `logical` only pretends to work (it becomes `integer`), other values fail too:

```r
> factor(sample(c(1.0,2.0), 10, TRUE), levels=c("FALSE", "TRUE"))    # cannot sample double() for factor
 [1] <NA> <NA> <NA> <NA> <NA> <NA> <NA> <NA> <NA> <NA>
Levels: FALSE TRUE
> 
```

eddelbuettel: Similarly if I make your script concrete and try an actual data object ... it becomes int + string via arrow.  

```sh
$ cat enum_non_int.R
#!/usr/bin/env Rscript

library(tiledb)
uri <- tempfile()
schema <- arrow::schema(arrow::field("soma_joinid", arrow::int64(), nullable = FALSE),
                        arrow::field("int", arrow::int32(), nullable = FALSE),
                        arrow::field("string", arrow::string(), nullable = FALSE),
                        arrow::field("enum", arrow::dictionary(index_type = arrow::int8(),
                                                               value_type = arrow::boolean(), #type_name_goes_here(), # <---- here
                                                               ordered = FALSE),
                                     nullable=FALSE))
cat("\nSCHEMA\n")
print(schema)

#sdf <- tiledbsoma::SOMADataFrameCreate(uri, schema = schema)

data <- arrow::arrow_table(soma_joinid = bit64::as.integer64(0L:4L),
                           int = 100:104L,
                           string = c("red", "yellow", "green", "red", "green"),
                           enum = factor(c(1,2,1,1,2), levels=c("green","red"))) # <--- values depending on type

cat("\nDATA\n")
print(data)

#sdf$write(data)
#sdf$close()
edd@rob:~/git/tiledb-adhoc/soma(master)$ Rscript enum_non_int.R

SCHEMA
Schema
soma_joinid: int64 not null
int: int32 not null
string: string not null
enum: dictionary<values=bool, indices=int8> not null

DATA
Table
5 rows x 4 columns
$soma_joinid <int64>
$int <int32>
$string <string>
$enum <dictionary<values=string, indices=int8>>
$ 
```

johnkerl: @eddelbuettel here's an example of writing float64 value-type enums from Python:

```
#!/usr/bin/env python

import tiledbsoma as soma
import pyarrow as pa
import pandas as pd
import numpy as np
import os, shutil

uri = "bkmc-arrow-vtypes-py"
if os.path.exists(uri):
    shutil.rmtree(uri)

schema = pa.schema(
    [
        ("soma_joinid", pa.int64()),
        ("int", pa.int32()),
        ("string", pa.large_string()),
        ("enum", pa.dictionary(pa.int8(), pa.float64())),
    ]
)

with soma.DataFrame.create(
    uri,
    schema=schema,
) as soma_dataframe:
    pydict = {
        "soma_joinid": [0, 1, 2, 3, 4],
        "int": [100, 101, 102, 103, 104],
        "string": ["red", "yellow", "green", "red", "green"],
        "enum": pd.Categorical(
            np.asarray([1.2, 3.4, 5.6, 1.2, 5.6]),
            ordered=True,
            categories=np.asarray([1.2, 3.4, 5.6]),
            #dtype=np.float32,
        ),
    }
    tbl = pa.Table.from_pydict(pydict)
    soma_dataframe.write(tbl)
```

When I do that write from Python, and then attempt a readback from Python, I get:

```
>>> with tiledb.open('bkmc-arrow-vtypes-py') as A:
...     print(A[:])
...
OrderedDict([('int', array([100, 101, 102, 103, 104], dtype=int32)), ('string', array(['red', 'yellow', 'green', 'red', 'green'], dtype=object)), ('enum', array([1.2, 3.4, 5.6, 1.2, 5.6])), ('soma_joinid', array([0, 1, 2, 3, 4]))])
>>>
>>>
>>> with tiledb.open('bkmc-arrow-vtypes-py') as A:
...     print(A.df[:])
...
   soma_joinid  int  string enum
0            0  100     red  1.2
1            1  101  yellow  3.4
2            2  102   green  5.6
3            3  103     red  1.2
4            4  104   green  5.6
```

and

```
>>> with tiledbsoma.DataFrame.open('bkmc-arrow-vtypes-py') as sdf:
...     print(sdf.read().concat().to_pandas())
...
   soma_joinid  int  string enum
0            0  100     red  1.2
1            1  101  yellow  3.4
2            2  102   green  5.6
3            3  103     red  1.2
4            4  104   green  5.6
>>>
>>>
```

When I try to read using R, I get

```
> tiledb::tiledb_array('bkmc-arrow-vtypes-py', return_as="tibble")[]
Error in as.character.factor(x) : malformed factor
```

and

```
sdf <- SOMADataFrameOpen('bkmc-arrow-vtypes-py')
> sdf$read()$concat()$enum
ChunkedArray
<dictionary<values=string, indices=int8>>
[

  -- dictionary:
    []
  -- indices:
    [
      0,
      1,
      2,
      0,
      2
    ]
]
```

so at minimum we should do:

* If presented with an enum value-type we cannot write from R, error out informatively;
* If presented with an enum value-type we cannot read from R, error out informatively

cc @pablo-gar @mlin @bkmartinjr for feedback

eddelbuettel: Thanks @johnkerl that is a good example:  because R cannot represent a vector of numerics as the payload of a factor, we effectively get the R level error of 'malformed factor' _when it tries to print_.  That is the same error we get on out 'empty' factors.  R cannot print them. 

Can you try instead of the implicit `print()` leading to `Error in as.character.factor(x) : malformed factor` to wrap a `str()` around i.e.

```r
str(tiledb::tiledb_array('bkmc-arrow-vtypes-py', return_as="tibble")[])
```

That display our zero-element factors (after 'create' and before 'write') correctly and I wonder if it can cope here too.

johnkerl: @eddelbuettel 

```
> sdf <- SOMADataFrameOpen('bkmc-arrow-vtypes-py')
> str(tiledb::tiledb_array('bkmc-arrow-vtypes-py', return_as="tibble")[])
tibble [5 × 4] (S3: tbl_df/tbl/data.frame)
 $ soma_joinid:integer64 [1:5] 0 1 2 3 4
 $ int        : int [1:5] 100 101 102 103 104
 $ string     : chr [1:5] "red" "yellow" "green" "red" ...
 $ enum       : Factor w/ 0 levels: 1 2 3 1 3
 - attr(*, "query_status")= chr "COMPLETE"
```

johnkerl: @eddelbuettel when the enum value-type written by Python is string, we get the following when reading from R:

```
> str(tiledb::tiledb_array('bkmc-arrow-py', return_as="tibble")[])
tibble [5 × 4] (S3: tbl_df/tbl/data.frame)
 $ soma_joinid:integer64 [1:5] 0 1 2 3 4
 $ int        : int [1:5] 100 101 102 103 104
 $ string     : chr [1:5] "red" "yellow" "green" "red" ...
 $ enum       : Factor w/ 3 levels "red","green",..: 1 2 3 1 3
 - attr(*, "query_status")= chr "COMPLETE"
```

and when the enum value-type written by Python is float64, we get the following when reading from R:

```
> str(tiledb::tiledb_array('bkmc-arrow-vtypes-py', return_as="tibble")[])
tibble [5 × 4] (S3: tbl_df/tbl/data.frame)
 $ soma_joinid:integer64 [1:5] 0 1 2 3 4
 $ int        : int [1:5] 100 101 102 103 104
 $ string     : chr [1:5] "red" "yellow" "green" "red" ...
 $ enum       : Factor w/ 0 levels: 1 2 3 1 3
 - attr(*, "query_status")= chr "COMPLETE"
```

johnkerl (approved):

apis/r/R/SOMADataFrame.R
eddelbuettel: Even in straight up `arrow` (the R package) going away from `int` seems to have issues:

```r
> dictionary(int8(), utf8())
DictionaryType
dictionary<values=string, indices=int8>
>
> dictionary(float32(), utf8())
Error: Type error: Dictionary index type should be integer, got float
> 
> dictionary(float64(), utf8())
Error: Type error: Dictionary index type should be integer, got double
> 
```

eddelbuettel: I finally have an existence proof:

```r
> dictionary(int8(), float64())
DictionaryType
dictionary<values=double, indices=int8>
> sch <- arrow::schema(arrow::field("ind", arrow::int8()), arrow::field("fct", arrow::dictionary(int8(), float64())))
> sch
Schema
ind: int8
fct: dictionary<values=double, indices=int8>
> arrow_table(ind=1:3, fct=c(2.1, 3.2, 4.3))
Table
3 rows x 2 columns
$ind <int32>
$fct <double>
> tibble::as_tibble(arrow_table(ind=1:3, fct=c(2.1, 3.2, 4.3)))
# A tibble: 3 × 2
    ind   fct
  <int> <dbl>
1     1   2.1
2     2   3.2
3     3   4.3
> 
```

but note that even here we loose the `factor`-ness.  The second column is now a straight-up `double` with no labels:

```r
> str(tibble::as_tibble(arrow_table(ind=1:3, fct=c(2.1, 3.2, 4.3))))
tibble [3 × 2] (S3: tbl_df/tbl/data.frame)
 $ ind: int [1:3] 1 2 3
 $ fct: num [1:3] 2.1 3.2 4.3
> 
```

johnkerl: Thanks @eddelbuettel !

PR #1838, opened 2023-10-30
[c++/python] Depend on core 2.17.3 and TileDB-Py 0.23.2

johnkerl:
TileDB-R 0.21.2 in `apis/r/DESCRIPTION` to follow once that becomes available


eddelbuettel (approved):
:shipit:

Issue #1607, opened 2023-08-17
Deploy profiler in the github


beroy:
In order for the profiler to be really effective, it needs to be embed in our Github flow to trigger the right performance tests for each major release or patch.

beroy:
Few candidates for next steps:

Platform:

- Github actions —> not the best choice
- GH runner with custom runner
- Do we even do it in CZI under tileDB SOMA or TileDB
-      If we can’t we can provide documentation on how exactly to do documentation 
- EC2 instance 
- SLOs —> where we were last time

Benchmarks:

- Experiment Axis Query test ON Reference data not on CellCensus
- Benchmark generate data and write it and then another test read it

PR #1840, opened 2023-10-31
Perf checker

beroy:
Early version of the profiler.

**Changes:** https://github.com/single-cell-data/TileDB-SOMA/issues/1607

**Notes for Reviewer:**

### DIFF
--- a/.github/workflows/profiler.yml
+++ b/.github/workflows/profiler.yml
@@ -1,14 +1,14 @@
 name: TileDB-SOMA Profiler run
 
 on:
+  pull_request:
+
   push:
     branches:
-  pull_request:
-    branches:
-      - "main"
+      - main
+      - 'release-*'
+  workflow_dispatch:
 
-paths:
-      - ".github/workflows/profiler"
 jobs:
 
   check-links:
--- a/.github/workflows/profiler.yml
+++ b/.github/workflows/profiler.yml
@@ -1,0 +1,24 @@
+name: TileDB-SOMA Profiler run
+
+on:
+  push:
+    branches:
+  pull_request:
+    branches:
+      - "main"
+
+paths:
+      - ".github/workflows/profiler"
+jobs:
+
+  check-links:
+
+  name: Setup profiler
+
+  run: |
+    python -m venv profiler_env
+    source profiler_env/bin/activate
+    pip install -e ./profiler
+
+  run: |
+    ./perf_checker.sh
--- a/profiler/ann_data.py
+++ b/profiler/ann_data.py
@@ -4,13 +4,12 @@ import cellxgene_census
 
 import tiledbsoma as soma
 
-census_S3_latest = dict(census_version="latest")
-census_local_copy = dict(uri="/Users/brobatmili/projects/census_data/")
+census_S3_latest = dict(census_version="2023-10-23")
 
 
 def main():
     t1 = perf_counter()
-    with cellxgene_census.open_soma(**census_local_copy) as census:
+    with cellxgene_census.open_soma(**census_S3_latest) as census:
         with census["census_data"]["homo_sapiens"].axis_query(
             measurement_name="RNA",
             obs_query=soma.AxisQuery(value_filter="""tissue_general == 'eye'"""),
--- a/profiler/ann_data.py
+++ b/profiler/ann_data.py
@@ -1,0 +1,23 @@
+from time import perf_counter
+
+import cellxgene_census
+
+import tiledbsoma as soma
+
+census_S3_latest = dict(census_version="latest")
+census_local_copy = dict(uri="/Users/brobatmili/projects/census_data/")
+
+
+def main():
+    t1 = perf_counter()
+    with cellxgene_census.open_soma(**census_local_copy) as census:
+        with census["census_data"]["homo_sapiens"].axis_query(
+            measurement_name="RNA",
+            obs_query=soma.AxisQuery(value_filter="""tissue_general == 'eye'"""),
+        ) as query:
+            query.to_anndata(X_name="raw")
+    t2 = perf_counter()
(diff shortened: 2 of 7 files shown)


profiler/ann_data.py
```diff
@@ -0,0 +1,23 @@
+from time import perf_counter
+
+import cellxgene_census
+
+import tiledbsoma as soma
+
+census_S3_latest = dict(census_version="latest")
+census_local_copy = dict(uri="/Users/brobatmili/projects/census_data/")
```
atolopko-czi, line 8: rm user path

.github/workflows/profiler.yml
```diff
@@ -0,0 +1,24 @@
...
+
+on:
+  push:
+    branches:
+  pull_request:
+    branches:
+      - "main"
+
+paths:
+      - ".github/workflows/profiler"
```
atolopko-czi, line 11: was this intended to be indented under `on`?

profiler/ann_data.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def main():
+    t1 = perf_counter()
+    with cellxgene_census.open_soma(**census_local_copy) as census:
```
atolopko-czi, line 13: I don't think we should use Census API here. We can do the equivalent with just TileDBSOMA API.

profiler/setup.py
```diff
@@ -4,5 +4,5 @@
     name="soma-profiler",
     version="1.0",
     packages=find_packages(),
-    requires=["gitpython", "psutil"],
+    requires=["gitpython", "comacore", "psutil", "tiledbsoma", "cellxgene_census"],
```
atolopko-czi, line 7: `somacore`. Though that shouldn't be needed since `tiledbsoma` is required (transitive dependency)

profiler/top_profiler.py
```diff
@@ -0,0 +1,25 @@
+import data
+
+# Processes the set of previously written logs
+
+threshold = 1.10  # Percent difference
+
+db = data.FileBasedProfileDB()
+actual_max_ts = 0
+dt = db.find("python ann_data.py")
```
atolopko-czi, line 9: Would parameterize the profile run name(s) to run this on.

profiler/top_profiler.py
atolopko-czi, line 1: Would rename to `profile_report.py`, since `top`-like behavior implies that is monitoring the profiler while it runs.

profiler/ann_data.py
atolopko-czi: since query times can change with census release versions, all profiling comparisons need to be done against a single census data release.  So the data release should be fixed. It should be updated on a reasonable schedule, such as when a new LTS data release is made available.

profiler/top_profiler.py
```diff
@@ -0,0 +1,25 @@
...
+last_two = dt[-2:]
+c = 0
+
+for s in dt:
```
atolopko-czi, line 13:
```suggestion
for s in last_two:
```
I _think_ this is the intention...

profiler/top_profiler.py
```diff
@@ -0,0 +1,25 @@
...
+    new_db = sorted(dt, key=lambda ProfileData: ProfileData.timestamp)
+
+    L = [1, 2]
+    L[0] = dt[0].user_time_sec
+    L[1] = dt[1].user_time_sec
+    for i in range(0, len(dt)):
+        print(f"{i} dt[{i}].user_time_sec = {dt[i].user_time_sec} ts {dt[i].timestamp}")
+    print(f"L0 = {L[0]} L1 {L[1]}")
```
atolopko-czi, line 21:
```suggestion
    print(f"L0 = {L[0]}, L1 = {L[1]}")
```

profiler/top_profiler.py
```diff
@@ -0,0 +1,25 @@
...
+dt = db.find("python ann_data.py")
+last_two = dt[-2:]
+c = 0
+
+for s in dt:
+    new_db = sorted(dt, key=lambda ProfileData: ProfileData.timestamp)
+
+    L = [1, 2]
+    L[0] = dt[0].user_time_sec
+    L[1] = dt[1].user_time_sec
```
atolopko-czi, line 18: maybe elapsed time? or both.

profiler/top_profiler.py
```diff
@@ -0,0 +1,25 @@
...
+    for i in range(0, len(dt)):
+        print(f"{i} dt[{i}].user_time_sec = {dt[i].user_time_sec} ts {dt[i].timestamp}")
+    print(f"L0 = {L[0]} L1 {L[1]}")
+
+    if threshold * float(L[1]) < float(L[0]) or float(L[1]) > threshold * float(L[0]):
+        raise SystemExit(f"Potential performance degradation detected {L[0]} va {L[1]}")
+    print("No recent performance degradation detected")
```
atolopko-czi, line 25: report should also output the previous and current tiledbsoma versions

atolopko-czi (changes requested):
A few lower level comments. But per recent meeting, consensus is that we should avoid running this on GHA at all, unless we can do so on an instance with consistent specs.  Also, the more expensive the operations that are being profiled, the more informative the results will be.  So we might consider [GHA large runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners#machine-sizes-for-larger-runners), if we can get work the paid-plan details between the CZI and TileDB organizations.

beroy:
I think we may still have the spec issue with GHA large instances. Maybe infra can help by providing us with a fixed instance.

beroy:
Closing the profiler harness will go to the census repo

johnkerl:
Noting the above

> Closing the profiler harness will go to the census repo

-- closing this PR

PR #1843, opened 2023-10-31
[r] Write string attrs as UTF-8 (Python compatibility)

johnkerl:
**Issue and/or context:** Adapt #99 to R as well as Python

**Changes:**

**Notes for Reviewer:**


apis/r/R/utils-arrow.R
```diff
@@ -88,6 +89,10 @@ tiledb_type_from_arrow_type <- function(x) {
     dictionary = "INT32", 			# for a dictionary the 'values' are ints, levels are character
```
eddelbuettel, line 89: This seems to miss the PR I just made / which is pending ie #1835.  Maybe get that one out of the way first and / or rebase against it ?

johnkerl: @eddelbuettel they're separate PRs, each fixing what it fixes -- I am OK merging them both separately

aaronwolen (approved):
Thanks!

PR #1848, opened 2023-11-02
[c++/python] Expanded enumeration support in `ArrowAdapter::to_arrow`

nguyenv:
**Issue and/or context:**

Updates in Pandas 2.0 has introduced changes to `DictionaryArray.from_arrays` that required refactoring in `pytiledbsoma.cc`. This bug motivated a change in the C++ `ArrowAdapter::to_arrow` method to further handle all supported enumerated types.

**Changes:**

- Several pre-existing C++ and Pybind11 modifications have pulled from https://github.com/single-cell-data/TileDB-SOMA/pull/1793
- Previously in the Python API the `ArrowTable` was converted from `Table.from_arrays` and then enumerations were added in an additional step with `DictionaryArray.from_arrays`. `DictionaryArray.from_arrays` has now been completely removed and now only uses `Table.from_arrays`
- The required changes for the above to work required refactoring of the C++ `ArrowAdapter::to_arrow` method. Previous work has been done to support enumerated string values. The changes in this PR have expanded this now handle all supported enumerated types including all integral numeric types, floating point numbers, and Boolean
- `to_arrow_format` now takes a `use_large` Boolean argument. For enumerations, we need to use `string` or `binary` rather than `large_string` or `large_binary`

**Notes for Reviewer:**

- Pre-existing changes to the `ColumnBuffer` class to handle string enumerations, and the steps used in `ArrowAdapter::to_arrow` to populate the data and offet buffers remain untouched. However, I have written a TODO note on how this can be cleaned up in a future refactor
- Boolean enumerations require special handling from other numeric types as `std::vector<bool>` is a specialized vector that does not store values contigously in memory. We must iterate through the container and then use bit masking to set each bit in least significant bit ordering
- Awaiting TileDB-Py 0.23.4 release

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -429,6 +429,11 @@ class DataFrame(TileDBArray, somacore.DataFrame):
                     # only extend if there are new values
                     if update_vals:
                         se = tiledb.ArraySchemaEvolution(self.context.tiledb_ctx)
+                        if not (
+                            np.issubdtype(enmr.dtype.type, np.str_)
+                            or np.issubdtype(enmr.dtype.type, np.bytes_)
+                        ):
+                            update_vals = np.array(update_vals, enmr.dtype)
                         new_enmr = enmr.extend(update_vals)
                         se.extend_enumeration(new_enmr)
                         se.array_evolve(uri=self.uri)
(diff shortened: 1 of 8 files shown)


nguyenv:
The unit test fail is corrected by https://github.com/TileDB-Inc/TileDB-Py/pull/1853 which requires a TileDB-Py 0.23.4 release.

johnkerl (approved):

eddelbuettel (approved):
(Focusing on the libtiledbsoma/src changes)  Looks good.  I wonder if, now that value is defaulting to `true` wheter we can remove the `use_enum` toggle ?

nguyenv:
Yeah you're right; we can get rid of that now.

eddelbuettel:
@nguyenv Cool. Do you want to fold that edit into the PR?

nguyenv:
Yup just made the change.

Issue #1733, opened 2023-09-28
[python] Revisit `numba` 0.58 dependency


johnkerl:
This was updated in #1723 (and `release-1.5` branch #1726).

Meanwhile I ran a condability check -- which in hindsight I should have run _before_ accepting #1723 -- at https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/48.

This failed as noted by @jdblischak at https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/48#issuecomment-1739947985 since numba 0.58.0 is not yet available on conda-forge

numba 0.58.0 isn't available on conda-forge yet

* https://anaconda.org/conda-forge/numba/files
* https://github.com/conda-forge/numba-feedstock/pull/126 open, blocked by
* https://github.com/conda-forge/llvmlite-feedstock/pull/74, which has all-red CI

We may need to revert our dependency back to numba 0.57 until numba 0.58.0 is available at conda-forge.

cc @thetorpedodog @mlin @jdblischak

jdblischak:
Numba 0.58.1 was merged this morning for conda-forge (https://github.com/conda-forge/numba-feedstock/pull/128, https://github.com/conda-forge/numba-feedstock/commit/172c489224b795438fdbf289d04cbc188ddb480a)

johnkerl:
https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/48 resolves this issue -- the moment we drop support for Python 3.7 (#1790), which is imminent. Leaving this open until then.

bkmartinjr:
@johnkerl - what (if any) is the action to be taken? Or is this resolved?  My understanding is that with the new `>=0.58` pin, both conda and pypi backed installs should be fine, as 0.58 is available (in both locations). Closing, but please re-open if I misunderstand the issue.

johnkerl:
Sorry @bkmartinjr I forgot to signal my intent -- yes, #2122 resolves this, and I wanted to honor/acknowledge your contribution on that front. Thank you!

PR #1852, opened 2023-11-03
[c++] Remove `spdlog` requirement

nguyenv:
**Issue and/or context:**

`libtiledbsoma` still erroneously required `spdlog` as a dependency due to exposing `logger.h` in the public API headers. This was reported in a channel internal to TileDB [here](https://tiledb.slack.com/archives/C052CFC4D2Q/p1698953405794929). 

**Changes:**

- Public headers using `logger.h` have been replaced by `logger_public.h`
- The `fmt` library is a part of `spdlog` which means it cannot be used or accessed except internally
- When possible, code that was using `fmt` in the public headers have now been moved into the `cc` source code which may continue to use the internal `logger.h`
- For `fmt` usage in templated functions, unit tests, and `pytiledbsoma.cc`, string formatting has been refactored to use either `std::ostringstream` or basic `std::string` appending with `+` for compatibility with C++17. In C++20, we can switch to using `std::format`. TODO comments have been left in these areas

**Notes for Reviewer:**

- It is not necessary to include these changes in the 1.5.0 release
- These code changes have been tested and confirmed fixed by @jdblischak

### DIFF
--- a/libtiledbsoma/src/cli/cli.cc
+++ b/libtiledbsoma/src/cli/cli.cc
@@ -33,6 +33,8 @@
 #include "soma/enums.h"
 #include "soma/soma_array.h"
 #include "utils/arrow_adapter.h"
+#include "utils/carrow.h"
+#include "utils/logger.h"
 
 using namespace tiledbsoma;
 
--- a/libtiledbsoma/src/soma/column_buffer.h
+++ b/libtiledbsoma/src/soma/column_buffer.h
@@ -39,7 +39,6 @@
 #include <tiledb/tiledb_experimental>
 
 #include "../utils/common.h"
-#include "../utils/logger.h"
 #include "span/span.hpp"
 
 namespace tiledbsoma {
@@ -128,9 +127,7 @@ class ColumnBuffer {
     ColumnBuffer(const ColumnBuffer&) = delete;
     ColumnBuffer(ColumnBuffer&&) = default;
 
-    ~ColumnBuffer() {
-        LOG_TRACE(fmt::format("[ColumnBuffer] release '{}'", name_));
-    }
+    ~ColumnBuffer();
 
     /**
      * @brief Attach this ColumnBuffer to a TileDB query.
--- a/libtiledbsoma/src/soma/soma_array.h
+++ b/libtiledbsoma/src/soma/soma_array.h
@@ -40,6 +40,7 @@
 #include <tiledb/tiledb>
 #include <tiledb/tiledb_experimental>
 #include "enums.h"
+#include "logger_public.h"
 #include "managed_query.h"
 
 namespace tiledbsoma {
@@ -251,12 +252,13 @@ class SOMAArray {
         int partition_count) {
         // Validate partition inputs
         if (partition_index >= partition_count) {
-            throw TileDBSOMAError(fmt::format(
-                "[SOMAArray] partition_index ({}) must be < "
-                "partition_count "
-                "({})",
-                partition_index,
-                partition_count));
+            // TODO this use to be formatted with fmt::format which is part of
+            // internal header spd/log/fmt/fmt.h and should not be used.
+            // In C++20, this can be replaced with std::format.
+            std::ostringstream err;
+            err << "[SOMAArray] partition_index (" << partition_index
+                << ") must be < partition_count (" << partition_count;
+            throw TileDBSOMAError(err.str());
         }
 
         if (partition_count > 1) {
@@ -268,19 +270,17 @@ class SOMAArray {
                 partition_size = points.size() - start;
             }
 
-            LOG_DEBUG(fmt::format(
-                "[SOMAArray] set_dim_points partitioning: sizeof(T)={} "
-                "dim={} "
-                "index={} "
-                "count={} "
-                "range=[{}, {}] of {} points",
-                sizeof(T),
-                dim,
-                partition_index,
-                partition_count,
-                start,
-                start + partition_size - 1,
-                points.size()));
+            // TODO this use to be formatted with fmt::format which is part of
+            // internal header spd/log/fmt/fmt.h and should not be used.
+            // In C++20, this can be replaced with std::format.
+            std::ostringstream log_dbg;
+            log_dbg << "[SOMAArray] set_dim_points partitioning:"
+                    << " sizeof(T)=" << sizeof(T) << " dim=" << dim
+                    << " index=" << partition_index
+                    << " count=" << partition_count << " range =[" << start
+                    << ", " << start + partition_size - 1 << "] of "
+                    << points.size() << "points";
+            LOG_DEBUG(log_dbg.str());
 
             mq_->select_points(
                 dim, tcb::span<T>{&points[start], partition_size});
@@ -301,7 +301,8 @@ class SOMAArray {
     template <typename T>
     void set_dim_points(const std::string& dim, const std::vector<T>& points) {
         LOG_DEBUG(
-            fmt::format("[SOMAArray] set_dim_points: sizeof(T)={}", sizeof(T)));
+            "[SOMAArray] set_dim_points: sizeof(T)=" +
+            std::to_string(sizeof(T)));
         mq_->select_points(dim, points);
     }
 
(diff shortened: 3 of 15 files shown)


johnkerl (approved):

johnkerl:
> It is not necessary to include these changes in the 1.5.0 release

@nguyenv I've been keeping `release-1.5` in sync with `main` (modulo https://github.com/single-cell-data/TileDB-SOMA/issues/1733) and I want to keep it that way until we get 1.5.0. I don't want to be omitting just a PR or two and then get confused.

libtiledbsoma/src/cli/cli.cc
```diff
@@ -33,6 +33,8 @@
 #include "soma/enums.h"
 #include "soma/soma_array.h"
 #include "utils/arrow_adapter.h"
+#include "utils/carrow.h"
+#include "utils/logger.h"
```
teo-tsirpanis, line 37: Why were these headers added?

libtiledbsoma/src/soma/soma_array.h
teo-tsirpanis, line 1: Can we move the generic method definitions to the `.cc` file and add explicit instantiations?

teo-tsirpanis (approved):

nguyenv:
@eddelbuettel, if you don't mind, could you take a look at the R failures and push the necessary changes to this branch? Today I'd like to focus on getting the necessary Python stuff working for 1.5.0.

libtiledbsoma/src/cli/cli.cc
nguyenv: `logger.h` is used for the `LOG_*` macros and `fmt`. `carrow.h` is used for `ArrowSchema`. `soma_array.h` use to include `logger.h` but that has now been removed.

jdblischak (approved):
Thanks!

libtiledbsoma/src/soma/soma_array.h
nguyenv: Are you talking about separating implemention of the templated functions in the `cc`? I considered doing that, but I feel very ambivalent about having a long list of explicit instantiations for multiple functions.

johnkerl:
> > It is not necessary to include these changes in the 1.5.0 release
> 
> @nguyenv I've been keeping `release-1.5` in sync with `main` (modulo https://github.com/single-cell-data/TileDB-SOMA/issues/1733) and I want to keep it that way until we get 1.5.0. I don't want to be omitting just a PR or two and then get confused.

Of course if this doesn't land in `main` in time for 1.5.0 that's fine -- this is not a 1.5.0 blocker.

libtiledbsoma/src/cli/cli.cc
teo-tsirpanis: But `soma_array.h` includes `logger_public.h`. Isn't it enough?

libtiledbsoma/src/soma/soma_array.h
nguyenv: Is this something that we do in core?

libtiledbsoma/src/cli/cli.cc
nguyenv: No, `logger_public.h` does not expose `fmt`

eddelbuettel:
These are compilation errors coming from the header re-arrangement:

```sh
using C++17
g++ -std=gnu++17 -I"/usr/share/R/include" -DNDEBUG -I. -I../inst/include/ -I../inst/tiledb/include -I../inst/tiledbsoma/include   -I'/usr/lib/R/site-library/Rcpp/include' -I'/usr/lib/R/site-library/RcppSpdlog/include' -I'/usr/local/lib/R/site-library/RcppInt64/include'     -fpic  -g -O2 -ffile-prefix-map=/build/r-base-MHXHhT/r-base-4.3.1=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2  -c RcppExports.cpp -o RcppExports.o
In file included from ../inst/tiledbsoma/include/tiledbsoma/utils/arrow_adapter.h:10,
                 from ../inst/tiledbsoma/include/tiledbsoma/tiledbsoma:45,
                 from ../inst/include/tiledbsoma_types.h:21,
                 from RcppExports.cpp:4:
../inst/tiledbsoma/include/tiledbsoma/utils/carrow.h:16:8: error: redefinition of ‘struct ArrowSchema’
   16 | struct ArrowSchema {
      |        ^~~~~~~~~~~
In file included from ../inst/include/tiledbsoma_types.h:18,
                 from RcppExports.cpp:4:
./nanoarrow.h:83:8: note: previous definition of ‘struct ArrowSchema’
   83 | struct ArrowSchema {
      |        ^~~~~~~~~~~
In file included from ../inst/tiledbsoma/include/tiledbsoma/utils/arrow_adapter.h:10,
                 from ../inst/tiledbsoma/include/tiledbsoma/tiledbsoma:45,
                 from ../inst/include/tiledbsoma_types.h:21,
                 from RcppExports.cpp:4:
../inst/tiledbsoma/include/tiledbsoma/utils/carrow.h:32:8: error: redefinition of ‘struct ArrowArray’
   32 | struct ArrowArray {
      |        ^~~~~~~~~~
In file included from ../inst/include/tiledbsoma_types.h:18,
                 from RcppExports.cpp:4:
./nanoarrow.h:99:8: note: previous definition of ‘struct ArrowArray’
   99 | struct ArrowArray {
      |        ^~~~~~~~~~
make: *** [/usr/lib/R/etc/Makeconf:200: RcppExports.o] Error 1
ERROR: compilation failed for package ‘tiledbsoma’
* removing ‘/tmp/RtmpPq5dYB/Rinst[192](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/6746709009/job/18341227889?pr=1852#step:10:193)9110b964d/tiledbsoma’
```

Any chance we can keep changess local?  The idea was less spill not more?

nguyenv:
Oh opps... I think that's happening because I removed these lines https://github.com/single-cell-data/TileDB-SOMA/pull/1852/files#diff-e419a62446641fa84b7b42a6ab37141a624cb49a0784add4f49823cd1abef3d5L7-L9. I didn't know what that macro was for or where it was being set, and my tests were passing without it so just kept the change in🤦🏻‍♀️

eddelbuettel:
@nguyenv Per your last comment making that `#include` conditional again lets me compile the R package.  Running tests now, not expecting anything so will commit these two lines in a moment.

johnkerl (approved):

libtiledbsoma/src/cli/cli.cc
eddelbuettel: FWIW one can (somewhat poorly) drop in `tinyfmt` which is a less powerful than `fmt` but a single-header C++11 library.  The key is, as @nguyenv points out, that the public logger only takes a string (or `const char*`, I forgot) to 'report' and does no formatting whatsoever itself.

libtiledbsoma/src/soma/column_buffer.h
```diff
@@ -128,9 +127,7 @@ class ColumnBuffer {
     ColumnBuffer(const ColumnBuffer&) = delete;
     ColumnBuffer(ColumnBuffer&&) = default;
 
-    ~ColumnBuffer() {
-        LOG_TRACE(fmt::format("[ColumnBuffer] release '{}'", name_));
-    }
+    ~ColumnBuffer();
```
eddelbuettel, line 130: Why?  While `trace` is super-voluminous it can be helpful to see that dtors have been reached.

eddelbuettel (approved):
There are _a lot of changes here_ and my head is a little dizzy but I take your word for it that things will work with `fmt` only in `src/` and not the public interface.  Thanks for cleaning this up.

libtiledbsoma/src/soma/column_buffer.h
nguyenv: https://github.com/single-cell-data/TileDB-SOMA/pull/1852/files#diff-e563ff857d3a241fa155b2c78fdd243bbe474e8bf28a7782d03f56f00d00f6b5R156-R158

It's because of `fmt::format`. I've moved it into `column_buffer.cc`, so we still do a get a log trace.

PR #1871, opened 2023-11-06
[r] No more "Last 13 lines of output:" in CI

johnkerl:
As discussed in Slack

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -14,6 +14,7 @@ on:
 env:
   COVERAGE_FLAGS: "r"
   COVERAGE_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+  _R_CHECK_TESTS_NLINES_: 0
 
 jobs:
   ci:


bkmartinjr (approved):

eddelbuettel:
Likely redundant because of https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/r/tools/r-ci.sh#L42

mojaveazure (approved):

johnkerl:
I am seeing output I didn't see before. This is a good thing.

eddelbuettel:
What's the test command?  Are you going through `r-ci.sh` which should have the value, or via `R CMD ...` directly?

johnkerl:
@eddelbuettel I am seeing more output before in CI logs linked to from these PRs here in GitHub. I am happy with that fact.

eddelbuettel:
@johnkerl I am happy you are happy but you did not answer my question to which my curious mind would still love an answer:  how did you trigger the tests so that setting YML fixed?

(Longer story R also reflects env variables just for checking from a file `~/.R/check.Renviron`, if it exists, so I have been contemplating having the `bootstrap` step during `r-ci.sh` run create it.  For that more info hence my question.)

johnkerl:
@eddelbuettel I don't know the answer to your question.

* Before this PR I was seeing "Last 13 lines of output" when clicking through "Checks" links on PRs in this repo
* After this PR I am seeing more output
* That is all I know, and all I need to know

eddelbuettel:
@johnkerl So you are saying that the test was triggered as before by `cd apis/r && tools/r-ci.sh run_tests`. I asked you because during the ongoing discussions around the CI fail we had brought the topic of triggering the tests differently (e.g. via a call to `R CMD check ...`.    So what this shows then is that `r-ci.sh` was supposed to already supply the setting but didn't.  This gives me more to work with so thanks for all your continued help in finding this.

PR #1874, opened 2023-11-06
[WIP] [r] Remove `expect_no_condition()` calls for `write_soma.SingleCellExperiment()`

mojaveazure:
For some reason, `write_soma.SingleCellExperiment()` seems to be calling a version of `as.data.frame.numeric()` that throws a deprecation warning. I believe this is causing downstream errors where testthat kills the write, but continues through the test suite. This PR is simply to test this theory.

Tracking issue: #1875.

PR #1876, opened 2023-11-06
[r] Suppress deprecation warnings in `write_soma.DataFrame()`

### LINKED ISSUE #1875: [r/ci] Red R CI since Saturday 2023-11-04
As discussed in Slack

mojaveazure:
Bioc 3.18 and R 4.3.2 now throw deprecation warnings when casting from `S4Vectors::DataFrame` to `data.frame`; this PR suppresses deprecation warnings when doing that cast, but allows additional warnings to percolate through

[sc-36559]

replaces #1874
resolves #1875

### DIFF
--- a/apis/r/R/write_bioc.R
+++ b/apis/r/R/write_bioc.R
@@ -21,7 +21,7 @@ write_soma.DataFrame <- function(
     }
   }
   index <- attr(x, which = 'index')
-  x <- as.data.frame(x)
+  x <- suppressWarnings(as.data.frame(x), classes = "deprecatedWarning")
   attr(x, which = 'index') <- index
   return(write_soma(
     x = x,


johnkerl (approved):

apis/r/R/write_bioc.R
```diff
@@ -21,7 +21,7 @@ write_soma.DataFrame <- function(
     }
   }
   index <- attr(x, which = 'index')
-  x <- as.data.frame(x)
+  x <- suppressWarnings(as.data.frame(x), classes = "deprecatedWarning")
```
eddelbuettel, line 24: Naive question:   If the use is flagged as 'deprecated', there is usually an alternative.  Is there here?

mojaveazure: Not one that we can easily access; the deprecation warning is triggered from code called by `S4Vectors::as.data.frame,DataFrame-method` and is inaccessible to us outside of rewriting the conversion method

eddelbuettel: Weird. Might be worth asking on either r-package-devel or the bioc slack.  I'd rather code around a warning with newly-blessed better code than to muffle it.

(Don't get me wrong: _here and now_ the muffler is right tool.  No criticism implied.  Mostly 'just curious'.)

Issue #1862, opened 2023-11-04
1.5.0 release checklist


johnkerl:
Following https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases, and any additional hmm moments

* [x] Post in the common Slack channel that you want to do a release
* [x] Validate the version number and wait for two people to agree
* [x] Considering the scope of changes in the release, a release candidate (e.g. `i.j.k-rc0`) might be required (we already have)
* [x] Pre-merge CI in the TileDB-SOMA repo is green
* [x] Post-merge CI in the TileDB-SOMA repo is green
  * [x] Red R CI since Saturday 2023-11-04 #1875
  * [x] Python CI sometimes (not always) timing out after 6 hours in CI -- when it happens, usually Python 3.11 only -- see also #1866
* [x] Core 2.17.4 Conda is available https://anaconda.org/conda-forge/tiledb/labels
* [x] TileDB-Py 0.23.4 Conda is available https://anaconda.org/conda-forge/tiledb-py/labels
* [x] TileDB-R 0.21.3 Conda is available https://anaconda.org/conda-forge/r-tiledb/labels
* [x] Check all `main` PRs to be included are merged, and have been backported: `git pull` on `main` and `release-1.5`, then `git diff -r release-1.5 -r main`
* [x] Conda pre-check PR has green CI
  * [x] https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/55
  * [x] But since we've gotten some bugfix PRs for weekend events, there are now more merges since that PR 55 -- we'll need another before tagging 1.5.0 (after all other open `blocks-1.5` PRs are resolved) -- https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/56
* [x] Sandbox-evaluate `soma-release-checks-py-prod` cloud notebook without errors (note cloud-REST already has core 2.17.4)
* [x] Sandbox-evaluate `soma-release-checks-r-prod` cloud notebook without errors (note cloud-REST already has core 2.17.4 -- #1867
  * [x] https://github.com/TileDB-Inc/TileDB-R/pull/614 merged
  * [x] `TileDB-R` 0.21.3 tagged and available at [r-universe](https://tiledb-inc.r-universe.dev/tiledb)
  * [x] `TileDB-SOMA` PR to depend on `TileDB-R` 0.21.3 #1882 
  * [x] `TileDB-SOMA` validation #1884
[...]

PR #1877, opened 2023-11-06
[python] Make Python blockwise tests less resource-intensive

bkmartinjr:
Changes:
* increase sparsity of test arrays for blockwise iterator unit tests
* add an explicit soma context to some tests that benefit from a bit more memory in their initial query

Context: #1862


johnkerl (approved):
🚢

Issue #1867, opened 2023-11-05
[r] Fix query conditions on enumerated columns


johnkerl:
Works in Python: https://gist.github.com/johnkerl/3d1279b0e3bd9e599d541774419faa54
Not in R: https://gist.github.com/johnkerl/d22373ff55cca23cd23672dae547b3de

Note `s3://tiledb-johnkerl/s/v/pbmc3k` in the above gists is nothing other than `tiledbsoma.io.from_h5ad` of
https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/python/testdata/pbmc3k.h5ad

Needs:

* Bugfix
* Unit-test coverage

See also #1862

johnkerl:
[sc-34800]

eddelbuettel:
Fixed by https://github.com/TileDB-Inc/TileDB-R/pull/614 as the actual parsing is done in the `tiledb` package.

johnkerl:
@eddelbuettel thanks! Looking forward to closing this PR once this functionality is rolled into TileDB-SOMA. (Remaining PRs mine, after you've got `TileDB-R` tagged & in [Conda](https://github.com/conda-forge/r-tiledb-feedstock).)

PR #1882, opened 2023-11-07
[r] Depend on TileDB-R 0.21.3

johnkerl:
**Issue and/or context:** #1867 #1862

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -34,7 +34,7 @@ Imports:
     Matrix,
     stats,
     bit64,
-    tiledb (>= 0.21.2),
+    tiledb (>= 0.21.3),
     arrow,
     utils,
     fs,


apis/r/DESCRIPTION
```diff
@@ -34,7 +34,7 @@ Imports:
     Matrix,
     stats,
     bit64,
-    tiledb (>= 0.21.2),
+    tiledb (>= 0.21.3),
```
eddelbuettel, line 37: We can also consider leaving it as is. By pointing at the very installation sources we point at, version 0.21.3 will be installed once available (as it is the highest version).  By putting constraints up like this we effectively only risk breaking builds where the constraint is not yet satisfied, i.e. at CRAN it can take days to get macOS builds, and we also have seen Conda builds delayed or failing for various reasons.  

Luckily we currently only depend on r-universe which is quicker.  But I am wary of putting up constraints when we don't really have to.

eddelbuettel (approved):

johnkerl:
@eddelbuettel queries on enumerated columns is broken without this -- which is spec-incompliant. And it needs to work in Conda, passing requirements, before Conda-build and tagging. See also https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases.

Issue #1890, opened 2023-11-08
[Bug] `ResultOrder` validation in `SOMAArray::reset`


nguyenv:
**Describe the bug**

Currently `SOMAArray::reset` can accept junk `ResultOrder` values. This sets the layout to `TILEDB_COL_MAJOR` which can cause performance degredation. 

**Versions (please complete the following information):**
 - TileDB-SOMA version: 1.5.0
 - Language and language version (e.g. Python 3.8, R 4.2.2): C++
 - OS (e.g. MacOS, Ubuntu Linux): -

**Proposed fix**

`SOMAArray::reset` should use switch-case statements to handle all valid `ResultOrder` values and throw an exception in the default case.

nguyenv:
Fixed in #1891.

PR #1891, opened 2023-11-08
`ResultOrder` validation in `SOMAArray::reset`

nguyenv:
**Issue and/or context:**

#1890

**Changes:**

`SOMAArray::reset` uses switch-case statements to handle all valid `ResultOrder` values and throws `std::invalid_argument` in the default case.


libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -196,18 +196,23 @@ void SOMAArray::reset(
...
-        ResultOrder::automatic) {  // default ResultOrder::automatic is set in
-                                   // soma_array.h
-        tiledb_layout_t layout = (result_order == ResultOrder::rowmajor) ?
-                                     TILEDB_ROW_MAJOR :
-                                     TILEDB_COL_MAJOR;
-        mq_->set_layout(layout);
-        result_order_ = result_order;
+    switch (result_order) {
+        case ResultOrder::automatic:
+            // default ResultOrder::automatic is set in soma_array.h
```
bkmartinjr, line 201: Is there any downside to explicitly setting the value here, rather than depending on defaulting?  Also, in the case where a user is re-using the reader, and is changing the result order from "col-major" to "automatic", won't this leave the value unchanged (i.e, be a noop)?

libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -196,18 +196,23 @@ void SOMAArray::reset(
...
+            break;
+        case ResultOrder::rowmajor:
+            mq_->set_layout(TILEDB_ROW_MAJOR);
+            break;
+        case ResultOrder::colmajor:
+            mq_->set_layout(TILEDB_COL_MAJOR);
+            break;
+        default:
+            throw std::invalid_argument(
```
johnkerl, line 210: bonus points if we can show them what that argument value was

libtiledbsoma/test/unit_soma_array.cc
```diff
@@ -458,4 +458,20 @@ TEST_CASE("SOMAArray: Enumeration") {
...
+    auto soma_array = SOMAArray::open(OpenMode::read, ctx, uri);
+    REQUIRE(soma_array->result_order() == ResultOrder::automatic);
+    soma_array->reset({}, "auto", ResultOrder::rowmajor);
+    REQUIRE(soma_array->result_order() == ResultOrder::rowmajor);
+    soma_array->reset({}, "auto", ResultOrder::colmajor);
+    REQUIRE(soma_array->result_order() == ResultOrder::colmajor);
+    REQUIRE_THROWS_AS(
+        soma_array->reset({}, "auto", static_cast<ResultOrder>(3)),
+        std::invalid_argument);
 }
```
johnkerl, line 477: please add a final newline

johnkerl (approved):

bkmartinjr (commented):
I think we need to set the 'automatic" value explicitly rather than relying on constructor defaulting.

nguyenv:
Failure on macOS R CI not related to changes in this PR.

eddelbuettel:
No such luck.  #1894 brings R 4.3.2 to macOS but still ends in red.

bkmartinjr (approved):

PR #1892, opened 2023-11-08
[python] Fix an f-string

johnkerl:
Found while looking at a unit-test failure


apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -414,8 +414,8 @@ def write(
                 if attr.enum_label is not None and col.num_chunks != 0:
                     if not pa.types.is_dictionary(col_info.type):
                         raise ValueError(
-                            "Expected dictionary type for enumerated attribute "
-                            "{name} but saw {col_info.type}"
+                            f"Expected dictionary type for enumerated attribute "
```
thetorpedodog, line 417: no `f` on this line (I thought ruff would catch this?)

thetorpedodog (approved):

nguyenv (approved):

PR #1894, opened 2023-11-08
[r] Determine the R version dynamically in r-ci

eddelbuettel:
**Issue and/or context:**

CI on macOS can currently warns 'package xyz built under R 4.3.2' because the version installed is still hard-wired to 4.3.1 -- and because the [default version of R at GitHub Actions on macOS](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md) is still 4.3.1.

**Changes:**

The script is changed to use a fall-back default of 4.3.2.

**Notes for Reviewer:**

[SC 36713](https://app.shortcut.com/tiledb-inc/story/36713/r-determine-r-version-at-run-time-in-r-ci-for-macos)

### DIFF
--- a/apis/r/tools/r-ci.sh
+++ b/apis/r/tools/r-ci.sh
@@ -17,7 +17,7 @@ set -e
 
 CRAN=${CRAN:-"https://cloud.r-project.org"}
 OS=$(uname -s)
-RVER=${RVER:-"4.3.1"}
+RVER=${RVER:-$(Rscript -e 'cat(format(getRversion()))')}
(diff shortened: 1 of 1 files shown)


eddelbuettel:
This PR was 'a hunch' about a possible CI issue that has since been fixed in #1898 in a different way.  The was caused more by the package than the R version so we can consider closing this.  On the other hand it does no harm, R 4.3.2 is there and can be used.

johnkerl (approved):

PR #1895, opened 2023-11-08
[r] [DONT MERGE] Dummy PR to get testing backtrace

mojaveazure:
R CI is failing because of a warning being thrown in `write_soma.Seurat()`, which is currently wrapped in `expect_no_condition()`; this change simply unwraps that test, then immediately fails with `expect_true(FALSE)` to try to recover the backtrace from the warning in `write_soma.Seurat()`


mojaveazure:
After investiation, the cause of the R CI failures is due to the SeuratObject binary being built against Matrix 1.6-1.1, which had classes that were removed in Matrix 1.6-2; the S4 cache from SeuratObject referred to the classes that no longer existed, thus causing errors/CI failures. One solution is to install SeuratObject from source, as proposed in #1898

eddelbuettel:
(And I also build us an updated `SeuratObject` package from source in r2u.)

PR #1898, opened 2023-11-08
[r] Install SeuratObject from source on macOS, rely on update r2u binary on linux

mojaveazure:
SeuratObject imports S4 classes from Matrix, which results in S4 method/class caches that can become out of date. If Matrix gets updated, but the binary from CRAN is not rebuilt, then the S4 cache can refer to methods/classes that no longer exist (cause of R CI failure on 2023-11-08). This forces SeuratObject to be rebuilt from source to make use of the updated S4 methods/classes in Matrix

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -73,6 +73,7 @@ jobs:
         run: Rscript -e 'update.packages(ask=FALSE)'
 
       - name: Install SeuratObject from source
+        if: ${{ matrix.os != 'macOS-latest' }}
         run: install.packages("SeuratObject", type = "source", repos = "https://cloud.r-project.org")
         shell: Rscript {0}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/r-ci.yml
```diff
@@ -72,6 +72,10 @@ jobs:
       - name: Update Packages
         run: Rscript -e 'update.packages(ask=FALSE)'
 
+      - name: Install SeuratObject from source
+        run: install.packages("SeuratObject", type = "source", repos = "https://cloud.r-project.org")
+        shell: Rscript {0}
```
eddelbuettel, line 77: Please condition this on macOS only by adding a line such as

```
```

which you see in other blocks in the file.  As mentioned over DM on slack, you may also need to take care of the build dependencies for SeuratObject.

r2u should supply us with a suitable update binary I injected earlier.

eddelbuettel: Ah dang so it was my fault copy and pasting the wrong (non-macOS) line.  Doh.

Well we're getting there now with the rebuilt r2u binary...

eddelbuettel (approved):
I think this one is ready to :ship:

PR #1909, opened 2023-11-14
[python] Normalize unit-test exception coverage

johnkerl:
As we run our unit tests on more and more environments, including on-image from `docker run`, we more often run into an old unit-test error. Namely, `apis/python/tests/test_dataframe_index_columns.py::test_types_read_errors` may either throw `RuntimeError` or `tiledb.cc.TileDBError`. Yet we only catch the former. This means, for example, when running unit tests from `docker run` we need to mentally 'subtract out' these spurious fails. (This has already bitten developers, who have become used to it.) What we should be doing is catching both. (Also note: grepping for `RuntimeError` in `apis/python/tests` show we _already_ do this in other unit-test files -- we need to do it in one more.)

### DIFF
--- a/apis/python/tests/test_dataframe_index_columns.py
+++ b/apis/python/tests/test_dataframe_index_columns.py
@@ -1805,77 +1805,77 @@ def test_types_write_errors(
 
 
 @pytest.mark.parametrize(
-    "name,index_column_names,domain,coords,error",
+    "name,index_column_names,domain,coords",
     [
         [
             "int32-pa-array-untyped",
             ["int32"],
             None,
             [pa.array([3202, 3203])],
-            RuntimeError,  # Static type (INT64) does not match expected type (INT32)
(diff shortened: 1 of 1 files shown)


teo-tsirpanis (approved):

nguyenv (approved):
Potentially corrected [here](https://github.com/single-cell-data/TileDB-SOMA/pull/1793/files#diff-994c39cd29f8569a4e0b2a3f91fc308720bd33fcc4cb32827b83110ef1cfa32cR16-R27) and [here](https://github.com/single-cell-data/TileDB-SOMA/pull/1793/files#diff-a217b4fc51d28f6eac008375b394d0df0df79a5683e83b64290c2f7bc4797825R25-R44) in #1793 where we translate `RuntimeError` to `SOMAError`. Approving this PR in the interim before #1793 gets merged.

PR #1919, opened 2023-11-16
[python] Append-mode ingestion with disjoint measurement names

### LINKED ISSUE #1917: [python] Append-mode ingestion with disjoint measurement names
Use-case:

* Ingest H5AD/AnnData to measurement `one`
* Append-ingest to measurement `two`

Based on user feedback.

johnkerl:
Resolves #1917.

More context:

* Existing append-mode unit-test coverage is thorough when the user ingests data to one measurement, then appends to that same measurement
* Unit-test coverage was lacking (and a bug was present) when the append is to a second measurement.

Specifically, the second measurement's `var` already exists in the former case, but does not in the latter case, so this PR is about checking for that.


aaronwolen (approved):

johnkerl:
Thanks @aaronwolen !

PR #1925, opened 2023-11-17
[python] Update pyarrow dependency

johnkerl:
**Issue and/or context:** https://github.com/single-cell-data/TileDB-SOMA/issues/1926

This is for the `pyarrow` CVE. @bkmartinjr has verified we don't use the vulnerable code path, but, it's good optics for us to update.

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -280,7 +280,7 @@ setuptools.setup(
         "numba==0.56.4; python_version<'3.8'",
         "numpy>=1.18,<1.24; python_version<'3.8'",
         "pandas",
-        "pyarrow>=9.0.0",
+        "pyarrow>=9.0.0,!=14.0.0",
         "scanpy>=1.9.2",
         "scipy",
         "somacore==1.0.4",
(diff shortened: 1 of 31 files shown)


thetorpedodog:
Is this from the [pyarrow CVE](https://www.cve.org/CVERecord?id=CVE-2023-47248)? If so, it looks like we would want to eliminate everything <=v14.0.0 rather than specifically 14.0.0. For py3.7 our only option might be [the hotfix](https://pypi.org/project/pyarrow-hotfix/)?

johnkerl:
Commit https://github.com/single-cell-data/TileDB-SOMA/pull/1925/commits/96088595d6b84c9db607191ed79a0575aca5ccb7 implements the plan at https://github.com/single-cell-data/TileDB-SOMA/issues/1926#issuecomment-1821208184

johnkerl:
R/Python interop is failing, and will continue to fail, until we have core 2.18 along all callpaths

bkmartinjr (approved):

johnkerl:
Thanks @bkmartinjr !

PR #1927, opened 2023-11-20
[c++/python] Add `SOMAArray::stats` to get query stats

nguyenv:
**Issue and/or context:**

Requested by @johnkerl. Context: #1932.

**Changes:**

- Addition of `SOMAArray:;stats` which calls `tiledb::Query::stats` to get stats of query run
- Add `SOMAArray.stats` Python binding

### DIFF
--- a/apis/python/src/tiledbsoma/pytiledbsoma.cc
+++ b/apis/python/src/tiledbsoma/pytiledbsoma.cc
@@ -307,6 +307,8 @@ PYBIND11_MODULE(pytiledbsoma, m) {
             "platform_config"_a = py::dict(),
             "timestamp"_a = py::none())
 
+        .def("stats", &SOMAArray::stats)
+
         .def(
             "set_condition", 
             [](SOMAArray& reader, 
--- a/libtiledbsoma/src/soma/managed_query.cc
+++ b/libtiledbsoma/src/soma/managed_query.cc
@@ -67,6 +67,10 @@ void ManagedQuery::reset() {
     query_submitted_ = false;
 }
 
+std::string ManagedQuery::stats() {
+    return query_->stats();
+}
+
 void ManagedQuery::select_columns(
     const std::vector<std::string>& names, bool if_not_empty) {
     // Return if we are selecting all columns (columns_ is empty) and we want to
--- a/libtiledbsoma/src/soma/managed_query.h
+++ b/libtiledbsoma/src/soma/managed_query.h
@@ -76,6 +76,13 @@ class ManagedQuery {
     void reset();
 
     /**
+     * @brief Returns a JSON-formatted string of the stats.
+     *
+     * @return std::string
+     */
+    std::string stats();
+
+    /**
      * @brief Select columns names to query (dim and attr). If the
      * `if_not_empty` parameter is `true`, the column will be selected iff the
      * list of selected columns is empty. This prevents a `select_columns` call
--- a/libtiledbsoma/src/soma/soma_array.cc
+++ b/libtiledbsoma/src/soma/soma_array.cc
@@ -221,6 +221,10 @@ void SOMAArray::reset(
     submitted_ = false;
 }
 
+std::string SOMAArray::stats() {
+    return mq_->stats();
+}
+
 std::optional<std::shared_ptr<ArrayBuffers>> SOMAArray::read_next() {
     // If the query is complete, return `std::nullopt`
     if (mq_->is_complete(true)) {
--- a/libtiledbsoma/src/soma/soma_array.h
+++ b/libtiledbsoma/src/soma/soma_array.h
@@ -223,6 +223,13 @@ class SOMAArray {
         ResultOrder result_order = ResultOrder::automatic);
 
     /**
+     * @brief Returns a JSON-formatted string of the stats.
+     *
+     * @return std::string
+     */
+    std::string stats();
+
+    /**
      * @brief Set the dimension slice using one point
      *
      * @note Partitioning is not supported
--- a/libtiledbsoma/test/test_soma_array.py
+++ b/libtiledbsoma/test/test_soma_array.py
@@ -1,9 +1,10 @@
 #!/usr/bin/env python
 
-import os
 import json
+import os
 
 import pyarrow as pa
+
 import tiledbsoma.pytiledbsoma as clib
 
 VERBOSE = False
@@ -51,18 +52,20 @@ def test_soma_array_obs():
     assert sr.results_complete()
     assert arrow_table.num_rows == 2638
 
+
 def test_soma_array_stats():
     """Get query stats from an obs array."""
-    
+
     name = "obs"
     uri = os.path.join(SOMA_URI, name)
     sr = clib.SOMAArray(uri)
     sr.read_next()
     stats = json.loads(sr.stats())
-    
+
     assert "Context.StorageManager.Query.Reader.dowork.sum" in stats["timers"]
     assert "Context.StorageManager.Query.Reader.dowork.avg" in stats["timers"]
     assert "Context.StorageManager.Query.Reader.loop_num" in stats["counters"]
+
 
 def test_soma_array_var():
     """Read all values from var array into an arrow table."""
--- a/libtiledbsoma/test/test_soma_array.py
+++ b/libtiledbsoma/test/test_soma_array.py
@@ -1,9 +1,9 @@
 #!/usr/bin/env python
 
 import os
+import json
 
 import pyarrow as pa
-
 import tiledbsoma.pytiledbsoma as clib
 
 VERBOSE = False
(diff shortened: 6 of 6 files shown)


johnkerl (changes requested):
Marking as 'request changes' solely to put a 'pause' on this while discussing with @KiterLuc

My concerns -- not just about this PR; more to make sure this is _really_ what the limitations are with core and that we can't do anything about it in TileDB-SOMA:

* My mental model of how stats are supposed to work is
  * (a) `stats_enable`
  * (b) do something
  * (c) do something else
  * (d,e,f) make all sorts of nested whatever
  * (g) `stats_dump`
  * This produces one dump for whatever core accesses were done through all of points b-f

What appears to be the case now is this is broken, in the core API, in three ways:

* The final `stats_dump` does not include `query->stats()` output for various instances of `query` -- some things appear in the final dump, others in the "side data" from a single query
* We have to do `exp.ms["RNA"].X["data"].stats()` separately
* Even then: on this PR as-is, we get the results of _one_ query -- so in the case of `X.read().concat()` that takes three paginations, the stats will only have the results of the last query

eddelbuettel:
At the risk of getting this wrong, there are actually (AFAIK) two API.  The one you describe where we _globally_ enable and collect.  But there is also a second one, just on the query engine. TileDB-R calls the 'getter' if a toggle is on, and this function independently of the global state.  Mini example below the fold.  And I seems to me that the PR hits the same accessor and should therefor work the same way.

<details>

```r
> uri <- "/tmp/tiledb/penguins"
> res <- tiledb_array(uri, return_as="data.table", query_statistics=TRUE)[]    # ask for query stats
> stats <- attr(res, "query_statistics")      # we return JSON string as attribute of the return object
> str( jsonlite::fromJSON(stats) )
List of 2
 $ timers  :List of 36
  ..$ Context.StorageManager.Query.Reader.unfilter_coord_tiles.sum        : num 0.000301
  ..$ Context.StorageManager.Query.Reader.unfilter_coord_tiles.avg        : num 1e-04
  ..$ Context.StorageManager.Query.Reader.unfilter_attr_tiles.sum         : num 0.00264
  ..$ Context.StorageManager.Query.Reader.unfilter_attr_tiles.avg         : num 0.00033
  ..$ Context.StorageManager.Query.Reader.tile_offset_sizes.sum           : num 4.41e-06
  ..$ Context.StorageManager.Query.Reader.tile_offset_sizes.avg           : num 4.41e-06
  ..$ Context.StorageManager.Query.Reader.resize_fixed_results_to_copy.sum: num 1.85e-06
  ..$ Context.StorageManager.Query.Reader.resize_fixed_results_to_copy.avg: num 1.85e-06
  ..$ Context.StorageManager.Query.Reader.read_tiles.sum                  : num 0.000289
  ..$ Context.StorageManager.Query.Reader.read_tiles.avg                  : num 7.23e-05
  ..$ Context.StorageManager.Query.Reader.read_coordinate_tiles.sum       : num 0.000115
  ..$ Context.StorageManager.Query.Reader.read_coordinate_tiles.avg       : num 5.74e-05
  ..$ Context.StorageManager.Query.Reader.read_attribute_tiles.sum        : num 0.000183
  ..$ Context.StorageManager.Query.Reader.read_attribute_tiles.avg        : num 9.17e-05
  ..$ Context.StorageManager.Query.Reader.read_and_unfilter_coords.sum    : num 0.00045
  ..$ Context.StorageManager.Query.Reader.read_and_unfilter_coords.avg    : num 0.00045
  ..$ Context.StorageManager.Query.Reader.read_and_unfilter_attributes.sum: num 0.00284
  ..$ Context.StorageManager.Query.Reader.read_and_unfilter_attributes.avg: num 0.00284
  ..$ Context.StorageManager.Query.Reader.process_tiles.sum               : num 0.00323
  ..$ Context.StorageManager.Query.Reader.process_tiles.avg               : num 0.00323
  ..$ Context.StorageManager.Query.Reader.load_tile_var_sizes.sum         : num 3.21e-06
  ..$ Context.StorageManager.Query.Reader.load_tile_var_sizes.avg         : num 3.21e-06
  ..$ Context.StorageManager.Query.Reader.load_tile_offsets.sum           : num 0.000583
  ..$ Context.StorageManager.Query.Reader.load_tile_offsets.avg           : num 0.000194
  ..$ Context.StorageManager.Query.Reader.load_initial_data.sum           : num 0.000266
  ..$ Context.StorageManager.Query.Reader.load_initial_data.avg           : num 0.000266
  ..$ Context.StorageManager.Query.Reader.dowork.sum                      : num 0.00469
  ..$ Context.StorageManager.Query.Reader.dowork.avg                      : num 0.00469
  ..$ Context.StorageManager.Query.Reader.create_result_tiles.sum         : num 3.55e-06
  ..$ Context.StorageManager.Query.Reader.create_result_tiles.avg         : num 3.55e-06
  ..$ Context.StorageManager.Query.Reader.copy_fixed_data_tiles.sum       : num 0.000338
  ..$ Context.StorageManager.Query.Reader.copy_fixed_data_tiles.avg       : num 3.76e-05
  ..$ Context.StorageManager.Query.Reader.compute_tile_bitmaps.sum        : num 8.19e-05
  ..$ Context.StorageManager.Query.Reader.compute_tile_bitmaps.avg        : num 8.19e-05
  ..$ Context.StorageManager.Query.Reader.apply_query_condition.sum       : num 5.58e-07
  ..$ Context.StorageManager.Query.Reader.apply_query_condition.avg       : num 5.58e-07
 $ counters:List of 12
  ..$ Context.StorageManager.Query.Reader.result_num              : int 344
  ..$ Context.StorageManager.Query.Reader.read_unfiltered_byte_num: int 16856
  ..$ Context.StorageManager.Query.Reader.num_tiles_read          : int 9
  ..$ Context.StorageManager.Query.Reader.loop_num                : int 1
  ..$ Context.StorageManager.Query.Reader.internal_loop_num       : int 1
  ..$ Context.StorageManager.Query.Reader.ignored_tiles           : int 0
  ..$ Context.StorageManager.Query.Reader.dim_num                 : int 1
  ..$ Context.StorageManager.Query.Reader.dim_fixed_num           : int 1
  ..$ Context.StorageManager.Query.Reader.cell_num                : int 4128
  ..$ Context.StorageManager.Query.Reader.attr_num                : int 8
  ..$ Context.StorageManager.Query.Reader.attr_nullable_num       : int 5
  ..$ Context.StorageManager.Query.Reader.attr_fixed_num          : int 8
> 
```
</details>

johnkerl:
Thanks @eddelbuettel 

We're agreed there are two APIs, one on the global state, and the other on the query

I enumerated three concerns with that status quo (a status quo which I don't much like)

My articulated first choice/hope is that we can find a way to simply get _all the stats_ through one interface

My articulated second choice/hope is that we need to communicate to all users of this that there are two paths now, and furthermore, that `.concat()` queries will need to be unrolled with `stats()` calls _inserted within each loop iteration_ in order to get non-misleading results

nguyenv:
I will hold off on making additional changes for now.

johnkerl:
> I will hold off on making additional changes for now.

@nguyenv I think what we talked about in Slack will work, namely:

* Take what you have now
* Wire it up through `read` methods in the `DataFrame`, `SparseNDArray`, and `DenseNDArray` classes' `read` methods as an optional, backward-compatible argument
* This will automatically solve the `.concat` problem: anyone who wants per-query stats will need to access them on each `read()`, which is a bit of a limitation, but one we're stuck with, and the beauty of this is that this _naturally_ forces the caller to get stats on each `read()` iteration -- there's zero risk of a 3-page `.read().concat()` returning only the stats for the 3rd page
* The only open question is how the user will get the stats-string
  * Option 1: they don't -- we unconditionally print it out to `stdout` or `stderr` (hard-coded)
    * Not great, but, simple API
  * Option 2: optarg to `read` takes a stream -- e.g. `read(..., stats_stream=os.stdout, ...)` or `read(..., stats_stream=handle_they_have_opened, ...)`
  * Option 3: optarg to read is `read(..., provided_stats=True)` and the caller will need to do `string = X.get_stats()` after every .read()`
  * Other options I'm missing -- ? 🤔

nguyenv:
I like option 3 because it's most similar to how the global stats work where first you enable stats and then get the stats in another call.

eddelbuettel:
Had the some instinctive reaction yesterday when that came up: one query statistic dump per read attempt during the iterations, resulting in one json string per read attempt.  If the user is iterating over reads, the burden may be in the reader to fetch the individual strings which should work both R and Python.  And just how `concat()` builds up the overall result data structure it should be able to construct a vector (or list) of json strings.

johnkerl:
@nguyenv new info on #1932 ... I think there's a straight-up bug here (I don't know where) and we should not be having to create a separate per-query stats accessor. Details on #1932 but TL;DR query stats _do_ appear in global stats for the same data & same query in TileDB-Py and TileDB-SOMA-R, just not in TileDB-SOMA-Py ...

PR #1928, opened 2023-11-20
[r] Update tests

mojaveazure:
This PR is two-fold: 
1) remove use of SeuratObject code in ecosystem-agnostic workflows
2) update tests for `SOMAOpen()`
3) skip iterated matrix reader tests on CI

The current `write_soma.data.frame()` pipeline uses `SeuratObject::RandomName()`; this shouldn't happen as `write_soma.data.frame()` is supposed to be ecosystem agnostic. This PR re-implements `SeuratObject::RandomName()` as `random_name()` and uses that instead

The tests for `SOMAOpen` currently use `expect_true()` with a long call to `inherits()`; when these fail, the error message omits the call so all we see is
```r
inherits(...) is not TRUE

`actual`:   FALSE
`expected`: TRUE 
```
this PR replaces `expect_true()` with `expect_s3_class()` so that we get error messages about the class failure. In addition, this expands two tests to expect either sparse or dense arrays as persuant to the spec

The sparse matrix iterated reader test is segfaulting on CI, but not locally. Neither @eddelbuettel nor I have been able to repro the segfault despite setting up containers as close to identical to CI as we can. As such, I'm skipping these tests, but allowing them to run locally

### DIFF
--- a/apis/r/R/utils.R
+++ b/apis/r/R/utils.R
@@ -47,7 +47,10 @@ null <- function(...) {
 }
 
 random_name <- function(length = 5L, chars = letters, ...) {
-  stopifnot(is_integerish(length, n = 1L), is.character(chars))
+  stopifnot(
+    "'length' must be a single integer" = is_integerish(length, n = 1L),
+    "'chars' must be character" = is.character(chars)
+  )
   chars <- unique(unlist(strsplit(chars, split = '')))
   return(paste(sample(chars, size = length, ...), collapse = ''))
 }
--- a/apis/r/R/utils.R
+++ b/apis/r/R/utils.R
@@ -46,6 +46,12 @@ null <- function(...) {
   return(NULL)
 }
 
+random_name <- function(length = 5L, chars = letters, ...) {
+  stopifnot(is_integerish(length, n = 1L), is.character(chars))
+  chars <- unique(unlist(strsplit(chars, split = '')))
+  return(paste(sample(chars, size = length, ...), collapse = ''))
+}
+
 #' Pad Names of a Character Vector
 #'
 #' Fill in missing names of a vector using missing values of said vector
--- a/apis/r/R/write_soma.R
+++ b/apis/r/R/write_soma.R
@@ -351,7 +351,6 @@ write_soma.TsparseMatrix <- function(
     prefix = 'tiledbsoma',
     ...
 ) {
-  check_package('SeuratObject', version = .MINIMUM_SEURAT_VERSION())
   stopifnot(
     "'x' must be a data frame" = is.data.frame(x) || inherits(x, 'DataFrame'),
     "'alt' must be a single character value" = is_scalar_character(alt),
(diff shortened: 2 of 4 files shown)


apis/r/R/utils.R
```diff
@@ -46,6 +46,12 @@ null <- function(...) {
   return(NULL)
 }
 
+random_name <- function(length = 5L, chars = letters, ...) {
+  stopifnot(is_integerish(length, n = 1L), is.character(chars))
```
eddelbuettel, line 50: It's nicer with error messages ie

```r
stopifnot("'length' must be integer" = is_integerish(length, n = 1L), 
                  '"chars" must be character" = is.character(chars))
```

or alike but the current form is of course fully functional.

eddelbuettel (changes requested):
There is a current snafu due possibly to me having updated the data package so that some of the tested assertion flipped, in particular this one:

```diff
@@ -18,7 +18,7 @@ test_that("SOMAOpen", {
     expect_true(inherits(SOMAOpen(file.path(uri, "pbmc3k_processed", "ms", "raw")), "SOMAMeasurement"))
     expect_true(inherits(SOMAOpen(file.path(uri, "pbmc3k_processed", "ms", "RNA")), "SOMAMeasurement"))
 
-    expect_true(inherits(SOMAOpen(file.path(uri, "pbmc3k_processed", "ms", "RNA", "obsm", "X_draw_graph_fr")), "SOMADenseNDArray"))
+    expect_true(inherits(SOMAOpen(file.path(uri, "pbmc3k_processed", "ms", "RNA", "obsm", "X_draw_graph_fr")), "SOMASparseNDArray"))
 
     expect_true(inherits(SOMAOpen(file.path(uri, "pbmc3k_processed", "ms", "raw", "X", "data")), "SOMASparseNDArray"))

```
 
Feel free to replace with a better form if you think it improves matters.

apis/r/R/utils.R
mojaveazure: I've left it without messages because this is an internal, non-exported function

eddelbuettel (commented):
That is way broader than what I suggested.  

The one change I suggested to make is made differently.  It no longer asserts a given class.  Rather it gets away with one of two.   Is that better?

apis/r/R/utils.R
eddelbuettel: I disagree.  The code I suggested is also a comment and an explanation.  Explicit beats implicit.

But you and I will never agree on R stylistic matters but as long as the right result comes out we don't need to split hairs.

eddelbuettel:
The reasoning in the (edited) main description is fine.   Happy to approve once it passes.

apis/r/R/utils.R
mojaveazure: Error messages added in 12dda97

eddelbuettel (approved):
Thanks!

PR #1950, opened 2023-12-05
[DONT MERGE] Test segfault fixes

mojaveazure:


eddelbuettel:
You could consider moving the actions you do not want to `.github/workflows/PARKED` or some such and maybe reduce the matrix: no need for coverage and linux if they both fail, and maybe no need for macOS.  I found it speeds cycles up a little.   But yes this remains extra frustrating.  Oh, and maybe add a block with (untested)

      - name: Show Linux Platform
        run: cat /proc/cpuinfo

PR #1943, opened 2023-12-04
[r] tiledbsoma-r 1.6.0

johnkerl:
I won't merge this until #1931 is complete, on the off chance we need to change something that #1931 reveals.

PR #1953, opened 2023-12-05
[c++] Modify `ManagedQuery` to perform async queries

nguyenv:
**Issue and/or context:**

Across both APIs we have observed the `SOMAArray` intermittenly segfault when accessing the array. This occurs due to a race condition where we have multiple read queries and one query closes the array before the other queries have completed. When a query that has not completed tries to access a closed array, it segfaults. The solution is to modify the C++ `ManagedQuery` so that it performs async queries and waits for all query threads to complete before closing the `SOMAArray`.

**Changes:**

- Changes pulled from #1817

### DIFF
--- a/libtiledbsoma/src/soma/managed_query.cc
+++ b/libtiledbsoma/src/soma/managed_query.cc
@@ -166,8 +166,13 @@ std::shared_ptr<ArrayBuffers> ManagedQuery::results() {
         return buffers_;
     }
 
-    LOG_DEBUG(fmt::format("[ManagedQuery] [{}] Waiting for query", name_));
-    query_future_.wait();
+    if (query_future_.valid()) {
+        LOG_DEBUG(fmt::format("[ManagedQuery] [{}] Waiting for query", name_));
+        query_future_.wait();
+    } else {
+        throw TileDBSOMAError(
+            fmt::format("[ManagedQuery] [{}] 'query_future_' invalid", name_));
+    }
 
     auto status = query_->query_status();
 
--- a/libtiledbsoma/src/soma/managed_query.cc
+++ b/libtiledbsoma/src/soma/managed_query.cc
@@ -54,9 +54,16 @@ ManagedQuery::ManagedQuery(
     reset();
 }
 
+void ManagedQuery::close() {
+    if (query_future_.valid()) {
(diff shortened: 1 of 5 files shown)


eddelbuettel (commented):
LGTM but I am not very deep into `std::future<>` so merely commenting for now.  

I am also a little fuzzy about when we actually launch multiple queries at once -- I had the (wrong, clearly) mental model that we launch one, and iterate if its status is incomplete.

johnkerl (changes requested):
This needs to be branched off `main` and then we can backport to `release-1.6`

At present it's branched off @mojaveazure 's feature-dev work which is unrelated to, and should not block, the 1.6.0 release we're trying to get out today

johnkerl:
Meanwhile @nguyenv  I really appreciate the debug, and am verifying it on #1943 -- which I will rebase later once this is merged to `release-1.6` -- so #1943 will go back to being only `apis/r/DESCRIPTION` and `apis/r/NEWS.md`

eddelbuettel:
Good catch @johnkerl -- I only saw 'one commit' and nodded and was asleep as to the branch it came from.

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -145,21 +152,24 @@ void ManagedQuery::submit_write() {
...
 
-    query_->submit();
-
-    // Poll status until query is not INPROGRESS
-    Query::Status status;
-    do {
-        status = query_->query_status();
-    } while (status == Query::Status::INPROGRESS);
+    LOG_DEBUG(fmt::format("[ManagedQuery] [{}] Waiting for query", name_));
+    query_future_.wait();
```
gspowley, line 170: We should check if the `query_future_` is valid like we do in `ManagedQuery::close()`. 

Something like this:
```c++
if (query_future_.valid()) {
    LOG_DEBUG(fmt::format("[ManagedQuery] [{}] Waiting for query", name_));
    query_future_.wait();
} else {
    // throw error
}
```

We expect it to be valid, but it's good to check.

gspowley (approved):
LGTM! One minor suggestion.

johnkerl (approved):

PR #1960, opened 2023-12-06
[python] Fix many `typeguard` warnings

### LINKED ISSUE #1116: [python] Update to typeguard 4
Tracks https://github.com/single-cell-data/TileDB-SOMA/pull/1115#issuecomment-1470297517

### LINKED ISSUE #2238: [python] Typeguard not running as intended
#2239 has a repro of typeguard failing to raise on non-test code with intentionally broken types.

I believe the issue is, in this block:

https://github.com/single-cell-data/TileDB-SOMA/blob/83bfd007c674fadce058729a542488b141d47fd4/apis/python/tests/__init__.py#L4-L7

Line 5 is inadvertently importing ≈all of `tiledbsoma` (it executes `tiledbsoma/__init__.py`, which [imports most things](https://github.com/single-cell-data/TileDB-SOMA/blob/83bfd007c674fadce058729a542488b141d47fd4/apis/python/src/tiledbsoma/__init__.py#L146-L169)) before the typeguard hook is initialized, meaning ≈nothing is being checked by typeguard.

I suspect this has been the case since [that line was added](https://github.com/single-cell-data/TileDB-SOMA/pull/496/files#diff-cadc56a7066928cc53f87f1f0e10a720a281a6636af66da6e056d98f247ad50bR4-R5), in #496 (Nov '22).

### Attempting to restore Typeguard checks
When I comment out the `_query_condition` line, typeguard raises in any `pytest` command that executes `tiledbsoma/__init__.py`, e.g.:

```bash
echo 'import tiledbsoma' > tests/test_typeguard.py
pytest tests/test_typeguard.py
# ERROR tests/test_typeguard.py - TypeError: Subscripted generics cannot be used with class and instance checks
```

<details><summary>Full stack trace</summary>

```
================================================== test session starts ==================================================
platform darwin -- Python 3.11.8, pytest-8.0.2, pluggy-1.4.0
rootdir: /Users/ryan/c/tiledb/tiledb-soma/apis/python
plugins: anyio-4.3.0, typeguard-2.13.3
collected 0 items / 1 error

======================================================== ERRORS =========================================================
_______________________________________ ERROR collecting tests/test_typeguard.py ________________________________________
tests/test_typeguard.py:3: in <module>
    import tiledbsoma
<frozen importlib._bootstrap>:1176: in _find_and_load
    ???
<frozen importlib._bootstrap>:1147: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:690: in _load_unlocked
    ???
/Users/ryan/.pyenv/versions/tiledb-soma-3.11.8/lib/python3.11/site-packages/typeguard/importhook.py:81: in exec_module
    return super().exec_module(module)
src/tiledbsoma/__init__.py:146: in <module>
    from ._collection import Collection
<frozen importlib._bootstrap>:1176: in _find_and_load
    ???
<frozen importlib._bootstrap>:1147: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:690: in _load_unlocked
    ???
/Users/ryan/.pyenv/versions/tiledb-soma-3.11.8/lib/python3.11/site-packages/typeguard/importhook.py:81: in exec_module
    return super().exec_module(module)
src/tiledbsoma/_collection.py:67: in <module>
    class CollectionBase(  # type: ignore[misc]  # __eq__ false positive
src/tiledbsoma/_collection.py:245: in CollectionBase
    @_funcs.forwards_kwargs_to(
/Users/ryan/.pyenv/versions/tiledb-soma-3.11.8/lib/python3.11/site-packages/typeguard/__init__.py:1032: in wrapper
    check_argument_types(memo)
/Users/ryan/.pyenv/versions/tiledb-soma-3.11.8/lib/python3.11/site-packages/typeguard/__init__.py:875: in check_argument_types
    raise TypeError(*exc.args) from None
E   TypeError: Subscripted generics cannot be used with class and instance checks
================================================ short test summary info ================================================
ERROR tests/test_typeguard.py - TypeError: Subscripted generics cannot be used with class and instance checks
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
=================================================== 1 error in 1.75s ====================================================
```
</details>

The culprit seems to be a `@_funcs.forwards_kwargs_to` annotation in `_collection.py`:

https://github.com/single-cell-data/TileDB-SOMA/blob/83bfd007c674fadce058729a542488b141d47fd4/apis/python/src/tiledbsoma/_collection.py#L245-L250

Adding `@typeguard_ignore` to [the `forwards_kwargs_to` declaration](https://github.com/single-cell-data/TileDB-SOMA/blob/83bfd007c674fadce058729a542488b141d47fd4/apis/python/src/tiledbsoma/_funcs.py#L39) didn't help, nor did adding it to the `@_funcs.forwards_kwargs_to` invocation above.

Open to discussion re: selectively ignoring/fixing problem cases, more explicitly enabling typeguard for known-compliant modules, etc.

johnkerl:
Some post-1.6.0 housekeeping.

We've been seeing literally hundreds of `pytest` warnings per CI run, which is frictional to sort through on CI fails.

Regarding the dominant error which is `no type annotations present -- not typechecking {}'.format(function_name(func))` I could not easily find any _where_ location. (Maddening.) But by comment-out bisection on `tiledbsoma.io` I found `_create_and_open_coll` and `_create_and_open_collection` (and `@overload` variants), the _existence_ of which on a typeguarded import results in these warnings.

I found https://github.com/agronholm/typeguard/issues/108 and in particular https://github.com/agronholm/typeguard/issues/108#issuecomment-1379547357 indicating that this is a fault of older typeguard versions.

Meanwhile thanks to yours truly on https://github.com/single-cell-data/TileDB-SOMA/pull/1115 we're currently pinned to `typeguard<3` which deprives us of the fix for the above.

On the other hand, typeguard 4 has a couple syntax changes to accommodate -- which this PR dutifully implements.

----
### @ryan-williams updates ca. Mar '24

- Fixes #1116
- Fixes #2238: restores typeguard to running on most of the `tiledbsoma` module (which was previously being skipped inadvertently)
- Fixes/Ignores many type errors that had crept in due to missing typeguard coverage

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -6,7 +6,7 @@
 """
 Implementation of a SOMA DataFrame
 """
-from typing import Any, Dict, Optional, Sequence, Tuple, Type, Union, cast
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Type, Union, cast
 
 import numpy as np
 import pandas as pd
@@ -29,8 +29,8 @@ from .options._soma_tiledb_context import _validate_soma_tiledb_context
 from .options._tiledb_create_options import TileDBCreateOptions
 
 _UNBATCHED = options.BatchSize()
-# TODO: in Python ≥3.10, this can be `Tuple[Any, Any] | List[Any]`. In Python <3.10 it raises `TypeError: unsupported operand type(s) for |: '_GenericAlias' and '_GenericAlias'`
-Domain = Sequence[Optional[Sequence[Any]]]
+AxisDomain = Union[None, Tuple[Any, Any], List[Any]]
+Domain = Sequence[AxisDomain]
 
 
 class DataFrame(TileDBArray, somacore.DataFrame):
@@ -721,18 +721,7 @@ def _build_tiledb_schema(
             )
 
     dims = []
-    for index_column_name, _slot_domain in zip(index_column_names, domain):
-        # TODO: this can be cleaned up a bit in Python ≥3.10 (see note on ``Domain`` type above)
-        if _slot_domain is not None:
-            if not isinstance(_slot_domain, tuple):
-                if len(_slot_domain) != 2:
-                    raise ValueError(
-                        f"Domain slots must be tuples or lists of length 2; received {len(_slot_domain)}-list {_slot_domain}"
-                    )
-            slot_domain = _slot_domain[0], _slot_domain[1]
-        else:
-            slot_domain = None
-
+    for index_column_name, slot_domain in zip(index_column_names, domain):
         pa_type = schema.field(index_column_name).type
         dtype = _arrow_types.tiledb_type_from_arrow_type(
             pa_type, is_indexed_column=True
@@ -828,7 +817,7 @@ def _build_tiledb_schema(
 
 
 def _fill_out_slot_domain(
-    slot_domain: Optional[Tuple[Any, Any]],
+    slot_domain: AxisDomain,
     index_column_name: str,
     pa_type: pa.DataType,
     dtype: Any,
@@ -864,7 +853,11 @@ def _fill_out_slot_domain(
                 raise ValueError(
                     f"soma_joinid indices cannot be negative; got upper bound {hi}"
                 )
-
+        if len(slot_domain) != 2:
+            raise ValueError(
+                f"domain must be a two-tuple; got {len(slot_domain)} elements"
+            )
+        slot_domain = slot_domain[0], slot_domain[1]
     elif isinstance(dtype, str):
         slot_domain = None, None
     elif np.issubdtype(dtype, NPInteger):
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -496,7 +496,11 @@ class DataFrame(TileDBArray, somacore.DataFrame):
         return self
 
     def _set_reader_coord(
-        self, sr: clib.SOMAArray, dim_idx: int, dim: tiledb.Dim, coord: object
+        self,
+        sr: clib.SOMAArray,
+        dim_idx: int,
+        dim: Union[pa.Field, tiledb.Dim],
+        coord: object,
     ) -> bool:
         if coord is None:
             return True  # No constraint; select all in this dimension
@@ -570,7 +574,7 @@ class DataFrame(TileDBArray, somacore.DataFrame):
         self,
         sr: clib.SOMAArray,
         dim_idx: int,
-        dim: tiledb.Dim,
+        dim: Union[pa.Field, tiledb.Dim],
         coord: object,
     ) -> bool:
         if isinstance(coord, np.ndarray):
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -6,7 +6,7 @@
 """
 Implementation of a SOMA DataFrame
 """
-from typing import Any, Dict, List, Optional, Sequence, Tuple, Type, Union, cast
+from typing import Any, Dict, Optional, Sequence, Tuple, Type, Union, cast
 
 import numpy as np
 import pandas as pd
@@ -29,7 +29,8 @@ from .options._soma_tiledb_context import _validate_soma_tiledb_context
 from .options._tiledb_create_options import TileDBCreateOptions
 
 _UNBATCHED = options.BatchSize()
-Domain = Sequence[Optional[Tuple[Any, Any] | List[Any]]]
+# TODO: in Python ≥3.10, this can be `Tuple[Any, Any] | List[Any]`. In Python <3.10 it raises `TypeError: unsupported operand type(s) for |: '_GenericAlias' and '_GenericAlias'`
+Domain = Sequence[Optional[Sequence[Any]]]
 
 
 class DataFrame(TileDBArray, somacore.DataFrame):
@@ -720,18 +721,22 @@ def _build_tiledb_schema(
             )
 
     dims = []
-    for index_column_name, slot_domain in zip(index_column_names, domain):
+    for index_column_name, _slot_domain in zip(index_column_names, domain):
+        # TODO: this can be cleaned up a bit in Python ≥3.10 (see note on ``Domain`` type above)
+        if _slot_domain is not None:
+            if not isinstance(_slot_domain, tuple):
+                if len(_slot_domain) != 2:
+                    raise ValueError(
+                        f"Domain slots must be tuples or lists of length 2; received {len(_slot_domain)}-list {_slot_domain}"
+                    )
+            slot_domain = _slot_domain[0], _slot_domain[1]
+        else:
+            slot_domain = None
+
         pa_type = schema.field(index_column_name).type
         dtype = _arrow_types.tiledb_type_from_arrow_type(
             pa_type, is_indexed_column=True
         )
-
-        if isinstance(slot_domain, list):
-            if len(slot_domain) != 2:
-                raise ValueError(
-                    f"Domain slots must be tuples or lists of length 2; received {len(slot_domain)}-list {slot_domain}"
-                )
-            slot_domain = slot_domain[0], slot_domain[1]
 
         slot_domain = _fill_out_slot_domain(
             slot_domain, index_column_name, pa_type, dtype
@@ -772,7 +777,8 @@ def _build_tiledb_schema(
         has_enum = pa.types.is_dictionary(pa_attr.type)
 
         if has_enum:
-            enmr_dtype: np.dtype[Any]
+            # TODO; make this `np.dtype[Any]` in Python ≥3.9
+            enmr_dtype: np.dtype  # type: ignore[type-arg]
             vtype = pa_attr.type.value_type
             if pa.types.is_large_string(vtype) or pa.types.is_string(vtype):
                 enmr_dtype = np.dtype("U")
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -401,7 +401,7 @@ class DataFrame(TileDBArray, somacore.DataFrame):
         dim_names_set = self.index_column_names
         n = None
 
-        se: tiledb.schema_evolution.ArraySchemaEvolution = None
+        se: Optional[tiledb.schema_evolution.ArraySchemaEvolution] = None
 
         for col_info in values.schema:
             name = col_info.name
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -6,7 +6,7 @@
 """
 Implementation of a SOMA DataFrame
 """
-from typing import Any, Dict, Optional, Sequence, Tuple, Type, Union, cast
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Type, Union, cast
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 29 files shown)


johnkerl:
Green on my laptop with Python 3.9; red CI for Python 3.7 .....

johnkerl:
@ryan-williams please take over this PR
https://github.com/single-cell-data/TileDB-SOMA/issues/1116#issuecomment-1989370557

ryan-williams:
Re-enabling `typeguard` on most of `tiledbsoma` (see [`tests/__init__.py` changes](https://github.com/single-cell-data/TileDB-SOMA/pull/1960/files#diff-cadc56a7066928cc53f87f1f0e10a720a281a6636af66da6e056d98f247ad50b)) surfaced a ton of type-checking errors.

[Current status](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/8291314674/job/22690816824?pr=1960#step:14:19669):
```
= 172 failed, 1472 passed, 1 skipped, 3 xfailed, 3222 warnings in 258.96s (0:04:18) =
```

It was much higher before! I fixed a bunch, and learned many typeguard/mypy tricks. Cautiously optimistic I can deal with the rest tomorrow.

apis/python/tests/__init__.py
```diff
@@ -1,11 +1,11 @@
...
-from typeguard.importhook import install_import_hook
-
-# avoid typeguard by importing before calling install_import_hook
-from tiledbsoma import _query_condition  # noqa: F401
+from typeguard import install_import_hook
 
 install_import_hook("tiledbsoma")
 
+# TODO: `pytest tests/test_basic_anndata_io.py` segfaults without this here (likely other things as well)
+from tiledbsoma import _query_condition  # noqa: F401, E402
```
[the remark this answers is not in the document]
johnkerl: @ryan-williams please file a follow-up issue and add the hyperlink to it here as a code comment

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -29,6 +29,8 @@
 from .options._tiledb_create_options import TileDBCreateOptions
 
 _UNBATCHED = options.BatchSize()
+# TODO: in Python ≥3.10, this can be `Tuple[Any, Any] | List[Any]`. In Python <3.10 it raises `TypeError: unsupported operand type(s) for |: '_GenericAlias' and '_GenericAlias'`
+Domain = Sequence[Optional[Sequence[Any]]]
```
thetorpedodog, line 33: This just looks like it’s related to the syntax of types, not the type itself: [`|` as a type union operator](https://docs.python.org/3.10/whatsnew/3.10.html#pep-604-new-type-union-operator) was introduced in 3.10. This can be represented as `Union[Tuple[Any, Any], List[Any]]`.

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -715,7 +721,18 @@ def _build_tiledb_schema(
...
+        # TODO: this can be cleaned up a bit in Python ≥3.10 (see note on ``Domain`` type above)
+        if _slot_domain is not None:
+            if not isinstance(_slot_domain, tuple):
+                if len(_slot_domain) != 2:
+                    raise ValueError(
+                        f"Domain slots must be tuples or lists of length 2; received {len(_slot_domain)}-list {_slot_domain}"
+                    )
+            slot_domain = _slot_domain[0], _slot_domain[1]
+        else:
+            slot_domain = None
```
thetorpedodog, line 734: I don’t follow this check. Why is the length only being checked if it’s a tuple? It seems like this could be reduced to:

```python
if slot_domain is not None:
  slot_domain = tuple(slot_domain)
  if len(slot_domain) != 2:
    raise ...
```

which nicely folds all non-tuple sequences down to tuples and ensures that they have the appropriate length.

apis/python/src/tiledbsoma/_index_util.py
```diff
@@ -16,7 +18,15 @@
...
 def tiledbsoma_build_index(
-    keys: np.typing.NDArray[np.int64],
+    keys: Union[  # type: ignore[type-arg]
+        np.typing.NDArray[np.int64],
+        pa.Array,
+        pa.IntegerArray,
+        pd.Series,
+        pd.arrays.IntegerArray,
+        pa.ChunkedArray,
+        list[int],
```
thetorpedodog, line 28: not something we have to do in this change, but now that py3.7 is finally behind us, it might be worth going back and cleaning up all the `List[whatever]`s and turning them into `list[whatever]`

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -148,7 +152,7 @@ def __init__(
     @classmethod
     def _validate_args(
         cls,
-        shape: NTuple,
+        shape: Union[NTuple, list[int]],
```
thetorpedodog, line 155: Is there a reason that this needs to be specifically a `list`? It seems like it could be a `Sequence` like the rest of the analogous arguments.

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -219,15 +223,13 @@ def concat(self) -> _RT:
...
-    _EagerRT = TypeVar("_EagerRT")
-
     def _maybe_eager_iterator(
         self, x: Iterator[_EagerRT], _pool: Optional[ThreadPoolExecutor] = None
     ) -> Iterator[_EagerRT]:
         """Private"""
         return EagerIterator(x, pool=_pool) if self.eager else x
 
-    def _table_reader(self) -> BlockwiseSingleAxisTableIter:
+    def _table_reader(self) -> Generator[BlockwiseTableReadIterResult, None, None]:
```
thetorpedodog, line 232: Since this generator only does the iteration part, [`Iterator[BlockwiseTableReadIterResult]`](https://docs.python.org/3.9/library/typing.html#typing.Generator:~:text=Alternatively%2C%20annotate%20your%20generator%20as%20having%20a%20return%20type%20of%20either%20Iterable%5BYieldType%5D%20or%20Iterator%5BYieldType%5D%3A) might communicate the intent better (and avoid making promises we don’t intend to keep).

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -350,7 +356,7 @@ def _opener(
 
     # Covariant types should normally not be in parameters, but this is for
     # internal use only so it's OK.
```
thetorpedodog, line 358: This comment is no longer needed since the type-ignore is gone.

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -214,6 +216,10 @@ class _FSPathWrapper:
     so here we just proxy all attributes except ``__fspath__``.
     """
 
+    # Needed to make typeguard happy
+    def __call__(self) -> None:
+        pass
```
thetorpedodog, line 221: might be worth making this `raise AssertionError("this should never be called")` since, well, it shouldn't

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -349,7 +358,15 @@ def from_h5ad(
...
-            anndata = ad.read_h5ad(_FSPathWrapper(input_handle, input_path), "r")
+            path: Path = (
+                input_path._path
+                if isinstance(input_path, _FSPathWrapper)
+                else input_path
+            )
+            anndata = ad.read_h5ad(
+                _FSPathWrapper(input_handle, path),
+                "r",
+            )
```
thetorpedodog, line 369: This check seems weird. If we’re being passed an `_FSPathWrapper` here, that means that we’ve been passed a VFS file handle rather than a thing that we should be opening with VFS.

apis/python/src/tiledbsoma/options/_tiledb_create_options.py
```diff
@@ -168,7 +173,7 @@ def from_platform_config(
         """
         create_entry = _dig_platform_config(platform_config, cls, ("tiledb", "create"))
         if isinstance(create_entry, dict):
-            attrs: Tuple["attrs_.Attribute[Any]", ...] = cls.__attrs_attrs__
+            attrs: Tuple[Attribute, ...] = cls.__attrs_attrs__  # type: ignore[type-arg]
```
thetorpedodog, line 176: since we already have `attrs_.Attribute` available here (the same object as `attr.Attribute`, the extra import can be avoided).

apis/python/tests/_util.py
```diff
@@ -0,0 +1,17 @@
+from contextlib import contextmanager
+from typing import Type
+
+import pytest
+from typeguard import suppress_type_checks
+
+
+@contextmanager
+def raises(exc: Type[Exception]):
```
thetorpedodog, line 9: This might be better named `raises_no_typeguard` or something like that, because otherwise it’s not immediately obvious what the difference between `with raises(...)` vs. `with pytest.raises(...)` is.

I would also recommend just forwarding all of `*args` and `**kwargs` to `pytest.raises`, so that the two are otherwise completely compatible.

thetorpedodog:
also, sorry if it didn’t come across in my previous message, but overall this change is a great improvement! my comments are relatively small in the scope of the entire thing.

ryan-williams (commented):
np at all @thetorpedodog, appreciate the thorough review! I believe I addressed everything.

I also resolved a final outstanding `typeguard_ignore` rough edge: typeguard was failing in an inscrutable way due to an interaction between `typing_extensions.TypedDict` and our `._funcs.typeguard_ignore` wrapper. I wrote it up at [ryan-williams/typeguard-issues](https://github.com/ryan-williams/typeguard-issues), along with another issue that resulted in some "ignore"s that maybe someday we can remove. Will try to file both against typeguard tmrw.

Update: filed https://github.com/agronholm/typeguard/issues/442 and https://github.com/agronholm/typeguard/issues/443

apis/python/src/tiledbsoma/io/ingest.py
ryan-williams: I actually removed this altogether, it didn't seem to be needed.

#2291 also includes https://github.com/single-cell-data/TileDB-SOMA/pull/2291/commits/2f00a9b3f7dd81910b43fec856d101e456f44f6d, which makes `_FSPathWrapper` inherit from `pathlib.Path`, which mitigates further type-checking issues surrounding `_FSPathWrapper` (and seems like a good change to make anyway).

ryan-williams (approved):

apis/python/tests/__init__.py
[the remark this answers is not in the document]
ryan-williams: Filed issue #2293, repro PR #2286

johnkerl:
@thetorpedodog can you approve or are these still changes outstanding?

apis/python/src/tiledbsoma/io/outgest.py
```diff
@@ -47,12 +47,13 @@
...
     Matrix,
 )
+from ._util import _FSPathWrapper
 
 
 # ----------------------------------------------------------------
 def to_h5ad(
     experiment: Experiment,
-    h5ad_path: Path,
+    h5ad_path: Union[Path, _FSPathWrapper],
```
thetorpedodog, line 56: one last `_FSPathWrapper` to clean up

thetorpedodog (approved):

apis/python/src/tiledbsoma/options/_tiledb_create_options.py
```diff
@@ -7,6 +7,7 @@
     Sequence,
     Tuple,
     Type,
+    TypedDict,
```
thetorpedodog, line 10: just a note for myself and anybody else who wonders about this change in the future: `TypedDict` was added to `typing` in py3.8, so we can indeed take it out of `typing_extensions`.

thetorpedodog (approved):

ryan-williams:
(R CI failures believed to be unrelated to this change, should be resolved later today…)

PR #1961, opened 2023-12-06
Add back Windows support for the Python API, but without CI.

teo-tsirpanis:
**Issue and/or context:**

**Changes:**

**Notes for Reviewer:**

Reverts #1959 (reinstating #1811), while removing Windows CI.

c.c. @johnkerl

PR #1811, opened 2023-10-19
Add Windows support for the Python API

teo-tsirpanis:
**Issue and/or context:**

**Changes:**

**Notes for Reviewer:**

[SC-35700](https://app.shortcut.com/tiledb-inc/story/35700/implement-windows-support-for-soma)

Fixes #1360.

PR #1962, opened 2023-12-06
[python] Full Windows pre-push CI matrix

johnkerl:
#1811 added Windows support. Pre-push/minimal CI was green, but post-push/full CI was not -- see notes on #1959 -- hence the revert on #1959. Then #1961 restores #1811 to `main` but without Windows CI. This is the sweet spot: now we have the large Windows PR in `main`, with opt-in CI testing here on this PR.  CC @teo-tsirpanis and @nguyenv .


johnkerl:
@teo-tsirpanis feel free to push commits to this PR whenever 🙏

PR #1963, opened 2023-12-06
[python] Fix `from_anndata()` to respect `var_id_name`

aaronwolen:
**Issue and/or context:** Bug report from user attempting to append an AnnData object to an existing SOMA experiment using a custom name for `var_id`.  

**Changes:** The `var_id_name` argument is no longer ignored. Relevant unit tests were updated to validate the set of columns read back from `obs` and `var` exactly match the input.


johnkerl:
[sc-38306]

PR #1967, opened 2023-12-08
[python] QueryCondition should error out if given empty set

nguyenv:
**Issue and/or context:**

This error was reported in an internal channel by @ckrilow.

**Changes:**

- These changes have already been applied to TileDB-Py https://github.com/TileDB-Inc/TileDB-Py/pull/1877 but need to also be ported over to TileDB-SOMA


aaronwolen (approved):
Thanks!

nguyenv:
I am not understanding why codecov is complaining. We do test the line that is it flagging.

Issue #1818, opened 2023-10-23
[Bug] Unable to install R API through conda or through R


Dzhan4:
Hello, I'm trying to install tiledbsoma R API to use for cellxgene census.

I have a custom conda environment with R 4.3.1 and Python 3.12.0 and all other dependencies installed except for tiledbsoma. I have already installed tiledb through conda as well. When I try to install either tiledbsoma or cellxgene.census, I get errors when I try to install through conda or through R. Any help would be greatly appreciated!

Errors below:

conda install error:

```
UnsatisfiableError: The following specifications were found to be incompatible with each other:

Output in format: Requested package -> Available versions

Package libgcc-ng conflicts for:
r-tiledbsoma -> libgcc-ng[version='>=12']
r-tiledbsoma -> r-arrow -> libgcc-ng[version='7.2.0.*|>=11.2.0|>=7.3.0|>=7.2.0']

Package libstdcxx-ng conflicts for:
r-tiledbsoma -> r-arrow -> libstdcxx-ng[version='7.2.0.*|>=11.2.0|>=7.3.0|>=7.2.0']
r-tiledbsoma -> libstdcxx-ng[version='>=12']

Package _libgcc_mutex conflicts for:
r-tiledbsoma -> libgcc-ng[version='>=12'] -> _libgcc_mutex==0.1=conda_forge
python=3.12 -> libgcc-ng[version='>=12'] -> _libgcc_mutex[version='*|0.1',build='main|main|conda_forge']The following specifications were found to be incompatible with your system:

  - feature:/linux-64::__glibc==2.17=0
  - feature:|@/linux-64::__glibc==2.17=0
  - python=3.12 -> libgcc-ng[version='>=11.2.0'] -> __glibc[version='>=2.17']

Your installed version is: 2.17

```

R install error:

```
[...]

PR #1972, opened 2023-12-11
[r] Update `DESCRIPTION` and `NEWS.md` for 1.5.2

johnkerl:
1.5.2 will deliver a Python-only bugfix on the `release-1.5` branch. R is not affected, other than our commitment to keep Python and R API versions synchronized.

Please also see https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -49,6 +49,7 @@ jobs:
         run: |
           cd apis/r
           wget https://github.com/TileDB-Inc/TileDB-R/archive/refs/tags/0.21.3.tar.gz
+          Rscript -e 'remotes::install_deps("0.21.3.tar.gz")'
           R CMD INSTALL 0.21.3.tar.gz    # as 0.21.3 is needed by SOMA 1.5.1
 
       - name: Install r-universe build of SeuratObject (macOS)
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -42,44 +42,30 @@ jobs:
       - name: Install BioConductor package SingleCellExperiment
         run: cd apis/r && tools/r-ci.sh install_bioc SingleCellExperiment
 
-      # Uncomment these next two stanzas as needed whenever we've just released a new tiledb-r for
-      # which source is available but CRAN releases (and hence update r2u binaries) are not yet:
-
-      - name: Install 0.21 build of tiledb-r with core 2.17
+      - name: Install tiledb-r 0.21 and core 2.17
         run: |
           cd apis/r
           wget https://github.com/TileDB-Inc/TileDB-R/archive/refs/tags/0.21.3.tar.gz
           Rscript -e 'remotes::install_deps("0.21.3.tar.gz")'
           R CMD INSTALL 0.21.3.tar.gz    # as 0.21.3 is needed by SOMA 1.5.1
-          Rscript -e 'print("VERSION CHECK")'
-          Rscript -e 'print(packageVersion("tiledb"))'
-          Rscript -e 'print(tiledb::tiledb_version())'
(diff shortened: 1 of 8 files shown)


apis/r/NEWS.md
```diff
@@ -1,4 +1,4 @@
-# tiledbsoma 1.5.1
+# tiledbsoma 1.5.2
```
eddelbuettel, line 1: Maybe add a line here that the update was procedural to keep Py + R aligned?   Better than a gap?

eddelbuettel (commented):
See inline

apis/r/NEWS.md
johnkerl: pushed

johnkerl:
See code comments and also https://github.com/single-cell-data/TileDB-SOMA/issues/1818

apis/r/NEWS.md
```diff
@@ -1,3 +1,10 @@
...
+
+## Changes
+
+1.5.2 delivers a Python-only bugfix on the `release-1.5` branch. R is not affected, other than our
```
aaronwolen, line 5: I like this!

aaronwolen (approved):
Thanks!

johnkerl:
[sc-38306]

.github/workflows/r-ci.yml
```diff
@@ -112,17 +101,22 @@ jobs:
           Rscript -e 'print(packageVersion("tiledb"))'
           Rscript -e 'print(tiledb::tiledb_version())'
 
-      - name: Test
-        if: ${{ matrix.covr == 'no' }}
+      - name: We want tiledb-r 0.21 and core 2.17
         run: |
           cd apis/r
+          Rscript -e 'remove.packages("tiledb")'
```
eddelbuettel, line 107: You don't need that.  It cleanly installs over it.

.github/workflows/r-ci.yml
```diff
@@ -112,17 +101,22 @@ jobs:
...
+          wget https://github.com/TileDB-Inc/TileDB-R/archive/refs/tags/0.21.3.tar.gz
+          Rscript -e 'remotes::install_deps("0.21.3.tar.gz")'
```
eddelbuettel, line 109: That is overly complicate `R CMD INSTALL "0.21.3.tar.gz"` is clearer/

.github/workflows/r-ci.yml
```diff
@@ -112,17 +101,22 @@ jobs:
...
+          R CMD INSTALL 0.21.3.tar.gz    # as 0.21.3 is needed by SOMA 1.5.1
```
eddelbuettel, line 110: That is doing work twice.  You just installed it.

eddelbuettel (commented):
Some comments inline.  Feel free to ignore.

.github/workflows/r-ci.yml
johnkerl: @eddelbuettel I am iterating until I get green CI with 2.17 and 0.22 surviving all the way until the test runs. I'm not there yet. When I finally succeed, I'll get more parsimonious. Thank you!

eddelbuettel: @johnkerl if you install the same package twice you are waiting longer.  Is that really on purpose?

johnkerl: @eddelbuettel again, I'm running experiments, iterating, trying to find out where the downgrade is happening ...

eddelbuettel: @johnkerl I understand I am just explaining that your wait is longer if there is a step you need _once_ but do _twice_.  Above you install the same package twice.  But whatevs.

johnkerl: @eddelbuettel thanks

CI is now green, and I will now prune

Issue #1973, opened 2023-12-12
[python/r] Multiple writes of `Dictionary` (categorical) to a `DataFrame` can result in loss of data


bkmartinjr:
The following code demonstrates a loss of data situation when there are multiple writes to a DataFrame column of type `dictionary` (enum). The essential issue appears to be the addition of a new enum/category value in the second write.

1. The first write works fine
2. The second write correctly extends the DataFrame enum list from `['A','B']` to `['A','B','C']`, but does _not_ write the correct values into the column, resulting in data corruption

The test case is pretty self-explanatory -- the data read back should match the data written, and the program should print out "all good".  However, when I run it, the data read back is incorrect.

Test code

```
import numpy as np
import pandas as pd
import pyarrow as pa
import tiledbsoma as soma

"""
Demonstrate that Enum value extension is not working.

"""


def main():
    fname = "test_dataframe"

    # --- create dataframe

    schema = pa.schema(
        [
            ("soma_joinid", pa.int64()),
            (
                "obs_col_like",
                pa.dictionary(
                    index_type=pa.int8(), value_type=pa.large_string(), ordered=False
                ),
            ),
        ]
    )
    soma.DataFrame.create(fname, schema=schema).close()

    # --- write dataframe, in two separate writes (NOTE: no overlapping joinids)

    pandas_df_1 = pd.DataFrame(
        {
            "soma_joinid": pd.Series([0, 1, 2], dtype=np.int64),
            "obs_col_like": pd.Series(["A", "B", "A"], dtype="category"),
        }
    )
    tbl = pa.Table.from_pandas(pandas_df_1, preserve_index=False)
    with soma.open(fname, mode="w") as A:
        A.write(tbl)

    pandas_df_2 = pd.DataFrame(
        {
            "soma_joinid": pd.Series([3, 4, 5], dtype=np.int64),
            "obs_col_like": pd.Series(["B", "C", "B"], dtype="category"),
        }
    )
    tbl = pa.Table.from_pandas(pandas_df_2, preserve_index=False)
    with soma.open(fname, mode="w") as A:
        A.write(tbl)

    # --- read back and validate
    with soma.open(fname) as A:
        df = A.read().concat().to_pandas()

    expected_df = pd.concat((pandas_df_1, pandas_df_2), ignore_index=True)

    if not df.equals(expected_df):
        print("NOT THE SAME!")
        print(df.compare(expected_df))


if __name__ == "__main__":
    main()
```

Version information:
[...]

Issue #1975, opened 2023-12-12
[r, python] `DataFrame` categorical handling does not handle extension past index type limit


bkmartinjr:
If a SOMA DataFrame is created with a dictionary column (ie., categorical), and in subsequent writes the number of categories exceeds the capacity of the dictionary index_type, the current implementation will silently lose data upon write.

Background:
* dictionary/enum values are stored as an int, in Arrow parlance the `index_type` of the `dictionary`.  E.g., int8, int16, etc.
* the actual dict/enum labels (cateogries) are stored separately.

The following code demonstrates what happens if you create a column with an `int8` index type, and then using multiple writes, attempt to write a 129th enum/category value.   

Expected behavior:  either raise an error, _or_ transparently extend / evolve the index type to a larger integer.  If the latter occurs, the test case below should return "all good". 

Actual behavior:  write completes, but stored data is different than the user provided (ie., silent data loss).  Digging into it a bit, it looks like the TileDB Array's enum list is extended with the 129th value and the wrong values are written into the column.

This may be intertwined with #1973, but a quick skim of the code base didn't show any support for detecting this edge case. 

Versions:
```
tiledbsoma.__version__        1.6.0
TileDB-Py tiledb.version()    (0, 24, 0)
TileDB core version           2.18.2
libtiledbsoma version()       libtiledb=2.18.2
python version                3.10.12.final.0
OS version                    Linux 6.2.0-1017-aws
```

Test code:
```
import numpy as np
import pandas as pd
import pyarrow as pa
import tiledbsoma as soma

"""
See what happens if we extend number of enums past index type limit,
i.e., if index type is int8, what happens when we add the 129th enum?
"""


def main():
    fname = "test_dataframe"

    # --- create dataframe

    schema = pa.schema(
        [
            ("soma_joinid", pa.int64()),
            (
                "obs_col_like",
                pa.dictionary(
                    index_type=pa.int8(), value_type=pa.large_string(), ordered=False
                ),
            ),
        ]
    )
    soma.DataFrame.create(fname, schema=schema).close()

    n_elem = 132
    n_cats = 128
    pandas_df_1 = pd.DataFrame(
        {
            "soma_joinid": pd.Series(np.arange(n_elem), dtype=np.int64),
            "obs_col_like": pd.Series(
                [f"enum_{i % n_cats}" for i in range(n_elem)], dtype="category"
            ),
        }
    )

[...]

PR #1978, opened 2023-12-13
[python] Fix data loss for multiple writes with `Enumeration` extend

nguyenv:
**Issue and/or context:**

#1973

When not all the categorical values are present in the column to be written, the index values are incorrect and results in data loss.

**Changes:**

Update the column with the extended enumerations so that the index values are mapped correctly.

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -422,9 +422,9 @@ class DataFrame(TileDBArray, somacore.DataFrame):
 
                     # get new enumeration values by taking the set difference
                     # while maintaining ordering
-                    update_vals = pd.Series(
-                        np.concatenate([enmr.values(), enmr.values(), col.dictionary])
-                    ).drop_duplicates(keep=False)
+                    update_vals = np.setdiff1d(
+                        col.dictionary, enmr.values(), assume_unique=True
+                    )
 
                     # only extend if there are new values
                     if len(update_vals) != 0:
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -436,6 +436,8 @@ class DataFrame(TileDBArray, somacore.DataFrame):
                         else:
                             extend_vals = np.array(update_vals, enmr.dtype)
                         new_enmr = enmr.extend(extend_vals)
+                        df = pd.Categorical(col.chunk(0).to_pandas(), new_enmr.values())
+                        col = pa.chunked_array(pa.DictionaryArray.from_pandas(df))
                         se.extend_enumeration(new_enmr)
                         se.array_evolve(uri=self.uri)
 
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -420,14 +420,14 @@ class DataFrame(TileDBArray, somacore.DataFrame):
 
                     enmr = self._handle.enum(attr.name)
 
-                    # get new enumeration values, maintain original ordering
-                    update_vals = []
-                    for new_val in col.dictionary.tolist():
-                        if new_val not in enmr.values():
-                            update_vals.append(new_val)
+                    # get new enumeration values by taking the set difference
+                    # while maintaining ordering
+                    update_vals = pd.Series(
+                        np.concatenate([enmr.values(), enmr.values(), col.dictionary])
+                    ).drop_duplicates(keep=False)
 
                     # only extend if there are new values
-                    if update_vals:
+                    if len(update_vals) != 0:
                         se = tiledb.ArraySchemaEvolution(self.context.tiledb_ctx)
                         if np.issubdtype(enmr.dtype.type, np.str_):
                             extend_vals = np.array(update_vals, "U")
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -427,7 +427,7 @@ class DataFrame(TileDBArray, somacore.DataFrame):
                     )
 
                     # only extend if there are new values
-                    if len(update_vals) != 0:
+                    if update_vals:
                         se = tiledb.ArraySchemaEvolution(self.context.tiledb_ctx)
                         if np.issubdtype(enmr.dtype.type, np.str_):
                             extend_vals = np.array(update_vals, "U")
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -404,14 +404,14 @@ class DataFrame(TileDBArray, somacore.DataFrame):
 
         for col_info in values.schema:
             name = col_info.name
-            col = values.column(name)
+            col = values.column(name).combine_chunks()
             n = len(col)
 
             if self._handle.schema.has_attr(name):
                 attr = self._handle.schema.attr(name)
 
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -436,6 +436,8 @@ def write(
                         else:
                             extend_vals = np.array(update_vals, enmr.dtype)
                         new_enmr = enmr.extend(extend_vals)
+                        df = pd.Categorical(col.chunk(0).to_pandas(), new_enmr.values())
```
bkmartinjr, line 439: This new code points out another potential bug -- what happens if the column's ChunkedArray has more than one chunk, where each chunk has a different set dictionary (enum) values?

Here is a real example I pulled from a multi-chunk column:
```
[<pyarrow.lib.ChunkedArray object at 0x7fd593a13ec0>
 [
 
   -- dictionary:
     [
       "A",
       "B"
     ]
   -- indices:
     [
       0,
       1,
       0
     ],
 
   -- dictionary:
     [
       "B",
       "C"
     ]
   -- indices:
     [
       0,
       1,
       0
     ]
 ]]```

As I read this code, it will only handle the information in chunk `0`, which is definitely incorrect....

Comment (bkmartinjr) [reply apis/python/src/tiledbsoma/_dataframe.py:439]: Actually I see more examples of the same bug in subsequent code (e.g., anywhere `col.chunk(0)` is hard-wired.

**Review by bkmartinjr — CHANGES_REQUESTED** [review]
see inline comments regarding bug in multi-chunk columns

File: apis/python/src/tiledbsoma/_dataframe.py [hunk 2]
```diff
@@ -422,7 +422,7 @@ def write(
 
                     # get new enumeration values, maintain original ordering
                     update_vals = []
-                    for new_val in col.chunk(0).dictionary.tolist():
+                    for new_val in col.dictionary.tolist():
```
bkmartinjr, line 425: A suggestion:   this loop is going to be slow where the enumeration gets large.  We commonly see enum's with thousands of values. For example, in a recent Census dataset, the number of unique values in our "string label" columns ranges from 2 to 5576, with an average of about 450.

Another way to implement would be a hash table or set, e.g., `pandas.unique`.   I benchmarked a couple of examples with a few thousand enums, and using `unique` is roughly 40X faster on a trivial case, with a variety of enum lengths:

Two implementations:  `unique1` is essentially the loop above.  `unique2` does the same thing using concat/unique.

```
In [26]: def unique1(existing_cats, new_cats):
    ...:     update_vals = []
    ...:     for new_val in new_cats:
    ...:         if new_val not in existing_cats:
    ...:             update_vals.append(new_val)
    ...:     return existing_cats + update_vals
    ...: 
    ...: def unique2(existing_cats, new_cats):
    ...:     return pd.unique(np.concatenate((existing_cats, new_cats)))
    ...: 

In [27]: %timeit unique1( np.arange(1000).tolist(), np.arange(2000,0,-1).tolist())
7.16 ms ± 69 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [28]: %timeit unique2( np.arange(1000).tolist(), np.arange(2000,0,-1).tolist())
175 µs ± 306 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [29]: %timeit unique1( np.arange(400).tolist(), np.arange(500,400,-1).tolist())
204 µs ± 6.1 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [30]: %timeit unique2( np.arange(400).tolist(), np.arange(500,400,-1).tolist())
42.9 µs ± 955 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [31]: %timeit unique1( np.arange(5000).tolist(), np.arange(5500,4000,-1).tolist())
32.8 ms ± 1.35 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [32]: %timeit unique2( np.arange(5000).tolist(), np.arange(5500,4000,-1).tolist())
396 µs ± 12.9 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [33]: %timeit unique1( np.arange(10).tolist(), np.arange(11,20).tolist())
2.46 µs ± 117 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [34]: %timeit unique2( np.arange(10).tolist(), np.arange(11,20).tolist())
14.9 µs ± 64.2 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

```

So my thought is that using `unique` could improve the worst case by a lot, and doesn't hurt the trivial case very much.

bkmartinjr (changes requested):
Overall this looks good.

I do have one performance suggestion inline, which I think also is slightly simpler code, and is worth implementing.

Side note:  this change still leaves issue #1975 open.  That could be resolved with a simple error check on the new "extended" enumeration list length.  If that is bounds check is added, it needs to be done _before_ the array evolution.  I'm fine if that fix (the bounds check) is done in a separate PR.

apis/python/src/tiledbsoma/_dataframe.py
nguyenv: Thanks for the tip about using Pandas unique. I'll make ther modification tomorrow. 

Originally I had used Python's built-in set and set operations but had issues with the enum values getting shuffled around.

bkmartinjr: Yeah, the nice thing about pandas.unique is that it preserves order.  From their docs:

> Uniques are returned in order of appearance. This does NOT sort.

nguyenv: I had to use a different approach to get the set difference, but it is still faster than before:

```
import pandas as pd
import numpy as np

def unique1(existing_cats, new_cats):
    update_vals = []
    for new_val in new_cats:
        if new_val not in existing_cats:
            update_vals.append(new_val)
    return existing_cats + update_vals
    
def unique2(existing_cats, new_cats):
    return pd.Series(np.concatenate([existing_cats, existing_cats, new_cats])).drop_duplicates(keep=False)
```

```
%timeit unique1( np.arange(1000).tolist(), np.arange(2000,0,-1).tolist())
%timeit unique2( np.arange(1000).tolist(), np.arange(2000,0,-1).tolist())
8.68 ms ± 216 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
292 µs ± 10.8 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

```
%timeit unique1( np.arange(5000).tolist(), np.arange(5500,4000,-1).tolist())
%timeit unique2( np.arange(5000).tolist(), np.arange(5500,4000,-1).tolist())
40 ms ± 1.35 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
654 µs ± 27.4 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -420,14 +420,14 @@ def write(
...
-                            update_vals.append(new_val)
+                    # get new enumeration values by taking the set difference
+                    # while maintaining ordering
+                    update_vals = np.setdiff1d(
+                        col.dictionary, enmr.values(), assume_unique=True
+                    )
 
                     # only extend if there are new values
-                    if update_vals:
+                    if len(update_vals) != 0:
```
bkmartinjr, line 430: nit: this can be `if update_vals: ...` as arrays are truthy when non-zero length

bkmartinjr (approved):
one optional nit.  otherwise LGTM

PR #1983, opened 2023-12-13
[r] Ensure factors are releveled before appending w/ expanded levels

eddelbuettel:
**Issue and/or context:**

When a `factor` variable `factor(c("A", "B", "A"))` is written, we write vector `c(1,2,1)` and labels `c("A", "B")`.  So far, so good.  As @bkmartinjr demonstrated if we then append a `factor(c("B", "C", "B"))` that is constructed without knowledge of the prior factor we also see `c(1,2,1)` and labels `c("B", "C"))`.   We were already correctly augmenting the stored enumeration labels to be `c("A", "B", "C")` but we still wrote `c(1,2,1)` -- when it should be `c(2, 3, 2)`.   

**Changes:**

The missing step of 're-levelling' the factor variable that is appended is now added.  

New tests have been added.

**Notes for Reviewer:**

[SC 38305](https://app.shortcut.com/tiledb-inc/story/38305/diagnose-multiple-categorical-write-issue-for-tiledb-soma-py-and-tiledb-soma-r)

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -19,6 +19,7 @@ env:
 jobs:
   ci:
     strategy:
+      fail-fast: false
       matrix:
         include:
           - name: linux
--- a/apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
+++ b/apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
@@ -152,7 +152,6 @@ test_that("Iterated Interface from SOMA Classes", {
 
 test_that("Iterated Interface from SOMA Sparse Matrix", {
     skip_if(!extended_tests() || covr_tests())
-    skip_on_ci()
     skip_if_not_installed("pbmc3k.tiledb")      # a Suggests: pre-package 3k PBMC data
 
     tdir <- tempfile()
--- a/apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
+++ b/apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
@@ -153,6 +153,7 @@ test_that("Iterated Interface from SOMA Classes", {
 test_that("Iterated Interface from SOMA Sparse Matrix", {
     skip_if(!extended_tests() || covr_tests())
     skip_if_not_installed("pbmc3k.tiledb")      # a Suggests: pre-package 3k PBMC data
+    skip_if(Sys.getenv("CI", "") != "")         # breaks only in CI so skipping
 
     tdir <- tempfile()
     tgzfile <- system.file("raw-data", "soco-pbmc3k.tar.gz", package="pbmc3k.tiledb")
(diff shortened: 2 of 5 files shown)


apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
```diff
@@ -152,8 +152,8 @@ test_that("Iterated Interface from SOMA Classes", {
 
 test_that("Iterated Interface from SOMA Sparse Matrix", {
     skip_if(!extended_tests() || covr_tests())
-    skip_on_ci()
     skip_if_not_installed("pbmc3k.tiledb")      # a Suggests: pre-package 3k PBMC data
+    skip_if(Sys.getenv("CI", "") != "")         # breaks only in CI so skipping
```
nguyenv, line 156: Does this segfault in CI? There seems to be issues with `SOMAArray` segfaulting recently and just want to make note.

.github/workflows/r-ci.yml
```diff
@@ -19,6 +19,7 @@ env:
 jobs:
   ci:
     strategy:
+      fail-fast: false
```
nguyenv, line 22: Revert?

apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
eddelbuettel: Yes. This one is driving me mad. Cannot replicate elsewhere (tried several machines and setups) but fails at CI. It already had 'skip this' which I erreneously removed (as it passes locally here for me, and in other places).

At CI, at least for now, we do need it.  Once I add it back in, it all turns :heavy_check_mark: on all runs as you saw.

.github/workflows/r-ci.yml
eddelbuettel: I would keep it simply because sometimes Coverage fails first and I find it informative to see what what macOS and standard non-Coverage Linux get up to.  Which is why I kept.

But not putting up a fight over it. If you (or someone else) feels the line must go we can delete it.

nguyenv: Doesn't matter to me either way

nguyenv (approved):
I can approve, but probably better to wait for someone who knows R to review

.github/workflows/r-ci.yml
mojaveazure: I would like to voice favor for keeping `fail-fast: false`; the coverage CI often fails first and swallows errors when it does. Allowing the Linux and macOS CIs to keep running will be beneficial for debugging errors

eddelbuettel:
No rush @nguyenv we can wait.  

The _key_ of the PR is one-line resetting the `factor` (aka dictionary) column to the expanded set of levels noticed -- the fix to the error @bkmartinjr reported for SOMA in Python, and the same issue @atolopko-czi reported to tiledb-py (for which I have an R repro).  The deeper issue really is the possibly disconnect between 'global' factor levels in the schema and the 'local' ones in the ephemeral data.frame object which are wrong as long as they ignore the global levels.  

The one-liner resets the factor to levels corresponding to the global set, and the unit test shows that.  There are two more CI tweaks in the PR (as we discussed) that also help to move us along.

PR #1986, opened 2023-12-14
[python] Error if extending enum past index type limit

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/1975

**Changes:**

Raise `ValueError` if the number of enumerations exceeds the limit of the index value type.


johnkerl (approved):
LGTM although I'd love to get @bkmartinjr 's eyes on it as well

PR #1987, opened 2023-12-15
Fix paths for  R codecov reporting

beroy:
**Issue and/or context:**

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -105,4 +105,4 @@ jobs:
 
       - name: Coverage
         if: ${{ matrix.os == 'ubuntu-latest' && matrix.covr == 'yes' }}
-        run: cd apis/r && tools/r-ci.sh coverage
+        run: apis/r/tools/r-ci.sh coverage && cd apis/r


.github/workflows/r-ci.yml
```diff
@@ -105,4 +105,4 @@ jobs:
 
       - name: Coverage
         if: ${{ matrix.os == 'ubuntu-latest' && matrix.covr == 'yes' }}
-        run: cd apis/r && tools/r-ci.sh coverage
+        run: apis/r/tools/r-ci.sh coverage && cd apis/r
```
eddelbuettel, line 108: I don't think this is right.

Many R add-on packages running in the source directory need the package layout: file `DESCRIPTION`, tests calles from `tests/*.R` etc.  Which is why we start with a `cd apis/r` to be in the right and _then_ run the R command invoked by my wrapper script `r-ci.sh`.  In full the call (after first installing R package `covr` and also taking care of some variables) is

    COVR="true" Rscript -e "covr::codecov(type = '${COVERAGE_TYPE}', quiet = FALSE, token = '${COVERAGE_TOKEN}', flags = '${COVERAGE_FLAGS}')"

and that has to run in that directory.

Can you describe what you are trying to change, and why?

eddelbuettel (changes requested):

PR #1990, opened 2023-12-15
[r] Tweak coverage data generation

eddelbuettel:
**Issue and/or context:**

Coverage for R has been generated with default values for some of the arguments leading to generated paths missing the initial two directories `apis/r/`.  By probing a little deeper @mojaveazure and I found we could accomodate this with a combination of setting the path relative to the repository root, _and_ setting the relative path argument too.  

**Changes:**

The runner script `r-ci.sh` has been updated accordingly.

**Notes for Reviewer:**

[SC 38463](https://app.shortcut.com/tiledb-inc/story/38463/refine-coverage)

### DIFF
--- a/.github/workflows/libtiledb-ci.yml
+++ b/.github/workflows/libtiledb-ci.yml
@@ -1,15 +1,15 @@
 name: libTileDB-SOMA CodeCov
 
 on:
-  # pull_request:
-  #   paths-ignore:
-  #     - "apis/python/**"
-  #     - "apis/r/**"
-  #     - ".pre-commit-config.yaml"
-  # push:
-  #   branches:
-  #     - main
-  #     - 'release-*'
+  pull_request:
+    paths-ignore:
+      - "apis/python/**"
+      - "apis/r/**"
+      - ".pre-commit-config.yaml"
+  push:
+    branches:
+      - main
+      - 'release-*'
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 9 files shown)


eddelbuettel:
I learned once again that the 'distance between "works here" and "works at GHA"' remainst constant at 'approximately four commits'.  It now runs but gets into a (new ?) failure under coverage for which I have to add a log dumper.  Later, or maybe tomorrow.

beroy (approved):

eddelbuettel:
One step forward, two steps back: it 'runs' given the path settings, but it doesn't attribute any of the tests. :sob:  So we get the correct filename, but a bunch of zeros reported.  In short, we are still fighting to adjust package `covr` to our repo layout.

eddelbuettel:
We seem to be getting in our own way.  On may laptop (no branches, main repo of TileDB Core, R, SOMA all a few days old) I can easily do

```r
> library(covr)
> setwd("apis/r")
> Sys.setenv("CI"="TRUE")                           # to have more tests run
> res <- package_coverage(relative_path="../..")    # all defaults but inform about relative paths
> report(res)                                       # results in browser
```

Plus we can easily use the `line_exclusion=list(...)` argument to exclude `nanoarrow.*` (vendored) and `roxygen.R` (doc generation) after which we end up at 83% which isn't a bad place.

eddelbuettel:
Some more progress.  

image

PR #1993, opened 2024-01-02
[python] Allow specification of X names on ingest

johnkerl:
**Issue and/or context:**

AnnData's `X` and `raw/X` were being named `X["data"]` unconditionally. Users need to be able to choose the name within the TileDB-SOMA experiment.

**Changes:**

**Notes for Reviewer:**


eddelbuettel:
Thumbs up on trying the 'controlled downgrade'.  It also needs to go to the inter-op test. I can take a peek if you want, the mechanics should be similar.

eddelbuettel:
BTW my grand plan of the 'controlled downgrade' got stuck in the same mud you have now:  The macOS variant does not build _before we downgrade_ meaning we cannot test and compare versions and then potentially downgrade.  I have no fix for that macOS.

johnkerl:
Thanks @eddelbuettel !

I'm puzzled, as I don't see a reason within `r-ci.yml` for MacOS and Linux to behave differently here ...

mlin:
@eddelbuettel @johnkerl We have also been seeing the `Symbol not found: (_tiledb_group_get_member_by_index_v2)` since Dec 21 in [cellxgene.census R macOS nightly test](https://github.com/chanzuckerberg/cellxgene-census/actions/workflows/r-dependency-check.yml) that installs all deps from scratch. I'm relieved it's on your radar already. Tracking this and please LMK if I can help

eddelbuettel:
@mlin I will admit to being something stunned by the error.  Maybe a fresh set of eyes can find some insight, I do not understand why/how it fails (apart from the known complication arising from various levels of pinning in different part of the SOMA repo coupled with the default of R pulling the newest available version from selected repos -- I worked on a helper function to overcome this but it too is blocked by this).

eddelbuettel:
Yes as we all know it's a long story, it's complicated, and it has a new-ish puzzle (as I obviously hit the same macOS issue now blocking you when I wrote the 'controlled downgrade' attempt, which works fine on Linux.

Part of the issue is that e.g. the line "TileDB-R 0.23 exists and depends on core 2.19" is exactly not correct as stated.  It will build with many versions down to 2.0.8 (tested) and up to dev (tested nighlty) but it has a _fallback_ of 2.19. Unless a different framework is put in place (for which 'controlled downgrade' was an attempt to supply one) it will use 2.19. But if overrides or alternative are present they will be used.  We are doing all this with the very best intentions, but we are making life somewhat complicated for both ourselves and our users.

johnkerl:
Please see also https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases#bug-fix-branches

johnkerl:
Controlled-downgrade issues have been resolved thanks to #2002, which this PR is now atop.

johnkerl:
@aaronwolen ready for review 🙏

johnkerl:
@aaronwolen ping 🙏

aaronwolen (approved):

johnkerl:
Thanks @aaronwolen !

PR #1994, opened 2024-01-02
[r] Controlled downgrade of TileDB-R to ensure matching TileDB Core

eddelbuettel:
**Issue and/or context:**

The TileDB-R package builds against TileDB Core, and while we can ensure a _minimum_ version
we cannot ensure an exact version.  That can lead to cases as right now where TileDB-R 0.23.0 is 
at CRAN with a default of TileDB Core 2.19.0, yet TileDB-SOMA is still set up to build with TileDB 
Core 2.18.2.  This mismatch is undesirable but a little difficult to avoid for example in continuous 
integration where R will always go to the newest package version.

As an experiment, @mojaveazure  and I have set up two helper repos. The first ensures we can still build
the _previous_ TileDB-R version at r-universe (which would otherwise pick newest versions from
CRAN and/or GitHub).  The second deploys these builds in a particular R repository hosted on
GitHun from which we can install.  Together, these provide a simple 'time machine' ability to
the previous binary.

To illustrate, by default we end with these packages (when running on Ubuntu 22.04 as for CI)

```r
  > show_package_versions()
  tiledbsoma:    1.6.0
  tiledb-r:      0.23.0
  tiledb core:   2.19.0
  libtiledbsoma: 2.18.2
  R:             R version 4.3.2 (2023-10-31)
  OS:            Ubuntu 22.04.1 LTS
  >
```

Whereas once we do a controlled downgrade we get

```r
  > library(tiledbsoma)
  > show_package_versions()
  tiledbsoma:    1.6.0
  tiledb-r:      0.22.0
  tiledb core:   2.18.2
  libtiledbsoma: 2.18.2
  R:             R version 4.3.2 (2023-10-31)
  OS:            Ubuntu 22.04.1 LTS
  >
```

**Changes:**

The PR adds a new helper script and invokes it from the r-valgrind action for continuous integration.

**Notes for Reviewer:**

[SC 38681](https://app.shortcut.com/tiledb-inc/story/38681/add-controlled-downgrade-of-tiledb-r-in-ci-to-ensure-matching-tiledb-core-versions)


apis/r/tools/controlled_downgrade.R
```diff
@@ -0,0 +1,63 @@
...
+
+if (!isLinux && !isMacX86) 								# we only support Linux and macOS/x86_64
+    q()
+
+if (isLinux && !grepl("Ubuntu 22.04", utils::osVersion))# on Linux we only support Ubuntu 22.04
+    q()
+
+core_from_tiledb <- tiledb::tiledb_version(compact=TRUE)
+core_from_soma <- tiledbsoma:::libtiledbsoma_version(compact=TRUE)
+
```
johnkerl, line 48: On #1993 I found it helpful to include

```
cat("core_from_tiledb", as.character(core_from_tiledb), "\n")
cat("core_from_soma", core_from_soma, "\n")
```

eddelbuettel: Yes -- I developed this in another repo first (to keep things calmer here) and did similar things there.

eddelbuettel:
I don't have the energy to debate this endlessly but we had reports of failed installations just days ago.  What we set up works _if and only if the stars align and tiledb-r is not ahead_.  I maintain what we can do better.

johnkerl:
@eddelbuettel these are all for CI, and CI works. When @nguyenv is done with C++-ification and removing our dependency on TileDB-R (and TileDB-Py), this will auto-solve.

eddelbuettel:
We reports from users being unable to install.  That is not CI.

PR #1995, opened 2024-01-04
[python] Allow specification of X names on ingest [release-1.5]

johnkerl:
**Issue and/or context:** #1993 for the `release-1.5` branch

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -274,11 +274,7 @@ def from_h5ad(
         platform_config: Platform-specific options used to create this array, provided in the form
         ``{"tiledb": {"create": {"sparse_nd_array_dim_zstd_level": 7}}}`` nested keys.
 
-        obs_id_name: Which AnnData ``obs`` column to use for append mode.  This may be either
-        an index column or a non-index column.
-
-        var_id_name: Which AnnData ``var`` column to use for append mode.  This may be either
-        an index column or a non-index column.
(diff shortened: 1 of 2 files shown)


johnkerl:
@aaronwolen ping 🙏

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -272,6 +274,14 @@ def from_h5ad(
         platform_config: Platform-specific options used to create this array, provided in the form
         ``{"tiledb": {"create": {"sparse_nd_array_dim_zstd_level": 7}}}`` nested keys.
 
+        obs_id_name: Which AnnData ``obs`` column to use for append mode.
+
+        var_id_name: Which AnnData ``var`` column to use for append mode.
```
aaronwolen, line 279: Thanks for documenting these! A little more info on how they're used for appending might out help users. I think these args are also used to rename the index of incoming obs/var dataframes prior to ingest, right?

aaronwolen (approved):
One minor comment re docs but otherwise looks perfect!

johnkerl:
@aaronwolen thank you for your review!

https://github.com/single-cell-data/TileDB-SOMA/pull/1995/commits/f3963c5ab2d12b85867c18bc2fc96b541e4e0ff7 offers a bit more detail -- however, I'm not sure what you want -- if you could state what you want (an example sentence perhaps) that would be helpful -- thank you!

aaronwolen:
Just noting we discussed and resolved on a call.

PR #1996, opened 2024-01-04
Allow specification of X names on ingest [release-1.6]

johnkerl:
**Issue and/or context:** #1993 for the `release-1.6` branch

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -85,7 +85,14 @@ jobs:
       #  run: sudo ldconfig
       #
       - name: Update Packages
-        run: Rscript -e 'update.packages(ask = FALSE, oldPkgs = setdiff(rownames(old.packages()), "tiledb"))'
+        run: Rscript -e 'update.packages(ask=FALSE)'
+
+      - name: Re-install 0.22 build of tiledb-r with core 2.18
+        run: |
+          cd apis/r
+          wget https://github.com/TileDB-Inc/TileDB-R/archive/refs/tags/0.22.0.tar.gz
+          Rscript -e 'remotes::install_deps("0.22.0.tar.gz")'
+          R CMD INSTALL 0.22.0.tar.gz    # as 0.22.0 is needed by SOMA 1.6
 
       - name: Test
         if: ${{ matrix.covr == 'no' }}
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -53,16 +53,12 @@ jobs:
       - name: Install BioConductor package SingleCellExperiment
         run: cd apis/r && tools/r-ci.sh install_bioc SingleCellExperiment
 
-      # Uncomment these next two stanzas as needed whenever we've just released a new tiledb-r for
-      # which source is available but CRAN releases (and hence update r2u binaries) are not yet:
-
-      - name: Install r-universe build of tiledb-r (macOS)
-        if: ${{ matrix.os == 'macOS-latest' }}
-        run: cd apis/r && Rscript -e "install.packages('tiledb', repos = c('https://eddelbuettel.r-universe.dev', 'https://cloud.r-project.org'))"
-
-      - name: Install r-universe build of tiledb-r (linux)
-        if: ${{ matrix.os != 'macOS-latest' }}
-        run: cd apis/r && Rscript -e "options(bspm.version.check=TRUE); install.packages('tiledb', repos = c('https://eddelbuettel.r-universe.dev/bin/linux/jammy/4.3/', 'https://cloud.r-project.org'))"
+      - name: Install 0.22 build of tiledb-r with core 2.18
+        run: |
+          cd apis/r
+          wget https://github.com/TileDB-Inc/TileDB-R/archive/refs/tags/0.22.0.tar.gz
+          Rscript -e 'remotes::install_deps("0.22.0.tar.gz")'
+          R CMD INSTALL 0.22.0.tar.gz    # as 0.22.0 is needed by SOMA 1.6
 
       - name: Install r-universe build of SeuratObject (macOS)
         if: ${{ matrix.os == 'macOS-latest' }}
@@ -90,6 +86,13 @@ jobs:
       #
       - name: Update Packages
         run: Rscript -e 'update.packages(ask=FALSE)'
+
+      - name: Re-install 0.22 build of tiledb-r with core 2.18
+        run: |
+          cd apis/r
+          wget https://github.com/TileDB-Inc/TileDB-R/archive/refs/tags/0.22.0.tar.gz
+          Rscript -e 'remotes::install_deps("0.22.0.tar.gz")'
+          R CMD INSTALL 0.22.0.tar.gz    # as 0.22.0 is needed by SOMA 1.6
 
       - name: Test
         if: ${{ matrix.covr == 'no' }}
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -85,14 +85,7 @@ jobs:
       #  run: sudo ldconfig
       #
       - name: Update Packages
-        run: Rscript -e 'update.packages(ask=FALSE)'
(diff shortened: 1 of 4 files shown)


johnkerl:
Note: the 2nd commit https://github.com/single-cell-data/TileDB-SOMA/pull/1996/commits/be8c808c6468d8b2ce446cf8735b5ff35f122d00 on this PR is a port of #1972's pin of TileDB-R from `release-1.5` (which it is known-good) to `release-1.6`.

.github/workflows/r-ci.yml
```diff
@@ -53,16 +53,12 @@ jobs:
...
-
-      - name: Install r-universe build of tiledb-r (linux)
-        if: ${{ matrix.os != 'macOS-latest' }}
-        run: cd apis/r && Rscript -e "options(bspm.version.check=TRUE); install.packages('tiledb', repos = c('https://eddelbuettel.r-universe.dev/bin/linux/jammy/4.3/', 'https://cloud.r-project.org'))"
+      - name: Install 0.22 build of tiledb-r with core 2.18
+        run: |
+          cd apis/r
+          wget https://github.com/TileDB-Inc/TileDB-R/archive/refs/tags/0.22.0.tar.gz
+          Rscript -e 'remotes::install_deps("0.22.0.tar.gz")'
+          R CMD INSTALL 0.22.0.tar.gz    # as 0.22.0 is needed by SOMA 1.6
```
eddelbuettel, line 61: The idea of 'controlled download' was to do just this, but via binaries.  Paul and I put some work into this.  But let's see if brute works.

.github/workflows/r-ci.yml
```diff
@@ -91,6 +87,13 @@ jobs:
       - name: Update Packages
         run: Rscript -e 'update.packages(ask=FALSE)'
 
+      - name: Re-install 0.22 build of tiledb-r with core 2.18
+        run: |
+          cd apis/r
+          wget https://github.com/TileDB-Inc/TileDB-R/archive/refs/tags/0.22.0.tar.gz
+          Rscript -e 'remotes::install_deps("0.22.0.tar.gz")'
+          R CMD INSTALL 0.22.0.tar.gz    # as 0.22.0 is needed by SOMA 1.6
```
eddelbuettel, line 95: You should not need to do it twice, you can can skip the `update.packages(ask=FALSE)` too.  That was from a different time fixing a different issue.

.github/workflows/r-ci.yml
johnkerl: @eddelbuettel in fact this has already worked on `release-1.5`.

.github/workflows/r-ci.yml
johnkerl: @eddelbuettel I went through multiple CI iterations on #1972 on the `release-1.5` branch to find, and use, a known-good pattern which we could then apply for subsequent PRs on subsequent branches.

This PR is a such a subsequent PR on a subsequent branch (`release-1.6`).

.github/workflows/r-ci.yml
eddelbuettel: I did the same in December, believe it or not, but stopped when I saw no light on either end of any tunnel.

.github/workflows/r-ci.yml
```diff
@@ -91,6 +87,13 @@ jobs:
       - name: Update Packages
         run: Rscript -e 'update.packages(ask=FALSE)'
```
eddelbuettel, line 88: I recommend to remove this block.  It would allow you to remove the next block (ie lines 90 to 95 as well).  And cut CI time down by a few minutes.

/CC @mojaveazure

eddelbuettel (commented):
See the comment inline.  @mojaveazure raised the same point, correctly, and may want to chime in.

It's good that it builds now -- big step forward.  But we don't need to build it twice.

.github/workflows/r-ci.yml
johnkerl: @eddelbuettel trying commit #3 https://github.com/single-cell-data/TileDB-SOMA/pull/1996/commits/de92e139b12dad908df1bf4da827a328143d0f7e

johnkerl: @eddelbuettel:

* Commit 2 https://github.com/single-cell-data/TileDB-SOMA/pull/1996/commits/be8c808c6468d8b2ce446cf8735b5ff35f122d00 had green CI
* There were 3 of 3 CI fails with commit number 3: https://github.com/single-cell-data/TileDB-SOMA/pull/1996/commits/de92e139b12dad908df1bf4da827a328143d0f7e
* Commit 4 https://github.com/single-cell-data/TileDB-SOMA/pull/1996/commits/5a76da15cd3a88e51370720fdba0df77dd3ace74 reverts to commit 2 and has green CI

eddelbuettel: If I look at the correct page not three for three.  One, plus two cancels. We know coverage is more fragile.

https://github.com/single-cell-data/TileDB-SOMA/actions/runs/7449708671/job/20267723363

johnkerl: @eddelbuettel I watched each one.

* Commit 3 had three fails
* Commit 4 had a cancel (I don't know why), then a successful re-run

johnkerl: @eddelbuettel I am happy to re-run the already-green CI another time for commit number 4.

eddelbuettel: Am I looking in the wrong place?

image

And 

image

Anyway, I have no beef with this but it may be a good idea to get someone else to look at this.

.github/workflows/r-ci.yml
```diff
@@ -91,6 +87,13 @@ jobs:
...
 
+      - name: Re-install 0.22 build of tiledb-r with core 2.18
+        run: |
+          cd apis/r
+          wget https://github.com/TileDB-Inc/TileDB-R/archive/refs/tags/0.22.0.tar.gz
+          Rscript -e 'remotes::install_deps("0.22.0.tar.gz")'
+          R CMD INSTALL 0.22.0.tar.gz    # as 0.22.0 is needed by SOMA 1.6
+
```
mojaveazure, line 96: Can we try updating all packages except tiledb-r? This should keep tiledb-r on the required version, allowing us to remove the re-install block (currently lines 90-95)
```suggestion
      - name: Update Packages
        run: Rscript -e 'update.packages(ask = FALSE, oldPkgs = setdiff(rownames(old.packages()), "tiledb"))'

```

johnkerl:
@mojaveazure with respect to commit https://github.com/single-cell-data/TileDB-SOMA/pull/1996/commits/3ce3836ca0c6dcd10e70952064f18d4ff1c25822 we have red CI at
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/7451362225/job/20272317064

johnkerl:
@mojaveazure CI fail with

```
Symbol not found: (_tiledb_group_get_member_by_index_v2)
```

(2.18 vs 2.19) on commit number 5: https://github.com/single-cell-data/TileDB-SOMA/pull/1996/commits/3ce3836ca0c6dcd10e70952064f18d4ff1c25822

I am again reverting to commits 4 and 2 (which are identical) -- this will be a commit number -- using the proven-good approach for those commits.

mojaveazure (approved):
Dissapointing that the controlled update didn't work; it seems like our best bet is the update + reinstall

eddelbuettel:
:crying_cat_face:   Sometimes blunt tools are all we have.  

Also worth reiterating that AFAICT all this work 'merely' lets us get by in CI, and may do little for users of SOMA who still fight similar battles.

eddelbuettel:
I just reiterated things internally in a conversartion, it is not ideal how things are set up because it makes builds from source harder than they maybe could be.  You are fighting a heroic battle with the build system, yet there are still more possible build permutations behind other curtains.  It's hard and complex ...

Thanks for getting this one finished!

PR #2000, opened 2024-01-08
Update to use somacore==1.0.7.

thetorpedodog:
somacore v1.0.7 is fully compatible with v1.0.6 and does not introduce any behavior changes. It does *prepare* us for the reindexer change: https://github.com/single-cell-data/TileDB-SOMA/pull/1728

Note: This can be folded into #1728 if we want, and I have no objections to doing so.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
     - id: mypy
       additional_dependencies:
         - "pandas-stubs==1.5.3.230214"
-        - "somacore==1.0.6"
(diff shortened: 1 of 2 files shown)


johnkerl:
Let's wait to merge this until TileDB-Py 0.25 is tagged (cc @ihnorton) per https://github.com/single-cell-data/TileDB-SOMA/pull/1993#issuecomment-1881150316

johnkerl:
@beroy @thetorpedodog I just merged in #2002

johnkerl (approved):

PR #2001, opened 2024-01-09
[python] Add default thread pool in `SOMATileDBContext`

### LINKED ISSUE #1836: [Feature request] [python] Consolidate thread-pool support
There are several thread pools created in the current Python SOMA:

* the new blockwise iterators (for eager read-ahead)
* ExperimentAxisQuery (also eager read-ahead)
* Fast CSR

There are likely to be more as we progress. This is not desirable as it can lead to over-subscription, etc.

It would be ideal if:
* there was a single thread pool for Python-level concurrency
* it supported user control (e.g. over level of concurrency), a la asyncio's `loop.set_default_executor()`

Suggestion from @thetorpedodog is to add a the thread pool to SOMATileDBContext, and provide user control over setting it.

As there is a lot of code in `somacore` that also uses this, we may need to consider refactoring the `somacore` code so that it can find the threadpool, or introducing a base/abstract "Context" class that contains the threadpool, and from which SOMATileDBContext is derived.

@thetorpedodog - if you have other thoughts, would be great if you could add them to this issue.

Note: this is entirely separate from the C++-level thread pool(s).

ebezzi:
**Issue and/or context:**
Resolves #1836 

**Changes:**
- Adds a `threadpool` parameter to `SOMATileDBContext` - instantiates one by default if not specified
- Uses this threadpool in the blockwise iterator (as mentioned, there is a possible issue as currently `EagerIterator` reserves one thread and could send the thread pool into starvation if there are not enough workers)
- Uses a threadpool in ExperimentAxisQuery: https://github.com/single-cell-data/SOMA/pull/184
- Adds tests.

**Notes for Reviewer:**
See comments in PR.


apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -129,6 +135,9 @@ def __init__(
         """The TileDB context to use, either provided or lazily constructed."""
         self._timestamp_ms = _maybe_timestamp_ms(timestamp)
 
+        """User specified threadpool. If None, we'll instantiate one ourselves."""
+        self._threadpool = threadpool or futures.ThreadPoolExecutor()
```
ebezzi, line 139: We could change the implementation so, if the thread pool isn't specified by the user, then it is created lazily. Not sure it's worth it as the overhead should be minimal.

apis/python/tests/test_experiment_query.py
```diff
@@ -83,6 +84,10 @@ def soma_experiment(
 
     return _factory.open((tmp_path / "exp").as_posix())
 
+def get_soma_experiment_with_context(soma_experiment, context):
```
ebezzi, line 87: Not a huge fan of this, but parametrizing the `context` in the `soma_experiment` fixture seemed even worse. Open to ideas on how to improve this.

apis/python/tests/test_experiment_query.py
```diff
@@ -871,3 +882,37 @@ def add_sparse_array(coll: CollectionBase, key: str, shape: Tuple[int, int]) ->
...
     a.write(tensor)
+
+@pytest.mark.parametrize("n_obs,n_vars", [(1001, 99)])
+def test_experiment_query_uses_threadpool_from_context(soma_experiment):
+    """
+    Verify that ExperimentAxisQuery uses the threadpool from the context
+    """
+
+    from concurrent import futures
+    class SentinelThreadPoolExecutor(futures.ThreadPoolExecutor):
```
ebezzi, line 893: This needs to be DRY'd and put somewhere else, but I don't see a suitable place to do it (normally, `conftest.py` would be a good candidate, but it isn't present). Any preference?

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
thetorpedodog: Yeah, just creating a threadpool doesn’t actually start any threads until they are called for. This is 100% fine.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -129,6 +135,9 @@ def __init__(
         """The TileDB context to use, either provided or lazily constructed."""
         self._timestamp_ms = _maybe_timestamp_ms(timestamp)
 
+        """User specified threadpool. If None, we'll instantiate one ourselves."""
```
thetorpedodog, line 138: this belongs immediately *below* setting `self._threadpool`.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -234,8 +246,10 @@ def replace(
             if timestamp is _SENTINEL:
                 # Keep the existing timestamp if not overridden.
                 timestamp = self._timestamp_ms
+            if threadpool is None:
+                threadpool = self._threadpool
```
thetorpedodog, line 250: Do we want this to be `_SENTINEL` by default, so that you can make this choice?

- leave the existing threadpool (i.e., unset; left as `_SENTINEL`)
- use your own threadpool (i.e., `threadpool=my_threadpool`)
- use a new default threadpool (i.e., `threadpool=None`)

This is a fairly niche use case so I am not _hugely_ concerned if we don’t do this, but just as a thought.

apis/python/tests/test_experiment_query.py
thetorpedodog: You can avoid the need for this by using a Mock:

```python

from unittest import mock

# ...

pool = mock.Mock(wraps=futures.ThreadPoolExecutor(...))
```

then use the [mock assertion methods](https://docs.python.org/3/library/unittest.mock.html#the-mock-class):

```python
pool.submit.assert_called()
# Or use the more advanced methods, like assert_called_with, etc.
```

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
bkmartinjr: I advocate for consistency with the rest of the class. I don't have an opinion which is correct, but if we think the `timestamp` behavior is right, the other params should follow this pattern.

ebezzi: Yes, I agree. It will be useful for testing at least. Thanks for suggesting the change.

apis/python/tests/test_experiment_query.py
ebezzi: Ah, didn't know about `wraps=`! This is great, thanks.

ebezzi:
> These changes look good.
> 
> I'd like to see the SOMA PR before approving. I think this PR implies that we need a "Context" base class in SOMA (or a protocol) to allow ExperimentAxisQuery to utilize the `_pool` slot from a context object. At the moment, I believe the somacore.SOMAObject.context is defined as `Any`....

PR [here](https://github.com/single-cell-data/SOMA/pull/184)

bkmartinjr (commented):
Another thought:  it would be prudent to add a check in SOMATileDBContext about likely deadlock scenarios.  I'm thinking of something like the following:

```
    if self._threadpool._max_workers < MAGIC_NUM:
        warning.warning("Insufficient workers in threadpool may cause deadlock - suggest increasing worker count")
```

or something like that.  Where MAGIC_NUM is hardwired to our best guess (3 or 4?)

thetorpedodog:
> Another thought: it would be prudent to add a check in SOMATileDBContext about likely deadlock scenarios. I'm thinking of something like the following:
> 
> ```
>     if self._threadpool._max_workers < MAGIC_NUM:
>         warning.warning("Insufficient workers in threadpool may cause deadlock - suggest increasing worker count")
> ```
> 
> or something like that. Where MAGIC_NUM is hardwired to our best guess (3 or 4?)

This is concerning—in my view, when things are dispatched to a thread pool, it should be safe to serialize the operations in question. That is to say, semantically, I look at a thread pool as just a work queue where the tasks are *allowed* to be executed at the same time. (For instance, the passed-in thread pool might already be doing N−1 other long-running tasks, leaving only one thread for us.)

Do we understand the mechanism by which the deadlock is caused?

bkmartinjr:
> Do we understand the mechanism by which the deadlock is caused?

Yep.  Each EagerIterator use consumes a thread, which it holds for the lifetime of the eager object. In some small number of places, there is eager iterator use nested in other code using eager iterator (max dep of 2 or 3 I think).  If the thread pool is small, and exhausted during this call stack, it will deadlock due to thread availability. 

Options discussed:
* have every eager iter allocate a stand-alone thread (do not participate in the shared pool)
* warn or error if the shared pool is smaller than the max required depth

We didn't really nail down the "right" answer. I've been leaning toward the first.  While threads are not cheap, eager iter is intended for use in heavy weight operations where the cost of thread create/delete should not be a major issue.

ebezzi:
Other possible options:

- Have `BlockwiseReadIterBase` (and any future classes using EagerIterator) manage its own ThreadPool (basically, revert this change for blockwise iterators)
- Have `BlockwiseReadIterBase` manage its own ThreadPool  _only if_ the context-derived iterator has less than the number of required workers.

None of the solutions are particularly challenging to implement, so @thetorpedodog let us know what you think.

thetorpedodog:
> > Do we understand the mechanism by which the deadlock is caused?
> 
> Yep. Each EagerIterator use consumes a thread, which it holds for the lifetime of the eager object. In some small number of places, there is eager iterator use nested in other code using eager iterator (max dep of 2 or 3 I think). If the thread pool is small, and exhausted during this call stack, it will deadlock due to thread availability.
> 
> Options discussed:
> 
> * have every eager iter allocate a stand-alone thread (do not participate in the shared pool)
> * warn or error if the shared pool is smaller than the max required depth
> 
> We didn't really nail down the "right" answer. I've been leaning toward the first. While threads are not cheap, eager iter is intended for use in heavy weight operations where the cost of thread create/delete should not be a major issue.

In the case of the second, we can still run out of resources (and this applies equally to @ebezzi’s second solution, which he posted while I was in the process of writing this):

1. have a sufficiently large threadpool in the context.
2. start two queries at the same time.
3. nested EagerIterators start on each.
4. oops! both got N−1 threads when they needed N.

Given that, I think that of the two the first one is the better option, but I think we can go even better and respect the threadpool while avoiding deadlock:

If the EagerIterator prefetch has not yet started, just compute it ourselves. I have a change in progress to do just this (https://github.com/single-cell-data/SOMA/pull/185).

The result is that we

- respect the limits of the passed-in thread pool,
- prefetch if possible, and
- don’t deadlock when we can’t prefetch.

bkmartinjr:
> If the EagerIterator prefetch has not yet started, just compute it ourselves

well, that is clever!  TY

bkmartinjr (approved):

bkmartinjr:
@ebezzi - bumping this. It LGTM, can we land the SOMA PR and get this into the release train?

ebezzi:
@bkmartinjr I just merged the related `somacore` PR. @johnkerl , when you have some time, would you mind doing a release? Thanks 🙏

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -261,19 +268,11 @@ class BlockwiseTableReadIter(BlockwiseReadIterBase[BlockwiseTableReadIterResult]
...
-            yield from (
-                self._reindexed_table_reader(_pool=self.eager_iterator_pool)
-                if self.axes_to_reindex
-                else self._table_reader()
-            )
+        yield from (
+            self._reindexed_table_reader(_pool=self._threadpool)
+            if self.axes_to_reindex
+            else self._table_reader()
+        )
```
thetorpedodog, line 275: Since this no longer uses a context manager, there’s no need to `yield from` here; you can just `return` the generator.

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -97,7 +97,14 @@ def __init__(
...
         self.sr = sr
         self.eager = eager
-        self.eager_iterator_pool = eager_iterator_pool
+
+        # Assign a thread pool from the context, or create a new one if no context
+        # is available
+        if context is not None:
+            self._threadpool = context.threadpool
+        else:
+            self._threadpool = futures.ThreadPoolExecutor()
```
thetorpedodog, line 106: since `context` is (presumably) always going to be a truthy object, this can be `if context` (thus reducing the amount of negatives)

thetorpedodog (approved):
two trivial things, neither of which are critical

PR #2005, opened 2024-01-09
[r] Bump `apis/r/DESCRIPTION` for `main` past current latest release

johnkerl:
* https://github.com/single-cell-data/TileDB-SOMA/releases/tag/1.6.1 exists
* `main` should be "past" that

### DIFF
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -6,7 +6,7 @@ Description: Interface for working with 'TileDB'-based Stack of Matrices,
     like those commonly used for single cell data analysis. It is documented at
     <https://github.com/single-cell-data>; a formal specification available is at
     <https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md>.
-Version: 1.6.0
+Version: 1.6.1.1
 Authors@R: c(
     person(given = "Aaron", family = "Wolen",
            role = c("cre", "aut"), email = "aaron@tiledb.com",
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -6,7 +6,7 @@ Description: Interface for working with 'TileDB'-based Stack of Matrices,
     like those commonly used for single cell data analysis. It is documented at
     <https://github.com/single-cell-data>; a formal specification available is at
     <https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md>.
-Version: 1.6.1.1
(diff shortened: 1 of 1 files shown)


eddelbuettel (commented):
Maybe 1.6.9, or 1.6.99, 1.6.9000 is better as we may well have releases 1.6.2, 1.6.3, ... and so on which would once again jump ahead of the number here.

If we pick a higher 'delta' we'd be safe from that confusion that 'newer sorts lower' which is not desirable.

aaronwolen (approved):

johnkerl:
Sorry, I changed from 1.6.1.1 to 1.6.99 on @eddelbuettel ' s advice but then realized @aaronwolen had accepted before my mod -- re-requesting review ...

johnkerl:
In Slack, Aaron propopsed 1.6.1.1.

@aaronwolen @eddelbuettel @mojaveazure I am serving in a clerical role here. Please offer a 2/3 majority opinion and I'll commit it.

mojaveazure:
My 2¢: if a theoretical 1.6.2 would pull from `main`, I vote for 1.6.1.1; otherwise, 1.6.99 or whatever

johnkerl:
@mojaveazure per our [established procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases) a hypothetical 1.6.2 would be on the `release-1.6` branch, not on `main`.

aaronwolen:
Agreed, 1.6.1.1 aligns with the convention we've been using to be consistent with tiledb-r.

mojaveazure (approved):

johnkerl:
* @eddelbuettel is OK with 1.6.99, and less OK with 1.6.1.1
* @mojaveazure prefers 1.6.99 if a 1.6.2 would be tagged from `release-1.6` rather than `main`, which is in fact the case
* @aaronwolen prefers 1.6.1.1
* As above, I am not an R expert
* 2 of 3 go with 1.6.99

johnkerl:
Our [established procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases) has been updated to reflect this decision:

https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases/_compare/a4d2ebf66de4128d8f2987290e8c0e3789098701...747bbebab373dbfeceff35ea03d66e8cb455c3b2

PR #2008, opened 2024-01-11
[r] Ensure addition of factor levels does not overflow capacity of index type

eddelbuettel:
**Issue and/or context:**

As discussed in #1975, index type can overflow when factor levels are added.

**Changes:**

When factor levels are added, the sum of 'before plus added' is compared to the capacity of the index type, and an error is signalled if the latter is exceeded.  An unit test has been where writing 130 levels exceeds the possible 128 on `int8_t` leading to error and passes on `uint8_t` indexing.

**Notes for Reviewer:**

[SC 39079](https://app.shortcut.com/tiledb-inc/story/39079/r-ensure-factor-levels-do-not-overflow)
[SC 38448](https://app.shortcut.com/tiledb-inc/story/38448/dataframe-categorical-handling-does-not-handle-extension-past-index-type-limit)
#1975

### DIFF
--- a/apis/r/tests/testthat/test-SOMADataFrame.R
+++ b/apis/r/tests/testthat/test-SOMADataFrame.R
@@ -820,3 +820,28 @@ test_that("factor levels can grow without overlap", {
     ref <- rbind( tibble::as_tibble(tbl_1), tibble::as_tibble(tbl_2) )
     expect_equal(tbl, ref)
 })
+
+test_that("factor levels cannot beyond index limit", {
+    for (tp in c("INT8", "UINT8")) {
+        uri <- tempfile()
+        idx_type <- if (tp == "INT8") arrow::int8() else arrow::uint8()
+        sch <- arrow::schema(soma_joinid = arrow::int64(),
+                             obs = arrow::dictionary(index_type = idx_type,
+                                                     value_type = arrow::string()))
+        df <- data.frame(soma_joinid = bit64::as.integer64(seq_len(65)),
+                         obs = factor(paste0("elem", seq_len(65))))
+        tbl <- arrow::as_arrow_table(df, schema = sch)
+        expect_silent(SOMADataFrameCreate(uri, sch)$write(tbl)$close())
+
+        df2 <- data.frame(soma_joinid =  bit64::as.integer64(65 + seq_len(65)),
+                          obs = factor(paste0("elem_", 65 + seq_len(65))))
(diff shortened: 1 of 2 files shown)


apis/r/tests/testthat/test-SOMADataFrame.R
```diff
@@ -820,3 +820,28 @@ test_that("factor levels can grow without overlap", {
     ref <- rbind( tibble::as_tibble(tbl_1), tibble::as_tibble(tbl_2) )
     expect_equal(tbl, ref)
 })
+
+test_that("factor levels cannot beyond index limit", {
```
johnkerl, line 824:
```suggestion
test_that("factor levels cannot extend beyond index limit", {
```

johnkerl (approved):

apis/r/tests/testthat/test-SOMADataFrame.R
eddelbuettel: Good catch.  Will add that as a quick `[cp skip]` commit!

eddelbuettel: Done.

apis/r/tests/testthat/test-SOMADataFrame.R
```diff
@@ -820,3 +820,28 @@ test_that("factor levels can grow without overlap", {
     ref <- rbind( tibble::as_tibble(tbl_1), tibble::as_tibble(tbl_2) )
     expect_equal(tbl, ref)
 })
+
+test_that("factor levels cannot extend beyond index limit", {
+    for (tp in c("INT8", "UINT8")) {
+        uri <- tempfile()
```
mojaveazure, line 826: This should probably be`withr::local_tempfile()` for consistency and to ensure automatic cleanup of test files

eddelbuettel: Strong disagree.  I remove these as they littler things in my use (often driven with `r`).

eddelbuettel: R itself has a guarantee to clean up testfile in its tempdir at the end of each session. I love to run tests in independent R sessions (via `r` or `Rscript`) yet this gets in the way and it is irritating:

```sh
edd@rob:/tmp$ ls -1d callr-sev-* devtools_install_* scoped_* soma-dataframe-* summarized-experiment* tiledb-rest-result-storage* | wc -l
174
edd@rob:/tmp$ 
```

PR #2012, opened 2024-01-12
[Backport release-1.6] [r] Ensure factors are releveled before appending w/ expanded levels …

johnkerl:
Backport #1983 to `release-1.6`

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -19,6 +19,7 @@ env:
 jobs:
   ci:
     strategy:
+      fail-fast: false
       matrix:
         include:
           - name: linux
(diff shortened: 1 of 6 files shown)


eddelbuettel (approved):
Indeed -- a backport branch is a prime example of where this is needed to pin through the pinned tiledb-r to a core version

johnkerl:
FYI the commit
https://github.com/single-cell-data/TileDB-SOMA/pull/2012/commits/9a30a5b5157f184beb4c57f25f346ff422bf7c1d
is unrelated to this PR, but addresses a merge issue. I backported three PRs today. In `_dataframe.py` there was an
```
if update_vals:
```
where `update_vals` is an `np.ndarray`. Note that Python lists can be truthy, but `np.ndarray` should not:
```
>>> if []: print(0)

>>> if np.asarray([]): print(0)
...
<stdin>:1: DeprecationWarning: The truth value of an empty array is ambiguous. Returning False, but in future this will result in an error. Use `array.size > 0` to check that an array is not empty.
```
Note that `main` has `if len(update_vals) > 0:`, so I did the same here.

PR #2019, opened 2024-01-17
[c++] Append `TILEDBSOMA_COVERAGE` to cmake flags

nguyenv:
**Issue and/or context:**

This issue was reported in https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/71 by @jdblischak.

**Changes:**

Previously the `CMAKE_C{,XX}_FLAGS` variable was being replaced by `$ENV{TILEDBSOMA_COVERAGE}`, causing issues when compiling in the [tiledbsoma-feedstock](https://github.com/TileDB-Inc/tiledbsoma-feedstock) nightly builds. The change appends this variable to any existing flags instead of replacing it.


eddelbuettel (approved):
Nice catch!

PR #2026, opened 2024-01-19
[python] Chunked writes for DataFrame; byte-caps for Arrow-table writes

johnkerl:
**Issue and/or context:** Implement chunked writes for `DataFrame` in addition to `NDArray`. Along the way, offer a centralized nbytes cap for remote object stores.

**Notes for Reviewer:**

* For `SparseNDArray`, even before this PR, COO cell sizes are already capped at 24 bytes: 8 each for `soma_dim_0` and `soma_dim_1`, and 1,2,4,8 for `soma_data` depending on the cell-value type (8 bytes for `float64` and `int64`). This means that before this PR, given `goal_chunk_nnz = 100_000_000` times cell cap of 24 bytes, we already had a hard cap of 2_400_000_000 bytes per chunked write. Now we do explicit byte arithmetic, which is a logging/visibility improvement.
* For `DenseNDArray` aside from `uns`, the same is true but cell sizes are capped at 8 not 24 since we don't offer IJV triples on the data writes. That means that before this PR, we already had a hard cap of 800_000_000 bytes per chunked write. Here, too, using explicit byte arithmetic, we obtain a logging/visibility improvement.
* For `DenseNDArray` within `uns`, we were not using chunked-write logic. Now we are.
* For `DataFrame`, also we were not using chunked-write logic. Now we are.

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -1165,6 +1165,30 @@ def _extract_new_values_for_append(
         return arrow_table
 
 
+def _write_arrow_table(
+    arrow_table: pa.Table,
+    handle: Union[DataFrame, SparseNDArray],
+    tiledb_create_options: TileDBCreateOptions,
+) -> None:
+    """Handles num-bytes capacity for remote object stores."""
+    cap = tiledb_create_options._remote_cap_nbytes
+    if arrow_table.nbytes > cap:
+        n = len(arrow_table)
+        if n < 2:
+            raise SOMAError(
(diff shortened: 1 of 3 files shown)


johnkerl:
Note: CI fail is Python 3.10 only, caused by SciPy 1.12, unrelated, and addressed in #2038

johnkerl:
@ihnorton @aaronwolen ping 🙏

aaronwolen:
Looks good to me but I'd like to have @ihnorton approve.

johnkerl:
@ihnorton ping 🙏

johnkerl:
@ihnorton ping 🙏

ihnorton (approved):

PR #2027, opened 2024-01-19
[r] Enable opt-in use of 'catchsegv', enable in CI on Linux

eddelbuettel:
**Issue and/or context:**

Sporadic and generally non-reproducible segmentation faults in CI have beeen an irritant. This PR enables support of a glibc feature (apparently now deprecated, but still supported by Linux distros) to wrap a script `catchsegv` around an execution leading to a stacktrace and register dump in case of segmentation fault.

**Changes:**

The `r-ci.sh` script has been altered to, if `catchsegv` is opted into and on Linux, install the `glibc-tools` package and prefix both coverage and test calls with `catchsegv`.

**Notes for Reviewer:**

[SC 39492](https://app.shortcut.com/tiledb-inc/story/39492/add-libsegfault-backtrace-dumping-to-soma-ci)

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -15,7 +15,8 @@ env:
   COVERAGE_FLAGS: "r"
   COVERAGE_TOKEN: ${{ secrets.CODECOV_TOKEN }}
   _R_CHECK_TESTS_NLINES_: 0
-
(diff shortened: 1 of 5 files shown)


johnkerl (approved):
🚢

Issue #2015, opened 2024-01-16
[python] Save original index names for `tiledbsoma.io`


johnkerl:
As discussed in Slack, we should store the original index name as dataframe metadata. It harms nothing to do so. And on outgest, we can use the name from the array metadata if it's present.

Already we have this:
https://github.com/single-cell-data/TileDB-SOMA/blob/1.5.3/apis/python/src/tiledbsoma/io/ingest.py#L2687-L2696
so if the user asks for `obs_id_name="something_else"` on outgest we will still respect that. And if they don't, we can consult the array metadata. And if that is also missing we can use the default "obs_id" / "var_id" as we already do. So I think this is entirely backward compatible, and an improvement.

johnkerl:
[sc-39271]

PR #2028, opened 2024-01-19
[python] Save off ingest-time obs/var index names for use at outgest

johnkerl:
**Issue and/or context:** #2015

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -2845,6 +2845,7 @@ def to_anndata(
     # * If the desired names are passed in, use them.
     # * Else if the names used at ingest time are available, use them.
     # * Else use the default/fallback name.
+
     if obs_id_name is None:
         if _DATAFRAME_ORIGINAL_INDEX_NAME in experiment.obs.metadata:
             if (
@@ -2873,6 +2874,24 @@ def to_anndata(
                 f"requested obs IDs column name {obs_id_name} not found in input: {obs_df.keys()}"
             )
         obs_df.set_index(obs_id_name, inplace=True)
+    else:
+        # There are multiple cases to be handled here, all tested in CI.
+        # This else-block handle this one:
+        #
+        #                 orig.ident  nCount_RNA  ...
+        # ATGCCAGAACGACT           0        70.0  ...
+        # CATGGCCTGTGCAT           0        85.0  ...
+        # GAACCTGATGAACC           0        87.0  ...
+        #
+        # Namely:
+        # * The input AnnData dataframe had an index with no name
+        # * In the SOMA experiment we name that column "obs_id" and our index is "soma_joinid"
+        # * On outgest we drop "soma_joinid"
+        # * The thing we named "obs_id" needs to become the index again ...
+        # * ... and it needs to be nameless.
+        if "obs_id" in obs_df:
+            obs_df.set_index("obs_id", inplace=True)
+            obs_df.index.name = None
 
     var_df = measurement.var.read().concat().to_pandas()
 
@@ -2883,6 +2902,10 @@ def to_anndata(
                 f"requested var IDs column name {var_id_name} not found in input: {var_df.keys()}"
             )
         var_df.set_index(var_id_name, inplace=True)
+    else:
+        if "var_id" in var_df:
+            var_df.set_index("var_id", inplace=True)
+            var_df.index.name = None
 
     nobs = len(obs_df.index)
     nvar = len(var_df.index)
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -9,6 +9,7 @@ This module contains methods to generate SOMA artifacts starting from
 other formats. Currently only ``.h5ad`` (`AnnData <https://anndata.readthedocs.io/>`_) is supported.
 """
 
+import json
 import math
 import time
 from typing import (
@@ -91,7 +92,6 @@ _UNS_OUTGEST_COLUMN_PREFIX_2D = "values_"
 
 _TILEDBSOMA_TYPE = "soma_tiledbsoma_type"
 _DATAFRAME_ORIGINAL_INDEX_NAME = "soma_dataframe_original_index_name"
-_DATAFRAME_ORIGINAL_INDEX_NAME_IS_NONE = "soma_original_index_name_is_python_None"
 
 
 # ----------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


johnkerl:
@aaronwolen @mojaveazure ping 🙏

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1253,6 +1268,15 @@ def _write_dataframe_impl(
         return soma_df
 
     soma_df.write(arrow_table)
+
+    # Save the original index name for outgest
+    if original_index_name is None:
+        soma_df.metadata[
+            _DATAFRAME_ORIGINAL_INDEX_NAME
+        ] = _DATAFRAME_ORIGINAL_INDEX_NAME_IS_NONE
```
thetorpedodog, line 1276: I thought we could put `null`/`None` values into metadata entries. If that’s not true, it might be better to make this value a JSON-encoded value, so `soma_df.metadata[...] = json.dumps(original_index_name)`; then when loading it back up, we don’t have any special-casing; it can just be `json.loads(soma_df.metadata[...])`. Alternately maybe set a special key, like `soma_df.metadata[_..._IS_NONE] = True` (and maybe set `soma_df.metadata[_..._NAME] = ''`) so that there’s not a special value of this key that we have to watch out for.

thetorpedodog (commented):
I have a suggestion for how you might implement storing the name in metadata, but apart from that this looks pretty good after a quick review.

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: @thetorpedodog thanks.

Whether or not Python `None` writes into metadata, I don't want to know -- I want this language-neutral as much as possible.

I really like the `json.dumps`/`json.loads` approach. :)

johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/2028/commits/da57874e0ef7ce3ec97e4973a5c9ca64851474f2

johnkerl: @thetorpedodog also, experimentally:

```
>>> exp=tiledbsoma.open('foo', 'w')

>>> exp.metadata['hm']=None
  File "/Users/johnkerl/git/single-cell-data/TileDB-SOMA/apis/python/src/tiledbsoma/_tdb_handles.py", line 493, in __setitem__
    _check_metadata_type(key, value)
  File "/Users/johnkerl/git/single-cell-data/TileDB-SOMA/apis/python/src/tiledbsoma/_tdb_handles.py", line 543, in _check_metadata_type
    raise TypeError(f"cannot store {type(obj)} instance as metadata")
TypeError: cannot store <class 'NoneType'> instance as metadata
```

johnkerl:
@thetorpedodog @aaronwolen ping 🙏

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -2650,13 +2670,17 @@ def to_h5ad(
...
     obsm_varm_width_hints: Optional[Dict[str, Dict[str, int]]] = None,
     uns_keys: Optional[Sequence[str]] = None,
 ) -> None:
     """Converts the experiment group to `AnnData <https://anndata.readthedocs.io/>`_
-    format and writes it to the specified ``.h5ad`` file. Arguments are as in ``to_anndata``.
+    format and writes it to the specified ``.h5ad`` file.
+
+    Arguments are as in ``to_anndata``.
+
+    TO DO: doc more params
```
aaronwolen, line 2683: Can we we create an issue/story and remove this from user-facing docs?

aaronwolen (approved):
One minor comment re docs but otherwise really nice.

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/issues/2071

https://github.com/single-cell-data/TileDB-SOMA/pull/2028/commits/896cbc0d4118436d2e89ec2d9e57dc13a1c4258a

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -89,6 +90,9 @@
 _UNS_OUTGEST_COLUMN_NAME_1D = "values"
 _UNS_OUTGEST_COLUMN_PREFIX_2D = "values_"
 
+_TILEDBSOMA_TYPE = "soma_tiledbsoma_type"
+_DATAFRAME_ORIGINAL_INDEX_NAME = "soma_dataframe_original_index_name"
```
thetorpedodog, line 94: maybe suffix this with `_json` to indicate that it is JSON-formatted data

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -2801,21 +2835,71 @@ def to_anndata(
...
+    if obs_id_name is None:
+        # Restore the original index name for outgest. We use JSON for elegant indication of index
+        # name being None (in Python anyway).
+        if _DATAFRAME_ORIGINAL_INDEX_NAME in experiment.obs.metadata:
+            # Maybe 'null' which maps to Pyhton None
+            obs_id_name = json.loads(
+                experiment.obs.metadata[_DATAFRAME_ORIGINAL_INDEX_NAME]
+            )
+        else:
+            obs_id_name = "obs_id"
```
thetorpedodog, line 2852: to avoid the need for the `in` check, you can `.get` with the default value we want:

```python
if obs_id_name is None:
  obs_id_name = json.loads(experiment.obs.metadata.get(_DATAFRAME_ORIGINAL_INDEX_NAME, '"obs_id"'))
```

(or, to make it the ultimate one-liner, `obs_id_name = obs_id_name or json.loads(...)`)

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -2801,21 +2835,71 @@ def to_anndata(
...
+        #
+        # Namely:
+        # * The input AnnData dataframe had an index with no name
+        # * In the SOMA experiment we name that column "obs_id" and our index is "soma_joinid"
+        # * On outgest we drop "soma_joinid"
+        # * The thing we named "obs_id" needs to become the index again ...
+        # * ... and it needs to be nameless.
+        if "obs_id" in obs_df:
+            obs_df.set_index("obs_id", inplace=True)
+            obs_df.index.name = None
```
thetorpedodog, line 2888: does it make sense to fold this `if` in to become an `elif` above?

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -2801,21 +2835,71 @@ def to_anndata(
...
-        )
-    var_df.set_index(var_id_name, inplace=True)
+    if var_id_name is not None:
+        if var_id_name not in var_df.keys():
+            raise ValueError(
+                f"requested var IDs column name {var_id_name} not found in input: {var_df.keys()}"
+            )
+        var_df.set_index(var_id_name, inplace=True)
+    else:
+        if "var_id" in var_df:
```
thetorpedodog, line 2900: ibid.

thetorpedodog (approved):
couple minor things; otherwise looks good overall!

apis/python/src/tiledbsoma/io/ingest.py
thetorpedodog: (adding `_json` to the metadata key, so it’s obvious to an otherwise-unfamiliar reader of the dataset)

Issue #2029, opened 2024-01-20
[Feature request] Update documentation on the version of C++ compiler to install `tiledbsoma` R package


CharlesBordet:
### Is your feature request related to a problem? Please describe.

Edit: I forgot to mention that this is related on the R package. The documentation I mention is `apis/r/README.md`

I ran into issues installing `tiledbsoma` on Amazon Linux 2. The documentation mentions requirements such as `cmake` and `git`, and I found that a recent enough version of a C++ compiler is also necessary.

Amazon Linux 2 comes with `gcc` version 7.3.1.

Installing `tiledbsoma` fails at the step `Building CXX object src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_dataframe.cc.o` with the following error message:

```
/tmp/Rtmp1MSC8v/R.INSTALL375e29e92bdb/tiledbsoma/src/libtiledbsoma/src/soma/soma_dataframe.cc:33:10: fatal error: filesystem: No such file or directory
 #include <filesystem>
          ^~~~~~~~~~~~
compilation terminated.
```

I found that using `#include <filesystem>` is not possible with `gcc` version 7. I upgraded it to version 8, which resolved this issue, but met another one later during the installation:

```
g++ -std=gnu++17 -I"/opt/R/4.1.3/lib64/R/include" -DNDEBUG -I. -I../inst/include/ -I../inst/tiledb/include -I../inst/tiledbsoma/include   -I'/home/user/project/renv/library/R-4.1/x86_64-pc-linux-gnu/Rcpp/include' -I'/home/user/project/renv/library/R-4.1/x86_64-pc-linux-gnu/RcppSpdlog/include' -I'/home/user/project/renv/library/R-4.1/x86_64-pc-linux-gnu/RcppInt64/include' -I/usr/local/include   -fpic  -g -O2  -c RcppExports.cpp -o RcppExports.o
In file included from RcppExports.cpp:4:0:
../inst/include/tiledbsoma_types.h:21:10: fatal error: tiledbsoma/tiledbsoma: No such file or directory
 #include <tiledbsoma/tiledbsoma>
          ^~~~~~~~~~~~~~~~~~~~~~~
```

This time I upgraded to `gcc` version 10.5.0, and it finally worked.

It took me quite some time to figure out what the issue was and how to resolve it. It would be great if the documentation states the version of `gcc` that is required.

### Describe the solution you'd like

[...]

PR #2031, opened 2024-01-20
[r] Update documentation on g++ requirement for R package

CharlesBordet:
**Issue and/or context:**

Issue #2029 on updating documentation to mention the version of C++ compiler to install the R package

**Changes:**

Update `aps/r/README.md` 

**Notes for Reviewer:**

N/A

### DIFF
--- a/apis/r/README.md
+++ b/apis/r/README.md
@@ -14,7 +14,7 @@ install.packages('tiledbsoma', repos = c('https://tiledb-inc.r-universe.dev',
                                          'https://cloud.r-project.org'))
 ```
 
-Installing from source on Unix-like platforms requires `cmake` and `git`.
+Installing from source on Unix-like platforms requires `cmake`, `git`, and a recent version of a C++ compiler (it was tested on `g++` version 10.5.0).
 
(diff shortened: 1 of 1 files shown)


apis/r/README.md
```diff
@@ -14,7 +14,7 @@ install.packages('tiledbsoma', repos = c('https://tiledb-inc.r-universe.dev',
                                          'https://cloud.r-project.org'))
 ```
 
-Installing from source on Unix-like platforms requires `cmake` and `git`.
+Installing from source on Unix-like platforms requires `cmake`, `git`, and a recent version of a C++ compiler (it was tested on `g++` version 10.5.0).
```
eddelbuettel, line 17: Very nice.   Maybe reword to 

```suggestion
Installing from source on Unix-like platforms requires `cmake`, `git`, and a recent-enough C++ compiler (`g++` version 10.* or above).
```

We may want to add the `clang++` version too but I am unsure which it is. Maybe `clang++-16` ?

CharlesBordet:
Thanks for the suggestion, I have committed your rewriting.

About clang++: I don't know what version is required, I didn't have to install it specifically, nor can I actually find it (`clang` or `clang++` is not found).

eddelbuettel (approved):
That looks good to me!  

Thank you for helping to make our documentation better!

Issue #2038, opened 2024-01-22
[python] Latest SciPy 1.12 and Pandas 2.2.0 releases causing Python CI failures


nguyenv:
The latest releases [Pandas 2.2.0](https://pandas.pydata.org/docs/dev/whatsnew/v2.2.0.html) and [SciPy 1.12.0](https://pypi.org/project/scipy), released Jan 19 and Jan 20 respectively, are causing the following CI failures:

```
FAILED apis/python/tests/test_dataframe.py::test_extend_enumerations - AssertionError: assert CategoricalDtype(categories=['A', 'B'], ordered=False, categories_dtype=object) == CategoricalDtype(categories=[b'A', b'B'], ordered=False, categories_dtype=object)
 +  where CategoricalDtype(categories=['A', 'B'], ordered=False, categories_dtype=object) = 0    A\n1    B\n2    A\n3    B\n4    B\n5    B\nName: byte, dtype: category\nCategories (2, object): ['A', 'B'].dtype
 +  and   CategoricalDtype(categories=[b'A', b'B'], ordered=False, categories_dtype=object) = 0    b'A'\n1    b'B'\n2    b'A'\n3    b'B'\n4    b'B'\n5    b'B'\nName: byte, dtype: category\nCategories (2, object): [b'A', b'B'].dtype
FAILED apis/python/tests/test_experiment_query.py::test_experiment_query_all[101-11-X_layer_names0] - TypeError: iteration over a 0-d array
FAILED apis/python/tests/test_sparse_nd_array.py::test_blockwise_scipy_iter_eager[0.01-shape0] - TypeError: iteration over a 0-d array
```

This only affects Python 3.10, which runs on the latest versions. Python 3.7 remain unaffected as it uses older versions of  Pandas and Scipy.

PR #2039, opened 2024-01-22
[python] Update unit tests for pandas 2.2 and scipy 1.12

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2038

**Changes:**

* `scipy.sparse.vstack` no longer takes in generator argument directly - relevant unit tests have been replaced with list comprehension semantics
* Temporarily compare categorical string-like dtypes by `kind` to get CI passing - bytes are being output as str and needs to be fixed in tiledb-py, not tiledb-soma


thetorpedodog (approved):
weird that they made that change to `sparse.vstack`.

PR #2049, opened 2024-01-23
[python] Pass config options to `DataFrame.read`

nguyenv:
**Issue and/or context:**

As reported internally, the code below breaks as the AWS S3 settings are ignored.

```
uri = "s3://cellxgene-census-public-us-west-2/cell-census/2024-01-16/soma/"

with soma.open(
    uri,
    context=soma.SOMATileDBContext(
        tiledb_config={
            "vfs.s3.no_sign_request": "true",
            "vfs.s3.region": "us-west-2",
        }
    ),
) as census:
    census["census_data"]["homo_sapiens"].obs.read()

```

```
SOMAError: Error opening array: 's3://cellxgene-census-public-us-west-2/cell-census/2024-01-16/soma/census_data/homo_sapiens/obs'
  [TileDB::Array] Error: Caught std::exception: [TileDB::S3] Error: Error while listing with prefix 's3://cellxgene-census-public-us-west-2/cell-census/2024-01-16/soma/census_data/homo_sapiens/obs/' and delimiter '/'[Error Type: 100] [HTTP Response Code: 301] [Exception: PermanentRedirect] [Remote IP: 54.231.161.178] [Request ID: 9630132PVWN1ZHJ4] [Headers: 'content-type' = 'application/xml' 'date' = 'Mon, 22 Jan 2024 21:43:18 GMT' 'server' = 'AmazonS3' 'transfer-encoding' = 'chunked' 'x-amz-bucket-region' = 'us-west-2' 'x-amz-id-2' = '/JFJGX5VnF2DA+Y0nUuv6pjEvg/2YFKxvZjYO7eDF/XRO8VOaIP4LwbI5Bxp8tEOwlO/ym0c8ug=' 'x-amz-request-id' = '9630132PVWN1ZHJ4'] : Unable to parse ExceptionName: PermanentRedirect Message: The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.
```

**Changes:**

Pass the already set `config_options` into `read`, not the default config.

**Notes for Reviewer:**

I am not sure how to add a unit test for this as anything I can think of uses cloud. I tested locally with the code snippet above, and it now works.

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -339,6 +339,9 @@ class DataFrame(TileDBArray, somacore.DataFrame):
         self._check_open_read()
 
         handle = self._handle._handle
+        
+        config = handle.config()
+        config.update(platform_config or {})
 
         ts = None
         if handle.timestamp is not None:
@@ -347,7 +350,7 @@ class DataFrame(TileDBArray, somacore.DataFrame):
         sr = clib.SOMADataFrame.open(
             uri=handle.uri,
             mode=clib.OpenMode.read,
-            platform_config=handle.config().update(platform_config or {}),
+            platform_config=config,
             column_names=column_names or [],
             result_order=_util.to_clib_result_order(result_order),
             timestamp=ts,
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -339,7 +339,7 @@ class DataFrame(TileDBArray, somacore.DataFrame):
         self._check_open_read()
 
         handle = self._handle._handle
-        
+
         ts = None
         if handle.timestamp is not None:
             ts = (0, handle.timestamp)
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -339,7 +339,7 @@ class DataFrame(TileDBArray, somacore.DataFrame):
         self._check_open_read()
 
         handle = self._handle._handle
-
+        
         ts = None
         if handle.timestamp is not None:
             ts = (0, handle.timestamp)
@@ -347,7 +347,7 @@ class DataFrame(TileDBArray, somacore.DataFrame):
         sr = clib.SOMADataFrame.open(
             uri=handle.uri,
             mode=clib.OpenMode.read,
-            platform_config=platform_config or handle.config(),
+            platform_config=handle.config().update(platform_config or {}),
             column_names=column_names or [],
             result_order=_util.to_clib_result_order(result_order),
             timestamp=ts,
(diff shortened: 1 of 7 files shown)


johnkerl (approved):
🚢

johnkerl (commented):
🚢

johnkerl (approved):
🚢

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -338,21 +338,23 @@ def read(
...
-            ts = (0, self._handle._handle.timestamp)
+        if handle.timestamp is not None:
+            ts = (0, handle.timestamp)
 
         sr = clib.SOMADataFrame.open(
-            uri=self._handle._handle.uri,
+            uri=handle.uri,
             mode=clib.OpenMode.read,
-            platform_config=platform_config or {},
+            platform_config=platform_config or handle.config(),
```
bkmartinjr, line 350: shouldn't these be merged?  I'm thinking about the case where:
* handle.config() has the context-provided configuration (from open)
* user provides a platform_config to override something (e.g., buffer size)

I'm essentially suggesting that the clib `platform_config` argument should be:
>  platform_config=handle.config() | (platform_config or {})

bkmartinjr (changes requested):
One small concern about how we handle the case where config is provided _both_ via `open` and `read`, and if they should be merged.  My understanding of the normal TileDB config handling is that they would merge.

apis/python/src/tiledbsoma/_dataframe.py
nguyenv: I think I understand. Do you mean `platform_config = handle.config().update(platform_config or {})`? The `handle.config()` (config that's passed to `open`) gets updated with any new key-value pairs in `platform_config` (config that's passed to `read`) but otherwise retains what's originally there.

bkmartinjr: yep!  The read should use the union of the config, prioritizing the read() argument if specified.

@thetorpedodog - do you concur?

johnkerl:
But honestly, what I _want_ to test is build-for-pkgdown nightly, and so I'm truly happiest with the setup on this PR, which does that:

* build for pkgdown nightly
* build and deploy at release-tag time

That what I want it to do, and it's what it does, and the filenames and CI-job names correctly indicate what's going on.

bkmartinjr (approved):

apis/python/src/tiledbsoma/_dataframe.py
thetorpedodog: > I think I understand. Do you mean `platform_config = handle.config().update(platform_config or {})`? The `handle.config()` (config that's passed to `open`) gets updated with any new key-value pairs in `platform_config` (config that's passed to `read`) but otherwise retains what's originally there.

> yep! The read should use the union of the config, prioritizing the read() argument if specified.

Well, we wouldn’t want to do `.update()` since that would end up permanently changing whatever gets returned from `handle.config()`, but yes, the union with passed-in value taking priority is what I would expect from a call like this.

nguyenv: Thanks for the heads up. I have made a deep copy of the `handle.config()` before updating it. I tried to use `{**handle.config(), **(platform_config or {})}` but mypy was giving me a cryptic typing error.

Issue #2052, opened 2024-01-23
[r/ci] Debug pkgdown CI fail at release-tag time


johnkerl:
# Problem

Tagging
https://github.com/single-cell-data/TileDB-SOMA/releases/tag/1.7.0rc0
triggered
https://github.com/single-cell-data/TileDB-SOMA/blob/5cc41b323d02dc7015e95f4e07d24742b6c14dd3/.github/workflows/pkgdown.yaml
which only runs on releases.

There, we got the compile error
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/7618476508/job/20749733125#step:5:916
namely
```
using C compiler: 'gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0'
using C++ compiler: 'g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0'
using C++17
g++ -std=gnu++17 -I"/opt/R/4.3.2/lib/R/include" -DNDEBUG -I. -I../inst/include/ -I../inst/tiledb/include -I../inst/tiledbsoma/include   -I'/home/runner/work/_temp/Library/Rcpp/include' -I'/home/runner/work/_temp/Library/RcppSpdlog/include' -I'/home/runner/work/_temp/Library/RcppInt64/include' -I/usr/local/include    -fpic  -g -O2  -Wall -pedantic -c RcppExports.cpp -o RcppExports.o
In file included from RcppExports.cpp:4:
../inst/include/tiledbsoma_types.h:21:10: fatal error: tiledbsoma/tiledbsoma: No such file or directory
   21 | #include <tiledbsoma/tiledbsoma>
      |          ^~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make: *** [/opt/R/4.3.2/lib/R/etc/Makeconf:200: RcppExports.o] Error 1
ERROR: compilation failed for package 'tiledbsoma'
* removing '/tmp/RtmpU14Loy/Rinst1b795ff84a8f/tiledbsoma'
      -----------------------------------
ERROR: package installation failed
```

This is odd for several reasons:

* We build from source in our other CIs all the time
* It's not at all clear why this path is using `-Werror=pedantic`:
  * `.github/workflows/pkgdown.yaml`
  * `./apis/r/tools/install-pkgdown-dependencies.sh`
  * `Rscript -e 'remotes::install_local()'`
[...]

PR #2053, opened 2024-01-23
[r/ci] Run pkgdown build, but not deploy, in nightly cron

johnkerl:
**Issue and/or context:** As described in #2052.

**Changes:** Run the install nightly, but not the deploy, as the install surfaces a `libtiledbsoma` build path that we want to flex regularly, without waiting for it to fail at release-tag time.

**Notes for Reviewer:**

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -1,5 +1,6 @@
 # Workflow derived from https://github.com/r-lib/actions/tree/master/examples
 # Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
+
 on:
   #push:
   #  # To publish docs from your branch: list the branch name here instead of main.
@@ -9,6 +10,8 @@ on:
   #  branches: [main]
   release:
     types: [published]
+  schedule:
+    - cron: "42 9 * * *"
   workflow_dispatch:
 
 name: pkgdown
@@ -31,8 +34,12 @@ jobs:
           use-public-rspm: true
           working-directory: "apis/r"
 
+      # Run this daily to surface errors as tracked at
+      # https://github.com/single-cell-data/TileDB-SOMA/issues/2052
       - name: Install dependencies
         run: ./apis/r/tools/install-pkgdown-dependencies.sh
 
+      # Run this on releases, or on workflow dispatch
       - name: Deploy package
+        if: github.event_name != 'schedule'
         run: ./apis/r/tools/deploy-pkgdown.sh


eddelbuettel:
Unless I am mistaken this would overwrite documentation for the released version with documentation for the in-development unreleased version which is a change in behaviour.

You could also rewrite the CI to use the same steps the existing nightly job and get the same effect.

johnkerl:
@eddelbuettel on this PR we just install packages nightly. The deploy is not done nightly.

johnkerl:
@eddelbuettel if by the "existing one" you mean `.github/workflows/r-valgrind.yaml`, then sure, but only if we rename that YAML file. (This build-for-pkgdown is _not_ valgrind.)

johnkerl:
But honestly, what I _want_ to test is build-for-pkgdown nightly, and so I'm truly happiest with the setup on this PR, which does that:

* build for pkgdown nightly
* build and deploy at release-tag time

This PR as-is does what I want, and the filenames and CI-job names correctly indicate what's going on.

eddelbuettel:
Sure, renaming to `r-nightly.yaml` works.

What I am pointing here, possibly somewhat indirectly, is that the older scripts in apis/r/tools/ used here were most likely part of the fragility we observed, and I think we would be better off if we had fewer rather than more moving parts where we have to chase installation and package version details.  As best as I can tell, we had no issue with `pkgdown`.  We had an issue with our setups which fell over semicolons in one setup (still not quite sure why...  and the semicolons are gone now so that is a plus) but not another.

But I have no bone in this fight.  If you feel this is the way to go feel free to merge.

johnkerl:
Thanks @eddelbuettel . I get the fragility.

That's precisely one of my concerns. With the semicolon issue in place I _did_ repro a fail, and I got that repro with the current setup, with `remotes::install_local()` precisely as articulated here. This _was_ a bug-catcher that caught a bug (once it was permitted to). And I want to preserve and value that.

I appreciate the dialog, and your offer to merge, which I accept.

Thanks again!

johnkerl:
> But I have no bone in this fight. If you feel this is the way to go feel free to merge.

@eddelbuettel happy to accept the offer; lacking an approval. (Perhaps @aaronwolen or @mojaveazure can help.)

eddelbuettel:
I am fairly certain that the frailty came from scripts that we ... don't need or use in other settings apart from this.  If we remove them, then there is no need to add another nightly job.  `pkgdown` by itself works.   :man_shrugging:

mojaveazure (approved):

PR #2055, opened 2024-01-23
[python/c++] Debug CI fails for #2049

johnkerl:
**Issue and/or context:** #2049 is failing CI, on MacOS only. Meanwhile i cannot reproduce the failuure on my own Mac. So, I'm copying the PR, bare-bonesing the test cases, and adding some debug visibility in an attempt to hone in on the problem this way.

**Changes:**

**Notes for Reviewer:**


johnkerl:
#2049 issued is solved on #2049

PR #2060, opened 2024-01-25
[c++] Fix reindexer for when workload is small and threadpool is large

johnkerl:
**Issue and/or context:** As discussed in Slack.

@bkmartinjr repro:

```
import numpy as np
import tiledbsoma as soma
from tiledbsoma import pytiledbsoma as clib

reindexer = clib.IntIndexer()
keys = np.arange(0, 10000, dtype=np.int64)
reindexer.map_locations(keys, 64)
ids = np.array([525, 1293, 1805, 5802, 7636, 7754, 7791, 7957, 7959, 8067, 8340, 8736, 8806, 9329,9377, 9653], dtype=np.int64)

# this hangs
reindexer.get_indexer(ids)
```

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/tests/test_indexer.py
+++ b/apis/python/tests/test_indexer.py
@@ -1,6 +1,7 @@
 import numpy as np
 import pandas as pd
 
+from tiledbsoma import pytiledbsoma as clib
 from tiledbsoma._index_util import build_index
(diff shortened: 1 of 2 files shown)


johnkerl:
Duplicate of #2061 by @beroy

PR #2061, opened 2024-01-25
[C++] Fix zero-size lookup chunk size for the C++ re-indexer

beroy:
**Issue and/or context:**
At lookup time, if chunk size (unit work size) is computed to zero (while happens when the concurrency level is larger than lookup array size)

**Changes:**
Fix the chunk size to be reset to 1 if it is calculated to zero. Also proper unittests were added.

**Notes for Reviewer:**


johnkerl (approved):

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -111,6 +111,9 @@ void IntIndexer::lookup(const int64_t* keys, int64_t* results, int size) {
     std::vector<tiledbsoma::ThreadPool::Task> tasks;
 
     size_t thread_chunk_size = size / tiledb_thread_pool_->concurrency_level();
+    if (thread_chunk_size == 0) {
```
bkmartinjr, line 114: style nit: suggest std::max(1, ...)

PR #2074, opened 2024-01-29
Needed changes for census profiling

beroy:
Including reporting script for detecting performance violations

**Issue and/or context:**
Currently census code does not go through performance checks


profile_report.py
```diff
@@ -0,0 +1,31 @@
...
+c = 0
+
+if len(dt) >= 2:
+    exec_time = [0.0, 1.0]
+    last_two = dt[-2:]
+    exec_time[0] = last_two[0].user_time_sec
+    exec_time[1] = last_two[1].user_time_sec
+
+    if threshold * float(exec_time[1]) < float(exec_time[0]) or float(exec_time[1]) > threshold * float(exec_time[0]):
+        raise SystemExit(f"Potential performance degradation detected {exec_time[0]} vs {exec_time[1]}")
```
atolopko-czi, line 26: If we allow for 10% degradation every time, we'll soon have unacceptable performance. Might we instead alert when a 10% degradation is detected compared to the _first_ benchmark?  If that makes sense, perhaps a larger degradation threshold is allowable.

Also, are the timings always for the same input data? If the data changes, we can't meaningfully compare the performance.

profile_report.py
```diff
@@ -0,0 +1,31 @@
...
+
+parser = argparse.ArgumentParser()
+parser.add_argument("benchmark", type=str)
+parser.add_argument("path", type=str)
+
+args = parser.parse_args()
+
+# Processes the set of previously written logs
+
+threshold = 1.10  # Percent difference
```
atolopko-czi, line 12:
```suggestion
# The threshold (ratio) of allowable performance degradation between profiling runs
threshold = 1.10 
```

profile_report.py
```diff
@@ -0,0 +1,31 @@
...
+dt = db.find(f"{args.benchmark}")
+c = 0
+
+if len(dt) >= 2:
+    exec_time = [0.0, 1.0]
+    last_two = dt[-2:]
+    exec_time[0] = last_two[0].user_time_sec
+    exec_time[1] = last_two[1].user_time_sec
+
+    if threshold * float(exec_time[1]) < float(exec_time[0]) or float(exec_time[1]) > threshold * float(exec_time[0]):
```
atolopko-czi, line 25: Would alert on these two conditions separately, with specific messages. The first condition is not a "degradation", of course.

profile_report.py
```diff
@@ -0,0 +1,31 @@
...
+        raise SystemExit(f"Potential performance degradation detected {exec_time[0]} vs {exec_time[1]}")
+    print(f"Prev vs Curr execution time {exec_time[0]} vs {exec_time[1]}")
+    print("No recent performance degradation detected")
+    print(
+        f"TileDB version prev = {dt[0].tiledbsoma_version} curr = {dt[1].tiledbsoma_version}"
+    )
```
atolopko-czi, line 31: might want to output this before the check, since it's relevant to the user in the case of an alert

profile_report.py
```diff
@@ -0,0 +1,31 @@
...
+parser.add_argument("path", type=str)
+
+args = parser.parse_args()
+
+# Processes the set of previously written logs
+
+threshold = 1.10  # Percent difference
+
+db = data.FileBasedProfileDB(args.path)
+actual_max_ts = 0
```
atolopko-czi, line 15: unused

profile_report.py
```diff
@@ -0,0 +1,31 @@
...
+dt = db.find(f"{args.benchmark}")
+c = 0
+
+if len(dt) >= 2:
+    exec_time = [0.0, 1.0]
+    last_two = dt[-2:]
+    exec_time[0] = last_two[0].user_time_sec
+    exec_time[1] = last_two[1].user_time_sec
```
atolopko-czi, line 23: more idiomatic as:
```suggestion
exec_times = dt[-2].user_time_sec, dt[-1].user_time_sec
```
(note the pluralization, which would require edits below)

profile_report.py
```diff
@@ -0,0 +1,31 @@
+from profiler import data
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument("benchmark", type=str)
```
atolopko-czi, line 5: Would rename `benchmark` to `command`, if only to keep the terminology consistent with the rest of the Profiler code. Consider the signature: `FileBaseProfileDB.find(command: str)`

profile_report.py
```diff
@@ -0,0 +1,31 @@
...
+parser.add_argument("path", type=str)
```
atolopko-czi, line 6: maybe `db-path`?

profile_report.py
```diff
@@ -0,0 +1,31 @@
+from profiler import data
```
atolopko-czi, line 1: Add a comment explaining the purpose and usage of the script.

profile_report.py
atolopko-czi, line 1: move this into `profiler` dir?

profile_report.py
atolopko-czi, line 26: When a performance problem is detected, is there more information that could be output which is useful to the reader? Maybe just print out both of the `ProfileData` objects, since that's the probably the very next thing the reader would want to see.

profile_report.py
atolopko-czi: Also, there is already a `report.py` module. Since this script is actually about detecting performance regressions, perhaps name like `perf_checker.py`

Also, would this be worth moving to the `tools/perf_checker` dir?

profiler/__main__.py
```diff
@@ -1,4 +1,4 @@
-from .profiler import main
+import profiler
```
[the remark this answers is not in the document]
beroy: No, the relative imports caused trouble actually. I could not figure the exact issue but managed to fix it this way.

profile_report.py
beroy: I see your point about the 10% and also believe 10% compared to the first performance point makes more sense.

That's very true. I am not really sure how to handle the changes of data. I think that needs some brainstorming. One option is to consider new data as a new benchmark or reset the benchmark data when data changes significantly. However, none of these can be automated. What do you think?

profile_report.py
beroy: Sounds good

profile_report.py
beroy: Great suggestion

profiler/src/profiler/profiler.py
```diff
@@ -183,12 +191,13 @@ def main():
             print(
                 f"Third profiler {args.prof2} missing output flamegraph file location"
             )
-
+    args.command.split(" ")
```
ebezzi, line 194: This line looks like a no-op.

profiler/src/profiler/profiler.py
```diff
@@ -183,12 +191,13 @@ def main():
...
     p_stdout, p_stderr = p.communicate()
     if p1 is not None:
         p1.wait()
     if p2 is not None:
         p2.wait()
+    print("Done profiler run", file=stderr)
```
ebezzi, line 200:
```suggestion
    print("Profiler run done", file=stderr)
```

ebezzi (approved):

Issue #2067, opened 2024-01-26
[python] add public API for new reindexer


bkmartinjr:
The new indexer should be exposed via a public API, allowing downstream consumers (eg., cellxgene-census package) to migrate from the Pandas indexer. Suggest that this is a `tiledbsoma` specific API, not a generic part of SOMA.

johnkerl:
@bkmartinjr do you have thoughtst on the right venue for API proposals? As a starting point I'd suggest you offer a sketch of cellxgene-census needs, and we can iterate from there ... either here, or in a GDoc, or maybe a call next week ...

bkmartinjr:
All we need is public wrappers for a drop-in equivalent of `pandas.Index()` and `get_indexer()` (i.e., indexer factory and the get_indexer method).  So I propose something super simple like (caution: typed in without running/testing):

```
from _index_util import build_index

class SOMATileDBIndexer:
    def __init__(keys: np.typing.NDArray[np.int64], context: SOMATileDBContext):
        self._index = build_index(keys, context)

    def get_indexer(target: npt.NDArray[np.int64]) -> Any:
        return self._index.get_indexer(target)
```

We could have this subclass `somacore.query.types.IndexLike` if we want to expose that protocol,but I don't think it is entirely necessary (I defer to @thetorpedodog )

thetorpedodog:
I don’t think there’s a need to make this its own class—`build_index` works as a factory for an `IndexLike` object. Exporting that function would suffice (which I think was one of the reasons it got put in a semi–public-facing namespace called `utils.py` originally).

thetorpedodog:
Naming I’m not sure about but I think specifying the return as `IndexLike` (rather than `clib.IntIndexer`) would be helpful in terms of just making the intent obvious. There’s nothing that has to change on the somacore side, since `IndexLike` is already in a public namespace and doesn’t really need a top-level export (it’s not something end users will ever likely explicitly use directly; only _maybe_ see as a parameter or return type).

bkmartinjr:
Re: IndexLike - sounds fine.

Naming:   propose  `tiledbsoma.build_index` as I can't think of anything better :-)

@beroy - suggest we just get this into a PR, add to the top-level namespace, and we can hash out the actual name in the PR disucssion.

beroy:
https://github.com/single-cell-data/TileDB-SOMA/pull/2080

PR #2080, opened 2024-01-29
[python] Reindexer external python API

beroy:
**Issue and/or context:** #2067
We need an external python API for the C++ reindexer

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/__init__.py
+++ b/apis/python/src/tiledbsoma/__init__.py
@@ -115,6 +115,7 @@ from ._general_utilities import (
     get_storage_engine,
     show_package_versions,
 )
+from ._index_util import build_index
 from ._measurement import Measurement
 from ._sparse_nd_array import SparseNDArray
 from .options import SOMATileDBContext, TileDBCreateOptions
@@ -130,6 +131,8 @@ __version__ = get_implementation_version()
 __all__ = [
     "AxisColumnNames",
     "AxisQuery",
+    "build_index",
+    "build_index_tiledb_context",
     "Collection",
     "DataFrame",
     "DenseNDArray",
--- a/apis/python/src/tiledbsoma/__init__.py
+++ b/apis/python/src/tiledbsoma/__init__.py
@@ -115,7 +115,7 @@ from ._general_utilities import (
     get_storage_engine,
     show_package_versions,
 )
-from ._index_util import build_index
+from ._index_util import tiledbsoma_build_index
 from ._measurement import Measurement
 from ._sparse_nd_array import SparseNDArray
 from .options import SOMATileDBContext, TileDBCreateOptions
@@ -131,7 +131,7 @@ __version__ = get_implementation_version()
 __all__ = [
     "AxisColumnNames",
     "AxisQuery",
-    "build_index",
+    "tiledbsoma_build_index",
     "Collection",
     "DataFrame",
     "DenseNDArray",
--- a/apis/python/src/tiledbsoma/__init__.py
+++ b/apis/python/src/tiledbsoma/__init__.py
@@ -132,7 +132,6 @@ __all__ = [
     "AxisColumnNames",
     "AxisQuery",
     "build_index",
-    "build_index_tiledb_context",
     "Collection",
     "DataFrame",
     "DenseNDArray",
--- a/apis/python/src/tiledbsoma/_index_util.py
+++ b/apis/python/src/tiledbsoma/_index_util.py
@@ -2,29 +2,40 @@
 This file is separate from _util.py, due to a circular-import issue with
 SOMATileDBContext which would otherwise ensue.
 """
+from __future__ import annotations
 
-from typing import Optional
+from typing import TYPE_CHECKING, Optional
 
 import numpy as np
-import pandas as pd
 from somacore.query.types import IndexLike
 
 from tiledbsoma import pytiledbsoma as clib
 
-from .options import SOMATileDBContext
+if TYPE_CHECKING:
+    from .options import SOMATileDBContext
 
 
 def tiledbsoma_build_index(
     keys: np.typing.NDArray[np.int64],
     *,
-    context: Optional[SOMATileDBContext] = None,
+    context: Optional["SOMATileDBContext"] = None,
     thread_count: int = 4,
 ) -> IndexLike:
-    """Builds an indexer object compatible with :meth:`pd.Index.get_indexer`."""
-    if len(np.unique(keys)) != len(keys):
-        raise pd.errors.InvalidIndexError(
-            "Reindexing only valid with uniquely valued Index objects"
-        )
+    """
+    Returns a IndexLike re-indexer
+    The reindxer has an API similar to :meth:`pd.Index.get_indexer`
+
+    Args:
+        keys:
+           integer keys used to build the index (hash) table.
+       context:
+           tiledbcontext object containing concurrecy level (exclusive with thread_count).
+       thread_count:
+           concurrecy level when the user does not want to use `context`.
+
+    Lifecycle:
+        Experimental.
+    """
     if context is not None:
         tdb_concurrency = int(
             context.tiledb_ctx.config().get("sm.compute_concurrency_level", 10)
--- a/apis/python/src/tiledbsoma/_index_util.py
+++ b/apis/python/src/tiledbsoma/_index_util.py
@@ -11,31 +11,20 @@ from tiledbsoma import pytiledbsoma as clib
 from .options import SOMATileDBContext
 
 
-def build_index_tiledb_context(
-    keys: np.typing.NDArray[np.int64], context: SOMATileDBContext
-) -> clib.IntIndexer:
-    """Builds an indexer object compatible with :meth:`pd.Index.get_indexer`."""
-    if len(np.unique(keys)) != len(keys):
-        raise pd.errors.InvalidIndexError(
-            "Reindexing only valid with uniquely valued Index objects"
-        )
-    tdb_concurrency = int(
-        context.tiledb_ctx.config().get("sm.compute_concurrency_level", 10)
-    )
-    thread_count = tdb_concurrency // 2
-    reindexer = clib.IntIndexer()
-    reindexer.map_locations(keys, thread_count)
-    return reindexer
-
-
 def build_index(
-    keys: np.typing.NDArray[np.int64], thread_count: int = 4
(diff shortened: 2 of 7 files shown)


johnkerl (commented):
LGTM but I defer to @bkmartinjr

apis/python/tests/test_reindexer_api.py
```diff
@@ -0,0 +1,15 @@
+import numpy as np
+
+from tiledbsoma import build_index
+
+
+def main():
```
nguyenv, line 6: Please rename this so that the function begins with `test` (ie. `test_indexer_basic`). This way it can be picked up by pytest automatically. If it appropriate to move this to `text_indexer.py` too then that would preferable as it would be nice to group similar tests together in the same file.

apis/python/src/tiledbsoma/_index_util.py
```diff
@@ -11,7 +11,7 @@
 from .options import SOMATileDBContext
```
nguyenv, line 11: I realize this is not part of this PR, but I only just noticed this. We don't need to have a separate `_index_util.py` to solve this. `SOMATileDBContext` is only imported for typing, and the `typing` library has a special macro to deal with this circular import issue.

Here are some references for `TYPE_CHECKING`:
1. https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING
2. https://mypy.readthedocs.io/en/stable/runtime_troubles.html#typing-type-checking

apis/python/tests/test_reindexer_api.py
bkmartinjr, line 6: shouldn't this be a pytest test?

apis/python/src/tiledbsoma/__init__.py
```diff
@@ -115,6 +115,7 @@
     get_storage_engine,
     show_package_versions,
 )
+from ._index_util import build_index
```
nguyenv, line 118: Are the functions in `_util.py` (which includes `_index_util.py`) supposed to be for internal only? If so, we might want to move these into a `_indexer.py`.

apis/python/src/tiledbsoma/_index_util.py
```diff
@@ -26,3 +26,16 @@ def build_index(
     reindexer = clib.IntIndexer()
     reindexer.map_locations(keys, thread_count)
     return reindexer
+
+
+def build_index(
```
bkmartinjr, line 31: Given that this function is only present in the tiledb variant of SOMA, we should follow the naming convention used for other tiledbsoma-exlusive functions.  Propose `tiledbsoma_build_index()`

bkmartinjr (changes requested):
Two suggestions:
* DRY the build_index function
* rename it to make clear it is a tiledbsoma-specific function

apis/python/src/tiledbsoma/_index_util.py
```diff
@@ -11,7 +11,7 @@
 from .options import SOMATileDBContext
 
 
-def build_index(
+def build_index_tiledb_context(
```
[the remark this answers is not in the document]
thetorpedodog: This is a good idea. Might also make sense to make `context` and `thread_count` mutually exclusive, though your suggested behavior of "use thread count if context is not present otherwise do not care about it" is also reasonable.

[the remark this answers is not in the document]
beroy: Sounds good to me.

apis/python/src/tiledbsoma/__init__.py
beroy: As you mentioned, lets address these in a separate PR if that's ok

beroy:
@bkmartinjr @thetorpedodog the changes you requested are implemented. LMK if all look good. Thanks again.

apis/python/src/tiledbsoma/_index_util.py
```diff
@@ -11,18 +11,20 @@
 from .options import SOMATileDBContext
 
 
-def build_index(
-    keys: np.typing.NDArray[np.int64], context: SOMATileDBContext
+def tiledbsoma_build_index(
+    keys: np.typing.NDArray[np.int64], context: SOMATileDBContext = None, thread_count: int = 4  # type: ignore
```
thetorpedodog, line 15: A few things here:

1. the `type: ignore` here is hiding a real error: `context = None` is not a valid value, since None is not a SOMATileDBContext. Instead, it should be specified as an `Optional[SOMATileDBContext]`.
2. When you start getting into optional arguments, it’s best to use keyword-only arguments, since there is no logic to the order that things might be omitted in (e.g. `tiledbsoma_build_index(some_keys, None, 5)` makes no more or less sense than `tiledbsoma_build_index(some_keys, 5)`
3. (minor) So that users get better type annotation information on `tiledbsoma_build_index`, let’s have its return be annotated as `IndexLike` rather than our internal-only type.

So, overall, this should look like:

```python
from somacore.query.types import IndexLike

def tiledbsoma_build_index(
    keys: np.typing.NDArray[np.int64],
    *,
    context: Optional[SOMATileDBContext] = None,
    thread_count: int = 4,
) -> IndexLike:
```

beroy: All sound good.

beroy: Done

apis/python/src/tiledbsoma/_index_util.py
```diff
@@ -3,26 +3,34 @@
...
+    if context is not None:
+        tdb_concurrency = int(
+            context.tiledb_ctx.config().get("sm.compute_concurrency_level", 10)
+        )
+        thread_count = tdb_concurrency // 2
+
     reindexer = clib.IntIndexer()
     reindexer.map_locations(keys, thread_count)
-    return reindexer
+    return cast(IndexLike, reindexer)
```
thetorpedodog, line 36: You can minimize the amount of errors that will be caught with a type-ignore comment:

```
return reindexer  # type: ignore[no-any-return]
```

In this case I think using this particular version of `type: ignore` is better than using a `cast`, since this ends up being more specific about what should be allowed (and it can be dropped if `clib.IntIndexer` gets type annotation data somehow). `cast` is a lot like a blanket `# type: ignore` for an assignment statement.

It is fine to leave it as-is, though; this is an incredibly minor difference.

thetorpedodog (commented):
Looks good to me! I leave it to somebody more directly on the project to give full approval.

apis/python/src/tiledbsoma/_index_util.py
beroy: sounds good. I'll use this. I did not know `ignore[no-any-return]` exists ;)

apis/python/src/tiledbsoma/_index_util.py
bkmartinjr: Lets add this change to this PR so the later reorganization is simple.  E.g.,

```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from .options import SOMATileDBContext
```

apis/python/src/tiledbsoma/_index_util.py
```diff
@@ -3,26 +3,34 @@
...
-def build_index(
-    keys: np.typing.NDArray[np.int64], context: SOMATileDBContext
-) -> clib.IntIndexer:
+def tiledbsoma_build_index(
+    keys: np.typing.NDArray[np.int64],
+    *,
+    context: Optional[SOMATileDBContext] = None,
+    thread_count: int = 4,
+) -> IndexLike:
     """Builds an indexer object compatible with :meth:`pd.Index.get_indexer`."""
```
bkmartinjr, line 23: nit: docstring needs to define arguments and return

apis/python/src/tiledbsoma/_index_util.py
```diff
@@ -3,26 +3,34 @@
...
     if len(np.unique(keys)) != len(keys):
```
bkmartinjr, line 24: this test can (is) quite expensive for large indices -- np.unique effectively has to build a data structure similar to the index in order to perform the check. For example, on my x66 box, it takes 2-3 seconds to perform this test for 100M values.

Can we just use the index, which we are building, to do the same check?

E.g., if the index is a hash table insertion, where dups overwrite, then:
1. create the index
2. check that the resulting length == the input length

This would remove the duplication of work, which works against the entire point of having a fast indexer ;-)

bkmartinjr (commented):
With one large caveat, this looks good.

The caveat is the performance cost of how the duplicate input key check is performed (currently by a separate call to numpy.unique).  This effectively doubles the cost of creating the indexer, and some trivial benchmarking shows it is a material amount of time for moderate to large keys.

Example:
```
In [14]: a = np.arange(0,100_000_000)

In [15]: %time np.unique(a)
CPU times: user 2.43 s, sys: 590 ms, total: 3.02 s
Wall time: 3.02 s
Out[15]: array([       0,        1,        2, ..., 99999997, 99999998, 99999999])
```

Can we leverage the fact that we are already building a hash table to perform this check as part of the hash table creation?

apis/python/src/tiledbsoma/_index_util.py
beroy: That sounds like a great idea @bkmartinjr. Let me see how quick I can implement that.

apis/python/src/tiledbsoma/_index_util.py
beroy: Done

apis/python/tests/test_reindexer_api.py
```diff
@@ -0,0 +1,22 @@
+import numpy as np
```
bkmartinjr, line 1: please add at least one test that an exception is thrown if there are non-unique keys passed to build_index

bkmartinjr: Oh, NM - I see there is a test in the other file covering this case

apis/python/tests/test_indexer.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
```
bkmartinjr, line 1: absolutely *NOT* required for this PR, but htis file should use pytest conventions, e.g.,  pytest.parametrize, pytest.raises, etc.

beroy: I'll make sure to create a follow up small task for that.

bkmartinjr (approved):

Issue #2066, opened 2024-01-26
[python] Integrate fast-indexer support into blockwise iterator


bkmartinjr:
The blockwise iterator is still using the Pandas indexer. It should be updated to utilize the new reindexer.

beroy:
https://github.com/single-cell-data/TileDB-SOMA/pull/2081

PR #2081, opened 2024-01-29
[python] Using optimized reindexer in blockwise iterators

beroy:
**Issue and/or context:** #2066
Block-wise iterators still use panda indexers and this must to switch to our new indexer.

**Changes:**

**Notes for Reviewer:**
I may add a benchmark for this change later.

### DIFF
--- a/apis/python/src/tiledbsoma/_read_iters.py
+++ b/apis/python/src/tiledbsoma/_read_iters.py
@@ -35,8 +35,9 @@ import tiledbsoma.pytiledbsoma as clib
 
 from . import _util
 from ._exception import SOMAError
-from ._index_util import build_index
+from ._index_util import tiledbsoma_build_index
 from ._types import NTuple
+from .options import SOMATileDBContext
 
 if TYPE_CHECKING:
     from . import SparseNDArray
@@ -88,6 +89,7 @@ class BlockwiseReadIterBase(somacore.ReadIter[_RT], metaclass=abc.ABCMeta):
         reindex_disable_on_axis: Optional[Union[int, Sequence[int]]] = None,
         eager: bool = True,
         eager_iterator_pool: Optional[ThreadPoolExecutor] = None,
+        context: Optional[SOMATileDBContext] = None,
     ):
         super().__init__()
 
@@ -96,6 +98,7 @@ class BlockwiseReadIterBase(somacore.ReadIter[_RT], metaclass=abc.ABCMeta):
         self.sr = sr
         self.eager = eager
         self.eager_iterator_pool = eager_iterator_pool
+        self.context = context
 
         # raises on various error checks, AND normalizes args
         self.axis, self.size, self.reindex_disable_on_axis = self._validate_args(
@@ -124,7 +127,7 @@ class BlockwiseReadIterBase(somacore.ReadIter[_RT], metaclass=abc.ABCMeta):
         # build indexers, as needed
         self.axes_to_reindex = set(range(self.ndim)) - set(self.reindex_disable_on_axis)
         self.minor_axes_indexer = {
-            d: build_index(self.joinids[d].to_numpy())
+            d: tiledbsoma_build_index(self.joinids[d].to_numpy(), context=context)
             for d in (self.axes_to_reindex - set((self.major_axis,)))
         }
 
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -124,7 +124,7 @@ def __init__(
         # build indexers, as needed
         self.axes_to_reindex = set(range(self.ndim)) - set(self.reindex_disable_on_axis)
         self.minor_axes_indexer = {
-            d: pd.Index(self.joinids[d].to_numpy())
+            d: build_index(self.joinids[d].to_numpy())
```
bkmartinjr, line 127: I suspect this should pull the concurrency from the current context, not just take the default (as is done in ExperimentAxisQuery).  Unfortunately, the context is not passed down to the iterator (it is accessible a couple of calls up the stack).

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -240,9 +240,11 @@ def _reindexed_table_reader(
                 col = tbl.column(f"soma_dim_{d}")
                 if d in self.axes_to_reindex:
                     if d == self.major_axis:
-                        col = pd.Index(coords[self.major_axis]).get_indexer(col.to_numpy())  # type: ignore[no-untyped-call]
+                        col = build_index(coords[self.major_axis]).get_indexer(
```
bkmartinjr, line 243: See previous comment about where the concurrency level is found - suggest using the context

beroy: Sounds good. I will go down through the stack to get the context here.

beroy:
@bkmartinjr I revised the PR as you recommended to propagate the context from top of the stack. The key to this change is I borrowed the context from ```SparseNDArrayBlockwiseRead::array field``` and brought it down to the block blockwise iterators.

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -88,6 +89,8 @@ def __init__(
         reindex_disable_on_axis: Optional[Union[int, Sequence[int]]] = None,
         eager: bool = True,
         eager_iterator_pool: Optional[ThreadPoolExecutor] = None,
+        compress: bool = True,
```
bkmartinjr, line 92: Moving the `compress` flag from BlockwiseScipyReadIter to the base class seems incorrect.  The flag only has meaning in the context of the scipy iterators (i.e., it is a flag which controls the choice of a compressed sparse matrix vs. a COO matrix).

apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -573,6 +573,8 @@ def tables(self) -> BlockwiseTableReadIter:
             size=self.size,
             reindex_disable_on_axis=self.reindex_disable_on_axis,
             eager=self.eager,
+            compress=True,
```
bkmartinjr, line 576: `compress` is meaningless in this context - tables never have a compresses axis.  Why is the flag being added here?

apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -615,7 +617,7 @@ def scipy(self, *, compress: bool = True) -> BlockwiseScipyReadIter:
             self.coords,
             self.axis,
             size=self.size,
-            compress=compress,
```
bkmartinjr, line 618: removing this means the default value (True) will be passed, which means a CSR/CSC matrix is created.  That will be a bug....

bkmartinjr (changes requested):
Lots of errors in the CI tests, which I believe are all a result of the `compress` flag being modified in the Blockwise*ReadIterator classes.

beroy:
You are right. I think somehow I confused the logic and flags. I'll fix them now and ping you as soon as CI is clear.

apis/python/src/tiledbsoma/_index_util.py
```diff
@@ -3,26 +3,45 @@
...
+
 import numpy as np
-import pandas as pd
+from somacore.query.types import IndexLike
 
 from tiledbsoma import pytiledbsoma as clib
 
 from .options import SOMATileDBContext
 
+"""
```
bkmartinjr, line 15: docstring goes inside function, on first line, not above it.

See examples elsewhere in the code base

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -282,8 +290,10 @@ def __init__(
         reindex_disable_on_axis: Optional[Union[int, Sequence[int]]] = None,
         eager: bool = True,
         compress: bool = True,
+        context: SOMATileDBContext = None,  # type: ignore
```
bkmartinjr, line 293: typing is wrong - see comment I left in the super class

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -88,6 +89,7 @@ def __init__(
         reindex_disable_on_axis: Optional[Union[int, Sequence[int]]] = None,
         eager: bool = True,
         eager_iterator_pool: Optional[ThreadPoolExecutor] = None,
+        context: SOMATileDBContext = None,  # type: ignore
```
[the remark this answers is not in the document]
beroy: You are right.

bkmartinjr (approved):

Issue #2014, opened 2024-01-16
[python] Support for nullable columns


johnkerl:
Problem report: Unable to ingest Dataframe column with extension type `Int64` that contains one or more `np.nan` values.

```python
df = pd.DataFrame()

df['empty_int_column'] = pd.Series([1] * 9 + [np.nan], dtype=pd.Int64Dtype())

# try to convert dataframe using TileDB function
arrow_df = df_to_arrow(df1)
arrow_df
```
Produces

```
ArrowInvalid: Could not convert <NA> with type NAType: did not recognize Python value type when inferring an Arrow data type
```

We can actually reproduce this error using arrow directly:

```py
pa.infer_type(df['empty_int_column'])
```

Interestingly, arrow converts the entire dataframe to a table without issue:

```py
pa.Table.from_pandas(df)

## pyarrow.Table
## empty_int_column: int64
## ----
## empty_int_column: [[1,1,1,1,1,1,1,1,1,null]]
```

johnkerl:
[sc-39118]

PR #2083, opened 2024-01-29
[python] Specially handle cast for object and category dtypes

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2014

**Changes:**

* "object" and "category" need explicit casting as the converter functions `pa.infer_type` and `pa.from_numpy_dtype` do not map to the correct associated pandas extension dtype
* Use `None` instead of `pa.null` as `pa.null` can only be used for numerics

**Notes for Reviewer:**


johnkerl (approved):
🚢

PR #2087, opened 2024-01-30
[r] Let `get_tarball.R` interop with `update-tiledb-version.py`

johnkerl:
## Issue

When we do releases and depedency bumps following our [established procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases), `apis/r/tools/get_tarball.R` has always required a manual-edit step.

This PR lets `apis/r/tools/get_tarball.R` be updated hands-off, automatically, along with `libtiledbsoma/cmake/Modules/FindTileDB_EP.cmake`.

This PR centralizes the hash-computation and file-update logic in a way that no longer requires the operator to be familiar with C++/Python/R build internals.

## To give this a test drive

* Check out today's `main`
* `./scripts/update-tiledb-version.py 2.19.0`
* `git diff`

## Before this PR

`scripts/update-tiledb-version.py 2.19.0`

```
diff --git a/apis/r/tools/get_tarball.R b/apis/r/tools/get_tarball.R
index 34d435d6..52b87072 100644
--- a/apis/r/tools/get_tarball.R
+++ b/apis/r/tools/get_tarball.R
@@ -10,7 +10,7 @@ if ( ! dir.exists("inst/") ) {
 }

 makeUrl <- function(arch, ver=tiledb_core_version, sha1=tiledb_core_sha1) {
-    sprintf("https://github.com/TileDB-Inc/TileDB/releases/download/%s/tiledb-%s-%s-%s.tar.gz", ver, arch, ver, sha1)
+    sprintf("https://github.com/TileDB-Inc/TileDB/releases/download/2.19.0/tiledb-2.19.0-2.19.0-2.19.0.tar.gz", ver, arch, ver, sha1)
 }

 isMac <- Sys.info()["sysname"] == "Darwin"
```

which then requires manual edits to restore the line that had `%s`, as well as doing manual edits here:

https://github.com/single-cell-data/TileDB-SOMA/blob/b3e802de05e220ae1d9f6e8a27f8059d4cf0eb24/apis/r/tools/get_tarball.R#L4-L5

## With this PR

`scripts/update-tiledb-version.py 2.19.0`

```
diff --git a/apis/r/tools/get_tarball.R b/apis/r/tools/get_tarball.R
index d2b79600..ff8f3a38 100644
--- a/apis/r/tools/get_tarball.R
+++ b/apis/r/tools/get_tarball.R
@@ -13,9 +13,9 @@ isLinux <- Sys.info()["sysname"] == "Linux"

 if (isMac) {
     arch <- system('uname -m', intern = TRUE)
-    url <- "https://github.com/TileDB-Inc/TileDB/releases/download/2.19.1/tiledb-macos-x86_64-2.19.1-29ceb3e7.tar.gz"
+    url <- "https://github.com/TileDB-Inc/TileDB/releases/download/2.19.0/tiledb-macos-x86_64-2.19.0-fa30a88a.tar.gz"
 } else if (isLinux) {
-    url <- "https://github.com/TileDB-Inc/TileDB/releases/download/2.19.1/tiledb-linux-x86_64-2.19.1-29ceb3e7.tar.gz"
+    url <- "https://github.com/TileDB-Inc/TileDB/releases/download/2.19.0/tiledb-linux-x86_64-2.19.0-fa30a88a.tar.gz"
 } else {
     stop("Unsupported platform for downloading artifacts. Please have TileDB Core installed locally.")
 }
```

No manual edits are required.

### DIFF
--- a/apis/r/tools/get_tarball.R
+++ b/apis/r/tools/get_tarball.R
@@ -13,7 +13,13 @@ isLinux <- Sys.info()["sysname"] == "Linux"
 
 if (isMac) {
     arch <- system('uname -m', intern = TRUE)
-    url <- "https://github.com/TileDB-Inc/TileDB/releases/download/2.19.1/tiledb-macos-x86_64-2.19.1-29ceb3e7.tar.gz"
+    if (arch == "x86_64") {
+      url <- "https://github.com/TileDB-Inc/TileDB/releases/download/2.19.1/tiledb-macos-x86_64-2.19.1-29ceb3e7.tar.gz"
+    } else if (arch == "arm64") {
+      url <- "https://github.com/TileDB-Inc/TileDB/releases/download/2.19.1/tiledb-arm64-x86_64-2.19.1-29ceb3e7.tar.gz"
+    } else {
+      stop("Unsupported Mac architecture. Please have TileDB Core installed locally.")
+    }
 } else if (isLinux) {
     url <- "https://github.com/TileDB-Inc/TileDB/releases/download/2.19.1/tiledb-linux-x86_64-2.19.1-29ceb3e7.tar.gz"
 } else {
--- a/apis/r/tools/get_tarball.R
+++ b/apis/r/tools/get_tarball.R
@@ -1,16 +1,11 @@
 #!/usr/bin/env Rscript
 
-## version pinning info
-tiledb_core_version <- "2.19.1"
-# 8-nybble hash for 2.19.0 and 2.19.1 only. Please see https://github.com/TileDB-Inc/TileDB/pull/4599.
-tiledb_core_sha1 <- "29ceb3e7"
+# Please use scripts/update-tiledb-version.py in this repo's base directory to
+# update URLs in this file as well as the FindTileDB_EP.cmake file within the
+# libtiledbsoma directory.
 
 if ( ! dir.exists("inst/") ) {
     stop("No 'inst/' directory. Exiting.", call. = FALSE)
-}
-
-makeUrl <- function(arch, ver=tiledb_core_version, sha1=tiledb_core_sha1) {
-    sprintf("https://github.com/TileDB-Inc/TileDB/releases/download/%s/tiledb-%s-%s-%s.tar.gz", ver, arch, ver, sha1)
 }
 
 isMac <- Sys.info()["sysname"] == "Darwin"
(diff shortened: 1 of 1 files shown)


apis/r/tools/get_tarball.R
```diff
@@ -1,26 +1,21 @@
...
-    sprintf("https://github.com/TileDB-Inc/TileDB/releases/download/%s/tiledb-%s-%s-%s.tar.gz", ver, arch, ver, sha1)
-}
-
 isMac <- Sys.info()["sysname"] == "Darwin"
 isLinux <- Sys.info()["sysname"] == "Linux"
 
 if (isMac) {
     arch <- system('uname -m', intern = TRUE)
-    url <- makeUrl(paste0("macos-", arch))
+    url <- "https://github.com/TileDB-Inc/TileDB/releases/download/2.19.1/tiledb-macos-x86_64-2.19.1-29ceb3e7.tar.gz"
```
eddelbuettel, line 16: This compute the macOS architecture, stores it in `arch` and then ignores it.  So this breaks installation on non-x86_64 macOS machine based on the arm64 architecture (ie M1, M2, M3).

eddelbuettel (changes requested):
The macOS case used to reflect both architectures, now it only support x86_64.

apis/r/tools/get_tarball.R
johnkerl: Thanks @eddelbuettel !

eddelbuettel:
I saw but I may be sitting this out as I cannot shake the feeling that assigning the required values in just one place is both simpler and more robust.

mojaveazure (dismissed):

mojaveazure:
I noticed an inncorrect URL for macOS-arm64. With that
- if the URL is fixed and `update-tiledb-version.py` is run, will the fixed URL remain or will it become incorrect again?
- if the latter; is there a way to get the `update-tiledb-version.py` to work with the original script where the version/SHA were stashed as variables and R was in charge of creating the URLs? (Not necessary if q1 shows that `update-tiledb-version.py` respects the correct URL)

johnkerl:
@mojaveazure thanks for the review!

* Typo fixed on https://github.com/single-cell-data/TileDB-SOMA/pull/2087/commits/fcacb53ba0db5d3466c592ab19d697410ca2fabc
* This script indeed automatically updates as described (with output shown) in the description field of this PR: https://github.com/single-cell-data/TileDB-SOMA/blob/1.7.0/scripts/update-tiledb-version.py

mojaveazure (approved):
Awesome, thanks! I'll echo Dirk's sentiment that defining the version/SHA in a single place feels more robust. However, I cannot at this time see a reason to not approve this PR, especially if it makes your life easier

🚢

Issue #2086, opened 2024-01-30
[Bug] SOMA datasets written to S3 and moved to new bucket still point to old bucket. 


royassis:
**Describe the bug**
SOMA datasets generated in S3 using `from_anndata` and copied to another bucket, still point to old bucket. 
The datasets were probably generate with the default parameter for `use_relative_uri` in `from_anndata`.

**Versions (please complete the following information):**
```
tiledbsoma.__version__        1.6.1
TileDB-Py tiledb.version()    (0, 24, 0)
TileDB core version           2.18.2
libtiledbsoma version()       libtiledb=2.18.2
python version                3.10.13.final.0
OS version                    Linux 5.15.133.1-microsoft-standard-WSL2
```

Is there a way to fix this without generating new datasets?

johnkerl:
Hi @royassis and thanks for writing!

Indeed when we ingest from local disk or S3, we default to `use_relative=True`, precisely to make these datasets relocatable.

To investigate the current status, can you please run the following and share the output?
https://gist.github.com/johnkerl/2b04c9089d3159c4b7bd4a43b6847734

royassis:
Hey @johnkerl, thanks for the quick response. 

Ok, I got some more info. I see that I get an error message when I try to read the .uns. 
I remember that I added them separately after I've created the SOMA objects on S3. 

I used this code:

```
import tiledbsoma
import boto3
from tiledbsoma.io import from_anndata
import anndata
import pandas as pd
import pyarrow as pa

bucket_name = "madeupbucket"
s3_resource = boto3.resource('s3')
bucket = s3_resource.Bucket(bucket_name)

s3_path = f"s3://{bucket_name}/test"

df = pd.DataFrame(data=[[1]], columns=["gene_a"], index=["cell_a"])
ad = anndata.AnnData(X=df)
ad.uns["k"] = "a"

bucket.object_versions.filter(Prefix="test").delete()
from_anndata(experiment_uri=s3_path, anndata=ad, measurement_name="RNA")

exp: tiledbsoma.Experiment = tiledbsoma.open(s3_path)

uns_uri = exp.ms['RNA'].uri + "/uns"
uns_key = "annotations"
annotations_uri = f"{uns_uri}/{uns_key}"
exp_prefix = exp.uri.split("/", 3)[-1]
annotations_prefix = annotations_uri.split("/", 3)[-1]
annotations_df = pd.DataFrame(data=[[1, 2]], columns=["val", "key"])
[...]

PR #2088, opened 2024-01-30
[python] Elaborate docstring re relativity for add-new

johnkerl:
Context: https://github.com/single-cell-data/TileDB-SOMA/issues/2086#issuecomment-1917669879


aaronwolen (approved):
Good idea.

PR #2096, opened 2024-01-31
[c++] Fix for error when the key array size is smaller than thread count

beroy:
**Issue and/or context:**
When key size is smaller than thread count, we incorrectly throw an exception

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/reindexer/reindexer.cc
+++ b/libtiledbsoma/src/reindexer/reindexer.cc
@@ -51,14 +51,6 @@ void IntIndexer::map_locations(const int64_t* keys, int size, int threads) {
     if (size == 0) {
         return;
     }
-    if (size < 10) {
-        threads = 1;
-    }
-    if (size < threads)
-        throw std::runtime_error(
-            "The number of keys " + std::to_string(size) +
-            " must be larger than the number of threads " +
-            std::to_string(threads) + " .");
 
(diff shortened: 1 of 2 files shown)


johnkerl (approved):

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -51,14 +51,6 @@ void IntIndexer::map_locations(const int64_t* keys, int size, int threads) {
     if (size == 0) {
```
bkmartinjr, line 51: several concerns here with param checking:
1. do you want map_locations to create a thread pool in the case where `threads == 1`?  It looks like it will never be used.
2. Why is `threads` a signed value?  ThreadPool uses `size_t`.  If there is meaning to negative values, don't you need to bounds check the arg?
3. What happens if `size` is <= 0?  Does this need a check?  Should size be unsigned?

johnkerl: @beroy please let's do a follow-up. I shouldn't have approved this; @bkmartinjr 's eyes are sharper than mine. :)

beroy: Follow up PR: https://github.com/single-cell-data/TileDB-SOMA/pull/2098

PR #2098, opened 2024-01-31
[C++] Handling edge cases for C++ re-indexer

beroy:
**Issue and/or context:**
Some of the edge case like negative size table and handling thread sizes smaller or equal to one could be handled better
**Changes:**

**Notes for Reviewer:**


libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -48,9 +48,19 @@ namespace tiledbsoma {
 
 void IntIndexer::map_locations(const int64_t* keys, int size, int threads) {
```
bkmartinjr, line 49: Why are `threads` and `size` params signed int?  I note that ThreadPool accepts a `size_t`, so there is already a cast occurring....

Using appropriate types seems like a better solution, unless there is some sort of upstream requirement for signed values.

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -48,9 +48,19 @@ namespace tiledbsoma {
...
     map_size_ = size;
+    thread_count_ = threads;
+
+    // Handling edge cases
     if (size == 0) {
         return;
     }
+    if (size < 0) {
+        throw std::runtime_error("Re-indexer map size cannot be native.");
```
bkmartinjr, line 58: `native`?   Did you mean negative?

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -48,9 +48,19 @@ namespace tiledbsoma {
...
+    }
+    if (threads <= 0) {
+        throw std::runtime_error(
+            "Re-indexer thread count cannot be zero or native.");
```
bkmartinjr, line 62: native - typo?

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -78,18 +88,22 @@ void IntIndexer::map_locations(const int64_t* keys, int size, int threads) {
...
-        tiledb_thread_pool_->concurrency_level(),
-        size));
-    if (tiledb_thread_pool_->concurrency_level() == 1) {
+    if (tiledb_thread_pool_) {
+        LOG_DEBUG(fmt::format(
+            "Lookup with thread concurrency {} on data size {}",
+            tiledb_thread_pool_->concurrency_level(),
+            size));
+    }
+    if (thread_count_ == 1) {
```
bkmartinjr, line 106: storing both `tiledb_thead_pool` and `thread_count` seems redundant.  Would be simpler to just use the former, e.g.,

```
if (tiledb_thread_pool_ == nullptr) {
    // do the single thread loop
    return
}

// else do the multi-thread path
```

You could remove `thread_count`, and move the new logging (line 101) into the MT branch

libtiledbsoma/src/reindexer/reindexer.h
```diff
@@ -96,6 +96,10 @@ class IntIndexer {
      * Number of elements in the map set by map_locations
      */
     int map_size_ = 0;
+    /*
+     * Number of requested threads
+     */
+    int thread_count_ = 0;
```
bkmartinjr, line 102: this new value seems redundant with tiledb_thread_pool, at least as currently used

bkmartinjr (commented):
functionally looks fine, but it could be simplified and cleaned up a bit:
* use unsigned types for arguments which should never (can never) be negative.
* remove thread_count_ instance value which, as used, seems redundant with the tiledb_thread_pool_ slot

libtiledbsoma/src/reindexer/reindexer.cc
beroy: Im setting both of them to size_t

libtiledbsoma/src/reindexer/reindexer.h
beroy: Good catch

beroy:
@bkmartinjr done with fixing both issue and now waiting for CI

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -46,11 +46,17 @@ KHASH_MAP_INIT_INT64(m64, int64_t)
...
+
+    // Handling edge cases
     if (size == 0) {
         return;
     }
+    if (threads == 0) {
+        throw std::runtime_error("Re-indexer thread count cannot be zero.");
+    }
 
     LOG_DEBUG(fmt::format(
```
bkmartinjr, line 61: any reason to have duplicate logging here - same information is in both statements....

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -101,6 +105,10 @@ void IntIndexer::lookup(const int64_t* keys, int64_t* results, int size) {
         }
         return;
     }
+    LOG_DEBUG(fmt::format(
+        "Lookup with thread concurrency {} on data size {}",
+        tiledb_thread_pool_->concurrency_level(),
+        size));
```
bkmartinjr, line 111: nit: these two log lines are largely redundant as well

bkmartinjr (commented):
tiny nit:  comment in line 30 is incorrect (looks like copy & paste error)

bkmartinjr (approved):
a couple of minor nits to pick, otherwise LGTM. TY.

libtiledbsoma/src/reindexer/reindexer.cc
beroy: fixed

libtiledbsoma/src/reindexer/reindexer.cc
beroy: fixed

PR #2102, opened 2024-02-02
[r] Fix quoting for a particular Docker-build environment (`release-1.7`)

johnkerl:
**Issue and/or context:** This arose while doing a Docker build within a particular TileDB-internal build environment. There exists platforms within which

```
root@1f09efc85df8:/# /bin/sh
# export CC=`${R_HOME}/bin/R CMD config CC`
# export CXX=`${R_HOME}/bin/R CMD config CXX`
/bin/sh: 2: export: -std: bad variable name
```

**Changes:** Properly double-quote variables which may contain spaces.

**Notes for Reviewer:**

As discussed in Slack.

This will also be applied to the `main` branch: #2103

PR #2103, opened 2024-02-02
[r] Fix quoting for a particular Docker-build environment (`main`)

johnkerl:
Sibling of #2102 on `main`

### DIFF
--- a/apis/r/configure
+++ b/apis/r/configure
@@ -41,9 +41,9 @@ ${R_HOME}/bin/Rscript tools/get_tarball.R
 ${R_HOME}/bin/Rscript tools/check_cmake_and_git.R
 
 ## Make libtiledbsoma library using cmake (and an added git dependency)
-export CC=`${R_HOME}/bin/R CMD config CC`
-export CXX=`${R_HOME}/bin/R CMD config CXX`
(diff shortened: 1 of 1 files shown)


eddelbuettel:
I had my battles over the years with similar things between shell script (snippets), `configure`, and various (CRAN particularly) build systems ... but I also have the gnarling suspicion that _maybe_ we (also ?) need to protect it where these variables are _used_ rather than where they are defined.

I recall you showed an error where the `-std=` broke off.  Was that a `make` call?

johnkerl:
@eddelbuettel I have a failed Docker build without this fix, and succeeded with it

Quoting at the use-spot would result in things like `"g++ -arch x86_64 -std=gnu++17"` being a single argv0 which, in this particular case, is not desired.

eddelbuettel (approved):
:shipit: to robustify against annoying `/bin/sh` implementations

johnkerl:
Thank you @eddelbuettel !

johnkerl:
Re the above backport-fail: by one means or another (it doesn't matter which, and it isn't work even taking the time to search through commit history) the `release-1.7` branch already has this fix.

PR #2112, opened 2024-02-06
[python] Accumulate obs/var IDs across batched reads in `register_anndatas()`

aaronwolen:
**Issue and/or context:** `tiledbsoma.io.register_anndatas()` would fail to return all obs/var IDs on disk if they exceeded the configured column-buffer size.

**Changes:** IDs are now accumulated across batches.


johnkerl (approved):
🚢

Issue #2079, opened 2024-01-29
[python] `tiledbsoma.io.to_anndata`/`to_h5ad` should populate all `adata.layers` from measurement's `X`


johnkerl:
**Describe the bug**
On ingest, we write `adata.X` and `adata.layers` into the SOMA experiment's `X` collection.

On outgest, we only create `adata.X`, not `adata.layers`.

johnkerl:
[sc-40225]

PR #2119, opened 2024-02-08
[python] Outgest additional X layers to AnnData

johnkerl:
**Issue and/or context:** Issue #2079

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -9,7 +9,6 @@ This module contains methods to generate SOMA artifacts starting from
 other formats. Currently only ``.h5ad`` (`AnnData <https://anndata.readthedocs.io/>`_) is supported.
 """
 
-import collections
 import json
 import math
 import time
@@ -17,6 +16,7 @@ from typing import (
     Any,
     ContextManager,
     Dict,
+    KeysView,
     List,
     Mapping,
     Optional,
@@ -2945,9 +2945,7 @@ def to_anndata(
     measurement_name: str,
     *,
     X_layer_name: Optional[str] = "data",
-    extra_X_layer_names: Optional[
-        Union[Sequence[str], collections.abc.KeysView[str]]
-    ] = None,
+    extra_X_layer_names: Optional[Union[Sequence[str], KeysView[str]]] = None,
     obs_id_name: Optional[str] = None,
     var_id_name: Optional[str] = None,
     obsm_varm_width_hints: Optional[Dict[str, Dict[str, int]]] = None,
@@ -3075,7 +3073,7 @@ def to_anndata(
     #   extra_X_layer_names=exp.ms["RNA"].X.keys()
     # while avoiding
     #   TypeError: 'ABCMeta' object is not subscriptable
-    if isinstance(extra_X_layer_names, collections.abc.KeysView):
+    if isinstance(extra_X_layer_names, KeysView):
         extra_X_layer_names = list(extra_X_layer_names)
 
     if X_layer_name is None and extra_X_layer_names:
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -9,6 +9,7 @@ This module contains methods to generate SOMA artifacts starting from
 other formats. Currently only ``.h5ad`` (`AnnData <https://anndata.readthedocs.io/>`_) is supported.
 """
 
+import collections
 import json
 import math
 import time
@@ -2909,12 +2910,44 @@ def to_h5ad(
 
 
 # ----------------------------------------------------------------
+def _extract_X_key(
+    measurement: Measurement,
+    X_layer_name: str,
+    nobs: int,
+    nvar: int,
+) -> Matrix:
+    """Helper function for to_anndata"""
+
+    if X_layer_name not in measurement.X:
+        raise ValueError(
+            f"X_layer_name {X_layer_name} not found in data: {measurement.X.keys()}"
+        )
+
+    # Acquire handle to TileDB-SOMA data
+    soma_X_data_handle = measurement.X[X_layer_name]
+    logging.log_io(None, f"FOO {X_layer_name} {soma_X_data_handle.uri}")
+
+    # Read data from SOMA into memory
+    if isinstance(soma_X_data_handle, DenseNDArray):
+        data = soma_X_data_handle.read((slice(None), slice(None))).to_numpy()
+    elif isinstance(soma_X_data_handle, SparseNDArray):
+        X_mat = soma_X_data_handle.read().tables().concat().to_pandas()
+        data = conversions.csr_from_tiledb_df(X_mat, nobs, nvar)
+    else:
+        raise TypeError(f"Unexpected NDArray type {type(soma_X_data_handle)}")
+
+    return data
+
+
+# ----------------------------------------------------------------
 def to_anndata(
     experiment: Experiment,
     measurement_name: str,
     *,
     X_layer_name: Optional[str] = "data",
-    extra_X_layer_names: Optional[Sequence[str]] = None,
+    extra_X_layer_names: Optional[
+        Union[Sequence[str], collections.abc.KeysView[str]]
+    ] = None,
     obs_id_name: Optional[str] = None,
     var_id_name: Optional[str] = None,
     obsm_varm_width_hints: Optional[Dict[str, Dict[str, int]]] = None,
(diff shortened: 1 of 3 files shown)


johnkerl:
@aaronwolen @mojaveazure ping 🙏

johnkerl:
@aaronwolen @mojaveazure ping 🙏

aaronwolen (changes requested):
Limiting the user's export options to either all or none of the X layers may be too inflexible. Ideally, they should be able to specify any subset of the layers they want.

johnkerl:
@aaronwolen thanks!

To avoid length iteration cycles, please take a minute or two to simply offer a candidate method signature, and I will do what you ask

Thank you in advance!

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -2817,7 +2815,11 @@ def to_anndata(
 
     The ``X_layer_name`` is the name of the TileDB-SOMA measurement's
     ``X`` collection which will be outgested to the resulting AnnData object's
-    ``adata.X``.
+    ``adata.X``. If this is ``None``, then the return value's ``adata.X`` will
+    be None, and ``adata.layers`` will be unpopulated. If this is not ``None``,
```
mojaveazure, line 2819: This doesn't make sense; `adata.X` should always be populated. I would recommend having a new parameter  (eg. `layers`) to handle adding additional layers to `adata.layers`

Ideally, this would accept three parameters
- A single `str` or collection of `str` for specific `exp.ms.X` to load to `adata.layers`
- `True` to pull all remaining `exp.ms.X` into `adata.layers`
- `False` or `""` to prevent populating `adata.layers`

mojaveazure (changes requested):

apis/python/src/tiledbsoma/io/ingest.py
aaronwolen: > This doesn't make sense; adata.X should always be populated. 

There have been specific use cases from users for creating `AnnData` objects with empty `X` fields.

mojaveazure: Interesting; I can't think of a case where you'd want an empty `X` and wasn't aware that AnnData allowed `adata.X` to be empty (neither Seurat nor SCE allow it)

mojaveazure: @johnkerl Then I retract that part of my review, but `adata.layers` should be populated by a separate parameter rather than be controlled by `X_layer_name` (although if `X_layer_name` is `None`, populating `adata.layers` should be suppressed regardless)

johnkerl: @mojaveazure maybe this:

```
* `X_layer_name: Optional[str] = None` for what will become `adata.X` -- `None` for those customers who want to suppress that
* `X_layer_restrictions: Optional[Sequence[str]] = None` for what will become `adata.layers` -- `None` means no restrictions, all SOMA Experiment layers other than `X_layer_name` will go to `adata.layers; non-none sequence means take `adata.layers` to be those

Example:

* SOMA Measrurement has `"a"`, `"b"`, `"c"`
* `to_anndata` with `X_layer_name="a"` and `X_layer_restrictions=None` results in `"a"` going to `adata.X` and `"b", `"c"` going to `adata.layers["b"]` and `adata.layers["c"]`
* `to_anndata` with `X_layer_name="a"` and `X_layer_restrictions=["b"]` results in `"a"` going to `adata.X` and `"b" going to `adata.layers["b"]`

Would this check all your boxes?

Comment (johnkerl) [reply apis/python/src/tiledbsoma/io/ingest.py:2819]: Thanks for the suggestions -- I'll  re-engage tomorrow morning

Comment (johnkerl) [reply apis/python/src/tiledbsoma/io/ingest.py:2819]: (Also we can't change `X_layer_name` -- it's existing API.)

Comment (johnkerl) [reply apis/python/src/tiledbsoma/io/ingest.py:2819]: @mojaveazure  I like the option of True for all, False for none at all, sequence of string for something in between ...

I'll get the green light on the function signature from you and/or @aaronwolen tomorrow before writing any more code that doesn't satisfy your needs

Comment (johnkerl) [reply apis/python/src/tiledbsoma/io/ingest.py:2819]: @mojaveazure I feel like the mixing of boolean and list-of-string is not quite Pythonic. I'm thinking of a nullable-list-of-string approach and wanted to get your & @aaronwolen 's thoughts on the following proposal:

```
X_layer_name: Optional[str] = "data",
alt_X_layer_restrictions: Optional[Sequence[str]] = None,
```

* Note: for Python typing, `Optional` is the way one says "nullable" -- it simply means the thing can be `None`
* We cannot change the current syntax or semantics of `X_layer_name` without breaking backward compatibility
* `X_layer_name=None` means the output `adata`'s `adata.X` will be `None` and also `adata.layers` will be empty. If `X_layer_name` is not None, then that key in the measurement's `X` must exist (it's an error if it's not), then this will be used to populate `adata.X`.
* If `alt_X_layer_restrictions` is `None` (the default), then there are no restrictions. All of the measurement's `X` slots _except_ `X_layer_name` will be used to produce `adata.layers`.
* If `alt_X_layer_restrictions` is a specified list of keys, then each of those keys must exist in the measurement (it's an error if any of them don't), and these will be used to populate `adata.layers`
* If `alt_X_layer_restrictions` is an empty list, then `adata.layers` will be empty.

@mojaveazure @aaronwolen does this meet everyone's needs? 

I'll block any attempts at implementation until we get the API agreed on.

Comment (mojaveazure) [reply apis/python/src/tiledbsoma/io/ingest.py:2819]: I think so? I'm fine with the existing semantics around `X_layer_name` and will not be suggesting changes to it

The reason I suggested using `True`/`False` as shortcuts is this is how the R API works. For example, when specifying `varm` layers to pull in `$to_seurat()`, the user can pass:
- a vector of characters (Python `Sequence[str]`) to select specific `varm` layers to pull
- `FALSE` to prevent pulling any `varm` layers
- `TRUE` to force pulling _all_ `varm` layers
- `NULL` (Python `None`) (default) to pull `varm` layers if and only if we can match them to an `obsm` layer (Seurat stores `obsm` and `varm` together, and it generally does not make sense to have a `varm` layer without an `obsm` layer, but it's possible)

For this `alt_X_layer`, my concern is finding an intuitive way to express both "pull all `alt_X_layers`" and "do not pull any `alt_X_layers`"; my gut says that `True`/`False` (with `True` being equivalent to `None`) would be more intuitive, but if the Pythonic way of expressing these is `None`/`[]` or `None`/`""`, then go for it

I'm also not entirely a fan of `alt_X_layer_restrictions`: this would be the first outgestion parameter across the APIs called this. On the R side, these "maybe outgest" items are named to convey "include these extras if available" (eg. `obsm_layers`, `var_column_names`). I feel `alt_layers` or `alt_X_layers` would be a more appropriate parameter name to convey "include these alternate X layer" rather than "restrict the alternate X layers to only these ones"

Comment (johnkerl) [reply apis/python/src/tiledbsoma/io/ingest.py:2819]: On discussion with @aaronwolen: 

* The default for additional layers should `adata.layers` being empty -- whereas above I was having it default to all
* I used the name `alt_X_layer_restrictions` above so that `None` would mean "no restrctions" -- there needs to be a way to say "all" that is easy --- but this also makes this argument _sound_ like a block-list even though it would actually be an allow-list

Next round of proposal:

```
X_layer_name: Optional[str] = "data",
additional_X_layer_names: Optional[Sequence[str]] = None,

```

* As above: We cannot change the current syntax or semantics of `X_layer_name` without breaking backward compatibility
* As above: `X_layer_name=None` means the output `adata`'s `adata.X` will be `None` and also `adata.layers` will be empty.
  * It will be an error for `X_layer_name` to be `None` and `additional_X_layer_names` to not be `None`.
* If `X_layer_name` is not None, then that key in the measurement's `X` must exist (it's an error if it's not), then this will be used to populate `adata.X`.
* If `additional_X_layer_names` is `None` (or empty list) then `adata.layers` will be empty
* If `additional_X_layer_names` is a provided list then we will use those for populating `adata.layers`
* There still needs to be an easy way to say "all layers"
  * They will be able to use `exp.ms["RNA"].X.keys()` and I'll include an example in the doc-block
  * I'll do the user this favor: say X keys are `"a"`, `"b"`, `"c"`, `"d"` and they want
    `X_layer_name` to be `"b"` for the output `adata.X`.  They want `"a"`, `"c"`, `"d"` to be
    `adata.layers`.  Instead of forcing them to remove `"b"` from their list of all keys, I'll
    simply say that the primary `X_layer_name` won't be redundantly added to `adata.layers`.
    That will give them a simple way to say "all" that doesn't force them into awkward
    keystroking.

@mojaveazure @aaronwolen please let me know if this works for you.

Comment (aaronwolen) [reply apis/python/src/tiledbsoma/io/ingest.py:2819]: This sounds great! Thanks for the summary. 

Could we substitute `extra` for `additional` to save the user a few key strokes and reduce opportunities for typos?

Comment (mojaveazure) [reply apis/python/src/tiledbsoma/io/ingest.py:2819]: I'll second @aaronwolen's approval and request to change `additional` to `extra` (`extra_X_layer_names` instead of `additional_X_layer_names`)

**Comment by johnkerl** [comment]
I don't know why this is suddenly failing CI in files I did not touch.

I am actively debugging.

**Review by aaronwolen — APPROVED** [review]
Thanks for the productive discussions. Tested it out and it works great.

**Comment by johnkerl** [comment]
I don't know why this is now failing Python 3.7 CI on tests which I did not touch. It passes tests for me on my laptop. I will iteratively debug on #2137.

File: apis/python/src/tiledbsoma/io/ingest.py [hunk 2]
```diff
@@ -2885,12 +2884,45 @@ def to_h5ad(
...
+
+
 # ----------------------------------------------------------------
 def to_anndata(
     experiment: Experiment,
     measurement_name: str,
     *,
     X_layer_name: Optional[str] = "data",
+    extra_X_layer_names: Optional[
+        Union[Sequence[str], collections.abc.KeysView[str]]
```
ihnorton, line 2924: This needs to be `typing.KeysView` for 3.7 compatibility.

johnkerl: Thank you @ihnorton !!!!!

aaronwolen:
[sc-40032]

PR #2131, opened 2024-02-13
Add type checks to `obsm`, `obsp`, `varm` and `varp` early in ingestion

### LINKED ISSUE #2118: [python] Early error for dataframe objects presented as `obsm` etc. on ingest
The `obsm`, `obsp`, `varm`, and `varp` must be 2D arrays, not dataframes. In those cases when an input AnnData object does contain a dataframe in any of these slots, the ingest fails (as it should) but it fails partway through the ingest, and it fails with a stack trace.

* We should have a type-check early on in `from_anndata` (next to the other existing error checks) -- before the ingest starts -- not within the ingest-matrix function at which point the experiment-level ingest would crash partway through
* The error message should be user-friendly

XanthosXanthopoulos:
**Issue and/or context:**
closes #2118

**Changes:**
Adds explicit type checks to `from_anndata` for `obsm`, `obsp`, `varm` and `varp` before the actual ingestion start to catch the case where a `DataFrame` is passed.

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -420,6 +420,30 @@ def from_anndata(
         raise TypeError(
             "Second argument is not an AnnData object -- did you want from_h5ad?"
         )
+    
+    for key in anndata.obsm.keys():
+        if not isinstance(anndata.obsm[key], Matrix.__args__):
+            raise TypeError(
+                f"Obsm value at {key} in not of type {list(cl.__name__ for cl in Matrix.__args__)}"
+            )
+        
+    for key in anndata.obsp.keys():
+        if not isinstance(anndata.obsp[key], Matrix.__args__):
+            raise TypeError(
+                f"Obsp value at {key} in not of type {list(cl.__name__ for cl in Matrix.__args__)}"
+            )
+        
+    for key in anndata.varm.keys():
+        if not isinstance(anndata.varm[key], Matrix.__args__):
+            raise TypeError(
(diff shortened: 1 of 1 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -421,6 +422,30 @@ def from_anndata(
             "Second argument is not an AnnData object -- did you want from_h5ad?"
         )
 
+    for key in anndata.obsm.keys():
+        if not isinstance(anndata.obsm[key], get_args(Matrix)):
+            raise TypeError(
+                f"Obsm value at {key} in not of type {list(cl.__name__ for cl in get_args(Matrix))}"
```
johnkerl, line 428: Users will expect `obsm` not `Obsm` -- same for the other three

johnkerl (changes requested):
@XanthosXanthopoulos thanks for working on this! :) 

Let's also add a unit-test case

* In `apis/python/tests/test_basic_anndata_io.py`
* You can get `adata`
* Make a `bdata` with `bdata = ad.AnnData(X=adata.X, obs=adata.obs, var=adata.var, obsm={"testing": adata.obs})` -- this will result in a dataframe being in an `obsm` slot
* Wrap a `from_anndata` call in `with pytest.raises(TypeError)`
* Assert the destination SOMA URI doesn't exist at all (i.e. there was no partial write)

johnkerl (approved):
🚢 Thanks @XanthosXanthopoulos !!! 🚢

johnkerl:
Please feel free to self-merge

PR #2148, opened 2024-02-15
[c++/python] Use a shared threadpool for the reindexer

beroy:
**Issue and/or context:** #2149

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/_index_util.py
+++ b/apis/python/src/tiledbsoma/_index_util.py
@@ -36,12 +36,6 @@ def tiledbsoma_build_index(
     Lifecycle:
         Experimental.
     """
-    if context is not None:
-        tdb_concurrency = int(
-            context.tiledb_ctx.config().get("sm.compute_concurrency_level", 10)
-        )
-        thread_count = max(1, tdb_concurrency // 2)
-
-    reindexer = clib.IntIndexer()
-    reindexer.map_locations(keys, thread_count)
+    reindexer = clib.IntIndexer(context)
+    reindexer.map_locations(keys)
     return reindexer  # type: ignore[no-any-return]
--- a/libtiledbsoma/src/reindexer/reindexer.cc
+++ b/libtiledbsoma/src/reindexer/reindexer.cc
@@ -36,7 +36,7 @@
 #include <thread>
 #include "khash.h"
 #include "soma/enums.h"
-#include "soma/soma_array.h"
+#include "soma/soma_context.h"
 #include "utils/arrow_adapter.h"
 #include "utils/common.h"
 #include "utils/logger.h"
@@ -46,16 +46,12 @@ KHASH_MAP_INIT_INT64(m64, int64_t)
 
 namespace tiledbsoma {
 
-void IntIndexer::map_locations(
-    const int64_t* keys, size_t size, size_t threads) {
+void IntIndexer::map_locations(const int64_t* keys, size_t size) {
     map_size_ = size;
 
     // Handling edge cases
     if (size == 0) {
         return;
-    }
-    if (threads == 0) {
-        throw std::runtime_error("Re-indexer thread count cannot be zero.");
     }
 
     hash_ = kh_init(m64);
@@ -64,10 +60,8 @@ void IntIndexer::map_locations(
     khint64_t k;
     int64_t counter = 0;
     // Hash map construction
-    LOG_DEBUG(fmt::format(
-        "[Re-indexer] Start of Map locations with {} keys and {} threads",
-        size,
-        threads));
+    LOG_DEBUG(
+        fmt::format("[Re-indexer] Start of Map locations with {} keys", size));
     for (size_t i = 0; i < size; i++) {
         k = kh_put(m64, hash_, keys[i], &ret);
         assert(k != kh_end(hash_));
@@ -79,9 +73,7 @@ void IntIndexer::map_locations(
     }
     auto hsize = kh_size(hash_);
     LOG_DEBUG(fmt::format("[Re-indexer] khash size = {}", hsize));
-    if (threads > 1) {
-        tiledb_thread_pool_ = std::make_unique<tiledbsoma::ThreadPool>(threads);
-    }
+
     LOG_DEBUG(
         fmt::format("[Re-indexer] Thread pool started and hash table created"));
 }
@@ -90,7 +82,9 @@ void IntIndexer::lookup(const int64_t* keys, int64_t* results, int size) {
     if (size == 0) {
         return;
     }
-    if (tiledb_thread_pool_ == nullptr) {  // When concurrency is 1
+    // Single thread checks
+    if (context_ == nullptr || context_->thread_pool() == nullptr ||
+        context_->thread_pool()->concurrency_level() == 1) {
         for (int i = 0; i < size; i++) {
             auto k = kh_get(m64, hash_, keys[i]);
             if (k == kh_end(hash_)) {
@@ -104,12 +98,13 @@ void IntIndexer::lookup(const int64_t* keys, int64_t* results, int size) {
     }
     LOG_DEBUG(fmt::format(
         "Lookup with thread concurrency {} on data size {}",
-        tiledb_thread_pool_->concurrency_level(),
+        context_->thread_pool()->concurrency_level(),
         size));
 
     std::vector<tiledbsoma::ThreadPool::Task> tasks;
 
-    size_t thread_chunk_size = size / tiledb_thread_pool_->concurrency_level();
+    size_t thread_chunk_size = size /
+                               context_->thread_pool()->concurrency_level();
     if (thread_chunk_size == 0) {
         thread_chunk_size = 1;
     }
@@ -122,7 +117,7 @@ void IntIndexer::lookup(const int64_t* keys, int64_t* results, int size) {
         }
         LOG_DEBUG(fmt::format(
             "Creating tileDB task for the range from {} to {} ", start, end));
-        tiledbsoma::ThreadPool::Task task = tiledb_thread_pool_->execute(
+        tiledbsoma::ThreadPool::Task task = context_->thread_pool()->execute(
             [this, start, end, &results, &keys]() {
                 for (size_t i = start; i < end; i++) {
                     auto k = kh_get(m64, hash_, keys[i]);
@@ -142,17 +137,13 @@ void IntIndexer::lookup(const int64_t* keys, int64_t* results, int size) {
             start,
             end));
     }
-    tiledb_thread_pool_->wait_all(tasks);
+    context_->thread_pool()->wait_all(tasks);
 }
 
 IntIndexer::~IntIndexer() {
     if (map_size_ > 0) {
         kh_destroy(m64, this->hash_);
     }
-}
-
-IntIndexer::IntIndexer(const int64_t* keys, int size, int threads) {
-    map_locations(keys, size, threads);
 }
 
 }  // namespace tiledbsoma
--- a/libtiledbsoma/src/reindexer/reindexer.h
+++ b/libtiledbsoma/src/reindexer/reindexer.h
@@ -43,7 +43,7 @@ struct kh_m64_s;
 
 namespace tiledbsoma {
 
-class ThreadPool;
+class SOMAContext;
 
 class IntIndexer {
    public:
@@ -53,9 +53,9 @@ class IntIndexer {
      * @param size yhr number of keys in the put
      * @param threads number of threads in the thread pool
      */
-    void map_locations(const int64_t* keys, size_t size, size_t threads = 4);
-    void map_locations(const std::vector<int64_t>& keys, size_t threads = 4) {
-        map_locations(keys.data(), keys.size(), threads);
+    void map_locations(const int64_t* keys, size_t size);
+    void map_locations(const std::vector<int64_t>& keys) {
+        map_locations(keys.data(), keys.size());
     }
     /**
      * Used for parallel lookup using khash
@@ -74,12 +74,8 @@ class IntIndexer {
         lookup(keys.data(), results.data(), keys.size());
     }
     IntIndexer(){};
-    /**
-     * Constructor with the same arguments as map_locations
-     */
-    IntIndexer(const int64_t* keys, int size, int threads);
-    IntIndexer(const std::vector<int64_t>& keys, int threads)
-        : IntIndexer(keys.data(), keys.size(), threads) {
+    IntIndexer(std::shared_ptr<tiledbsoma::SOMAContext> context) {
+        context_ = context;
     }
     virtual ~IntIndexer();
 
@@ -88,10 +84,8 @@ class IntIndexer {
      * The created 64bit hash table
      */
     kh_m64_s* hash_;
-    /*
-     * TileDB threadpool
-     */
-    std::shared_ptr<tiledbsoma::ThreadPool> tiledb_thread_pool_ = nullptr;
+
+    std::shared_ptr<SOMAContext> context_ = nullptr;
     /*
      * Number of elements in the map set by map_locations
      */
--- a/libtiledbsoma/src/soma/soma_context.cc
+++ b/libtiledbsoma/src/soma/soma_context.cc
@@ -1,0 +1,49 @@
+/**
+ * @file   soma_context.h
+ *
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2024 TileDB, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @section DESCRIPTION
+ *
+ *   This file defines the SOMAContext class.
+ */
+#include "soma_context.h"
+#include <thread_pool/thread_pool.h>
+
+namespace tiledbsoma {
+
+void SOMAContext::create_thread_pool() {
+    // Extracting concurrency from tiledb config
+    auto cfg = tiledb_config();
+    int concurrency = 10;
+    if (cfg.find("sm.compute_concurrency_level") != cfg.end()) {
+        concurrency = std::stoi(cfg["sm.compute_concurrency_level"]);
+    }
+    int thread_count = std::max(1, concurrency / 2);
+    if (thread_count > 1) {
+        thread_pool_ = std::make_shared<ThreadPool>(thread_count);
+    }
+}
+}  // namespace tiledbsoma
--- a/libtiledbsoma/src/soma/soma_context.h
+++ b/libtiledbsoma/src/soma/soma_context.h
@@ -34,10 +34,12 @@
 #define SOMA_CONTEXT
 
 #include <map>
+#include <mutex>
 #include <string>
 #include <tiledb/tiledb>
 
 namespace tiledbsoma {
+class ThreadPool;
 
 using namespace tiledb;
 
@@ -47,10 +49,12 @@ class SOMAContext {
     //= public non-static
     //===================================================================
     SOMAContext()
-        : ctx_(std::make_shared<Context>(Config({}))){};
+        : ctx_(std::make_shared<Context>(Config({})))
+        , thread_pool_mutex_(){};
 
     SOMAContext(std::map<std::string, std::string> platform_config)
-        : ctx_(std::make_shared<Context>(Config(platform_config))){};
+        : ctx_(std::make_shared<Context>(Config(platform_config)))
+        , thread_pool_mutex_(){};
 
     bool operator==(const SOMAContext& other) const {
         return ctx_ == other.ctx_;
@@ -67,6 +71,17 @@ class SOMAContext {
         return cfg;
     }
 
+    std::shared_ptr<ThreadPool>& thread_pool() {
+        const std::lock_guard<std::mutex> lock(thread_pool_mutex_);
+        // The first thread that gets here will create the context thread pool
+        if (thread_pool_ == nullptr) {
+            create_thread_pool();
+        }
+        return thread_pool_;
+    }
+
+    void create_thread_pool();
+
    private:
     //===================================================================
     //= private non-static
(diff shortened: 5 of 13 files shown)


apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -128,6 +131,18 @@ def __init__(
...
+        if soma_indexer_concurrency is not None:
+            clib.IntIndexer.start_thread_pool(soma_indexer_concurrency)
+        elif tiledb_ctx is not None:
+            tdb_concurrency = int(
+                tiledb_ctx.config().get("sm.compute_concurrency_level", 10)
+            )
+            thread_count = max(1, tdb_concurrency // 2)
+            clib.IntIndexer.start_thread_pool(thread_count)
+        else:
+            clib.IntIndexer.start_thread_pool(5)
```
thetorpedodog, line 145: This and `__del__` lead to several race conditions—if I have two threads, and an indexer running in one, then starting a second one will lead to the following:

1. Start indexer A. It starts the global thread pool.
2. Indexer A runs.
3. Start indexer B. It starts a new global thread pool, thus deleting the old one.
4. Indexer B runs. (What happens to the threads of indexer A that are still running under the old pool?)
5. Indexer A completes and is garbage collected (assuming some other problem does not arise). It deletes the thread pool entirely in `__del__`.
6. Indexer B cannot continue to run. (Does it crash? Just freeze?)

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -46,28 +46,37 @@ KHASH_MAP_INIT_INT64(m64, int64_t)
...
+    tiledb_thread_pool_ = std::make_unique<tiledbsoma::ThreadPool>(
+        thread_count);
+    LOG_DEBUG(fmt::format(
+        "[Re-indexer] Thread pool started with {} threads", thread_count));
+}
+
+void IntIndexer::stop_thread_pool() {
+    tiledb_thread_pool_.reset();
+    LOG_DEBUG("[Re-indexer] Thread stopped");
+}
```
thetorpedodog, line 62: Apart from the problems noted above, the way this thread pool is accessed globally would need lock protection—otherwise, the same pointer can be racily accessed and reset across multiple threads with no guarantee of atomicity, leading to the dreaded Undefined Behavior.

thetorpedodog (commented):
This looks like it is going to have one single, global thread pool, across contexts and everything else. Is this what we actually want?

bkmartinjr:
> This looks like it is going to have one single, global thread pool, across contexts and everything else. Is this what we actually want?

I think it would be better to have a per-SOMATileDBContext thread pool, following the tiledb.Ctx design pattern.

Open question: is the context passed to the indexer _init_ and bound at indexer create time or passed to get_indexer.  For Pandas compat, I prefer the former

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
beroy: @thetorpedodog thanks for the comments. Thinking about them, some clarifications:

1. The thread pool lifetime is tied to creation and deletion of soma context and not the indexers
2. The thread pool supports parallel indexers with no race issues assuming they are both attached to the same context (it uses indexers as keys). I can add unittests for that.
3. I believe the only situation that can lead to the race condition you mentioned is when we have two parallel indexers each with a different context. In that case the scenario you mentioned can take place

beroy:
@bkmartinjr, this PR assumes there is only one SOMATileDBContext existing at any given point. In other words, we support parallel indexers as long as there is no parallel (or overlapping) SOMATileDBContext. Does this assumption make sense?

libtiledbsoma/src/reindexer/reindexer.cc
beroy: I'll work on implementing the design I mentioned @thetorpedodog @bkmartinjr

thetorpedodog:
> @bkmartinjr, this PR assumes there is only one SOMATileDBContext existing at any given point. In other words, we support parallel indexers as long as there is no overlapping SOMATileDBContext. Does this assumption make sense? And wrt your question, the context is not bound to `get_indexer`.

No—the idea of the SOMATileDBContext is that multiple contexts may exist which have access to different sets of resources, etc., within the same process. Having implicit global state very much contravenes that design.

Since it’s possible to access Python objects from within C++ code, it seems like the indexer should be able to run its tasks using the thread pool that already exists within the context. One way to do this might be:

1. Expose the function we run within the thread pool to Python as a private (i.e., `_`-prefixed) member of the IntIndexer (something like `IntIndexer._lookup_internal(...)`, maybe?). (This might not be needed if you can easily pass an inner function as a Callable from C++ to Python; I am not sure how advanced pybind11 is.)
2. Where [the thread task is run](https://github.com/single-cell-data/TileDB-SOMA/blob/f4999d8ea89f7fcfb710380e64f3572acd01d5ca/libtiledbsoma/src/reindexer/reindexer.cc#L125), instead of using the internal threadpool, get the threadpool from the SOMATileDBContext and call `soma_thread_pool.submit(our_task, params, as, needed)`.
3. Join on the Future returned by the threadpool and examine the result.

bkmartinjr:
Apologies for slow response - out of pocket today.  I agree with Paul's summary of the design goal:
> idea of the SOMATileDBContext is that multiple contexts may exist which have access to different sets of resources

beroy:
@thetorpedodog and @bkmartinjr: @nguyenv is working on a way for the context threadpool to be directly accessible by the C++. So with that in place, on the C++ side,  when creating an indexer I just need to grab the the tiledb context threadpool and keep it in the indexer object to be used at the lookup time.

apis/python/src/tiledbsoma/_index_util.py
```diff
@@ -28,20 +20,10 @@ def tiledbsoma_build_index(
...
-           Concurrency level when the user does not want to use ``context``.
 
     Lifecycle:
         Experimental.
     """
-    if context is not None:
-        tdb_concurrency = int(
-            context.tiledb_ctx.config().get("sm.compute_concurrency_level", 10)
-        )
-        thread_count = max(1, tdb_concurrency // 2)
```
nguyenv, line 43: I'm looking at this again more closely, and what you can do instead is 
`tdb_concurrency = int(context.tiledb_config.get("sm.compute_concurrency_level", 10))`

Using `context.tiledb_ctx` will create a new context if it doesn't exist. Using `context.tiledb_config` will not. If you don't actually need to use the context and all you need to do is get this setting out of the config then just use the latter.

nguyenv:
> @thetorpedodog and @bkmartinjr: @nguyenv is working on a way for the context threadpool to be directly accessible by the C++. So with that in place, on the C++ side, when creating an indexer I just need to grab the the tiledb context threadpool and keep it in the indexer object to be used at the lookup time. This would require minimum changes to the indexer as well.

I think I need clarification. Currently, the there is a unique `TheadPool` created for each `IntIndexer`? And we want to modify this so that we create a `ThreadPool` for each `SOMATileDBContext`? And the `IntIndexer` uses the threadpool associated with the context its been given?

bkmartinjr:
> clarification

Yes, I believe you have it correct.  Goal is a thread pool associated with each context, and indexer lookups occur within that context (much like queries, etc), submitting to the context's threadpool.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
nguyenv: Already discussed in DMs but x-posting here: Instead of a global map, we should add `tiledb_threadpool` as a property of `TileDBSOMAContext`. It could also be lazy loaded just like how we do with `TileDBSOMAContext.tiledb_ctx`. Unless I'm misunderstanding something below, there isn't already a preexisting threadpool that can be directly accessed via `tiledb::Context` and has to be instantiated within `TileDBSOMAContext` in Python or the upcoming `SOMAContext` in C++.

beroy:
Now with @nguyenv work on sharing context between C++ and python, I think I can just achieve this goal by adding threadpools to C++ TileDBSOMAContext so each indexer keeps track of its own context + threadpool (lazily create the pool when lookup is needed for the first time)

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
thetorpedodog: We have the start of a thread pool in somacore (`threadpool` member on the `ContextBase` protocol), but it seems to have not made it fully to TileDB-SOMA yet? https://github.com/single-cell-data/SOMA/blob/76b2d2bbca7c8baf543b5cc9cf71fc5c9138a39a/python-spec/src/somacore/types.py#L87

nguyenv: OIC - that makes more sense now. Yeah `SOMATIleDBContext` doesn't inherit from `soma.ContextBase` yet, but that'll be easy to add and then we can start utilizing the context's threadpool.

beroy: OK, for the time being I'm gonna implement the design with the thread on top of the C++ Soma context.

beroy:
@bkmartinjr, @thetorpedodog and @nguyenv I update the PR with rebasing to @nguyenv C++ context and used that lazily create threadpool.

bkmartinjr:
@beroy - can I suggest you base this PR on the `shared_threadpool_indexer` branch so your changes are more evident?

beroy:
@bkmartinjr sorry about the mix up. I jsut rebased the PR against Vivan's PR as all my work is against her PR. I think now my changes are clear.

bkmartinjr:
> rebased the PR

Thank you! Greatly appreciated.

apis/python/src/tiledbsoma/_index_util.py
```diff
@@ -40,8 +40,8 @@ def tiledbsoma_build_index(
         tdb_concurrency = int(
```
bkmartinjr, line 40: this all looks like dead code, or possibly wrong:
* if no context, it should use the default context
* the code to calculate thread count is dead code and can be removed

nguyenv:
Where / when does the threadpool get created? It looks like it's set to the `nullptr` but then not ever constructed anywhere.

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -79,18 +77,33 @@ void IntIndexer::map_locations(
...
+    context_ = context;
 }
 
 void IntIndexer::lookup(const int64_t* keys, int64_t* results, int size) {
     if (size == 0) {
         return;
     }
-    if (tiledb_thread_pool_ == nullptr) {  // When concurrency is 1
+    // Creating threadpool lazily when the first lookup happens
+    if (context_ != nullptr && context_->thread_pool() == nullptr) {
```
bkmartinjr, line 91: this code doesn't look thread safe to me -- e.g., if there are two concurrent calls to `lookup()`, they will both attempt to create a thread pool and write to `context_`.  This needs to be a critical section.

TL;DR - contexts must be multi-thread safe

bkmartinjr: Related:  it seems to me that the creation of the thread pool for a context should be code that lives _inside_ the SOMAContext C++ class, not outside of it.   @nguyenv - thoughts?

nguyenv: oic - this is where the threadpool gets created.  Yes I agree that it makes more sense to be in the C++ code.

bkmartinjr: and in a critical section, right?  (unless I grossly misunderstand, it is acceptable for multiple threads to all use the same SOMAContext much like tiledb.Ctx)

bkmartinjr (changes requested):
left various comments inline:
* dead code in _index_util.py
* need to centralize thread pool creation inside the SOMAContext C++ class so it is centralized
* missing critical section for thread pool creation

libtiledbsoma/src/reindexer/reindexer.cc
nguyenv: Yes that's correct. `SOMAContext` should behave just like `tiledb::Context` which includes running multiple threads using the same context, so we should create the threadpool in a thread-safe manner.

beroy: Great catch. The older version of lookup was thread safe but with shared context that's no longer the case. Moving to context makes sense. My goal here was merely lazily creation.

apis/python/src/tiledbsoma/_index_util.py
beroy: You're right.

nguyenv:
Does this conflict with work in https://github.com/single-cell-data/TileDB-SOMA/pull/2001?

libtiledbsoma/src/soma/soma_context.cc
```diff
@@ -0,0 +1,49 @@
+/**
+ * @file   soma_dataframe.h
```
nguyenv, line 2:
```suggestion
 * @file   soma_context.cc
```

libtiledbsoma/src/soma/soma_context.cc
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2023 TileDB, Inc.
```
nguyenv, line 8:
```suggestion
 * @copyright Copyright (c) 2024 TileDB, Inc.
```

beroy:
@nguyenv I wouldn't call it a conflict. I think if we later want to use python threadpool, we always have that option.

libtiledbsoma/src/soma/soma_context.h
```diff
@@ -47,10 +48,14 @@ class SOMAContext {
...
-        : ctx_(std::make_shared<Context>(Config({}))){};
+        : ctx_(std::make_shared<Context>(Config({}))) {
+        create_thread_pool();
+    };
 
     SOMAContext(std::map<std::string, std::string> platform_config)
-        : ctx_(std::make_shared<Context>(Config(platform_config))){};
+        : ctx_(std::make_shared<Context>(Config(platform_config))) {
+        create_thread_pool();
+    };
```
nguyenv, line 59: I don't think we should create the threadpool whenever we construct the `SOMAContext`. Only when we need to use it.

libtiledbsoma/src/soma/soma_context.h
```diff
@@ -67,13 +72,22 @@ class SOMAContext {
         return cfg;
     }
 
+    std::shared_ptr<ThreadPool>& thread_pool() {
+        return thread_pool_;
+    }
```
nguyenv, line 77: I'm thinking something like if `thread_pool_ == nullptr` then create it. Right now you're lazy loading `SOMAContext` in the reindexer class which does technically lazy load the threadpool. But I don't think you need to lazy load the context in the `IntIndexer`. Just the threadpool in `SOMAContext` if that makes sense.

bkmartinjr (commented):
I am confused by the IntIndexer state handling for context. 
* there is a private instance variable called `context_`
* there is a `instance` argument provided to map_locations
* the constructor doesn't take a context
* map_locations accepts a context, and sets the instance variable with it (without a lock, which will race)

What I expected to see is:
* The IntIndexer constructor accepts a context.  This can be done w/o a lock, as construction will not race
* map_locations uses the instance value (`context_`) and does not accept an argument called `context`
* the Python calls match this

In other words, the code in _index_util.py would look like:

```
    reindexer = clib.IntIndexer(context)
    reindexer.map_locations(keys)
    return reindexer  # type: ignore[no-any-return]
```

This solves both the race condition, and makes the API match other uses of context

libtiledbsoma/src/soma/soma_context.h
beroy: Yes, you are right. that makes sense. I should not load lazily.

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -151,8 +150,11 @@ IntIndexer::~IntIndexer() {
...
-IntIndexer::IntIndexer(const int64_t* keys, int size, int threads) {
-    map_locations(keys, size, threads);
+IntIndexer::IntIndexer(
+    const int64_t* keys,
+    int size,
+    std::shared_ptr<tiledbsoma::SOMAContext> context) {
+    map_locations(keys, size, context);
 }
 
 }  // namespace tiledbsoma
```
bkmartinjr, line 160: nit - missing EOL

libtiledbsoma/src/reindexer/reindexer.h
```diff
@@ -53,9 +54,14 @@ class IntIndexer {
      * @param size yhr number of keys in the put
      * @param threads number of threads in the thread pool
      */
-    void map_locations(const int64_t* keys, size_t size, size_t threads = 4);
-    void map_locations(const std::vector<int64_t>& keys, size_t threads = 4) {
-        map_locations(keys.data(), keys.size(), threads);
+    void map_locations(
+        const int64_t* keys,
+        size_t size,
+        std::shared_ptr<tiledbsoma::SOMAContext> context);
```
bkmartinjr, line 60: see other comment related to this -- the context should be provided and set during construction, not in a call to map_locations

beroy: @bkmartinjr I think the code for context looks a bit confusion. Based on a discussion I had with @nguyenv the context is created beforehand and when the call to `pybind` happens we get a copy of the context (as a shared_ptr). So the call to amp_location gets the same context that is passed to `tiledbsoma_build_index(keys, context.native_context)` which is created previously. The only reason to that have to pass the context to map_location is to keep a pointer to it (as lookup doesn't get called with context) and use its threadpool during lookup. @nguyenv does it make sense?

bkmartinjr: To clarify, I mean it should be passed to the IntIndexer during IntIndexer construction.

In other words, with a pre-existing context, code in index_util would look like:
```
    reindexer = IntIndexer(context)
    reindexer.map_locations(keys)
    return reindexer
```

Without this, it allows the possibility that the IntIndexer context is mutable after IntIndexer creation, and I'm pretty sure that is not desirable

libtiledbsoma/src/soma/soma_context.h
nguyenv: I would still prefer to see `create_thread_pool` get called in this `SOMAContext` class rather than in the `IntIndexer`. As mentioned above by @bkmartinjr, we'll also want this to be thread-safe - I would probably look into using `std::unique_lock`.

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -142,7 +135,14 @@ void IntIndexer::lookup(const int64_t* keys, int64_t* results, int size) {
...
             end));
     }
-    tiledb_thread_pool_->wait_all(tasks);
+    thread_pool_->wait_all(tasks);
+}
+
+IntIndexer::IntIndexer(std::shared_ptr<tiledbsoma::SOMAContext> context) {
+    if (context != nullptr) {
+        context->create_thread_pool();
+        thread_pool_ = context->thread_pool();
```
nguyenv, line 144: I don't think we should call `create_thread_pool` in here. See other comment.

libtiledbsoma/src/soma/soma_context.h
bkmartinjr: I fully agree

johnkerl:
[sc-42198]

beroy:
@nguyenv and @bkmartinjr I updated the PR based on your last set of reviews. Will appreciate feedback.

libtiledbsoma/src/reindexer/reindexer.h
```diff
@@ -74,24 +75,16 @@ class IntIndexer {
...
     /*
      * The created 64bit hash table
      */
     kh_m64_s* hash_;
-    /*
-     * TileDB threadpool
-     */
-    std::shared_ptr<tiledbsoma::ThreadPool> tiledb_thread_pool_ = nullptr;
+
+    std::shared_ptr<ThreadPool> thread_pool_ = nullptr;
```
bkmartinjr, line 87: Probably a question for @nguyenv ... should resources like the IntIndexer hold a reference to the ThreadPool, or to the Context?  I had assumed that when the Context is destructed, it would also tear down the thread pool.  I don't _think_ the above would cause problems as it uses a shared_ptr, but wanted to confirm...

beroy: Actually it was originally her suggestion that IntIndex holds the reference to the threadpool.

libtiledbsoma/src/soma/soma_context.h
bkmartinjr: @beroy - did this comment get addressed?  Unless I misunderstand the code, it it not lazy?

libtiledbsoma/src/reindexer/reindexer.h
bkmartinjr: Great - ty for clarifying!

bkmartinjr (commented):
My primary remaining question is around lazy-creation of the thread pool.  It looks like @nguyenv recommended this, and it makes sense to me.  As I read the code, it looks like that was not completed?

libtiledbsoma/src/reindexer/reindexer.h
nguyenv: There might have been some miscommunication because I don't think the `ThreadPool` should reside in here but in the `SOMAContext`.

libtiledbsoma/src/soma/soma_context.h
nguyenv: To clarify, what I'm expecting in the implementation file is something like:
```
    std::shared_ptr<ThreadPool>& SOMAContext::threadpool() {
        std::unique_lock;
        if(thread_pool_ == nullptr)
             create_thread_pool();
        return thread_pool_;
    }
```

nguyenv: I agree that we don't actually really need a separate `create_thread_pool()` method though. Was just for the example.

bkmartinjr: I'm on the same page...this is what I was imagining as well.

bkmartinjr (approved):

libtiledbsoma/src/reindexer/reindexer.h
```diff
@@ -74,12 +74,8 @@ class IntIndexer {
...
     }
     IntIndexer(){};
-    /**
-     * Constructor with the same arguments as map_locations
-     */
-    IntIndexer(const int64_t* keys, int size, int threads);
-    IntIndexer(const std::vector<int64_t>& keys, int threads)
-        : IntIndexer(keys.data(), keys.size(), threads) {
+    IntIndexer(std::shared_ptr<tiledbsoma::SOMAContext> context) {
+        context_ = context;
```
nguyenv, line 78: I think that we should always have a context.

```
    IntIndexer() : context_(make_shared<SOMAContext>()){}; // or IntIndexer() = delete;?
    IntIndexer(std::shared_ptr<tiledbsoma::SOMAContext> context) : context_(context) {};
```

libtiledbsoma/src/reindexer/reindexer.cc
```diff
@@ -90,7 +82,9 @@ void IntIndexer::lookup(const int64_t* keys, int64_t* results, int size) {
     if (size == 0) {
         return;
     }
-    if (tiledb_thread_pool_ == nullptr) {  // When concurrency is 1
+    // Single thread checks
+    if (context_ == nullptr || context_->thread_pool() == nullptr ||
+        context_->thread_pool()->concurrency_level() == 1) {
```
nguyenv, line 87: I think we can just check `context_->thread_pool()->concurrency_level() == 1`

libtiledbsoma/src/soma/soma_context.h
```diff
@@ -67,13 +71,30 @@ class SOMAContext {
...
+    std::shared_ptr<ThreadPool>& thread_pool() {
+        const std::lock_guard<std::mutex> lock(thread_pool_mutex_);
+        // The first thread that gets here will create the context thread pool
+        if (thread_pool_ == nullptr) {
+            create_thread_pool();
+        }
+        return thread_pool_;
+    }
+
+    void create_thread_pool();
```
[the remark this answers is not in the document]
beroy: Why is that?

libtiledbsoma/src/reindexer/reindexer.h
beroy: I wouldn't create a new context in IntIndexer but agree with other suggestions.

libtiledbsoma/src/reindexer/reindexer.cc
beroy: I get your point but in cases like this, I really prefer to be ultra conservative

johnkerl:
Backports failed, again :(

PR #2151, opened 2024-02-16
Fix inconsistency in DictFilterSpec docstring.

thetorpedodog:
We mistakenly used `_name` in the literal rather than `_type`.


nguyenv (approved):

thetorpedodog:
> We mistakenly used `_name` in the literal rather than `_type`.

for the record we are using the royal we here

PR #2152, opened 2024-02-17
[r] Port blockwise iterator/reader to R

### LINKED ISSUE #1853: [r] Port blockwise sparse iterator from Python to R
Port #1792

mojaveazure:
Implement the blockwise iterator and reader for the R API

This PR parallels #1792; it implements new classes for blockwise iteration through a SOMA sparse nd-array. Blockwise iteration is implemented through `SOMASparseNDArrayRead$blockwise()` (paralleling the Python implenetation) and enabled for Arrow tables (`$blockwise()$tables()`) and COO sparse matrices (`$blockwise()$sparse_matrix()`)

New classes:
- `CoordsStrider`: new class to iterate through coordinate similar to Python's `_coords_strider`
- `SOMASparseNDArrayReadBase`: base class for sparse array reads
- `SOMASparseNDArrayBlockwiseRead`: new reader class for blockwise iterated reads
- `BlockwiseReadIterBase`: base class for blockwise iteration
- `BlockwiseTableReadIter`: blockwise iterator returning Arrow tables
- `BlockwiseSparseReadIter`: blockwise iterator returning sparse matrices

New SOMA methods:
- `SOMASparseNDArrayRead$blockwse()`: perform a blockwise read

resolves #1853


apis/r/R/BlockwiseIter.R
```diff
@@ -52,9 +52,10 @@ BlockwiseReadIterBase <- R6::R6Class(
     },
     read_complete = function() !self$coords_axis$hasNext() || super$read_complete(),
     read_next = function() {
-      #if (is.null(private$soma_read_pointer)) {  # TODO: FIXME
-      #  return(NULL)
-      #}
+      message("blockwise read next")
+      if (is.null(private$soma_reader_pointer)) {
```
eddelbuettel, line 56: Nice catch.

apis/r/R/BlockwiseIter.R
```diff
@@ -93,18 +92,10 @@ BlockwiseTableReadIter <- R6::R6Class(
   classname = "BlockwiseTableReadIter",
   inherit = BlockwiseReadIterBase,
   public = list(
-    concat = function() {
-      message("blockwise table concat")
-      # TODO: implement concat() of blockwise table iterator
-      .NotYetImplemented()
-    }
+    concat = function() soma_arraw_to_arrow_table_concat(self)
```
eddelbuettel, line 95: arraw?  arrow!

apis/r/R/utils-readerTransformers.R
```diff
@@ -12,6 +12,16 @@ soma_array_to_arrow_table <- function(x) {
   )
 }
 
+soma_arraw_to_arrow_table_concat <- function(it) {
```
eddelbuettel, line 15: More arraw

apis/r/R/BlockwiseIter.R
mojaveazure: 🤦 One day I'll learn to type

aaronwolen:
[sc-42211]

apis/r/R/SOMASparseNDArrayRead.R
```diff
@@ -7,51 +117,153 @@
...
+      ...,
+      size,
+      reindex_disable_on_axis = NULL
+    ) {
+      super$initialize(sr, array, coords)
+      stopifnot(
+        "'size' must be a single integer value" = is.null(size) ||
+          rlang::is_integerish(size, 1L, finite = TRUE) ||
+          (inherits(size, 'integer64') && length(size) == 1L && is.finite(size)),
+        "'reindex_disable_on_axis' must be  avector of integers" = is.null(reindex_disable_on_axis) ||
```
mlin, line 220:
```suggestion
        "'reindex_disable_on_axis' must be a vector of integers" = is.null(reindex_disable_on_axis) ||
```

mlin:
@mojaveazure @eddelbuettel Thanks for all the work on this! One high-level question at this stage:

The Python `.blockwise().scipy()` iterator returns a `csr_matrix` (when stepping by the rows) or a `csc_matrix` (stepping by the columns, less common) where here it looks like we're always returning a COO (`dgTmatrix`). I personally don't have a strong opinion that the R version needs to reflect that exactly; but I thought we should at least discuss it here.

*In Python* (scipy) `coo_matrix` doesn't support all the operations that `csr_matrix` does, vector math ops in particular; so the latter is usually a little more convenient for the user.

However, also *in Python*, the internal representation of `csr_matrix` unfortunately uses memory proportional to the largest row index, not the number of actual rows. The result is that the reindexing feature is practically necessary for large datasets, to keep the memory usage of the returned `csr_matrix` objects under control. The reindexing is also practically helpful just to know what you're getting in each block.

I emphasized *in Python* because of course I don't know if we're working under the same design constraints in R.

apis/r/R/SOMASparseNDArrayRead.R
eddelbuettel: Thanks @mlin, corrected in https://github.com/single-cell-data/TileDB-SOMA/pull/2152/commits/839411a3433cd181f91b6d1642562fd4cfcddfe2

eddelbuettel:
@mlin There is a (currently unused) argument `repr` for the sparse matrix layout.  We use 'T' (Matrix code for COO, internally called 'dgTMatrix') as it is the format on disc.  The very simple change

```diff
modified   apis/r/R/BlockwiseIter.R
@@ -193,7 +193,7 @@ BlockwiseSparseReadIter <- R6::R6Class(
       coords,
       axis,
       ...,
-      repr = "T",
+      repr = c("T","R","C"),
       reindex_disable_on_axis = NULL
     ) {
       super$initialize(
```

permits return also of, respectively, a row- or column-compressed variant (dgRMatrix or dgCMatrix) corresponding to those compression formats.

mojaveazure:
@mlin I would vote for only returning a COO matrix. The existing `SparseReadIter` class hardcodes a return representation of COO and the current blockwise behavior is in-line with that https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/r/R/SparseReadIter.R#L20-L31

I can see a case for removing the `repr` parameter from the blockwise side, espeically since the existing `SparseReadIter` does not have a `repr` parameter, but until we get alternative representations flowing through the existing `SparseReadIter` I think the blockwise should only return COO

mojaveazure:
Also, @eddelbuettel's fix isn't as simple as expanding the allowed values to `repr`; it needs to be plumbed through and constantly checked for (`Matrix` has a bad habit of silently changing the representations based on what it deems most efficient; for example, `conat()` uses `+` to grow the resulting matrix, and row-based growth often results in CSC/`CsparseMatrix` getting cast to CSR/`RsparseMatrix`)

eddelbuettel:
(Well it passed the unit test where we do `$read_next()$concat()` -- that appears to wire through.)

But thumbs for 'simpler is better'.  COO seems fine as default.

mlin:
@mojaveazure @eddelbuettel Ok, no objection from me to COO-only in principle, subject to:

What are we thinking about reindexing? As I mentioned, *in Python* it's practically essential, albeit because of an implementation detail in `csr_matrix`.

Besides that peculiarity, it's still useful for the iterator to tell you the range of the major axis (the one being strided) you're getting in each block. Reindexing is one, not the only, way of providing that info. Without, imagine getting the sparse matrix with a huge shape, and you know only one small block/stripe in it is populated, but you don't know where that is.

Then the minor axis- there are some use cases for which it is and isn't helpful to reindex that too. I'm less concerned about that.

mojaveazure:
Reindexing would be useful, and essential to offering CSC/CSR output in R as well. However, that has yet to be ported to R, so I don't think we can offer that now.

As for knowing the range of the major axis, that poses different problem in R, and one I hadn't thought of. R does not allow tuple outputs, and has no native unpacking like Python. The ways around this that I can think of are:
- returning a list with the values in one entry and the major-axis indices in another (`list(block = read_next(), indices = indices, axis = axis)`); this would change the return type and make it harder to move code from the existing iterators to the blockwise iterator
- store the indices as an attribute on the return values; this would allow the return type of the blockwise iterators to be the same as the existing iterators
```R
block <- read_next()
attr(block, "indices") <- indices
attr(block, "axis") <- axis
```
- offer a mode to extract just the indices returned
```
block <- read_next()
if (axis == 0) {
  return(block[indices, ])
else {
  return(block[, indices])
}
```

As for minor-axis, this PR currently doesn't do anything there as reindexing is not a part of this PR

aaronwolen:
As we discussed, overall this looks great! 

I do think we need to think carefully about @mlin's comment:

> ...it's still useful for the iterator to tell you the range of the major axis (the one being strided) you're getting in each block.

and your suggestion to return a list with the values might be the way to go. Perhaps @bkmartinjr and/or @pablo-gar could weigh in since this is used in the cellxgene-census package.

eddelbuettel:
The single-return-object constraint is indeed real.  In other (simpler) contexts I have often used the `list()` method but its lack of elegance is really stark.  Given how refined a class structure we built (see below), it seems options 2 or 3 above are more natural?

tldsm

bkmartinjr:
Re: reindexing - it has uses outside of generating CSR/CSC.  For example, if you want accumulate across a slice (say compute mean X value for a cell or gene slice), it is much easier to do it with reindexing.  Without it, the user effectively has to do their own reindexing implementation to map between the joinids and the offset-indexed accumulator (or use a more complicated accumulator data structure, such as a hash table).  I've found that many of my uses for blockwise benefit from reindexing on one or the other axis.

I am not opposed to breaking large changes into multiple PRs, and am not advocating that reindexing necessarily be included in this PR. But I do think that many blockwise use cases strongly benefit from reindexing, and not supporting the feature pushes the task to the API user.

TL;DR - it isn't just a CSC/CSR thing.

eddelbuettel:
I am leaning towards decomposing this into multiple steps.   

We demonstrated last summer that accessing the C++ reindexing facitly is straightforward: we are passing standard vectors around.   But placing it in the already somewhat complicated R class structure and 'API' modeled after Python is not that obvious (at least to me) so it may be easier to deal with that in a second step given that we now have blockwise access in this PR.  Finishing it may be a good step en route to more complete support.

johnkerl (commented):
This PR has stagnated. For the sake of forward motion, let's get this PR merged and let's continue buillding the second half of it on a subsequent PR, which we all agree we need in order for this to be a complete deliverable.

mlin:
SGTM. I think we're all agreed that, in a subsequent installment, the `BlockwiseSparseReadIter` will need to produce not only the `sparseMatrix` object but also the reindexing vectors, on each iteration. I'd like to defer to @eddelbuettel and @mojaveazure to agree on the preferred way to produce multiple values like that, just to ensure we won't need to break the API soon after merging this.

I'd also like to note that it's not necessary to use the optimized C++ reindexer immediately. A slower version written in pure (vectorized) R would be fine to start with.

johnkerl (approved):
Re-approving, as discussed

eddelbuettel:
I don't think the :x: on R is related to the PR.  It is in one of those spots where we have seen it pop up, and disappear, without any hints to resolution which ... has been frustrating for all us as we tried. Again and again.

It all passes here. No fail, no skip, so it remain puzzling.

```sh
# earlier lines omitted

══ Results ═════════════════════════════
Duration: 127.2 s

[ FAIL 0 | WARN 0 | SKIP 0 | PASS 2669 ]
edd@rob:~/git/tiledb-soma/apis/r(ph/feat/blockwise-reader)$ 
```

johnkerl:
@eddelbuettel re the ❌  I did some debug with @mojaveazure yesterday and haven't typed up the result yet.

Agreed it's green on your machine; we need green here. I might not get a write-up until Monday.

eddelbuettel:
@johnkerl Eager to hear any conclusion.  The package has bombed in that spot before, it went away, it came back.  Would be nice to nail it.

eddelbuettel:
For completeness same thing on my AWS instance running 22.04 (after making sure I had the update `pbmc3k.tiledb` as well as added `iterators` and `itertools`:

```sh
# ...
══ Results ════════════════════════════════
Duration: 193.8 s

[ FAIL 0 | WARN 0 | SKIP 0 | PASS 2669 ]
ubuntu@aws:~/git/tiledb-soma/apis/r(ph/feat/blockwise-reader)$ 
ubuntu@aws:~/git/tiledb-soma/apis/r(ph/feat/blockwise-reader)$ git ls | head -5
* f4d27c10 - (HEAD -> ph/feat/blockwise-reader, origin/ph/feat/blockwise-reader) Correct typo spotted by @mlin (26 hours ago) <Dirk Eddelbuettel>
* 1c4f29de - Add error messages to `stopifnot()` calls (26 hours ago) <Paul Hoffman>
* 289a97a5 - Fix typo (26 hours ago) <Paul Hoffman>
* 4eeefbc4 - Clean up `soma_array_to_sparse_matrix_concat()` Have `SparseReadIter$concat()` use new helper function (26 hours ago) <Paul Hoffman>
* dffd0e31 - Make tests more explicit with `expect_s3_class()`/`expect_s4_class()` (26 hours ago) <Paul Hoffman>
ubuntu@aws:~/git/tiledb-soma/apis/r(ph/feat/blockwise-reader)$ 
```

johnkerl:
Yesterday @mojaveazure and I were able to repro the segfault within a clean docker-run environment.

Using
```
CI="." VALGRIND_OPTS="-s --trace-children=yes --max-threads=1024" R CMD check --use-valgrind --no-vignettes --no-manual *.tar.gz
```
we saw this in the valgrind output:

<details>

```
Mismatched free() / delete / delete []
   at 0x484B27F: free (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
   by 0x25474CDF: tiledbsoma::ArrowAdapter::release_array(ArrowArray*) (in /TileDB-SOMA/libtiledbsoma/build/libtiledbsoma/src/libtiledbsoma.so)
   by 0x25310B5D: ArrowArrayRelease (nanoarrow.c:1770)
   by 0x2012DD66: ??? (in /usr/lib/R/site-library/arrow/libs/arrow.so)
   by 0x1FA49879: std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() (in /usr/lib/R/site-library/arrow/libs/arrow.so)
   by 0x2012FE75: ??? (in /usr/lib/R/site-library/arrow/libs/arrow.so)
   by 0x1FB0921D: std::_Sp_counted_ptr_inplace<arrow::ArrayData, std::allocator<arrow::ArrayData>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() (in /usr/lib/R/site-library/arrow/libs/arrow.so)
   by 0x200A2D8D: arrow::SimpleRecordBatch::~SimpleRecordBatch() (in /usr/lib/R/site-library/arrow/libs/arrow.so)
   by 0x1FAA4699: cpp11::external_pointer<std::shared_ptr<arrow::RecordBatch>, &(void cpp11::default_deleter<std::shared_ptr<arrow::RecordBatch> >(std::shared_ptr<arrow::RecordBatch>*))>::r_deleter(SEXPREC*) (in /usr/lib/R/site-library/arrow/libs/arrow.so)
   by 0x49EF5CC: R_RunWeakRefFinalizer (in /usr/lib/R/lib/libR.so)
   by 0x49EF82B: ??? (in /usr/lib/R/lib/libR.so)
   by 0x49ADCD9: Rf_eval (in /usr/lib/R/lib/libR.so)

 Address 0xdc3792b0 is 0 bytes inside a block of size 24 alloc'd
   at 0x484A2F3: operator new[](unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
   by 0x25474592: tiledbsoma::ArrowAdapter::to_arrow(std::shared_ptr<tiledbsoma::ColumnBuffer>) (in /TileDB-SOMA/libtiledbsoma/build/libtiledbsoma/src/libtiledbsoma.so)
   by 0x2534D2B6: sr_next(Rcpp::XPtr<tiledbsoma::SOMAArray, Rcpp::PreserveStorage, &(void Rcpp::standard_delete_finalizer<tiledbsoma::SOMAArray>(tiledbsoma::SOMAArray*)), false>) (riterator.cpp:219)
   by 0x252FB4A1: _tiledbsoma_sr_next (RcppExports.cpp:139)
   by 0x49562DD: ??? (in /usr/lib/R/lib/libR.so)
   by 0x495685C: ??? (in /usr/lib/R/lib/libR.so)
   by 0x49AE367: Rf_eval (in /usr/lib/R/lib/libR.so)
   by 0x49B1377: ??? (in /usr/lib/R/lib/libR.so)
   by 0x49AE0FF: Rf_eval (in /usr/lib/R/lib/libR.so)
   by 0x49AFD85: ??? (in /usr/lib/R/lib/libR.so)
   by 0x49B0BB4: Rf_applyClosure (in /usr/lib/R/lib/libR.so)
   by 0x49ADE2B: Rf_eval (in /usr/lib/R/lib/libR.so)
```

</details>

the TL;DR of which is

```
Mismatched free() / delete / delete []
free
   by 0x25474CDF: tiledbsoma::ArrowAdapter::release_array libtiledbsoma.so
new
   by 0x25474592: tiledbsoma::ArrowAdapter::to_arrow libtiledbsoma.so
   by 0x2534D2B6: sr_next riterator.cpp:219
```

And indeed we see the `free` here
https://github.com/single-cell-data/TileDB-SOMA/blob/f4d27c10ec7225d164cf0d78bd3fa6ace88e8666/libtiledbsoma/src/utils/arrow_adapter.cc#L92
and the `new` here
https://github.com/single-cell-data/TileDB-SOMA/blob/f4d27c10ec7225d164cf0d78bd3fa6ace88e8666/libtiledbsoma/src/utils/arrow_adapter.cc#L301
which appears to be more of the same of #2260

I am putting up a separate PR (atop this one) to validate

eddelbuettel:
But these a Arrow data structures which use a C interface structures.  So it needs to be `malloc` and `free` with the one (clearly commented) exception 

https://github.com/single-cell-data/TileDB-SOMA/blob/6abaf6698a6b409037d6497ddf48c0d36c79f6f3/libtiledbsoma/src/utils/arrow_adapter.cc#L273-L283

which is freed here

https://github.com/single-cell-data/TileDB-SOMA/blob/6abaf6698a6b409037d6497ddf48c0d36c79f6f3/libtiledbsoma/src/utils/arrow_adapter.cc#L80-L83

Everything else should be matching `malloc` and `free` as it is throughout the rest of that code (and see the release helpers).  

At least that has been our thinking (as discussed with other colleagues too).  But hey I am easy.  If it makes the Heisenbug go away I am all for.

The embedding of the `shared_ptr<>` is however something we may want to revisit. Allocating Arrow buffers, running the query with them and then accessing results zero-copy is a valid plan.  For another time.

johnkerl:
@nguyenv as you noted at https://github.com/single-cell-data/TileDB-SOMA/pull/2259#issuecomment-1995555308 it would be ideal to remove _all_ instances of `new`, `new[]`, `delete`, and `delete[]` from this file.

A few things needing to be done:

* Shockingly, although there's a lot of `free` in today's `arrow_adapter.cc`, there is no `calloc` or `malloc` at all
  * Much of the allocation is done by `strdup` which calls `malloc` internall
* We need within `arrow_adapter.cc` which calls `calloc` (_not_ `malloc` as I do not want any uninitialized-memory issues)
  * That helper needs a uniform handling for out-of-memory cases
* That will handle most sites. However: what `new` and `delete` do is memory allocation _and_ invoking constructors/destructors.
  * Here's the [one spot](https://github.com/single-cell-data/TileDB-SOMA/blob/f4d27c10ec7225d164cf0d78bd3fa6ace88e8666/libtiledbsoma/src/utils/arrow_adapter.cc#L283) I've been head-scratching around, as it indeed invokes a constructor on a `struct` which has a `shared_ptr` in it ...

eddelbuettel:
@johnkerl That is not entirely correct. The `strdup` is just used in less than a handful of cases for arrow object 'schema' and 'name' strings.  The structures themselves are allocated by helper functions and offloaded to `nanoarrow` (in other places including tiledb-r and my nanoarrow refactor branch here, also held up by a Heisenbug.  Here we allocated
via

https://github.com/single-cell-data/TileDB-SOMA/blob/f4d27c10ec7225d164cf0d78bd3fa6ace88e8666/libtiledbsoma/src/utils/arrow_adapter.cc#L258-L259

and then take ownership via `std::move()` on the exit

https://github.com/single-cell-data/TileDB-SOMA/blob/f4d27c10ec7225d164cf0d78bd3fa6ace88e8666/libtiledbsoma/src/utils/arrow_adapter.cc#L386

That is for the outer schema and array data structures. Now when there is an enumeration, we have an embedded data structure and these two are wrong

https://github.com/single-cell-data/TileDB-SOMA/blob/f4d27c10ec7225d164cf0d78bd3fa6ace88e8666/libtiledbsoma/src/utils/arrow_adapter.cc#L331-L332

and is likely this

https://github.com/single-cell-data/TileDB-SOMA/blob/f4d27c10ec7225d164cf0d78bd3fa6ace88e8666/libtiledbsoma/src/utils/arrow_adapter.cc#L356

But we are now comingling reviewing the arrow treatment (for which I have a branch, which is stalled for reasons I do not understand _at all_ failing _only_ in coverage (which, because the world is a mean place, runs yet another build/test setup).  I spent a full day on that yesterday, I have nothing to show.  If you want to try your `valgrind` magic (I seem to have a different option `--leak-check=full` in my ~/.valgrindrc which may be why valgrind dies on me (!!) when running the full test battery.

johnkerl:
@eddelbuettel thanks. I can recommend `--leak-check=full` when chasing memory leaks; here I am not chasing a memory leak so that option's output isn't helpful here.

johnkerl:
I also often have `valgrind` dying (including the infamous "the impossible happened!") -- which is the reasoning behind my current recommend of
```
valgrind --trace-children=yes -s --max-threads=1024
```
or even
```
valgrind --trace-children=yes -s --max-threads=2048
```
...

johnkerl:
> and these two are wrong

> and is likely this

We are in agreement. I am advocating for getting rid of _all_ `new` and `delete` within `arrow_adapter.cc`. And I noted a concern around one particular callsite, above, on which I defer to @nguyenv .

johnkerl:
> But we are now comingling reviewing

@eddelbuettel this PR has persistently red CI. I am unwilling to commit that to `main`, because once `main` has persistently red CI, our QA experience (candidly -- mainly my own as project manager and release manager) becomes unacceptable. I'm unwilling to "let the red spread" beyond this still-unmerged PR.

eddelbuettel:
By all means commit a change, test it, and maybe you get the winner.  But some things you jotted do not apply (yes, I also refer `calloc` over `malloc` in general and in principle, here it is irrelevant as `arrow` overwrites _all_ bytes) and I fear that in the above you also cross an allocation to `array->buffers` with an argument over freeing `dict->buffer`.  But I am all for experimenting: we could split off a branch and if, say, commenting out all `delete` helps.

Lastly, this is a long-running PR and manifestedly was not 'always red' as a [sorted display](https://github.com/single-cell-data/TileDB-SOMA/actions/workflows/r-ci.yml?query=branch%3Aph%2Ffeat%2Fblockwise-reader) will show.  It had a fair number, and then it didn't.  Which makes is quite frustrating.  So I am for nuking the cause and I really wish I knew what the cause was.

eddelbuettel:
> I am advocating for getting rid of all new and delete

Again, one is in there by design and has from the beginning, see [my comment](https://github.com/single-cell-data/TileDB-SOMA/pull/2152?notification_referrer_id=NT_kwDOAApFYbE5NTAzNDU5ODE2OjY3MzEyMQ#issuecomment-2027691963) above.

Commenting out all others may be worth a try.  I did such tests here in all those weeks working on it, it generally did not survive a first local testing path.  Maybe you have more luck.

johnkerl (changes requested):
Looks good at the app level; just doing request-changes explicitly to ensure clarity on the fact that we have persistently red CI on this PR and we need it green in order to merge.

eddelbuettel:
You keep calling it persistently red and it wasn't:

image

So maybe we got ourselves a regression in when rebasing?  I wish I knew!

johnkerl:
@eddelbuettel it's red now, stuck red, and I don't know how to make it green other than #2352 and that is also red.

eddelbuettel:
Yes #2352 may have confused array->buffers and dict-buffers.  But I have nothing have better, sadly.

nguyenv:
I should have said we should move everything towards the C++ `new` and `delete`. There is really no reason to be using `malloc` in C++. You can use `new` with C structure as well so that is not an issue. 

Though thinking about this more, I actually don't know an alternative for `strdup`/`free`...

eddelbuettel:
@nguyenv Please reconsider that we are talking about a file `arrow_adapter.cc` here.  A C API, we have C context there.

nguyenv:
I should have said we should move everything towards the C++ `new` and `delete`. There is really no reason to be using `malloc` in C++. You can use `new` with C structure as well so that is not an issue. 

Though thinking about this more, I actually don't know an alternative for `strdup`/`free`...

eddelbuettel:
There is more to all this.  _There are Arrow wrappers properly using `std::unique_ptr` together with `std::move`_.  Arrow is picky about its ownership.  This is not just about allocators.

johnkerl:
https://github.com/single-cell-data/TileDB-SOMA/pull/2352/commits has the bugfix that @eddelbuettel found (my bad) on its commit #2, and the just-don't-free-anything on its commit #3

Still it is segfaulting

So while I do believe -- as proved by `valgrind` as noted above -- that my commit #2 there is a positive step, there is something going on here other than mixes of C and C++ heap managers.

I'm running out of week but will need to help continue debugging Monday, if we're to move this work forward.

eddelbuettel:
What I actually _really_ do not understand and am moderately mad about is why compiler is kind to me letting it pass when it clearly balks in similar environments.  That the tests pass here only to fail on (arguable more similar than different) Ubuntu Linux setups is maddening too.

eddelbuettel:
@johnkerl I concur that the `valgrind` check is excellent. I really should have thought of that. (I was so focus on leak checks only, and we're still not in as good a place as tiledb-r is.)

johnkerl:
> clearly balks in similar environments

@mojaveazure and I had some success in doing a `docker run`of a vanilla image, then running lines from `apis/r/tools/r-ci.sh` from `.github/workflows/r-ci.yml`, then using `valgrind` (with above flags) on various unit-test invocations. Not without various intermittent errors from `valgrind` sometimes; which maddens.

Nonetheless we were able to also repro some instances of the segfaults. Along that path, promise lies.

nguyenv:
There's some work Dirk and I did very recently on the nanoarrow branch where missing `strdup`s got added and some corrections were made in the `release_array` and `release_schema` functions. This corrected segfaults we were seeing in the Python CI. I think those corrections need to be applied here.

eddelbuettel:
You can do a little faster than `r-ci.sh` or `run.ci` if you already start from `rocker/r2u`.  Do `apt update -qqq; apt install cmake git curl` and then you are down to R commands.  I have shortcuts like `installDeps.r` and now `installDeps2.r` (passing the needed arg for 'all' depends down) for the fetching of all depends, and then you are basically there.

That said I too of course sometimes run from `ubuntu:22.04` trying to replicate but it's all hard -- GHA does not run vanilla Ubuntu either.  Overall as said this is madness in the ability to not replicate.  We lost weeks to this now.

eddelbuettel:
@nguyenv Correct.  The 'nanoarrow refactor' branch is clear in CI for all Python and R tests.  But because the world is still a mean place it blows up coverage and I have no idea why.  (Double-checked the `covr` package, it recommends / uses -O0 but to no avail.)

nguyenv:
Note the changes within `arrow_adapter.cc` in [here](https://github.com/single-cell-data/TileDB-SOMA/compare/de/sc-42316/update_and_refactor_nanoarrow...viviannguyen/fix-segfault-issues-in-release-schema#diff-c66edd736f9a9cb059c6f6e0f4cc6d1cdfa0fd91a2b9220dd2e83c12c2546089). These corrections were made on top of the nanoarrow conversion so they'll need to be picked out manually.

nguyenv:
> C API, we have C context there.

I guess I don't see the issue though. I don't think this necessarily means we need to be using C `malloc`/`free`. Yes they are C structs, but those can all be managed with `new`/`delete` without needing to do anything special.

eddelbuettel:
@nguyenv In principle right in practice you are not I fear.  See https://arrow.apache.org/docs/format/CDataInterface.html, in only references `malloc`.   Grep leans a certain way too

```sh
edd@rob:~/git/tiledb-soma/apis/r/src(ph/feat/blockwise-reader)$ grep -ci malloc nanoarrow.*
nanoarrow.c:21
nanoarrow.h:8
edd@rob:~/git/tiledb-soma/apis/r/src(ph/feat/blockwise-reader)$ grep -c 'new ' nanoarrow.*
nanoarrow.c:0
nanoarrow.h:0
edd@rob:~/git/tiledb-soma/apis/r/src(ph/feat/blockwise-reader)$ grep -c free nanoarrow.*
nanoarrow.c:3
nanoarrow.h:9
edd@rob:~/git/tiledb-soma/apis/r/src(ph/feat/blockwise-reader)$ grep -c delete nanoarrow.*
nanoarrow.c:0
nanoarrow.h:0
edd@rob:~/git/tiledb-soma/apis/r/src(ph/feat/blockwise-reader)$ 
```

They of course is that pairing must match.  Each Arrow structure has a `release` (function) pointer.  When we suggest `malloc` may be more appropriate it is because we find the matching `free` in these.  As I said, there are possible C++ wrappers on top of this but they generally operate with (bare) unique pointers and move.   We _could_ allocate with `new` if we are very careful to pass C function pointers for `release` down with matching `delete` calls.

I have no priors.  I want the Heisenbug gone so I am with whatever achieves that.

eddelbuettel:
Also maybe no free cigar from valgrind.  Per [this page](https://bytes.usc.edu/cs104/wiki/valgrind#:~:text=Mismatched%20free%2Fdelete,-%3D%3D91%3D%3D&text=This%20is%20caused%20by%20incorrectly,need%20to%20use%20delete%20%5B%5D%20.) the 'mismatched free / delete / delete[]' hits only the case of freeing a pointer vs pointer-pointer (as we have with `buffers` array of pointers).  Doesn't look like we a magic 'wrong release for allocator used' here.

eddelbuettel:
I did a broader replacement of the few `delete` with `free` which of course
- passes here
- blows up at GHA in a side-repo
- does not say anything under valgrind either.

Sigh.

nguyenv:
Yeah I know the carrow docs use `malloc` in their examples, but I think that's just to show how to use it purely in C. Since we are using this within C++, we can favor using `new` instead. I think it's similar for the nanoarrow stuff where the code is written so that it _can_ be used in pure C as needed, but we don't need to limit ourselves to using C memory management that when using it in C++ . But honestly I don't care too much about needing to use `new` and would prefer to just do whatever eliminates the segfaults in the easiest way.

nguyenv:
I suspect this is more about the missing strdups and trying to free memory that was never actually allocated (see branch above)

johnkerl:
> You can do a little faster than r-ci.sh or run.ci if you already start from rocker/r2u. Do apt update -qqq; apt install cmake git curl and then you are down to R commands. I have shortcuts like installDeps.r and now installDeps2.r (passing the needed arg for 'all' depends down) for the fetching of all depends, and then you are basically there

@eddelbuettel please share a Dockerfile :)

> I think those corrections need to be applied here

@nguyenv commits are welcome on this PR or #2352 :)

nguyenv:
I will make a new branch with changes on top of #2352 then

eddelbuettel:
The Dockerfile is [rocker/r2u](https://rocker-project.org/images/) as I wrote. Then

```sh
docker run --rm -ti -v $PWD:/work -w work rocker/r2u bash
```

to lauch it followed by 

```sh
apt update -qqq; apt install cmake curl git
```

followed by this (where `local-buildMe.sh` is my default `bootstrap` or `cmake` caller)

```sh
cd libtiledbsoma/; mkdir temp-build; cd temp-build; ../local-buildMe.sh; ldconfig; cd ../..
``` 

and preparing the R build via

```sh
cd apis/r; Rscript -e 'remotes::install_deps(".", TRUE)'
```

as the newer `installDeps2.r -a` will only be in the next rebuild.  Then eg

```sh
install.r
```

and you're done.   I use a `testthat::test_local()` wrapper that set CI too via `tttl.r -c` but that too is more recent addition so 

```sh
CI=true /usr/lib/R/site-library/littler/examples/tttl.r
```

would do as would the Rscript call.  Of course, doing this does not lead to an error because the world is a mean place.

```sh
....
Duration: 124.3 s

[ FAIL 0 | WARN 0 | SKIP 0 | PASS 2669 ]
root@6520af1b9fff:/work/apis/r# 
```

eddelbuettel:
@nguyenv I have done that in a side repo (quicker to open a new GitHub repo and to both remove the other yaml files and reduce the matrix than to wait on runs one may not care about).  But no luck:

image

nguyenv:
https://github.com/single-cell-data/TileDB-SOMA/actions?query=branch%3Aviviannguyen%2Fblockwise-reader-patch-for-arrow-adapter

I spent some time today trying to figure this out but also no luck on my end. Cannot repro locally or on an EC2 instance either.

eddelbuettel:
Same. I even went as far a disabling the code inside the release functions. No change, still bombs in the same spot.  Tried a few other things, all bombed in the same spot in CI.

There must be a subtle race that CI / GHA trigger and our machines don't.  There is very similar code in tiledb-r which has no issues: the one key difference is what SOMA does with the private data use to hold on the shared pointer to an ArrayBuffer.  I am starting to think that it may come from there.  Alas, I have no "proof" of anything.

eddelbuettel:
@mojaveazure Thanks for the rebase, that was on my TODO as well but it has been a busy day.  Looks like we inherited some good state from `main` is nice.

johnkerl (approved):

johnkerl:
Regarding
https://github.com/single-cell-data/TileDB-SOMA/pull/2152#pullrequestreview-1964735631
we were waiting on:

* Having the reindexer in this PR, or another;
* Redness of CI

As per discussion above we decided collectively that the reindexer will be a follow-on PR, and the CI issue has been resolved via #2363 which has been merged to `main` and which this current PR is now rebased on top of.

So I believe we are good to merge this PR.

Issue #2149, opened 2024-02-15
[python] version 1.7 allocates more threads than 1.6 when reading array


bkmartinjr:
Test case below, which when run on a 128 CPU box:
* using 1.6.2, the peak threads allocated are 783
* using 1.7.0, the peak threads allocated are 1039 (additional 256)

The test case opens a SOMA DataFrame and does a simple read. Both the open and read steps seem to allocate increased number of threads. The difference is material on a high-CPU box where there is a worker per CPU:
* after opening the dataframe, 321 additional threads over 1.6
* after read, additional 256 threads over 1.6

A work-around is to restrict per process thread count using  `sm.compute_concurrency_level` and `sm.io_concurrency_level`, but that is undesirable in the long term.

----

Output of test shows thread count after key steps (noted in code).

On tiledbsoma==1.6.2:
```
$ pip install tiledbsoma==1.6.2 -q
$ python tmp/morethreads.py 
tiledbsoma.__version__        1.6.2
TileDB-Py tiledb.version()    (0, 24, 0)
TileDB core version           2.18.2
libtiledbsoma version()       libtiledb=2.18.2
python version                3.11.7.final.0
OS version                    Linux 6.2.0-1018-aws
None
Start                           : pid=106037, cpus=128 number of threads=136
After context                   : pid=106037, cpus=128 number of threads=136
After open dataframe            : pid=106037, cpus=128 number of threads=457
After read                      : pid=106037, cpus=128 number of threads=783
After close                     : pid=106037, cpus=128 number of threads=783
After GC                        : pid=106037, cpus=128 number of threads=783
After context                   : pid=106037, cpus=128 number of threads=783
After open dataframe            : pid=106037, cpus=128 number of threads=783
After read                      : pid=106037, cpus=128 number of threads=783
After close                     : pid=106037, cpus=128 number of threads=783
After GC                        : pid=106037, cpus=128 number of threads=783
```

On tiledbsoma==1.7.0:
```
$ pip install tiledbsoma==1.7.0 -q
$ python tmp/morethreads.py 
tiledbsoma.__version__        1.7.0
TileDB-Py tiledb.version()    (0, 25, 0)
TileDB core version           2.19.0
libtiledbsoma version()       libtiledb=2.19.1
python version                3.11.7.final.0
OS version                    Linux 6.2.0-1018-aws
None
Start                           : pid=107929, cpus=128 number of threads=136
After context                   : pid=107929, cpus=128 number of threads=136
[...]

PR #2158, opened 2024-02-21
[c++/python] Addition of `SOMAContext` to reduce thread count

nguyenv:
**Issue and/or context:**

#2149

Locally, the thread count with these changes is now:

```
tiledbsoma.__version__        1.5.0rc0.post208.dev4244460900
TileDB-Py tiledb.version()    (0, 25, 0)
TileDB core version           2.19.0
libtiledbsoma version()       libtiledb=2.21.0
python version                3.11.6.final.0
OS version                    Linux 4.19.128-microsoft-standard
None
Start                           : pid=31616, cpus=8 number of threads=16
After context                   : pid=31616, cpus=8 number of threads=37
After open dataframe            : pid=31616, cpus=8 number of threads=37
After read                      : pid=31616, cpus=8 number of threads=53
After close                     : pid=31616, cpus=8 number of threads=53
After GC                        : pid=31616, cpus=8 number of threads=53
After context                   : pid=31616, cpus=8 number of threads=53
After open dataframe            : pid=31616, cpus=8 number of threads=53
After read                      : pid=31616, cpus=8 number of threads=53
After close                     : pid=31616, cpus=8 number of threads=53
After GC                        : pid=31616, cpus=8 number of threads=53
After context                   : pid=31616, cpus=8 number of threads=53
After open dataframe            : pid=31616, cpus=8 number of threads=53
After read                      : pid=31616, cpus=8 number of threads=53
After close                     : pid=31616, cpus=8 number of threads=53
After GC                        : pid=31616, cpus=8 number of threads=53
After context                   : pid=31616, cpus=8 number of threads=53
After open dataframe            : pid=31616, cpus=8 number of threads=53
After read                      : pid=31616, cpus=8 number of threads=53
After close                     : pid=31616, cpus=8 number of threads=53
After GC                        : pid=31616, cpus=8 number of threads=53
```
With 1.7, after the DataFrame refactoring, the counts were
```
tiledbsoma.__version__        1.7.0
TileDB-Py tiledb.version()    (0, 25, 0)
TileDB core version           2.19.0
libtiledbsoma version()       libtiledb=2.19.1
python version                3.11.0.final.0
OS version                    Linux 4.19.128-microsoft-standard
None
Start                           : pid=712, cpus=8 number of threads=16
After context                   : pid=712, cpus=8 number of threads=16
After open dataframe            : pid=712, cpus=8 number of threads=58
After read                      : pid=712, cpus=8 number of threads=74
After close                     : pid=712, cpus=8 number of threads=74
After GC                        : pid=712, cpus=8 number of threads=58
After context                   : pid=712, cpus=8 number of threads=58
After open dataframe            : pid=712, cpus=8 number of threads=74
After read                      : pid=712, cpus=8 number of threads=74
After close                     : pid=712, cpus=8 number of threads=74
After GC                        : pid=712, cpus=8 number of threads=58
After context                   : pid=712, cpus=8 number of threads=58
After open dataframe            : pid=712, cpus=8 number of threads=74
After read                      : pid=712, cpus=8 number of threads=74
After close                     : pid=712, cpus=8 number of threads=74
After GC                        : pid=712, cpus=8 number of threads=58
After context                   : pid=712, cpus=8 number of threads=58
After open dataframe            : pid=712, cpus=8 number of threads=74
After read                      : pid=712, cpus=8 number of threads=74
After close                     : pid=712, cpus=8 number of threads=74
After GC                        : pid=712, cpus=8 number of threads=58
```

With 1.6, prior to the DataFrame refactor:
```
tiledbsoma.__version__        1.6.0
TileDB-Py tiledb.version()    (0, 24, 0)
TileDB core version           2.18.2
libtiledbsoma version()       libtiledb=2.18.2
python version                3.11.6.final.0
OS version                    Linux 4.19.128-microsoft-standard
None
Start                           : pid=13843, cpus=8 number of threads=16
After context                   : pid=13843, cpus=8 number of threads=16
After open dataframe            : pid=13843, cpus=8 number of threads=37
After read                      : pid=13843, cpus=8 number of threads=58
After close                     : pid=13843, cpus=8 number of threads=58
After GC                        : pid=13843, cpus=8 number of threads=58
After context                   : pid=13843, cpus=8 number of threads=58
After open dataframe            : pid=13843, cpus=8 number of threads=58
After read                      : pid=13843, cpus=8 number of threads=58
After close                     : pid=13843, cpus=8 number of threads=58
After GC                        : pid=13843, cpus=8 number of threads=58
After context                   : pid=13843, cpus=8 number of threads=58
After open dataframe            : pid=13843, cpus=8 number of threads=58
After read                      : pid=13843, cpus=8 number of threads=58
After close                     : pid=13843, cpus=8 number of threads=58
After GC                        : pid=13843, cpus=8 number of threads=58
After context                   : pid=13843, cpus=8 number of threads=58
After open dataframe            : pid=13843, cpus=8 number of threads=58
After read                      : pid=13843, cpus=8 number of threads=58
After close                     : pid=13843, cpus=8 number of threads=58
After GC                        : pid=13843, cpus=8 number of threads=58
```

**Changes:**

- Introduce `SOMAContext` to replace `tiledb::Context` arguments. Ultimately it is just a very simple wrapper around `tiledb::Context` but will aid in transitioning us out of tiledb-py usage by slowly eliminating usage of `tiledb.Ctx()` in favor of `SOMAContext`
- We can now pass `SOMAContext` between C++ and Python and avoid needing to create new Contexts unless necessary
- Modify `SOMACollection` to directly inherit from `SOMAGroup` so that they may be dynamically casted in the `SOMAObject::open` factory

**Notes for Reviewer:**

- Applying query conditions greatly increases the number of threads. Without applying the filter, the thread count remains at 37 threads even after the dataframe read. It is not really possible to work around until the parser is implemented in core


apis/python/src/tiledbsoma/_tiledb_object.py
```diff
@@ -82,20 +82,17 @@ def open(
...
         try:
-            handle = _tdb_handles._open_with_clib_wrapper(
-                uri, mode, context, tiledb_timestamp
-            )
+            handle = _tdb_handles.open(uri, mode, context, tiledb_timestamp)
         except SOMAError:
             handle = cls._wrapper_type.open(uri, mode, context, tiledb_timestamp)
         return cls(
-            handle,
+            handle,  # type: ignore
```
nguyenv, line 89: @thetorpedodog I'm still having issues with typing for these. I get this error "apis/python/src/tiledbsoma/_tiledb_object.py:89: error: Argument 1 to "TileDBObject" has incompatible type "Wrapper[Any | Any | Any]"; expected "_WrapperType_co"  [arg-type]". I'm not sure what do add to get this to work.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -47,7 +48,9 @@ def _maybe_timestamp_ms(input: Optional[OpenTimestamp]) -> Optional[int]:
 """Sentinel object to distinguish default parameters from None."""
 
 
-class SOMATileDBContext:
+class SOMATileDBContext(clib.SOMAContext):  # type: ignore
+    # (Class cannot subclass "SOMAContext" (has type "Any"))
```
nguyenv, line 52: This one too. Nothing that I found via Google on my own resolved it.

apis/python/src/tiledbsoma/_tiledb_object.py
thetorpedodog: At this point, `_WrapperType_co` is essentially "resolved", so it refers specifically to the wrapper type of the given SOMA object.

Take this example:

```python
def returns_something() -> int | str:
  ...

_T_co = TypeVar("_T_co", bound=Union[int, str], covariant=True)
class Holder(Generic[_T_co]):
  _it: _T_co
  def __init__(self, obj: Optional[_T_co]):
    if obj is None:
      self._it = returns_something()
    else:
      self._it = obj

class IntHolder(Holder[int]):
  pass

class StrHolder(Holder[str]):
  pass
```

In this case, this use of `returns_something` is not valid, because `returns_something` may return either of an `int` or a `str`. So an `IntHolder` might get a `str`, or a `StrHolder` might get an `int`.

I think the right thing to do here is to add a generic entry for the specialized SOMA type we expect, and use *that* rather than the return of `_tdb_handles.open`.

nguyenv: I think I partially understand but not really. I've been staring at this code and trying to follow it but confusing myself now. Is this on the right track? To make this valid you would modify the constructor to be:

```
_T_co = TypeVar("_T_co", bound=Union[int, str], covariant=True)
class Holder(Generic[_T_co]):
  _it: _T_co
  def __init__(self, obj: Optional[_T_co]):
    if obj is None:
      self._it = obj

class IntHolder(Holder[int]):
  pass

class StrHolder(Holder[str]):
  pass
```

And then you would use it like this which would create you either the `IntHolder` or `StringHolder`. You wouldn't actually use `returns_something()` inside `Holder.__init__()`?
```
holder = Holder(returns_something())
```

thetorpedodog: I probably should have mentioned the fact that in the example, Holder is a purely abstract class (like TileDBObject), and IntHolder/StrHolder are the concrete types which provide additional functionality (like, e.g., Collection or DataFrame). So calling `Holder(returns_something())` by itself wouldn't make sense, but its type would end up being `Holder[int | str]`, which is equivalent to `Holder[int] | Holder[str]` due to covariance.

The important thing I was trying to get at is that `returns_something` is like `_tdb_handles.open`.

There are a couple of ways we could resolve this. Both involve putting the held class in the concrete types:

```python
_T_co = TypeVar("_T_co", bound=Union[int, str], covariant=True)
class Holder(Generic[_T_co]):
  _CLS: ClassVar[Type[_T_co]]
  _it: _T_co

  @classmethod
  def open(cls, obj: Optional[_T_co]) -> Self:
    ...

class IntHolder(Holder[int]):
  _CLS = int
  ...

class StrHolder(Holder[str]):
  _CLS = str
  ...
```

Either we check the type of the thing returned from `returns_something`:

```python
class Holder(Generic[_T_co]):
  ...
  @classmethod
  def open(cls, obj: Optional[_T_co]) -> Self:
    if obj is None:
      maybe_obj = returns_something()
      if not isinstance(maybe_obj, cls._CLS):
        raise TypeErrror("something went wrong")
      obj = maybe_obj
    self._it = obj
```

or we use a property of the type we passed in to return the value we want:

```python
class Holder(Generic[_T_co]):
  ...
  @classmethod
  def open(cls, obj: Optional[_T_co]) -> Self:
    # This isn't a perfect example since there aren't any classmethods on both
    # int and str which return their own type.
    self._it = cls(0) if obj is None else obj
```

Since C++ doesn’t really have an equivalent to this type of class method (i.e., you can’t define static `SOMAObject::open` and have it work as `SOMADataFrame::open` and `SOMASparseNDArray::open` etc.), I think we’re left with the first version, where we check the type that the open operation returned.

This means that in the end, we probably want something like:

```python
# AnyWrapper is renamed to AnyRWWrapper.
_RWWrapper_co = TypeVar("_RWWrapper_co", bound=_tdb_handles.AnyRWWrapper, covariant=True)
# The new AnyWrapper includes wrappers for both the read-only and read-write handles.
_ROWrapper_co = TypeVar("_ROWrapper_co", bound=_tdb_handles.AnyWrapper, covariant=True)
class TileDBObject(somacore.SOMAObject, Generic[_RWWrapper_co, _ROWrapper_co]):
  _rw_wrapper_type: Type[_RWWrapper_co]
  _ro_wrapper_type: Type[_ROWrapper_co]  # these should maybe be ClassVars

  _handle: Union[_RWWrapper_co, _ROWrapper_co]

  def __init__(self, handle: Union[_RWWrapper_co, _ROWrapper_co], ...) -> None:
    ...
    self._handle = handle

  @classmethod
  def open(cls, ...) -> Self:
    try:
      handle = _tdb_handles.open(...)
      if not isinstance(handle, cls._ro_wrapper_type):
        raise WhateverError(...)
      # Now we're guaranteed that `handle` is a `_ROWrapper_co` instance.
    except ...:
      handle = self._rw_wrapper_type.open(...)
    return cls(handle, ...)
```

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
thetorpedodog: This one is, thankfully, a little more straightforward. It’s because [the type checker has no information about what a `clib.SOMAContext`](https://stackoverflow.com/questions/49888155/class-cannot-subclass-qobject-has-type-any-using-mypy) is, so it treats it as an `Any` type in some ways. mypy disallows this by default, since that means that `SOMATileDBContext` would effectively be an `Any` as far as accessing it is concerned. So if you had `my_context.this_member_will_never_exist`, it would always be forced to say "yeah that maybe exists" and call the result of that call an `Any`.

I would handle this by using composition rather than inheritance, so maybe have a `SOMATileDBContext._native_context` field that contains the `clib.SOMAContext`.

apis/python/src/tiledbsoma/_tiledb_object.py
nguyenv: Thanks so much for explaining! I have something that works for this now.

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -51,36 +51,45 @@ def open(
...
+        if soma_object.type in (
+            "SOMADataFrame",
+            "SOMASparseNDArray",
+            "SOMADenseNDArray",
+        ):
             return ArrayWrapper.open(uri, mode, context, timestamp)
-        if obj_type == "group":
+        if soma_object.type in ("SOMACollection", "SOMAExperiment", "SOMAMeasurement"):
             return GroupWrapper.open(uri, mode, context, timestamp)
+    except (SOMAError, RuntimeError):
```
bkmartinjr, line 74: I found this code a bit confusing and with some redundancy -- it suggests that there exists a situation where an exception is thrown in the first part (above), which can be resolved by the "use tiledb-py code path" code below.  The actual code in each path seems only different in that it uses `tiledb.object_type` instead of the tiledbsoma `open` to determine object type (and lacks the tiledbsoma DataFrame read path).

It seems like it could be simplified quite a bit, unless I'm missing some not-obvious logic (in which case a comment might help).  Example of what I had in mind (caveat: not tested):

```
    soma_object = clib.SOMAObject.open(
        uri, open_mode, context.native_context, (0, timestamp_ms)
    )

    obj_type = soma_object.type or tiledb.object_type(uri, ctx=context.tiledb_ctx)

    if open_mode == clib.OpenMode.read and obj_type == "SOMADataFrame":
        return DataFrameWrapper._from_soma_object(soma_object, context)

    if obj_type in (
            "SOMADataFrame",
            "SOMASparseNDArray",
            "SOMADenseNDArray",
            "array",
        ):
            return ArrayWrapper.open(uri, mode, context, timestamp)
        if soma_object.type in ("SOMACollection", "SOMAExperiment", "SOMAMeasurement", "group"):
            return GroupWrapper.open(uri, mode, context, timestamp)

    # Invalid object
    raise SOMAError(f"{uri!r} has unknown storage type {obj_type!r}")
```

apis/python/src/tiledbsoma/soma_collection.cc
```diff
@@ -0,0 +1,54 @@
...
+namespace py = pybind11;
+using namespace py::literals;
+using namespace tiledbsoma;
+
+void load_soma_collection(py::module &m) {
+    py::class_<SOMACollection, SOMAGroup, SOMAObject>(m, "SOMACollection");
+    py::class_<SOMAExperiment, SOMAGroup, SOMAObject>(m, "SOMAExperiment");
+    py::class_<SOMAMeasurement, SOMAGroup, SOMAObject>(m, "SOMAMeasurement");
+}
+}
```
bkmartinjr, line 54: missing EOL (lint?)

apis/python/src/tiledbsoma/soma_array.cc
```diff
@@ -199,9 +199,7 @@ void load_soma_array(py::module &m) {
             auto pa_schema_import = pa.attr("Schema").attr("_import_from_c");
             return pa_schema_import(py::capsule(reader.arrow_schema().get()));
         })
-        .def("config", [](SOMAArray& reader) -> py::dict {
-            return py::cast(reader.config());
-        })
+        .def("ctx", &SOMAArray::ctx)
```
bkmartinjr, line 202: The Python bindings use the name `context`, rather than `ctx` (or in TileDB-PY, `Ctx`). Is there a reason to introduce a third option (`ctx`) rather than stick with the SOMA name `context`?

(this is a nit: if there is a reason, I have no objection - just spent a few minutes trying to understand if they were different concepts or the same, just renamed)

apis/python/src/tiledbsoma/soma_group.cc
```diff
@@ -0,0 +1,45 @@
...
+
+namespace py = pybind11;
+using namespace py::literals;
+using namespace tiledbsoma;
+
+void load_soma_group(py::module &m) {
+    py::class_<SOMAGroup, SOMAObject>(m, "SOMAGroup")
+    .def_property_readonly("type", &SOMAGroup::type);
+}
+}
```
bkmartinjr, line 45: nit: missing EOL (lint?)

nguyenv: The pybind11 code doesn't go through the clang-format linter.

apis/python/src/tiledbsoma/soma_array.cc
nguyenv: No there isn't a reason for using `ctx`. I'll make them all `context`.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -146,6 +151,15 @@ def timestamp(self) -> Optional[datetime.datetime]:
...
+        """The C++ SOMAContext for this SOMA context."""
+        if self._native_context is None:
+            self._native_context = clib.SOMAContext(
+                {k: str(self.tiledb_config[k]) for k in self.tiledb_config}
+            )
+        return self._native_context
+
     @property
     def tiledb_ctx(self) -> tiledb.Ctx:
         """The TileDB Context for this SOMA context."""
```
bkmartinjr, line 165: to clarify this comment in light of the `native_context`, maybe rephrase as `The TileDB-Py Context ...`?

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -146,6 +151,15 @@ def timestamp(self) -> Optional[datetime.datetime]:
             return None
         return ms_to_datetime(self.timestamp_ms)
 
+    @property
+    def native_context(self) -> clib.SOMAContext:
+        """The C++ SOMAContext for this SOMA context."""
+        if self._native_context is None:
```
bkmartinjr, line 157: This is lock-free, but lazily instantiated.  does it need a concurrency guard like that used in `tiledb_ctx()` below?

bkmartinjr (changes requested):
The only substantive comment is related to a missing lock around context instantiation in _soma_tiledb_context.py.  It looks like it needs a lock.

apis/python/src/tiledbsoma/_tdb_handles.py
nguyenv: Thanks for the suggestion. I was able to clean up this code block into something easier to read.

bkmartinjr (approved):
LGTM!

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -146,9 +151,19 @@ def timestamp(self) -> Optional[datetime.datetime]:
...
 
+    @property
+    def native_context(self) -> clib.SOMAContext:
+        """The C++ SOMAContext for this SOMA context."""
+        with self._lock:
+            if self._native_context is None:
+                self._native_context = clib.SOMAContext(
+                    {k: str(self.tiledb_config[k]) for k in self.tiledb_config}
+                )
+            return self._native_context
```
nguyenv, line 162: How do you use this lock correctly? Something seems to be holding on to it and preventing anything from running. I just tested locally and Pytest also freezes but commenting this out makes everything run again.

bkmartinjr: the lock is not re-entrant, and this code calls self.tiledb_config which also tries to acquire it.

Solutions:
* unwind the nested locks
* or switch to a re-entrant lock (threading.RLock instead of .Lock)

bkmartinjr: Doc: https://docs.python.org/3/library/threading.html#rlock-objects

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -146,9 +151,23 @@ def timestamp(self) -> Optional[datetime.datetime]:
...
+    @property
+    def native_context(self) -> clib.SOMAContext:
+        """The C++ SOMAContext for this SOMA context."""
+        with self._lock:
+            if self._native_context is not None:
+                return self._native_context
+
+        # tiledb_config also acquires the lock so we must first release it and
+        # then reacquire it
+        cfg = self.tiledb_config
```
thetorpedodog, line 163: I think you just want to call `self._internal_tiledb_config()` here while holding onto the lock. Right now this ends up creating a race condition where we might end up constructing two separate `clib.SOMAContext`s and using them at different times.

Issue #2099, opened 2024-02-01
[python] indexer creation performance is slow for Pandas Series vs numpy ndarray


bkmartinjr:
if you build an indexer with a Pandas Series, it is 25X slower than if you build it with a Numpy NDArray.  Example which uses the exact same object, but does a cast before calling the index creator.

Expected result: these should be identical

This is on the (pre-release) `release-1.7` branch as of 8d858e6

```
In [36]: import pandas as pd, tiledbsoma as soma

In [37]: type(obs.soma_joinid)
Out[37]: pandas.core.series.Series

In [38]: type(obs.soma_joinid.to_numpy())
Out[38]: numpy.ndarray

In [39]: obs.soma_joinid.shape
Out[39]: (41939851,)

In [40]: %time soma.tiledbsoma_build_index(obs.soma_joinid)
CPU times: user 50.8 s, sys: 405 ms, total: 51.2 s
Wall time: 51.2 s
Out[40]: <tiledbsoma.pytiledbsoma.IntIndexer at 0x7fa0daa04530>

In [41]: %time soma.tiledbsoma_build_index(obs.soma_joinid.to_numpy())
CPU times: user 1.29 s, sys: 285 ms, total: 1.58 s
Wall time: 1.58 s
Out[41]: <tiledbsoma.pytiledbsoma.IntIndexer at 0x7f9e32a2fb70>
```

bkmartinjr:
We should also ensure that we correctly (performantly) handle:
* Arrow Array (contiguous)
* Arrow ChunkedArray (non-contiguous)
* pandas.Series (original bug)
* numpy.ndarray (which already works)

beroy:
I tried multiple scenarios and used the np.array as my base line (keys and lookups are equal and I perform one map_location (setup) and one lookup):
**Baseline:**
    `keys = np.array(range(1, 100000000))`
Setup time: 1.7088468329966418
Lookup time: 0.49534945800405694    
**pyarrow array:**
   `keys = pa.array(list(range(1, 100000000)))`
Setup time: 1.7942252089997055
Lookup time: 0.5838396660037688
**pyarrow Chunked Array:** 
    `keys = pa.chunked_array([list(range(1, 100000000)), list(range(100000000, 200000000)), list(range(300000000, 400000000))])`
**3x size the baseline data.**
Setup time: 10.952384541000356
Lookup time: 6.603933874997892    
    
**Pandas series:** 
[...]

PR #2159, opened 2024-02-21
[c++] Optimizing indexer for pandas and pyarrow

beroy:
**Issue and/or context:**
https://github.com/single-cell-data/TileDB-SOMA/issues/2099

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/reindexer.cc
+++ b/apis/python/src/tiledbsoma/reindexer.cc
@@ -31,7 +31,7 @@
  */
 
 #include <tiledbsoma/reindexer/reindexer.h>
-
+#include <tiledbsoma/utils/carrow.h>
 #include "common.h"
 
 #define DENUM(x) .value(#x, TILEDB_##x)
@@ -40,6 +40,95 @@ namespace libtiledbsomacpp {
 namespace py = pybind11;
 using namespace py::literals;
 using namespace tiledbsoma;
+
+/***
+ * Handle general lookup for Re-indexer
+ * @param indexer reference to the indexer
+ * @param lookups input values to be looked up
+ * @return looked up values
+ */
+py::array_t<int64_t> get_indexer_general(
+    IntIndexer& indexer, py::array_t<int64_t> lookups) {
+    auto input_buffer = lookups.request();
+    int64_t* input_ptr = static_cast<int64_t*>(input_buffer.ptr);
+    size_t size = input_buffer.shape[0];
+    auto results = py::array_t<int64_t>(size);
+    auto results_buffer = results.request();
+    size_t results_size = results_buffer.shape[0];
+    int64_t* results_ptr = static_cast<int64_t*>(results_buffer.ptr);
+    indexer.lookup(input_ptr, results_ptr, size);
+    return results;
+}
+
+/***
+ * Helper function to provide data and schema for an arrow object
+ * @param object python object
+ * @param arrow_array extracted array data
+ * @param arrow_schema extracted array schema
+ */
+//
+void extract_py_array_schema(
+    const pybind11::handle object,
+    ArrowArray& arrow_array,
+    ArrowSchema& arrow_schema) {
+    uintptr_t arrow_schema_ptr = (uintptr_t)(&arrow_schema);
+    uintptr_t arrow_array_ptr = (uintptr_t)(&arrow_array);
+
+    // Using array._export_to_c to get arrow array and schema.
+    object.attr("_export_to_c")(arrow_array_ptr, arrow_schema_ptr);
+}
+
+/***
+ * Handle pyarrow-based lookup for Re-indexer
+ * @param indexer reference to the indexer
+ * @py_arrow_array pyarrow inputs to be looked up
+ * @return looked up values
+ */
+py::array_t<int64_t> get_indexer_py_arrow(
+    IntIndexer& indexer, py::object py_arrow_array) {
+    // Check if it is not a pyarrow array or pyarrow chunked array
+    if (!py::hasattr(py_arrow_array, "_export_to_c") &&
+        !py::hasattr(py_arrow_array, "chunks") &&
+        !py::hasattr(py_arrow_array, "combine_chunks")) {
+        // Handle the general case (no py arrow objects)
+        return get_indexer_general(indexer, py_arrow_array);
+    }
+
+    py::list array_chunks;
+    if (py::hasattr(py_arrow_array, "chunks")) {
+        array_chunks = py_arrow_array.attr("chunks").cast<py::list>();
+    } else {
+        array_chunks.append(py_arrow_array);
+    }
+
+    // Calculate the total size of the input chunked array.
+    int total_size = 0;
+    for (const pybind11::handle array : array_chunks) {
+        ArrowSchema arrow_schema;
+        ArrowArray arrow_array;
+        extract_py_array_schema(array, arrow_array, arrow_schema);
+        total_size += arrow_array.length;
+    }
+
+    // Allocate the output
+    auto results = py::array_t<int64_t>(total_size);
+    auto results_buffer = results.request();
+    size_t results_size = results_buffer.shape[0];
+    int64_t* results_ptr = static_cast<int64_t*>(results_buffer.ptr);
+
+    // Write output (one chunk at a time)
+    int write_offset = 0;
+    for (const pybind11::handle array : array_chunks) {
+        ArrowSchema arrow_schema;
+        ArrowArray arrow_array;
+        extract_py_array_schema(array, arrow_array, arrow_schema);
+        auto input_ptr = (int64_t*)arrow_array.buffers[1];
+        indexer.lookup(
+            input_ptr, results_ptr + write_offset, arrow_array.length);
+        write_offset += arrow_array.length;
+    }
+    return results;
+}
 
 void load_reindexer(py::module& m) {
     // Efficient C++ re-indexing (aka hashing unique key values to an index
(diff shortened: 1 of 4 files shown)


bkmartinjr:
@beroy - do you have post-change benchmarking numbers for all cases?  Just want to ensure removing this does not add any significant regression.  I noted the post-change measurement of Pandas types, but not the others (e.g., Arrow).

Most concerned about:
* Arrow Array & ChunkedArray
* Pandas Series
* NDArray
* Python list (of least concern, but it does see some use)

Appreciate it if you could check these.

Somewhat related:  the pytests do not seem to test all types (e.g., Arrow types).  It would be excellent if we had a unit test to ensure the expected types work in a basic manner.

bkmartinjr:
> posted all the benchmarks

I don't see "after change" benchmarks for all cases.  Are they identical to "before change" numbers?   (I'm being explicit as I'm just looking for any potential regression)

beroy:
Yes, for the rest of the cases the benchmark results are identical. Actually I only posted the one with major regression. LMK if you need all the results there.

beroy:
Updated all results

beroy:
@bkmartinjr I published all the result plus Panda's results. Some take away:

- We clearly are way better in lookup across the board.
- Panda's does not perform in a few cases python list and pa.chunked_array
- Panda's setup (hash creation) is significantly faster than us. Something that for sure needs investigation.

beroy:
A couple of points:

- The results you mentioned are from Panda's runs
- The size of chunked_array is 3 times bigger

I'll create a spreadsheet with all the results and the arrays all of the same size soon. Easier to discuss that way.

bkmartinjr:
> The results you mentioned are from Panda's runs

Not sure I understand, as the items I pasted are Arrow.  

> chunked_array is 3 times bigger

That completely escaped me - apologies!  Results still should not scale sub-linearly, right?  My rationale: setup and lookup do not inherentlyrequire a flattened vector or a copy - the should be approximately the same speed regardless of chunked or flat.   Under the covers, they are all just 1-n_chunks flat vectors, so any overhead is caused by either:
* your interface forcing a single flat vector rather than a list of vectors (ake ChunkedArray)
* unexpected time conversions which do a copy

I suspect the ChunkedArray issue is primarily the first.  Given that this is the standard Arrow array type we receive from Table, etc, it is worth handling correctly.

beroy:
@bkmartinjr you're absolutely right. When I changed the chunked array to the same size with 10 chunks my result became consistent with np.array. [Here](https://docs.google.com/spreadsheets/d/1gGJ-2lqd1I0UqNLj3W8XjMd0IG42prmJXq7lt7XLzBw/) is the spreadsheet with all details. For me, right now one concern is in general our setup time vs pandas.

bkmartinjr:
> spreadsheet

thanks for positing that including the benchmark code.  There is a bug in the benchmark which explains some of the odd results (e.g., time==0 for Pandas setup) - sent you a slack with the details.

beroy:
@bkmartinjr updated the doc with random large lookups. There are multiple major observations:

- The fix take care of panda's setup slowdown
- We do worse than panda's on pyarrow (maybe worth looking into)
- Both us and panda do really bad on python lists (we do much better than pandas still)

bkmartinjr:
@beroy - much better results!   I'd focus on pyarrow, numpy and pandas keys, as they should all be roughly equivalent in speed (all are flat int64 arrays underneath, so any difference is due to the wrappers and/or excess copies).   The `list` is another beast and low priority IMHO.

apis/python/tests/test_indexer_dtatye_perf.py
```diff
@@ -0,0 +1,99 @@
+from time import perf_counter
```
bkmartinjr, line 1: Do we want to start adding perf tests to pytest unit tests?  Without some sort of pass/fail criteria, they don't seem very useful.  Historically we have separated unit tests (function, correctness) from performance, and only the former were run in CI

CC: @johnkerl for any thoughts he may have

bkmartinjr (approved):
overall LGTM.  Not sure the perf test belongs in our suite of unit tests without a clear pass/fail criteria, but I'll leave that decision to John, et al.

apis/python/tests/test_indexer_dtatye_perf.py
beroy: @bkmartinjr I moved the perf benchmark outside the CI. Agree totally they should not be in the CI. Also I have correctness version of those tests (smaller memory footprint) in the CI.

libtiledbsoma/src/reindexer/test_indexer_dtatye_perf.py
```diff
@@ -0,0 +1,99 @@
+from time import perf_counter
```
johnkerl, line 1: @beroy is `test_indexed_dtatye_perf.py` a typo? What does `dtatye` mean?

libtiledbsoma/src/reindexer/test_indexer_dtatye_perf.py
```diff
@@ -0,0 +1,99 @@
...
+import numpy as np
+import pandas as pd
+import pyarrow as pa
+
+from tiledbsoma._index_util import tiledbsoma_build_index
+from tiledbsoma.options import SOMATileDBContext
+from tiledbsoma.options._soma_tiledb_context import _validate_soma_tiledb_context
+
+"""
+Performance test evaluating the reindexer performance compared to pnadas.Index for different data types
```
johnkerl, line 12: typo: `pnadas` -> `pandas`

libtiledbsoma/src/reindexer/test_indexer_dtatye_perf.py
```diff
@@ -0,0 +1,99 @@
...
+
+    keys = list(range(1, 100000000))
+    lookups = list(np.random.randint(0, 1000, 1000000000))
+
+
+#   Commented out as it takes a very long time for pandas
+#   indexer_test_build("python list", keys, lookups)
+
+
+main()
```
nguyenv, line 99: This file should be written to be run through Pytest and not call `main()` like this.

nguyenv: Oh wait sorry. Bruce already addressed that it shouldn't. Ignore this comment.

johnkerl:
@beroy there are several post-merge comments here -- let's get a follow-up PR going

apis/python/src/tiledbsoma/reindexer.cc
```diff
@@ -57,48 +146,19 @@ void load_reindexer(py::module& m) {
...
-                int64_t* results_ptr = static_cast<int64_t*>(
-                    results_buffer.ptr);
-                size_t results_size = input_buffer.shape[0];
-                indexer.lookup(input_ptr, input_ptr, size);
-            });
+        // If the input is not arrow (does not have _export_to_c attribute),
+        // it will be handled using a general input method.
+        .def("get_indexer", [](IntIndexer& indexer, py::object py_arrow_array) {
+            return get_indexer_py_arrow(indexer, py_arrow_array);
+        });
```
nguyenv, line 161: I think these should have distinct names as we discussed Pybind11 will not handle dispatching to the correct arguments. You can also simplify the binding by just passing in the function name.

```
.def("get_indexer_general", get_indexer_general)
.def("get_indexer_py_arrow", get_indexer_py_arrow);
```

Instead of doing the type checking within the Pybind11 function, they should be done in the Python code. And then call the correct `get_indexer_general` or `get_indexer_py_arrow` for that object.

apis/python/src/tiledbsoma/reindexer.cc
```diff
@@ -41,6 +41,95 @@ namespace py = pybind11;
...
+ */
+py::array_t<int64_t> get_indexer_py_arrow(
+    IntIndexer& indexer, py::object py_arrow_array) {
+    // Check if it is not a pyarrow array or pyarrow chunked array
+    if (!py::hasattr(py_arrow_array, "_export_to_c") &&
+        !py::hasattr(py_arrow_array, "chunks") &&
+        !py::hasattr(py_arrow_array, "combine_chunks")) {
+        // Handle the general case (no py arrow objects)
+        return get_indexer_general(indexer, py_arrow_array);
+    }
```
nguyenv, line 95: I know I said something different over DMs a few days ago, but I think it's better to remove this check from here and do these on the Python side.

nguyenv (commented):
I don't think my comments need to be addressed immediately, but we should definitely correct the typos @johnkerl pointed out.

apis/python/src/tiledbsoma/reindexer.cc
beroy: This has a huge implication to our entire indexer API! Changing the API is huge change that has ripple effects. I strongly rather not change API specially given the fact that our API is intentionally copying the pandas one.

apis/python/src/tiledbsoma/reindexer.cc
beroy: Doing it on python is not practical as the API call goes directly to the indexer meaning there's no python layer between the python `get_index` and this.

beroy: I will create one PR with minor changes needed for this and other PRs.

libtiledbsoma/src/reindexer/test_indexer_dtatye_perf.py
beroy: It's a typo. It should say `data_types`

apis/python/src/tiledbsoma/reindexer.cc
nguyenv: I don't think this should have any effect on the API. You would implement in Python by doing
```
def get_indexer():
    if(is pyarrow object):
        return get_indexer_py_arrow()
    else:
        return get_indexer_general()
```

nguyenv: Oh I see the issue now. I think just keep how you have it for now, but we should have the `clib` stuff just internal. In the future, have a `class IntIndex` that holds a member to `clib.IntIndexer`.

PR #2168, opened 2024-02-23
[c++/ci] Update `make format` and CI format-checker to find more C++ files

johnkerl:
The `make format` and the CI checker were coded back when all the C++ source was located in `libtiledbsoma`. Since then, there are C++ source files located elsewhere, but `make format` and the CI checker were never updated to find them. So we've been running `clang-format` on some files and not others.


eddelbuettel:
> So we've been running clang-format on some files and not others.

Intentional for the R package embedded here (just like the external R package). `clang-format` was never used.

nguyenv (approved):

PR #2173, opened 2024-02-26
[python] Use somacore v1.0.8

johnkerl:
For @ebezzi 

See also https://github.com/single-cell-data/SOMA/releases/tag/python-v1.0.8


ebezzi:
This is unfortunately due to some coupling between `ExperimentAxisQuery` (defined in `somacore`) and `SOMATileDBContext` (defined here). https://github.com/single-cell-data/TileDB-SOMA/pull/2001 will fix this issue.

johnkerl:
Superseded by #2001

Issue #2172, opened 2024-02-26
[Bug] - Ingesting multiple H5ADs with `.from_h5ad`


srdsam:
**Describe the bug**

When I ingest multiple `h5ad` files with `tiledbsoma.io.from_h5ad` using a `registration_mapping` (generated with `tiledbsoma.io.register_h5ads`), the `from_h5ad` function didn't throw any errors. However only the data from the first `h5ad` seemed to populate the SOMA object. Downstream objects populate `obs` and `var` correctly, but not `X` (is just filled with 0s). 

I ended up circumventing this error by using the `append` functions (e.g. `append_obs`, `append_var`, and `append_X`). Unsure if I was misusing the `tiledbsoma.io.from_h5ad` function or registration mapping? It data ingestion to work file with fewer than 10 h5ads (didn't profile exactly). 

**To Reproduce**

The ingest function took in a list of h5ads converted from GTEx's bulk tissue data. Happy to provide full code or list of h5ads if that helps. Just want to sanity check my use of TileDB-SOMA first. 

```
def ingest(
    data: Union[List[Union[os.PathLike, str]], Union[os.PathLike, str]],
    modality: str,
    base_path: str,
    data_source: str,
    registration_mapping: Optional[ExperimentAmbientLabelMapping] = None,
) -> str:
    """
    Ingest data into a TileDB-SOMA experiment.
    """
    # Validate AnnData Objects
    data = validate(data)
    data = [data] if isinstance(data, (str, os.PathLike)) else data

    # Generate registration mapping
    if registration_mapping is None:
        registration_mapping = register_h5ads(
            experiment_uri=(
                base_path if Experiment.exists(base_path, context=SOMA_CTX) else None
            ),
            h5ad_file_names=data,
            measurement_name=modality,
            obs_field_name="obs_id",
            var_field_name="var_id",
            context=SOMA_CTX,
        )

    # Ingest raw h5ad files
    soma_uri = _ingest_h5ads(data, base_path, modality, registration_mapping)

    logger.info(f"Size of SOMA: {get_size_of_dir(base_path):.2f} MB")
    return soma_uri


@profile
def _ingest_h5ads(
    file_paths: List[str],
    base_path: str,
    modality: str,
    registration_mapping: ExperimentAmbientLabelMapping,
) -> str:
    """
    Handles ingestion of data from a list of file paths.
    """
    total_size = 0
    uri = None
    num_files = len(file_paths)
[...]

PR #2175, opened 2024-02-28
[python] Expose `tiledbsoma.io.ExperimentAmbientLabelMapping`

johnkerl:
**Issue and/or context:** The subpackage `tiledbsoma.io._registration` is mostly internals, and rightly so. There are `tiledbsoma.io.register_anndatas` and `tiledbsoma.io.register_h5ads` exposed at `tiledbsoma.io`, and this is _almost_ sufficient.

I noted on #2172, though, something I should have thought of -- namely, if any user wants to do Python type-checking on the _result_ of `tiledbsoma.io.register_anndatas` or `tiledbsoma.io.register_h5ads` then tney currently need to cal this `tiledbsoma.io._registration.ExperimentAmbientLabelMapping`. This is non-ideal since they have to type out something with a leading underscore.

**Changes:** Expose `tiledbsoma.io.ExperimentAmbientLabelMapping`

**Notes for Reviewer:**

### DIFF
--- a/apis/python/tests/test_registration_mappings.py
+++ b/apis/python/tests/test_registration_mappings.py
@@ -1098,3 +1098,9 @@ def test_registration_with_batched_reads(tmp_path, soma_larger, use_small_buffer
 
 def test_ealm_expose():
     """Checks that this is exported from tiledbsoma.io._registration"""
+    # All we want to check is that the import doesn't throw. Job done. Period.
+    # However, the pre-commit hook will strip out this import statement as "unused".
+    # So, assert something.
+    from tiledbsoma.io import ExperimentAmbientLabelMapping
+
+    assert ExperimentAmbientLabelMapping.from_h5ad_appends_on_experiment is not None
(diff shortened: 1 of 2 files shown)


apis/python/tests/test_registration_mappings.py
```diff
@@ -1094,3 +1094,7 @@ def test_registration_with_batched_reads(tmp_path, soma_larger, use_small_buffer
     )
 
     assert len(rd.obs_axis.data) == 1000
+
+
+def test_ealm_expose():
+    """Checks that this is exported from tiledbsoma.io._registration"""
```
thetorpedodog, line 1100: Is there something missing here?

thetorpedodog (approved):

apis/python/tests/test_registration_mappings.py
johnkerl: it does the import and if the import fails it'll throw

johnkerl: .... oh crap, the pre-commit hook removed that line :headdesk:

thetorpedodog: I believe `import whatever as _` (or in this case, `from what import ever as _`) will also preserve a side-effect import. Alternately, since `tiledbsoma.io` is already imported by this file, you could just say `assert tiledbsoma.io.ExperimentLabelMapping is not None`.

thetorpedodog (approved):

PR #2186, opened 2024-03-01
[c++] Address issues related to reindexer advanced features

beroy:
**Issue and/or context:**
Addressing minor issues related to reindexer's shared threadpool and its optimization for pandas and pyarrow: review comments from #2148 and #2159.

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/soma/soma_context.h
+++ b/libtiledbsoma/src/soma/soma_context.h
@@ -71,16 +71,7 @@ class SOMAContext {
         return cfg;
     }
 
-    std::shared_ptr<ThreadPool>& thread_pool() {
-        const std::lock_guard<std::mutex> lock(thread_pool_mutex_);
-        // The first thread that gets here will create the context thread pool
-        if (thread_pool_ == nullptr) {
-            create_thread_pool();
-        }
-        return thread_pool_;
-    }
-
-    void create_thread_pool();
+    std::shared_ptr<ThreadPool>& thread_pool();
 
    private:
     //===================================================================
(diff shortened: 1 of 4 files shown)


libtiledbsoma/src/soma/soma_context.h
```diff
@@ -71,14 +71,7 @@ class SOMAContext {
...
-        const std::lock_guard<std::mutex> lock(thread_pool_mutex_);
-        // The first thread that gets here will create the context thread pool
-        if (thread_pool_ == nullptr) {
-            create_thread_pool();
-        }
-        return thread_pool_;
-    }
+    std::shared_ptr<ThreadPool>& thread_pool();
 
     void create_thread_pool();
```
nguyenv, line 76: We should be able to remove this from the header.

beroy: @nguyenv can you elaborate how? It's function of this class and I'm just putting declaration here. The implementation is in C++.

nguyenv: I would just literally delete the entire create_thread_pool and just do

```
std::shared_ptr<ThreadPool>& SOMAContext::thread_pool() {
    const std::lock_guard<std::mutex> lock(thread_pool_mutex_);
    // The first thread that gets here will create the context thread pool
    if (thread_pool_ == nullptr) {
        auto cfg = tiledb_config();
        int concurrency = 10;
        if (cfg.find("sm.compute_concurrency_level") != cfg.end()) {
            concurrency = std::stoi(cfg["sm.compute_concurrency_level"]);
        }
        int thread_count = std::max(1, concurrency / 2);
        if (thread_count > 1) {
            thread_pool_ = std::make_shared<ThreadPool>(thread_count);
        }
    }
    return thread_pool_;
}
```

beroy: OK, I will inline the function.

nguyenv (approved):
Thanks! LGTM

johnkerl (approved):

PR #2188, opened 2024-03-01
[r] [WIP - Do Not Merge Yet] Update and refactor nanoarrow

eddelbuettel:
**Issue and/or context:**

We rely on `nanoarrow` to take advantage of higher-level structures for the (otherwise 'naked' `void *` of the C API of Arrow.  `nanoarrow` helps here, and has made good strides in crecent releases.

**Changes:**

We update the 'vendored' `nanoarrow` to release 0.4.0, and make extended use of it.  This works well: simple examples work fine and show the desired memory behaviour.   However, on larger examples we see interactions with (local) handling of Arrow objects in `libtiledbsoma` so the scope of the PR will have to extended to handle allocating, releases and setting up Arrow objects via `nanoarrow`.

**Notes for Reviewer:**

This PR is informational to run some CI; all local tests pass.  However, _at present_ an invasice change in `libtiledbsoma` makes this unsuitable for a merge _at present_ meaning it is also not yet fully ready for a review.


.github/workflows/r-ci.yml
```diff
@@ -79,13 +79,13 @@ jobs:
       #  if: ${{ matrix.os != 'macOS-latest' }}
       #  run: cd apis/r && Rscript -e "options(bspm.version.check=TRUE); install.packages('tiledb', repos = c('https://eddelbuettel.r-universe.dev/bin/linux/jammy/4.3/', 'https://cloud.r-project.org'))"
 
-      - name: Install r-universe build of SeuratObject (macOS)
-        if: ${{ matrix.os == 'macOS-latest' }}
-        run: cd apis/r && Rscript -e "install.packages('SeuratObject', repos = c('https://mojaveazure.r-universe.dev', 'https://cloud.r-project.org'))"
+      #- name: Install r-universe build of SeuratObject (macOS)
```
johnkerl, line 82: @mojaveazure do these two for SeuratObject need to be "in" or "out"? The (very helpful) comment material above tells us under which conditions we need the two tiledb-r stanzas in or out ... but what about these two SeuratObject stanzas?

apis/python/src/tiledbsoma/reindexer.cc
```diff
@@ -31,7 +31,7 @@
  */
 
 #include <tiledbsoma/reindexer/reindexer.h>
-#include <tiledbsoma/utils/carrow.h>
+// #include <tiledbsoma/utils/carrow.h>
```
johnkerl, line 34: Remove?

apis/r/src/riterator.cpp
```diff
@@ -198,16 +216,25 @@ Rcpp::List sr_next(Rcpp::XPtr<tdbs::SOMAArray> sr) {
...
+   exitIfError(ArrowArrayInitFromType(arr, NANOARROW_TYPE_STRUCT), "Bad array init");
+   exitIfError(ArrowArrayAllocateChildren(arr, ncol), "Bad array children alloc");
+
+   arr->length = 0;             // initial value
 
    for (size_t i=0; i<ncol; i++) {
        // this allocates, and properly wraps as external pointers controlling lifetime
-       Rcpp::XPtr<ArrowSchema> chldschemaxp = schema_owning_xptr();
-       Rcpp::XPtr<ArrowArray> chldarrayxp = array_owning_xptr();
+       //Rcpp::XPtr<ArrowSchema> chldschemaxp = schema_owning_xptr();
```
johnkerl, line 236: Remove?

apis/r/src/riterator.cpp
```diff
@@ -217,21 +244,19 @@ Rcpp::List sr_next(Rcpp::XPtr<tdbs::SOMAArray> sr) {
        // this is pair of array and schema pointer
        auto pp = tdbs::ArrowAdapter::to_arrow(buf);
 
-       memcpy((void*) chldschemaxp, pp.second.get(), sizeof(ArrowSchema));
-       memcpy((void*) chldarrayxp, pp.first.get(), sizeof(ArrowArray));
-
-       schemaxp->children[i] = chldschemaxp;
-       arrayxp->children[i] = chldarrayxp;
+       //memcpy((void*) sch->children[i], pp.second.get(), sizeof(ArrowSchema));
```
johnkerl, line 247: Remove?

apis/r/src/rutilities.h
```diff
@@ -62,3 +62,17 @@ struct ContextWrapper {
     std::shared_ptr<tiledb::Context> ctxptr;
 };
 typedef struct ContextWrapper ctx_wrap_t;
+
+// some lipstick on the (plain C language) pig that is a SEXP:
```
johnkerl, line 66:
```suggestion
// some decoration on the (plain C language) SEXP:
```

apis/r/tests/testthat/test-SCEOutgest.R
```diff
@@ -1,4 +1,5 @@
 test_that("Load SCE object from ExperimentQuery mechanics", {
+  #if (Sys.getenv("GITHUB_ACTION") != "")  set_log_level("trace")
```
johnkerl, line 2: remove?

apis/r/tests/testthat/test-SCEOutgest.R
```diff
@@ -358,4 +359,5 @@ test_that("Load SCE object from indexed ExperimentQuery", {
   )
   expect_identical(SingleCellExperiment::colPairNames(obj), 'connectivities')
   expect_identical(SingleCellExperiment::rowPairNames(obj), 'network')
+  #if (Sys.getenv("GITHUB_ACTION") != "")  set_log_level("warn")
```
johnkerl, line 362: remove?

libtiledbsoma/CMakeLists.txt
```diff
@@ -187,10 +187,6 @@ if(MSVC)
 else()
   add_compile_options(-Wall -Wextra)
 
-  if(TILEDBSOMA_ENABLE_WERROR)
```
johnkerl, line 190: Why is this being removed on this PR?

libtiledbsoma/src/cli/cli.cc
```diff
@@ -33,7 +33,7 @@
 #include "soma/enums.h"
 #include "soma/soma_array.h"
 #include "utils/arrow_adapter.h"
-#include "utils/carrow.h"
+// #include "utils/carrow.h"
```
johnkerl, line 36: Remove?

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -8,9 +8,11 @@
 // https://arrow.apache.org/docs/format/Columnar.html#buffer-listing-for-each-layout
 // https://arrow.apache.org/docs/format/CDataInterface.html#exporting-a-simple-int32-array
 
-#ifndef ARROW_SCHEMA_AND_ARRAY_DEFINED
-#include "carrow.h"
-#endif
+#include "nanoarrow.hpp"
+// #ifndef ARROW_SCHEMA_AND_ARRAY_DEFINED
```
johnkerl, line 12: Remove?

johnkerl (commented):
Some lignt neatens, and some questions.

Additionally: what mechanism (if any) do we have to keep the two copies of `nanoarrow` source in this package in sync?

```
johnkerl@exade[de/sc-42316/update_and_refactor_nanoarrow][TileDB-SOMA]$ md5sum $(find . -name nanoarrow\*) | sort
0d4b6c9774faf52256a0089c1125dcaf  ./apis/r/src/nanoarrow.h
16ff8d1a344f1f6abc5d9ef96262ea5a  ./apis/r/src/nanoarrow.hpp
16ff8d1a344f1f6abc5d9ef96262ea5a  ./libtiledbsoma/src/utils/nanoarrow.hpp
20c90c6ed488402344d9792254bf87af  ./apis/r/src/nanoarrow.c
20c90c6ed488402344d9792254bf87af  ./libtiledbsoma/src/utils/nanoarrow.c
e2e271a604732d0c2be136a3f0dd2fe7  ./libtiledbsoma/src/utils/nanoarrow.h
```

suggests that the two copies of `nanoarrow.c` are (currently) identical as are the two copies of `nanoarrow.hpp` but the two copies of `nanoarrow.h` are distinct.

libtiledbsoma/CMakeLists.txt
eddelbuettel: It has been a while. Maybe try locally without it?  I think it may break.

Keep in mind that some people have settings newer than Ubuntu 22.04 and hence newer (pickier) compilers.  I believe we have a remnant compiler warning.

Redoing CMake in a follow-up is a TODO.

eddelbuettel:
The dual nanoarrows (from inside the R package, and inside libtiledbsoma) can likely be remedied. The location in tiledbsoma should be in externals/ anyway -- a follow-up PR seems apt.

johnkerl:
> The dual nanoarrows (from inside the R package, and inside libtiledbsoma) can likely be remedied. The location in tiledbsoma should be in externals/ anyway -- a follow-up PR seems apt.

https://github.com/single-cell-data/TileDB-SOMA/issues/2362

libtiledbsoma/CMakeLists.txt
johnkerl: > Maybe try locally without it? 

@eddelbuettel I'd like you to try locally without it.

.github/workflows/r-ci.yml
mojaveazure: These two install development versions of SeuratObject to ensure compatibility with up-coming. However, they are not necessary as the current version (SeuratObject v5.0.1) has all the fixes that these stanzars were originally written for

johnkerl: @eddelbuettel in that case let's please remove those stanzas.

eddelbuettel:
@johnkerl One triplet of `nanoarrow` files is now gone.

libtiledbsoma/CMakeLists.txt
eddelbuettel: @johnkerl It is possible that I can't as it doesn't (fully) build for me as I may have communicated a few times.  I have libfmt9 leading to errors with `bld`.  Which is why I suggested you may try too.

eddelbuettel: @johnkerl It seems to work here _for the more limited builds_ I do.  I can turn it back on, if stuff breaks we can deal with it.

apis/r/src/rutilities.h
johnkerl: I want  you to not use this language please

johnkerl (changes requested):

apis/r/src/rutilities.h
```diff
@@ -63,7 +63,7 @@ struct ContextWrapper {
 };
 typedef struct ContextWrapper ctx_wrap_t;
 
-// some lipstick on the (plain C language) pig that is a SEXP:
+// some lipstick on the (plain C language) [redacted] that is a SEXP:
```
johnkerl, line 66:
```suggestion
// some decoration on the (plain C language) SEXP:
```

johnkerl (changes requested):

johnkerl (approved):
🚢

johnkerl:
Thanks @eddelbuettel !!

Just to set an expectation, this may not backport automatically to `release-1.9` since `apis/r/DESCRIPTION` is touched and that differs between `main` and `release-1.9` -- not to worry, I'll do a manual backport tomorrow morning if so

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -108,17 +154,20 @@ std::unique_ptr<ArrowSchema> ArrowAdapter::arrow_schema_from_tiledb_array(
     auto nattr = tiledb_schema.attribute_num();
 
     std::unique_ptr<ArrowSchema> arrow_schema = std::make_unique<ArrowSchema>();
-    arrow_schema->format = "+s";
+    arrow_schema->format = strdup("+s");
```
nguyenv, line 157: I know I'm the one who added this but I'm not sure that it's necessary here. I may have overcorrected.

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -443,4 +540,51 @@ std::string_view ArrowAdapter::to_arrow_format(
...
+        return NANOARROW_TYPE_INT64;  // NB time resolution set indepedently
+    else if (sv == "tdD")
+        return NANOARROW_TYPE_INT32;  // R Date: fractional days since epoch
+    else if (sv == "z")
+        return NANOARROW_TYPE_BINARY;
+    else if (sv == "Z")
+        return NANOARROW_TYPE_LARGE_BINARY;
+    else
+        throw TileDBSOMAError(fmt::format(
+            "ArrowAdapter: Unsupported TileDB datatype string: {} ", sv));
```
nguyenv, line 587: Nanoarrow format strings here, not TileDB

libtiledbsoma/src/utils/arrow_adapter.cc
eddelbuettel: I think it is. It is just stylistic. If you peek into what `nanoarrow` does it actually allocates explicitly (as do I over in tiledb-r)

```c
ArrowErrorCode ArrowSchemaSetFormat(struct ArrowSchema* schema, const char* format) {
  if (schema->format != NULL) {
    ArrowFree((void*)schema->format);
  }

  if (format != NULL) {
    size_t format_size = strlen(format) + 1;
    schema->format = (const char*)ArrowMalloc(format_size);
    if (schema->format == NULL) {
      return ENOMEM;
    }

    memcpy((void*)schema->format, format, format_size);
  } else {
    schema->format = NULL;
  }

  return NANOARROW_OK;
}
```

This is different from what it also has 'view' inits for schema which just borrow the pointed to memory because the contract the viewer can assume the viewed object exists.

In short, your `strdup()` was fine.  And after 30+ plus years of C it possibly taught me another K+R (at style) function so all good!

libtiledbsoma/src/utils/arrow_adapter.cc
eddelbuettel: Actually Arrow format string (which is the same as Arrow).  Will adjust.

PR #1937, opened 2023-11-30
[python] Do not copy `libtiledbsoma.so` into `package_data` if it already exists

nguyenv:
**Issue and/or context:**

- The copying issue was reported in internal channel by @jdblischak: https://tiledb.slack.com/archives/C052CFC4D2Q/p1700678097748879
- This PR also includes a fix for https://github.com/single-cell-data/TileDB-SOMA/issues/1991

**Changes:**

| TileDB Exists  | TileDB SOMA Exists | Action |
| -------------- | ---------------------- | -------------- | 
| No                 | No                            | Both TileDB and TileDB-SOMA get copied |
| No                 | Yes                            | Error: The TileDB shared object that is linked to TileDB-SOMA shared object needs to exist somewhere |
| Yes                 | No                            | Copy TileDB-SOMA only |
| Yes                 | Yes                            | No copies |

PR #2208, opened 2024-03-05
[ci] Run `python-so-copying.yml` on Linux and Mac both [WIP]

johnkerl:
**Issue and/or context:** My current theory on https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/95 and https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/91 is that this code path, not being regression-tested on MacOS, is perhaps faulty on MacOS. Regardless, the logic which was developed on #1937 needs testing on MacOS, and it was an oversight on #1937 for us to have only tested on Linux.

See also #2210 for tracking.

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/.github/workflows/python-so-copying.yml
+++ b/.github/workflows/python-so-copying.yml
@@ -22,7 +22,6 @@ defaults:
 
 jobs:
   build:
-    runs-on: ${{ matrix.os }}
     name: "TILEDB_EXISTS: ${{ matrix.TILEDB_EXISTS }} TILEDBSOMA_EXISTS: ${{ matrix.TILEDBSOMA_EXISTS }}"
     strategy:
       fail-fast: false
@@ -33,6 +32,7 @@ jobs:
         exclude:
           - TILEDB_EXISTS: "no"
             TILEDBSOMA_EXISTS: "yes"
+    runs-on: ${{ matrix.os }}
     container:
       image: ${{ matrix.os }}
     steps:
--- a/.github/workflows/python-so-copying.yml
+++ b/.github/workflows/python-so-copying.yml
@@ -22,18 +22,19 @@ defaults:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


.github/workflows/python-so-copying.yml
```diff
@@ -22,18 +22,19 @@ defaults:
...
+        os: [ubuntu-22.04, macos-12]
         TILEDB_EXISTS: ["no", "yes"]
         TILEDBSOMA_EXISTS: ["no", "yes"]
         exclude:
           - TILEDB_EXISTS: "no"
             TILEDBSOMA_EXISTS: "yes"
+    runs-on: ${{ matrix.os }}
     container:
-      image: ubuntu:22.04
+      image: ${{ matrix.os }}
```
jdblischak, line 37: This is the tricky part. This needs to be a Docker image, not the name of the GitHub Actions runner.

The linux job runs inside of a Docker container (this made it easier for me to test locally). Thus it's not as easy as adding `macos-12` to the build matrix. We need to either 1) make the container directive only apply to the  ubuntu job (probably possible, but not immediately clear to me how), or 2) create a second job that runs similar code on macOS (which might be worth it given all the changes to accommodate running the same steps on linux and macOS)

jdblischak (changes requested):
Thanks for getting this started!

.github/workflows/python-so-copying.yml
johnkerl: @jdblischak I'm totally happy with well-documented code-duplication here.

jdblischak: I did find some [workarounds](https://stackoverflow.com/questions/76919426/is-it-possible-to-use-docker-for-linux-in-a-matrix-of-operating-systems) to create a build matrix that builds natively on macOS but in Docker on Ubuntu. There is no `if`, so the key is to pass `null` to `container` for the macOS job. Overall the readability is not great though.

Also, I realized that we'll also need to download a different pre-built libtiledb binary for the macOS build, which makes the option to have a completely separate macOS build all the more appealing.

PR #2209, opened 2024-03-05
[release-1.8] Revert #1937

johnkerl:
Temporarily reverts #1937 for the `release-1.8` branch until @nguyenv is back from PTO.

Per discussion with @jdblischak, #1937 is broken for MacOS. #2208 is my WIP attempt to deal with that.

Meanwhile we'll revert to unblock https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/95 and https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/91 in order to unblock `tiledbsoma` 1.7.3 and 1.8.0. Then, end of this week, with @nguyenv back in-office, we can try again.

Do note that I'm leaving #1937 in place on `main`, but reverting on the `release-1.7` and `release-1.8` branches, so we can pick up on the debug of #2208.

See also https://github.com/single-cell-data/TileDB-SOMA/issues/2210 for tracking.

See also #2219 for tracking.


jdblischak (approved):

johnkerl:
Closing: #2220 proves that #1937 did not entirely break dylib-copying for MacOS. All we need is an upcoming PR from @jdblischak.

PR #2216, opened 2024-03-06
[release-1.7] Append `LD_LIBRARY_PATH` for census

johnkerl:
Per Slack conversation with @bkmartinjr 


bkmartinjr (approved):
See simplification proposed in #2218

PR #2217, opened 2024-03-06
[release-1.8] Append `LD_LIBRARY_PATH` for census

johnkerl:
Sibling of #2216


bkmartinjr (approved):
See simplification proposed in #2218

PR #2218, opened 2024-03-06
Append `LD_LIBRARY_PATH` for census

johnkerl:
Sibling of #2216 and #2217

### DIFF
--- a/scripts/bld
+++ b/scripts/bld
@@ -70,17 +70,29 @@ if [ -n "${tiledb}"  ]; then
   extra_opts+=" -DFORCE_BUILD_TILEDB=OFF"
   export TileDB_DIR="${tiledb}"
 
-  # Since set -u:
-  if [[ -v LD_LIBRARY_PATH ]]; then
-    export LD_LIBRARY_PATH="${tiledb}:${LD_LIBRARY_PATH}"
-  else
+  # It's more elegant to use
+  #   if [[ -v LD_LIBRARY_PATH ]]; then
+  #     ...
+  #   fi
+  # -- however, this is supproted by bash not sh. And this script can
+  # be invoked by python via its shell-out which seems to use sh.
+  # It's simplest to just pause set -u.
+
+  set +u
+
+  if [ -z $"LD_LIBRARY_PATH" ]; then
     export LD_LIBRARY_PATH="${tiledb}"
-  fi
-  if [[ -v DYLD_LIBRARY_PATH ]]; then
-    export DYLD_LIBRARY_PATH="${tiledb}:${DYLD_LIBRARY_PATH}"
   else
-    export DYLD_LIBRARY_PATH="${tiledb}"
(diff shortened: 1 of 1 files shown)


scripts/bld
```diff
@@ -69,8 +69,30 @@ if [ -n "${tiledb}"  ]; then
...
+  #   if [[ -v LD_LIBRARY_PATH ]]; then
+  #     ...
+  #   fi
+  # -- however, this is supproted by bash not sh. And this script can
+  # be invoked by python via its shell-out which seems to use sh.
+  # It's simplest to just pause set -u.
+
+  set +u
+
+  if [ -z $"LD_LIBRARY_PATH" ]; then
```
bkmartinjr, line 83: I am under the impression that the linker ignored leading/trailing semi-colons, in which case you can just do:
> export LD_LIBRARY_PATH="${tiledb}:${LD_LIBRARY_PATH}"

that said, the extra code should not hurt anything...

bkmartinjr (approved):
left a suggested simplification inline.

scripts/bld
johnkerl: @bkmartinjr I'd rather not count on that; who knows what system's linker somewhere is non-compliant -- safety is best -- thank you though!

ryan-williams: `$"LD_LIBRARY_PATH"` evaluates to the string `LD_LIBRARY_PATH`; the `$` needs to be inside the quotes:

```bash
docker run --rm --entrypoint bash bash -c 'foo=1; echo $"foo" "$foo"'
# foo 1
```

`[ -z $"LD_LIBRARY_PATH" ]` is always false, so we've inadvertently been doing what @bkmartinjr suggested (always prepending `${tiledb}:`). I think this is also why the `set +u` seemed necessary; the "append" block was running even with `LD_LIBRARY_PATH` unset, which `set -u` was flagging as an error.

A common way of doing these two `if` blocks is:
```bash
export LD_LIBRARY_PATH="$tiledb${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
export DYLD_LIBRARY_PATH="$tiledb${DYLD_LIBRARY_PATH:+:$DYLD_LIBRARY_PATH}"
```
cf. [SO](https://stackoverflow.com/a/46892074). This also works in `sh`:
```sh
docker run --rm -it --entrypoint sh ubuntu -c 'foo=; echo "prepended_val${foo:+:$foo}"'
# prepended_val
docker run --rm -it --entrypoint sh ubuntu -c 'foo=val; echo "prepended_val${foo:+:$foo}"'
# prepended_val:val
```
though I'd think this file's shebang line ensures we're always running `bash`.

I will send a PR with a fix, as well as other changes recommended by [shellcheck](https://www.shellcheck.net/) (which flagged this issue to me).

johnkerl: Thank you @ryan-williams -- `$"` is very clearly wrong and I'm absolutely appalled that I missed a basic shell-syntax error on reviewing this :(

Issue #2210, opened 2024-03-05
[ci] dylib-copy logic appears to be broken on MacOS


johnkerl:
Please see:

* https://github.com/single-cell-data/TileDB-SOMA/pull/2209
* https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/95
* https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/91
* https://github.com/single-cell-data/TileDB-SOMA/pull/2208
* https://github.com/single-cell-data/TileDB-SOMA/pull/2207
* To do:
  * #2207 has backports #2212 and #2213
  * #2211 and #2209 remove the so-copy logic, and as a result, they also remove #2212 and #2213
  * So once we unbreak the so-copy logic on `main`, and resolve either #2208 or a TBD PR which supersedes it, we must be sure to re-apply #2212 and #2213 on the `release-1.7` and `release-1.8` branches

Also note:
* https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/91 and https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/95 final commits have `cmake` added to their `build` dependencies
  * This is a workaround not a fix since the feedstocks should not _have_ to build from source
  * The fact that they're building from source at all on Mac is because the so-copy logic is broken on Macs as narrated above

See also https://github.com/single-cell-data/TileDB-SOMA/issues/2219 for tracking.

johnkerl:
PR #2220 proves that PR #1937 did not entirely break dylib-copying for MacOS. All we need is an upcoming PR from @jdblischak.

johnkerl:
#2221 should be the final fix; then we can close this issue.

PR #2220, opened 2024-03-06
[ci] Add MacOS job to test shared object copying

jdblischak:
**Issue and/or context:**

* Follow up to #1937 
* Linked Issue: #2210

**Changes:**

Expands SO copying pipeline to include a macOS job

**Notes for Reviewer:**

I have good news and bad news. The good news is that I added a macOS job, and its behavior is correct! The bad news of course is that this means it doesn't explain the feedstock failures where the macOS builds fail to find the shared objects.

I suspect this is because the feedstock is still using the older `setup.py` with the flag `--libtiledbsoma` ([script](https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/6e28425785020e98e088390bec528a13eeea2dc6/recipe/build-tiledbsoma-py.sh#L8)). This was required pre-#1937. However, now that #1937 has been merged, we can now upgrade to the more modern alternative:

```sh
export TILEDB_PATH=$PREFIX
export TILEDBSOMA_PATH=$PREFIX
$PYTHON -m pip install --no-deps . -vv
```

**update:** replaced jinja2 `{{ PYTHON }}` (for use in `meta.yaml`) with `$PYTHON` (for use in separate build script)

### DIFF
--- a/.github/workflows/python-so-copying.yml
+++ b/.github/workflows/python-so-copying.yml
@@ -16,14 +16,11 @@ on:
       - 'libtiledbsoma/cmake/**'
   workflow_dispatch:
 
-defaults:
-  run:
-    shell: bash
-
 jobs:
-  build:
+
+  docker:
     runs-on: ubuntu-latest
-    name: "TILEDB_EXISTS: ${{ matrix.TILEDB_EXISTS }} TILEDBSOMA_EXISTS: ${{ matrix.TILEDBSOMA_EXISTS }}"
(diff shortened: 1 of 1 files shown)


johnkerl (approved):
🚢

johnkerl:
Backports to `release-1.8` and `release-1.7` have failed to launch in CI multiple times, e.g.

https://github.com/single-cell-data/TileDB-SOMA/actions/runs/8176608919

```
Backport
Error: The process '/usr/bin/git' failed with exit code 128
Backport
HttpError: Resource not accessible by integration
Has anyone seen this before? I can debug but am already debugging multiple things simultaneously in an attempt to unblock multiple teams, so I thought I'd crowdsource first ...
```

I'm not convinced this is urgently needed in the branches

PR #2229, opened 2024-03-07
[ci] Only run build tests when build files are modified

jdblischak:
**Issue and/or context:**

* Follow up to #1937, #2220, and #2221 

**Changes:**

* Only runs this CI workflow that tests the build steps if the PR/commit modifies a file that involves finding libtiledb and/or libtiledbsoma at build/run time

**Notes for Reviewer:**

* These tests are quick, but we noticed that the recently added macOS builds are extremely flaky, often requiring a restart to succeed. It's unclear to me why the macOS builds are so less reliable than the Docker ones

* For the Python package, I included both `setup.py` (finds libtiledbsoma at build time) and `src/tiledbsoma/__init__.py` (finds libtiledbsoma at runtime)

### DIFF
--- a/.github/workflows/python-so-copying.yml
+++ b/.github/workflows/python-so-copying.yml
@@ -7,13 +7,25 @@ on:
   push:
     paths:
       - '.github/workflows/python-so-copying.yml'
-      - 'apis/python/**'
-      - 'libtiledbsoma/cmake/**'
+      - 'apis/python/setup.py'
+      - 'apis/python/src/tiledbsoma/__init__.py'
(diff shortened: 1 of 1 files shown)


johnkerl (approved):
🚢

jdblischak:
And a general comment, we may want to be more aggressive with the path-based filters for our other GitHub Actions workflows as well. An awful lot of CI is running on this PR that only modifies another CI file

johnkerl:
Backporting manually since #2231 :(

Issue #1791, opened 2023-10-15
[python] Type signature for `SparseNDArray.read coord` is overly restrictive


bkmartinjr:
SparseNDArray.read accepts a Arrow chunked array of integers as an acceptable coordinate.  However, the SparseNDCoord type signature does not accept ChunkedArray, creating false errors with mypy, etc.

The current definition uses pa.IntegerArray, which is a sub-class of pa.Array.  Also needed is a pa.ChunkedArray (AFAIK, there is no way to express a more specialized sub-class of ChunkedArray in the current pyarrow).

```
SparseNDCoord = Union[
    None,
    ValSliceOrSequence[int],
    npt.NDArray[np.integer],
    pa.IntegerArray,
]
"""A single coordinate range for one dimension of a sparse nd-array."""
SparseNDCoords = Sequence[SparseNDCoord]
"""A sequence of coordinate ranges for reading sparse ndarrays."""
```

johnkerl:
To be assigned to @ryan-williams

johnkerl:
[sc-41751]

johnkerl:
@ryan-williams please put up a PR on `TileDB-SOMA` bumping to `somacore` 1.0.9 here:

```
.pre-commit-config.yaml
18:        - "somacore==1.0.8"

apis/python/setup.py
332:        "somacore==1.0.8",
```

ryan-williams:
@johnkerl #2251 is merged, lmk if anything else should happen here

PR #2234, opened 2024-03-07
[python] Verify `pa.{Chunked,Integer}Array` args to `SparseNDArray.read`

ryan-williams:
### Issue and/or context
#1791 

### Changes
Add a test case to `test_sparse_nd_array_table_slicing`, verifying that `pa.ChunkedArray` (of integers) and `pa.IntegerArray` work as `SparseNDCoord`s.

### Notes for Reviewer
Ideally we'd have a repro where:
- Type checker fails when `pa.ChunkedArray` passed to `SparseNDArray.read` (before any changes to `SparseNDCoord` [in somacore](https://github.com/single-cell-data/SOMA/blob/76b2d2bbca7c8baf543b5cc9cf71fc5c9138a39a/python-spec/src/somacore/options.py#L172-L179); see https://github.com/single-cell-data/SOMA/pull/187).
- Same checks succeed when `pa.ChunkedArray` is added to `SparseNDCoord`.

The existing test I've modified here apparently isn't subjected to such runtime type-checking. A more explicit version in https://github.com/single-cell-data/TileDB-SOMA/commit/0d93315e34d86549524959ba5fec98f7818779cd also doesn't exhibit a type-check failure. I'll keep looking into how to accomplish that.

### DIFF
--- a/apis/python/tests/test_sparse_nd_array.py
+++ b/apis/python/tests/test_sparse_nd_array.py
@@ -829,6 +829,19 @@ def test_csr_csc_2d_read(tmp_path, shape):
             },
             "throws": None,
         },
+        {
+            "name": "2D coords=(pa_chunked[1, 2], pa64[3, 4])",
+            "shape": (9, 11),
+            "coords": (
+                pa.chunked_array([[1], [2]]),
+                pa.array([3, 4]),
+            ),
+            "dims": {
+                "soma_dim_0": [1, 1, 2, 2],
+                "soma_dim_1": [3, 4, 3, 4],
+            },
+            "throws": None,
+        },
     ],
     ids=lambda io: io.get("name"),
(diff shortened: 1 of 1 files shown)


johnkerl:
> The existing test I've modified here apparently isn't subjected to such runtime type-checking. A more explicit version in https://github.com/single-cell-data/TileDB-SOMA/commit/0d93315e34d86549524959ba5fec98f7818779cd also doesn't exhibit a type-check failure. I'll keep looking into how to accomplish that

This surprises me as we do use `typeguard` in our unit tests

ryan-williams:
It seems like typeguard may not be running as intended: https://github.com/single-cell-data/TileDB-SOMA/issues/2238

johnkerl (approved):

PR #2235, opened 2024-03-07
Typo fixes, broken README link

ryan-williams:
Re: the first commit, https://github.com/single-cell-data/TileDB-SOMA/commit/9870a3fe7baf016c192519b773edf85194478d6a:
- [3rd paragraph of README](https://github.com/single-cell-data/TileDB-SOMA?tab=readme-ov-file#tiledb-soma) ("Get started…") links to `#quick-start`, which doesn't exist.
- Relevant section appears to be [APIs Installation and Quick Start](https://github.com/single-cell-data/TileDB-SOMA?tab=readme-ov-file#apis-installation-and-quick-start) a.k.a. `#apis-installation-and-quick-start`.
- Rather than "hard-code" that, I appended an empty anchor tag with a canonical `id` (`#quick-start`) to the end of the &lt;h2&gt; line. I've used this technique in the past and found it preferable / never had problems, but can use Github's default auto-generated dash-cased version if that is preferred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ TileDB-SOMA provides interoperability with existing Python or R data structures:
 channel `#cellxgene-census-users`.
 
 
-## APIs Installation and Quick Start
+## APIs Installation and Quick Start <a id="quick-start"></a>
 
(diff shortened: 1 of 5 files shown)


johnkerl (approved):
🚢

Issue #1565, opened 2023-08-01
[python] Second `add_matrix_to_collection` in one `with` fails "not open for write"


johnkerl:
Repro snippet:

```
with tiledbsoma.open(exp.uri, "w") as exp_w:
    tiledbsoma.io.add_matrix_to_collection(
        exp=exp_w,
        measurement_name="RNA",
        collection_name="obsm",
        matrix_name="logcounts_pca",
        matrix_data=pbmc3k.obsm["X_pca"],
    )

    tiledbsoma.io.add_matrix_to_collection(
        exp=exp_w,
        measurement_name="RNA",
        collection_name="varm",
        matrix_name="logcounts_pcs",
        matrix_data=pbmc3k.varm["PCs"],
    )
```

The second fails with

```
tiledbsoma._exception.SOMAError: Collection must be open for write: file:///var/folders/7l/_wsjyk5d4p3dz3kbz7wxn7t00000gn/T/tmpw49h7i_o/ms/RNA
```

which AFAICT is due to this here:
https://github.com/single-cell-data/TileDB-SOMA/blob/86b0298eccb70d9d27ea4b46498b79fc1a868145/apis/python/src/tiledbsoma/io/ingest.py#L885

Namely, the inner `with` in `io/ingest.py` closes the measurement, while the outer `with` at the callsite has the experiment held open.

Needs some unit-testing of two `add_matrix_to_collection` in a single callsite `with`.

johnkerl:
[sc-41750]

PR #2236, opened 2024-03-07
[python] Attempt to repro #1565

ryan-williams:
**Issue and/or context:** #1565 

**Notes for Reviewer:** unsuccessful first attempt to repro #1565. Either the issue was fixed, or I'm missing something necessary to repro it. Thoughts welcome.


johnkerl:
> Thoughts welcome

IMO your new unit-test case proves the problem is solved ... I think this is plenty

johnkerl (approved):

PR #2244, opened 2024-03-11
Fix building TileDB if a CMake toolchain is not specified.

teo-tsirpanis:
**Issue and/or context:**

**Changes:**

When configuring the Core, if the variable `CMAKE_TOOLCHAIN_FILE` is not specified, we will pass `-DCMAKE_TOOLCHAIN_FILE=`, which means that the variable is empty but also _defined_, which will cause [this branch](https://github.com/TileDB-Inc/TileDB/blob/faa42b9d70695f0a94ae062a27f8866720bff23b/cmake/Options/TileDBToolchain.cmake#L39) to not be entered, which will cause vcpkg to not be downloaded, which will cause failures in `find_package`.

We now pass `-DCMAKE_TOOLCHAIN_FILE` to `ep_tiledb` only if the variable in the parent project has a value.

**Notes for Reviewer:**

Validated by @ryan-williams; thanks for finding the issue!


ryan-williams (approved):
lgtm, thanks.

Here is the validation I ran, for reference: [ryan-williams/tiledb-scratch@dbg-build](https://github.com/ryan-williams/tiledb-scratch/tree/dbg-build)

johnkerl (approved):

PR #2257, opened 2024-03-13
[r/ci] Set `options(pkgType = "binary")` for MacOS

### LINKED ISSUE #2254: [r/ci] Analyze `arrow` build-from-source on MacOS in R CI
## Problem

Slow CI time (30 extra minutes per run) on MacOS due to build-from-source of the R `arrow` package.

As raised by @eddelbuettel.

## Details

CI runs https://github.com/single-cell-data/TileDB-SOMA/actions/workflows/r-ci.yml

Reading log-files there reveal this leads to "Dependencies" https://github.com/single-cell-data/TileDB-SOMA/blob/1.8.0/.github/workflows/r-ci.yml#L77-L78. This calls https://github.com/single-cell-data/TileDB-SOMA/blob/1.8.0/apis/r/tools/r-ci.sh#L356-L358

Logs:

* Full-output example https://gist.github.com/johnkerl/45747955555e1cb8ac1e5670b9e8f819

* Issue 
  * https://gist.github.com/johnkerl/45747955555e1cb8ac1e5670b9e8f819#file-gistfile1-txt-L3391
  * https://gist.github.com/johnkerl/45747955555e1cb8ac1e5670b9e8f819#file-gistfile1-txt-L3149-L3391

```
2024-03-11T21:24:54.7197580Z   There are binary versions available but the source versions are later:
2024-03-11T21:24:54.7219220Z              binary   source needs_compilation
2024-03-11T21:24:54.7219850Z digest       0.6.34   0.6.35              TRUE
2024-03-11T21:24:54.7220420Z tidyselect    1.2.0    1.2.1              TRUE
2024-03-11T21:24:54.7221570Z arrow      14.0.0.2 14.0.2.1              TRUE
...
2024-03-11T21:26:09.0335010Z *** Building libarrow from source
```

Artifacts:

 * https://cran.r-project.org/package=arrow
 * Shows clearly: macOS binaries:	r-release (arm64): arrow_14.0.0.2.tgz, r-oldrel (arm64): arrow_14.0.0.2.tgz, r-release (x86_64): arrow_14.0.0.2.tgz
* Meanwhile source 14.0.2.1 exists
* In summary: the source version is newer than the latest version for which MacOS binaries are provided

## Solution

TBD. Maybe we need to work with the Arrow folks to provide current MacOS binaries? Or somehow install a pinned version for which binaries _are_ available?

mojaveazure:
Force binary-only installation for macOS. By setting as an option in `~/.Rprofile`; this should affect both package installation and updating through `utils::install.packages()`, `utils::update.packages()`, and `remotes::install_deps()`

resolves #2254

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -16,7 +16,7 @@ env:
   COVERAGE_TOKEN: ${{ secrets.CODECOV_TOKEN }}
   _R_CHECK_TESTS_NLINES_: 0
   CATCHSEGV: "TRUE"
-  
+
 jobs:
   ci:
     strategy:
@@ -51,6 +51,11 @@ jobs:
 
       - name: Bootstrap
         run: cd apis/r && tools/r-ci.sh bootstrap
+
+      - name: Set pkgType to binary (macOS)
+        if: ${{ matrix.os == 'macOS-latest' }}
+        run: cat("\noptions(pkgType = 'binary')\n", file = "~/.Rprofile", append = TRUE)
+        shell: Rscript {0}
 
       - name: Install BioConductor package SingleCellExperiment
         run: cd apis/r && tools/r-ci.sh install_bioc SingleCellExperiment
@@ -116,4 +121,3 @@ jobs:
       - name: View Logs
         run: apis/r/tools/r-ci.sh dump_logs
         if: failure()
-
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -82,6 +82,10 @@ jobs:
       - name: Dependencies
         run: cd apis/r && tools/r-ci.sh install_all
 
+      - name: Install dataset packages from source (macOS)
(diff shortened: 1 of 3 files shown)


mojaveazure:
Note:

As the two pbmc3k dataset packages are source-only, I've added a new stanza to install these separately on macOS. This calls a separate R script `apis/r/tools/install_missing_deps.R` which
1. finds dependencies of tiledbsoma that were not installed due to the binary-only installation earlier
2. find the upstream dependencies of the missing packages that are source-only (because Bioconductor)

This stanza runs on macOS only

johnkerl (approved):
LGTM to me from the release-management point of view!! :) CI is green and I am happy

I defer to @eddelbuettel on the implementation details ...

eddelbuettel:
I'll stay out of it.  I provided a much simpler code example to @mojaveazure that attacked a simpler sub-task and that is what I might have done, but I didn't and I won't rewrite this. If it works it works.

PR #2259, opened 2024-03-13
[c++] Match `delete[]` with `new[]`

johnkerl:
As discussed in Slack.

### DIFF
--- a/libtiledbsoma/src/utils/arrow_adapter.cc
+++ b/libtiledbsoma/src/utils/arrow_adapter.cc
@@ -83,7 +83,7 @@ void ArrowAdapter::release_array(struct ArrowArray* array) {
     delete arrow_buffer;
 
     if (array->buffers != nullptr) {
-        free(array->buffers);
+        delete [] array->buffers;
     }
 
     struct ArrowArray* dict = array->dictionary;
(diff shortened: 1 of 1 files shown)


nguyenv (approved):

johnkerl:
@nguyenv would you prefer we use `malloc` and `free` rather than `new[]` and `delete[]`?

johnkerl:
@nguyenv hmm ... really a toss-up IMHO ... I do notice:

* There's a lot of other `new` and `delete` in that file
* No other `malloc` or `free`
* As long as C++ news memory, hands it off to C which uses it, then deletes that memory, I think all is well ...

eddelbuettel:
The line to allocate these buffers should be this one

https://github.com/single-cell-data/TileDB-SOMA/blob/4a71f19f20c3b6ad4774942b0211df50087adb1b/libtiledbsoma/src/utils/arrow_adapter.cc#L288

so it is indeed a `new` that should be a matched by a `delete`.   Or ... as there is a `free` the allocations could a `malloc`.  But point taken that it is currently mismatched.

johnkerl:
@eddelbuettel that is correct, lines 288 and 86 as in Slack.

nguyenv:
I think this PR is fine as is to fix for the upcoming 2.21.0 release. Then it can be changed to malloc/free in a future PR since that's not as high priority with everything now matching correctly.

PR #2275, opened 2024-03-14
[r/ci] Point at correct TileDB-R location

johnkerl:
See comments in the affected CI YAML file.

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -60,10 +60,18 @@ jobs:
       - name: Install BioConductor package SingleCellExperiment
         run: cd apis/r && tools/r-ci.sh install_bioc SingleCellExperiment
 
+      # The next two stanzas are necessary given propagation delay for binaries at
+      # https://r2u.stat.illinois.edu/ which may lack binaries (for short periods of usually a day)
+      # when sources have been updated. The default installation could switch to installation from
+      # sources --- but that would require installing all build dependencies. To see what the most
+      # recent binary is run e.g.  docker run --rm -ti rocker/r2u bash -c 'apt update -qq &&
+      # apt-cache show r-cran-tiledb' Using the r-universe builds (as below) is a suitable fallback
+      # as they update more frequently than CRAN.
+      #
       # Uncomment these next two stanzas as needed whenever we've just released a new tiledb-r for
       # which source is available but CRAN releases (and hence update r2u binaries) are not yet:
       #
-      # IMPORTANT: these two stanzas should remain uncomment _only_ during the propagation time
+      # IMPORTANT: these two stanzas should remain uncommented _only_ during the propagation time
       # between (a) publication of source for a new TileDB-R _upon which_ TileDB-SOMA depends in its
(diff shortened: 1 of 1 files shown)


johnkerl:
[sc-43172]

.github/workflows/r-ci.yml
```diff
@@ -60,16 +60,28 @@ jobs:
       - name: Install BioConductor package SingleCellExperiment
         run: cd apis/r && tools/r-ci.sh install_bioc SingleCellExperiment
 
+      # The next two stanzas are necessary given propagation delay for binaries at
+      # https://r2u.stat.illinois.edu/ which may lack binaries (for short periods of usually a day)
+      # when sources have been updated. The default installation could switch to installation from
+      # sources --- but that would require installing all build dependencies. To see what the most
+      # recent binary is run e.g.  docker run --rm -ti rocker/r2u bash -c 'apt update -qq &&
+      # apt-cache show r-cran-tiledb' Using the r-universe builds (as below) is a suitable fallback
+      # as they update more frequently than CRAN.
```
eddelbuettel, line 69: I am not really sure I agree with the wording.  Sadly I cannot quite put my finger on something better.  

But riffing for a moment we seem to have the three situations here:
- tiledb-r and tiledbsoma agree on versions, the world is good
- tiledb-r is
  - at release at CRAN (which is what we generally want)
  - at release-candidate in the GH repo and available via selected r-universe builds 
 
That last bit is "unusual but common" and I think the bit we have here today.  

And just how he have / had the existing ("sleeping") PR #1994 we have this now.   Part of me thinks we may want to discuss how to drive this better from one top-level (possibly outside the repo?) 'declarative' file. 

It's tricky.

eddelbuettel (commented):
Can you remove the new leading paragraph?  I do not think it illuminates sufficiently.

What we have here is that sometimes we need jump a step ahead and sometimes we need to not do that.  So we comment these stanza 'in' and 'out' as they give us the binaries we otherwise want to speed things up.

johnkerl:
> Can you remove the new leading paragraph? I do not think it illuminates sufficiently.

I put it in a while back because I found it to be necessary information for me. And I do release management for this project -- so I believed that was important.

Then, on another commit (not by me), that paragraph disappeared.

Then, on this commit I'm trying to put it back in.

Here, you're asking me to take it back out again.

What I will do is to take the wording I find crucial -- to avoid CI breakages like the one we're having right now -- and I will put that wording at https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases, and I will add that hyperlink as a comment within the `r-ci.yaml` file on this PR, and I will ask you to not remove this crucial hyperlink on future commits.

mojaveazure (approved):

eddelbuettel:
@johnkerl My comments in this ticket were aimed at illustrating the complexity of the issue. The comment in the yaml did not reflect it to my reading, but focussed on a different (and here less relevant)  issue.  No more no less.   It is by our chosing that we sometimes want what r-universe offers, and sometimes we don't.  I maintain that a more central switch may help us all do our job and more easily.

PR #2283, opened 2024-03-15
[python] Fix an ingestion corner case with empty chunks

johnkerl:
**Issue and/or context:** This is from a non-GitHub bug report.

**Changes:** Avoid division by zero on entirely empty chunks.

**Notes for Reviewer:** I tried to create synthetic data that would trigger the corner case, but failed to repro. Had I succeeded, I'd have created a new unit-test case on this PR. As is, I was only able to repro with some larger data I'm not willing to check in. Regardless, the change has been tested successfully against that dataset, and furthermore, it's a clear and solid code mod.


johnkerl:
[sc-42637]

Shelnutt2 (approved):

PR #2290, opened 2024-03-19
[python] Extract fs wrapper to util file

johnkerl:
**Issue and/or context:** This will allow us to use append mode with multiple `.h5ad` files stored on S3.

**Changes:** Extract the fs-wrapper function to a util location, and call it everywhere we were using `anndata.read_h5ad` before.

**Notes for Reviewer:**


ryan-williams (approved):
lgtm, will follow on with #2291 (contextmanager version)

PR #2294, opened 2024-03-21
Use TileDB core 2.21.0, TileDB-Py 0.27.0, TileDB-R 0.25.0

johnkerl:
[sc-43172]

Procedure:
https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases#incrementing-core-versions


eddelbuettel (approved):
Looks good as a far as I can tell.  It might be easier to use a single source of truth in the repo, or switch to accessing the now-existing [single source of truth](https://github.com/TileDB-Inc/TileDB/releases/download/2.21.0/releases.csv).

Issue #1988, opened 2023-12-15
[python/r] DataFrame: value filter on enum/dict column generates internal error when sought value not in enumeration


bkmartinjr:
I have an empty dataframe containing dictionary/enum attributes.  When a value filter / query condition is applied to it, it triggers an internal Arrow error.  It should return an empty result.  All works fine for non-dictionary attributes, so it appears that value filters do not always work correctly with dict/enum attributes.

Note that the `tiledb` package also has questionable behavior here, returning an exception if the value filter attempts to test for a value not in the enumeration.  So it is likely that the Arrow error is unique to the libtiledbsoma codepath, but both behaviors make the combination of filters and enums problematic.

What I think should happen:  the value filter should have identical behavior (ie., results) for a column of type "T" and a column of type "enum-of-T", where T is string, int, etc (e.g., a query against a "dict of strings" column should perform the same as a query against a string column). 

\<late edit\>
The empty dataframe is unrelated.  It fails in exactly the same way for non-empty arrays.  I'll add an example of that below.
\</late edit\>

The schema (abbreviated for ease of reading):
 
``` 
In [102]: obs.schema
Out[102]: 
soma_joinid: int64
dataset_id: dictionary<values=string, indices=int8, ordered=0>
is_primary_data: bool
observation_joinid: large_string
# lots of other columns removed for brevity
```

Reading the entire thing works correctly (output abbreviated):
```
In [103]: obs.read().concat()
Out[103]: 
pyarrow.Table
soma_joinid: int64
dataset_id: dictionary<values=string, indices=int8, ordered=0>
is_primary_data: bool
observation_joinid: large_string
----
soma_joinid: [[]]
dataset_id: [  -- dictionary:
[]  -- indices:
[]]
assay: [  -- dictionary:
[]  -- indices:
[]]
...
```

Read with a value filter on a string attribute works fine (output abbreviated):

```
In [104]: obs.read(value_filter="observation_joinid == 'foobar'").concat()
Out[104]: 
pyarrow.Table
soma_joinid: int64
dataset_id: dictionary<values=string, indices=int8, ordered=0>
is_primary_data: bool
observation_joinid: large_string
----
soma_joinid: [[]]
dataset_id: [  -- dictionary:
[]  -- indices:
[]]
...
```

Reading with a value filter *on a dict column* fails an internal Arrow error check:
```
In [105]: obs.read(value_filter="""dataset_id == 'foobar'""").concat()
[...]

PR #2299, opened 2024-03-21
Factor `pytest.fixture`s, verify #1988

ryan-williams:
**Issue and/or context:** #1988

**Changes:**
- Add test to verify #1988
- Factor `pytest.fixture`s, misc test cleanup

**Notes for Reviewer:**
- Test cleanups can be factored out to another PR, if desired.
- I verified that the new test, [`test_empty_categorical_query`](https://github.com/single-cell-data/TileDB-SOMA/pull/2299/commits/6dbb9a0084a03f4b3c390fb51b226b966adb6909#diff-356b52ccf7c4b2025a8d80701d49afb9ee699e70f42e5e54f3357a8cbb29bba3R910), raises `ArrowInvalid` with TileDB 2.20.1, but passes here with 2.21.0.

### DIFF
--- a/apis/python/src/tiledbsoma/_read_iters.py
+++ b/apis/python/src/tiledbsoma/_read_iters.py
@@ -13,7 +13,6 @@ from concurrent.futures import ThreadPoolExecutor
 from typing import (
     TYPE_CHECKING,
     Dict,
-    Generator,
     Iterator,
     List,
     Optional,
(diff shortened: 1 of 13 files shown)


johnkerl (approved):
This is epic -- thank you!

I didn't set an expectation ahead of time, but for the future: let's do refactor PRs and behavior-addition PRs as separate. That makes them easier to review. (And easier to deal with in case of reverts, which admittedly don't often happen.)

PR #2307, opened 2024-03-22
skip `test_empty_categorical_query`

ryan-williams:
Skip test added in #2299 which requires core 2.21.0, and is currently failing builds `release-1.8` builds ([example](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/8392617173/job/22989959668)).

**Issue and/or context:** #1988, #2299


johnkerl:
Replaced by #2308

Issue #2312, opened 2024-03-25
[python/ci] Issues with typeguard 4.2


johnkerl:
We've been at 4.1.5 since last week. Meanwhile typeguard 4.2 was just released and we are seeing in CI

```
TypeError: issubclass() argument 2 cannot be a parameterized generic
```

@ryan-williams will either fix for typeguard 4.2, or, pin our repo to typeguard 4.1.5.

ryan-williams:
#2313 pinned 4.1.5

Looking into getting 4.2.x working

PR #2314, opened 2024-03-25
[python/ci] `typeguard==4.2.1`, make `requirements_dev.txt` canonical

ryan-williams:
**Issue and/or context:** #2312

**Changes:**
- use requirements_dev.txt in CI, setup.py: https://github.com/single-cell-data/TileDB-SOMA/commit/c47fa2a03b97c2b1550956f114556953374c5020
- use typeguard>=4.2, fixup some type checks: https://github.com/single-cell-data/TileDB-SOMA/pull/2314/commits/74c530b77c0b630b1540b9a662ebc4b20eb83733

**Notes for Reviewer:**
`requirements_dev.txt` seemed to be unused. I:
- populated `requirements_dev.txt` with the actual `[dev]` reqs from `setup.py` (plus a few that were only listed in CIs: `pytest-cov`, `sparse`)
- made `setup.py` use the new reqs file
- made CIs install `-e apis/python[dev]` (they were previously installing sans `[dev]`, and pre-installing a few select dev reqs)

Previously the typeguard pin existed in 4 places, for no good reason (afaict).

Let me know if there were actually two distinct sets of "dev" reqs here, which I am now conflating.

### DIFF
--- a/apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
+++ b/apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
@@ -8,7 +8,7 @@ import functools
 import threading
 import time
 from concurrent.futures import ThreadPoolExecutor
-from typing import Any, Dict, Mapping, Optional, Type, Union, cast
+from typing import Any, Dict, Literal, Mapping, Optional, Union
 
 import tiledb
 from somacore import ContextBase
@@ -46,10 +46,8 @@ def _maybe_timestamp_ms(input: Optional[OpenTimestamp]) -> Optional[int]:
     return to_timestamp_ms(input)
 
 
-class _SENTINEL:
-    """Sentinel object to distinguish default parameters from None."""
-
-    pass
+_Unset = Literal["__unset__"]
+_UNSET: _Unset = "__unset__"
 
 
 class SOMATileDBContext(ContextBase):
@@ -221,8 +219,8 @@ class SOMATileDBContext(ContextBase):
         *,
         tiledb_config: Optional[Dict[str, Any]] = None,
         tiledb_ctx: Optional[tiledb.Ctx] = None,
-        timestamp: Union[None, OpenTimestamp, Type[_SENTINEL]] = _SENTINEL,
-        threadpool: Union[None, ThreadPoolExecutor, Type[_SENTINEL]] = _SENTINEL,
+        timestamp: Union[None, OpenTimestamp, _Unset] = _UNSET,
+        threadpool: Union[None, ThreadPoolExecutor, _Unset] = _UNSET,
     ) -> Self:
         """Create a copy of the context, merging changes.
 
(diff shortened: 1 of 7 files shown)


johnkerl (approved):
🚀

ryan-williams:
@thetorpedodog, you had good feedback on #1960, wanted to give you an opportunity to comment here, in case anything jumps out.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -259,17 +261,19 @@ def replace(
...
                 # Keep the existing threadpool if not overridden.
                 threadpool = self.threadpool
+
         return type(self)(
             tiledb_config=tiledb_config,
             tiledb_ctx=tiledb_ctx,
-            timestamp=timestamp,
-            threadpool=threadpool,
+            timestamp=cast(Optional[OpenTimestamp], timestamp),
+            threadpool=cast(Optional[ThreadPoolExecutor], threadpool),
```
ryan-williams, line 276: Surprised I had to resort to `cast`ing here. I thought mypy would understand that the `if`s on L265-L270 above should remove `SENTINEL` from the `timestamp` and `threadpool` Unions, but mypy was erroring.

Previously, the `# type: ignore[assignment]`s on L224-L225 were working around the issue. Localizing the issue here seems cleaner, but open to ideas.

thetorpedodog: It can't infer that `_SENTINEL` is the only instance of a Type that we will accept; i.e., it thinks that there might be some other type, like `class MyBogusSentinel(_soma_tiledb_context.SENTINEL): pass`. This would mean that a user could pass `ctx.replace(timestamp=MyBogusSentinel)`, and doing so is valid within the constraints of the type system.

To do this without requiring changes to the type-checker would require the `if whatever is _SENTINEL` lines to be changed to `if isinstance(whatever, type)` (since `type` is the type of the `_SENTINEL` class-value).

I think there are a few reasonable solutions here:

 1. Change the if-statement to check whether the value in question is a type (since `_SENTINEL` is a class rather than an instance of the class):

    ```python
    def some_func(
        very_optional_param: Union[None, int, Type[_SENTINEL]] = _SENTINEL,
    ) -> ...:
      if isinstance(very_optional_param, type):
        very_optional_param = int_from_somewhere()
      # All types are gone from possible values of very_optional_param
      # leaving it as an Optional[int]
    ```

    (You can’t do `issubclass(very_optional_param, _SENTINEL)` because the first parameter to `issubclass` must be a class itself.)

 2. Use an _instance_ of a sentinel type as a default value, rather than the type itself. This might look like:

    ```python
    # this is an empty attrs class so it is immutable
    @attrs.define()
    class _Unset:
      pass

    def some_func(
        very_optional_param: Union[None, int, _Unset] = _Unset(),
    ) -> ...:
      if isinstance(very_optional_param, _Unset):
        very_optional_param = int_from_somewhere()
      # now very_optional_param is definitely an Optional[int]
      # since all possible _Sentinels are gone
      ...
    ```

 3. A simpler solution for this particular case: Make the default value a simple literal that doesn’t belong in any of the other types.

    ```python
    # My thinking behind defining these literals and constants:
    # We can document that people shouldn't be relying on
    # the exact value "__unset__" if they use the API.
    _Unset = Literal["__unset__"]
    _UNSET: _Unset = "__unset__"

    def some_func(
        very_optional_param: Union[None, int, _Unset] = _UNSET,
    ) -> ...:
      if very_optional_param == _UNSET:
        very_optional_param = int_from_somewhere()
      # "__unset__" is also now gone from the possible values
      # of very_optional_param.
    ```

I actually like the latter, since it means we just use types and instances that are already there, doesn’t have the weirdness of `Type[_SENTINEL]`, and we can always upgrade it to the immutable-class-instance–based system by changing internal code without affecting the appearance of the signature.

It would be neat if `Literal` accepted just a raw object:

```python
_SENTINEL = object()
_LiteralSentinel = Literal[_SENTINEL]
```

to mean that you could only pass the literal object `my_module._SENTINEL` the function, but it doesn’t (for sensible reasons). But if it did, we could avoid all this!

thetorpedodog (approved):
Up to you how you want to handle the unset sentinel thing; I have three possible ideas below and you can choose if you like one.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
ryan-williams: Thanks @thetorpedodog, great explanation!

I've updated this to use `_Unset`/`_UNSET`, per your example. Typeguard and mypy seem happy with it, and it seems cleaner than my previous `cast`-based workaround.

Issue #2311, opened 2024-03-25
TileDB-SOMA 1.9 release


johnkerl:
Tracks things needing to be done to get there.

* Blockers: https://github.com/single-cell-data/TileDB-SOMA/issues?q=is%3Aissue+is%3Aopen+label%3Ablocks-1.9
* This will cause the [Conda pre-check PR](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases) to fail: https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/108#issuecomment-2018088174

[[sc-43172]](https://app.shortcut.com/tiledb-inc/story/43172)
[[sc-43511]](https://app.shortcut.com/tiledb-inc/story/43511)
[[sc-43673]](https://app.shortcut.com/tiledb-inc/story/43673)

jdblischak:
> This will cause the [Conda pre-check PR](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases) to fail: https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/108#issuecomment-2018088174

This conda solver error was specific to tiledb 2.20. The recently successful https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/110 confirmed that it did not affect tiledb 2.21

Also, I fixed the issue with tiledb 2.20, so the nightlies are working again (https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/108#issuecomment-2020565393)

PR #2342, opened 2024-03-27
[r] tiledbsoma-r 1.9.1

johnkerl:

We need a 1.9.1 for our upcoming Python-only 1.9.1

#2311
[sc-43673]

### DIFF
--- a/apis/r/NEWS.md
+++ b/apis/r/NEWS.md
@@ -1,3 +1,7 @@
+# 1.9.1
+
+* This release contains a single Python-only mod
+
 # 1.9.0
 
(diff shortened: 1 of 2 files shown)


apis/r/NEWS.md
```diff
@@ -1,3 +1,7 @@
+# 1.9.1
+
+* This release contains a single Python-only mod
```
eddelbuettel, line 3: It's fine but a few added words such 'Python-only modification for its build process' are more informative

eddelbuettel (approved):

PR #2350, opened 2024-03-29
[c++/python] Add `soma_type` argument for `SOMAObject::open`

nguyenv:
**Issue and/or context:**

`SOMAObject::open` introduced a performance degradation due to its overzealous usage of `tiledb::Object::object`, which checks for the type by opening the purpoted TileDB object at the given URI. In some cases, the type of the TileDB object is already known beforehand which makes these operations unnecessary.

The example below takes ~15 seconds to run and yields three 404 errors.
```
SOMA_URI = "tiledb://TileDB-Inc/soma-exp-tabula-sapiens-immune"
var_columns = ["soma_joinid", "var_id", "ensemblid", "feature_type"]

with tiledbsoma.open(SOMA_URI, context=sctx) as exp:
    var = (
        exp.ms["RNA"]
        .var.read(column_names=var_columns)
        .concat()
        .to_pandas()
        .set_index("soma_joinid")
    )
```

With the changes in this PR, the example takes ~8 seconds and yields no 404 errors.

**Changes:**

Bypass the need to use `tiledb::Object::object` by explcitly passing in a `soma_type` argument of either `SOMAArray` or `SOMAGroup`.

### DIFF
--- a/apis/python/src/tiledbsoma/_collection.py
+++ b/apis/python/src/tiledbsoma/_collection.py
@@ -143,7 +143,7 @@ class CollectionBase(  # type: ignore[misc]  # __eq__ false positive
             tiledb_timestamp=tiledb_timestamp,
             context=context,
             platform_config=platform_config,
-            soma_type="SOMAGroup",
+            clib_type="SOMAGroup",
         )
 
     # Subclass protocol to constrain which SOMA objects types  may be set on a
@@ -457,12 +457,8 @@ class CollectionBase(  # type: ignore[misc]  # __eq__ false positive
             timestamp = self.tiledb_timestamp_ms
 
             try:
-                soma_type = None
-                if entry.entry.wrapper_type == _tdb_handles.ArrayWrapper:
-                    soma_type = "SOMAArray"
-                elif entry.entry.wrapper_type == _tdb_handles.GroupWrapper:
-                    soma_type = "SOMAGroup"
-                wrapper = _tdb_handles.open(uri, mode, context, timestamp, soma_type)
+                clib_type = entry.entry.wrapper_type.clib_type
+                wrapper = _tdb_handles.open(uri, mode, context, timestamp, clib_type)
                 entry.soma = _factory.reify_handle(wrapper)
             except SOMAError:
                 entry.soma = _factory._open_internal(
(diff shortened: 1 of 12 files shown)


aaronwolen:
[sc-44157]

johnkerl (dismissed):
🚢

johnkerl:
@aaronwolen @nguyenv in order to take full advantage of this PR we should be using
```
with tiledbsoma.Experiment.open(SOMA_URI, context=sctx) as exp:
```
at the notebook level -- making `tiledbsoma.open` more performant will require core mods this PR cannot implement.

apis/python/src/tiledbsoma/_collection.py
```diff
@@ -436,7 +436,12 @@ def __getitem__(self, key: str) -> CollectionElementType:
             timestamp = self.tiledb_timestamp_ms
 
             try:
-                wrapper = _tdb_handles.open(uri, mode, context, timestamp)
+                soma_type = None
+                if entry.entry.wrapper_type == _tdb_handles.ArrayWrapper:
+                    soma_type = "SOMAArray"
+                elif entry.entry.wrapper_type == _tdb_handles.GroupWrapper:
+                    soma_type = "SOMAGroup"
```
thetorpedodog, line 443: These seem like they could be members of the `_tdb_handles.[whatever]Wrapper` classes, eliminating the need for this test.

Would `storage_type` or `tiledb_type` be a better name for this? I would expect `soma_type` to refer to the name of the type of SOMA object that is being opened (like `SOMADataFrame`).

apis/python/src/tiledbsoma/soma_object.cc
```diff
@@ -55,21 +55,23 @@ void load_soma_object(py::module& m) {
...
-               std::optional<std::pair<uint64_t, uint64_t>> timestamp)
-                -> py::object {
+               std::optional<std::pair<uint64_t, uint64_t>> timestamp,
+               std::optional<std::string> soma_type) -> py::object {
                 try {
-                    auto obj = SOMAObject::open(uri, mode, context, timestamp);
-                    if (obj->type() == "SOMADataFrame")
+                    auto obj = SOMAObject::open(
+                        uri, mode, context, timestamp, soma_type);
+                    auto obj_type = obj->type();
```
thetorpedodog, line 63: this variable might be better named `soma_type`

johnkerl (changes requested):
There is a missed callsite -- more details in Slack

The root cause is `Experiment.open` does not have its own impl and neither does `CollectionBase.open` and so therefore `tiledbsoma.Experiment.open`'s impl is `TileDBObject.open` (along the inheritance hierarchy) and there is not a `soma_type` plumbed through to there

johnkerl (commented):
🚢 
Thanks @nguyenv !!!! :)

johnkerl (approved):

PR #2357, opened 2024-04-01
[python] Skip writing empty `arrow_table`s

ryan-williams:
**Issue and/or context:** [sc-43541]; all-`nan` column of `string` dtype cause `RuntimeError` when appending an AnnData with the same `obs` IDs as already exist (i.e. the append would no-op / no new rows are added).

**Changes:** skip writing `arrow_table` when it is empty

**Notes for Reviewer:**
- Why is the issue specific to all-`nan`, `string` dtype columns?
- Are there any other side effects related to opening and/or writing that should be accounted for?


johnkerl (approved):
🚢

PR #2360, opened 2024-04-02
[python] `tiledbsoma.io.from_h5ad`: document new `additional_metadata` param

### LINKED ISSUE #2359: [python] Add docstring for `additional_metadata`
#2258 is missing docstring content

ryan-williams:
**Issue and/or context:** fixes #2359

Documentation follow-up to #2258

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -311,6 +311,14 @@ def from_h5ad(
           The default is to ingest them all. Use ``uns_keys=[]``
           to not ingest any ``uns`` keys.
 
+        additional_metadata: Optional metadata to add to the ``Experiment`` and all descendents.
+          This is a coarse-grained mechanism for setting key-value pairs on all SOMA objects in an
+          ``Experiment`` hierarchy. Metadata for particular objects is more commonly set like:
+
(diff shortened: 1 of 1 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -311,6 +311,8 @@ def from_h5ad(
           The default is to ingest them all. Use ``uns_keys=[]``
           to not ingest any ``uns`` keys.
 
+        additional_metadata: Optional metadata to add to the ``Experiment`` and all descendents.
```
johnkerl, line 314: Let's emphasize that this is a blanket same-for-all convenience. And tell them that if they want to set metadata on individual elements they can do so using the SOMA API

```
with soma.open(uri, 'w') as exp:
    exp.metadata.update({"foo":"bar"})
    exp.obs.metadata.update({"foo":"whatever"})
    ...
```

ryan-williams: Good idea, lmk how the latest version looks.

jp-dark (approved):

johnkerl (approved):
🚢 
LGTM; unless @jp-dark has any other thoughts! :)

PR #2364, opened 2024-04-03
[python] More `nan_append` checks: not-all-NaN, non-dupe IDs

ryan-williams:
**Issue and/or context:** [sc-43541](https://app.shortcut.com/tiledb-inc/story/43541/failed-to-convert-buffer-for-attribute-on-ingest), extends #2357

**Changes:** sweep over more configurations when testing AnnData appends with empty columns:
- not all values `nan`
- not all `obs` IDs duplicated between initial and appended AnnData

Per offline discussion with @aaronwolen and @johnkerl.

### DIFF
--- a/apis/python/tests/test_basic_anndata_io.py
+++ b/apis/python/tests/test_basic_anndata_io.py
@@ -1258,13 +1258,11 @@ def test_outgest_X_layers(tmp_path):
         assert sorted(list(bdata.layers.keys())) == ["data2", "data3"]
 
 
-@pytest.mark.parametrize("dtype", ["float64", "string"])  # new column dtype
-@pytest.mark.parametrize(
-    "nans", ["all", "none", "some"]
-)  # how many `nan`s in new column?
-@pytest.mark.parametrize(
-    "new_obs_ids", ["all", "none", "half"]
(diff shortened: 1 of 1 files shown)


ryan-williams:
Here's a sweep I just ran locally, with my fix from #2357 disabled, and more test cases than are included in this PR (https://github.com/single-cell-data/TileDB-SOMA/commit/4dad5808097d8e34b554ee8cc9bd4ddd29d9c3ff):

image


The two failures occur with parameters:

1. "none" (no new obs IDs present)
    - "all" and "half" always pass, i.e. the failure was specific to when the `obs` append is effectively empty
2. "all" or "some" elems of added column are `nan`
    - "none" always passes, i.e. the failure requires at least 1 nan in the column being added
3. "string" dtype on the added column
    - "float64" always passes; failure requires adding a `string` column containing at least one `nan`)

Not sure it's necessary to add this larger complement of test cases to trunk, just pinpointing the original error conditions a bit more precisely…

aaronwolen (commented):
Thank you!

aaronwolen (approved):

johnkerl (approved):

PR #2366, opened 2024-04-03
[c++] Set `nanoarrow.c` to use `CMAKE_CXX_COMPILER`

nguyenv:
By default, file extensions ending in .c use `CMAKE_C_COMPILER`. When compiling `nanoarrow.c` with gcc, this yields the warning `command-line option '-fvisibility-inlines-hidden' is valid for C++/ObjC++ but not for C` which causes the nightly feedstock to fail as we promote all warnings to errors.


jdblischak (approved):

eddelbuettel:
So am I summarizing it correctly as 
- compiling `nanoarrow.c` as C++ helps with a visibility issue (maybe only on Conda?) that is created by using  -Werror
- but creates a side effect for R only of hiding three symbols which we then 'unvendor' as a fix

Is that what we did?

nguyenv:
> So am I summarizing it correctly as
> 
> * compiling `nanoarrow.c` as C++ helps with a visibility issue (maybe only on Conda?) that is created by using  `-Werror`
> * but creates a side effect for R only of hiding three symbols which we then 'unvendor' as a fix
> 
> Is that what we did?

Yes that is correct.

eddelbuettel (approved):

eddelbuettel:
Yep, all good.  That is a lesser evil than turning off `-Werror`.

PR #2367, opened 2024-04-03
[NOMERGE] [c++/python] Move `DataFrame` write path to use C++ bindings

nguyenv:
This will be broken up into multiple smaller PRs.


nguyenv:
This PR has now been fullly merged in as #2418, #2425, and #2431 and can be closed.

PR #2372, opened 2024-04-03
[r] Add `$reopen()` to `TileDBObject`

mojaveazure:
Expose `$reopen()` as a public API for all `TileDBObjects` (groups and arrays); as R does not have context managers, there's no native way to do the Python construct
```python
cls = type(arr)
with cls.open(arr.uri, "r") as readarr:
    # do a read operation even though the array is open in write
```
This PR exposes `$reopen()` to reopen an array or collection in a new mode

Modified SOMA methods:

 - `TileDBObject$reopen()`: pushed down from `TileDBArray` and now public; includes new default for reopening: if `arr$mode()` is `READ`, will automatically reopen in `WRITE` and vice versa
 - `TileDBObject$is_open()`: now uses `TileDBObject$mode()` to determine if the array is open as `TileDBObject$mode()` already processes `TileDBObject$private$.mode`


apis/r/R/TileDBArray.R
```diff
@@ -49,6 +49,29 @@ TileDBArray <- R6::R6Class(
...
+    #'  \item \dQuote{\code{WRITE}}
+    #' }
+    #' By default, reopens in the opposite mode of the current mode
+    #'
+    #' @return Invisibly returns \code{self}
+    #'
+    reopen = function(mode = NULL) {
+      modes <- c(READ = 'WRITE', WRITE = 'READ')
+      oldmode <- self$mode()
+      mode <- mode %||% modes[oldmode]
```
eddelbuettel, line 66: The flipping read/write is clever but maybe too hard too read?   

Also what happens if mode were "RESUME" ?

apis/r/R/TileDBArray.R
```diff
@@ -289,17 +312,17 @@ TileDBArray <- R6::R6Class(
...
+    # reopen = function(mode = c("READ", "WRITE")) {
+    #   mode <- match.arg(mode)
+    #   if (private$.mode == mode) {
+    #     return(invisible(self))
+    #   }
+    #   self$close()
+    #   invisible(
+    #     self$open(mode = mode, internal_use_only = "allowed_use")
+    #   )
+    # },
```
eddelbuettel, line 325: We tend to remove commented-out code.

apis/r/R/TileDBArray.R
johnkerl: Per conversation with @mojaveazure `mode` and `ingest_mode` are distinct in Python, and must be in R as well.

mojaveazure: For `$open()` (and therefore `$reopen()`); mode cannot be `"RESUME"`

On the Puython side, resume-mode exists only in `tiledbsoma.io`; for parity, this means that resume-mode exists only in the `write_soma()` methods and is handled there, but the mode of the array itself cannot be `"RESUME"`

apis/r/R/TileDBArray.R
mojaveazure: 🤦 One day, I'll learn to read-through my PRs and remove unused code. Thanks for catching!

eddelbuettel (approved):
That will be good to have exposed.

johnkerl:
@mlin @bkmartinjr @pablo-gar I'd love to get your feedback

johnkerl (approved):
🔥

mojaveazure:
[sc-48522]

PR #2373, opened 2024-04-03
[r/ci] Controlled downgrade for TileDB-R 0.25

johnkerl:
**Issue and/or context:** Applies the same pattern on the `release-1.9` branch as #1972 did for the `release-1.5` branch, #1996 did for the `release-1.6` branch, #2157 did for the `release-1.7` branch, and #2295 did for the `release-1.8` branch.

Standard procedure:
https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases#incrementing-core-versions

**Changes:** As on #1972, #1996, #2157, and #2295. This is necessary since [TileDB-R 0.26.0](https://tiledb-inc.r-universe.dev/tiledb) will soon exist, and the R language's `DESCRIPTION` file does not syntactically support `tiledb < 0.26` or `tiledb <= 0.26`. Therefore (see also this comment https://github.com/single-cell-data/TileDB-SOMA/pull/1996#issuecomment-1881702859) we do a controlled downgrade to TileDB-R 0.25 on this PR, for the `release-1.9` branch.

**Notes for Reviewer:** Every stanza in this PR has been proved necessary via the dialog on #1972, #1996, #2157, and #2295. See in particular https://github.com/single-cell-data/TileDB-SOMA/pull/1996#pullrequestreview-1809807165.

PR #2157, opened 2024-02-20
[r/ci] Controlled downgrade for TileDB-R 0.23

johnkerl:
**Issue and/or context:** Applies the same pattern as #1972 did for the `release-1.5` branch and `#1996` did for the `release-1.6` branch.

**Changes:** As on #1972 and #1996. This is necessary since [TileDB-R 0.24.0](https://tiledb-inc.r-universe.dev/tiledb) now exists, and the R `DESCRIPTION` file does not syntactically support `tiledb < 0.24` or `tiledb <= 0.24`. Therefore (see also this comment https://github.com/single-cell-data/TileDB-SOMA/pull/1996#issuecomment-1881702859) we do a controlled downgrade to TileDB-R 0.23 on this PR, for the `release-1.7` branch.

**Notes for Reviewer:**

Every stanza in this PR has been proved necessary via the dialog on #1972 and #1996. See in particular https://github.com/single-cell-data/TileDB-SOMA/pull/1996#pullrequestreview-1809807165.

PR #2374, opened 2024-04-03
[r] Experiment with syntax for lower and upper pin bounds on `tiledb-r`

johnkerl:
**Issue and/or context:** Per https://github.com/single-cell-data/TileDB-SOMA/pull/2373#issuecomment-2035526682 following advice https://stackoverflow.com/questions/63834412/enforce-upper-bound-for-r-package-version-dependency which is at variance with https://github.com/rstudio/renv/issues/764 and, in fact, everything I've ever been told. I'm currently uncertain whether the bidirectional-pin syntax is genuinely new.

**Changes:** Try it out!

**Notes for Reviewer:** This PR is not ready for review. I will take this PR out of draft status as such time as it becomes ready  for review.

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -105,7 +105,7 @@ jobs:
       - name: Update Packages
         run: Rscript -e 'update.packages(ask=FALSE)'
 
-      # WIP (NOT READY FOR REVIEW) -- LOOKING FOR GREEN CI
+      # NOT SURE THIS IS CORRECT YET -- LOOKING FOR GREEN CI
       #- name: Install 0.25 build of tiledb-r with core 2.21
       #  run: |
       #    cd apis/r
@@ -113,7 +113,7 @@ jobs:
       #    Rscript -e 'remotes::install_deps("0.25.0.tar.gz")'
       #    R CMD INSTALL 0.25.0.tar.gz    # as 0.25.0 is needed by TileDB-SOMA 1.9
 
-      # WIP (NOT READY FOR REVIEW) -- LOOKING FOR GREEN CI
+      # NOT SURE THIS IS CORRECT YET -- LOOKING FOR GREEN CI
       #- name: Re-install 0.25 build of tiledb-r with core 2.21
       #  run: |
       #    cd apis/r
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -36,31 +36,15 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-      - uses: actions/checkout@v3
-
-      - name: Show matrix OS
-        run: echo "matrix.os:" ${{ matrix.os }}
-
-      - name: Linux CPU info
-        if: ${{ matrix.os == 'ubuntu-latest' }}
-        run: cat /proc/cpuinfo
-
-      - name: MacOS CPU info
-        if: ${{ matrix.os == 'macOS-latest' }}
-        run: sysctl -a | grep cpu
+      - uses: actions/checkout@v4
 
       - name: Bootstrap
         run: cd apis/r && tools/r-ci.sh bootstrap
 
-      - name: Set pkgType to binary (macOS)
-        if: ${{ matrix.os == 'macOS-latest' }}
-        run: cat("\noptions(pkgType = 'binary')\n", file = "~/.Rprofile", append = TRUE)
-        shell: Rscript {0}
-
       - name: Install BioConductor package SingleCellExperiment
         run: cd apis/r && tools/r-ci.sh install_bioc SingleCellExperiment
 
-
+        
       # Please see https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases which
       # is crucial for anyone doing releases of TileDB-SOMA.
       #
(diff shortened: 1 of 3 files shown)


eddelbuettel:
To be plain, and at least to me, `renv` does not matter.  It's an add-on with its own behavior.  (The author is a close friend though.)

We whould need the suite of package building and checking command, plus `install.packages()`, `available.packages()` and `update.packages()` to support it.

I also learned only this week (on list, from CRAN maintainer Uwe Ligges) that these packages support a `filter` argument that can be function (!!) so apparently one could express one's own preferences or constraints. I would recommend playing with small ad-hoc package on the side rather than burning cycles in our (constrained) CI.  But that's just me.

johnkerl:
> There is a lot of experimenting with bidirectional pins, repo combos, and install chains that I'd like to play around with before adopting this in any way

@mojaveazure +1.

#2373 is a known-good pattern which, factually, _does_ solve the problem for CI. And I'm unaware of solutions I'm able to implement which are available to solve the problem outside of CI.

Those facts, combined with your assertion above which I agree with, mean:

* I will merge #2373 to keep our CI green, as it's a proved fact (four releases now) that it does keep our CI green
* Experiment here
* Anyone else who puts up a PR with a solution for non-CI users, I will be excited to read (and, who knows, this may end up _being_ that PR ...)

apis/r/DESCRIPTION
```diff
@@ -34,7 +34,7 @@ Imports:
     Matrix,
     stats,
     bit64,
-    tiledb (>= 0.25.0),
+    tiledb (>= 0.25.0), tiledb (< 0.26.0),
```
eddelbuettel, line 37: Conceptually this is correct but it has two issue.  First off, R only allows `<=`, not `<`.  Second, the way the releases run in the tiledb-r an earlier 'x' in 0.25.0.6 brought in 2.22.0-rc0.  So I have used `(<= 0.25.0.3)` with luck, looking at the commits `(<= 0.25.0.5)` should also work.  This should make little difference in practice.

```suggestion
    tiledb (>= 0.25.0), tiledb (<= 0.25.0.5)
```

eddelbuettel (dismissed):

.github/workflows/r-ci.yml
```diff
@@ -70,6 +77,13 @@ jobs:
       #
       # Please see https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases which
       # is crucial for anyone doing releases of TileDB-SOMA.
+      #
+      # Please edit both files in the same way:
+      # * r-ci.yml
+      # * r-python-interop-testing.yml
```
eddelbuettel, line 83: I would recommend farming out the logic to a single file that is then included -- as we do here already with with `python-ci-single.yml` which is included by 'minimal' and 'full'.

johnkerl: @eddelbuettel thanks for the recommendation.

I will consider that for follow-on work, which is tracked here:
https://github.com/single-cell-data/TileDB-SOMA/issues/2447

.github/workflows/r-ci.yml
```diff
@@ -116,12 +105,21 @@ jobs:
       - name: Update Packages
         run: Rscript -e 'update.packages(ask=FALSE)'
 
-      - name: Re-install 0.25 build of tiledb-r with core 2.21
-        run: |
-          cd apis/r
-          wget https://github.com/TileDB-Inc/TileDB-R/archive/refs/tags/0.25.0.tar.gz
-          Rscript -e 'remotes::install_deps("0.25.0.tar.gz")'
-          R CMD INSTALL 0.25.0.tar.gz    # as 0.25.0 is needed by SOMA 1.9
+      # WIP (NOT READY FOR REVIEW) -- LOOKING FOR GREEN CI
```
eddelbuettel, line 108: As we're now a CI and not a draft, remove?

apis/r/DESCRIPTION
```diff
@@ -34,7 +34,7 @@ Imports:
     Matrix,
     stats,
     bit64,
-    tiledb (>= 0.25.0),
+    tiledb (>= 0.25.0), tiledb (<= 0.25.0.5)
```
eddelbuettel, line 37: Yes, correct for release 1.9.* as long as the selected r-universe has a suitable version.

eddelbuettel:
That looks like an unrelated error with the Additional_repositories: entry we use for the two data packages.  We made no changes to that aspect, and I think I had a similar 'intra-GH' fluke once before.  (That Additional_repositories: is hosted at GH, but mortals like us do not know if in one or more data centers etc).  

(And it just worked for me in a Docker container here.)

mojaveazure:
I've done some experimentation with double-pinning and I don't think it actually works. I've got a [branch](https://github.com/single-cell-data/TileDB-SOMA/tree/ph/test/doublepin) off of `release/1.8` with a double-pin of tiledb-r to [`>= 0.24.0`, `<= 0.24.99`](https://github.com/single-cell-data/TileDB-SOMA/blob/ph/test/doublepin/apis/r/DESCRIPTION#L37-L38) that's built on [R-universe](https://mojaveazure.r-universe.dev/tiledbsoma) ([source](https://github.com/mojaveazure/mojaveazure.r-universe.dev/blob/da14a871b6fef2aa6a4c2e1b6a9584ea02c38a9f/packages.json#L35-L39)). Using this branch, I've tried testing out double-pins and their effect on the installation and loading/attachment process, and so far I see no effect

R sees two versions of tiledb-r: 0.24.0 from my R-universe and 0.26.0 from CRAN/P3M
```
#8 [5/9] RUN Rscript -e "db <- utils::available.packages(filters = c('R_version', 'OS_type'))"     -e "db[db[, 'Package'] == 'tiledb', c('Version', 'Repository')]"
#8 2.478        Version  Repository
#8 2.478 tiledb "0.24.0" "https://mojaveazure.r-universe.dev/src/contrib"
#8 2.478 tiledb "0.26.0" "https://p3m.dev/cran/__linux__/jammy/latest/src/contrib"
#8 DONE 2.5s
```

R also sees my branch of tiledbsoma with the double-pin
```
#10 [7/9] RUN Rscript -e "uri <- paste(contrib.url(getOption('repos')[1L], 'source'), 'tiledbsoma_1.8.1.tar.gz', sep = '/')"     -e "dest <- tempfile()"     -e "dir.create(dest, showWarnings = FALSE, recursive = TRUE)"     -e "download.file(uri, destfile = file.path(dest, basename(uri)))"     -e "untar(file.path(dest, basename(uri)), files = 'tiledbsoma/DESCRIPTION', exdir = dest)"     -e "desc <- read.dcf(file.path(dest, 'tiledbsoma/DESCRIPTION'), fields = 'Imports')"     -e "unlink(dest, recursive = TRUE, force = TRUE)"     -e "print(desc)"
#10 0.740 trying URL 'https://mojaveazure.r-universe.dev/src/contrib/tiledbsoma_1.8.1.tar.gz'
#10 1.270 Content type 'application/x-gzip' length 720118 bytes (703 KB)
#10 1.275 ==================================================
#10 1.388 downloaded 703 KB
#10 1.388
#10 1.417      Imports
#10 1.417 [1,] "R6, methods, Matrix, stats, bit64, tiledb (>= 0.24.0), tiledb\n(<= 0.24.99), arrow, utils, fs, glue, urltools, Rcpp,\ndata.table, spdl, rlang, tools, tibble"
#10 DONE 1.5s
```

However, when installing this branch of tiledbsoma, R opts for 0.26.0 from CRAN; the double-pin also allows full loading/attachment of tiledbsoma despite the higher version of tiledb-r
```
#12 [9/9] RUN Rscript -e "library(tiledbsoma)"     -e "tiledbsoma::show_package_versions()"
#12 2.375 tiledbsoma:    1.8.1
#12 2.375 tiledb-r:      0.26.0
#12 2.375 tiledb core:   2.22.0
#12 2.375 libtiledbsoma: 2.20.1
#12 2.375 R:             R version 4.3.2 (2023-10-31)
#12 2.375 OS:            Ubuntu 22.04.3 LTS
#12 DONE 2.4s
```

Dockerfile with all steps is available as a [gist](https://gist.github.com/mojaveazure/2c455a6cafc3b4ea2d7731634093c89e); the build used was
```
docker build -t doublepin -f ./doublepin.Dockerfile --progress=plain .
```

eddelbuettel:
@mojaveazure There is a difference in behavior between _installing_ and _loading_ / _attaching_.  The effect of the dual inequality is more on the _latter_ as it seems to have not effect on the former (which would be a 'nice to have' but it not implemented by R).  It remains hard / difficult to 'hide' a higher version available at CRAN.  But see eg my [PR #2450](https://github.com/single-cell-data/TileDB-SOMA/pull/2450) which uncorks the issue here (via explicit installation steps).

johnkerl:
@eddelbuettel @mojaveazure I always appreciate your R-expert dialog!

As I, an R non-expert, am functioning in a clerical role on this PR -- @eddelbuettel and/or @mojaveazure please accept this PR as-is, or accept with proposed changes for a follow-up PR, or request changes on this PR with specific proposed changes -- thank you!

.github/workflows/r-python-interop-testing.yml
```diff
@@ -107,12 +111,13 @@ jobs:
       - name: Update Packages
         run: Rscript -e 'update.packages(ask=FALSE)'
```
eddelbuettel, line 112: We should remove these two lines as we did in the other yaml file.  I can send a follow-up commit.

.github/workflows/r-python-interop-testing.yml
```diff
@@ -107,12 +111,13 @@ jobs:
...
-          wget https://github.com/TileDB-Inc/TileDB-R/archive/refs/tags/0.25.0.tar.gz
-          Rscript -e 'remotes::install_deps("0.25.0.tar.gz")'
-          R CMD INSTALL 0.25.0.tar.gz    # as 0.25.0 is needed by SOMA 1.9
+      # NOT SURE THIS IS CORRECT YET -- LOOKING FOR GREEN CI
+      #- name: Re-install 0.25 build of tiledb-r with core 2.21
+      #  run: |
+      #    cd apis/r
+      #    wget https://github.com/TileDB-Inc/TileDB-R/archive/refs/tags/0.25.0.tar.gz
+      #    Rscript -e 'remotes::install_deps("0.25.0.tar.gz")'
+      #    R CMD INSTALL 0.25.0.tar.gz    # as 0.25.0 is needed by SOMA 1.9
```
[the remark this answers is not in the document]
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/2374/commits/fe8d232f29e5c5abfcca27b4961d6a2e5514df2a

eddelbuettel:
(Had this typed earlier an apparently never sent it / saved it.  Sorry.)

@mojaveazure At the risk of drifting further 'inside baseball' you can look at R's own `src/packages.R` in base packages `utils`.  I took the existing 'CRAN' filter (that makes a CRAN repo version win) and (brute force) added a 'but not for `tiledb`'.  That works.

```r
## This one a 'CRAN' filter in file src/packages.R in 'utils'
## So let's build not-CRAN one for tiledb
available_packages_filters_db_not_CRAN_tiledb <- function(db) {
    packages <- db[, "Package"]
    dups <- packages[duplicated(packages)]
    drop <- integer()
    CRAN <- getOption("repos")["CRAN"]
    ## do nothing if there is no CRAN repos on the list
    if(is.na(CRAN)) return(db)
    for(d in dups) {
        pos <- which(packages == d)
        pkg <- db[pos[1], "Package"]
        if (pkg == "tiledb") {
            ind <- startsWith(db[pos, "Repository"], CRAN)
            if(!all(ind)) drop <- c(drop, pos[ind])
        } else {
            ind <- !startsWith(db[pos, "Repository"], CRAN)
            if(!all(ind)) drop <- c(drop, pos[ind])
        }
    }
    if(length(drop)) db[-drop, , drop = FALSE] else db
}
```

and use it (note that it picks up the CRAN repo via the named entry in `getOptions("repos")` which is not terribly robust against an unnamed list of repo):

```r
> options("repos" = c(univ="https://tiledb-inc.r-universe.dev", getOption("repos")))
> AP1 <- data.frame(available.packages(filters=list(available_packages_filters_db_not_CRAN)))
> AP1[with(AP1, Package=="tiledb"), c(1:2,17)]
       Package Version                                    Repository
tiledb  tiledb  0.25.0 https://tiledb-inc.r-universe.dev/src/contrib
>
```

eddelbuettel:
@johnkerl @mojaveazure I suggest two more changes I can make if we're all on board with them.

.github/workflows/r-python-interop-testing.yml
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/2374/commits/3cb2e01a228d626adf1f827450940435a4b7dbec

johnkerl:
@eddelbuettel @mojaveazure all requested changes have been made

mojaveazure:
@eddelbuettel I showed in the Docker that the double-pin had no effect on loading/attaching. R was more than happy to load tiledbsoma with a version of tiledb-r that exceeded the pin. At this point, I'm not convinced that this PR will do anything to handle mismatches between tiledb-r and tiledbsoma

That being said, we could rig up something in `.onLoad()` to read `DESCRIPTION` and use the double-pin to do its own checking of tiledb-r (I would do it today, but I'm OOO)

I won't object to this PR being merged, I just wanted to flag that R doesn't seem to respect it 🤷

eddelbuettel:
@mojaveazure We can take this off-line. Happy to discuss and dive further. `install.packages()` can use `available.packages()` via a filter as shown above, there is no claim that the dual inequality helps there per se as it appears to it other parts.

eddelbuettel (approved):

eddelbuettel:
@johnkerl Before merging maybe renaming 'with bidirectional-pin' ?  It's not actually bi-directional as I understand it as no information flows back,is there?

johnkerl:
@mojaveazure I appreciate your concerns about correctness.

We do have:

* Green CI on this PR
* On the `release-1.9` branch which needs TileDB-R 0.25 and core 2.21
* While TileDB-R 0.26.0 does exist and is available at CRAN
* Without the controlled-downgrade procedure used on the previous five releases: #1972 #1996 #2157 #2295 #2373

This means we do have correctness signal on this PR -- I believe this means we at least have break-even with #1972 #1996 #2157 #2295 #2373 -- not to say that more can't be done (and I note #2406 will eliminate this problem entirely)

I'll of course be watching everything very closely (as usual) during the 1.10 release process happening this week

Issue #2293, opened 2024-03-21
[python] Importing `tiledb` before `tiledbsoma` leads to segfault in `clib.SOMAObject.open` on MacOS


ryan-williams:
Repro at [ryan-williams/tiledb-repros@import-segfault](https://github.com/ryan-williams/tiledb-scratch/tree/import-segfault):

Importing `tiledb` before `tiledbsoma` causes `open` to segfault on macOS:

```python
import tiledb  # ❌ This unused import causes a segfault in `open` below
import tiledbsoma
tiledbsoma.Experiment.open("pbmc-small")  # ❌ segfaults
```

Click each ✅/❌ to see a full log in Github Actions:
<table>
  <tr>
    <th></th>
    <th colspan="2">
      <code>tiledb</code>,
      <code>tiledbsoma</code>
      versions
    </th>
  </tr>
  <tr>
    <th>Arch/OS</th>
    <th>2.20.1, 1.8.1</th>
    <th>2.21.1, <a href="https://github.com/single-cell-data/TileDB-SOMA/tree/e6822fb4f0a02256ceb487448bd0d13b073946c8"><code>e6822fb</code></a></th>
  </tr>
    <tr>
        <th>Ubuntu</th>
        <td>
          <a href="https://github.com/ryan-williams/tiledb-scratch/actions/runs/8457612701/job/23169976714">✅</a>
          <a href="https://github.com/ryan-williams/tiledb-scratch/actions/runs/8457612701/job/23170035198">✅</a>
          <a href="https://github.com/ryan-williams/tiledb-scratch/actions/runs/8457612701/job/23170201677">✅</a>
        </td>
        <td>
          <a href="https://github.com/ryan-williams/tiledb-scratch/actions/runs/8457048651/job/23168150126#step:9:20">✅</a>
          <a href="https://github.com/ryan-williams/tiledb-scratch/actions/runs/8457048651/job/23169324907">✅</a>
          <a href="https://github.com/ryan-williams/tiledb-scratch/actions/runs/8457048651/job/23170218569">✅</a>
        </td>
    </tr>
    <tr>
        <th>Intel Mac</th>
        <td>
          <a href="https://github.com/ryan-williams/tiledb-scratch/actions/runs/8457636565/job/23170045934#step:10:20">✅</a>
          <a href="https://github.com/ryan-williams/tiledb-scratch/actions/runs/8457636565/job/23170117058#step:10:15">✅</a>
[...]

PR #2379, opened 2024-04-04
[python] Respect import-ordering

johnkerl:
**Issue and/or context:** #2293 

**Changes:**

**Notes for Reviewer:**


ryan-williams:
https://github.com/single-cell-data/TileDB-SOMA/pull/2381 may be a better way to achieve the import-order we want.

johnkerl:
#2381 is way better -- thanks @ryan-williams ! 😎

PR #2380, opened 2024-04-04
[python] Check for uniqueness of registration obs/var field-name inputs

johnkerl:
**Issue and/or context:** When non-unique values are present in `obs` or `var` then append-mode ingest, if it uses that non-uniquely-valued field name as registration inputs, will fail at ingestion time with

```
Error: Duplicate coordinates (1413) are not allowed
```

This problem happens during the data write.

**Changes:**

The error should be surfaced earlier, at registration time:

* Before the write has begun
* WIth a more informative error message
* 
**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/io/_registration/id_mappings.py
+++ b/apis/python/src/tiledbsoma/io/_registration/id_mappings.py
@@ -68,12 +68,22 @@ class ExperimentIDMapping:
         return cls(obs_axis=obs_mapping, var_axes=var_axes)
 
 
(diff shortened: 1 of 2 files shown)


johnkerl:
[sc-44549]

ryan-williams (approved):
lg! I added https://github.com/single-cell-data/TileDB-SOMA/commit/6f452c58d7d548947015f00a55cee7a612a7c728 as a minor cleanup (per side-convo)

johnkerl:
[sc-44675]

PR #2390, opened 2024-04-04
[ci] Debug wheel-build failure [WIP]

johnkerl:
**Issue and/or context:** This PR is an experiment, not for review. Context: 

**Changes:**

**Notes for Reviewer:**

PR #2398, opened 2024-04-05
[r/ci] Unbreak wheel builds

johnkerl:
**Issue and/or context:**

#2393

**Changes:**

`nanoarrow.c` mods were passing all regular per-commit CI but were failing wheel builds which are done nightly, and at release time.

Compile error appearing in wheel builds:
```
    /project/tiledbsoma-1.5.0rc0.post284.dev1878736499/dist_links/libtiledbsoma/src/utils/nanoarrow.c:1775:28: error: null destination pointer [-Werror=format-truncation=]
     1775 |     n_chars_last = snprintf(out, n, ">");
          |                    ~~~~~~~~^~~~~~~~~~~~~
```

**Notes for Reviewer:**

This unblocks my attempt to tag 1.9.4.

Validation: I manually ran https://github.com/single-cell-data/TileDB-SOMA/actions/runs/8572181618/job/23494196523 with PR #2390 which is essentially this PR.

### DIFF
--- a/libtiledbsoma/src/external/src/nanoarrow/nanoarrow.c
+++ b/libtiledbsoma/src/external/src/nanoarrow/nanoarrow.c
@@ -1712,6 +1712,10 @@ int64_t ArrowSchemaToString(const struct ArrowSchema* schema, char* out, int64_t
     return snprintf(out, n, "[invalid: schema is released]");
   }
 
+  if (out == NULL) {
+    return 0;
+  }
+
   struct ArrowSchemaView schema_view;
   struct ArrowError error;
 


eddelbuettel:
Nice catch:

https://github.com/single-cell-data/TileDB-SOMA/blob/5483ea21a927ed05e355f2e6527497230a4848ad/libtiledbsoma/src/external/src/nanoarrow/nanoarrow.c#L1705-L1717

I found upstream quite receptive for the few issues I filed so you could consider mentioning it [at their tracker](https://github.com/apache/arrow-nanoarrow/issues). From following their repo, I am fairly certain they build under Python, do not yet (AFAIK) build at PyPi (but "should be soon") and may not have hit this wheels issue.

johnkerl:
@nguyenv I'll accept that request as a follow-up task.

We need to tag 1.9.4 today and this is blocking the build. I won't be waiting on upstream today.

Happy to accommodate as follow-up.

eddelbuettel:
@johnkerl I did not suggest to wait for upstream to fix it, you have a fix.  But as it is a vendored file, and to save us from future surprises if we ever update the vendoring to a newer version,  it may be beneficial to communicate the fix upstream.

johnkerl:
[sc-44675]

jp-dark (approved):

aaronwolen (approved):

nguyenv (approved):

johnkerl:
@eddelbuettel we are in agreement

PR #2403, opened 2024-04-08
[c++] Address final `valgrind` issue

eddelbuettel:
**Issue and/or context:**

When running under `valgrind` (by for example calling an R script or expression and starting as `R -q -d valgrind`) and with the proper `memcheck` instrumentation for chasing leaks, we were reminded that we lost a small fixed amoumt of 248 bytes per column retrieved.  This was due to allocating the (opaque) `private_data` by default and then (as we do here for buffer management) store the ArrowBuffer / ColumnBuffer container handle.  

**Changes:**

By freeing the default allocation before assigning, we no longer leak.  We also removed another small assignment (and helper function used) taken care of now by `nanoarrow` following #2188.

**Notes for Reviewer:**

[SC 44774](https://app.shortcut.com/tiledb-inc/story/44774/c-squash-final-valgrind-leak)

### DIFF
--- a/libtiledbsoma/src/utils/arrow_adapter.cc
+++ b/libtiledbsoma/src/utils/arrow_adapter.cc
@@ -298,14 +298,6 @@ std::pair<const void*, std::size_t> ArrowAdapter::_get_data_and_length(
     }
 }
 
-bool ArrowAdapter::_isstr(const char* format) {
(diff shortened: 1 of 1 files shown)


eddelbuettel:
Thanks!  It's all valgrind.  

And I still fail at my 'pro license' as I _once again_ forgot the value-add provided by `clang-format-14`.

nguyenv (approved):

PR #2410, opened 2024-04-08
[r] Check that TileDB Core library versions match

### LINKED ISSUE #1837: [python/r] Enforce identical core library version

We have by now learned that TileDB Core version differences between the API package (tiledb-py, tiledb-r) and the tiledbsoma package can lead to undesirable outcomes.   We actually have the versions as the debug helpers (ie in R `show_package_version()`) display them.

We should consider aborting with error when the versions differ.  Possibly to be overriden by an enviroment variable.

If and when all read and writes are done directly by tiledbsoma this will no longer be an issue.  So we _could_ just wait...

**Additional context**

From discussion on Slack with @johnkerl

eddelbuettel:
closes #1837 

**Issue and/or context:**

Issue #1837 proposed a check to ensure library versions of TileDB Core as used by the TileDB R package and the TileDB SOMA package match, and to error if not 

**Changes:**

A check has been added

**Notes for Reviewer:**

[SC 43169](https://app.shortcut.com/tiledb-inc/story/43169/r-py-check-at-package-startup-and-load-and-warn-or-even-die-if-two-different-core-libraries-are-seen)
[SC 44807](https://app.shortcut.com/tiledb-inc/story/44807/r-error-on-mismatched-core-libraries)

### DIFF
--- a/apis/r/R/Init.R
+++ b/apis/r/R/Init.R
@@ -1,0 +1,25 @@
+## .onLoad is called whether code from the package is used and the packages is 'loaded'. An
+## example is calling `tiledbsoma::show_package_versions()`. So this is most elementary check,
+## .onAttach is also called when the package is 'attached' via 'library(tiledbsoma)'
+## During package build and byte-code compilation and load check, both are called.
+.onLoad <- function(libname, pkgname) {
+    rpkg_lib_version <- tiledb::tiledb_version(compact=TRUE)
+    soma_lib_version <- libtiledbsoma_version(compact=TRUE)
+    if (rpkg_lib_version != soma_lib_version) {
+        msg <- sprintf("TileDB Core version %s used by TileDB-R package, but TileDB-SOMA uses %s",
+                       sQuote(rpkg_lib_version), sQuote(soma_lib_version))
+        stop(msg, call. = FALSE)
+    }
+}
+
+## A _possible_ .onAttach() function -- these are not allowed to use cat() etc but _must_ communicate
+## via packageStartupMessage() as this function can be 'muzzled' as desired. See Writing R Extensions.
+## .onAttach <- function(libname, pkgname) {
+##     if (interactive()) {
+##         packageStartupMessage("TileDB-SOMA R package ", packageVersion(pkgname),
+##                               " with TileDB Embedded ", format(tiledb::tiledb_version(TRUE)),
+##                               " on ", utils::osVersion,
+##                               ".\nSee https://github.com/single-cell-data for more information ",
+##                               "about the SOMA project.")
+##     }
+## }
--- a/apis/r/R/Init.R
+++ b/apis/r/R/Init.R
@@ -12,14 +12,14 @@
     }
 }
 
-## A _possible_ .onAttach() function -- these are not allowed to use cat() etc but _must_ communicate
-## via packageStartupMessage() as this function can be 'muzzled' as desired. See Writing R Extensions.
-## .onAttach <- function(libname, pkgname) {
-##     if (interactive()) {
-##         packageStartupMessage("TileDB-SOMA R package ", packageVersion(pkgname),
-##                               " with TileDB Embedded ", format(tiledb::tiledb_version(TRUE)),
-##                               " on ", utils::osVersion,
-##                               ".\nSee https://github.com/single-cell-data for more information ",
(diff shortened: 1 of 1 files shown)


apis/r/R/Init.R
```diff
@@ -0,0 +1,25 @@
...
+## via packageStartupMessage() as this function can be 'muzzled' as desired. See Writing R Extensions.
+## .onAttach <- function(libname, pkgname) {
+##     if (interactive()) {
+##         packageStartupMessage("TileDB-SOMA R package ", packageVersion(pkgname),
+##                               " with TileDB Embedded ", format(tiledb::tiledb_version(TRUE)),
+##                               " on ", utils::osVersion,
+##                               ".\nSee https://github.com/single-cell-data for more information ",
+##                               "about the SOMA project.")
+##     }
+## }
```
johnkerl, line 25: Please remove, or uncomment and utilize

johnkerl (approved):
Thanks @eddelbuettel !

apis/r/R/Init.R
eddelbuettel: I knew this would come but I kept it _on purpose_:  it is near impossible to test or reason about .onAttach and .onLoad without having a working example.  (Illustration: this one is hard to test via a unit test, so you may want to check out the package and flip the `!=` on line eight to a `==` just to see.  It is also useful to see if the test could be in .onAttach (tl;dr: not really).  So give it another minute of thought.  And while you are there, we could also contemplate actually using the .onAttach message just like tiledb-r does:

```r
> library(tiledbsoma)
TileDB-SOMA R package 1.9.99.3 with TileDB Embedded 2.23.0 on Ubuntu 23.10.
See https://github.com/single-cell-data for more information about the SOMA project.
> 
```

But because this is more-or-less outside of the scope of this PR I did not feel like forcing it in.  We could though.

Thoughts?  Easy to adjust either way.  My weak vote is on uncommenting .onAttach.  If you feel that needs another PR it can of course be done.

johnkerl: @eddelbuettel I've accepted the PR, with advice -- feel free to merge at will

eddelbuettel: @johnkerl Good. I didn't mean to or want to 'impose' or 'force' the added note but I think of it as an improvement so I will add it with a second commit.  If it is found to be a nuisance it is equally easy to remove.

johnkerl:
> Illustration: this one is hard to test via a unit test

@eddelbuettel indeed. I did this diff locally:

```diff
diff --git a/apis/r/R/Init.R b/apis/r/R/Init.R
index 78cd0ef6..70badfad 100644
--- a/apis/r/R/Init.R
+++ b/apis/r/R/Init.R
@@ -4,7 +4,7 @@
 ## During package build and byte-code compilation and load check, both are called.
 .onLoad <- function(libname, pkgname) {
     rpkg_lib_version <- tiledb::tiledb_version(compact=TRUE)
-    soma_lib_version <- libtiledbsoma_version(compact=TRUE)
+    soma_lib_version <- '2.99.999' # libtiledbsoma_version(compact=TRUE)
     if (rpkg_lib_version != soma_lib_version) {
         msg <- sprintf("TileDB Core version %s used by TileDB-R package, but TileDB-SOMA uses %s",
                        sQuote(rpkg_lib_version), sQuote(soma_lib_version))
```

and then ran the usual R build+install which failed (as expected) like this:

```
...
** R
** inst
** byte-compile and prepare package for lazy loading
** help
Error : EphemeralCollection.Rd:7: .onLoad failed in loadNamespace() for 'tiledbsoma', details:
  call: NULL
  error: TileDB Core version ‘2.21.2’ used by TileDB-R package, but TileDB-SOMA uses ‘2.99.999’
ERROR: installing Rd objects failed for package ‘tiledbsoma’
* removing ‘/tmp/RtmpFw3iE4/Rinst22dd45d4a363/tiledbsoma’
      -----------------------------------
ERROR: package installation failed
```

so I am feeling good about this.

eddelbuettel:
Yes I did similar things.  I also realized after the fact we could _at least_ check for 'a == b' in a unit test (where, as it runs, they have to be equal) ensuring that the comparison works.  (That said, the little R class for `package_version` is just lovely as in the automagic cast from your '2.99.999' above.)

Issue #2228, opened 2024-03-07
[c++] Create ArrowSchema from TileDB Array Schema (keep-open)


nguyenv:
- [x] Pass in `ColumnIndexInfo` that contains domain and tile extents for each index column
- [x] Set attributes
- [x] Set dimensions
- [x] Handle nullables
- [x] Handle enumerations
- [ ] Set filters
- [ ] Set duplicates

See also #2406

johnkerl:
This issue exists as a coat-hanger for any schema-releated issues that may (or may not) arise while @eddelbuettel is porting @nguyenv 's C++ and Python/pybind11 work to R/Rcpp.

PR #2418, opened 2024-04-09
[c++] Addition of `ArrowSchema` to TileDB `ArraySchema` Converter

nguyenv:
**Issue and/or context:** #2228 

**Changes:**

- Addition of `tiledb_schema_from_arrow_schema` converter that takes in an `ArrowSchema` with `ColumnIndexInfo`, which contains info about the dimension's extent and tiles, and `PlatformConfig`, which contains info for other TileDB `ArraySchema` settings such as filters, etc.
- All `SOMAObject` `create` functions now take in `ArrowSchema`, `ColumnIndexInfo`, and `PlatformConfig` arguments; return signature is void to avoid opening the TileDB object until necessary
- Add unit test helper functions in `common.cc` and `common.h`

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/soma/soma_dataframe.h
+++ b/libtiledbsoma/src/soma/soma_dataframe.h
@@ -52,17 +52,17 @@ class SOMADataFrame : public SOMAArray {
     /**
      * @brief Create a SOMADataFrame object at the given URI.
      *
-     * @param uri URI to create the SOMAArray
+     * @param uri URI to create the SOMADataFrame
      * @param schema TileDB ArraySchema
-     * @param ctx SOMAContext
-     * @param timestamp Optional pair indicating timestamp start and end
-     * @return std::unique_ptr<SOMADataFrame>
+     * @param platform_config Optional config parameter dictionary
      */
-    static std::unique_ptr<SOMADataFrame> create(
+    static void create(
         std::string_view uri,
-        ArraySchema schema,
+        std::shared_ptr<ArrowSchema> schema,
+        ColumnIndexInfo index_columns,
         std::shared_ptr<SOMAContext> ctx,
-        std::optional<TimestampRange> timestamp = std::nullopt);
+        std::optional<PlatformConfig> platform_config = std::nullopt,
+        std::optional<std::pair<uint64_t, uint64_t>> timestamp = std::nullopt);
 
     /**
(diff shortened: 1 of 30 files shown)


eddelbuettel:
Very timely!  Just what I may need for tiledb-r too :sparkles:   If you can maybe try to exclude nanoarrow.h from the enforced linting as it only gets us fairly pointless whitespace diffs to an already vendored external file.  Working with @dudoslav to see if we can make it a more formal external project to avoid those woes.

libtiledbsoma/src/soma/soma_dataframe.h
```diff
@@ -142,7 +142,7 @@ class SOMADataFrame : public SOMAArray {
     /**
      * Return the data schema, in the form of a ArrowSchema.
      *
-     * @return std::unique_ptr<ArrowSchema>
+     * @return std::shared_ptr<ArrowSchema>
      */
     std::unique_ptr<ArrowSchema> schema() const;
```
eddelbuettel, line 147: That seems to a mismatch.

libtiledbsoma/src/soma/soma_dense_ndarray.h
```diff
@@ -52,13 +52,12 @@ class SOMADenseNDArray : public SOMAArray {
...
+     * @param uri URI to create the SOMADenseNDArray
      * @param schema TileDB ArraySchema
-     * @param ctx SOMAContext
-     * @param timestamp Optional pair indicating timestamp start and end
-     * @return std::unique_ptr<SOMADenseNDArray>
+     * @param platform_config Optional config parameter dictionary
+     * @return std::shared_ptr<SOMADenseNDArray> opened in read mode
      */
-    static std::unique_ptr<SOMADenseNDArray> create(
+    static void create(
```
eddelbuettel, line 60: See line 58 about the return value ?

eddelbuettel (commented):
This looks really good now, and easier for me to use than in an earlier form.  I just spotted some small documentation entries.  Worth another pass before merging?

nguyenv:
@eddelbuettel I have taken a closer scan of all the `SOMAArray` and derived classes' docstrings and made necessarily corrections.

eddelbuettel (approved):
Nice work!  I think this is ready to move along and have its tires kicked from my side too!

PR #2425, opened 2024-04-09
[c++] Write data in `ArrowArray` to `SOMAArray`

nguyenv:
**Issue and/or context:** #2426 

**Changes:**

For all `SOMAArray` classes, write by using `set_column_data` with the column name, number of elements to write, pointer to the begining of the buffer, optional pointer to the offsets, and optional pointer for the validity mask. The enumerations for the column are set by using `extend_enumeration` with the column name, number of elements, pointer to the beginning of the buffer, and optional offset buffer. These arguments were chosen to make it convenient to pull directly out of the `ArrowArray` struct.

```
auto soma_dataframe = SOMADataFrame::open(uri, OpenMode::write, ctx);
soma_dataframe->set_column_data("a0", a0.size(), a0.data());
soma_dataframe->set_column_data("d0", d0.size(), d0.data());
soma_dataframe->write();
soma_dataframe->close();
```

**Notes for Reviewer:**

I am not sure why CI did not automatically run the Python tests as expected, but here is a manual run to show they are all passing: https://github.com/single-cell-data/TileDB-SOMA/actions/runs/8622574819


eddelbuettel:
This looks pretty solid.  I have two questions, both are a little meta.  One may reveal that I need to look more closely at the 'schema' branch you plan to merge into.  In your (nice !) example of `soma_dataframe->set_column_data("a0", a0.size(), a0.data());` (ignoring for a second that this is the `array` case not `dataframe`) how do we know what the payload is?  The schema already got covered?   Also, I see 

```c++
void SOMAArray::set_column_data(
    std::string_view name,
    uint64_t num_elems,
    const void* data,
    uint64_t* offsets,
    uint8_t* validity) {
```

which makes sense to but I see no example here with offsets (which may well be null in most cases) or validity (more common maybe?).  I guess you wanted to keep it simple but maybe another example or test would help?

The other questions is even more meta.  I think I have been schooled by `nanoarrow` to prefer `std::unique_ptr` for the Arrow objects, and I see you flipped back to `std::shared_ptr`.  A number of the internal uses (and a lot of the tests) in `nanoarrow` use a copy of 'view' onto schema or array.  That may be something worth considering.  In the standard examples shipping with `nanoarrow` views are often used for printing etc leaving the incoming pointers and their use unchanged.  I haven't really thought or worked this through but could our 'read from the incoming data structure to fill our buffers for the write query' not benefit from a similar approach?

Lastly, even more meta, do we also want a write method directly from `struct ArrowArray*` and `struct ArrowSchema*` (or possibly even their typedef'ed C++ equivalents `UniqueSchema` and `UniqueArray`) ?

eddelbuettel:
This looks pretty good.  From my side as a user of the interface provided a wee bit more on the interface would be nice too. 

One thing I noticed that when running `valgrind libtiledbsoma/test/unit_soma 2>&1 | less` (and the standard `leak-check=full` argument in `~/.valgrindrc`) then the `main` branch looses only the standard 216 bytes via the AWS SDK whereas the new branch looses for a few call of the arrow schema helper (and 260 bytes on each).  Not the biggest problem in the world, and from a tests-only helper function, but maybe worth cleaning up just so it may be easier to notice other (real) issues.

eddelbuettel (approved):
Looks good to me (some comments in the main thread notwithstanding).  Thanks for pushing this forward.

nguyenv:
https://github.com/single-cell-data/TileDB-SOMA/blob/14be77da81448613bc85f22233016b5b8ece3869/apis/python/src/tiledbsoma/soma_array.cc#L42-L91

It is actually in https://github.com/single-cell-data/TileDB-SOMA/pull/2431 where I handle manipulating the `ArrowTable` coming in from Python (more specifically, in the Pybind11 portion) and pass the data, offset, and validity buffers into C++ from there.

I have now switched to using `unique_ptr<ArrowSchema>` in #2418.

nguyenv:
I also realize there's a mistake in that code where the validity argument is hardcoded to `nullptr` but should be `validity`.

nguyenv:
To this PR I've added `SOMAArray::set_array_data` which takes in an Arrow table/batch as represented by an Arrow schema and Arrow array. This sets the write buffers for all columns. I've decided to also keep `SOMAArray::set_column_data` which allows for setting the write buffer for a single column by passing in the pointers to the data and optional offset and validity buffers. I've requested a rereview to make sure everything looks good before merging.

libtiledbsoma/src/soma/column_buffer.h
```diff
@@ -136,6 +116,45 @@ class ColumnBuffer {
...
+     *
+     * @param data pointer to the beginning of the data to write
+     * @param num_elems the number of elements in the column
+     */
+    void set_data(
+        uint64_t num_elems,
+        const void* data,
+        uint64_t* offsets = nullptr,
+        uint8_t* validity = nullptr) {
+        num_cells_ = num_elems;
```
eddelbuettel, line 130: num_cells does not seem be used here

libtiledbsoma/src/soma/soma_array.h
```diff
@@ -515,9 +522,9 @@ class SOMAArray : public SOMAObject {
     /**
      * @brief Get the Arrow schema of the array.
      *
-     * @return std::unique_ptr<ArrowSchema> Schema
+     * @return std::shared_ptr<std::unique_ptr<ArrowSchema>> Schema
```
eddelbuettel, line 525: shared_ptr of unique_ptr of ArrowSchema ?

eddelbuettel:
I added two minor questions inline. Othere than the reversal to shared_ptr is slight concern.

libtiledbsoma/src/soma/column_buffer.h
nguyenv: This sets `this->num_cells_` which gets used as `ColumnBuffer:size`.

eddelbuettel: Thank you!

libtiledbsoma/src/soma/soma_array.h
nguyenv: Thanks for catching this!

nguyenv:
Turns out I forgot to push some changes before rerequesting reviews.

libtiledbsoma/src/soma/soma_array.h
```diff
@@ -515,7 +521,7 @@ class SOMAArray : public SOMAObject {
     /**
      * @brief Get the Arrow schema of the array.
      *
-     * @return std::unique_ptr<ArrowSchema> Schema
+     * @return std::unique_ptr<std::unique_ptr<ArrowSchema>> Schema
```
eddelbuettel, line 524: Dual unique pointer.

```suggestion
     * @return std::unique_ptr<ArrowSchema> Schema
```

johnkerl: Uber-unique 😎

eddelbuettel (approved):
I found a remaining minor documentation wart but otherwise this looks as if it is good to go.   These are all large-ish PRs so can never be sure but we'll revise as needed if issues pop up.

libtiledbsoma/src/soma/soma_array.h
eddelbuettel: Good point. That means it doesn't exist.  Less-than-zero copy.  We should patent that.

aaronwolen:
[sc-46181]

PR #2431, opened 2024-04-10
[python] Use C++ `DataFrame` bindings for create and writes

nguyenv:
**Issue and/or context:** #2427

**Changes:**

Remove TileDB-Py usage for `DataFrame` by using `create` and `write` C++ bindings.


apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -155,7 +156,9 @@ def open(
                     handle._do_initial_reads(auxiliary_reader)
             else:
                 handle._do_initial_reads(tdb)
-        except tiledb.TileDBError as tdbe:
+
+        # TODO macOS throws RuntimeError
```
johnkerl, line 160: ✨

johnkerl (approved):

eddelbuettel:
This once again whitespace changed nanoarrow.h which, as an external file, is not really needed -- and makes comparison to upstream more difficult.

jp-dark (approved):
Looks good to me.

ryan-williams (approved):
lgtm, thanks for the walk-through yesterday!

Issue #2433, opened 2024-04-11
[python] Ingestion performance


johnkerl:
There are opportunities within the ingestor to try/catch create then open, rather than the other way around, at zero functional change but with improved processing times.

Idea:

* There are times we want to open something for write if it exists, or create it if it doesn't
  * Sometimes we simply do that: just to get a thing existing at a given URI, whether by creating or opening for write
  * Other times it's an error if the thing already exists (e.g. ingest without `ingest_mode="resume"`)
* Currently we've been doing a try-open-for-write, and catching a does-not-exist error with trying a create
  * All else being equal this would be as good as doing try-create, catching an already-exists error with trying an open for write
  * However in certain contexts, the if-exists logic in the try-open-for-write case can take unnecessarily long
  * This is particularly relevant for _ab nihilo_ ingests for which the thing not existing is the normal case
  * In either _ab nihilo_ ingests, or append-mode ingests, it turns out to be quite performant to do the try-create _first_ and then fail back to try-open-for-write
  * In summary: we have create and open, and open can be more expensive, so it's always best to try create then open

[sc-44853]

PR #2434, opened 2024-04-11
[python] Ingestion performance

johnkerl:
**Issue and/or context:** #2433

**Changes:** As narrated on #2433. The only additional details is that to balance the `DoesNotExistError` for `open`, we need a completely analogous `AlreadyExistsError` for `create`.

**Notes for Reviewer:** There is a bit more performance improvement to be made, but it needs to be done in core; I will track that work separately. Also, as this is for Python; I'll track the R work -- both a needs-analysis, and any possible (maybe no) dev work -- separately.

### DIFF
--- a/apis/python/src/tiledbsoma/_exception.py
+++ b/apis/python/src/tiledbsoma/_exception.py
@@ -82,10 +82,7 @@ def is_already_exists_error(e: tiledb.TileDBError) -> bool:
     # Local-disk, S3, and TileDB Cloud exceptions all have the substring
     # "already exists". Here we lower-case the exception message just
     # in case someone ever uppercases it on the other end.
-    if "already exists" in stre.lower():
-        return True
-
-    return False
+    return "already exists" in stre.lower()
 
 
 def is_duplicate_group_key_error(e: tiledb.TileDBError) -> bool:
--- a/apis/python/src/tiledbsoma/_exception.py
+++ b/apis/python/src/tiledbsoma/_exception.py
@@ -55,6 +55,46 @@ def is_does_not_exist_error(e: tiledb.TileDBError) -> bool:
     return False
 
 
+class AlreadyExistsError(SOMAError):
+    """Raised when attempting to create an already existing SOMA object.
+
+    Lifecycle: experimental
+    """
+
+    pass
+
+
+def is_already_exists_error(e: tiledb.TileDBError) -> bool:
+    """Given a TileDBError, return true if it indicates the object already exists
+
+    Lifecycle: experimental
+
+    Example:
+        XXX EDIT ME
+        try:
+            with tiledb.open(uri):
+                ...
+        except tiledb.TileDBError as e:
+            if is_does_not_exist_error(e):
+                ...
+            raise e
+    """
+    stre = str(e)
+    # Local-disk/S3 does-not-exist exceptions say 'Group does not exist'; TileDB Cloud
+    # does-not-exist exceptions are worded less clearly.
+    if (
+        "lready exists"
+        in stre
+        # XXX
+        # or "Unrecognized array" in stre
+        # or "HTTP code 401" in stre
+        # or "HTTP code 404" in stre
+    ):
+        return True
+
+    return False
+
+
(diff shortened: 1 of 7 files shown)


apis/python/src/tiledbsoma/_exception.py
```diff
@@ -55,6 +55,46 @@ def is_does_not_exist_error(e: tiledb.TileDBError) -> bool:
...
+        except tiledb.TileDBError as e:
+            if is_does_not_exist_error(e):
+                ...
+            raise e
+    """
+    stre = str(e)
+    # Local-disk/S3 does-not-exist exceptions say 'Group does not exist'; TileDB Cloud
+    # does-not-exist exceptions are worded less clearly.
+    if (
+        "lready exists"
```
nguyenv, line 86: Typo?

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -2749,15 +2749,15 @@ def _ingest_uns_ndarray(
...
-    except DoesNotExistError:
         soma_arr = DenseNDArray.create(
             arr_uri,
             type=pa_dtype,
             shape=value.shape,
             platform_config=platform_config,
             context=context,
         )
+    except AlreadyExistsError:
+        soma_arr = _factory.open(arr_uri, "w", soma_type=DenseNDArray, context=context)
```
nguyenv, line 2760: We can actually just directly call `DenseNDArray.open` here (and same with `DataFrame.open` above).

apis/python/src/tiledbsoma/_exception.py
```diff
@@ -55,6 +55,46 @@ def is_does_not_exist_error(e: tiledb.TileDBError) -> bool:
...
+        in stre
+        # XXX
+        # or "Unrecognized array" in stre
+        # or "HTTP code 401" in stre
+        # or "HTTP code 404" in stre
```
nguyenv, line 91: Did we need to check these error messages too?

apis/python/src/tiledbsoma/_exception.py
```diff
@@ -55,6 +55,46 @@ def is_does_not_exist_error(e: tiledb.TileDBError) -> bool:
...
+    pass
+
+
+def is_already_exists_error(e: tiledb.TileDBError) -> bool:
+    """Given a TileDBError, return true if it indicates the object already exists
+
+    Lifecycle: experimental
+
+    Example:
+        XXX EDIT ME
```
ryan-williams, line 73: intended?

ryan-williams (commented):
looks good, just one additional q re: a `XXX` comment

apis/python/src/tiledbsoma/_exception.py
johnkerl: No, I wanted `Already` or `already`

But it looks wrong -- I'll add ☝️ as a comment

apis/python/src/tiledbsoma/_exception.py
johnkerl: I'll remove them -- they're copypasta from does-not-exist which don't apply in the already-exists case

apis/python/src/tiledbsoma/_exception.py
johnkerl: 😬 

Derp, I put up this PR too quick -- my apologies

johnkerl:
@nguyenv @ryan-williams ready for re-review! :)

apis/python/src/tiledbsoma/_exception.py
```diff
@@ -55,6 +55,39 @@ def is_does_not_exist_error(e: tiledb.TileDBError) -> bool:
...
+            raise e
+    """
+    stre = str(e)
+    # Local-disk, S3, and TileDB Cloud exceptions all have the substring
+    # "already exists". Here we lower-case the exception message just
+    # in case someone ever uppercases it on the other end.
+    if "already exists" in stre.lower():
+        return True
+
+    return False
```
nguyenv, line 88:
```suggestion
    return "already exists" in stre.lower()
```

nguyenv (approved):
Looks goods to me. Just have one minor suggestion.

PR #2445, opened 2024-04-15
Fix Windows build regressions and stop using deprecated TileDB APIs.

teo-tsirpanis:
**Issue and/or context:** As part of validating that libtiledbsoma does not use deprecated TileDB APIs, I tried to build it on my Windows machine with a build of TileDB with all deprecated APIs removed from the headers. I encountered a couple of errors and this PR fixes them. I also migrated two deprecated API usages in the tests.

**Changes:** With an exception in two tests, we mostly replace constructs that caused errors for MSVC with constructs that do not. The deprecated API migration is very straightforward.

Validated by successfully building libtiledbsoma and `unit_soma` on my local machine and running the tests. To prevent further regressions we should eventually work towards reenabling Windows CI in #1962.

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/reindexer/reindexer.cc
+++ b/libtiledbsoma/src/reindexer/reindexer.cc
@@ -32,7 +32,6 @@
 
 #include "reindexer.h"
 #include <thread_pool/thread_pool.h>
-#include <unistd.h>
 #include <thread>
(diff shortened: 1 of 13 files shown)


nguyenv (approved):
Thank you!

PR #2446, opened 2024-04-16
[python/c++] Use C++ bindings for `SparseNDArray` write path

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2428

**Changes:**

* Use C++ bindings for `SparseNDArray` write path
* Also modify SOMA typechecks to be case insensitive


apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -93,26 +93,26 @@ def open(
 
     if not soma_type:
         raise DoesNotExistError(f"{uri!r} does not exist")
+    
+    soma_type = soma_type.lower()
```
johnkerl, line 97: @nguyenv why?

The "real names" in metadata are camelcased and I fear the lowercasing, which leads to two different case-sensitive spellings now, will be error-prone later on.

Why not simply keep the agreed-on, standard, single camel-cased spellings and use them consistently and error-free everywhere?

johnkerl (dismissed):
As noted re lowercasing. Unless I'm missing something I don't see a good reason for this, and I think it'll be error-prone in the future. If there's a good reason I'm missing, please eludicate it as a code comment. Thanks!

apis/python/src/tiledbsoma/_tdb_handles.py
nguyenv: https://github.com/single-cell-data/TileDB-SOMA/blob/viviannguyen/ndarray-write-path/apis/python/tests/test_factory.py#L41-L78

I modified this because the pre-existing unit tests actually do explicitly check that the type is case insensitive (the d in `NDArray` may be lower or upper case).

johnkerl: :( @nguyenv I really don't like mixing cases :( -- would it be too much to ask for us to simply take this opportunity to keystroke out correct casings for once and for all?

johnkerl: @nguyenv I really don't like this at all but it appears not to be new to this PR. Thanks.

johnkerl:
R CI is a KP, as TileDB-R 0.26.0 exists but we've not updated TileDB-SOMA yet. We are actively addressing this and will have a fix soon. Meanwhile the red R CI should not be used as a reason to not merge this PR.

nguyenv:
I'm not so sure that my failure is related to the TileDB-R 0.2.26 because the only test it is failing is the SparseNDArray Python write  and R readback (and not vice versa) which is exactly what this PR changes. However, I'll go ahead and switch to working on other tasking.

johnkerl:
@nguyenv I may be mistaken -- CI fail for this PR
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/8706640564/job/23879684263?pr=2446
is not related to core versions.

johnkerl:
@nguyenv I just rebased on top of #2448 (the R CI thing) & indeed an issue does still remain ...

apis/python/src/tiledbsoma/_arrow_types.py
```diff
@@ -50,6 +50,24 @@
     pa.date32(): TypeError("32-bit date - unsupported type (use TimestampType)"),
     pa.date64(): TypeError("64-bit date - unsupported type (use TimestampType)"),
 }
+
+_PYARROW_TO_CARROW: Dict[pa.DataType, str] = {
```
johnkerl, line 54: Awesome 😎

apis/python/src/tiledbsoma/_exception.py
```diff
@@ -138,8 +138,23 @@ def is_duplicate_group_key_error(e: Union[SOMAError, tiledb.TileDBError]) -> boo
...
-    stre = str(e)
-    if "member already exists in group" in stre:
-        return True
+    return "member already exists in group" in str(e)
 
-    return False
+
+def is_domain_setting_error(e: SOMAError) -> bool:
+    """Given a SOMAError, return whether it attempted to create the ArraySchema
+    but resulted in a
```
johnkerl, line 146: resulted in a ... ?

apis/python/src/tiledbsoma/_util.py
```diff
@@ -322,3 +324,86 @@ def verify_obs_var(ad0: AnnData, ad1: AnnData, nan_safe: bool = False) -> None:
...
+) -> pa.Table:
+    """
+    When writing data to a SOMAArray, the values that the user passes in may not
+    match the schema on disk.
+
+    1. Cast the values to the correct dtypes
+    2. Perform special casting for Boolean as it is represented in TileDB as
+    int8 but bits in Arrow
+    3. If there are new enumeration values in the values that are not yet
+    present in the schema, exten the enumeration to include these new values
```
johnkerl, line 340:
```suggestion
    present in the schema, extend the enumeration to include these new values
```

johnkerl (commented):
Deferring approval to @eddelbuettel who has higher stakes here :)

apis/python/src/tiledbsoma/_arrow_types.py
```diff
@@ -50,6 +50,24 @@
...
+    pa.uint64(): "L",
+    pa.float32(): "f",
+    pa.float64(): "g",
+    pa.timestamp("s"): "tss:",
+    pa.timestamp("ms"): "tsm:",
+    pa.timestamp("us"): "tsu:",
+    pa.timestamp("ns"): "tsn:",
+}
 """Dict of types unsupported by to_pandas_dtype, which require overrides for
 use in TileDB Attributes (aka DataFrame non-indexe columns).
```
thetorpedodog, line 72: This comment has been separated from its home :(

apis/python/src/tiledbsoma/_arrow_types.py
```diff
@@ -237,3 +255,9 @@ def df_to_arrow(df: pd.DataFrame) -> pa.Table:
         arrow_table = arrow_table.replace_schema_metadata(md)
 
     return arrow_table
+
+
+def pyarrow_to_carrow_type(pa_type: pa.DataType) -> str:
+    if pa_type not in _PYARROW_TO_CARROW:
+        raise TypeError(f"Invalid pyarrow type {pa_type}")
+    return _PYARROW_TO_CARROW[pa_type]
```
thetorpedodog, line 263: eafp:

```python
try:
  return _PYARROW_TO_CARROW[pa_type]
except KeyError:
  raise ... from None
```

apis/python/src/tiledbsoma/_exception.py
```diff
@@ -138,8 +138,23 @@ def is_duplicate_group_key_error(e: Union[SOMAError, tiledb.TileDBError]) -> boo
...
+
+    Lifecycle: maturing
+    """
+    return "Cannot set domain" in str(e)
+
+
+def map_exception_for_create(e: SOMAError, uri: str) -> None:
```
thetorpedodog, line 153: If this is intended to be used in error-handling code, it might be better to have this return the error and raise the new one, so that the caller can put `from None` or `from prev_error` on it. Otherwise, instead of getting:

```
AlreadyExistsError: "some_uri" already exists"
... traceback ...
the cause of this error was ...
```

you get

```
SOMAError: ...
While that was being handled, this happened:
AlreadyExistsError: ...
```

so then at the callsite it looks like:

```python
try:
  ...
except SOMAError as se:
  raise map_exception_for_create(se, ...) from se  # or "from None" if you prefer

File: apis/python/src/tiledbsoma/_sparse_nd_array.py [hunk 8]
```diff
@@ -348,15 +424,14 @@ def used_shape(self) -> Tuple[Tuple[int, int], ...]:
...
-        with tiledb.open(self.uri, ctx=self.context.tiledb_ctx):
-            for i in range(20):
-                lower_key = f"soma_dim_{i}_domain_lower"
-                lower_val = self.metadata.get(lower_key)
-                upper_key = f"soma_dim_{i}_domain_upper"
-                upper_val = self.metadata.get(upper_key)
-                if lower_val is None or upper_val is None:
-                    break
-                retval.append((lower_val, upper_val))
+        for i in range(20):
```
thetorpedodog, line 427: not directly related to this change but: where did 20 come from?

apis/python/src/tiledbsoma/_util.py
```diff
@@ -322,3 +324,86 @@ def verify_obs_var(ad0: AnnData, ad1: AnnData, nan_safe: bool = False) -> None:
...
+            )
+
+        if pa.types.is_boolean(input_field.type):
+            target_schema.append(target_field.with_type(pa.uint8()))
+        else:
+            target_schema.append(target_field)
+    return values.cast(pa.schema(target_schema, values.schema.metadata))
+
+
+def set_clib_platform_config(
```
thetorpedodog, line 384: `build_clib_platform_config` might be a better name.

apis/python/src/tiledbsoma/_sparse_nd_array.py
johnkerl: I put that in a long time ago & didn't comment it well : (

apis/python/src/tiledbsoma/_arrow_types.py
nguyenv: Fixed now but just curious -- I think this is the second time in this codebase where the documentation is beneath the data struct instead of above it. Is that typical?

apis/python/src/tiledbsoma/_sparse_nd_array.py
nguyenv: Good question. I think this should the number of dimensions in the array and not hardcoded to 20? 

@johnkerl

johnkerl: > Good question. I think this should the number of dimensions in the array and not hardcoded to 20?

@nguyenv yes

Issue #2447, opened 2024-04-16
[r/ci] Centralize r-universe stanzas from various R CI YAML files


johnkerl:
At present we have three R CI YAML files which involve installing a correct/compatible version of TileDB-R:

* `.github/workflows/r-ci.yml`
* `.github/workflows/r-python-interop-testing.yml`
* `.github/workflows/r-valgrind.yaml`

It is possible that we would extract out the common/reused blocks into a helper YAML file.

But also note that via C++ification work we will be entirely eliminating TileDB-SOMA's dependency on TileDB-R and TileDB-Py both, at which point the above-referenced stanzas will disappear entirely.

PR #2450, opened 2024-04-17
[r/ci] Required changes to `r-ci.yml`, plus simplifications

eddelbuettel:
**Issue and/or context:**

TileDB-SOMA uses TileDB Embedded and needs to ensure it builds / tests with a TileDB-R version using the same TileDB Embedded version

**Changes:**

Ensure the corresponding version from tiledb-inc.r-universe.dev is used on macOS and Linux

The r-ci.yml contained a number of paragraphs we had accumulated over the last several month when debugging other issues.  They are not needed, and may distract overall.  So this PR removes them but they can be restored as desired. This PR aims to serve as an existence proof of working CI as e.g. provided in [this run](
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/8714532260/job/23904928763) moments ago.

**Notes for Reviewer:**

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -36,31 +36,15 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-      - uses: actions/checkout@v3
-
-      - name: Show matrix OS
-        run: echo "matrix.os:" ${{ matrix.os }}
-
-      - name: Linux CPU info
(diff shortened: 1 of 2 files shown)


johnkerl (approved):
Thank you!

eddelbuettel:
My pleasure!   It's all still a tad experimental but this may be helping us along.  Also another score for making the changes in a single file that is sourced.  When we have time to get to it...

johnkerl:
#2447

PR #2451, opened 2024-04-17
Bump `DESCRIPTION` version past 1.10 (`main` branch)

johnkerl:

Please see #2328 for previous

[sc-44519]

### DIFF
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -6,7 +6,7 @@ Description: Interface for working with 'TileDB'-based Stack of Matrices,
     like those commonly used for single cell data analysis. It is documented at
     <https://github.com/single-cell-data>; a formal specification available is at
     <https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md>.
-Version: 1.9.99.3
+Version: 1.10.99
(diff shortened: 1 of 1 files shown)


johnkerl:
[sc-44519]

apis/r/DESCRIPTION
```diff
@@ -34,7 +34,7 @@ Imports:
     Matrix,
     stats,
     bit64,
-    tiledb (>= 0.26.0),
+    tiledb (>= 0.26.0), tiledb (<= 0.26.99),
```
eddelbuettel, line 37: Yep.  If and when tiledb-r changes version and if and when that would bring in a newer TileDB Embedded we'd have to lower the '99'.   But it is a 'known unknown' now and this, even just a signal value, may help.

eddelbuettel (approved):
:shipit:

apis/r/DESCRIPTION
johnkerl: @eddelbuettel as always a bump to core 2.23 would be accompanied by a bump of TileDB-R to 0.27 -- if a TileDB-R 0.26.anything depends someday on a core 2.23, that would be an error.

PR #2455, opened 2024-04-19
[python] Append-mode enum bit-width mods

johnkerl:
**Issue and/or context:** Presented in detail on #2456.

**Changes:** As on #2456.

**Notes for Reviewer:** See also updates below.

### DIFF
--- a/apis/python/tests/test_registration_mappings.py
+++ b/apis/python/tests/test_registration_mappings.py
@@ -25,7 +25,6 @@ def _create_anndata(
     X_value_base: int,
     measurement_name: str,
     raw_var_ids: Optional[Sequence[str]] = None,
-    X_density: float = 0.3,
 ):
     n_obs = len(obs_ids)
     n_var = len(var_ids)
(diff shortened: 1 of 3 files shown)


johnkerl:
Update: @nguyenv 's recent cppification has fixed the case when a larger-bit-width enum exists on disk, with a smaller-bit-width append -- this now works correctly.

The only remaining issue is: when a smaller bit width is done for the initial ingest, and then we overflow.

* For the`tiledbsoma` API: this is intentional: if a user supplies say `int8` index type in their user-provided Arrow schema and user-provided data exceed that capacity, then we must fail -- and we do
* For the `tiledbsoma.io` API: users do _not_ pass in Arrow schema -- they simply pass in AnnData objects and they are entitled to expect appendability. The necessary step here is for `tiledbsoma.io` to _broaden_ our inference for index types to `int32` -- regardless of the initial input data -- since any enumerated column on any dataset could be appended to.

apis/python/tests/test_registration_mappings.py
```diff
@@ -1217,3 +1216,105 @@ def test_append_with_nonunique_field_values(
...
+        # XXX assert schema enum is categ
+        print()
+        print()
+        print("-- SOMA SCHEMA")
+        print(exp.obs.schema)
+        print("-- ARROW SCHEMA")
+        print(obs.schema)
+        print("-- PANDAS DATA")
+        print(obs.to_pandas())
+        print()
```
nguyenv, line 1312: Need to rm print statements from unit tests.

johnkerl: Sorry @nguyenv ... `apis/python/tests/test_update_dataframes.py` was _not_ ready for review (at all) -- I shouldn't have put it up as-is ...

johnkerl:
@nguyenv https://github.com/single-cell-data/TileDB-SOMA/pull/2455/commits/83fd5c45be2b758f8ddd914a6af93db5f1199a4a

nguyenv (approved):

Issue #2413, opened 2024-04-09
[python/ci] Provisional/evaluation Linux + arm64 builds


johnkerl:
# Context

As discussed in Slack. Non-urgent but worth some experimentation.

At present we have Linux x86, MacOS x86, MacOS arm.

Use cases:

* AWS Graviton
* Anyone doing docker-build/docker-run of Linux images on MacOS M-series systems

Many thanks to @ihnorton and @jdblischak for support on making these evals happen.

# Evaluation results: Docker

To build: please place the following in an otherwise empty directory as `Dockerfile.eval`:

```
FROM python:3.11

SHELL ["/bin/bash", "-c"]

RUN apt-get update && \
    apt-get install -y build-essential zip unzip tar ninja-build && \
    apt-get install -y libhdf5-dev && \
    apt-get clean && rm -rf /var/lib/apt/lists/*

WORKDIR /app

# required for libtiledb source build

ENV VCPKG_FORCE_SYSTEM_BINARIES=1
ENV VENV="/opt/env"
ENV PATH="$VENV/bin:$PATH"

RUN python3 -m venv $VENV

RUN pip3 install --upgrade pip wheel

RUN pip3 install cellxgene-census
```

Then:

```
docker build -f Dockerfile -t linux-aarch64-eval .
```

To run:

```
alias dreb='docker run --rm -it --entrypoint=/bin/bash'
dreb linux-aarch64-eval
export MAMBA_NO_BANNER=1
mamba install -c jdblischak/label/experimental -c conda-forge -c tiledb tiledbsoma-py
python
```

```
import tiledbsoma
tiledbsoma.show_package_versions()
# ...
```

# Evaluation results: Conda

Artifacts: https://anaconda.org/jdblischak/tiledbsoma-py/files?channel=experimental

Usage:

```
alias dreb='docker run --rm -it --entrypoint=/bin/bash'
dreb condaforge/miniforge3
export MAMBA_NO_BANNER=1
mamba install -c jdblischak/label/experimental -c conda-forge -c tiledb tiledbsoma-py
```

# Full support

Tracked in #2482 and [sc-46090]

johnkerl:
[sc-44849]

johnkerl:
Completed provisional test. Follow-up on #2482.

PR #2470, opened 2024-04-22
[ci] Linux/ARM64 conda-forge debug [WIP / do not merge]

johnkerl:
**Issue and/or context:** #2413

**Changes:** Debug-only for https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/138

**Notes for Reviewer:**

### DIFF
--- a/apis/r/tools/get_tarball.R
+++ b/apis/r/tools/get_tarball.R
@@ -21,6 +21,7 @@ if (isMac) {
       stop("Unsupported Mac architecture. Please have TileDB Core installed locally.")
     }
 } else if (isLinux) {
+    # TODO: LOOK AT THIS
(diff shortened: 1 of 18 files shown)


apis/r/tools/get_tarball.R
```diff
@@ -21,6 +21,7 @@ if (isMac) {
       stop("Unsupported Mac architecture. Please have TileDB Core installed locally.")
     }
 } else if (isLinux) {
+    # TODO: LOOK AT THIS
```
eddelbuettel, line 24: Right on. The Linux side, as far as pre-made blobs are concerned, only knows x86_64 per the following line.

(Same thing is happening with Windows now where R build start for arm64, but not a concern for us yet as we do not do Windows, and also do not ship to CRAN).

Theses nested build scripts are not great but "evolved".  You can test a little in, say, a container by running each of one of them (down from the initial call from configure) and see where things are after each steps.

PR #2472, opened 2024-04-22
[c++] Pass `ColumnIndexInfo` as `ArrowTable`

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2471

**Changes:**

Change the argument dtype of `ColumnIndexInfo` to `ArrowTable`


eddelbuettel (approved):
Nice generalisation which, _mea maxima culpa_, was driven by an oversight and stupid error in my test code.

All better now.

PR #2479, opened 2024-04-25
Remove arbitrary dimension limit in `used_shape`.

thetorpedodog:
We were previously using `range(20)` as an arbitrarily high number of dimensions that people were unlikely to use. Instead, `itertools.count` provides a range-like counter that goes on forever. The function is still guaranteed to exit because there will always be a finite number of `soma_dim_{i}_domain_{endpoint}` metadata values.

---

I'm going to wait until #2446 is in to submit this, to avoid stepping on that change, but am putting it out now so that when that is in, I can just rebase and be done.


johnkerl (approved):
🙇

PR #2481, opened 2024-04-25
[c++] Add documentation and better checking to `PlatformConfig`

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2480

**Changes:**

Add inline documentation for `PlatformConfig` options and add more robust checking for passed in values.

### DIFF
--- a/libtiledbsoma/src/utils/arrow_adapter.cc
+++ b/libtiledbsoma/src/utils/arrow_adapter.cc
@@ -259,8 +259,14 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
         if (platform_config->offsets_filters.size() != 0) {
             FilterList offset_filter_list(*ctx);
             for (auto offset : platform_config->offsets_filters) {
-                offset_filter_list.add_filter(
-                    Filter(*ctx, convert_filter[offset]));
+                try {
+                    offset_filter_list.add_filter(
+                        Filter(*ctx, convert_filter.at(offset)));
+                } catch (std::out_of_range& e) {
+                    throw TileDBSOMAError(fmt::format(
+                        "Invalid offset filter {} passed to PlatformConfig",
+                        offset));
+                }
             }
             schema.set_offsets_filter_list(offset_filter_list);
         }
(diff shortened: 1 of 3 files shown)


libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -259,32 +259,60 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
...
         }
 
         schema.set_allows_dups(platform_config->allows_duplicates);
 
-        if (platform_config->tile_order)
-            schema.set_tile_order(
-                platform_config->tile_order == "row" ? TILEDB_ROW_MAJOR :
-                                                       TILEDB_COL_MAJOR);
+        if (platform_config->tile_order) {
+            if (platform_config->tile_order == "row") {
```
eddelbuettel, line 292: I am wondering if we could extend the comparison. I get passed "ROW_ORDER" and "COL_ORDER", respectively, from @mojaveazure

eddelbuettel (changes requested):
Looks good and helps, left one minor question for possible values.

Please a sport for sparse vs dense though (unless that will come in under a parallel, earlier PR in which case I can of course approve).

libtiledbsoma/src/utils/arrow_adapter.cc
nguyenv: There is a check for both "row" and "col" here, so I'm not sure what you mean.

eddelbuettel: To add an `|| platform_config->tile_order == "ROW_ORDER"` as that is the value I (as the middle man) receive.  Saves me from mapping it.

Do we what if anything the spec suggests?  I suppose you got `row` from Python. If R uses `ROW_ORDER` then one of the two may be wrong...

nguyenv: OIC. Yeah I can modify this.

eddelbuettel: The R package currently has these strings as defaults:

```r
.default_tile_order        <- function() { "ROW_MAJOR" }
.default_cell_order        <- function() { "ROW_MAJOR" }

# ...

.default_offsets_filters  <- function() { list("DOUBLE_DELTA", "BIT_WIDTH_REDUCTION", "ZSTD")}
```

There is a bit of a larger issue looming that maybe we want the C++ side to own the defaults and have the API just instantiate the difference from defaults.  Will try to chat with @mojaveazure about his side of things...

nguyenv:
I have extended the options to include col-major and row-major. I also added the option for creating a sparse or dense array as a method argument instead of within `PlatformConfig` because I don't think this should be configurable by the user.

eddelbuettel (approved):

nguyenv:
The array type is dependent on the SOMAObject type so it doesn't make sense to be able to arbitrarily set it. The other options you listed don't have the same restrictions.

eddelbuettel:
Fair point!

nguyenv:
Do you think instead of passing in a `tiledb_array_type_t` we should pass in a Boolean indicating whether the array should be sparse or not?

eddelbuettel:
Or fold it into another PR ...   But yeppers a `bool` seems apt here (and is easy to deal with as a plain old data type).

PR #2483, opened 2024-04-25
[r] Set default create options during `TileDBCreateOptions$new()`

mojaveazure:
Set default creation options during `TileDBCreateOptions$initialize()` instead of on-the-fly in the accessors. This allows easy serialization of `TileDBCreateOptions` with `TileDBCreateOptions$to_list()`

Also add a new `build_filters` parameter to `TileDBCreateOptions$to_list()`; if `TRUE`, builds filters with `tiledb::tiledb_filter()`, otherwise leaves as characters/lists of characters describing the filters

Also simplify a lot of the accessors/filter building methods, and move `$.dim()`, `$.attr()`, `$.build_filters()`, and `$.build_filter()` to be private methods instead of public ones


johnkerl (approved):
🚢

mojaveazure:
[sc-48524]

PR #2487, opened 2024-04-26
[c++] Pass Boolean to indicate sparse or dense in ArraySchema

nguyenv:
**Issue and/or context:**

This is a follow up PR to #2481.

**Changes:**

Use `bool` instead of `tiledb_array_type_t` to indicate whether the ArraySchema should be sparse or dense.


eddelbuettel (approved):
:rocket: 

For style, is `is_sparse` more common / readable than `issparse` ?  Either works for me.

Issue #2429, opened 2024-04-09
[python] Use `DenseNDArray::create` and `write` bindings in Python


nguyenv:
Parent task: https://github.com/single-cell-data/TileDB-SOMA/issues/2241

johnkerl:
@nguyenv please link to the parent task

PR #2494, opened 2024-04-29
[python,c++] Use C++ `DenseNDArray` bindings in Python API

nguyenv:
**Issue and/or context:** 
https://github.com/single-cell-data/TileDB-SOMA/issues/2429

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -1785,9 +1785,6 @@ def _write_matrix_to_denseNDArray(
             )
             return
 
-    if isinstance(matrix, sp.csc_matrix):
-        matrix = sp.csr_matrix(matrix)
-
     # Write all at once?
     if not tiledb_create_options.write_X_chunked:
         if not isinstance(matrix, np.ndarray):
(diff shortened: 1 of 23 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1785,6 +1785,9 @@ def _write_matrix_to_denseNDArray(
             )
             return
 
+    if isinstance(matrix, sp.csc_matrix):
+        matrix = sp.csr_matrix(matrix)
```
nguyenv, line 1789: This is a workaround for getting CSC matrices to write properly with the new `clib.DenseNDArray` bindings. I would appreciate it if someone could suggest a better solution.

johnkerl (changes requested):
https://github.com/single-cell-data/TileDB-SOMA/pull/2494#discussion_r1584926419

apis/python/src/tiledbsoma/io/ingest.py
nguyenv: Thank you so much @johnkerl! This is extremely helpful, and I really appreciate it. I couldn't figure it out after a weekend of debugging.

johnkerl (approved):
🚢 
Awesome @nguyenv !! :)

PR #2497, opened 2024-04-30
[r] Better handling of timestamps in `$reopen()`

mojaveazure:
Internal timestamps of not `NULL` cause `$reopen()` to fail when reopening in `"WRITE"` mode; this PR allows `$reopen()` to do one of the following:
- set the time to the current time (required for `$reopen("WRITE")`)
- restore the time to the previously set timestamp
- open at a custom timestamp by passing a `POSIXct` object

Currently, the existing timestamp causes an error when reopening for write and modifying an existing array (eg. schema evolution). For example
```r
library(SeuratObject)
data(pbmc_small)

uri <- "/path/to/uri"
tiledbsoma::write_soma(pbmc_small, uri = uri)
exp <- tiledbsoma::SOMAExperimentOpen(uri)

obs_df <- as.data.frame(exp$obs$read()$concat())
# Add new column
obs_df$colors <- sample(
  c("red", "blue", "green"),
  size = ncol(pbmc_small),
  replace = TRUE
)

# `as.data.frame.Table()` coerces integer64 to integer, need to cast back
obs_df$soma_joinid <- bit64::as.integer64(obs_df$soma_joinid)

# Reopen in WRITE mode for schema evolution
exp$reopen("WRITE")
exp$obs$update(obs_df) # fails here
exp$reopen("READ")
```

The checks in `TileDBArray$open()` https://github.com/single-cell-data/TileDB-SOMA/blob/65e6bafa808d06ebdb5aa79d7f5dc83e8af422a8/apis/r/R/TileDBArray.R#L29-L33 and `TileDBGroup$open()` https://github.com/single-cell-data/TileDB-SOMA/blob/65e6bafa808d06ebdb5aa79d7f5dc83e8af422a8/apis/r/R/TileDBGroup.R#L62-L70 are what's causing the error. Rather than change the checks, I chose to change `$reopen()` to handle the situation these checks are working on

~Question for reviewers: do we want to cache the timestamp when reopening in `"WRITE"` mode or reset it?~ ~Now changed to parameterize timestamp behavior in `$reopen()`~ Now changed to always reset to current time; re-opening at a custom time can be had with a close-and-open cycle. Adding custom timestamp behavior to `$reopen()` can be evaluated at a later date

### DIFF
--- a/apis/r/tests/testthat/test-SeuratIngest.R
+++ b/apis/r/tests/testthat/test-SeuratIngest.R
@@ -160,7 +160,7 @@ test_that("Write DimReduc mechanics", {
   ))
   on.exit(ms_pca2$close(), add = TRUE, after = FALSE)
   expect_no_condition(write_soma(pbmc_small_tsne, soma_parent = ms))
-  ms$reopen(ms$mode(), force = TRUE)
+  ms$reopen(ms$mode())
   expect_identical(sort(ms$names()), sort(c('X', 'var', 'obsm', 'varm')))
   expect_identical(sort(ms$obsm$names()), sort(paste0('X_', c('pca', 'tsne'))))
   expect_identical(ms$varm$names(), 'PCs')
--- a/apis/r/tests/testthat/test-SeuratIngest.R
+++ b/apis/r/tests/testthat/test-SeuratIngest.R
@@ -4,7 +4,7 @@ test_that("Write Assay mechanics", {
 
   uri <- withr::local_tempdir("write-assay")
   collection <- SOMACollectionCreate(uri)
-  on.exit(collection$close(), add = TRUE)
+  on.exit(collection$close(), add = TRUE, after = FALSE)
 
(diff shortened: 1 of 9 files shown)


eddelbuettel (approved):
Looks good to me.

Maybe add a NEWS item before merging and maybe reference the 'why' a little more in PR (I got briefed over DM which was helpful indeed).

apis/r/tests/testthat/test-SeuratIngest.R
```diff
@@ -26,6 +26,10 @@ test_that("Write Assay mechanics", {
       info = layers[i]
     )
   }
+
+  ms$close()
+  gc()
```
eddelbuettel, line 31: Whether `gc()` helps or not is never clear, but your chances will be marginally better if you first `rm()` the objects holding the memory.  (With sufficient logging you can see finalizers entered upon `gc()` _if_ objects have been removed.)  The trouble with `gc()` is that doesn't not help much on fragmented memory.

Issue #2241, opened 2024-03-08
[python/c++] Correctly write and read back arrays using C++ bindings (parent)


nguyenv:
Parent: #1632

- [x] #2426
- [x] #2427
- [x] #2428
- [x] #2429

johnkerl:
This will be subtasked per @nguyenv , per class (`DataFrame`, `SparseNDArray`, `DenseNDArray`).

johnkerl:
This can be closed when the four PRs here are merged

PR #2498, opened 2024-04-30
[c++,python] Remove tiledb-py for all arrays

nguyenv:
**Issue and/or context:**

#2241

This completes the moveover for `SOMAArray`s to use `clib.SOMAArray` instead of tiledb-py arrays.

**Changes:**

* Add vacuuming and consolidation to C++
* Remove TileDB `ArraySchema` creation
* Rename TileDBArray to `SOMAArray` to reflect that arrays are no longer TileDB-Py arrays but `clib.SOMAArray`


johnkerl (approved):
🚀

PR #2339, opened 2024-03-27
[python/ci] Unbreak wheel builds

johnkerl:
**Issue and/or context:** Attempt on #2337

**Changes:** per @nguyenv 

**Notes for Reviewer:**

Issue #2337, opened 2024-03-26
[python/ci] Unbreak PyPI wheel builds


johnkerl:
See
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/8443323587/job/23126755537
```
FileNotFoundError: [Errno 2] No such file or directory: 'requirements_dev.txt'
```

What happened with 1.9.0 today:

* We only wheel-build on nightlies, or releases
* I tagged a 1.9.0
* I should have tagged a 1.9.0rc0 first :disappointed:
* The `requirements_dev.txt` PR landed before the nightly wheel-build run could happen

#2311 
[sc-43673]

jdblischak:
This error was discovered by the workflow [`python-packaging.yml`](https://github.com/single-cell-data/TileDB-SOMA/blob/36d5b6e7e9657ba9d41e9f71c0b19d0468dc1f79/.github/workflows/python-packaging.yml) and not by [`python-so-copying.yml`](https://github.com/single-cell-data/TileDB-SOMA/blob/36d5b6e7e9657ba9d41e9f71c0b19d0468dc1f79/.github/workflows/python-so-copying.yml) because the former first builds a source tarball (which doesn't contain `requirements_dev.txt`) and then builds the wheel from this source tarball, whereas the latter runs `python setup.py bdist_wheel` directly in the Git repository (where `requirements_dev.txt` exists)

https://github.com/single-cell-data/TileDB-SOMA/blob/36d5b6e7e9657ba9d41e9f71c0b19d0468dc1f79/.github/workflows/python-packaging.yml#L31-L33

Here is a reprex I ran to reproduce the error locally with Python 3.10 on Linux:


```sh
python --version
## Python 3.10.14
uname
## Linux
git checkout 1.9.0
cd apis/python
python setup.py sdist
tar --list -f dist/tiledbsoma-1.9.0.tar.gz | grep requirements

mkdir -p /tmp/soma
cp dist/tiledbsoma-1.9.0.tar.gz /tmp/soma
cd /tmp/soma/
tar -xzf tiledbsoma-1.9.0.tar.gz
cd tiledbsoma-1.9.0/
python -m venv ./venv-soma
conda deactivate
source ./venv-soma/bin/activate
pip install wheel
python setup.py bdist_wheel
## Traceback (most recent call last):
##   File "/tmp/soma/tiledbsoma-1.9.0/setup.py", line 335, in <module>
[...]

PR #2506, opened 2024-05-03
[python/ci] Add CI job to test building wheel from sdist

jdblischak:
**Issue and/or context:**

TileDB-SOMA-Py is built and distributed in many different ways (from source, PyPI wheels, conda binaries). We want to identify any packaging-related problems immediately in the PR that generates them instead of in a nightly build or worse after the release has been made.

**Changes:**

This PR includes 2 main changes:

* The first is rebranding my existing CI pipeline to `python-ci-packaging.yml`. The original name was from my initial PR to test the shared object copying behavior (#1937, #2220). However, I had already previously expanded the pipeline to test the pure setuptools installation setup (which we use in tiledbsoma-feedstock) (#2221). With the addition of a 3rd job (below), I wanted to update the CI to better match its purpose

* The 1.9.0 release was delayed because the PyPI wheels could not be built (#2337, #2339). While we build wheels in many of the existing CI pipelines, none build the wheel from a source tarball (sdist) as is done in [`python-packaging.yml`](https://github.com/single-cell-data/TileDB-SOMA/blob/36d5b6e7e9657ba9d41e9f71c0b19d0468dc1f79/.github/workflows/python-packaging.yml#L31), so this error wasn't caught until post release. With this new CI job, we will immediately detect if a crucial file is absent from the source tarball 

**Notes for Reviewer:**

See the wiki page [Debugging wheel‐build issues](https://github.com/single-cell-data/TileDB-SOMA/wiki/Debugging-wheel%E2%80%90build-issues) for more details

### DIFF
--- a/.github/workflows/python-ci-packaging.yml
+++ b/.github/workflows/python-ci-packaging.yml
@@ -142,6 +142,9 @@ jobs:
           find . -name '*tile*.so*'
           ./venv-soma/bin/python -c "import tiledbsoma; print(tiledbsoma.pytiledbsoma.version())"
 
+  # Confirm shared object copying when building the Python package
+  # https://github.com/single-cell-data/TileDB-SOMA/pull/1937
+  # Same as job above, but running on macOS instead of in a Docker container
   macos:
     runs-on: macos-13
     name: "SO copying (macos) TILEDB_EXISTS: ${{ matrix.TILEDB_EXISTS }} TILEDBSOMA_EXISTS: ${{ matrix.TILEDBSOMA_EXISTS }}"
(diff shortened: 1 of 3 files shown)


johnkerl:
@jdblischak is this partially duplicating https://github.com/single-cell-data/TileDB-SOMA/blob/main/.github/workflows/python-packaging.yml ?

jdblischak:
> is this partially duplicating https://github.com/single-cell-data/TileDB-SOMA/blob/main/.github/workflows/python-packaging.yml?

I wouldn't describe it as duplication. I would describe it more as preparation for the official building of the PyPI wheels.

Differences:

* Uses transparent python commands instead of the opaque action `pypa/cibuildwheel`, and therefore easier to debug
* Only builds two wheels total (python 3.11 on Ubuntu and macOS), so it is much faster. It's not creating the wheels for distribution, but instead just providing a quick check that all the required files are distributed in the source tarball

In summary, the goal is to identify any potential problems with building the PyPI wheels when a PR is submitted, and not at release time.

johnkerl:
@jdblischak can we please get this verbatim into the affected YAML file as a block comment?

> Differences:
> 
> Uses transparent python commands instead of the opaque action pypa/cibuildwheel, and therefore easier to debug
Only builds two wheels total (python 3.11 on Ubuntu and macOS), so it is much faster. It's not creating the wheels for distribution, but instead just providing a quick check that all the required files are distributed in the source tarball
In summary, the goal is to identify any potential problems with building the PyPI wheels when a PR is submitted, and not at release time.

johnkerl (commented):
@jdblischak SGTM as long as CI resource consumption on every PR doesn't become excessive ... main issue being macos runners: choices are 5, 5, 5, or enterprise with no "step up" to 10 or 20 say ...

johnkerl (approved):
@jdblischak SGTM as long as CI resource consumption on every PR doesn't become excessive ... main issue being macos runners: choices are 5, 5, 5, or enterprise with no "step up" to 10 or 20 say ...

jdblischak:
> as long as CI resource consumption on every PR doesn't become excessive

This pipeline does spawn a lot of jobs. Two reasons why I think it avoids being excessive:

1. Aggressive use of path-based filters. This pipeline is only triggered if a file that could affect the Python installation process is affected (eg `setup.py`, `pyproject.toml`, etc). A PR that edits the C++ code, R code, or even the non-installation Python files will never trigger these jobs
2. The macOS jobs all finish in under 10 minutes. So even if it hogs our parallel macOS jobs, it will be for a brief period of time

If it does start to become excessive, the first macOS job I would drop is the one added in this PR, `sdist`. It is mainly to catch files missing from the source tarball, and that behavior should be consistent across Linux and macOS.

johnkerl:
Cool and thanks @jdblischak ! I indeed just looked for `paths:` in your affected YAML and we have protection there against over-running so my concern was premature (I should have checked there first). But thanks for looking, and thanks for the prioritization if/when "something has to go" someday -- really appreciate it! :)

PR #2154, opened 2024-02-20
[r] Fix Seurat ingestion with missing command logs collection

mojaveazure:
Fix a bug in Seurat ingestion where `exp/uns` is present but `exp/uns/seurat_commands` is not

fixes #2153

PR #2507, opened 2024-05-03
[r] Silence missing command logs warning

mojaveazure:
Muffle warnings about missing command logs in `SOMAExperimentAxisQuery$to_seurat()`; warnings can be restored with `options(verbose = TRUE)`

### DIFF
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -6,7 +6,7 @@ Description: Interface for working with 'TileDB'-based Stack of Matrices,
     like those commonly used for single cell data analysis. It is documented at
     <https://github.com/single-cell-data>; a formal specification available is at
     <https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md>.
-Version: 1.10.99.3
+Version: 1.10.99.4
 Authors@R: c(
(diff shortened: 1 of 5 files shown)


aaronwolen (approved):
Thanks!

eddelbuettel:
Do you have an example of the warnings to be muffled?  I run the tests all the time too and don't think I have seen anything...

(Adding a link to the SC story is also always appreciated, saves one step...)

mojaveazure:
[SC story](https://app.shortcut.com/tiledb-inc/story/46694/silence-warnings-for-missing-command-logs) (not sure why the bot didn't automagically add a link)

Warning that's being muffled for those who can't access Shortcut
image

eddelbuettel:
Thanks, and thanks for the link (I still add mine by hand, I know that works).  Is that a version issue? Or maybe .... I never saw this as I may not have run that inject step interactively?   All good now :)

mojaveazure:
Nope, this warning has been around since #2154; it only occurs when trying to read a SOMA that was written with `uns` but without Seurat's command logs. It's a relatively niche case and I was asked to silence the warning. For users who still want it, setting `options(verbose = TRUE)` will re-enable the warning (and this is tested for as part of this PR)

Issue #2517, opened 2024-05-07
[python] "Failed to convert buffer" for nullable attribute on `update_obs` after initial non-null ingest


johnkerl:
Repro of the problem:

* The `batch_id` column of `obs` is all non-null on initial ingest
* It has nulls on `tiledbsoma.io.update_obs`

<details>

```
#!/usr/bin/env python

import tiledbsoma
import tiledbsoma.io

import pandas as pd
import scanpy as sc

import sys, os, shutil
uri = '46654a'
if len(sys.argv) == 2:
    uri = sys.argv[1]
if os.path.exists(uri):
    shutil.rmtree(uri)

# Load example anndata
adata = sc.datasets.pbmc3k()

# pre-process
sc.pp.calculate_qc_metrics(adata, inplace=True)

# Add a string column to obs with strings (no missings)
adata.obs['batch_id'] = ['batch1'] * adata.n_obs

# Initial ingest
tiledbsoma.io.from_anndata(
    experiment_uri=uri,
    anndata=adata,
    measurement_name="RNA"
)

# Replace strings with missings
adata.obs['batch_id'] = pd.NA

# Update on disk
with tiledbsoma.Experiment.open(uri, "w") as exp:
    tiledbsoma.io.update_obs(exp, adata.obs)
```

</details>

Non-problem: the `batch_id` column of `obs` has nulls on initial ingest

<details>

```
#!/usr/bin/env python

import tiledbsoma
import tiledbsoma.io

import pandas as pd
import scanpy as sc

import sys, os, shutil
uri = '46654b'
if len(sys.argv) == 2:
    uri = sys.argv[1]
if os.path.exists(uri):
    shutil.rmtree(uri)

# Load example anndata
adata = sc.datasets.pbmc3k()

# pre-process
sc.pp.calculate_qc_metrics(adata, inplace=True)

# Add a string column to obs with strings (no missings)
adata.obs['batch_id'] = ['batch1'] * adata.n_obs

# Replace strings with missings
adata.obs['batch_id'] = pd.NA

# Initial ingest
tiledbsoma.io.from_anndata(
    experiment_uri=uri,
    anndata=adata,
    measurement_name="RNA"
)

```

</details>

[sc-46654]

PR #2518, opened 2024-05-07
[python] Fix `failed to convert buffer` on `update_obs`/`update_var`

johnkerl:
**Issue and/or context:** #2517 / [sc-46654]

**Changes:** As described in the PR

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/_arrow_types.py
+++ b/apis/python/src/tiledbsoma/_arrow_types.py
@@ -248,6 +248,25 @@ def df_to_arrow(df: pd.DataFrame) -> pa.Table:
                 values=column, categories=categories, ordered=ordered
             )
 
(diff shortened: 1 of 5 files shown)


johnkerl:
Moving back to draft status until I add a unit-test case

johnkerl:
@nguyenv @ryan-williams ready for review 🙏

ryan-williams (approved):

johnkerl:
@nguyenv thoughts?

nguyenv (approved):
Looks good to me. I know this already existed prior to this PR, but I was just wondering it was necessary for `nullable_fields` to be a `set` since it would be more efficient to append to a list than add to an (immutable) set. But I don't think this is something that needs to be changed.

Issue #2511, opened 2024-05-06
[c++] `ColumnBuffer::set_data` needs to handle 32-bit string


nguyenv:
Currently, we only correctly handle 64-bit large string and binary. 

Relates to #2471

PR #2519, opened 2024-05-07
[c++] Handle 32-bit string and binary in writes

nguyenv:
**Issue and/or context:**

#2511

**Changes:**

We only dealt with large string and binary before, which use 64-bit offsets. This also supports small string and binary, which use 32-bit offsets.

### DIFF
--- a/apis/python/src/tiledbsoma/soma_dataframe.cc
+++ b/apis/python/src/tiledbsoma/soma_dataframe.cc
@@ -66,8 +66,7 @@ void load_soma_dataframe(py::module& m) {
                     for (int64_t i = 0; i < schema.n_children; ++i) {
                         auto child = schema.children[i];
                         auto val = metadata.attr("get")(
-                            py::str(child->name)
-                                .attr("encode")("utf-8"));
+                            py::str(child->name).attr("encode")("utf-8"));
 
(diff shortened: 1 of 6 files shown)


eddelbuettel (approved):
This looks pretty good and compact and is probably what we need.

Any chance you can add a quick test for both utf8() and large_utf8() ?   Or too involved in libtiledbsoma's tests?

I could add something from the R side if you want (leaning on nanoarrow and / or the arrow R package).

nguyenv:
Yeah I can add something to test small and large strings in the C++ unit tests in a bit. Did you test this branch against your work to make sure it corrected your issue with 32-bit offsets?

eddelbuettel:
> Did you test this branch against your work to make sure it corrected your issue with 32-bit offsets?

Not yet.

nguyenv:
@eddelbuettel Actually can you help me out by creating the unit test in R as you offered? I just attempted to write one in C++ but then realized I'm unclear how I'd actually set up data buffers of 32-bit vs 64-bit strings.

eddelbuettel:
@nguyenv Maybe I need to defer this til my PR #2534 is landed. I need a little bit of infrastructure from it.

PR #2521, opened 2024-05-07
[python] Clarify `test_update_dataframe.py` and `_arrow_types.py`

johnkerl:
**Issue and/or context:** Pre-PR to facilitate #2518

**Changes:** Just renames and comments -- zero functional changes

**Notes for Reviewer:**

### DIFF
--- a/apis/python/tests/conftest.py
+++ b/apis/python/tests/conftest.py
@@ -13,33 +13,35 @@ from ._util import TESTDATA
 
 
 @pytest.fixture
-def pbmc0_h5ad_path(request) -> Path:
+def conftest_pbmc_small_h5ad_path(request) -> Path:
     """Path to a tiny (80x20) h5ad, useful for unit-test / CI runs."""
     return TESTDATA / "pbmc-small.h5ad"
 
 
 @pytest.fixture
-def pbmc0_adata(pbmc0_h5ad_path) -> AnnData:
+def conftest_pbmc_small(conftest_pbmc_small_h5ad_path) -> AnnData:
     """Tiny (80x20) AnnData, useful for unit-test / CI runs."""
-    return anndata.read_h5ad(pbmc0_h5ad_path)
+    return anndata.read_h5ad(conftest_pbmc_small_h5ad_path)
 
 
 @pytest.fixture
-def pbmc0_exp(pbmc0_h5ad_path) -> Experiment:
+def conftest_pbmc_small_exp(conftest_pbmc_small_h5ad_path) -> Experiment:
     """Ingest an ``AnnData``, yield a ``TestCase`` with the original and new AnnData objects."""
     with TemporaryDirectory() as exp_path:
-        tiledbsoma.io.from_h5ad(exp_path, pbmc0_h5ad_path, measurement_name="RNA")
+        tiledbsoma.io.from_h5ad(
+            exp_path, conftest_pbmc_small_h5ad_path, measurement_name="RNA"
+        )
         with tiledbsoma.Experiment.open(exp_path) as exp:
             yield exp
 
 
 @pytest.fixture
-def pbmc3k_h5ad_path(request) -> Path:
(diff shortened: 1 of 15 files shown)


apis/python/tests/test_basic_anndata_io.py
```diff
@@ -85,17 +85,18 @@ def h5ad_file_X_none(request):
     "X_kind",
     [tiledbsoma.SparseNDArray, tiledbsoma.DenseNDArray],
 )
-def test_import_anndata(adata, ingest_modes, X_kind):
-    original = adata.copy()
-    adata = adata.copy()
+# Magical conftest_adata fixture
+def test_import_anndata(conftest_adata, ingest_modes, X_kind):
```
ryan-williams: I've replaced `conftest_` with more descriptive prefixes in #2529, and struck the `Magical conftest_adata fixture` comments (per reasoning above).

apis/python/tests/test_update_dataframes.py
```diff
@@ -11,81 +11,122 @@
...
-                o1=o1,
-                v1=v1,
+                obs_schema=obs_schema,
+                var_schema=var_schema,
             )
 
 
-def create_member_fixture(name):
-    """Create a ``pytest.fixture`` for a ``TestCase`` field."""
+for field in fields(TestCase):
```
ryan-williams: I think this block was a main culprit making things too magical; unrolled it into 7 features (one for each `TestCase` field) in #2529.

apis/python/tests/conftest.py
```diff
@@ -10,31 +10,31 @@
 
 
 @pytest.fixture
-def h5ad_path(request):
+def conftest_h5ad_path(request):
```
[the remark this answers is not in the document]
johnkerl: > Defining fixtures in conftest.py is a relatively core pytest pattern ([docs](https://docs.pytest.org/en/8.2.x/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files)), and the more unique/descriptive names in https://github.com/single-cell-data/TileDB-SOMA/pull/2529 should make it easy for people to make the connection, in the worst case

> IntelliJ

> I'd rather strike a balance where this pytest "feature" is reasonably discoverable, without scare-{naming,commenting} them at every use-site

Here is where we differ

I'm disinteresting in biasing toward individual toolchains (e.g. IntelliJ). Also since we are multi-lingual it's really crucial that we bias toward less Python expertise. Whatever may seem "standard" to highly evolved Python devs such as yourself -- whom I admire & envy -- there are lots of folks "dipping in" to Python and we need to lower the barrier to entry. I don't find these to be "scare" names -- rather, this is simple inclusivity.

ryan-williams (approved):

apis/python/tests/test_basic_anndata_io.py
johnkerl: > Defining fixtures in conftest.py is a relatively core pytest pattern ([docs](https://docs.pytest.org/en/8.2.x/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files)), and the more unique/descriptive names in https://github.com/single-cell-data/TileDB-SOMA/pull/2529 should make it easy for people to make the connection, in the worst case

> IntelliJ

> I'd rather strike a balance where this pytest "feature" is reasonably discoverable, without scare-{naming,commenting} them at every use-site

@ryan-williams thanks for your perspective. Here is where we differ:

I'm disinterested in biasing toward individual toolchains (e.g. IntelliJ). Also since we are multi-lingual it's really crucial that we bias toward less Python expertise. Whatever may seem "standard" to highly evolved Python devs such as yourself -- whom I admire & envy -- there are lots of folks "dipping in" to Python and we need to lower the barrier to entry. I don't find these to be "scare" names -- rather, this is simple inclusivity.

johnkerl:
Thanks for all the many improvements via #2529 @ryan-williams ! :)

PR #2405, opened 2024-04-08
[r] Port resume-mode to R

mojaveazure:
Implement resume-mode ingestion in the R API

This PR parallels #664; it adds support for resume-mode in factory functions, which allows `SOMA*Create()` to check for an already existing TileDB object at `uri` and if so, simply connect to it rather than try to re-create

It also adds support for resume-mode in `write_soma()` methods; these methods check the `soma_joinids` that are present in the input data and that already exist on disk, and only writes data for `soma_joinids` that are missing from disk

The following SOMA functions have been modified to with an `ingest_mode` parameter:
 - `SOMADataFrameCreate()`
 - `SOMASparseNDarrayCreate()`
 - `SOMACollectionCreate()`
 - `SOMAMeasurementCreate()`
 - `SOMAExperimentCreate()`

The following methods of `write_soma()` have been modified with an `ingest_mode` parameter:
 - `write_soma.character()`/`write_soma.data.frame()`/`write_soma.Dataframe()`
 - `write_soma.matrix()`/`write_soma.TsparseMatrix()`
 - `write_soma.Seurat()` and other Seurat-subobject methods
 - `write_soma.SummarizedExperiment()`/`write_soma.SingleCellExperiment()`

resolves #1399

PR #2523, opened 2024-05-07
[r] Add initial support for ragged array writing for Seurat v5

### LINKED ISSUE #2658: [r] Add support for ragged arrays in `write_soma()`
Seurat v5 includes support for ragged arrays, where not every layer has the same obs and vars. Seurat handles this with character indexing, but as SOMA uses integer indexing, layers will need to be padded to the minimum of `max(axis_soma_joinid), dim(layer)`

[[sc-46644]](https://app.shortcut.com/tiledb-inc/story/46644/add-support-for-ragged-arrays-in-write-soma)

mojaveazure:
Seurat v5 adds support for ragged arrays, where not every `X` layer has exactly the same cells and features. To handle these ragged arrays on ingestion, re-indexing the soma join IDs is necessary to pad the `X` layer to the full domain of the SOMA measurement

Implemented SOMA methods:
- `write_soma.Assay5()`: write a Seurat v5 assay to a SOMA measurement. When writing `X` layers, if a layer is ragged:
  - cast layer to `TsparseMatrix` for COO representation
  - re-index Seurat's character IDs to SOMA join IDs
  - re-index COO coordinates to SOMA join IDs
  - write array using `SOMASparseNDArray$private$.write_coo_dataframe()`

Notes:
 - This PR does not implement alternate matrix (eg. DelayedArray, BPCells) ingestion

[SC-46644](https://app.shortcut.com/tiledb-inc/story/46644/)

resolves #2658


apis/r/R/utils.R
```diff
@@ -110,6 +110,65 @@ uns_hint <- function(type = c('1d', '2d')) {
   }
 }
 
+.type_hint <- function(type) {
+  nm <- 'soma_r_type_hint'
```
johnkerl, line 114: I think asking `nm` to be spelled out `name` is a reasonable ask

apis/r/R/utils.R
```diff
@@ -110,6 +110,65 @@ uns_hint <- function(type = c('1d', '2d')) {
   }
 }
 
+.type_hint <- function(type) {
```
johnkerl, line 113: I know this is internal & doesn't result in any `.Rd` content but it's a a non-trivial function with zero comments

please add at least some doc couments at the top: why is this function here? what is its nominal use? what are one or two sample inputs & outputs?

apis/r/R/write_seurat.R
```diff
@@ -207,6 +208,210 @@ write_soma.Assay <- function(
...
+  on.exit(X$close(), add = TRUE, after = FALSE)
+
+  X$set_metadata(.layer_hint(SeuratObject::DefaultLayer(x)))
+
+  # Pull presence matrices from the v5 assay
+  cmat <- methods::slot(x, name = 'cells')
+  fmat <- methods::slot(x, name = 'features')
+
+  # Write `X` matrices
+  for (lyr in SeuratObject::Layers(x)) {
```
johnkerl, line 320: I think spelling out `lyr` as `layer` is a reasonable ask

also I think it's a `layer_name` not a `layer` -- given `key = lyr` below

apis/r/R/write_seurat.R
```diff
@@ -207,6 +208,210 @@ write_soma.Assay <- function(
...
+    ldat <- SeuratObject::LayerData(x, layer = lyr)
```
johnkerl, line 321: Likewise `layer_data`

apis/r/R/write_seurat.R
```diff
@@ -207,6 +208,210 @@ write_soma.Assay <- function(
...
+  withCallingHandlers(
+    .register_soma_object(X, soma_parent = ms, key = 'X', relative = relative),
+    existingKeyWarning = .maybe_muffle
+  )
+  on.exit(X$close(), add = TRUE, after = FALSE)
+
+  X$set_metadata(.layer_hint(SeuratObject::DefaultLayer(x)))
+
+  # Pull presence matrices from the v5 assay
+  cmat <- methods::slot(x, name = 'cells')
```
johnkerl, line 316:
```suggestion
  cells_matrix <- methods::slot(x, name = 'cells')
```

apis/r/R/write_seurat.R
```diff
@@ -207,6 +208,210 @@ write_soma.Assay <- function(
...
+  fmat <- methods::slot(x, name = 'features')
```
johnkerl, line 317:
```suggestion
  features_matrix <- methods::slot(x, name = 'features')
```

apis/r/R/write_seurat.R
```diff
@@ -207,6 +208,210 @@ write_soma.Assay <- function(
...
+      soma_dim_1 = bit64::as.integer64(jdx[ldat@j + 1L] - 1L),
+      soma_data = ldat@x
+    )
+    shape <- c(max(coo$soma_dim_0), max(coo$soma_dim_1)) + 1L
+    arr <- X$add_new_sparse_ndarray(
+      key = lyr,
+      type = arrow::infer_type(coo$soma_data),
+      shape = as.integer(shape)
+    )
+    arr$.__enclos_env__$private$.write_coo_dataframe(coo)
```
johnkerl, line 372: smells like an antipattern -- is this truly the only R-tastic way to do this? if function `F` is invoking class `C`'s private method -- it isn't really private and we're deceiving ourselves

apis/r/R/write_seurat.R
```diff
@@ -207,6 +208,210 @@ write_soma.Assay <- function(
...
+    arr$set_metadata(.ragged_array_hint())
+    arr$set_metadata(type)
+  }
+
+  # Write feature-level meta data
```
johnkerl, line 377:
```suggestion
  # Write feature-level metadata
```

apis/r/R/write_seurat.R
```diff
@@ -207,6 +208,210 @@ write_soma.Assay <- function(
...
+    x = var_df,
+    uri = 'var',
+    soma_parent = ms,
+    key = 'var',
+    ingest_mode = ingest_mode,
+    platform_config = platform_config,
+    tiledbsoma_ctx = tiledbsoma_ctx
+  )
+
+  # Return
```
johnkerl, line 396: I'm generally in favor of more comments than fewer -- don't get me wrong -- but this particular one doesn't seem helpful

apis/r/R/write_seurat.R
```diff
@@ -207,6 +208,210 @@ write_soma.Assay <- function(
...
+  if (class(x)[1L] != 'Assay5') {
+    warning(
+      paste(
+        strwrap(paste0(
+          "Extended assays (eg. ",
+          class(x)[1L],
+          ") are not fully supported; core Assay data has been written but ",
```
johnkerl, line 403: Is 'core' Seurat terminology here? What should the word 'core' mean to me as a reader of this message when/if I see it appearing at runtime in an R session?

apis/r/R/write_seurat.R
```diff
@@ -541,24 +748,14 @@ write_soma.Seurat <- function(
   )
   on.exit(experiment$close(), add = TRUE, after = FALSE)
 
-  # Write cell-level meta data (obs)
-  spdl::info("Adding cell-level meta data")
+  # Prepare cell-level meta daa (obs)
```
johnkerl, line 751:
```suggestion
  # Prepare cell-level metadata (obs)
```

apis/r/tests/testthat/test-SeuratIngest.R
```diff
@@ -137,6 +137,96 @@ test_that("Write Assay mechanics", {
...
+  expect_equivalent(
+    ms$get_metadata(names(assay_hint)),
+    assay_hint[[1L]]
+  )
+  expect_identical(ms$names(), c("X", "var"))
+  expect_s3_class(ms$var, "SOMADataFrame")
+  expect_identical(setdiff(ms$var$attrnames(), "var_id"), names(rna[[]]))
+  expect_s3_class(ms$X, "SOMACollection")
+  expect_identical(ms$X$names(), SeuratObject::Layers(rna))
+  fmat <- methods::slot(rna, name = "features")
```
johnkerl, line 166: Same as above with spelling out `fmat`, `cmat`, `lyr` -- just a few keystrokes for you -- single-time spend of those seconds! -- with added clarity for every reader forever after

apis/r/R/write_seurat.R
```diff
@@ -207,6 +208,210 @@ write_soma.Assay <- function(
   return(ms)
 }
 
+#' @rdname write_soma_seurat_sub
```
johnkerl, line 211: I'd love to learn why this line is here

johnkerl (changes requested):

apis/r/R/write_seurat.R
mojaveazure: This was needed due to the way the writer worked prior to #3204; now that that PR is in, we can switch to using the newly-exposed `$.write_coordinates()` (working on rebasing this PR against `main` since it's gotten quite stale)

apis/r/R/utils.R
mojaveazure: `nm` is using the R standard name (see [`?setNames`](https://stat.ethz.ch/R-manual/R-devel/library/stats/html/setNames.html))

apis/r/R/write_seurat.R
mojaveazure: "core" refers to the components shared by all Assays; this wording was agreed to last year when the Seurat v3/v4 -> SOMA ingestor was built https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/r/R/write_seurat.R#L195-L199

apis/r/R/write_seurat.R
mojaveazure: This is following the same pattern as used in the v3 Assay ingestor https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/r/R/write_seurat.R#L191; happy to remove if requested

apis/r/R/write_seurat.R
mojaveazure: "meta data" is how Seurat spell this out https://satijalab.github.io/seurat-object/reference/sub-sub-.Assay5.html, and was used previously in the v3 -> SOMA ingestor https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/r/R/write_seurat.R#L172

apis/r/R/write_seurat.R
johnkerl: Thanks, TIL!

apis/r/R/utils.R
johnkerl: I concede to the R world ✅

apis/r/R/write_seurat.R
mojaveazure: Resolved in 5cf3f8d

apis/r/R/utils.R
mojaveazure: Added in 568dd6b

apis/r/R/write_seurat.R
```diff
@@ -207,6 +208,215 @@ write_soma.Assay <- function(
...
+  parents <- unique(sys.parents())
+  idx <- which(vapply_lgl(
+    parents,
+    FUN = function(i) identical(sys.function(i), write_soma.Seurat)
+  ))
+  shape <- if (length(idx) == 1L) {
+    get("shape", envir = sys.frame(parents[idx]))
+  } else {
+    NULL
+  }
```
aaronwolen, line 276: Accessing variables from parent frames seems fragile. Could we just add a `shape` argument?

aaronwolen (commented):
Functionally this is great! I really like the new `_hint()` helpers and the tests appear very comprehensive.

My main ask is to address the significant code duplication between the existing `write_soma.Assay()` method and the new `write_soma.Assay5()` method. Both methods share similar logic for creating measurements, writing X matrices, handling feature-level metadata, etc.

This will make the code more maintainable, easier to update, and less likely to create bugs by making changes in one method but not the other.

apis/r/R/write_seurat.R
mojaveazure: I had thought about that, but I didn't want to expose `shape` in `write_soma()`; this logic is needed for resume-mode #2405, and more specifically for the tests. This functionality isn't used otherwise, which is why I didn't want to expose `shape` for `write_soma()`

apis/r/R/write_seurat.R
```diff
@@ -506,6 +716,8 @@ write_soma.Seurat <- function(
   platform_config = NULL,
   tiledbsoma_ctx = NULL
 ) {
+  op <- options(tiledbsoma.write_soma.internal = TRUE)
```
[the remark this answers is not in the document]
mojaveazure: Added in 21bd333

aaronwolen (approved):
Thanks for the updates!

PR #2530, opened 2024-05-08
[c++] Add `SOMABaseNDArray::shape`

nguyenv:
**Issue and/or context:**

As brought up internally, `SOMADataFrame::shape` should not exist. 

**Changes:**

As `shape` is only viable for the NDArray classes, this PR removes `SOMAArray::shape` and creates a separate base class `SOMABaseNDArray` from which only `SOMASparseNDArray` and `SOMADenseNDArray` can inherit `shape` from.


libtiledbsoma/src/soma/soma_base_ndarray.cc
```diff
@@ -0,0 +1,113 @@
+/**
+ * @file   soma_base_ndarray.cc
+ *
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2022-2023 TileDB, Inc.
```
eddelbuettel, line 8: 2024

eddelbuettel (approved):
This looks good, I have not tested it yet.  One nit over copyright year.

nguyenv:
1. The addition of a `SOMABaseNDArray` subclass is incompatible with the tiledbsoma-r API
2. `shape` has been completely refactored

PR #2531, opened 2024-05-08
[r] Ensure `SOMADataFrame$shape()` throws a not-yet-implemented error

mojaveazure:
New `$shape()` method for `SOMADataFrame` that throws a not-yet-implemented error. This is implemneted as a custom error instead of `.NotYetImplmented()` because:
- `.NotYetImplemented()` does not understand R6 methods and produces a janky error message
```R
Error: `$` is not yet implmeneted`sdf` is not yet implemented`shape()` is not yet implemented
```
- `.NotYetImplemented()` throws a simple error (same as `stop()`) which is harder to catch; I'm custom-classing this error as `notYetImplementedError` so that any `tryCatch()` can catch this and only this

New SOMA methods:
 - `SOMADataFrame$shape()`: throw a `notYetImplementedError`

This parallels the Python's API #2533

### DIFF
--- a/apis/r/NEWS.md
+++ b/apis/r/NEWS.md
@@ -10,7 +10,7 @@
 * Add support for resume-mode in `write_soma()`
 * Push default-setting for `TileDBCreateOptions` to `$initialize()` instead of in the accessors
 * Muffle warnings for missing command logs when outgesting SOMA to `Seurat`
-* Ensure `SOMADataFrame$shape()` always returns `NA`
+* Have `SOMADataFrame$shape()` throw a not-yet-implemented error
 
 # 1.7.0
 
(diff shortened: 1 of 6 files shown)


eddelbuettel (approved):
LGTM

apis/r/NEWS.md
```diff
@@ -10,6 +10,7 @@
 * Add support for resume-mode in `write_soma()`
 * Push default-setting for `TileDBCreateOptions` to `$initialize()` instead of in the accessors
 * Muffle warnings for missing command logs when outgesting SOMA to `Seurat`
+* Ensure `SOMADataFrame$shape()` always returns `NA`
```
eddelbuettel, line 13: That now needs an update as it errors

mojaveazure: Updated in [`fc9e5d7`](https://github.com/single-cell-data/TileDB-SOMA/pull/2531/commits/fc9e5d7ef91445f92a5596d8167ecc24edbb88d5)

eddelbuettel (approved):
:shipit:

PR #2534, opened 2024-05-08
[r] DataFrame write via `libtiledbsoma`

eddelbuettel:
**Issue and/or context:**

Connect DataFrame write to new C++ code

**Changes:**

Added support for new scheme of passing configuration parameters as either an Arrow Table (for domain and extent, passed via `nanoarrow` instances of Arrow's `void*` pointers) as well as via JSON strings for various other options including all filter settings.

**Notes for Reviewer:**

[SC 44885](https://app.shortcut.com/tiledb-inc/story/44885/r-c-rcpp-libtiledbsoma-integration-somadataframe-write-path) #2421


apis/r/tests/testthat/test-write-soma-resume.R
```diff
@@ -106,7 +106,7 @@ test_that("Resume-mode data frames", {
   }
 
   # Expect error when writing to existing array
-  expect_error(write_soma(co2, uri = uri, soma_parent = collection))
+  # FIXME: why no fail?   expect_error(write_soma(co2, uri = uri, soma_parent = collection))
```
mojaveazure, line 110: We should be getting an error here. This test is "error if opening existing without explicitly allowing for existing"; in effect, the following behaviors with SOMADataFrame creation need to be preserved:
- create SOMADataFrame that doesn't exist (should succeed)
- create SOMADataFrame that does exist in resume-mode (should simply return an handle to the existing on-disk group in write-mode)
- create SOMADataFrame that does exist in write-mode (should fail due to existing on-disk group)

mojaveazure (changes requested):
I can't speak to the C++ side, but overall the R-side looks good (I like the JSON stuff for filters). However, the error-on-create-if-existing behavior needs to be preserved. It seems that this PR has changed how that check works. I'm not sure where that's happening but the error-on-create-if-existing behavior should be restored

apis/r/tests/testthat/test-write-soma-resume.R
eddelbuettel: I didn't change anything here.  Could it be that this farms off to collections so that we may be caught between 'half of this approach' and 'half of the other' ?  Any thoughts you may have @nguyenv ?

apis/r/R/TileDBCreateOptions.R
```diff
@@ -31,7 +31,9 @@
   # Filter-related schema parameters
   dataframe_dim_zstd_level = 3,
   sparse_nd_array_dim_zstd_level = 3,
-  offsets_filters = list('DOUBLE_DELTA', 'BIT_WIDTH_REDUCTION', 'ZSTD'),
+  dense_nd_array_dim_zstd_level = 3,
+  offsets_filters = list("DOUBLE_DELTA", "BIT_WIDTH_REDUCTION",
+                         list("name"="ZSTD", "COMPRESSION_LEVEL"=-1)),
```
nguyenv, line 36: Does ZSTD here need the compression level set? I thought without providing a compression level it should just default set this to -1.

apis/r/tests/testthat/test-write-soma-resume.R
eddelbuettel: So 'someone' here needs to do ` stat()` call and see if the to-be-written object exists already.  That could either be on the R side, or (for consistency?) on the C++ side. I wonder if that is something we should handle with an additional option toggle because default behavior, when confronted with an existing file, is to simply overwrite isn't it?

mojaveazure: I believe the default behavior when confronted with an existing file is to error out, and the toggle (resume-mode) handles the alternate case of reopen-in-write. The previous error was bubbling up from (IIRC) `tiledb::tiledb_array_create()`; it wasn't something that SOMA threw explicitly

eddelbuettel: Fixed, I think!  

This was me: I had a well-intentioned but wrong 'if uri exists, recursively delete' in there that did not belong.  Removed in 75e710a1

mojaveazure: Awesome, thank you!

mojaveazure (approved):
Approving from a read of the R code; leaving the C++ code for someone more experienced than I

apis/r/R/TileDBCreateOptions.R
nguyenv: Do you recall what error you were seeing. Because in the C++ API the default for this argument is `R"(["DOUBLE_DELTA", "BIT_WIDTH_REDUCTION", "ZSTD"])"`, so we don't set the compression level there and it works.

apis/r/tests/testthat/test-write-soma-resume.R
eddelbuettel: No worries -- I am glad you put the test in!

nguyenv (approved):
The C++ side looks good to me except for the ZSTD compression level needing to be set. I want to figure out if the bug(?) resides in the C++ code or not.

apis/r/R/TileDBCreateOptions.R
eddelbuettel: I get a segfault, no error message and if memory serves it comes from expecting a second array element that is not coming.

eddelbuettel: For what it is worth it happens (when running all tests) when it comes to `test-reopen.R` so about a quarter through.

eddelbuettel:
The ZSTD compression looks like I tricked myself, possibly by a brief appearance of a Heisenbuglet.  Now cleaned up in 89190bc7.

johnkerl:
Blocked by #2629 per @eddelbuettel and @nguyenv

johnkerl:
#2629 may be of help

eddelbuettel:
Done via #2704 instead.

PR #2535, opened 2024-05-08
[python] Error on registration of multi-modal experiment with distinct var field names

### LINKED ISSUE #2532: [python] Error on registration of multi-modal experiment with distinct var field names
Scenario:

* Ingest an AnnData object with whatever `obs.index.name` and `var.index.name` and a particular measurement name (e.g. RNA modality)
* Have a second AnnData object with (necessarily) the same `obs.index.name` and different `var.index.name` and a different measurement name (e.g. protein modality)
* Registration failure with "Index out of bounds"

Of course, it is not a requirement that the `var` dataframes for two different measurements within the same SOMA experiment have the same registration-dedupe field names.

Full repro available as unit-test code on #2535

[sc-46930]

johnkerl:
**Issue and/or context:** Resolves #2532

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/io/_registration/ambient_label_mappings.py
+++ b/apis/python/src/tiledbsoma/io/_registration/ambient_label_mappings.py
@@ -233,11 +233,13 @@ class ExperimentAmbientLabelMapping:
                     obs_map.update(dict(zip(obs_ids, soma_joinids)))
 
                 for measurement_name in exp.ms:
-                    exp.ms[measurement_name].var
+                    expvar = exp.ms[measurement_name].var
                     if "var" not in exp.ms[measurement_name]:
                         continue
+                    if var_field_name not in expvar.schema.names:
+                        continue
                     var_map = {}
-                    for batch in exp.ms[measurement_name].var.read(
+                    for batch in expvar.read(
                         column_names=["soma_joinid", var_field_name]
                     ):
                         var_ids = [e.as_py() for e in batch[1]]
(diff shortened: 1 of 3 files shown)


apis/python/src/tiledbsoma/io/_registration/ambient_label_mappings.py
```diff
@@ -233,21 +233,23 @@ def from_isolated_soma_experiment(
                     obs_map.update(dict(zip(obs_ids, soma_joinids)))
 
                 for measurement_name in exp.ms:
-                    exp.ms[measurement_name].var
+                    expvar = exp.ms[measurement_name].var
                     if "var" not in exp.ms[measurement_name]:
                         continue
+                    if var_field_name not in expvar.schema.names:
```
johnkerl, line 239: It would be possible to not even consult all measurments -- only the one of interest higher up in the call stack.

However, this isn't neat -- since there is not _one_ of interest. At multiple spots in today's ingestor, we take the current measurement name _and_ "raw". That's a non-ideal implementation. However, to change it -- to require users to now explicitly and separately ingest "raw" -- would be a breaking change to the user-facing API. And perhaps one worth making. But one which would require discussion, design, implmentation, and a deprecation cycle.

apis/python/src/tiledbsoma/io/_registration/ambient_label_mappings.py
```diff
@@ -233,21 +233,23 @@ def from_isolated_soma_experiment(
                     obs_map.update(dict(zip(obs_ids, soma_joinids)))
 
                 for measurement_name in exp.ms:
-                    exp.ms[measurement_name].var
+                    expvar = exp.ms[measurement_name].var
                     if "var" not in exp.ms[measurement_name]:
```
ryan-williams, line 237: Slightly cleaner to reference `exp.ms[measurement_name].var` after the `"var" in exp.ms[measurement_name]` check?

I see that it works either way, but the sequencing looks a little strange.

ryan-williams (approved):
one minor q, otherwise lgtm

apis/python/src/tiledbsoma/io/_registration/ambient_label_mappings.py
johnkerl: Done -- thanks @ryan-williams ! :)

PR #2539, opened 2024-05-09
[r] Suppress Warnings on SOMA -> `Seurat` Outgestion

mojaveazure:
SeuratObject v5.0.2 throws a warning when the layer requested does not exist. During SOMA -> `Seurat` outgestion, this occurs when running `SeuratObject::.CalcN` eg.
```r
> obj <- query$to_seurat(X_layers = c('data' = 'data'))
Warning: Layer counts isn't present in the assay object; returning NULL
```
Since the results of `.CalcN()` should already exist in the SOMA, we can disable this processing and suppress the warning


eddelbuettel (approved):
:shipit:

PR #2540, opened 2024-05-09
[r/ci] Utilize pin of tiledb-r during install

mojaveazure:
Update `r-ci.yml` to respect the doublepin of tiledb-r in `DESCRIPTION`; this is accomplished by doing the following:
- setting environment variable `R_REMOTES_UPGRADE` to `"never"` to prevent upgrading of dependencies in `tools/r-ci.sh install_all`
- setting `tiledb-inc.r-universe.dev` as an additional repo for installing dependencies
- parsing `apis/r/DESCRIPTION` to find constraints on tiledb-r
- searching through all repos to find all accessible versions of tiledb-r
- identifying the version of tiledb-r that matches our constraints
- installing upstream dependencies of tiledb-r from all accessible repos
- installing the correctly-constrained version of tiledb-r

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -62,8 +62,8 @@ jobs:
         run: |
           rversion <- paste(strsplit(as.character(getRversion()), split = '\\.')[[1L]][1:2], collapse = '.')
           codename <-  system('. /etc/os-release; echo ${VERSION_CODENAME}', intern = TRUE)
-          repo <- sprintf("https://tiledb-inc.r-universe.dev/bin/linux/%s/%s/", codename, rversion)
-          (opt <- sprintf('options(repos = c("%s", getOption("repos")))', repo))
+          repo <- "https://tiledb-inc.r-universe.dev"
+          (opt <- sprintf('options(repos = c("%s/bin/linux/%s/%s", "%s", getOption("repos")))', repo, codename, rversion, repo))
           cat(opt, "\n", file = "~/.Rprofile", append = TRUE)
         shell: Rscript {0}
 
@@ -140,7 +140,7 @@ jobs:
       #  run: cd apis/r && Rscript -e "options(bspm.version.check=TRUE); install.packages('tiledb', repos = c('https://eddelbuettel.r-universe.dev/bin/linux/jammy/4.3/', 'https://cloud.r-project.org'))"
 
       - name: Dependencies
-        run: R_REMOTES_UPGRADE=never cd apis/r && tools/r-ci.sh install_all
+        run: cd apis/r && Rscript -e "remotes::install_deps(dependencies = TRUE, upgrade = FALSE)"
 
       # - name: Install dataset packages from source (macOS)
       #   if: ${{ matrix.os == 'macOS-latest' }}
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -68,47 +68,7 @@ jobs:
         shell: Rscript {0}
 
       - name: Install tiledb-r
-        run: |
-          options(bspm.version.check = TRUE)
-          # Find tiledb-r constraints
-          deps <- read.dcf('apis/r/DESCRIPTION', fields = 'Imports')[1, ]
-          deps <- trimws(strsplit(gsub(pattern = ',', replacement = '\n', x = deps), split = '\n')[[1L]])
-          deps <- Filter(nzchar, deps)
-          deps <- Filter(function(x) grepl('^tiledb', x), deps)
-          deps <- gsub(pattern = '[[:space:]]', replacement = '', x = deps)
-          deps <- gsub(pattern = '.*\\(|\\)', replacement = '', x = deps)
-          const <- data.frame(comp = gsub(pattern = '[[:digit:]\\.]*', replacement = '', x = deps))
-          const$vers <- gsub(pattern = paste(const$comp, collapse = '|'), replacement = '', x = deps)
-          const
-          # Find correct version of tiledb-r
-          db <- utils::available.packages(filters = c("R_version", "OS_type", "subarch"))
-          idx <- which(rownames(db) == 'tiledb')
-          valid <- vapply(
-            idx,
-            FUN = function(i) {
-              v <- db[i, 'Version']
-              res <- vector(mode = 'logical', length = nrow(const))
-              for (i in seq_along(res)) {
-                res[i] <- do.call(const$comp[i], args = list(v, const$vers[i]))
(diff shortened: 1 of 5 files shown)


johnkerl (approved):

.github/workflows/r-ci.yml
```diff
@@ -52,6 +53,63 @@ jobs:
...
+        run: echo 'options(repos = c("https://tiledb-inc.r-universe.dev", getOption("repos")))' | tee -a ~/.Rprofile
+
+      - name: Set additional repositories (Linux)
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+        run: |
+          rversion <- paste(strsplit(as.character(getRversion()), split = '\\.')[[1L]][1:2], collapse = '.')
+          codename <-  system('. /etc/os-release; echo ${VERSION_CODENAME}', intern = TRUE)
+          repo <- "https://tiledb-inc.r-universe.dev"
+          (opt <- sprintf('options(repos = c("%s/bin/linux/%s/%s", "%s", getOption("repos")))', repo, codename, rversion, repo))
+          cat(opt, "\n", file = "~/.Rprofile", append = TRUE)
```
eddelbuettel, line 67: This won't work, as I just discovered.  r-universe already jumped to noble whereas ubuntu-latest will be jammy for some time.  

image

So we are once again between a rock and a hard place.  Once r2u jumps to 0.27.0 (by default tomorrow we're in trouble and the above won't save us.  :disappointed:

eddelbuettel, line 67: This won't work, as I just discovered.  r-universe already jumped to noble whereas ubuntu-latest will be jammy for some time.  

image

So we are once again between a rock and a hard place.  Once r2u jumps to 0.27.0 (by default tomorrow we're in trouble and the above won't save us.  :disappointed:

.github/workflows/r-ci.yml
```diff
@@ -52,6 +53,63 @@ jobs:
...
+          db <- db[-idx[!valid], , drop = FALSE]
+          # Install upstream deps
+          ups <- tools::package_dependencies("tiledb", db = db, recursive = TRUE)$tiledb
+          utils::install.packages(intersect(ups, rownames(db)))
+          # Install correct version of tiledb-r
+          ctb <- contrib.url(getOption("repos"))
+          names(ctb) <- getOption("repos")
+          dbr <- db[idx[valid], 'Repository']
+          repos <- names(ctb[ctb %in% dbr])
+          utils::install.packages("tiledb", repos = repos)
```
eddelbuettel, line 110: This is IMHO hard to read (inside YAML) and impossible to test.  Can we make it standalone script?

eddelbuettel, line 110: This is IMHO hard to read (inside YAML) and impossible to test.  Can we make it standalone script?

Also the install step here (before dependencies) may pull in a few other things at this step.

.github/workflows/r-ci.yml
mojaveazure: That's why I've added the source-only repo as an additional repo https://github.com/single-cell-data/TileDB-SOMA/actions/runs/9022913531/job/24793545830?pr=2540#step:8:1; if possible, CI prefers a binary install but if none matches the constrains, we have the source to fallback on
image

eddelbuettel: I fear you may need to also deal with `bspm` to re-enable source and deal with the sorting.  I am sure you can sort that out tomorrow when the need arises...

eddelbuettel: And there is simply no linux/jammy/4.4 -- I wish there were. (You can hit eg the URL https://tiledb-inc.r-universe.dev/bin/linux/noble/4.4/src/contrib/PACKAGES in the browser -- this one corresponds to one of the builds from my screenshot above -- but other combos such as jammy/4.4 failed for me and likely do not exist.

We're still between a rock and a hard place.  So maybe starting tomorrow source builds for tiledb-r until SOMA flips to the newer one.  We'll survive...

mojaveazure: When doing the actual install, I do the `options(bspm.version.check = TRUE)` to allow bspm to search elsewhere https://github.com/single-cell-data/TileDB-SOMA/pull/2540/files#diff-5331055c5808c739bfd30c63013fd35ce1e28e3edef1df4476550617d393467cR72; in addition, I tell `install.packages()` to use the specific repository for tiledb-r that has the valid version https://github.com/single-cell-data/TileDB-SOMA/pull/2540/files#diff-5331055c5808c739bfd30c63013fd35ce1e28e3edef1df4476550617d393467cR109-R110

I'm happy to hold off merging this until tomorrow when r2u gets 0.27.0 and re-run CI to test

.github/workflows/r-ci.yml
mojaveazure: I've created three install steps
1. install upstream hard dependencies of tiledb-r https://github.com/single-cell-data/TileDB-SOMA/pull/2540/files#diff-5331055c5808c739bfd30c63013fd35ce1e28e3edef1df4476550617d393467cR102-R104
2. install tiledb-r from the specified repository https://github.com/single-cell-data/TileDB-SOMA/pull/2540/files#diff-5331055c5808c739bfd30c63013fd35ce1e28e3edef1df4476550617d393467cR110
3. install dependencies of tiledbsoma-r with an explicit `upgrade = "never"` to prevent tiledb-r and its dependencies from being upgraded https://github.com/single-cell-data/TileDB-SOMA/pull/2540/files#diff-5331055c5808c739bfd30c63013fd35ce1e28e3edef1df4476550617d393467cR143

I can move the "Install tiledb-r" step into a standalone script

mojaveazure: Resolving in favor of https://github.com/single-cell-data/TileDB-SOMA/pull/2540#pullrequestreview-2048782610

.github/workflows/r-ci.yml
eddelbuettel: Merge as you see fit, this is mostly a heads-up that I would not be surprised to see new and different breakage tomorrow.

Which can be dealt with tomorrow, I suppose.

.github/workflows/r-ci.yml
mojaveazure: Moved the installation steps into `apis/r/tools/install-tiledb-r.R` in [`88bafc4`](https://github.com/single-cell-data/TileDB-SOMA/pull/2540/commits/88bafc4213d624be8a2e6ff0f3cf3d2d253a77cc)

mojaveazure:
[sc-48525]

Issue #2520, opened 2024-05-07
[c++] `PlatformConfig` tile/cell order and filter options should be case insensitive 


nguyenv:
Currently the C++ `PlatformConfig` only handles options as put forth by the SOMA Python API. Some of the options in the SOMA R API differ slightly and need to be added in C++.

- Tile/cell order and filter options should be case insensitive
- Add all name to filter mappings (ie. "rle" and "rlefilter" should both map to `tiledb::Filter::RleFilter`)
- Handle global and unordered ordering

Relates to https://github.com/single-cell-data/TileDB-SOMA/issues/2471

PR #2541, opened 2024-05-09
[c++] `PlatformConfig` refactoring

nguyenv:
**Issue and/or context:**

#2520

**Changes:**

* Add `column-major` as valid column ordering value
* Add missing `dense_nd_array_dim_zstd_level` option
* `tiledb_schema_from_arrow_schema` needs to take in additional `std::string soma_type` argument to ensure `PlatformConfig` parameters `dataframe_dim_zstd_level`, `dense_nd_array_dim_zstd_level`, and `sparse_nd_array_dim_zstd_level` get applied correctly
* Use the default `PlatformConfig` constructor if one has not been provided, not `std::nullopt`. This is why the default levels were not getting set correctly
* Apply default filter zstd (at TileDB default level) to attributes
* Filter names coming in from R API need to be respected
* `PlatformConfig` should be representable in JSON format; options for `FilterList`s should take in JSON nested arrays:
```
{
    "attr_name": {
         "filters": ["XOR", {"name": name="GZIP", COMPRESSION_LEVEL=3]
    }
}
```
* Need to set `FilterList` for `attrs` and `dims`
* Break up `tiledb_schema_from_arrow_schema` into more readable, separate functions
* Test passing values to `PlatformConfig` in pytest, ensuring defaults get overridden

### DIFF
--- a/libtiledbsoma/src/utils/arrow_adapter.cc
+++ b/libtiledbsoma/src/utils/arrow_adapter.cc
@@ -301,7 +301,9 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
 
         if (tile_order == "row-order" or tile_order == "row") {
             schema.set_tile_order(TILEDB_ROW_MAJOR);
-        } else if (tile_order == "col-order" or tile_order == "col") {
+        } else if (
+            tile_order == "col-order" or cell_order == "column-major" or
+            tile_order == "col") {
             schema.set_tile_order(TILEDB_COL_MAJOR);
         } else {
             throw TileDBSOMAError(fmt::format(
--- a/libtiledbsoma/src/utils/arrow_adapter.cc
+++ b/libtiledbsoma/src/utils/arrow_adapter.cc
@@ -233,7 +233,7 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
 
     ArraySchema schema(*ctx, is_sparse ? TILEDB_SPARSE : TILEDB_DENSE);
     Domain domain(*ctx);
-    
+
     std::map<std::string, tiledb_filter_type_t> convert_filter = {
         {"GZIP", TILEDB_FILTER_GZIP},
         {"ZSTD", TILEDB_FILTER_ZSTD},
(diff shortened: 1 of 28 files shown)


libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -350,7 +350,13 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
                     dim_zstd_filter.set_option(
                         TILEDB_COMPRESSION_LEVEL,
                         platform_config.dataframe_dim_zstd_level);
-                } else if (soma_type == "SOMASparseNDArray") {
+                } else if (
+                    soma_type == "SOMASparseNDArray" ||
+                    soma_type == "SOMADenseNDArray") {
+                    // sparse_nd_array_dim_zstd_level being used for
```
johnkerl, line 356: Sorry @nguyenv , no -- this is not at all what I meant earlier.

* What I thought you meant was that we could/should have a single `platform_config.dim_zstd_level` which would be used for dataframe, sparse nd array, and dense nd array
  * And my answer was: no, we need separate configs per datatype
* What I was advocating for (however inarticulate I may have succeeded in being, despite my best efforts) was:
  * that we _do_ need separate:
    * `platform_config.dataframe_dim_zstd_level`
    * `platform_config.sparse_nd_array_dim_zstd_level`
    * `platform_config.dense_nd_array_dim_zstd_level`
  * that our not having had a `platform_config.dense_nd_array_dim_zstd_level` was an oversight that needs to be remedied

nguyenv: OK understood - I can change this.

johnkerl:
@nguyenv  I took the liberty of merging from latest `main` at https://github.com/single-cell-data/TileDB-SOMA/pull/2541/commits/42ea7e86fa7c8c2e2c2d4a4597e2ffbbf8caffd5 in order to bring in #2540 for CI ... heads up for your git-pulling information :)

eddelbuettel (approved):
It's a large-ish PR and still a little short on tests (outside the Python side I can't really focus on).  I think this deserves a qualified thumbs-up but it may be best to wait for another review or two.

Issue #2537, opened 2024-05-09
[c++] Error writing shuffled numpy data (regression) 


jp-dark:
When writing shuffled/transposed numpy data that has been shuffled, the data is written in the order of the underlying numpy data instead of the "expected" shuffled order. This is a regression that was introduced when dense writes were pushed down to C++.

**To Reproduce**
```
data = np.reshape(np.arange(24, dtype=np.uint8), (4, 3, 2))
data= np.transpose(data, (2,0, 1))
array = soma.DenseNDArray.create(
    uri,
    type=pa.from_numpy_dtype(data.dtype),
    shape=data.shape,
)
tensor = pa.Tensor.from_numpy(data)
array.write(
    (slice(None), slice(None), slice(None)), tensor
)
```

Printing `data`:
```
[[[ 0  2  4]
  [ 6  8 10]
  [12 14 16]
  [18 20 22]]

 [[ 1  3  5]
  [ 7  9 11]
  [13 15 17]
  [19 21 23]]]
```

Reading back data from `array` and printing:
```
[[[ 0  1  2]
  [ 3  4  5]
  [ 6  7  8]
  [ 9 10 11]]

 [[12 13 14]
  [15 16 17]
  [18 19 20]
  [21 22 23]]]
```

PR #2542, opened 2024-05-09
[python] Fix write from array that is not row or column order

jp-dark:
**Issue and/or context:** #2537

**Changes:**
When writing a dense array, this change will check if the data is neither row major nor column major and explicitly set it to be row-major before writing.


jp-dark:
Closes #2537

nguyenv (approved):
Awesome! Thanks so much for looking at this.

PR #2546, opened 2024-05-10
[r/ci] Add check to disable BSPM in R CI

mojaveazure:
Follow-up to #2540 

bspm does not respect the `repos` argument in `install.packages()`; this is an issue where the valid version of tiledb-r is on R-universe, but not on CRAN. In this situation, bspm will still install from APT rather than the repository specified in `install.pacakges(repos = )`

To get around this, I've added a check that asks
 - is bspm installed
 - is the valid tiledb-r avaialble only on alternate repos

If both of those are true, disable bspm for installing tiledb-r; bspm is still used for upstream dependencies of tiledb-r

Note, this has the unfortunate side-effect of installing tiledb-r from source on Linux CI. This is because Linux CI is currently using 22.04 jammy while [R-universe is now building binaries for 24.04 noble](https://tiledb-inc.r-universe.dev/tiledb). As seen in Docker using the Dockerfile [`doublepin22.Dockerfile`](https://gist.github.com/mojaveazure/c1e08930d9ee8717e02b9cce66ab5bf0#file-doublepin22-dockerfile), we build tiledb-r from source ([`jammy.log`](https://gist.github.com/mojaveazure/c1e08930d9ee8717e02b9cce66ab5bf0#file-jammy-log)); however, whenever Linux CI gets updated to 24.04 noble ([`doublepin24.Dockerfile`](https://gist.github.com/mojaveazure/c1e08930d9ee8717e02b9cce66ab5bf0#file-doublepin24-dockerfile)), we will regain binary installation of tiledb-r thanks to R-universe ([`noble.log`](https://gist.github.com/mojaveazure/c1e08930d9ee8717e02b9cce66ab5bf0#file-noble-log))

### DIFF
--- a/apis/r/tools/install-tiledb-r.R
+++ b/apis/r/tools/install-tiledb-r.R
@@ -42,4 +42,13 @@ ctb <- contrib.url(getOption("repos"))
 names(ctb) <- getOption("repos")
 dbr <- db[idx[valid], 'Repository']
 (repos <- names(ctb[ctb %in% dbr]))
+
+# BSPM doesn't respect `repos`
+# Check to see if any repo w/ valid tiledb-r is CRAN
+# If not, turn of BSPM
+cran <- getOption("repos")['CRAN']
+cran[is.na(cran)] <- ""
+if (requireNamespace("bspm", quietly = TRUE) && !any(dbr %in% cran)) {
+  bspm::disable()
+}
 utils::install.packages("tiledb", repos = repos)
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -73,7 +73,6 @@ jobs:
       - name: Install BioConductor package SingleCellExperiment
         run: cd apis/r && tools/r-ci.sh install_bioc SingleCellExperiment
 
-
       # Uncomment these next two stanzas as needed whenever we've just released a new tiledb-r for
       # which source is available but CRAN releases (and hence update r2u binaries) are not yet:
       #
--- a/.github/workflows/r-python-interop-testing.yml
+++ b/.github/workflows/r-python-interop-testing.yml
@@ -32,8 +32,21 @@ jobs:
       - name: Bootstrap
         run: cd apis/r && tools/r-ci.sh bootstrap
 
+      - name: Set additional repositories (Linux)
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+        run: |
+          rversion <- paste(strsplit(as.character(getRversion()), split = '\\.')[[1L]][1:2], collapse = '.')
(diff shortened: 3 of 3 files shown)


apis/r/tools/install-tiledb-r.R
```diff
@@ -42,4 +42,13 @@ ctb <- contrib.url(getOption("repos"))
 names(ctb) <- getOption("repos")
 dbr <- db[idx[valid], 'Repository']
 (repos <- names(ctb[ctb %in% dbr]))
+
+# BSPM doesn't respect `repos`
```
eddelbuettel, line 46: As a long-time user of bspm and partial contributor to it I think that is not a correct way of stating it.

After all bspm (and its cousin rspm using the p3m.dev binaries not packaged for apt, dnf, ...) is designed to use apt and dnf and alike.  It's simply a different approach.   I am fairly certain bspm does available.packages and hence repos but then has to reconcile different 'sets' of packages.  There are some options that control its behavior, and its author added more.

apis/r/tools/install-tiledb-r.R
```diff
@@ -42,4 +42,13 @@ ctb <- contrib.url(getOption("repos"))
...
+# Check to see if any repo w/ valid tiledb-r is CRAN
+# If not, turn of BSPM
+cran <- getOption("repos")['CRAN']
+cran[is.na(cran)] <- ""
+if (requireNamespace("bspm", quietly = TRUE) && !any(dbr %in% cran)) {
+  bspm::disable()
```
eddelbuettel, line 52: What is 'dbr' here?

mojaveazure: `dbr` is a vector of repositories that a valid version of tiledb-r is present in. In this case, it's `https://tiledb-inc.r-universe.dev`

See log line at https://github.com/single-cell-data/TileDB-SOMA/actions/runs/9035919602/job/24831666413?pr=2546#step:9:169

apis/r/tools/install-tiledb-r.R
mojaveazure: If there's an option to tell bspm to use r-universe of APT, then I'm all for using that. However, with bspm enabled and passing repos that doesn't contain CRAN à la
```
install.packages("tiledb", repos = "https://tiledb-inc.r-universe.dev")
```
We get tiledb-r from APT, not from R-universe

johnkerl (approved):
LGTM as long as CI is passing

apis/r/tools/install-tiledb-r.R
eddelbuettel: I am kinda bored by this because we go over all of this every three or four weeks.  Do whatever you need to do it is all fine by me.

PR #2548, opened 2024-05-10
Use core 2.23.0, TileDB-Py 0.29.0, TileDB-R 0.27.0

johnkerl:
**Issue and/or context:** [sc-46600]

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/.github/workflows/python-ci-packaging.yml
+++ b/.github/workflows/python-ci-packaging.yml
@@ -74,7 +74,7 @@ jobs:
         run: |
           mkdir -p external
           # Please do not edit manually -- let scripts/update-tiledb-version.py update this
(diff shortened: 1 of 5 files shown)


nguyenv (approved):

eddelbuettel (approved):
The same sha1-version information is copied to 15 places. It really would be cleaner (and likely also more robust) to store in (ideally just) one place and use that.

johnkerl:
> The same sha1-version information is copied to 15 places. It really would be cleaner (and likely also more robust) to store in (ideally just) one place and use that.

@eddelbuettel we have discussed this before. `scripts/update-tilebs-version.py 2.23.0` automates this process. I do releases and I'm 100% delighted with the ease of use.

PR #2549, opened 2024-05-13
[r] Two-sided R pin for `release-1.10` branch; backport #2540, #2546, #2549

johnkerl:
**Issue and/or context:** Following our [established procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases), I am working on tiledbsoma 1.11 today

[sc-46600]

**Changes:**

**Notes for Reviewer:**

@mojaveazure I am following https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases#incrementing-core-versions

However, since that was updated we do have #2540 and #2546 on `main`; I'm not sure what's the right thing to do here on the `release-1.10` branch, as I am going to create a `release-1.11` branch today -- @mojaveazure let's work together on this.

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -16,6 +16,7 @@ env:
   COVERAGE_TOKEN: ${{ secrets.CODECOV_TOKEN }}
   _R_CHECK_TESTS_NLINES_: 0
   CATCHSEGV: "TRUE"
+  R_REMOTES_UPGRADE: "never"
 
 jobs:
   ci:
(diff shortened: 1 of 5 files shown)


mojaveazure:
If tiledbsoma 1.10 uses core 2.22/tiledb-r 0.26, then in addition to the following change we'll also need #2540 and #2546 backported to `release-1.10`

[As shown previously](https://github.com/single-cell-data/TileDB-SOMA/pull/2374#issuecomment-2060236286), the doublepin does not affect installation or library loading, only `R CMD check`; as such, you need the two aforementioned PRs to get CI to install the correct version of tiledb-r

.github/workflows/r-ci.yml
```diff
@@ -82,7 +99,7 @@ jobs:
       #  run: cd apis/r && Rscript -e "options(bspm.version.check=TRUE); install.packages('tiledb', repos = c('https://eddelbuettel.r-universe.dev/bin/linux/jammy/4.3/', 'https://cloud.r-project.org'))"
 
       - name: Dependencies
-        run: cd apis/r && tools/r-ci.sh install_all
+        run: cd apis/r && Rscript -e "remotes::install_deps(dependencies = TRUE, upgrade = FALSE)"
```
johnkerl, line 102: @mojaveazure should this be
```
remotes::install_deps(".", dependencies = TRUE, upgrade = FALSE)"
```
?

mojaveazure: Not necessarily, the `"."` is the default https://remotes.r-lib.org/reference/install_deps.html

mojaveazure (approved):

Issue #2547, opened 2024-05-10
[Bug] `pip install` fails in shallow clones (GHA, RTD); unable to generate SCM version


ryan-williams:
Local installs end up with `setuptools_scm`-generated versions like `1.5.0rc0.post332.dev6296020074`, because [1.5.0](https://github.com/single-cell-data/TileDB-SOMA/tree/1.5.0) is the most recent tag that's an ancestor of `main` (subsequent releases have all been on release branches). However, shallow clones (e.g. in GHAs and RTD builds) don't have enough context to figure that out, and fail.

This is currently failing non-release RTD builds ([readthedocs.org/projects/tiledbsoma/builds](https://readthedocs.org/projects/tiledbsoma/builds/)):

<img src="https://github.com/single-cell-data/TileDB-SOMA/assets/465045/a5b1a3a5-8691-4c2f-8901-2b9e420fd1d8" width="60%" />

The current [most recent failure](https://readthedocs.org/projects/tiledbsoma/builds/24345280/) there shows `python -m pip install -v apis/python` failing with:

```
    File "/home/docs/checkouts/readthedocs.org/user_builds/tiledbsoma/checkouts/latest/apis/python/version.py", line 147, in getVersion
      raise ValueError("Cannot find the version number")
  ValueError: Cannot find the version number
  error: subprocess-exited-with-error
```

I've typically worked around this (in test GHA workflows) with a line like:
```bash
echo "1.11.0rc0" > apis/python/RELEASE-VERSION
```

but we should standardize on a solution.

jdblischak:
For GHA workflows, I thought we had standardized on passing [`fetch-depth: 0`](https://github.com/search?q=repo%3Asingle-cell-data%2FTileDB-SOMA%20fetch-depth&type=code) to `actions/checkout` to perform a unshallow clone.

From a quick search, [here](https://docs.readthedocs.io/en/latest/build-customization.html#unshallow-git-clone) is how to perform an unshallow clone on RTD.

ryan-williams:
[RTD builds](https://readthedocs.org/projects/tiledbsoma/builds/) are passing, and I've subscribed to email alerts for when any future failures.

As of #2563, we should also be getting sensible version numbers during development (in both shallow and full clones).

PR #2563, opened 2024-05-15
[python] Prototype: better version numbers during development

ryan-williams:
## Issue and/or context: #2547, extends #2560

[`1.5.0`](https://github.com/single-cell-data/TileDB-SOMA/tree/1.5.0) (Nov '23) is typically the most recent tag that's an ancestor of `main`, because subsequent release tags all exist on release branches.

This means local installs get versions like `1.5.0rc0.post332.dev6296020074`, adapted from `git describe` (the `dev` number is the Git short SHA, `git log -1 --format=%h`, converted to base 10, for [PEP440](https://peps.python.org/pep-0440/)-compliance).

It's a little weird to see `1.5.0*` on dev installs, and it crashes in shallow clones (which GHA / RTD use by default).

### 1. Fallback version number based on most recent release tag (e.g. `1.11.1.post0.dev67466703006`)

If `git describe` returns `1.5.0` as the nearest tagged ancestor, synthesize a more meaningful version number:

1. Find the latest release tag in the local repo (or tracked remote, if there are no local tags, e.g. in case of a shallow clone).
2. Return a version like `A.B.C.post0.devN`:
    - `A.B.C` is the most recent release tag in the repo, and
    - `post0` is intended to signal that we are "ahead" of `A.B.C`, though the number of commits is not computed/specified.
      - We are also "behind" by an unknown number of commits, but it's not important to specify either number in the version number.
      - Something that roughly tracks what the "next" SemVer will be, and encodes the Git SHA, seems better than the status quo.
    - `N` is the current short Git SHA, converted to base 10 (as above).

### 2. "[CalVer](https://calver.org/)" fallback (e.g. `2024.05.14.dev67466703006`)

The first commit in this PR adds an additional "CalVer"  fallback (suggested by @jp-dark 🙏):

```python
def generate_cal_version():
    today = date.today().strftime("%Y.%m.%d")
    return f"{today}.dev{get_sha_base10()}"
```

#### CalVer currently unused
Fallback **1.** above should always find the most recent release tag (in the upstream remote, worst case) so this CalVer is never expected to be used, as written, but I've left it in for discussion.

We could merge it, to keep the option available, or "stash" it for later.

### 3. Preview/Check version number: `python apis/python/version.py`
I added an `__name__ == '__main__'` hook to make it easy to interact with this code without running a full `pip install -e .`.

I also cleaned up a few things (e.g. use `open` as a ContextManager, convert camelCase names to snake_case).

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -8,6 +8,5 @@ build:
   tools:
     python: "3.11"
   commands:
-    - git fetch --unshallow || true
     - pip install --upgrade pip
     - doc/build.sh -r -V  # install deps, don't make a venv
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -339,5 +339,5 @@ setuptools.setup(
     },
     python_requires=">=3.8",
     cmdclass={"build_ext": build_ext, "bdist_wheel": bdist_wheel},
-    version=version.getVersion(),
+    version=version.get_version(),
 )
--- a/apis/python/version.py
+++ b/apis/python/version.py
@@ -71,60 +71,136 @@ __all__ = "getVersion"
 
 import os
 import re
-import subprocess
+import shlex
 import sys
 from datetime import date
 from os.path import basename
-from subprocess import CalledProcessError, check_output
+from subprocess import DEVNULL, CalledProcessError, check_output
+from typing import List, Optional
 
 RELEASE_VERSION_FILE = os.path.join(os.path.dirname(__file__), "RELEASE-VERSION")
 
 # http://www.python.org/dev/peps/pep-0386/
 _PEP386_SHORT_VERSION_RE = r"\d+(?:\.\d+)+(?:(?:[abc]|rc)\d+(?:\.\d+)*)?"
 _PEP386_VERSION_RE = r"^%s(?:\.post\d+)?(?:\.dev\d+)?$" % _PEP386_SHORT_VERSION_RE
-_GIT_DESCRIPTION_RE = r"^(?P<ver>%s)-(?P<commits>\d+)-g(?P<sha>[\da-f]+)$" % (
-    _PEP386_SHORT_VERSION_RE
+_GIT_DESCRIPTION_RE = (
+    r"^(?P<ver>%s)-(?P<commits>\d+)-g(?P<sha>[\da-f]+)$" % _PEP386_SHORT_VERSION_RE
 )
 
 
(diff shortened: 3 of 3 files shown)


ryan-williams:
I've also removed the `fetch --unshallow` from `.readthedocs.yml`, which is no longer necessary.

[Here's](https://readthedocs.org/projects/tiledbsoma/builds/24388201/) a passing RTD build, [corresponding docsite](https://tiledbsoma.readthedocs.io/en/rw-ver/) built from this branch.

jdblischak (approved):
I always wondered why the version in my nightly builds started with 1.5.0. Now I know why!

jdblischak:
I'm fine with this PR, but I can't pass up the opportunity to propose a radical simplification of the versioning. I know Python developers like to automate the version bumps (either via setuptools-scm or a custom script like we do here). But this automation is not without its own headaches and maintenance requirements. Getting setuptools-scm to behave in our conda feedstock builds is a pain, and with this PR `version.py` has grown to over 200 lines.

Manually bumping the version at each release isn't that terrible. When you go to `git tag` a release, also bump the version in `setup.py` and `DESCRIPTION`. Or you could even create a super simple script like `bump-version.sh X.X.X` that bumps the version to X.X.X in `setup.py` and `DESCRIPTION` and then runs `git tag -a X.X.X`.

I know this proposal is unlikely to be adopted, but I feel compelled to point out this alternative.

jdblischak:
> What version would you suggest dev builds get right now, when the most recent release tag is [1.11.1](https://github.com/single-cell-data/TileDB-SOMA/tree/1.11.1)?

It depends. What function do we need the dev versions to serve? Do we expect end users to ever see them? Do developers ever actually refer to the dev version, or instead to the exact Git sha? Are dev versions required for RTD to function properly?

For personal and smaller projects (1-4 developers), I tend to manually bump a 4th dev version, so it would be `1.11.1.1`, `1.11.1.2`, etc., until release time when it would be bumped to either `1.12.0` or `1.11.2`. But in a higher-velocity repository like this with many developers, that can get cumbersome (eg having to resolve merge conflicts due to version numbers if someone else's PR is merged before yours).

A simple option would be to add a big number like `9999` after each release to indicate it is a dev version. Then no one touches it until the next release.

And if no one is using the dev version as a diagnostic, we could just leave the version untouched between releases.

johnkerl (approved):

Issue #2054, opened 2024-01-23
[c++] Plumb global-order-write concat through TileDB-SOMA


johnkerl:
Per @gspowley and @KiterLuc there is an option in the C++ core API such that when global-order writes are done, fragments from multiple chunk-writes may be extended/grown rather than resulting in multiple fragments.

The ask is to plumb this through `libtiledbsoma` once we have the write path fully in C++ (@nguyenv).

The context is for CZI's census writer (@bkmartinjr): for them, there are a lot of fragments, resulting in time-consuming consolidation, and this problem can be avoided from the get-go by our simply creating fewer fragments in the first place.

PR #2564, opened 2024-05-15
[c++/python] Enable optional global-order writes for sparse arrays

nguyenv:
**Issue and/or context:** #2054

The C++-ification of the SOMA API now allows us to utilize global writes when writing to sparse arrays, thus saving time by skipping the coordinate sorting step when the coordinates are already known to be ordered.

**Changes:**

- Add `sort_coords` setting in the `PlatformConfig`, where by default, it is set to `False` and thus still uses `TILEDB_UNORDERED`
- Add `sort_coords` parameter to `SOMAArray::write` and `ManagedQuery::submit_write`

### DIFF
--- a/apis/python/src/tiledbsoma/_sparse_nd_array.py
+++ b/apis/python/src/tiledbsoma/_sparse_nd_array.py
@@ -9,6 +9,7 @@ Implementation of SOMA SparseNDArray.
 from __future__ import annotations
 
 import itertools
+import warnings
 from typing import (
     Dict,
     Optional,
@@ -45,7 +46,7 @@ from .options._soma_tiledb_context import (
     SOMATileDBContext,
     _validate_soma_tiledb_context,
 )
-from .options._tiledb_create_options import TileDBCreateOptions
+from .options._tiledb_create_options import TileDBCreateOptions, TileDBWriteOptions
 
 _UNBATCHED = options.BatchSize()
 
@@ -269,9 +270,17 @@ class SparseNDArray(NDArray, somacore.SparseNDArray):
             Experimental.
         """
 
-        tiledb_create_options = TileDBCreateOptions.from_platform_config(
-            platform_config
-        )
+        write_options: Union[TileDBCreateOptions, TileDBWriteOptions]
+        sort_coords = None
+        if isinstance(platform_config, TileDBCreateOptions):
+            warnings.warn(
+                "The write parameter now takes in TileDBWriteOptions instead of TileDBCreateOptions",
+                DeprecationWarning,
+            )
+            write_options = TileDBCreateOptions.from_platform_config(platform_config)
+        else:
+            write_options = TileDBWriteOptions.from_platform_config(platform_config)
+            sort_coords = write_options.sort_coords
 
         clib_sparse_array = self._handle._handle
 
@@ -289,7 +298,7 @@ class SparseNDArray(NDArray, somacore.SparseNDArray):
                 np.array(
                     data, dtype=self.schema.field("soma_data").type.to_pandas_dtype()
                 ),
-                tiledb_create_options.sort_coords,
+                sort_coords or True,
             )
 
             # Write bounding-box metadata. Note COO can be N-dimensional.
@@ -297,7 +306,7 @@ class SparseNDArray(NDArray, somacore.SparseNDArray):
             bounding_box = self._compute_bounding_box_metadata(maxes)
             self._set_bounding_box_metadata(bounding_box)
 
-            if tiledb_create_options.consolidate_and_vacuum:
+            if write_options.consolidate_and_vacuum:
                 # Consolidate non-bulk data
                 clib_sparse_array.consolidate_and_vacuum()
             return self
@@ -321,7 +330,7 @@ class SparseNDArray(NDArray, somacore.SparseNDArray):
                 np.array(
                     sp.data, dtype=self.schema.field("soma_data").type.to_pandas_dtype()
                 ),
-                tiledb_create_options.sort_coords,
+                sort_coords or True,
             )
 
             # Write bounding-box metadata. Note CSR and CSC are necessarily 2-dimensional.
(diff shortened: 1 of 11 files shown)


eddelbuettel (approved):
:+1: for the non-Python side of things

johnkerl (changes requested):
@nguyenv this is a great step forward! Global-order writes are awesome.

But the spelling is a bit off. I went to give this two test drives: one with writing to local disk; the other with writing to a cloud URI. I didn't do the second yet but realized I have some feedback for you after doing the first.

Namely, I checked out this branch & put some prints in to make sure I was getting `sort_coords` when I asked for them.

Using a performance script we've been looking at this week
https://gist.github.com/johnkerl/d5002b994caccdfb059efba081f28bb5
I put in a `"sort_coords": False` line into the `platform_config`.

Then I noticed that `sort_coords` was _not_ being used in the code, even though I'd included it in my `platform_config` in the above script.

Then I realized that I needed to now pass `platform_config` to `write`. So far so good.

However, our current nesting is like

```
platform_config =  {
    "tiledb": {
        "create": {
            "key": "value",
            ...
        },
    }
}
```

and then we use `TileDBCreateOptions.from_platform_config(platform_config)` to extract out the `create_options = platform_config["tiledb"]["create"]` subdict and turn it into more of a struct-accessible form so devs can do `create_options.key`: for example:

https://github.com/single-cell-data/TileDB-SOMA/blob/1.11.1/apis/python/src/tiledbsoma/_sparse_nd_array.py#L131

What confused me -- and is confusing -- is that I had to (a) put `sort_coords` into my `create` subkey, and (b) pass that to `write`.

What we should do is:

* Keep `TileDBCreateOptions` for pulling create-time parameters out of `platform_config["tiledb"]["create"]`
* Make a new `TileDBWriteOptions` for pulling write-time parameters out of `platform_config["tiledb"]["write"]`
* Users need to be instructed how to set their `platform_config["tiledb"]["write"]["sort_coords"] = False`

Other that that, I like this! It's a step forward in that it's still 'off by default' and we can subsequently look into connecting this automatically for the times it's correct to automatically do so (CSR and CSC inputs) and not when it's not (`pa.Table` input to `SparseNDArray`'s `write`)

nguyenv:
Thanks for the feedback @johnkerl. I'll definitely make changes so that this goes into `write` rather than `create`, and it's more clear to the user how to set this up in the platform config.

johnkerl:
Also: I think we should consider the `sort_coords` to be an _override_ on what we do

* For CSR/CSC -- _unless_ they explicitly say `sort_coords=True` then we should not sort the coords -- since we can guarantee ordering
* For `pa.Table`/COO -- then we should sort the coords _unless_ they assert to us the coords are pre-sorted

apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -269,9 +270,17 @@ def write(
...
+        sort_coords = None
+        if isinstance(platform_config, TileDBCreateOptions):
+            warnings.warn(
+                "The write parameter now takes in TileDBWriteOptions instead of TileDBCreateOptions",
+                DeprecationWarning,
+            )
+            write_options = TileDBCreateOptions.from_platform_config(platform_config)
+        else:
+            write_options = TileDBWriteOptions.from_platform_config(platform_config)
+            sort_coords = write_options.sort_coords
```
nguyenv, line 283: @johnkerl Need your feedback here. The `write` function has been taking in `TileDBCreateOptions` so that the user can set the `consolidate_and_vacuum` flag. This parameter is only used for writes so would be appropriate to also move it into `TileDBWriteOptions`. I've added this `DeprecationWarning` here for now so that it can still take in `TileDBCreateOptions` without breaking pre-exsting code while also allowing the new `TileDBWriteOptions`. I am not sure if this is what I am supposed to do though.

johnkerl: Yes @nguyenv that's perfect, and having the consolidate-and-vacuum in create options always was an oversight.

Please fill a tracking issue in GH reminding us to remove consolidate-and-vacuum from create options in, say, tiledbsoma 1.13.

nguyenv: #2579

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -279,8 +279,7 @@ void ManagedQuery::submit_write(bool sort_coords) {
         query_->set_layout(
             sort_coords ? TILEDB_UNORDERED : TILEDB_GLOBAL_ORDER);
     }
-    query_->submit();
-    query_->finalize();
+    query_->submit_and_finalize();
```
johnkerl, line 282: Only if not `sort_coords` IIUC

https://github.com/TileDB-Inc/TileDB/blob/2.23.0/tiledb/sm/query/query.cc#L449-L453

nguyenv:
Tested writing to cloud, and it is working. (I had to change the `soma_init_buffer_bytes` to something smaller, so that's why the write times look faster).

```
(soma-3.11) vivian@mangonada:~/tiledb-bugs$ python sparse-perf-degradation.py
BP.PY INPUT_PATH  norm4
BP.PY OUTPUT_PATH tiledb://vivian/s3://tiledb-vivian/gow
BP.PY COMPRESSION_LEVEL 3
BP.PY BUFFER_SIZE 2097152
BP.PY BATCH_OUT   None

BP.PY OPENING  norm4
BP.PY OPENED   norm4 SECONDS = 0.013
BP.PY CREATING tiledb://vivian/s3://tiledb-vivian/gow
BP.PY CREATED  tiledb://vivian/s3://tiledb-vivian/gow SECONDS = 4.385
BP.PY READING  norm4 batch 0
BP.PY READ     norm4 batch 0 SECONDS = 0.064
BP.PY WRITING  tiledb://vivian/s3://tiledb-vivian/gow batch 0
BP.PY WROTE    tiledb://vivian/s3://tiledb-vivian/gow batch 0 SECONDS = 1.823
BP.PY READ     norm4 batch 1 SECONDS = 0.012
BP.PY WRITING  tiledb://vivian/s3://tiledb-vivian/gow batch 1
BP.PY WROTE    tiledb://vivian/s3://tiledb-vivian/gow batch 1 SECONDS = 1.909
BP.PY READ     norm4 batch 2 SECONDS = 0.010
BP.PY WRITING  tiledb://vivian/s3://tiledb-vivian/gow batch 2
BP.PY WROTE    tiledb://vivian/s3://tiledb-vivian/gow batch 2 SECONDS = 1.891
BP.PY READ     norm4 batch 3 SECONDS = 0.011
```

johnkerl (approved):
Awesome, thanks!!!!

Issue #2569, opened 2024-05-16
Ingesting Multiple H5ad Projects into a single SOMA Object


danishzmalik:
This is regarding a query we have related to the ingestion of h5ad files to SOMA experiments. 
We have data from around 400 in-house Single Cell Studies in h5ad format, stored in an s3 bucket.
Our end goal is to create an application where a user can select a study from a list of of our own studies and have its visualization rendered. 

For this we have been trying to use tiledb.soma python package on databricks. 
We have been successful in converting individual h5ads to the SOMA format, however in an attempt to ingest 2 h5ad files to the same SOMA experiment, we faced an issue where  
`experiment.obs.read().concat().to_pandas()` returned data from only the first file that was ingested. 
                                                                                                                                                              
The following function was used for the ingestion:

```
tiledbsoma.io.from_h5ad(
experiment_uri="/tmp/soma_object",
measurement_name="RNA",
input_path = H5AD_URI,
ingest_mode= “resume”

)
```
Where the ingest mode was set to ‘resume’.

Is there a way to achieve this? And are there any pre-reqs that we need to consider? E.g. Same columns (obs) in every file

johnkerl:
Hi @danishzmalik ! Indeed, we have an "append mode" designed and well-suited for this very purpose. @ryan-williams is taking some notebook example material and formatting it for more broad use -- we'll be in touch soon!

johnkerl:
Related: https://github.com/chanzuckerberg/cellxgene-census/issues/1132

johnkerl:
Note for @ryan-williams: 
https://github.com/single-cell-data/TileDB-SOMA/pull/2571#pullrequestreview-2061916226

ryan-williams:
Hi @danishzmalik, [looks like](https://github.com/single-cell-data/TileDB-SOMA/issues/2619) you found the tutorial notebook added above, just wanted to make sure you caught "part 2" ("[Ingesting multiple datasets to a SOMA Experiment](https://tiledbsoma.readthedocs.io/en/stable/notebooks/tutorial_soma_append_mode.html#Ingesting-multiple-datasets-to-a-SOMA-Experiment)", #2582), which demonstrates ingesting multiple h5ads from scratch (as opposed to adding single h5ads onto an existing SOMA).

I believe this is done, but feel free to comment/reopen if you have further q's/issues.

PR #2571, opened 2024-05-16
Tutorial: TileDB-SOMA append-mode

ryan-williams:
**Issue and/or context:** #2569

**Changes:**
- Add "Tutorial: TileDB-SOMA append-mode" notebook (created by @johnkerl a while back, mistakenly not published)
- RTD automatically picks it up; preview: https://tiledbsoma.readthedocs.io/en/rw-apptut/notebooks/tutorial_soma_append_mode.html

### DIFF
--- a/apis/python/notebooks/tutorial_soma_append_mode.ipynb
+++ b/apis/python/notebooks/tutorial_soma_append_mode.ipynb
@@ -1,0 +1,699 @@
+{
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "tags": []
+      },
+      "source": [
+        "# Tutorial: TileDB-SOMA append-mode"
+      ]
+    },
+    {
+      "cell_type": "markdown",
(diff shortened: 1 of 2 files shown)


johnkerl:
Reviewer-friendly link FYI:
https://github.com/single-cell-data/TileDB-SOMA/blob/rw/apptut/apis/python/notebooks/tutorial_soma_append_mode.ipynb

johnkerl (approved):
Thanks @ryan-williams , this is awesome!

As discussed in Slack, and totally OK to have on a separate PR:

* This is the "Monday/Tuesday" model: ingest one, register another, ingest that other
* The same append-mode API also supports an "atlas model": register multiples, then ingest them in a loop (or in parallel)

We should have either a separate section in this notebook, or a parallel notebook (preferably the former I think) demonstrating the atlas-mode use case for append mode.

aaronwolen (approved):

PR #2573, opened 2024-05-16
[r] Fix builds against current TileDB Embedded 

johnkerl:
**Issue and/or context:** It never made sense to set row/cell order in array-schema creation. But https://github.com/TileDB-Inc/TileDB/pull/4973 recently formalized this.

We shouldn't ask for unordered at schema-create time.

See also: https://github.com/TileDB-Inc/TileDB-Py/pull/1970

Note the above core change isn't in 2.23, only in `dev` -- but https://github.com/TileDB-Inc/tiledbsoma-feedstock has nightlies which point at core dev: https://github.com/jdblischak/centralized-tiledb-nightlies/issues/6#issuecomment-2115841783

**Changes:**

**Notes for Reviewer:**


jdblischak (approved):
xref: https://github.com/jdblischak/centralized-tiledb-nightlies/issues/6#issuecomment-2115841783

eddelbuettel:
Suggested subject change above.  It also broke tests here were I tend to be on current TileDB Embedded as the nightlies are -- but the need goes beyond the nightlies to all build with 'dev' versions of TileDB Embedded and hence likely next versions.

PR #2575, opened 2024-05-16
[r] Switch 'UNORDERED' cell order to 'ROW_MAJOR' on schema creation

eddelbuettel:
**Issue and/or context:**

"UNORDERED" is no longer a permitted cell order value on schema creation.

**Changes:**

This follows #2573 and applies a similar change to the sparse and dense array cases.

**Notes for Reviewer:**

[SC 47577](https://app.shortcut.com/tiledb-inc/story/47577/r-no-more-unordered-cell-order)


eddelbuettel:
I had failures here (against TileDB Embedded at 93dab0ad0 from earlier today) so I suspect the nightly may need this too.

johnkerl (approved):

Issue #2513, opened 2024-05-06
[c++] Eliminate unused-variable and deprecated-API warnings in `libtiledbsoma` compile


johnkerl:
Goal: A clean C++ compile

PR #2581, opened 2024-05-17
[c++/python] Address C++ warnings in Python build

jp-dark:
**Issue and/or context:** #2513

**Changes:**
1. Add `TILEDB_REMOVE_DEPRECATIONS` as a CMake option that can be passed to the external TileDB build.
2. Update CI to build TileDB without the deprecated API.
3. Address a few misc. warnings in PyBind. 
   * Use `is` instead of `==` for PyBind type checks.
   * Use `TileDBSOMAPyError` instead of `TileDBSOMAError` when generating errors in `query_conditions.cc` (remove redefinition).
   * Remove unused variables.


jp-dark:
Note: This change removes the TileDB deprecation warnings from some but not most CI. When TileDB is built downstream (e.g. as part of TileDB-Py), then the deprecations are still built into the API.

johnkerl (approved):
🚢

Issue #2588, opened 2024-05-20
[Bug] Python client no longer able to use version from file `RELEASE-VERSION`


jdblischak:
PR https://github.com/single-cell-data/TileDB-SOMA/pull/2563 was merged Thursday. It updated `version.py` to improve the auto-generated versions for the Python package. However, this seems to have broken the package build workflow that relies on obtaining the version from the file `RELEASE-VERSION` instead of the Git repository.

The nightly jobs that build wheels from an sdist (which has `RELEASE-VERSION` bundled with it) are failing because `version.py` is running a Git command which fails in this context with no Git repository

xref: https://github.com/single-cell-data/TileDB-SOMA/issues/2578#issuecomment-2117781606, https://github.com/single-cell-data/TileDB-SOMA/issues/2583, https://github.com/single-cell-data/TileDB-SOMA/issues/2584, https://github.com/single-cell-data/TileDB-SOMA/issues/2585

```python
subprocess.CalledProcessError: Command '('git', 'tag', '--list', '--sort=v:refname', '[0-9].*.*')' returned non-zero exit status 128.
```

note to self: this scheduled job runs at 5:44 AM each day, which is why the job passed on Thursday and the first failure was Friday morning

The nightly tiledbsoma-feedstock builds are failing with a similar albeit different error:

xref: https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/153#issuecomment-2120605388

```python
    File "${SRC_DIR}/apis/python/version.py", line 124, in get_latest_remote_tag
      raise RuntimeError(f"No tags found in remote {remote}")
  RuntimeError: No tags found in remote origin
```

The conda recipe build script creates `RELEASE-VERSION` using the version specified in the recipe ([source](https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/f537d2dbc8d7e195dac350578fd8b0ca031cd67a/recipe/build-tiledbsoma-py.sh#L7)): `echo "$PKG_VERSION" >> RELEASE-VERSION`

Frustratingly my packaging CI job didn't catch this. Even though I build from an sdist, I unpacked it within the Git repository, so `version.py` was still able to successfully run Git commands. I'll update the pipeline to be more rigorous.

ryan-williams:
[...]

PR #2589, opened 2024-05-20
[python/ci] Build wheel from sdist outside of Git repository

jdblischak:
**Issue and/or context:**

The recent update to `version.py` broke package builds that rely on `RELEASE-VERSION` outside the context of a Git repo

xref: #2588, https://github.com/single-cell-data/TileDB-SOMA/pull/2563, https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/153#issuecomment-2120605388

**Changes:**

This is exactly the type of build error that my packaging CI workflow is supposed to catch. I've updated it so that it now extracts the sdist outside of the Git repo that was cloned, which should catch this type of error in the future

**Notes for Reviewer:**

The packaging CI is expected to fail. I will rebase it after we merge the fix from @ryan-williams

### DIFF
--- a/.github/workflows/python-ci-packaging.yml
+++ b/.github/workflows/python-ci-packaging.yml
@@ -10,6 +10,7 @@ on:
       - 'apis/python/MANIFEST.in'
       - 'apis/python/pyproject.toml'
       - 'apis/python/setup.py'
+      - 'apis/python/version.py'
       - 'apis/python/src/tiledbsoma/__init__.py'
       - 'libtiledbsoma/cmake/inputs/Config.cmake.in'
       - 'libtiledbsoma/cmake/inputs/tiledbsoma.pc.in'
@@ -23,6 +24,7 @@ on:
       - 'apis/python/MANIFEST.in'
       - 'apis/python/pyproject.toml'
       - 'apis/python/setup.py'
+      - 'apis/python/version.py'
       - 'apis/python/src/tiledbsoma/__init__.py'
       - 'libtiledbsoma/cmake/inputs/Config.cmake.in'
       - 'libtiledbsoma/cmake/inputs/tiledbsoma.pc.in'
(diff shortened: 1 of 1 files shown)


jdblischak:
As expected, the "Wheel from sdist" jobs failed with:

```python
subprocess.CalledProcessError: Command '('git', 'tag', '--list', '--sort=v:refname', '[0-9].*.*')' returned non-zero exit status 128.
```

jdblischak:
Potential fix from @ryan-williams in #2590. I manually triggered the branch [rw/ver](https://github.com/single-cell-data/TileDB-SOMA/tree/rw/ver) to run the packaging CI workflow

https://github.com/single-cell-data/TileDB-SOMA/actions/runs/9161462896

(in the future this CI will automatically be triggered when `version.py` is edited)

.github/workflows/python-ci-packaging.yml
```diff
@@ -351,6 +353,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
+          path: TileDB-SOMA
           fetch-depth: 0 # for setuptools-scm
```
ryan-williams, line 357: Just noting, `fetch-depth: 0` shouldn't be necessary anymore (since #2563). Fine to leave it in as well.

ryan-williams (approved):

.github/workflows/python-ci-packaging.yml
jdblischak: I agree it isn't strictly necessary. But I want to keep this test job as close to as possible to the upstream official packaging job that it is designed to emulate

https://github.com/single-cell-data/TileDB-SOMA/blob/6108f733e1d616f20a285a92a8695f124a795d44/.github/workflows/python-packaging.yml#L27-L30

PR #2597, opened 2024-05-20
[c++] Correct `platform_config` to `tiledb_config`

nguyenv:
**Issue and/or context:**

The argument to a `SOMAContext` is a `tiledb::Config` object (as represent by a `map<string, string>`), not a `PlatformConfig` object. `PlatformConfig` is for configuring `SOMAArray`s with settings that cannot be represented by an `ArrowSchema`.

**Changes:**

- Rename argument name `platform_config` to `tiledb_config`
- Add explicit check to ensure passed config settings are honored in `SOMAContext`


eddelbuettel (approved):
:+1:

PR #2599, opened 2024-05-20
[c++] Also map "NONE" to `TILEDB_FILTER_NONE`

nguyenv:
**Changes:**

Allow both "NONE" and "NOOP" as acceptable strings to map to `TILEDB_FILTER_NONE`.


eddelbuettel (approved):
:shipit: 

And thanks for filling that in.

PR #2641, opened 2024-05-24
[python/ci] Bump up cibuildwheel version, use macos-14, and use `MACOSX_DEPLOYMENT_TARGET`

dudoslav:


.github/workflows/python-packaging.yml
```diff
@@ -58,7 +58,7 @@ jobs:
             cibw_archs_macos: x86_64
             wheel-name: macos-x86_64
           - build: macosx_arm64
-            os: macos-12
+            os: macos-14
```
ihnorton, line 61: We should add a note inline that this image is _only_ arm64

PR #2643, opened 2024-05-24
[python] Include guidance for older CPU architectures

### LINKED ISSUE #2580: Develop guidance for support for older `linux-x86` processor variants
From time to time we have questions like this:
https://github.com/chanzuckerberg/cellxgene-census/issues/1121

Issues involved:

* TileDB Core, TileDB-Py, TileDB-R, and TileDB-SOMA (Python and R) have precompiled binaries for OS the major architecture:
  * `linux-x86` fully supported
  * `macos-x86` fully supported
  * `macos-arm64` fully supported
  * `linux-aarch64` support is forthcoming: #2482 
* Issue here is support for processor _variants_ within `linux-x86`
  * Symptom is `Illegal instruction (core dumped)` ([see here for example](2482))
  * The binaries we have pre-compiled for include certain instruction sets
  * Sometimes users run in environments not containing support for some newer instructions
 
Work to do:

* We need to decide, and document, which processor variants we support -- and "support" means:
  * Necessarily people will need to build from source -- since our precompiled binaries aren't right for older variants
  * People will need clear build-from-source instructions
* We need to decide, and document, which processor variants we don't support -- and that means:
  * Listing out processor variants for which we don't have a suggested recipe, and for which we don't intend to develop one

[[sc-47640]](https://app.shortcut.com/tiledb-inc/story/47640/develop-guidance-for-support-for-older-linux-x86-processor-variants)

johnkerl:
Resolves #2580.

### DIFF
--- a/apis/python/README.md
+++ b/apis/python/README.md
@@ -26,6 +26,8 @@ To update to the latest version:
 $ python -m pip install --upgrade tiledbsoma
 ```
 
+In case of `illegal instruction` errors when running on older architectures --- e.g. Opteron, non-AVX2 --- the issue is that the pre-compiled binaries available at Conda or PyPI aren't targeted for all processor variants over time. You can install from source, as shown below.
+
 ## From source
 
 * This requires [`tiledb`](https://github.com/TileDB-Inc/TileDB-Py) (see [./setup.cfg](setup.cfg) for version), in addition to other dependencies in [setup.cfg](./setup.cfg).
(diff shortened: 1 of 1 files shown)


aaronwolen (approved):

eddelbuettel:
Suggestion:  Maybe we should a hint there or on the wiki that helps identify the issue. In the case at hand here as well as in the earlier case it was that `grep avx2 /proc/cpuinfo`  came up empty as the machines were in fact pre-AVX2 and that happens to be a CMake check we have in Core.

Issue #2644, opened 2024-05-24
[ c++] Segfault in some contexts as of 1.11.2


johnkerl:
CZI reports a segfault in `cellxgene_census` test
```
pytest -v -rP --durations=20 --experimental ./api/python/cellxgene_census/tests/experimental
```
on TileDB-SOMA 1.11.2, and no segfault on 1.11.1.

This has been bisected to #2572 which is the `release-1.11` backport of #2541.

[[sc-48124]](https://app.shortcut.com/tiledb-inc/story/48124/segfault-in-some-contexts-as-of-1-11-2)

PR #2645, opened 2024-05-24
[c++] Pass in a `shared_ptr<Context>` to `_create_dim_aux`

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2644

* Derefencing and passing a `Context` object into `_create_dim_aux` prematurely decrements a reference count to the `Context` which then garbage collects the object. When we attempt to reference the `shared_ptr<Context>` again, it is invalid and segfaults
* Fixes intermittent bug in Release mode that is consistently reproducible in Debug mode

### DIFF
--- a/libtiledbsoma/src/utils/arrow_adapter.cc
+++ b/libtiledbsoma/src/utils/arrow_adapter.cc
@@ -413,25 +413,25 @@ Dimension ArrowAdapter::_create_dim(
             return Dimension::create(
                 *ctx, name, type, (uint64_t*)buff, (uint64_t*)buff + 2);
         case TILEDB_INT8:
-            return ArrowAdapter::_create_dim_aux(*ctx, name, (int8_t*)buff);
+            return ArrowAdapter::_create_dim_aux(ctx, name, (int8_t*)buff);
         case TILEDB_UINT8:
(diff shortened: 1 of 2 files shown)


libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -413,25 +413,25 @@ Dimension ArrowAdapter::_create_dim(
             return Dimension::create(
                 *ctx, name, type, (uint64_t*)buff, (uint64_t*)buff + 2);
```
eddelbuettel, line 414: What about lines 414 (here) and 404 (just above) ?

eddelbuettel: I.e. I see that they are different (Dimension objects as opposed to ArrowAdapter method calls -- but why?

nguyenv: nvm. As mentioned over DM but repeating here for anyone else reviewing, the datetime types need to be explicitly passed. Replacing with `_create_dim_aux<uint64_t>` would incorrectly create these datetime dims as uint64 dims.

johnkerl (approved):

PR #2629, opened 2024-05-22
[c++] Cast extended enumeration values and boolean types on writes

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2628

**Changes:**

When writing to the array, handle casting enumeration values and indexes when extending. Also handle bit to uint8_t conversion when writing Boolean columns.

PR #2650, opened 2024-05-29
[c++] Fix validity buffer bit unpacking

nguyenv:
**Issue and/or context:**

This bug was caught when C++-ifying the SOMA R API. https://github.com/single-cell-data/TileDB-SOMA/pull/2629 originally contains the bug fix but has now separated out into this new PR.

**Changes:**

* In Arrow, the validity buffer is stored as bitpacked buffers whereas in TileDB, they are stored as uint8_t buffers. We need to properly unpack the bits and cast them to uint8_t on writes
* Although we have unit tests in the SOMA Python API for nullable attributes in several areas, they somehow happened to write and read back fine so this issue went uncaught. This PR adds a new unit test that fails currently on main but is fixed by this PR

### DIFF
--- a/apis/python/tests/test_dataframe.py
+++ b/apis/python/tests/test_dataframe.py
@@ -1437,12 +1437,23 @@ def test_enum_schema_report(tmp_path):
 def test_nullable(tmp_path):
     uri = tmp_path.as_posix()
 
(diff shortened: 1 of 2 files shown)


nguyenv:
[sc-48000]

eddelbuettel (approved):
This passses the one-off test I have here on a round-trip with NA i.e. "missing" in three different columns.

PR #2681, opened 2024-06-04
[c++] Improve default buffer sizes

johnkerl:
**Issue and/or context:** As on #2682: switch the default from small to medium. The small/CI regime is preserved via opt-in environment variable.

See also https://github.com/single-cell-data/TileDB-SOMA/pull/531 wherein I found that even 512MB was too big for Windows CI -- which is why on this PR I have explicit CI logic.

### DIFF
--- a/.github/workflows/python-ci-single.yml
+++ b/.github/workflows/python-ci-single.yml
@@ -123,7 +123,7 @@ jobs:
     - name: Run libtiledbsoma unit tests
       run: ctest --output-on-failure --test-dir build/libtiledbsoma -C Release --verbose
       env:
-        CI_MEMORY_USAGE: true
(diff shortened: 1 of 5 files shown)


nguyenv (approved):

eddelbuettel (commented):
I think this PR takes the wrong approach.  It adds a small added cost of env var check _to every allocation of a column_.  That is not great design especially that we _already have a variable_.  It think it would be simper to just set it to larger valuue (1gb seems fine) and at the same time ensure our three CI files each set a smaller value.  Less complexity, reuse of existing capabilities.

eddelbuettel:
Thanks for making the change. I think you still need to alter the default in 

https://github.com/single-cell-data/TileDB-SOMA/blob/c6bea4d2fd4f08849c992a70597e4b30dd568d56/libtiledbsoma/src/soma/column_buffer.h#L55

Issue #2684, opened 2024-06-05
[r] Examine dependency of `SOMAArrayReader-Iterated` test on buffer sizes


johnkerl:
Follow-up from https://github.com/jdblischak/centralized-tiledb-nightlies/issues/11#issuecomment-2150323719 after yesterday's #2681

Also: we have this guidance
https://github.com/single-cell-data/TileDB-SOMA/blob/1.11.4/apis/r/vignettes/soma-experiment-queries.Rmd#L84
which says to use `soma.init_buffer_bytes` but doesn't show _how_. This needs to be addressed as well.

Details will surface on #2685.

PR #2685, opened 2024-06-05
[r] Fix dependency of a test on fixed buffer sizes

johnkerl:
Addresses #2684 .

* In our repo CI we have `export TILEDB_SOMA_INIT_BUFFER_BYTES=16777216`
* In the nightlies https://github.com/jdblischak/centralized-tiledb-nightlies/issues/11#issuecomment-2150323719 we do not
* On my laptop I do not
* If I set `export TILEDB_SOMA_INIT_BUFFER_BYTES=16777216` then the `tests/testthat/test-SOMAArrayReader-Iterated.R"` tests pass
* We need to explicitly set `soma.init_buffer_bytes` for those few cases which depend on it.

### DIFF
--- a/apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
+++ b/apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
@@ -94,10 +94,14 @@ test_that("Iterated Interface from SOMA Classes", {
     ## parameterize test
     test_cases <- c("data.frame", "sparse")
 
+    # The read_complete et al. in this test case are designed to be verified
+    # against 16MB buffer size, and the particular provided input dataset.
+    ctx <- SOMATileDBContext$new(c(soma.init_buffer_bytes='16777216'))
+
     for (tc in test_cases) {
         sdf <- switch(tc,
-                      data.frame = SOMADataFrame$new(uri, internal_use_only = "allowed_use"),
-                      sparse = SOMASparseNDArray$new(uri, internal_use_only = "allowed_use"))
+                      data.frame = SOMADataFrameOpen(uri, tiledbsoma_ctx = ctx),
+                      sparse = SOMASparseNDArrayOpen(uri, tiledbsoma_ctx = ctx))
         expect_true(inherits(sdf, "SOMAArrayBase"))
         sdf$open("READ", internal_use_only = "allowed_use")
 
@@ -159,8 +163,9 @@ test_that("Iterated Interface from SOMA Sparse Matrix", {
     untar(tarfile = tgzfile, exdir = tdir)
     uri <- file.path(tdir, "soco", "pbmc3k_processed", "ms", "raw", "X", "data")
 
+    # The read_complete et al. in this test case are designed to be verified
+    # against 16MB buffer size, and the particular provided input dataset.
     ctx <- SOMATileDBContext$new(c(soma.init_buffer_bytes='16777216'))
-
     snda <- SOMASparseNDArrayOpen(uri, tiledbsoma_ctx = ctx)
 
     expect_true(inherits(snda, "SOMAArrayBase"))
--- a/apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
+++ b/apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
@@ -245,11 +245,3 @@ test_that("Dimension Point and Ranges Bounds", {
     gc()
 
 })
-
-# testthat::test_file("tests/testthat/test-SOMAArrayReader-Iterated.R")
-# ── Failure ('test-SOMAArrayReader-Iterated.R:130:13'): Iterated Interface from SOMA Classes ──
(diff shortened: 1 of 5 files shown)


jdblischak:
I kicked off a nightly run that uses [kerl/flex-buffer-sizes](https://github.com/single-cell-data/TileDB-SOMA/tree/kerl/flex-buffer-sizes) for TileDB-SOMA

https://github.com/jdblischak/centralized-tiledb-nightlies/actions/runs/9390901834

mojaveazure (approved):
LGTM
image

apis/r/vignettes/soma-experiment-queries.Rmd
```diff
@@ -41,6 +41,13 @@ We can see that our current experiment contains only a single measurement: `"RNA
 experiment$ms
 ```
 
+To use larger (or smaller) buffer sizes:
```
eddelbuettel, line 44: Maybe mention the environment variable setter as the usual fallback (as we are going to Core code here) ?

apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R
```diff
@@ -94,12 +94,15 @@ test_that("Iterated Interface from SOMA Classes", {
...
+    # The read_complete et al. in this test case are designed to be verified
+    # against 16MB buffer size, and the particular provided input dataset.
+    ctx <- SOMATileDBContext$new(c(soma.init_buffer_bytes=as.character(16777216)))
+
     for (tc in test_cases) {
         sdf <- switch(tc,
-                      data.frame = SOMADataFrame$new(uri, internal_use_only = "allowed_use"),
-                      sparse = SOMASparseNDArray$new(uri, internal_use_only = "allowed_use"))
+                      data.frame = SOMADataFrameOpen(uri, tiledbsoma_ctx = ctx),
+                      sparse = SOMASparseNDArrayOpen(uri, tiledbsoma_ctx = ctx))
```
eddelbuettel, line 104: Switch to factory method is good, the test was likely written before we added these.

In the interim @mojaveazure and I had huddled too and see that the create step has the entry for a ctx object so this seems like the best approach.

eddelbuettel (changes requested):
Please also remove the now no-longer needed env variable from the CI yaml files.

johnkerl:
@eddelbuettel as discussed in https://github.com/single-cell-data/TileDB-SOMA/pull/2681 I had on #531 attempted this before and ran into issues with various-sized runners.

The `TILEDB_SOMA_INIT_BUFFER_BYTES` lines on the current CI YAMLs are for robustness with respect to various GHA runners.

They are not related to the bugfix done on this PR.

This PR fixed a bug wherein two cases would pass only with fixed buffer sizes. That is fixed.

What remains, and what needs to remain, is "low ceiling" for all tests, all languages (Python and R), Linux, MacOS, and Windows.

eddelbuettel:
> The TILEDB_SOMA_INIT_BUFFER_BYTES lines on the current CI YAMLs are for robustness with respect to various GHA runners.
> 
> They are not related to the bugfix done on this PR.

By keeping them as env vars you ensure the values are picked up. That neuters any and all CI on this very PR.  As the values are already picked 'for certain' from the TileDB Core API under this, we cannot actually ascertain that this PR does as it claims.

johnkerl:
@eddelbuettel thank you for your feedback.

As noted I don't want to break future CI including Windows as linked on #531.

And we do have the nightlies.

The env vars here must be retained.

But I could bump them from say 16MB in the env vars to 32MB, to ensure bug-fix regression checking on `apis/r/tests/testthat/test-SOMAArrayReader-Iterated.R` _and_ maintaining the "low ceiling" needed for Windows CI.

eddelbuettel:
Conceptually speaking, now that have a handle on setting size as needed, should we not run with the defaults at CI (because hey, CI is there to check our defaults) and the override in test files as needed as we did here?   

(I am aware that we are smudging the lines in other places by suppressing some CI tests on some platforms, or under valgrind, or skip some, ... )

johnkerl:
This is a good idea.

eddelbuettel:
That's basically what I wanted to get at with 'remove the CI value' because it will not blind us from possible errors outside of CI (i.e. real users and clients...) as we are forcing the old small size back in.  Maybe worth a try in a branch.

The main thing is that the regression is fixed, and Paul and I had some good chats about configs. All that may need a revisit too.

johnkerl:
I will file a follow-up task.

There is other more pressing tech debt than this.

Thank you for your dialogue.

eddelbuettel:
I maintained for maybe one and a half years that the layout, build system and CI are all tech debt and got some verbal promises that 'after the release (i.e 1.5.0) this would get addressed.  But that's the way it goes with promises: some remain unfulfilled.

johnkerl:
Getting rid of TileDB-Py and TileDB-R remains tech debt number one. Glad we're getting close on those.

eddelbuettel:
Well we will see about that.  All the config calls we used today make use of tiledb-r config and ctx objects.

PR #2693, opened 2024-06-06
[python] Deprecate `tiledb_ctx` usage in `TileDBSOMAContext`

nguyenv:
**Issue and/or context:**

[sc-49124]

Deprecating usage of `tiledb.Ctx` for users is necessary for the complete removal of tiledb-py. Internally, the codebase now passes around a `TileDBSOMAContext.native_context` which is a C++ `SOMAContext` binding. We no longer use `tiledb.Ctx` anywhere.

**Changes:**

* Deprecate `tiledb_ctx` as an argument into the constructor and `replace`
* Deprecate `tiledb_ctx` as an attribute getter
* Use `tiledb_config` as an alternative 

**Notes for Reviewer:**

I have set the deprecation for 1.14.

### DIFF
--- a/apis/python/src/tiledbsoma/_general_utilities.py
+++ b/apis/python/src/tiledbsoma/_general_utilities.py
@@ -8,8 +8,8 @@
 import os
 import platform
 import sys
+import warnings
 from re import fullmatch
-from typing import Tuple
 
 import tiledb
 
@@ -53,9 +53,17 @@ def get_implementation_version() -> str:
             return "unknown"
 
 
-def get_release_version() -> Tuple[int, int]:
-    version = get_implementation_version().split(".")
-    return (int(version[0]), int(version[1]))
+def assert_version_before(major: int, minor: int) -> None:
+    version_string = get_implementation_version()
+    if version_string == "unknown":
+        warnings.warn(
+            "`assert_version_before` could not retrieve the current "
+            "implementation version"
+        )
+        return
+
+    version = version_string.split(".")
+    assert (int(version[0]), int(version[1])) < (major, minor)
 
 
 def get_storage_engine() -> str:
--- a/apis/python/src/tiledbsoma/_general_utilities.py
+++ b/apis/python/src/tiledbsoma/_general_utilities.py
@@ -9,6 +9,7 @@ import os
 import platform
 import sys
 from re import fullmatch
+from typing import Tuple
 
 import tiledb
 
@@ -50,6 +51,11 @@ def get_implementation_version() -> str:
             return importlib.metadata.version("tiledbsoma")
         except importlib.metadata.PackageNotFoundError:
             return "unknown"
+
+
+def get_release_version() -> Tuple[int, int]:
+    version = get_implementation_version().split(".")
+    return (int(version[0]), int(version[1]))
 
 
 def get_storage_engine() -> str:
--- a/apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
+++ b/apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
@@ -17,13 +17,13 @@ from typing_extensions import Self
 import tiledb
 
 from .. import pytiledbsoma as clib
-from .._general_utilities import get_release_version
+from .._general_utilities import assert_version_before
 from .._types import OpenTimestamp
 from .._util import ms_to_datetime, to_timestamp_ms
 
 
 def _warn_ctx_deprecation() -> None:
-    assert get_release_version() < (1, 14)
+    assert_version_before(1, 14)
     warnings.warn(
         "tiledb_ctx is now deprecated for removal in 1.14. "
         "Use tiledb_config instead by passing "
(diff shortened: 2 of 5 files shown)


apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -118,6 +120,16 @@ def __init__(
                 provided, a new ThreadPoolExecutor will be created with
                 default settings.
         """
+        if tiledb_ctx is not None:
+            version = get_implementation_version().split(".")
+            assert (int(version[0]), int(version[1])) < (1, 14)
```
thetorpedodog, line 125: It seems like it would make sense to have a function that returns this as a tuple or a `packaging.version.Version`.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -118,6 +120,16 @@ def __init__(
...
+            warnings.warn(
+                "The tiledb_ctx parameter will be deprecated in 1.14. "
+                "Use tiledb_config instead by passing "
+                "SOMATileDBContext(tiledb_config=ctx.config()).",
+                DeprecationWarning,
```
thetorpedodog, line 130: I’m a bit confused by the language here. If the parameter isn’t deprecated yet, then what is the warning for? If the intent is to *remove* it in v1.14, then I would say it’s deprecated _now_ for _removal_ in v1.14. As it stands, I would read “will be deprecated in v1.14” to mean that it works now, it will continue to work in v1.14 (but be even more deprecated somehow?), and at some indeterminate point after v1.14, it will stop working.

johnkerl: > deprecated now for removal in v1.14

+1

FWIW I had several co-workers at a former job who (incorrectly) used "deprecate" to mean "get rid of". And I suspect they're not alone; others may have a similar misread of what "deprecate" means. So it's good to be extra-clear on what we mean: "deprecating now" means "flagging that people should stop using this", and "removal in 1.14" meaning that's when it will no longer exist.

nguyenv: Gotcha - thanks for the clarification. I was conflating deprecation and removal and will clear up the language here.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -17,11 +17,21 @@
...
 from .._types import OpenTimestamp
 from .._util import ms_to_datetime, to_timestamp_ms
 
 
+def _tiledb_ctx_deprecation() -> None:
+    assert get_release_version() < (1, 14)
+    warnings.warn(
+        "tiledb_ctx is now deprecated for removal in 1.14. "
+        "Use tiledb_config instead by passing "
+        "SOMATileDBContext(tiledb_config=ctx.config()).",
```
johnkerl, line 30: Won't this still require `tiledb` (i.e. tiledb-py)?

I believe we'll need them to use `ctx.config().dict()`, IIUC.

apis/python/src/tiledbsoma/_general_utilities.py
```diff
@@ -52,6 +53,11 @@ def get_implementation_version() -> str:
             return "unknown"
 
 
+def get_release_version() -> Tuple[int, int]:
+    version = get_implementation_version().split(".")
+    return (int(version[0]), int(version[1]))
```
thetorpedodog, line 58: This is going to fail if the `get_implementation_version` function fails. Since this is now called only once, I would inline it into the `_tiledb_ctx_deprecation` function and if it does fail, just skip the assertion and issue the warning, so that if somebody’s setup metadata gets messed up it doesn’t break SOMA entirely.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -16,10 +17,21 @@
 import tiledb
 
 from .. import pytiledbsoma as clib
+from .._general_utilities import get_release_version
 from .._types import OpenTimestamp
 from .._util import ms_to_datetime, to_timestamp_ms
 
 
+def _tiledb_ctx_deprecation() -> None:
```
thetorpedodog, line 25: maybe `_warn_ctx_deprecation()`?

apis/python/src/tiledbsoma/_general_utilities.py
nguyenv: I actually just created a [new PR](https://github.com/single-cell-data/TileDB-SOMA/pull/2701) on top of this one that uses `get_implementation_version` in another file. Should I inline it for there as well or do better error handling for this

thetorpedodog: I would just add better error handling here—it might be as simple as returning `(0, 0)`, or perhaps this could be written as a function like `assert_version_before(major: int, minor: int)`, so that the assert code is in one place

johnkerl (approved):
🚢

PR #2704, opened 2024-06-10
[r,py,c++] Support data frame write path

eddelbuettel:
**Issue and/or context:**

This PR switches data frame write operations to using `libtiledbsoma`.  It contains the branch `viviannguyen/fix ordered dictionary flag` as well.

**Changes:**

Enable schema creation and data.frame writes.

**Notes for Reviewer:**

~This PR is (for now) still flagged WIP and 'draft' as a pass or two for cleanup may be needed removing some comments and alike.~  However, as it passes all tests locally it is ready for CI.

The commit history is missing timestamps because of rebasing.  The initial branch started a while ago, and Vivian's branch was merged another while back too.  My [predecessor branch is (for now) still up](https://github.com/single-cell-data/TileDB-SOMA/tree/de/sc-44885/dataframe_writeurl) as is [Vivian's branch merged here](https://github.com/single-cell-data/TileDB-SOMA/tree/viviannguyen/fix-ordered-dictionary-flag).

[SC 44885](https://app.shortcut.com/tiledb-inc/story/44885/r-c-rcpp-libtiledbsoma-integration-somadataframe-write-path)


nguyenv (approved):
All I can review is the C++ and Python, which I myself wrote. While I think my own code "looks good," I'm obviously not a good judge of that so best to wait for another review if you can.

To leave some comments for other review for the  C++ portion of this code:

- Prior to these changes, casting the `ArrowTable` values to be written to match what was in the schema on disk was done through Pyarrow in the Python codebase. This PR now moves the casting into the C++ codebase. We now now take the data in the Arrow buffers, place it into C++ vectors, iterate through the C++ vector to cast the user passed in value to the type as specified in the schema on disk via double templated functions, perform a `memcpy` to replace with the new casted buffers, and modify the `ArrowSchema` with the on disk type.
- Issues with casting between `large_string` and `string` enumerated types were indepedently reported by CZI in the Python code at the same time the issue was caught in R. There is a [hotfix](https://github.com/single-cell-data/TileDB-SOMA/pull/2668) solely in the Python side that allowed the census builder unit tests to pass. The fix was sufficient at the time as the R API had not moved over to use the C++ write path yet. This PR now fixes the problem more deeply in the C++ codebase.
- The other change required to get this PR passing, converting between bit and uint8_t representations for Boolean types and validity buffers between Arrow and TileDB, has already been merged: https://github.com/single-cell-data/TileDB-SOMA/pull/2650.

eddelbuettel:
@nguyenv We could also split this off if your PR #2629 went in by itself I could rebase.  I have no preference which way we do this but I _do_ have a preference for this getting done sooner rather than later, if possible, as subsequent work of mine depends on both.

nguyenv:
I am totally fine with leaving this is a single large PR. I think that would be easiest and fastest tbh. I just wanted to leave some context for my own changes for anyone else who is reviewing this.

apis/r/R/TileDBCreateOptions.R
```diff
@@ -229,9 +234,10 @@ TileDBCreateOptions <- R6::R6Class(
     #' @description ...
     #'
     #' @param build_filters Build filters into
-    #' \code{\link[tiledb:tiledb_filter-class]{tiledb_filter}} objects
+    #' \code{\link[tiledb:tiledb_filter-class]{tiledb_filter}} objects. If set to
+    #' \code{FALSE}, JSON strings are created instead of filter objects.
     #'
-    #' @return The create options as a list
+    #' @return The created options as a list
```
mojaveazure, line 240: I wrote this as "create options" because these are options for creation

apis/r/tests/testthat/test-write-soma-resume.R
```diff
@@ -105,8 +107,8 @@ test_that("Resume-mode data frames", {
     )
   }
 
-  # Expect error when writing to existing array
-  expect_error(write_soma(co2, uri = uri, soma_parent = collection))
+  # Expect no error when writing to existing array as it will be skipped
+  expect_silent(write_soma(co2, uri = uri, soma_parent = collection))
```
mojaveazure, line 112: We do need to have some kind of catch here

apis/r/R/TileDBCreateOptions.R
eddelbuettel: I see. Shall we put some ' around it then?

```suggestion
    #' @return The 'create options' as a list
```

apis/r/tests/testthat/test-write-soma-resume.R
eddelbuettel: As discussed with @nguyenv we currently just skip.

apis/r/R/TileDBCreateOptions.R
mojaveazure: I like the quotes

apis/r/tests/testthat/test-write-soma-resume.R
mojaveazure: Hmm, is there a way we can catch the no-op and throw an error from tiledbsoma-r? I don't like the silent fail; if a user is trying to write to something that already exists, and hasn't explicitly said it's okay, we should throw some kind of message as otherwise the user will think "hey this worked"

eddelbuettel: We could -- but I think the throw has to happen at the C++ level and propagate to R and Python.  And then be tested by both.

eddelbuettel: Nevermind -- I was wrong. I do that on the C++ side of the R package.  I can easily error out. Trying that now.

apis/r/src/arrow.cpp
```diff
@@ -99,7 +99,9 @@ void createSchemaFromArrow(const std::string& uri,
     tiledb::VFS vfs(*ctxsp);
     //spdl::warn("[createSchemaFromArrow] About to create {}, dir exists ? {}", uri, vfs.is_dir(uri));
 
-    if (!vfs.is_dir(uri)) {
+    if (vfs.is_dir(uri)) {
```
johnkerl, line 102: I'm curious how this works with `tiledb://...` URIs -- ?

eddelbuettel: It should if you have a env vars set etc.  I test more frequently with plain s3, the rest on top "is just us".

```r
> tiledb::tiledb_vfs_is_dir("s3://tiledb-dirk/penguins")
[1] TRUE
> tiledb::tiledb_vfs_is_dir("s3://tiledb-dirk/does_not_exist")
[1] FALSE
> 
```

eddelbuettel:
@mojaveazure Give it anoother look, please -- your comment was spot and I addressed it in follow-up commit https://github.com/single-cell-data/TileDB-SOMA/pull/2704/commits/f115d43c770f19b449e1ca23f245f0673dae58c7.

apis/r/src/arrow.cpp
johnkerl: @eddelbuettel I'm referring to `tiledb://...` URIs not `s3://...` URIs

johnkerl: The `$exists` method is storage-independent -- it works on more than only VFS.

mojaveazure (approved):

mojaveazure:
Thanks for adding in the error-on-exists! From an R standpoint, this looks good to me! I'll defer to others for Python and C++

eddelbuettel:
Thanks for catching that.  Much better with the explicit (earlier than in core) error.

apis/r/src/arrow.cpp
johnkerl: @eddelbuettel thank you.

```
$ rg bool.exists
src/soma/soma_dataframe.h
99:    static bool exists(std::string_view uri);

src/soma/soma_dense_ndarray.h
99:    static bool exists(std::string_view uri);

src/soma/soma_sparse_ndarray.h
99:    static bool exists(std::string_view uri);
```

`exists` is exposed to users in Python and R; also it's in `libtiledbsoma`.

`vfs` likewise exists in Python, R, and C++.

Regardless of language, `vfs` addresses local disk and S3 but does not handle `tiledb://...` URIs.

This PR is now merged, so we'll need to address this non-functioning with respect to `tiledb://...` URIs in a separate PR.

(Note as always that we don't unit-test `tiledb://...` URIs within this repo as this is an open-source repo.)

johnkerl: https://github.com/single-cell-data/TileDB-SOMA/issues/2714

johnkerl:
Backport unneeded: error on my part: this was merged to `main` just _before_ `release-1.12` was branched off of `main`

Issue #2657, opened 2024-05-30
Test `from_visium` ingestor with Visium dataset


johnkerl:
Experiment with using the `from_visium` on Visium datasets from 10x. Document any bugs that occur.

[[sc-44672]](https://app.shortcut.com/tiledb-inc/story/44672/test-from-visium-ingestor-with-visium-dataset)

johnkerl:
Tutorial notebook created

PR #2705, opened 2024-06-10
[spatial] Add tutorial notebook

ryan-williams:
**Issue and/or context:** #2657

**Changes:**
Add example notebook showing basic spatial SOMA ingestion and use

Demo: https://tiledbsoma.readthedocs.io/en/rw-spnb/notebooks/tutorial_spatial.html

**Notes for Reviewers:**
- Targets branch [spatial](https://github.com/single-cell-data/TileDB-SOMA/tree/spatial), not main
- Currently rebased on #2706

image


apis/python/notebooks/tutorial_spatial.ipynb
```diff
@@ -0,0 +1,727 @@
...
+    "    dst = join(data_dir_10x, dst_name)\n",
+    "    if exists(dst):\n",
+    "        if overwrite_10x:\n",
+    "            err(f\"{dst} exists, removing\")\n",
+    "            remove(dst)\n",
+    "    if not exists(dst):\n",
+    "        makedirs(data_dir_10x, exist_ok=True)\n",
+    "        sh('wget', '-qO', dst, src)\n",
+    "        if dst.endswith('.tar.gz'):\n",
+    "            sh('tar', '-C', data_dir_10x, '-xvf', dst)"
```
jp-dark, line 147: Should this also be skipped if `exp_uri` exists and is not being over-written?

jp-dark (approved):
Thanks for adding this! I have one small non-blocking comment, but feel free to merge with or without the change.

johnkerl (approved):
🚢

apis/python/notebooks/tutorial_spatial.ipynb
ryan-williams: I guess the use case is:
1. Run nb for first time, it downloads 10x data, and creates `exp_uri`
2. Clean up local dir by removing the 10x data, which is no longer needed
3. Run nb subsequent times, it will re-download 10x data

Makes sense to me to skip downloading 10x data in case 3., as you suggest, let me make that change…

Issue #2700, opened 2024-06-10
[c++] Add `SOMAArray.platform_config()` getter


nguyenv:
Add a method that returns the `PlatformConfig` object used to create the `ArraySchema` for the given `SOMAArray`.

johnkerl:
#2709 is closed

PR #2708, opened 2024-06-11
[c++] Add `SOMAArray::config_options` getter

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2700

**Changes:**

Get the `ArraySchema` options that cannot be represented in the `ArrowSchema` (such as filter lists, tile/cell order, etc) in the form of a `PlatformConfig`.


nguyenv:
I originally thought I needed to add this to grab info out of the PlatformConfig in Python from C++, but it ended up not being necessary.

PR #2609, opened 2024-05-21
[c++] Fix nightly build failure on `formattable` of `json` type

johnkerl:
https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/153#issuecomment-2122979236

#2608

PR #2718, opened 2024-06-12
[c++] Fix a nightly-build issue

johnkerl:
* Fixes https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/170
* Basically a recurrence of https://github.com/single-cell-data/TileDB-SOMA/pull/2609
  * There, the `fmt::format` argument which was not `formattable` was of type `json`
  * Here, the `fmt::format` argument which is not `formattable` is of core type `tiledb_datatype_t` which is a C++ int-backed `enum` type
* Open question in my mind: why is that these issues compile OK in our TileDB-SOMA repo CI but fail to compile in the nightlies? I'd love to surface these errors sooner.

### DIFF
--- a/libtiledbsoma/src/utils/arrow_adapter.cc
+++ b/libtiledbsoma/src/utils/arrow_adapter.cc
@@ -565,9 +565,9 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
                     *ctx, attr, child->name);
                 LOG_DEBUG(fmt::format(
                     "[ArrowAdapter] dictionary for {} as {} {}",
-                    child->name,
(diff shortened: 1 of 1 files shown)


jdblischak (approved):

johnkerl:
Evaluating post-merge as per @jdblischak instructions here:
https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/170#issuecomment-2163344686

jdblischak:
> which I can't believe included a full recompile and unit-test run 

xref: https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/170#issuecomment-2163510654

Issue #2699, opened 2024-06-10
[python] Deprecate `tiledb.Filter{List}` getters for `PlatformConfig`


nguyenv:
This is a part of the tiledb-py dep removal. We already have getters that return the JSON representation of the `tiledb.Filter{List}` but the getters that return tiledb-py objects need to be removed [alongside `tiledb.Ctx`](https://github.com/single-cell-data/TileDB-SOMA/pull/2693).

PR #2720, opened 2024-06-12
[python] Deprecate `tiledb.Filter` getters in `PlatformConfig`

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2699

As part of tiledb-py removal work. Related to https://github.com/single-cell-data/TileDB-SOMA/pull/2693.

**Changes:**

Deprecate `tiledb.Filter{List}` getters in `PlatformConfig` and inform users to use JSON getters as an alternative.


johnkerl (approved):
LGTM

PR #2675, opened 2024-06-03
[python] Move `SOMAGroup` to use C++ bindings

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2674

These changes are applied on top of https://github.com/single-cell-data/TileDB-SOMA/pull/2508

**Changes:**

- Replace `tiledb.Group` usage with `clib.SOMAGroup`
- Remove references to `tiledb.TileDBError` as we no longer have exceptions coming from tiledb-py
- Remove `ArrayWrapper` and `GroupWrapper` as these were wrappers around tiledb-py objects
- Addition of `CollectionWrapper`, `ExperimentWrapper`, and `MeasurementWrapper` to wrap around new C++ bindings
- Rename `TileDBObject` to `SOMAObject` and `_tiledb_object.py` to `_soma_object.py` to reflect moveover changes

**Notes for Reviewer:**

Although `SOMAArray` and `SOMAGroup` now fully use the C++ bindings, there's still areas of the code that use tiledb-py and need to be replaced in a follow up PR. For example, we use `tiledb.VFS` in several areas.

PR #2731, opened 2024-06-14
[c++] More `SOMAGroup` refinements

nguyenv:
**Issue and/or context:**

This work is originally a part of https://github.com/single-cell-data/TileDB-SOMA/pull/2675

**Changes:**

- Replace contents of `SOMACollection::get` with a call to `SOMAObject::open`
- Rename `SOMAGroup::member_to_uri_mapping` to `SOMAGroup::members` (this is a getter that returns a map of the child member's name and its associated `SOMAGroupEntry`)
- Refactor `SOMAGroup::members` mapping to store `SOMAGroupEntry` values. `SOMAGroupEntry` stores the URI of the child member and the type of `SOMAObject` it is. This type info gets passed into `SOMAObject::open` (via `SOMACollection::get`) to reduce the number of unnecessary `tiledb::open` calls
- Add an iterator for `SOMACollection`
- Add missing documentation
- Rethrow `TileDBError` as `TileDBSOMAError` where relevant

### DIFF
--- a/libtiledbsoma/test/unit_soma_collection.cc
+++ b/libtiledbsoma/test/unit_soma_collection.cc
@@ -71,7 +71,7 @@ TEST_CASE("SOMACollection: add SOMASparseNDArray") {
         "l",
         ArrowTable(
             std::move(index_columns.first), std::move(index_columns.second)));
-    REQUIRE(soma_collection->members() == expected_map);
+    REQUIRE(soma_collection->members_map() == expected_map);
     REQUIRE(soma_sparse->uri() == sub_uri);
     REQUIRE(soma_sparse->ctx() == ctx);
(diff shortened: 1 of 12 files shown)


libtiledbsoma/test/unit_soma_collection.cc
```diff
@@ -71,7 +71,7 @@ TEST_CASE("SOMACollection: add SOMASparseNDArray") {
         "l",
         ArrowTable(
             std::move(index_columns.first), std::move(index_columns.second)));
-    REQUIRE(soma_collection->member_to_uri_mapping() == expected_map);
+    REQUIRE(soma_collection->members() == expected_map);
```
johnkerl, line 74: Maybe `members_map`?

johnkerl (approved):
LGTM but I defer to @eddelbuettel for final approval

PR #2733, opened 2024-06-14
[python] Release GIL where necessary in pybind11 code

nguyenv:
**Issue and/or context:**

This issue was internally reported by @bkmartinjr:
> I am using the latest Python SOMA, and seeing some really weird serialization of my API requests that "feel" like the GIL is not being released for some I/O operations.
>
> For example:  soma.DataFrame.count.  If I'm reading the code right, it calls the underlying nnz method without releasing it.

**Changes:**

- Just as we do for `clib.SOMAArray.nnz`, we need to release the GIL for `clib.SOMADataFrame.count`
- Also release for all `create` and `open` operations

### DIFF
--- a/apis/python/src/tiledbsoma/soma_collection.cc
+++ b/apis/python/src/tiledbsoma/soma_collection.cc
@@ -60,7 +60,8 @@ void load_soma_collection(py::module& m) {
             py::kw_only(),
             "mode"_a,
             "context"_a,
-            "timestamp"_a = py::none())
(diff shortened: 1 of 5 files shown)


nguyenv:
[sc-49550]

bkmartinjr (approved):
I did a quick review and things look good.  I did not spend time looking for other places we might benefit from releasing the GIL, but these fixes should unblock me.  Thanks @nguyenv

PR #2740, opened 2024-06-14
[python] Remove tiledb-py from top-level src (sans `io` and `options`) 5/5

nguyenv:
**Issue and/or context:**

This is on top of https://github.com/single-cell-data/TileDB-SOMA/pull/2736

This removes `import tiledb` from `apis/python/src` but not in `io` or `options` which will be done in subsequent PRs.

**Changes:**

- Remove `tiledb_schema_to_arrow` as we no longer use TileDB `ArraySchema`
- Remove `get_tiledb_py_core_version` and `verify_core_versions`


johnkerl (approved):
🔥

PR #2730, opened 2024-06-13
[r,c++] Support dense nd array write case

eddelbuettel:
**Issue and/or context:**

This PR extends the C++-fication to dense n-dimensional array.

**Changes:**

Refactored code to use C++ classes to write.

**Notes for Reviewer:**

[SC 44887](https://app.shortcut.com/tiledb-inc/story/44887/r-c-rcpp-libtiledbsoma-integration-somadensendarray-write-path)

PR #2755, opened 2024-06-20
[r] Support `SparseNDArray` writes case

eddelbuettel:
**Issue and/or context:**

Following PRs #2704 (data frames) and #2730 (dense arrays) this PR extends to C++-ification to sparse n-dimensional arrays. 

**Changes:**

Refactored code to use C++ classes to write.

**Notes for Reviewer:**

[SC 44886](https://app.shortcut.com/tiledb-inc/story/44886/r-c-rcpp-libtiledbsoma-integration-somasparsendarray-write-path)

### DIFF
--- a/apis/r/R/SOMASparseNDArray.R
+++ b/apis/r/R/SOMASparseNDArray.R
@@ -177,9 +177,6 @@ SOMASparseNDArray <- R6::R6Class(
       self$set_metadata(bbox_flat)
       private$.write_coo_dataframe(coo)
 
-      # tiledb-r always closes the array after a write operation so we need to
-      # manually reopen it until close-on-write is optional
-      self$open("WRITE", internal_use_only = "allowed_use")
       invisible(self)
     },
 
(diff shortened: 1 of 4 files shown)


eddelbuettel:
I think this is now ready for review.  Locally I get `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 3475 ]`.

apis/r/R/SOMASparseNDArray.R
```diff
@@ -227,14 +224,26 @@ SOMASparseNDArray <- R6::R6Class(
...
+      nms <- colnames(values)
+
+      ## the 'soma_data' data type may not have been cached, and if so we need to fetch it
+      if (is.null(private$.type)) {
+          ## TODO: replace with a libtiledbsoma accessor as discussed
+          tpstr <- tiledb::datatype(tiledb::attrs(tiledb::schema(self$uri))[["soma_data"]])
+          arstr <- arrow_type_from_tiledb_type(tpstr)
+          private$.type <- arstr
+      }
+
```
nguyenv, line 236: Are you able to use [SOMAArray::arrow_schema](https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/src/soma/soma_array.h#L564-L567) or [SOMASparseArrayNDArray::schema](https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/src/soma/soma_sparse_ndarray.h#L154-L159) to get the dtype of `soma_data`?

apis/r/R/SOMASparseNDArray.R
```diff
@@ -227,14 +224,26 @@ SOMASparseNDArray <- R6::R6Class(
...
-      ##     tiledb::query_layout(arr) <- "GLOBAL_ORDER"
-      ##     cat("*********", tiledb::query_layout(arr), "*****\n")
-      ##     print(arr)
-      ## }
-      arr[] <- values
+      nms <- colnames(values)
+
+      ## the 'soma_data' data type may not have been cached, and if so we need to fetch it
+      if (is.null(private$.type)) {
+          ## TODO: replace with a libtiledbsoma accessor as discussed
```
nguyenv, line 231: Are you able to use [SOMAArray::arrow_schema](https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/src/soma/soma_array.h#L564-L567) or [SOMASparseArrayNDArray::schema](https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/src/soma/soma_sparse_ndarray.h#L154-L159) to get the dtype of `soma_data`?

eddelbuettel: I think it is not clear where the value would come from.  This function is called with just the values to be written, and this segment is where, if no 'cached' type info exists, we go to the schema on disk as a last resort.

https://github.com/single-cell-data/TileDB-SOMA/blob/f06cd71b3db4cd98d6263013285b8650849a5af2/apis/r/R/SOMASparseNDArray.R#L229-L235

In some of the use case we do not need that as `private$.type` has been set.

eddelbuettel: And it is similar for thw `write()` function from which this is called.  It gets called with just `values` (and an optional bbox).  No schema info.

nguyenv (approved):
As discussed over DM, I will add a dtype getter for the `soma_data` into libtiledbsoma.

PR #2756, opened 2024-06-21
Use core 2.23.1 on `release-1.11` branch

johnkerl:
Following our [established procedure](https://www.notion.so/785d165226074071872a8b797133d36c?v=c168da78516e487eac3746d38924fa97&p=d1c7a37761b04cf699a705bb8963dea9&pm=s).

Note in particular that core artifacts are updated using the error-proof `./scripts/update-tiledb-version.py 2.23.1`.

### DIFF
--- a/.github/workflows/python-ci-packaging.yml
+++ b/.github/workflows/python-ci-packaging.yml
@@ -76,7 +76,7 @@ jobs:
         run: |
           mkdir -p external
           # Please do not edit manually -- let scripts/update-tiledb-version.py update this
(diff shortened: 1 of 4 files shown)


johnkerl:
@ihnorton I believe we will need a 0.29.2 having core 2.23.1:
https://github.com/TileDB-Inc/TileDB-Py/blob/0.29.1/setup.py#L23

johnkerl:
We are moving forward with core 2.24 and tiledbsoma 1.12

PR #2762, opened 2024-06-27
[c++] Update CMake to use target-based properties

### LINKED ISSUE #2664: [c++] Update CMake to use target-based properties
Update CMake to use target-based properties to set compiler options, compiler defintions, include directories, etc. This task is motivated by the need to set all compiler options except `-Werror` for nanoarrow (see #2408).

jp-dark:
**Issue and/or context:** Closes #2664

**Changes:**
* Add options explicitly to targets instead of to everything. Note: this means compile options are no longer propagated to third-party libraries included with `FetchContent` and `ExternalProject_Add`.
* Move `find_package` calls and coverage flags to base `CMakeLists.txt` instead of adding them twice.

### DIFF
--- a/libtiledbsoma/CMakeLists.txt
+++ b/libtiledbsoma/CMakeLists.txt
@@ -165,48 +165,67 @@ if(MSVC)
   # C4702: unreachable code
   # C4800: warning implicit cast int to bool
   # C4996: deprecation warning about e.g. sscanf.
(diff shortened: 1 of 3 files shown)


eddelbuettel (approved):
This seems to be functionally equivalent but uses better (newer?) cmake idioms -- a good change!

Take what I say with a grain of salt though as I am far from being cmake-fluent.

eddelbuettel (approved):
:+1:

Issue #2408, opened 2024-04-08
[c++] Create a `cmake` rule for `nanoarrow`


johnkerl:
Following on from #2362:

* `nanoarrow.c` and `nanoarrow.h` are vendored
* They contain a bit of `sprintf` with destination named `out` which is, from the compiler's point of view, possibly null
* Different compilers handle this differently:
  * No problem in our main CI
  * Error in our wheels-build CI, fixed on #2398 
  * Another build failure on Centos 8 with GCC 8.5: 2402
    * As found on #2402, adding `-D TILEDBSOMA_ENABLE_WERROR=OFF` at `scripts/bld` did not result in `-Werror` being avoided at `nanoarrow.c`
* The bigger issue:
  * `nanoarrow.c` -- and _only_ `nanoarrow.c` -- needs to be compiled without `-Werror`, and _by default_
  * This will take some `cmake` work

johnkerl:
@nguyenv is our TileDB-SOMA `cmake` expert: yet I want to free up as much of her time as possible toward closing out #1632, upon which much depends.

I'm not a `cmake` expert by any means but am willing to attempt to factor out a `cmake` file for `nanoarrow.c`, and will ask for help as needed.

eddelbuettel:
Note that there already is _some_ support upstream:

- https://github.com/apache/arrow-nanoarrow/tree/main/examples/cmake-scenarios (more featureful)
- https://github.com/apache/arrow-nanoarrow/tree/main/examples/cmake-minimal (more minimal)

and there is some 'moar' cmake support by other projects in a recent PR / like to be in 0.5.0

- https://github.com/apache/arrow-nanoarrow/issues/350

What may be the real desired outcome here is a solid `FindNanoarrow_EP.cmake` in `libtiledbsoma/cmake/Modules/` where we could also pin the release we want (we are currently at 0.4.0).

PR #2763, opened 2024-06-28
Split nanoarrow into a separate object library

jp-dark:
**Issue and/or context:** #2408

**Changes:**
Create a separate object library for nanoarrow without `-Werror`.


eddelbuettel (approved):
Thanks for doing this, and kudos.  I had reached out to our two of our resident cmake whizzes but they did not find a narrow / short path.

The CI fail is unrelated and also currently holding a CI of mine.  I think we can merge this despite the one :x: .

eddelbuettel (approved):
While I fear CI may still fail for R here, I think it is unrelated.  So this should still be ready to go.

PR #2768, opened 2024-07-02
[python] Continue to use `tiledb.Ctx`'s config until it is removed

nguyenv:
**Issue and/or context:**

[sc-50154]

**Changes:**

https://github.com/single-cell-data/TileDB-SOMA/pull/2693 deprecated `tiledb.Ctx` in favor of `clib.SOMAContext`. When setting `tiledb_config`, we switched over to using the config in `clib.SOMAContext`. However, if the user passed in a `tiledb_ctx`, it would ignore it and not use the correct config. Since we are still allowing `tiledb_ctx` until it is removed in 1.14, ensure that we still pull the config from `tiledb_ctx` if that is what the user passed in.

**Notes for Reviewer:**

Passing this in as a `tiledb_config` works as expected.

### DIFF
--- a/apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
+++ b/apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
@@ -221,16 +221,22 @@ class SOMATileDBContext(ContextBase):
 
         Returns a new dict with the contents. Caller must hold ``_lock``.
         """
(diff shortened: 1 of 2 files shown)


johnkerl (approved):
Verified good using the scenario from sc-50154 -- thank you @nguyenv !

PR #2777, opened 2024-07-05
Use core 2.24.2

johnkerl:

As always, I ran the error-proof

```
./scripts/update-tiledb-version.py 2.24.2
```


johnkerl:
See also https://github.com/TileDB-Inc/TileDB-Py/pull/2001

nguyenv (approved):

Issue #2787, opened 2024-07-10
[performance] Implement/validate a non-opening member-list accessor


johnkerl:
Follow-up for https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/596.

We need to use, or create, an accessor as in that PR that uses `tiledbsoma-py` rather than `tiledb-py`

See also [[sc-49443]](https://app.shortcut.com/tiledb-inc/story/49443/performance-issue-in-soma-group-member-retrieval).

PR #2795, opened 2024-07-15
[python] Add `CollectionBase.members`

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2787

**Changes:**

The binding from Python to C++ `SOMAGroup::members_map` already exists. We are just exposing this at the API level so that the user can call `SOMA{Collection,Experiment,Measurement}.members()` to get a dictionary of member names to a tuple of (uri, type).

https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/src/soma/soma_group.cc#L142-L156
When creating or opening a `SOMAGroup` in C++, we call `fill_caches` which populates the member mapping by calling `tiledb::Object::member` rather than opening each object in the collection.

### DIFF
--- a/apis/python/src/tiledbsoma/_collection.py
+++ b/apis/python/src/tiledbsoma/_collection.py
@@ -531,6 +531,11 @@ class CollectionBase(  # type: ignore[misc]  # __eq__ false positive
         )
         return self
 
+    def members(self) -> Dict[str, Tuple[str, str]]:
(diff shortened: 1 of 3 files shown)


johnkerl (approved):
Sorry I didn't approve this sooner @nguyenv !!! :|

PR #2800, opened 2024-07-17
[python] Hold onto GIL for open methods

nguyenv:
We have observed CI randomly segfaulting since the merger of https://github.com/single-cell-data/TileDB-SOMA/pull/2733. Consistent in all these stacktraces is `open`. This PR reverts the addition of `gil_scoped_release` from the `SOMAArray::open` bindings.


johnkerl (approved):
I think this is a smart move

PR #2806, opened 2024-07-23
[python] Replace `py::none` with `py::list` for `column_names` default arg

nguyenv:
**Issue and/or context:**

As internally reported, the `open` bindings error out when `column_names` uses the default argument `py::none` because the C++ argument is a `std::vector<std::string>` and not `optional<...>`.

```
In [10]: clib.SOMAObject.open("tiledb://TileDB-Inc/4b53d109-27aa-4e17-b02a-d61710719b7b", mode=clib.OpenMode.read, context=native_context)
Out[10]: <tiledbsoma.pytiledbsoma.SOMADataFrame at 0x7bc8d4304eb0>

In [11]: clib.SOMADataFrame.open("tiledb://TileDB-Inc/4b53d109-27aa-4e17-b02a-d61710719b7b", mode=clib.OpenMode.read, context=native_context)
TypeError                                 Traceback (most recent call last)
Cell In[11], line 1
----> 1 clib.SOMADataFrame.open("tiledb://TileDB-Inc/4b53d109-27aa-4e17-b02a-d61710719b7b", mode=clib.OpenMode.read, context=native_context)

TypeError: open(): incompatible function arguments. The following argument types are supported:
    1. (uri: str, mode: tiledbsoma.pytiledbsoma.OpenMode, context: tiledbsoma.pytiledbsoma.SOMAContext, *, column_names: list[str] = None, result_order: tiledbsoma.pytiledbsoma.ResultOrder = <ResultOrder.automatic: 0>, timestamp: Optional[tuple[int, int]] = None) -> tiledbsoma.pytiledbsoma.SOMADataFrame

Invoked with: 'tiledb://TileDB-Inc/4b53d109-27aa-4e17-b02a-d61710719b7b'; kwargs: mode=<OpenMode.read: 0>, context=<tiledbsoma.pytiledbsoma.SOMAContext object at 0x7bc8d431c4f0>
```

**Changes:**

- Use an empty tuple rather than `None`
- Add tests using bindings directly

### DIFF
--- a/apis/python/tests/test_dataframe.py
+++ b/apis/python/tests/test_dataframe.py
@@ -111,6 +111,20 @@ def test_dataframe(tmp_path, arrow_schema):
         assert sorted([e.as_py() for e in list(table["baz"])]) == ["apple", "cat"]
         assert [e.as_py() for e in list(table["quux"])] == [True, False]
 
+    # Open and read with bindings
+    sdf = soma.pytiledbsoma.SOMADataFrame.open(
+        uri, soma.pytiledbsoma.OpenMode.read, soma.pytiledbsoma.SOMAContext()
+    )
+    table = sdf.read_next()
+    assert table.num_rows == 5
+    assert table.num_columns == 5
+    assert [e.as_py() for e in list(table["soma_joinid"])] == pydict["soma_joinid"]
+    assert [e.as_py() for e in list(table["foo"])] == pydict["foo"]
+    assert [e.as_py() for e in list(table["bar"])] == pydict["bar"]
+    assert [e.as_py() for e in list(table["baz"])] == pydict["baz"]
+    assert [e.as_py() for e in list(table["quux"])] == pydict["quux"]
+    sdf.close()
+
     # Validate TileDB array schema
     with tiledb.open(uri) as A:
         assert A.schema.sparse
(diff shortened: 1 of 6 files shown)


bkmartinjr (commented):
Question: is there any benefit from using tuple() rather than list()?  Just thinking about extra memory allocations.

bkmartinjr (approved):
No concerns other than the question about using tuple rather than list.

nguyenv:
Good point. I've changed it to `py::tuple()` for less overhead.

apis/python/tests/test_dataframe.py
```diff
@@ -111,6 +111,20 @@ def test_dataframe(tmp_path, arrow_schema):
...
+        uri, soma.pytiledbsoma.OpenMode.read, soma.pytiledbsoma.SOMAContext()
+    )
+    table = sdf.read_next()
+    assert table.num_rows == 5
+    assert table.num_columns == 5
+    assert [e.as_py() for e in list(table["soma_joinid"])] == pydict["soma_joinid"]
+    assert [e.as_py() for e in list(table["foo"])] == pydict["foo"]
+    assert [e.as_py() for e in list(table["bar"])] == pydict["bar"]
+    assert [e.as_py() for e in list(table["baz"])] == pydict["baz"]
+    assert [e.as_py() for e in list(table["quux"])] == pydict["quux"]
```
thetorpedodog, line 125: the `list(...)` calls here are unnecessary; these can just be `[e.as_py() for e in table["soma_joinid"]]`.

thetorpedodog (approved):

PR #2808, opened 2024-07-23
[python] Allow platform config to set extent and filter

johnkerl:
Found while experimenting with write-performance tuning, including dogfooding of #2564.

It turns out if the platform-config create spec for a dim has tile extent _and_ a filter, the filter was begin ignored.

I fixed this, and improved the unit-test coverage to catch the fixed bug.


nguyenv (approved):
Thank you for fixing this

Issue #2797, opened 2024-07-16
[c++] Add `soma_data` type getter for `NDArray`s


nguyenv:
As discussed in https://github.com/single-cell-data/TileDB-SOMA/pull/2755, we should add a `soma_data` type getter for the `NDArray` classes.

PR #2810, opened 2024-07-23
[c++] Add `NDArray::soma_data_type`

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2797

**Changes:**

As requested for the SOMA R API, add `soma_data_type` for `SOMADenseNDArray` and `SOMASparseNDArray`.

### DIFF
--- a/libtiledbsoma/src/soma/soma_sparse_ndarray.h
+++ b/libtiledbsoma/src/soma/soma_sparse_ndarray.h
@@ -157,6 +157,14 @@ class SOMASparseNDArray : public SOMAArray {
      * @return std::unique_ptr<ArrowSchema>
      */
     std::unique_ptr<ArrowSchema> schema() const;
+
+    /**
+     * @brief Get the soma_data's dtype in the form of an Arrow
+     * format string.
+     *
+     * @return std::string_view Arrow format string.
+     */
+    std::string_view soma_data_type();
 };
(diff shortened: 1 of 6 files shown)


libtiledbsoma/src/soma/soma_sparse_ndarray.h
```diff
@@ -157,6 +157,14 @@ class SOMASparseNDArray : public SOMAArray {
...
      */
     std::unique_ptr<ArrowSchema> schema() const;
+
+    /**
+     * @brief Get the soma_data's dtype in the form of an Arrow
+     * format string.
+     *
+     * @return std::string_view Arrow format string.
+     */
+    std::string_view soma_data_type();
```
eddelbuettel, line 167: From glancing at it, both functions appear to be identical so should they be promoted 'up' to SOMAArray?

nguyenv: The problem with that is `SOMADataFrame` does not use `soma_data`

eddelbuettel: Spot on, thank you. I thought of that (but didn't open the file).  Make sense as is then.

eddelbuettel (approved):
:shipit:

Issue #2765, opened 2024-07-02
[Feature request] `ExperimentAxisQuery.to_anndata()` (and R analogs) should drop unused categories on axis data frames 


pablo-gar:
**Is your feature request related to a problem? Please describe.**
Some analytical pipelines, specially those that relate to visualizations rely on the categories of `pandas.Categorical`. In the case of large SOMAExperiments, many times a query will result on unused categories for potentially many columns of `obs` or `var`, thus the user needs to **always** iterate on all columns and perform a `cat.remove_unused_categories()` operation.

See for example this reproducible example

```python
import cellxgene_census
import scanpy as sc
census = cellxgene_census.open_soma(census_version="2024-05-20")

human = census["census_data"]["homo_sapiens"]
query = human.axis_query(
    measurement_name = "RNA",
    obs_query = tiledbsoma.AxisQuery(
        value_filter = "tissue == 'tongue' and is_primary_data == True"
    )
)

adata = query.to_anndata(column_names={"obs": ["tissue"]}, X_name = "normalized")
sc.pp.pca(adata)
sc.pp.neighbors(adata)
sc.tl.umap(adata)
sc.pl.umap(adata, color="tissue")
```

Only one `"tissue"` was selected but **all** hundreds of tissues are drawn in the umap

<img width="756" alt="image" src="https://github.com/single-cell-data/TileDB-SOMA/assets/29237819/28b3a4d3-6063-454d-be82-f4d351d043a4">

**Describe the solution you'd like**
`ExperimentAxisQuery.to_anndata()` returns an anndata with unused categories already removed in the axis data frames

johnkerl:
Needs Python & R both. We can subtask.

eddelbuettel:
For R it is a pretty common task and a [base R function](https://stat.ethz.ch/R-manual/R-devel/library/base/html/droplevels.html)

```r
> example(droplevels)

drplvl> aq <- transform(airquality, Month = factor(Month, labels = month.abb[5:9]))

drplvl> aq <- subset(aq, Month != "Jul")

drplvl> table(           aq $Month)

May Jun Jul Aug Sep 
 31  30   0  31  30 

drplvl> table(droplevels(aq)$Month)

May Jun Aug Sep 
 31  30  31  30 
> 
```

mojaveazure:
[...]

PR #2811, opened 2024-07-24
[python] Drop unused categories in `ExperimentAxisQuery.to_anndata`

nguyenv:
**Issue and/or context:**

#2765

**Changes:**

- Create `ExperimentAxisQuery` which inherits from somacore's `query.ExperimentAxisQuery`
- `ExperimentAxisQuery.to_anndata` iterates through each column and drops unused categories

**Notes for Reviewer:**

Should this be implemented in somacore instead?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
     - id: mypy
       additional_dependencies:
         - "pandas-stubs==1.5.3.230214"
-        - "somacore==1.0.11"
+        - "somacore==1.0.13"
         - "types-setuptools==67.4.0.3"
       args: ["--config-file=apis/python/pyproject.toml", "apis/python/src", "apis/python/devtools"]
       pass_filenames: false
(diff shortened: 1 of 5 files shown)


johnkerl (changes requested):
This is awesome!

I see no reason not to do this in somacore ...

ivirshup:
fwiw, anndata will move to not doing this by default. This new behavior will be available as a feature flag in anndata 0.11.0.

ivirshup:
One of the reasons we are doing this is so that we can maintain consistent categorical labels across objects. Which would be particularly useful for objects extracted from a large corpus like census.

johnkerl:
@nguyenv sorry for the delay, I was AFK for a bit.

To quicken things, I pushed this commit just now to this PR:
https://github.com/single-cell-data/TileDB-SOMA/pull/2811/commits/f1ba80057954eb2f4be5225541e93857bf43530f

johnkerl:
@nguyenv I'll let you revert the changes to `_experiment.py` (now that they're in `somacore`), and, parameterize `drop_levels` in your unit-test cases

nguyenv:
Yes will do that now.

johnkerl (approved):
🚢

PR #2813, opened 2024-07-26
Use core 2.25, TileDB-Py 0.31, TileDB-R 0.29

johnkerl:

As always, core artifacts were updating using the error-proof and automated `./scripts/update-tiledb-version.py 2.25.0`.


eddelbuettel (approved):
:+1: 

New R package is at CRAN (since yesterday) and in r2u (since this morning) and our r-universe settings are under our control too

nguyenv (approved):

Issue #2812, opened 2024-07-24
[Bug] Updated error message is much less helpful/ provides incorrect information


ivirshup:
The new `DoesNotExist` error has a confusing message in one our tests. The test is:

```python
import os
import pytest
import tiledbsoma
import cellxgene_census

def test_opening_census_without_anon_access_fails_with_bogus_creds() -> None:
    os.environ["AWS_ACCESS_KEY_ID"] = "fake_id"
    os.environ["AWS_SECRET_ACCESS_KEY"] = "fake_key"
    # Passing an empty context
    with pytest.raises(
        tiledbsoma.DoesNotExistError,
        match=r"The AWS Access Key Id you provided does not exist in our records",
    ):
        cellxgene_census.open_soma(census_version="latest", context=tiledbsoma.SOMATileDBContext())
```

This test now fails with:

```python
>       with pytest.raises(
            tiledbsoma.DoesNotExistError,,
            match=r"The AWS Access Key Id you provided does not exist in our records",
        ):
E       AssertionError: Regex pattern did not match.
E        Regex: 'The AWS Access Key Id you provided does not exist in our records'
E        Input: "'s3://cellxgene-census-public-us-west-2/cell-census/2024-07-22/soma/' does not exist"
```

The new error message is confusing. The old error message says the provided key doesn't exist, which seems accurate. The new error says the URI doesn't exist, which isn't true.

**Versions (please complete the following information):**

I am currently working off of the spatial branch. You can see the failures on macOS and ubuntu CI runners here: https://github.com/chanzuckerberg/cellxgene-census/actions/runs/10070531059/job/27839203972

```
tiledbsoma.__version__              1.12.0rc0.post56.dev999411909
TileDB core version (libtiledbsoma) 2.24.2
python version                      3.11.9.final.0
OS version                          Linux 6.8.0-1010-aws
```

johnkerl:
#2814 is merged

PR #2814, opened 2024-07-26
[python] Use original error message when re-raising with `DoesNotExistError`

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2812

**Changes:**

Use original error message when re-raising with `DoesNotExistError`


johnkerl (approved):

johnkerl:
Most recent commit is a merge commit to current `main`

PR #2816, opened 2024-07-26
[c++] Check index in enumeration casting before using

nguyenv:
**Issue and/or context:**

As reported by @ryan-williams, `pytest test_dataframe.py::test_dataframe_with_enumeration` was segfaulting on an Amazon Linux 2 EC2 instance (although not reproducible on other platforms).

**Changes:**

For nullable dictionary columns, when the value is null, the corresponding index may be a negative integer (depending on whether the index type is a signed or unsigned integer type). Using a negative integer to index into the vector `enums_in_write` causes a segfault so check the index value first.

There are no added tests as `test_dataframe.py::test_dataframe_with_enumeration` no longer segfaulting on the EC2 instance proves that it is fixed.


eddelbuettel (approved):
:shipit:

PR #2817, opened 2024-07-29
[r] Remove default mode-flipping in `$reopen()`

mojaveazure:
As discussed on Slack, `$reopen()` should not default to switching open mode; instead `mode` should always be passed in by the user

Modified SOMA methods:
 - `TileDBObject$reopen()`: `mode` no longer has a default; instead, the user must supply a mode to reopen the object as

[SC-51935](https://app.shortcut.com/tiledb-inc/story/51935)


eddelbuettel (approved):
:shipit:

johnkerl (approved):
Thank you!

Issue #2698, opened 2024-06-08
[r/python] R and Python create groups with incompatible SOMA metadata


bkmartinjr:
Objects created by R or Python bindings should have identical metadata, but currently the R and Python packages tag SOMA objects with different and incompatible metadata tags for `dataset_type`, `soma_encoding_version` and `soma_object_type`.

* Python creates with Unicode strings (e.g., `"dataset_type": "soma"`)
* R package creates objects with byte arrays (e.g., `"dataset_type": b"soma"`)

Using TileDB-Py to inspect two arrays.

When array created by Python ([array info](https://cloud.tiledb.com/soma/TileDB-Inc/83acf862-b507-4020-8541-d553517f5ded/overview)):
```In [25]: tiledb.Group('tiledb://TileDB-Inc/83acf862-b507-4020-8541-d553517f5ded', ctx=tiledb.cloud.Ctx()).meta
Out[25]: {'dataset_type': 'soma', 'soma_encoding_version': '1', 'soma_object_type': 'SOMAExperiment'}
```

When array created by R ([array info](https://cloud.tiledb.com/soma/tiledb-genomics-dev/42e031a6-cbcf-435f-9b22-a42e16fa3a34/overview)):
```In [23]: tiledb.Group('tiledb://tiledb-genomics-dev/42e031a6-cbcf-435f-9b22-a42e16fa3a34', ctx=tiledb.cloud.Ctx()).meta
Out[23]: {'dataset_type': b'soma', 'soma_encoding_version': b'1', 'soma_object_type': b'SOMAExperiment'}
```

I also checked directly reading from S3, i.e., not using the `tiledb://` URI, and the result is the same.

Where a "string" or "byte array" is right, I think it is reasonably clear that there is a bug here - the mandatory metadata tags should be identical no matter which ingestion system is used, and which package is used to read it back.

Side note:  the current Python package seems to have a work-around for this, as it detects byte array metadata and converts it to utf-8.  This is nice, but doesn't seem like the right answer, as it requires any other user of that metadata (e.g., end-user code) to do the same encoding/decoding step for any/all metadata values.

In my _opinion_, we should be using utf-8 everywhere (and document that in the SOMA spec), but at a minimum, we should have common behavior across all reader/writer code.

```
tiledbsoma.__version__              1.11.4
TileDB-Py version                   0.29.0
TileDB core version (tiledb)        2.23.0
TileDB core version (libtiledbsoma) 2.23.0
[...]

PR #2819, opened 2024-07-29
[r] Metadata read/write support via libtiledbsoma

eddelbuettel:
**Issue and/or context:**

This PR adds support for metadata reading and writing via libtiledbsoma.  It does not yet add support for timestamp'ed create / write / read which will follow in subsequent PR building on this.

**Changes:**

The write-path for object creation and write is now fully utilizing the libtiledbsoma facilities (though the additional timestamp feature is yet to come, work in progress).

**Notes for Reviewer:**

[SC 51949](https://app.shortcut.com/tiledb-inc/story/51949/r-meta-data-read-and-write-support)

#2698


apis/r/src/arrow.cpp
```diff
@@ -118,14 +119,26 @@ void createSchemaFromArrow(const std::string& uri, naxpSchema nasp, naxpArray na
...
+        std::string datacoltype = sp->children[sp->n_children-1]->format;
+        tdbs::SOMASparseNDArray::create(uri, datacoltype,
+                                        std::pair(std::move(dimarr), std::move(dimsch)),
+                                        sctx, pltcfg);
+    } else if (datatype == "SOMADenseNDArray") {
+        // for arrays n_children will be three as we have two dims and a data col
+        std::string datacoltype = sp->children[sp->n_children-1]->format;
+        tdbs::SOMADenseNDArray::create(uri, datacoltype,
+                                       std::pair(std::move(dimarr), std::move(dimsch)),
+                                       sctx, pltcfg);
```
nguyenv, line 137: Not necessary to do, but an alternative to doing these if-else statements would be to create a `std::map<std::string, std::function<...>>` where the string datatype maps to the associated `create` static method.

libtiledbsoma/src/soma/soma_array.h
```diff
@@ -662,14 +662,17 @@ class SOMAArray : public SOMAObject {
...
      *
      * @note The writes will take effect only upon closing the array.
      */
     void set_metadata(
         const std::string& key,
         tiledb_datatype_t value_type,
         uint32_t value_num,
-        const void* value);
+        const void* value,
+        bool force = false);
```
nguyenv, line 675: Do you still need this `force` toggle now that you are using the SOMA create methods?

apis/r/src/arrow.cpp
eddelbuettel: Yes I adopted that in one or two places too and like it.  The explicit if/else if/else trees allow for logging and better debugging though.

libtiledbsoma/src/soma/soma_array.h
eddelbuettel: I do. I checked that this morning.   We have a number of entry points from the R code so this could be a ticket for @mojaveazure and me to resolve "medium-term" but for right now there is sadly no way around this: without it we get massive test failure.

nguyenv:
Why are the metadata accessors implemented as static methods? You should be able to use `SOMAArray::open` to open the array and then use the member functions `SOMAArray::get_metadata`, etc. (and same for `SOMAGroup`).

eddelbuettel:
Can you point to line and file?

apis/r/src/metadata.cpp
```diff
@@ -0,0 +1,179 @@
...
+//' @export
+// [[Rcpp::export]]
+std::string get_metadata(std::string& uri, std::string& key, bool is_array,
+                         Rcpp::XPtr<somactx_wrap_t> ctxxp) {
+    // shared pointer to SOMAContext from external pointer wrapper
+    std::shared_ptr<tdbs::SOMAContext> sctx = ctxxp->ctxptr;
+
+    // SOMA Object unique pointer (aka soup)
+    auto soup = getObjectUniquePointer(is_array, OpenMode::read, uri, sctx);
+    auto mv = soup->get_metadata(key);
```
nguyenv, line 96: Nevermind I think I understand now. You open the `SOMAObject` and then call `SOMAObject::get_metadata` which looks how it was also implemented before when you had it calling `tiledb::tiledb_put_metadata` calls. You are just switching the internals from using TileDB to SOMA C++ but it still looks the same to the user.

nguyenv: > Am I overlooking something?

No I think I just misunderstood how to read the R code

eddelbuettel: Phew.  All good then.

And yes given the 11800 lines of R code (!!) in apis/r/R/*, never mind the 10300 lines of test code, I am trying to stick with existing interfaces and "swap" out calling tiledb core with calling libtiledbsoma.

nguyenv (approved):

PR #2820, opened 2024-07-30
[c++/python] Add `tiledb_timestamp` argument for `reopen`

nguyenv:
**Issue and/or context:**

As reported internally, the `reopen` unit tests intermittenly fail due to a race condition when objects are opened at the same timestamp.

**Changes:**

Add `tiledb_timestamp` for `reopen` methods and modify tests to use explicit timestamps to prevent race conditions.

### DIFF
--- a/apis/python/tests/test_sparse_nd_array.py
+++ b/apis/python/tests/test_sparse_nd_array.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import contextlib
+import datetime
 import itertools
 import operator
 import pathlib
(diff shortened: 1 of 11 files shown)


bkmartinjr (changes requested):
Suggestion:  the timestamp param for reopen also accepts a `datetime.datetime` (as does the timestamp for open, etc).  Can we add at least one test to confirm this works as expected, in addition to the existing tests which cover None and int (ms).

Otherwise it looks good to me.

bkmartinjr (approved):

apis/python/tests/test_sparse_nd_array.py
```diff
@@ -84,33 +85,38 @@ def test_sparse_nd_array_reopen(tmp_path):
...
+            with A2.reopen("r", tiledb_timestamp=3) as A3:
                 assert A1.mode == "r"
                 assert A2.mode == "w"
                 assert A3.mode == "r"
-                assert A1.tiledb_timestamp < A2.tiledb_timestamp
-                assert A2.tiledb_timestamp < A3.tiledb_timestamp
-
-    with soma.SparseNDArray.open(tmp_path.as_posix(), "r") as A1:
-        with A1.reopen("r") as A2:
+                assert A1.tiledb_timestamp.timestamp() == 0.001
```
johnkerl, line 97: I'd prefer `A1.tiledb_timestamp_ms` so we can do int-to-int compares rather than relying on whether floating-point equality rounds correctly (there's no need for floating-point here)

johnkerl (changes requested):

johnkerl (approved):

johnkerl (approved):

PR #2821, opened 2024-07-31
pre-commit update

bkmartinjr:
Changes:
* ran `pre-commit autoupdate` which updated black, ruff and mypy versions
* ran pre-commit, which applied latest formatting rules
* updated pyproject.toml to fix obsolete syntax issue with ruff config
* fix lint pointed out by latest mypy

### DIFF
--- a/apis/python/src/tiledbsoma/_collection.py
+++ b/apis/python/src/tiledbsoma/_collection.py
@@ -204,8 +204,7 @@ class CollectionBase(  # type: ignore[misc]  # __eq__ false positive
         *,
         uri: Optional[str] = ...,
         platform_config: Optional[options.PlatformConfig] = ...,
-    ) -> "Collection[AnySOMAObject]":
-        ...
+    ) -> "Collection[AnySOMAObject]": ...
 
     @overload
     def add_new_collection(
@@ -215,8 +214,7 @@ class CollectionBase(  # type: ignore[misc]  # __eq__ false positive
         *,
         uri: Optional[str] = ...,
         platform_config: Optional[options.PlatformConfig] = ...,
-    ) -> _Coll:
-        ...
+    ) -> _Coll: ...
 
     def add_new_collection(
         self,
@@ -572,7 +570,7 @@ class CollectionBase(  # type: ignore[misc]  # __eq__ false positive
     def _contents_lines(self, last_indent: str) -> Iterable[str]:
         indent = last_indent + "    "
         if self.closed:
-            return ()
+            return
         for key, entry in self._contents.items():
             obj = entry.soma
             if obj is None:
--- a/apis/python/src/tiledbsoma/_types.py
+++ b/apis/python/src/tiledbsoma/_types.py
@@ -20,7 +20,7 @@ if TYPE_CHECKING:
     NPInteger = np.integer[npt.NBitBase]
     NPFloating = np.floating[npt.NBitBase]
     NPNDArray = npt.NDArray[np.number[npt.NBitBase]]
-    PDSeries = pd.Series[Any]  # type: ignore[misc]
+    PDSeries = pd.Series[Any]  # type: ignore[misc] # a pd.Series of "any" type will raise mypy error
 else:
     NPInteger = np.integer
     NPFloating = np.floating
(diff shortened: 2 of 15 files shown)


apis/python/src/tiledbsoma/_collection.py
```diff
@@ -572,7 +570,7 @@ def _my_repr(self) -> str:
     def _contents_lines(self, last_indent: str) -> Iterable[str]:
         indent = last_indent + "    "
         if self.closed:
-            return ()
```
jp-dark, line 575: Is this doing what we expect it to be doing?

jp-dark (approved):

apis/python/src/tiledbsoma/_collection.py
bkmartinjr: I believe so - we simply want the iterable to end, which it does.  No idea why the code used to return an empty tuple, which is definitely wrong for an iterator of strings

thetorpedodog: I think I was the author here, and it turns out this was a misconception that I had. What I didn’t realize was that the mere presence of a `yield` in a function turned it into a generator; it didn’t _just_ happen when the `yield` statement was hit.

So in my head:

```python
def some_fn(*, do_iter: bool) -> object:
  if do_iter:
    # If you hit this line, the function suspends itself and returns
    # a generator which yields the string "5" and then is exhausted.
    yield "5"
  else:
    # This is the part where I was WRONG: I thought that
    # if you hit this line, the function simply returns the value 1.
    return 1

some_fn(do_iter=False)
# -> 1
# i.e., list(some_fn(do_iter=False)) -> list(1) -> TypeError

some_fn(do_iter=True)
# -> generator which yields "5", then exits.
# i.e., list(some_fn(do_iter=True)) -> ["5"]
```

This is not the case; instead what happens is that the fact that there is a `yield` statement in the function at all means it is _always_ a generator. The value after the `return` is ignored, so what actually happens is:

```python
some_fn(do_iter=False)
# -> generator which immediately exits
# i.e., list(some_fn(do_iter=False)) -> []
```

For the case where I was “returning” `()`, that meant that we were actually returning an empty iterator, which was what I wanted to happen anyway. I’m surprised no tools actually caught that there shouldn’t be a value there!

apis/python/src/tiledbsoma/io/conversions.py
```diff
@@ -50,7 +52,7 @@ def to_tiledb_supported_array_type(name: str, x: _MT) -> _MT:
     ):
         # mypy issues a spurious error here, but only when
         # _to_tiledb_supported_dtype is decorated with @typeguard_ignore???
```
thetorpedodog, line 54: If the type-ignore is no longer needed, this comment can be removed.

apis/python/src/tiledbsoma/_types.py
```diff
@@ -18,7 +20,7 @@
     NPInteger = np.integer[npt.NBitBase]
     NPFloating = np.floating[npt.NBitBase]
     NPNDArray = npt.NDArray[np.number[npt.NBitBase]]
-    PDSeries = pd.Series[Any]
+    PDSeries = pd.Series[Any]  # type: ignore[misc]
```
thetorpedodog, line 23: could you add a note for why there's an ignore here?

bkmartinjr:
> I’m surprised no tools actually caught that there shouldn’t be a value there!

Actually, the latest mypy did catch it, which is why I made the modification.  Progress!

apis/python/src/tiledbsoma/_types.py
bkmartinjr: Will do. Without the ignore, mypy complains:
> apis/python/src/tiledbsoma/_types.py:23: error: Type application targets a non-generic function or class  [misc]

I have not yet figured out how to say "a Series of any type" with the current pandas-stubs.  My best guess is that it isn't possible, but instead you have to literally itemize every type that might be supported in the Series (e.g., pd.Series[int] | pd.Series[float], | ...), which seemed a bit onerous.

apis/python/src/tiledbsoma/io/conversions.py
bkmartinjr: fixed

PR #2825, opened 2024-08-02
[r] Add `drop_levels` to `SOMAExperimentAxisQuery` -> ecosystem outgestors

### LINKED ISSUE #2765: [Feature request] `ExperimentAxisQuery.to_anndata()` (and R analogs) should drop unused categories on axis data frames

mojaveazure:
R analog of #2811 and single-cell-data/SOMA#204; add a `drop_levels` paramter to the ecosystem outgestors to drop unused factor levels from resulting data frames

Modified SOMA methods:
 - `SOMAExperimentAxisQuery$to_seurat()`: add `drop_levels` to drop drop unused levels from `obs` and `var` data frames
 - `SOMAExperimentAxisQuery$to_seurat_assay()`: add `drop_levels` to drop unused levels from `var` data frame
 - `SOMAExperimentAxisQuery$to_single_cell_experiment()`: add `drop_levels` to drop unused levels from `obs` and `var` data frames

Also shifts `SOMAExperimentAxisQuery$to_seurat()` and `SOMAExperimentAxisQuery$to_seurat_assay()` to use `SOMAExperimentAxisQuery$private$.load_df()` for loading `obs` and `var`; removing standalone code and increase sharing with the SCE outgestor

resolves #2765

[SC-51945](https://app.shortcut.com/tiledb-inc/story/51945)

### DIFF
--- a/apis/r/R/SOMAExperimentAxisQuery.R
+++ b/apis/r/R/SOMAExperimentAxisQuery.R
@@ -515,6 +515,7 @@ SOMAExperimentAxisQuery <- R6::R6Class(
     #' @param obsm_layers \Sexpr[results=rd]{tiledbsoma:::rd_outgest_mlayers()}
     #' @param varm_layers \Sexpr[results=rd]{tiledbsoma:::rd_outgest_mlayers(axis = 'varm')}
     #' @param obsp_layers \Sexpr[results=rd]{tiledbsoma:::rd_outgest_players()}
+    #' @param drop_levels Drop unused levels from \code{obs} and \code{var} factor columns
     #'
     #' @return A \code{\link[SeuratObject]{Seurat}} object
     #'
@@ -526,7 +527,8 @@ SOMAExperimentAxisQuery <- R6::R6Class(
       var_column_names = NULL,
       obsm_layers = NULL,
       varm_layers = NULL,
-      obsp_layers = NULL
+      obsp_layers = NULL,
+      drop_levels = FALSE
     ) {
       check_package('SeuratObject', version = .MINIMUM_SEURAT_VERSION())
       op <- options(Seurat.object.assay.version = 'v3')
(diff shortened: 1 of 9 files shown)


apis/r/R/SOMAExperimentAxisQuery.R
```diff
@@ -515,6 +515,7 @@ SOMAExperimentAxisQuery <- R6::R6Class(
     #' @param obsm_layers \Sexpr[results=rd]{tiledbsoma:::rd_outgest_mlayers()}
     #' @param varm_layers \Sexpr[results=rd]{tiledbsoma:::rd_outgest_mlayers(axis = 'varm')}
     #' @param obsp_layers \Sexpr[results=rd]{tiledbsoma:::rd_outgest_players()}
+    #' @param drop_levels Drop unused levels from \code{obs} and \code{var} factor columns
```
eddelbuettel, line 518: I do not use these patterns when I write docs so pardon the naive question but why do all other params here follow a pattern and this one doesn't?

eddelbuettel (approved):
Looks good per the added tests.

One mostly curiously question about Rd writing style inline

pablo-gar (approved):

apis/r/R/SOMAExperimentAxisQuery.R
mojaveazure: The previous ones used `\Sexpr` for templated docs that are similar but with minor variations between different methods as `@template` does not work with R6; however, the use of `\Sexpr` is the cause of the "installing package to process help pages" step when building tiledbsoma-r from source. As this step has been a source of complaint, I've resorted to copy-pasting docs and will eventually switch out all `\Sexpr` to copy-pasted docs to make building tiledbsoma-r easier

eddelbuettel: I am 100% on board with the simplification and robustification.

PR #2834, opened 2024-08-05
[r/python] Deprecation notices for `used_shape`

johnkerl:
**Issue and/or context:** For #2407

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/_types.py
+++ b/apis/python/src/tiledbsoma/_types.py
@@ -20,7 +20,8 @@ if TYPE_CHECKING:
     NPInteger = np.integer[npt.NBitBase]
     NPFloating = np.floating[npt.NBitBase]
     NPNDArray = npt.NDArray[np.number[npt.NBitBase]]
-    PDSeries = pd.Series[Any]
+    # A pd.Series of "Any" type will raise mypy error:
+    PDSeries = pd.Series[Any]  # type: ignore[misc]
 else:
     NPInteger = np.integer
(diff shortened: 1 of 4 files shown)


apis/python/src/tiledbsoma/_types.py
```diff
@@ -20,7 +20,8 @@
     NPInteger = np.integer[npt.NBitBase]
     NPFloating = np.floating[npt.NBitBase]
     NPNDArray = npt.NDArray[np.number[npt.NBitBase]]
-    PDSeries = pd.Series[Any]  # type: ignore[misc] # a pd.Series of "any" type will raise mypy error
+    # A pd.Series of "Any" type will raise mypy error:
```
johnkerl, line 23: Not related to this PR. Lint flared up today right here. IDK why.

mojaveazure (approved):

mojaveazure:
For the R API, you could also call `.Deprecated()` in `used_shape()` to throw a deprecation warning when calling that method

johnkerl:
We have some flakiness around (I think) the particular linter version available -- something bumped today -- lint passing on Python 3.8 and failing on Python 3.11, and it appears they cannot be both appeased simultaneously ...

johnkerl:
> For the R API, you could also call .Deprecated() in used_shape() to throw a deprecation warning when calling that method

Oops @mojaveazure I meant to do that on this PR -- I'll put up another

Issue #2839, opened 2024-08-06
[python] CI `mypy` issue as of 2024-08-05


johnkerl:
To be investigated:

* Example CI fail https://github.com/single-cell-data/TileDB-SOMA/pull/2834/checks (not just that PR)
  * Lint ok on python 3.11 + ubuntu
  * Lint not ok on python 3.11 + macos
  * Maybe related to mypy 1.11.1 [just dropped a few days ago](https://pypi.org/project/mypy/1.11.1/) -- ? :thinking_face:
* Large-scale post-tag CI fail on [1.13.0rc0](https://github.com/single-cell-data/TileDB-SOMA/releases/tag/1.13.0rc0): https://github.com/single-cell-data/TileDB-SOMA/actions/runs/10257321585/job/28378154410

On the one hand this might seem to be a minor annoyance; on the other, it's clouding CI just when we're trying to make the TileDB-SOMA 1.13.0 release. Thus we should solve this sooner than later.

See also #2838 (just created) which has significant narrative by @ryan-williams on it.

Some background info (something that's been in place for a year now, I am reading up on now):

* https://github.com/single-cell-data/TileDB-SOMA/blame/1.13.0rc0/.github/workflows/python-ci-single.yml#L45-L49
* https://github.com/single-cell-data/TileDB-SOMA/pull/756

[[sc-52471]](https://app.shortcut.com/tiledb-inc/story/52471)

johnkerl:
Re-assigning to @ryan-williams

ryan-williams:
At a minimum, actions/setup-python is doing something buggy, I filed https://github.com/actions/setup-python/issues/919.

This issue is live on #2853, until we find a work-around, so I'm reopening it.

ryan-williams:
Issue [persists](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/10292424521/job/28486804717) with `setup-python` cache disabled (and no pre-commit cache found/restored), on both macos and ubuntu. Very strange.

ryan-williams:
I believe I've root-caused this:
- Our pandas dep ([in `setup.py`][pandas]) is unpinned, as is our `pandas-stub` ([in `.pre-commit-config.yaml`][pandas-stub]), meaning we can end up with Pandas 1.x or 2.x (and likewise for the "stub" types in mypy).
[...]

PR #2840, opened 2024-08-06
[python] Canary to check for mypy update

johnkerl:
For #2839. Here I have _no_ codemods and I'm solely seeing how CI has changed. Then on subsequent commits I'll experiment.


johnkerl:
Closing in favor of #2838 which @ryan-williams created concurrently

PR #2846, opened 2024-08-06
[python] X-write performance improvement for the non-append case

johnkerl:
Manual profiling on #2054 revealed an opportunity for a performance improvement of several seconds per X write on moderately sized ingests (thousands of cells).

### DIFF
--- a/apis/python/src/tiledbsoma/io/_registration/id_mappings.py
+++ b/apis/python/src/tiledbsoma/io/_registration/id_mappings.py
@@ -22,6 +22,12 @@ class AxisIDMapping:
     # Tuple not List so this can't be modified by accident when passed into some function somewhere
     data: Tuple[int, ...]
 
+    def is_identity(self) -> bool:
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/io/_registration/id_mappings.py
```diff
@@ -22,6 +22,12 @@ class tracks the mapping of input-data ``obs`` or ``var`` 0-up offsets to SOMA j
     # Tuple not List so this can't be modified by accident when passed into some function somewhere
     data: Tuple[int, ...]
 
+    def is_identity(self) -> bool:
+        for i in range(len(self.data)):
+            if self.data[i] != i:
+                return False
+        return True
```
nguyenv, line 29:
```suggestion
    def is_identity(self) -> bool:
        for i,data in enumerate(self.data):
            if data != i:
                return False
        return True
```

nguyenv (approved):

PR #2851, opened 2024-08-06
[c++] Only evolve schema when enumerations are extended

nguyenv:
**Issue and/or context:**

As caught be @eddelbuettel, the C++ write path was calling `SchemaEvolution.array_evolve` for every write. This call is only necessary if we have extended the enumerations. Otherwise, we end up bloating the `__schema` directory with unnecessary fragment files.

**Changes:**

- `SOMAArray` methods `_extend_enumeration`, `_extend_and_evolve_schema{,_str}`, and `_create_and_cast_column` return `bool` to indicate whether the column called `SchemaEvolution.extend_enumeration`
- Only call `SchemaEvolution.array_evolve` if any of the columns in the table called `SchemaEvolution.extend_enumeration`

**Notes for Reviewer:**

The test for this is written in Python as it is much more complicated to construct an `ArrowTable` with dictionary columns in C++.

### DIFF
--- a/apis/python/tests/test_dataframe.py
+++ b/apis/python/tests/test_dataframe.py
@@ -1589,9 +1589,9 @@ def test_only_evolve_schema_when_enmr_is_extended(tmp_path):
         data["bar"] = ["cat", "dog", "cat", "cat", "cat"]
         sdf.write(pa.Table.from_pydict(data))
 
-    # total 4 fragment files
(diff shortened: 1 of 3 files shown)


johnkerl (commented):
🔥

eddelbuettel (approved):
:tada:

PR #2854, opened 2024-08-08
[python] pin `pandas-stubs>=2` during pre-commit `mypy` hook

ryan-williams:
also simplify some `if TYPE_CHECKING` blocks

**Issue and/or context:** #2839, #2853

The unpinned `pandas-stubs` dep in `.pre-commit-config.yaml` allowed existing local environments (and old cached `setup-python` environments, which in some cases were being incorrectly used in our GHAs, per https://github.com/actions/setup-python/issues/919) to run the pre-commit `mypy` hook using `pandas-stubs<2`, where `pd.Series` is the type, and `pd.Series[Any]` requires a `# type: ignore[misc]` annotation.

With `pandas-stubs>=2` (which new environments have been picking up), `pd.Series[Any]` is the correct type, and the `# type: ignore[misc]` annotation was flagged as "unused" ([example](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/10291856865/job/28485041039?pr=2853#step:7:43)).

Pinning `pandas-stubs>=2`, and using the corresponding `pd.{Series,Index}[Any]` annotations when `TYPE_CHECKING`, should resolve this conflict, in both new and existing environments (in existing environments, `pre-commit` will recognize the new `pandas-stubs>=2` pin, and create a new venv that incorporates it).

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -5,10 +5,12 @@ on:
     paths:
       - '**'
       - '!**.md'
-      - '!apis/python/**'
-      - '!docs/**'
       - '!.github/**'
       - '.github/workflows/r-ci.yml'
+      - '!.pre-commit-config.yaml'
+      - '!apis/python/**'
+      - '!docs/**'
+      - '!LICENSE''
   push:
     branches:
       - main
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,10 @@ repos:
     hooks:
     - id: mypy
       additional_dependencies:
-        - pandas-stubs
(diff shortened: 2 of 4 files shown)


johnkerl (approved):
🔥

.github/workflows/r-ci.yml
```diff
@@ -5,10 +5,12 @@ on:
...
       - '**'
       - '!**.md'
-      - '!apis/python/**'
-      - '!docs/**'
       - '!.github/**'
       - '.github/workflows/r-ci.yml'
+      - '!.pre-commit-config.yaml'
+      - '!apis/python/**'
+      - '!docs/**'
+      - '!LICENSE''
```
ryan-williams, line 13: typo here, fix in #2856

PR #2871, opened 2024-08-09
[r] Expose timestamps publicly and plumb through for resume-mode

mojaveazure:
Expose timestamps publicly through a new active binding; replace calls to `private$tiledb_timestamp` with `self$tiledb_timestamp`

Also plumb timestamps through for `write_soma()` in resume-mode

### DIFF
--- a/apis/r/R/SOMASparseNDArray.R
+++ b/apis/r/R/SOMASparseNDArray.R
@@ -51,8 +51,8 @@ SOMASparseNDArray <- R6::R6Class(
                      config = cfg,
                      dim_points = coords,
                      result_order = result_order,
-                     timestamprange = if (is.null(private$tiledb_timestamp)) private$tiledb_timestamp
-                                      else c(0, private$tiledb_timestamp),
+                     timestamprange = if (is.null(self$tiledb_timestamp)) self$tiledb_timestamp
+                                      else c(0, self$tiledb_timestamp),
                      loglevel = log_level)
       private$ctx_ptr <- rl$ctx
       SOMASparseNDArrayRead$new(rl$sr, self, coords)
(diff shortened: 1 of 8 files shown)


apis/r/R/SOMASparseNDArray.R
```diff
@@ -51,8 +51,8 @@ SOMASparseNDArray <- R6::R6Class(
                      config = cfg,
                      dim_points = coords,
                      result_order = result_order,
-                     timestamprange = if (is.null(private$tiledb_timestamp)) private$tiledb_timestamp
-                                      else c(0, private$tiledb_timestamp),
+                     timestamprange = if (is.null(self$tiledb_timestamp)) self$tiledb_timestamp
+                                      else c(0, self$tiledb_timestamp),
```
eddelbuettel, line 55: Actually see the difference between this and the preceding block?  If we carry a two elem vector (or NULL) through we no longer need the if / else.  I just hadn't gotten to NDArray yet.  Likely cleanup?

apis/r/R/write_soma.R
```diff
@@ -466,7 +466,8 @@ write_soma.TsparseMatrix <- function(
     shape = shape %||% dim(x),
     ingest_mode = ingest_mode,
     platform_config = platform_config,
-    tiledbsoma_ctx = tiledbsoma_ctx
+    tiledbsoma_ctx = tiledbsoma_ctx,
+    tiledb_timestamp = Sys.time()
```
eddelbuettel, line 470: NULL may be better.  Let's core / libtiledbsoma do this.

mojaveazure: `NULL` wasn't working, setting to `Sys.time()` got the tests to pass

apis/r/R/SOMASparseNDArray.R
mojaveazure: All I've done is change all calls to `private$tiledb_timestamp` to `self$tiledb_timetamp`, leaving the code exactly as it was from your branch

eddelbuettel: Yeah. I was speaking half to myself here :)

eddelbuettel (approved):
Looks good, I can pick up the one now-bleeding new test from here.

Issue #2829, opened 2024-08-02
[Bug] Lossy ingest/outgest round-trips


ryan-williams:
(Factored out of #2804)

**Describe the bug**
Several categories of `pd.DataFrame` are modified during an ingest/outgest round-trip (e.g. if they are `obs` or `var` on an AnnData, ingested with `from_anndata`, and outgested with `to_anndata`).

**To Reproduce**
See [test_lossy_roundtrips_df.py].

**Versions (please complete the following information):**
 - TileDB-SOMA version: 1.12.3, seems like ingest/outgest have always worked this way though 
 - Language and language version: Python, all supported versions
 - OS: all

# Example lossy round-trips

(see [test_lossy_ingest_outgest_roundtrips.py][ROUND_TRIPS]; [obs-roundtrip.py] does similar and is used in the code snippets below)

<details>
<summary>
1. <code>df.index</code> is named "index":
<ul><li>
<code>df.index.name</code> is dropped (index becomes unnamed)
</li></ul>
</summary>

```bash
python obs-roundtrip.py -i index=xx,yy,zz -c col0=aa,bb,cc -c col1=AA,BB,CC 2>/dev/null
# Before:
#       col0 col1
# index
# xx      aa   AA
# yy      bb   BB
# zz      cc   CC
#
# SOMA DataFrame schema:
# soma_joinid: int64 not null
# obs_id: large_string
# col0: large_string
# col1: large_string
# metadata[soma_dataframe_original_index_name]:
# null
#
# After:
#    col0 col1
# xx   aa   AA
# yy   bb   BB
# zz   cc   CC
```
</details>

<details>
<summary>
2. DataFrame has a column named <code>obs_id</code>:
<ul>
<li><code>obs_id</code> column becomes unnamed index</li>
<li>Original <code>df.index</code> dropped</li>
</ul>
</summary>

```bash
python obs-roundtrip.py -i xx,yy,zz -c col0=aa,bb,cc -c obs_id=AA,BB,CC 2>/dev/null
# Before:
#    col0 obs_id
# xx   aa     AA
# yy   bb     BB
# zz   cc     CC
#
# SOMA DataFrame schema:
# soma_joinid: int64 not null
# col0: large_string
# obs_id: large_string
# metadata[soma_dataframe_original_index_name]:
# null
#
# After:
#    col0
# AA   aa
# BB   bb
# CC   cc
```
</details>

## DataFrame has a column named "index"
<details>
<summary>
[...]

PR #2804, opened 2024-07-22
[python] Ingest/outgest round-trip improvements

ryan-williams:
1. The first commit in this PR just adds [new unit tests][ROUND_TRIPS], which verify several categories of ingest/outgest "round-trip mutations."
   - See https://github.com/single-cell-data/TileDB-SOMA/issues/2829 for more details.
2. The second commit factors DF ingest/outgest logic currently ≈duplicated in 2 places each:
   - ingest: `ingest.py`/`signatures.py`
   - outgest: `obs`/`var`
   
   In the near future we'll want to use both for `uns` as well.
 
3. The third commit adds a 2nd set of tests, functionally identical to those added in 1., except they directly round-trip an "obs" DataFrame (using `_write_dataframe` and `_read_dataframe`; the latter is added in 2.).
   - The tests from 1. round-trip using `{from,to}_anndata`), and are still present.

[ROUND_TRIPS]: https://github.com/single-cell-data/TileDB-SOMA/blob/b967edabf1adb251d23c76297b626b6eb4d9f014/apis/python/tests/test_dataframe_io_roundtrips.py#L97-L154

PR #2873, opened 2024-08-09
[python] Update comments, remove unused test params

ryan-williams:
**Issue and/or context:** #2829, #2804, #2868

**Changes:**
- Tests introduced in #2804 included configurable `outgest_default_index_name` and `outgest_fallback_index_name` fields, that were never set to non-default values. Removing them here, to avoid confusion.
- Other comment updates / typo fixes

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -1145,14 +1145,17 @@ def _write_dataframe(
     axis_mapping: AxisIDMapping,
 ) -> DataFrame:
     """
-    The id_column_name is for disambiguating rows in append mode;
-    it may or may not be an index name in the input AnnData obs/var.
+    Convert and save a pd.DataFrame as a SOMA DataFrame.
 
-    The original_index_name is the index name in the AnnData obs/var.
+    This function adds a ``soma_joinid`` index to the provided ``pd.DataFrame``, as that is
+    required by libtiledbsoma; the original ``pd.DataFrame`` index is "reset" to a column, with its
+    original name stored as metadata on the SOMA ``DataFrame``.
 
-    This helper mutates the input dataframe, for parsimony of memory usage.
-    The caller should have copied anything pointing to a user-provided
-    adata.obs, adata.var, etc.
+    There are several subtleties related to naming the reset index column, and the ``id_column_name``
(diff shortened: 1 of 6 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1145,14 +1145,17 @@ def _write_dataframe(
     axis_mapping: AxisIDMapping,
 ) -> DataFrame:
     """
-    The id_column_name is for disambiguating rows in append mode;
-    it may or may not be an index name in the input AnnData obs/var.
+    Convert and save a pd.DataFrame as a SOMA DataFrame.
```
johnkerl, line 1148:
```suggestion
    Convert and save a ``pd.DataFrame`` as a SOMA ``DataFrame``.
```

And likewise throughout.

I may not have remember this on every PR review -- my apologies -- I don't always remember to flag it. But we _should_ always. Issue being our status-quo docsite is here

https://tiledbsoma.readthedocs.io/en/latest/

and we need RTD-style markup including double backticks

apis/python/tests/_util.py
```diff
@@ -1,12 +1,43 @@
...
+        return None, col_str.split(",")
+    elif len(pcs) == 2:
+        name, vals_str = pcs
+        vals = vals_str.split(",")
+        return name, vals
+    else:
+        raise ValueError(f"Invalid column string: {col_str}")
+
+
+def make_df(index_str: Optional[str] = None, **cols) -> pd.DataFrame:
```
johnkerl, line 25:
```suggestion
def make_pandas_dataframe(index_str: Optional[str] = None, **cols) -> pd.DataFrame:
```

johnkerl (changes requested):

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: BTW if we ever "miss any" I'm happy for one of us to put up an RTD-docstring-formats-only after-PR, no questions asked :)

apis/python/tests/_util.py
johnkerl: I don't like the short name but you're right it's the name as before, just moved from one file to another.

"df" is ambigious: there are Pandas and SOMA "dfs". I'll put up a separate PR.

johnkerl (approved):

PR #2872, opened 2024-08-09
[python] More AnnData equality assertions in tests

ryan-williams:
**Changes:**
- Introduce `assert_adata_equal` utilitiy
  - Replaces `verify_obs_and_var_eq`, `anndata_dataframe_unmodified`, `anndata_dataframe_unmodified_nan_safe`
- Use `assert_adata_equal` in tests
  - This surfaces I/O round-trip misbehavior in `test_basic_anndata_io.py::test_uns_io` (some covered by [#2829](https://github.com/single-cell-data/TileDB-SOMA/issues/2829), some specific to `uns` / not previously documented)
  - That test is updated here, to verify and document current behavior.
  - https://github.com/single-cell-data/TileDB-SOMA/compare/rw/eq...rw/uns includes follow-on work improving/fixing the issues.

#2874 builds on this, fixes some of the `uns` issues, and removes code added to `test_uns_io` here.

PR #667, opened 2023-01-07
[python] Complete basic anndata I/O tests

johnkerl:
Development on #664 revealed some underlying unit-test work that had been left incomplete. This PR fixes that problem.

PR #2874, opened 2024-08-09
[python] improve `uns` DataFrame I/O

ryan-williams:
**Issue and/or context:** builds on #2872

**Changes:**
- During `from_anndata` ingest, `copy` DataFrames supplied by user in `uns` before passing to `_write_dataframe`, to avoid mutating user-supplied objects
  - `obs` and `var` are already copied before being passed to `_write_dataframe`, by being passed through `conversions.decategoricalize_obs_or_var` ([example](https://github.com/single-cell-data/TileDB-SOMA/blob/main/apis/python/src/tiledbsoma/io/ingest.py#L512))
- During `to_anndata` outgest, use `_read_dataframe` on `uns` DataFrames, to better restore what was originally ingested
- Clean up tests (added in #2872) which codify previous behavior

### DIFF
--- a/apis/python/src/tiledbsoma/io/outgest.py
+++ b/apis/python/src/tiledbsoma/io/outgest.py
@@ -125,7 +125,9 @@ def _extract_X_key(
 
 
 def _read_dataframe(
-    df: DataFrame, default_index_name: Optional[str], fallback_index_name: str
+    df: DataFrame,
+    default_index_name: Optional[str] = None,
+    fallback_index_name: Optional[str] = None,
 ) -> pd.DataFrame:
     """Outgest a SOMA DataFrame to Pandas, including restoring the original index{,.name}.
 
@@ -175,7 +177,7 @@ def _read_dataframe(
         #
         # NOTE: several edge cases result in the outgested DF not matching the original DF; see
         # https://github.com/single-cell-data/TileDB-SOMA/issues/2829.
-        if fallback_index_name in pdf:
+        if fallback_index_name is not None and fallback_index_name in pdf:
             pdf.set_index(fallback_index_name, inplace=True)
             pdf.index.name = None
 
@@ -444,20 +446,18 @@ def _extract_uns(
             extracted[key] = _extract_uns(element, level=level + 1)
         elif isinstance(element, DataFrame):
             hint = element.metadata.get(_UNS_OUTGEST_HINT_KEY)
-            pdf = element.read().concat().to_pandas()
-            if hint is None:
-                extracted[key] = pdf
-            elif hint == _UNS_OUTGEST_HINT_1D:
+            if hint == _UNS_OUTGEST_HINT_1D:
+                pdf = element.read().concat().to_pandas()
                 extracted[key] = _outgest_uns_1d_string_array(pdf, element.uri)
             elif hint == _UNS_OUTGEST_HINT_2D:
+                pdf = element.read().concat().to_pandas()
                 extracted[key] = _outgest_uns_2d_string_array(pdf, element.uri)
             else:
-                msg = (
-                    f"Warning: uns {collection.uri}[{key!r}] has "
-                    + "{_UNS_OUTGEST_HINT_KEY} as unrecognized {hint}: leaving this as Pandas DataFrame"
-                )
-                logging.log_io_same(msg)
-                extracted[key] = pdf
+                if hint is not None:
+                    logging.log_io_same(
+                        f"Warning: uns {collection.uri}[{key!r}] has {_UNS_OUTGEST_HINT_KEY} as unrecognized {hint}: leaving this as Pandas DataFrame"
+                    )
+                extracted[key] = _read_dataframe(element, fallback_index_name="index")
         elif isinstance(element, SparseNDArray):
             extracted[key] = element.read().tables().concat().to_pandas()
         elif isinstance(element, DenseNDArray):
--- a/apis/python/tests/test_basic_anndata_io.py
+++ b/apis/python/tests/test_basic_anndata_io.py
@@ -15,7 +15,7 @@ from scipy.sparse import csr_matrix
 
 import tiledbsoma
 import tiledbsoma.io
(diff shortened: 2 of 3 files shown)


johnkerl (approved):
🚢

ryan-williams:
My idea was to merge #2872 first, then this PR's target would become `main`, then merge this.

apis/python/src/tiledbsoma/io/outgest.py
```diff
@@ -444,20 +446,18 @@ def _extract_uns(
...
-                    f"Warning: uns {collection.uri}[{key!r}] has "
-                    + "{_UNS_OUTGEST_HINT_KEY} as unrecognized {hint}: leaving this as Pandas DataFrame"
-                )
-                logging.log_io_same(msg)
-                extracted[key] = pdf
+                if hint is not None:
+                    logging.log_io_same(
+                        f"Warning: uns {collection.uri}[{key!r}] has {_UNS_OUTGEST_HINT_KEY} as unrecognized {hint}: leaving this as Pandas DataFrame"
+                    )
+                extracted[key] = _read_dataframe(element, fallback_index_name="index")
```
ryan-williams, line 460: @johnkerl just want to confirm: this change means outgested `uns` DataFrames no longer "gain" a `soma_joinid` column. The idea is to outgest DataFrames that match what was ingested (using the same function/logic that is applied to `obs`/`var`, `_read_dataframe`).

Does that seem right, or do we want to continue outgesting `uns` DataFrames containing a `soma_joinid` column that was not present on ingest?

apis/python/tests/test_basic_anndata_io.py
```diff
@@ -1047,24 +1035,6 @@ def test_uns_io(tmp_path, outgest_uns_keys):
 
     expected_adata = deepcopy(adata0)
 
-    # When outgesting `uns` DataFrames, `to_anndata` fails to remove the `soma_joinid` column added
```
ryan-williams, line 1050: N.B.: this block corresponds to the `_read_dataframe` line above. If we outgest the same thing we ingest, we no longer need the test to simulate the `soma_joinid` column addition.

apis/python/src/tiledbsoma/io/outgest.py
johnkerl: There have been asks in the past (maybe not well-tracked on my part) for:

* `soma_joinid` to not be included on outgest of `obs`/`var` -- because:
  * This is a SOMA join ID not an AnnData join ID
  * Gets us closer to round-trip fidelity
* `soma_joinid` to be included on outgest of `obs`/`var` -- because:
  * Customers wanting to see it, as they now consider it "part of the data" they're interested in

Some discussion on what to do about this -- again, not well-tracked on my part -- some in Slack/Shortcut/GitHub, I'll need to recover:

* Default exclude `soma_joinid` on outgest of `obs`/`var`; opt-in via some TBD yet-another-kwarg, or, some TBD config parameter
* Default include `soma_joinid` on outgest of `obs`/`var`;  opt-in via some TBD yet-another-kwarg, or, some TBD config parameter
* Maybe something else

This was never resolved AFAIK. I'll take it as an action item to track this.

Meanwhile let's be sure we're not changing the current behavior for `obs`/`var` -- the change we made to go from not outgesting `soma_joinid` (for non-uns), to outgesting `soma_joinid` (for non-uns) was per an explicit customer request. So we need to keep the current behavior for outgesting non-uns until we track this through product with @aaronwolen .

For outgesting `uns` dataframes (not `obs` or `var`) -- I believe you have the flexibility to do what you think is best, and to factor the code as you wish to (a) retain the current behavior for `obs`/`var` and (b) get what you think best for dataframes within `uns`.

TL;DR re your "Does that seem right" -- yes, but with the above just for complete clarity for all parties involved (not just you and me but everyone reading this).

apis/python/tests/test_basic_anndata_io.py
johnkerl: Yes this seems safe -- thank you.

I always get nervous changing existing unit-test cases so thank you again -- very much! -- for sequencing these PRs so it's clear to reviewer(s) including myself, and, anyone reading squash-merged PRs afterward (we have customers that read our GH like a newspaper :) ) , that we've made our lists and checked them twice, and that we are doing the right things regarding (a) important feature-improvement/refactor work, and (b) preserving tests which are our bulwark for maintaining customer-level expectations.

Again, I really appreciate the work you're doing here, and how you're sequencing it.

johnkerl (approved):

apis/python/src/tiledbsoma/io/outgest.py
ryan-williams: > the change we made to go from not outgesting `soma_joinid` (for non-uns), to outgesting `soma_joinid` (for non-uns) was per an explicit customer request

I'm confused by this… non-uns (`obs`/`var`) do **not** currently outgest a `soma_joinid`.

`obs_df.drop([SOMA_JOINID], …)` was added [in #667](https://github.com/single-cell-data/TileDB-SOMA/pull/667/files#diff-14a5396d5051d9449f58ae128b9e11f40bf6ef8b67e705a238cd08ba5afceb94R525) (Jan '23 / [1.0.0](https://github.com/single-cell-data/TileDB-SOMA/blob/1.0.0/apis/python/src/tiledbsoma/io/ingest.py#L1054-L1058)), and remains present [in 1.13.0](https://github.com/single-cell-data/TileDB-SOMA/blob/1.13.0/apis/python/src/tiledbsoma/io/outgest.py#L157).

This PR changes `uns` to drop `soma_joinid` on outgest, aligning `uns` with `obs`/`var`, and improving round-trip fidelity for `uns`. Existing `obs`/`var` behavior is preserved (i.e. `soma_joinid` is **removed** on outgest).

johnkerl: @ryan-williams thanks, my understanding was out of date.

Thanks for the history links!

PR #2876, opened 2024-08-12
[python] `update_uns_by_uri`, `_update_uns`

### LINKED ISSUE #1769: [python] `tiledbsoma.io.update_uns` phase one

ryan-williams:
## Issue and/or context
- Fixes #1769
- Extends #2892 

### [`_update_uns`][update_uns] helper
- Takes an already-open-for-write `Experiment`
- Mostly only supports writing keys that don't currently exist:
  - Existing scalars can be updated with other scalars.
  - If the existing or new value is an array or dataframe, the update is not performed
    - Overwriting scalars with arrays/dataframes should be support-able, can add here or in a future PR.
  - `strict` param controls behavior in the presence of unsupported updates:
    - Default (`strict=True`):
      - Performs a "dry run," which `raise`s on unsupported updates, but does not perform any writes (corresponds to `strict="dry_run"`)
      - If dry run succeeds, a second pass (with `strict="raise"`) performs the updates.
        - The expectation is nothing will be `raise`d by this pass.
      - This ensures updates are either all applied, or none are.
    - `strict="{debug,info,warn}"` can also be provided, in which case unsupported updates will be skipped, and a message logged at the specified level.

### `update_uns_by_uri`
Some more involved changes to an existing `uns` (e.g. updating dataframes/arrays) would require `open`ing the SOMA objects more than once, at different timestamps, meaning the user must provide a `uri: str` instead of an `Experiment.open(…, "w")`. Existing `update_{obs,var}` take an `Experiment`, so @aaronwolen was concerned about having something called `update_uns` that doesn't conform to the same API.

`update_uns_by_uri` takes a `uri: str`, but at the moment just `Experiment.open`s the URI, and passes that along to `_update_uns`, since we don't support any cases yet that require URI-based handling.

A possible downside of including this wrapper now is that we may want `update_uns_by_uri` to have a different API, when we actually support operations that require it. Open to discussion about which subset of `update_uns{,_by_uri}` should be included in the "first pass" here.

### Tests
[`test_update_uns.py`]: verifies several cases that succeed as written, as well as errors+messages for some cases that don't.

### Reusing "ingest" functions that may skip a write
With default params (`strict=True`), I've made a best effort to ensure all updates are applied, or none are. However, I delegate writing DataFrames and arrays (when nothing previously exists at that `uns` slot) to `ingest` functions ([`_write_soma_dataframe_impl`], [`_ingest_uns_array`]), but those functions will skip writing in some situations. Open to whether this "strictness" notion should be plumbed through general "ingest" functions as well…

### Breaking up `ingest.py`?
I've added the new "update `uns`" functionality in `io/update_uns.py`, instead of `io/ingest.py`. The latter is >2800 LOC, and may warrant factoring into more files, or perhaps an `ingest` sub-package?

I think most changes we might want to make can be done backwards-compatibly, e.g. with a `io/ingest/__init__.py` exposing symbols that previously lived in `ingest.py`, but may be moved to more specialized files in an `io/ingest/` package.

Not necessarily advocating doing this, just thinking through what options may exist, and trying not to inadvertently foreclose on things we might want to do in the future.

[`update_{obs,var}`]: https://github.com/single-cell-data/TileDB-SOMA/blob/b9ec3c5e8a89b3ed997f873772e0fb194eb0d34b/apis/python/src/tiledbsoma/io/ingest.py#L1367
[update_uns]: https://github.com/single-cell-data/TileDB-SOMA/pull/2876/files#diff-c82c57b8002263d8fc44e999a85fbd78b78eefcb6901799c368f0e9f44100472
[`test_update_uns.py`]: https://github.com/single-cell-data/TileDB-SOMA/pull/2876/files#diff-3d5d69b17928503976f475244d98ae2df94b7c62a454c36a7327b14a5643f666
[`_write_soma_dataframe_impl`]: https://github.com/single-cell-data/TileDB-SOMA/pull/2876/files#diff-dc1f1c04136350fc3134f40bd69db3715831b77579de8a425a1eacac03effcc7R1227
[`_ingest_uns_array`]: https://github.com/single-cell-data/TileDB-SOMA/pull/2876/files#diff-dc1f1c04136350fc3134f40bd69db3715831b77579de8a425a1eacac03effcc7R2622

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -1182,6 +1182,10 @@ def _write_dataframe_impl(
     platform_config: Optional[PlatformConfig] = None,
     context: Optional[SOMATileDBContext] = None,
 ) -> DataFrame:
+    """Save a Pandas DataFrame as a SOMA DataFrame.
+
+    Expects the required ``soma_joinid`` index to have already been added to the ``pd.DataFrame``.
+    """
     s = _util.get_start_stamp()
     logging.log_io(None, f"START  WRITING {df_uri}")
 
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -70,7 +70,6 @@ from .._types import (
     _INGEST_MODES,
     INGEST_MODES,
     IngestMode,
-    Metadatum,
     NPNDArray,
     Path,
     _IngestMode,
@@ -89,6 +88,7 @@ from ._common import (
     _UNS_OUTGEST_HINT_1D,
     _UNS_OUTGEST_HINT_2D,
     _UNS_OUTGEST_HINT_KEY,
+    AdditionalMetadata,
     Matrix,
     SparseMatrix,
     UnsMapping,
@@ -105,9 +105,6 @@ from ._util import get_arrow_str_format, read_h5ad
 
 _NDArr = TypeVar("_NDArr", bound=NDArray)
 _TDBO = TypeVar("_TDBO", bound=SOMAObject[RawHandle])
-
-
-AdditionalMetadata = Optional[Dict[str, Metadatum]]
 
 
 def add_metadata(obj: SOMAObject[Any], additional_metadata: AdditionalMetadata) -> None:
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -92,6 +92,7 @@ from ._common import (
     Matrix,
     SparseMatrix,
     UnsMapping,
+    UnsNode,
 )
 from ._registration import (
     AxisIDMapping,
@@ -2482,7 +2483,7 @@ def _maybe_ingest_uns(
 def _ingest_uns_dict(
     parent: AnyTileDBCollection,
     parent_key: str,
-    dct: Mapping[str, object],
+    dct: UnsMapping,
     *,
     platform_config: Optional[PlatformConfig],
     context: Optional[SOMATileDBContext],
@@ -2521,9 +2522,9 @@ def _ingest_uns_dict(
 
 
 def _ingest_uns_node(
-    coll: Any,
-    key: Any,
-    value: Any,
+    coll: AnyTileDBCollection,
+    key: str,
+    value: UnsNode,
     *,
     platform_config: Optional[PlatformConfig],
     context: Optional[SOMATileDBContext],
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -2576,36 +2576,57 @@ def _ingest_uns_node(
     if isinstance(value, list) or "numpy" in str(type(value)):
         value = np.asarray(value)
     if isinstance(value, np.ndarray):
-        if value.dtype.names is not None:
-            msg = f"Skipped {coll.uri}[{key!r}]" " (uns): unsupported structured array"
-            # This is a structured array, which we do not support.
-            logging.log_io(msg, msg)
-            return
-
-        if value.dtype.char in ("U", "O"):
-            # In the wild it's quite common to see arrays of strings in uns data.
-            # Frequent example: uns["louvain_colors"].
-            _ingest_uns_string_array(
-                coll,
-                key,
-                value,
-                use_relative_uri=use_relative_uri,
-                **ingest_platform_ctx,
-            )
-        else:
-            _ingest_uns_ndarray(
-                coll,
-                key,
-                value,
-                use_relative_uri=use_relative_uri,
-                **ingest_platform_ctx,
-            )
+        _ingest_uns_array(
+            coll,
+            key,
+            value,
+            use_relative_uri=use_relative_uri,
+            ingest_platform_ctx=ingest_platform_ctx,
+        )
         return
 
     msg = (
         f"Skipped {coll.uri}[{key!r}]" f" (uns object): unrecognized type {type(value)}"
     )
     logging.log_io(msg, msg)
+
+
+def _ingest_uns_array(
+    coll: AnyTileDBCollection,
+    key: str,
+    value: NPNDArray,
+    use_relative_uri: Optional[bool],
+    ingest_platform_ctx: IngestPlatformCtx,
+) -> None:
+    """Ingest an uns Numpy array.
+
+    Delegates to :func:`_ingest_uns_string_array` for string arrays, and to
+    :func:`_ingest_uns_ndarray` for numeric arrays.
+    """
+    if value.dtype.names is not None:
+        # This is a structured array, which we do not support.
+        logging.log_io_same(
+            f"Skipped {coll.uri}[{key!r}]" " (uns): unsupported structured array"
+        )
+
+    if value.dtype.char in ("U", "O"):
+        # In the wild it's quite common to see arrays of strings in uns data.
+        # Frequent example: uns["louvain_colors"].
+        _ingest_uns_string_array(
+            coll,
+            key,
+            value,
+            use_relative_uri=use_relative_uri,
+            **ingest_platform_ctx,
+        )
+    else:
+        _ingest_uns_ndarray(
+            coll,
+            key,
+            value,
+            use_relative_uri=use_relative_uri,
+            **ingest_platform_ctx,
+        )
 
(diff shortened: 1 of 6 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1473,14 +1496,144 @@ def update_var(
     )
 
 
+def update_uns(
+    exp: Experiment,
```
johnkerl, line 1500: I know this is a draft PR, but I don't want you to have to do time-consuming rework.

As in our internal design document, this _must_ take a URI not an experiment opened for write. Even if it works for you for step 1 of 5 of planned `update_uns` work. I'm happy to review previous conversations on this topic and/or the design document, offline.

ryan-williams: The doc says a function that takes a URI would need a different name (than `update_uns`), to avoid confusion vs. existing `update_{obs,var}`.

If simple cases can be supported using this API (which matches existing `update_{obs,var}`), that seems worthwhile, and doesn't preclude supporting a URI-based helper, down the road, for changes that require multiple `open`s.

That was my thinking, anyway. Can discuss more offline.

johnkerl: Thanks @ryan-williams . OK to discuss here as well, for transparency with our stakeholders.

* There are things (documented as above) which cannot be accommodated in a function which takes an experiment already opened for write. So we'll need that second function someday anyway.
* If we make an `update_uns` that takes an experiment opened for write, and later a `update_uns_from_uri` (or whatever name) later, then:
  * They will both have the same semantics, which is "make the TileDB-SOMA `uns` look like this new AnnData `uns` as much as you are capable"
  * We can keep around the `update_uns` and it will need to start throwing, for the things it can't do -- this means users will have to modify their notebooks/pipelines/etc to call the new function
  * Or, we go through a deprecation cycle and ask users to plan for the change, letting them know `update_uns` will be going away -- this means users will have to modify their notebooks/pipelines/etc to call the new function
* Or, we make an `update_uns_from_uri` (or whatever name) which takes an experiment URI and we simply do the opens within that function (as sketched in the above-referenced doc)
  * There will be one function with the semantics "make the TileDB-SOMA `uns` look like this new AnnData `uns` as much as you are capable"
  * The list of things of which it is capable will grow over time
  * No need for two not-quite the same functions, and/or deprecation cycle

@aaronwolen -- thoughts from the product POV?

ryan-williams: (Had a rewrite of the PR description saved as a draft; just submitted it, for reference)

One note on the above:

> We can keep around the `update_uns` and it will need to start throwing, for the things it can't do, that the other one can do

`update_uns` in this PR already throws for the things it can't do; it won't have to "start throwing." If anything, it can be made to support some cases that it doesn't currently (which don't require a URI / multiple `open`s). Cases that require a URI will necessarily be implemented only in the URI-based version, and would continue to throw in the `update_uns` introduced here.

Agreed that having two similar functions (one taking a URI, and handling a superset of cases) could be confusing. OTOH, it could be useful to refrain from committing to an API for the URI-based version now, in case we end up wanting a different API, when we get around to supporting those cases. This `update_uns` matches analogous `update_{obs,var}`, and handles some proportion of the cases we've heard requests for.

apis/python/src/tiledbsoma/io/update_uns.py
```diff
@@ -0,0 +1,242 @@
...
+            use_relative_uri=use_relative_uri,
+            context=context,
+            additional_metadata=additional_metadata,
+            platform_config=platform_config,
+            default_index_name=default_index_name,
+            strict=strict,
+        )
+
+
+def update_uns(
```
johnkerl, line 62: This is going to need to be `_update_uns`, otherwise it will be public API.

ryan-williams: done

apis/python/src/tiledbsoma/io/ingest.py
ryan-williams: The two functions are now `update_uns_by_uri` and `_update_uns` ("private"). I've reworked the PR description as well to reflect that.

apis/python/src/tiledbsoma/io/update_uns.py
johnkerl: Thanks!

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: Thanks!

ryan-williams: We might have landed in an OK place, that reflects the doc… @johnkerl wdyt? I will rename the functions or fold `_update_uns` into `update_uns_by_uri`, if that is still what you think is best.

ryan-williams:
@johnkerl just flagging a couple topics I've added to "Notes for Reviewers" in the PR description, in case you have thoughts on them.

johnkerl (commented):
More comments coming; just 'poking save' on the ones I have already

apis/python/src/tiledbsoma/io/update_uns.py
```diff
@@ -0,0 +1,242 @@
...
+        be opened for write.
+
+        measurement_name: Specifies which measurement's ``uns`` within the experiment
+        is to be updated.
+
+        uns: a Pandas dataframe with the desired contents.
+
+        use_relative_uri: If True, store the URI relative to the experiment's URI.
+
+        context: Optional :class:`SOMATileDBContext` containing storage parameters, etc.
```
johnkerl, line 95: double backticks please

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1218,14 +1220,37 @@ def _write_dataframe_impl(
             if _chunk_is_contained_in(dim_range, storage_ned):
                 logging.log_io(
                     f"Skipped {df_uri}",
-                    _util.format_elapsed(s, f"SKIPPED {df_uri}"),
+                    _util.format_elapsed(start, f"SKIPPED {df_uri}"),
                 )
                 return soma_df
 
+    return _write_soma_dataframe_impl(
```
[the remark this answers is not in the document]
johnkerl: ✨

johnkerl (commented):
Forgot to poke 'submit' on some comments from a few days ago -- coming here how & seeing there were still "unsubmitted" :| ...

apis/python/src/tiledbsoma/io/update_uns.py
johnkerl: Derp yes, good call @ryan-williams , thanks!

johnkerl:
> @johnkerl just flagging a couple topics I've added to "Notes for Reviewers" in the PR description, in case you have thoughts on them.

@ryan-williams thanks!

Yes, `ingest.py` is big -- it was already split out into `ingest.py` and `outgest.py` for the same reason, and KLOC is indeed a factor, and splitting it more (e.g. `io/ingest/` and `io/outgest/`) is fine with me.

Re

> Open to whether this "strictness" notion should be plumbed through general "ingest" functions as well…

I don't see a pressing need at this particular moment; let's keep this in mind though. All else being equal I'd rather leave the non-update-uns UX unmodified by the current update-uns work.

apis/python/src/tiledbsoma/io/update_uns.py
```diff
@@ -0,0 +1,242 @@
...
+    measurement_name: str,
+    *,
+    use_relative_uri: Optional[bool] = None,
+    context: Optional[SOMATileDBContext] = None,
+    additional_metadata: AdditionalMetadata = None,
+    platform_config: Optional[PlatformConfig] = None,
+    default_index_name: Optional[str] = None,
+    strict: Strict = True,
+) -> None:
+    """Wrapper around :func:`update_uns` that opens the experiment at the given URI.
```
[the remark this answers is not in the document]
ryan-williams: done

johnkerl (approved):

PR #2752, opened 2024-06-18
[python] Remove tiledb-py dependency

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/1632

This completes the SOMA Python API move over from tiledb-py to C++.

**Changes:**

- Remove `tiledb.ctx` usage
- Remove `tiledb.Filter` usage

PR #2883, opened 2024-08-12
[python] Remove tiledb-py from unit tests

nguyenv:
**Issue and/or context:**

This PR is separated out from https://github.com/single-cell-data/TileDB-SOMA/pull/2752

**Changes:**

- Addition of `config_options_from_schema` binding to retrieve `ArraySchema` info not encapsulated by the `ArrowSchema`
- Corrected JSON `dims` and `attrs` by returning `tile` for `dims` and using `filters` keys for `FilterList` values
- Replace `tiledb.FilterList` with `PlatformConfig`'s JSON formatted strings
- Add `is_relative` for `SOMAGroup`s
- Tests that checked fragment info count (either with `VFS` or `FragmentInfoList`) now use `libpath.Path`'s `iterdir`

### DIFF
--- a/apis/python/tests/test_basic_anndata_io.py
+++ b/apis/python/tests/test_basic_anndata_io.py
@@ -19,14 +19,6 @@ from tiledbsoma import Experiment, _constants, _factory
 from tiledbsoma._soma_object import SOMAObject
 from tiledbsoma.io._common import _TILEDBSOMA_TYPE, UnsDict, UnsMapping
 
-try:
-    import tiledb
-
-    hastiledb = True
-except ModuleNotFoundError:
-    hastiledb = False
-
-
 from ._util import TESTDATA, assert_adata_equal, make_pd_df
 
 
@@ -283,7 +275,6 @@ def _get_fragment_count(array_uri):
         TESTDATA / "pbmc-small-x-csc.h5ad",
     ],
 )
-@pytest.mark.skipif(not hastiledb, reason="tiledb-py not installed")
 def test_resume_mode(resume_mode_h5ad_file):
     """
     Makes sure resume-mode ingest after successful ingest of the same input data does not write
@@ -342,7 +333,6 @@ def test_resume_mode(resume_mode_h5ad_file):
 
 
 @pytest.mark.parametrize("use_relative_uri", [False, True, None])
-@pytest.mark.skipif(not hastiledb, reason="tiledb-py not installed")
 def test_ingest_relative(conftest_pbmc3k_h5ad_path, use_relative_uri):
     tempdir = tempfile.TemporaryDirectory()
     output_path = tempdir.name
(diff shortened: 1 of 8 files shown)


johnkerl (changes requested):
@nguyenv thanks!!!

I'm putting this as request-changes just to put a "hold" on it -- no indication anything's necessarily wrong but rather to flag some things for conversation:

* Will we be doing pip-install of tiledb-py within our CI jobs?
  * If so: may need some (light) handling around core versions in the case when
    * Core has bumped storage-format version
    * TileDB-Py has that new core available at PyPI and TileDB-SOMA Py does not (or vice versa)
  * Also if so: that means the tests you have here may go "uncovered" for a while until we restore pip-install of tiledb-py within our CI YAML -- ok as long as we understand/track what we're doing
  * If not:
    * We'll be losing test coverage of assertions we once made and now no longer can
    * How bad is this??

Since the core C++ API (and C API) are "full APIs" that means anything we can/do do in TileDB-Py we should be able to do ourselves -- I want us to at least have the _conversation_ about what to do with things like

```
            assert A.schema.sparse
            assert not A.schema.allows_duplicates
            assert A.dim("foo").filters == [tiledb.ZstdFilter(level=3)]
            assert A.attr("bar").filters == [tiledb.ZstdFilter()]
```

-- namely: can we either:

* Expose some of these things via `libtiledbsoma` -> `pybind11`/`Rcpp` -> `tiledbsoma-py`/`tiledbsoma-r` -> so we can still test them at the `pytest` level?
* Or, maybe expose something only at the `libtiledbsoma` level (no bindings) and then move these test cases from `pytest` (or `testthat` for R) to `ctest` for `libtiledbsoma`

nguyenv:
My idea for the latter conerns: When creating a `SOMAArray` in C++, we can store the `PlatformConfig` that's passed in to generate the `ArraySchema`. Then have a `platform_config` getter that we can bind in the APIs. In Python, we could return that as a JSON formatted string.

nguyenv:
I think we should install tiledb-py when running CI. We could have a `requirements_ci.txt` that pulls everything from `requirements_dev.txt` with just the additional requirement for `tiledb`. But yes we'd have to make sure the version of tiledb-py it pulls is compatible with the libtiledbsoma in tiledbsoma-py. I don't know how to deal with that yet but can look into it.

johnkerl:
> My idea for the latter conerns: When creating a SOMAArray in C++, we can store the PlatformConfig that's passed in to generate the ArraySchema. Then have a platform_config getter that we can bind in the APIs. In Python, we could return that as a JSON formatted string.
> 
> So essentially your first bullet point.

Hmm, I hadn't thought of that. TBH I don't think this is a good check -- it will be checking a copy of platform config stored somewhere, not checking against readback of the actual core schema that was created. We need to actually read the core schema and do these checks -- if we want to continue doing them at all ...

nguyenv:
All changes in this branch have been split out into multiple PRs and merged into main.

PR #2884, opened 2024-08-13
[c++] Add `dataset_type` into `SOMAExperiment` metadata

nguyenv:
Tracking: #2885 

As caught interally, the `dataset_type` needs to be set to `soma` for `SOMAExperiment` on tiledb-cloud upon creation.

### DIFF
--- a/libtiledbsoma/src/soma/soma_group.cc
+++ b/libtiledbsoma/src/soma/soma_group.cc
@@ -67,6 +67,14 @@ std::unique_ptr<SOMAGroup> SOMAGroup::create(
             static_cast<uint32_t>(ENCODING_VERSION_VAL.length()),
             ENCODING_VERSION_VAL.c_str());
 
+        // Root SOMA objects include a `dataset_type` entry to allow the
+        // TileDB Cloud UI to detect that they are SOMA datasets.
+        std::string cloud_scheme = "tiledb://";
+        if (soma_type == "SOMAExperiment" &&
+            uri.compare(0, cloud_scheme.length(), cloud_scheme) == 0) {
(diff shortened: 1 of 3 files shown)


libtiledbsoma/src/soma/soma_group.cc
```diff
@@ -67,6 +67,14 @@ std::unique_ptr<SOMAGroup> SOMAGroup::create(
             static_cast<uint32_t>(ENCODING_VERSION_VAL.length()),
             ENCODING_VERSION_VAL.c_str());
 
+        // Root SOMA objects include a `dataset_type` entry to allow the
+        // TileDB Cloud UI to detect that they are SOMA datasets.
+        std::string cloud_scheme = "tiledb://";
```
johnkerl, line 72: Let's go ahead & put this regardless of URI -- as it was before. People can ingest directly to TileDB Cloud, or, ingest to S3 and register it.

There's no harm in writing it unconditionally, and value to be gained.

johnkerl (changes requested):
Thanks!

Please add a unit-test case which creates an `Experiment` and asserts `dataset_type=="soma"` is in its metadata. (In hindsight, we should have done this before ...)

jp-dark (approved):
LGTM other than John's comment about doing it regardless of URI. Do we want to also remove the `_set_create_metadata` methods from `Experiment` and `SOMAObject` during this or a follow-on PR?

eddelbuettel (approved):
LGTM

johnkerl (approved):
🚀

johnkerl:
> Good point. I will remove those in a follow-up PR.

@nguyenv's going on well-deserved PTO so I've put up #2887

PR #2891, opened 2024-08-14
[c++] Remove a temporary `gdb` hook from #2888

johnkerl:
Follow-up from https://github.com/single-cell-data/TileDB-SOMA/pull/2888#discussion_r1716813679

I had used this to make it easier to get a breakpoint in the right spot in `gdb` while debugging #2888. I neglected to take it back out.


ryan-williams (approved):

eddelbuettel (approved):
:shipit:

PR #2911, opened 2024-08-17
[c++] Arrow utils with current-domain option

johnkerl:
**Issue and/or context:** This is a line-count-reducing under-diff split out from #2785 for issue #2407 / [[sc-51048]](https://app.shortcut.com/tiledb-inc/story/51048). It will help reduce distractions on upcoming review of more substantial PRs -- although this PR has a bit more substance than #2908, #2909, and #2910.

**Changes:**

* `create_arrow_schema` and `create_index_column_info` take flags for status quo (`use_current_domain=false`) as well as new-shape (`use_current_domain=true` -- #2407) . These flags will allow for a reasoned, well-tested transition to new-shape functionality at the R and Python levels, implemented on upcoming PRs.
* For knowledge management, I'll introduce some comment blocks about how these two methods work. They're an ingenious technique for sharing schema information in a language-neutral way between R/C++ and Python/C++.
* Unit-test support on this PR is mainly confined to demonstrating that no backward compatibility is broken: the Python and R APIs continue to work as they have before.
* Unit-test cases at the `libtiledbsoma` level will validate correctness of the libtiledbsoma-to-core contracts, thereby increasing confidence of the correctness of the libtiledbsoma-level implementation.
* Further unit-test expansion on subsequent PRs will introduce the `resize` logic to `libtiledbsoma`, and unit-test its correctness. However, this PR is already more than big enough as-is.

**Notes for Reviewer:**

As of 2024-08-18, this is stacked on top of #2910. #2910 should be approved and merged first, then, this should be rebased on the new `main`. (What should not happen is for this PR to be merged into #2910.)

As of 2024-08-18 this incorporates #2915 without being stacked on top of it. Here, too, #2915 should be approved and merged first, then, this should be rebased on the new `main`.

#2785 is unreviewable as-is: besides being a WIP, it's an all-in-one experimental area. Smaller PRs such as this one are being offered for manageable review.


eddelbuettel:
A more general comment orthogonal to this PR:  we really should think about making more use of nanoarrow (as I have) to remove most if not all of this arrow hand-holding boilerplate code.

johnkerl:
Thanks @eddelbuettel . That's a good idea for another PR. At the moment I want to separate refactors / dependency changes from the  careful sequencing of steps necesary for confident rollout of the current-domain-backed new-shape feature. Introducing nanoarrow here would be too many variables: a dependent-package change, along with the current-domain support. As follow-on -- separately -- yes, I am behind this. I'll file an issue to track this great idea.

libtiledbsoma/test/common.cc
```diff
@@ -33,11 +33,12 @@
...
 
 namespace helper {
-ArraySchema create_schema(Context& ctx, bool allow_duplicates) {
+ArraySchema create_schema(
+    Context& ctx, int64_t dim_max, bool allow_duplicates) {
     // Create schema
     ArraySchema schema(ctx, TILEDB_SPARSE);
 
-    auto dim = Dimension::create<int64_t>(ctx, "d0", {0, 1000});
+    auto dim = Dimension::create<int64_t>(ctx, "d0", {0, dim_max});
```
johnkerl, line 41: This is atop #2910, which should be approved/merged first, and then this distraction will rebase away.

libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -1178,6 +1178,25 @@ uint64_t SOMAArray::nnz_slow() {
...
+    }
+}
+
+std::vector<int64_t> SOMAArray::maxshape() {
+    return _tiledb_domain();
+}
+
+std::vector<int64_t> SOMAArray::_tiledb_domain() {
     std::vector<int64_t> result;
     auto dimensions = mq_->schema()->domain().dimensions();
```
johnkerl, line 1202: This is old code, but incorrect: the return value is `std::vector<int64_t>` but it accepts switch statements for other types.

Meanwhile, in the SOMA data model:

* `SparseNDArray` has dims which are all necessarily `int64_t`
* `DenseNDArray` has dims which are all necessarily `int64_t`
* `DataFrame` _default_ indexing is one dim

johnkerl (commented):
Comment here:
https://github.com/single-cell-data/TileDB-SOMA/pull/2911/files#r1721037749

libtiledbsoma/src/soma/soma_array.cc
johnkerl: Split out as #2915.

libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -1178,81 +1178,116 @@ uint64_t SOMAArray::nnz_slow() {
...
+        // * DenseNDArray has dims which are all necessarily int64_t
+        // * DataFrame _default_ indexing is one dim named "soma_dim_0" of type
+        //   int64_t, however:
+        //   * Users can (and do) add other additional dims
+        //   * The SOMA data model requires that soma_joinid be present in each
+        //     DataFrame either as a dim or an attr -- and there are DataFrame
+        //     objects for which soma_joinid is not a dim at all
+        //   * These cases are all actively unit-tested within apis/python/tests
+        if (dim.type() != TILEDB_INT64) {
+            throw TileDBSOMAError("Found unexpected non-int64 dimension type.");
```
eddelbuettel, line 1217: I know where you are coming from here, and I slept over it one night, but I am not sure I am fully on-board. 

We have the factory methods.  We have the SOMA definitions.  It is clear what a well-formed SOMA object is.  

But I  still cannot shake the feeling that this seems "wrong" in the sense of working, testing, testable, useful, ... general code.  

I'll sleep over it another night.  Not objecting to the change, but scratching my chin.  :thinking:

johnkerl: The removed code is dead ... unit tests would be failing if it were not. :)

eddelbuettel: That was not really my point / I understand that.  But such a converter is also useful outside of SOMA, the test is fine, and I have tested and developed too with non-soma dataframes.  That can be useful.  

Again, not vetoing anything here, just saying that I simultaneously understand why you propose what you and that I likely would not do it.  No more, no less, and surely no veto.

eddelbuettel: Also "we can remove because doing so does not break tests" is not really a viable rule because we all know that coverage is never perfect.  Anyway ...

libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -1183,81 +1183,116 @@ uint64_t SOMAArray::nnz_slow() {
...
+void SOMAArray::resize(const std::vector<int64_t>& newshape) {
+    if (mq_->query_type() != TILEDB_WRITE) {
+        throw TileDBSOMAError(
+            "[SOMAArray::resize] array must be opened in write mode");
+    }
+
+    auto tctx = ctx_->tiledb_ctx();
+    ArraySchema schema = arr_->schema();
+    Domain domain = schema.domain();
+    ArraySchemaEvolution schema_evolution(*tctx);
```
nguyenv, line 1269: Will we need to use a timestamp with this (when `_make_se` gets reintroduced)?

libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -1183,81 +1183,116 @@ uint64_t SOMAArray::nnz_slow() {
...
+    //   new-shape feature
+    // * Even after the new-shape feature is fully released, there will be old
+    //   arrays on disk that were created before this feature existed.
+    // So this is long-term code.
+    auto current_domain = _get_current_domain();
+    if (current_domain.is_empty()) {
+        return _tiledb_domain();
+    } else {
+        return _tiledb_current_domain();
+    }
```
nguyenv, line 1197:
```suggestion
    return current_domain.is_empty() ? _tiledb_domain() : _tiledb_current_domain();
```

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -586,6 +746,56 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
...
+                if (ArrowAdapter::_isvar(child->format)) {
+                    // In the core API:
+                    //
+                    // * domain for strings must be set as (nullptr, nullptr)
+                    // * current_domain for strings cannot be set as (nullptr,
+                    //   nullptr)
+                    //
+                    // Fortunately, these are ASCII dims and we can range
+                    // these accordingly.
+                    ndrect.set_range(col_name, "\x01", "\x7f");
```
nguyenv, line 779: Can you add a note too that this range sets all printable ASCII characters (I had to look it up)?

libtiledbsoma/test/unit_soma_collection.cc
```diff
@@ -49,43 +49,49 @@ TEST_CASE("SOMACollection: basic") {
...
-    REQUIRE(soma_sparse->ndim() == 1);
-    REQUIRE(soma_sparse->nnz() == 0);
-    REQUIRE(soma_sparse->timestamp() == ts);
-    soma_sparse->close();
-    soma_collection->close();
-
-    soma_collection = SOMACollection::open(base_uri, OpenMode::read, ctx);
-    REQUIRE(soma_collection->members_map() == expected_map);
-    soma_collection->close();
+    bool use_current_domains[] = {false, true};
```
nguyenv, line 52: Catch2 has a way to parameterize these options so that they run as separate unit tests. Here is an example from the `SOMAArray` tests: https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/test/unit_soma_array.cc#L177-L190

libtiledbsoma/src/soma/soma_array.cc
johnkerl: Most definitely -- this is an old pre-PTO PR, needing rebasing ...

libtiledbsoma/src/utils/arrow_adapter.cc
johnkerl: Oh nice -- sure! :)

libtiledbsoma/src/soma/soma_array.cc
johnkerl: But definitely, seeing this callsite (another construction of `ArraySchemaEvolution`) is exactly what led me to realize we needed to extract the helper function `_make_se`.

libtiledbsoma/src/utils/arrow_adapter.cc
johnkerl: Pushed

nguyenv (approved):

libtiledbsoma/test/unit_soma_collection.cc
johnkerl: Commit 3 https://github.com/single-cell-data/TileDB-SOMA/pull/2911/commits/c6f7f730b9400ca55cc5e45c0308cc886d990e10 -- thank you :)

PR #2916, opened 2024-08-18
[c++] Performant `DataFrame.shape`

johnkerl:
**Issue and/or context:** This follows on #2944 for issue #2407 / [[sc-51048]](https://app.shortcut.com/tiledb-inc/story/51048). Here we implement `shape` for SOMA `DataFrame`, including its variant-indexed flavors.

**Changes:**

While `SOMASparseNDArray` and `SOMADenseNDArray` must always and only have `int64` dims within the [SOMA data model](https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md), `SOMADataFrame` is different. The default behavior -- which almost everyone uses -- has a single `soma_joinid` dim which is indeed of type `int64`. (Also note that `exp.obs.shape` _does_ exist within TileDB-SOMA-Py, and people _do_ call it.) However, the spec only requires that `soma_joinid` exist as a dim or an attr: it can be a dim along with others, or it can not be a dim at all. Here we do the right thing, without and with current-domain support, to produce the user-expected `exp.obs.shape`.

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/soma/soma_array.h
+++ b/libtiledbsoma/src/soma/soma_array.h
@@ -804,8 +804,8 @@ class SOMAArray : public SOMAObject {
     }
 
    protected:
-    // For use nominally by SOMADataFrame. This could be moved in its entirety to
-    // SOMADataFrame, but it would entail moving several SOMAArray attributes
+    // For use nominally by SOMADataFrame. This could be moved in its entirety
+    // to SOMADataFrame, but it would entail moving several SOMAArray attributes
     // from private to protected, which has knock-on effects on the order of
     // constructor initializers, etc.: in total it's simplest to place this
     // here and have SOMADataFrame invoke it.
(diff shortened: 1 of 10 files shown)


libtiledbsoma/src/soma/soma_array.h
```diff
@@ -803,6 +803,14 @@ class SOMAArray : public SOMAObject {
         return _get_current_domain();
     }
 
+   protected:
+    // For use nominally by SOMADataFrame. This could be moved in its entrety to
```
nguyenv, line 807:
```suggestion
    // For use nominally by SOMADataFrame. This could be moved in its entirety to
```

libtiledbsoma/src/soma/soma_array.h
```diff
@@ -803,6 +803,14 @@ class SOMAArray : public SOMAObject {
...
+    // SOMADataFrame, but it would entail moving several SOMAArray attributes
+    // from private to protected, which has knock-on effects on the order of
+    // constructor initializers, etc.: in total it's simplest to place this
+    // here, and have SOMADataFrame invoke it.
```
nguyenv, line 811:
```suggestion
    // here and have SOMADataFrame invoke it.
```

nguyenv (approved):

PR #2895, opened 2024-08-14
[c++] Apply `ArraySchemaEvolution` to specified timestamp

johnkerl:
**Issue and/or context:** #2879 

**Changes:** If we've got the array open at a timestamp, and we need to evolve it, evolve from that same timestamp.

**Notes for Reviewer:**

Material from @nguyenv to be incorporated into unit-test material:

**Script:**

```
$ cat ts.py
import tiledb
import tiledbsoma as soma
import pyarrow as pa
import tempfile
import numpy as np
import pandas as pd
import time

with tempfile.TemporaryDirectory() as uri:
    asch = pa.schema([("foo", pa.dictionary(pa.int8(), pa.large_string()))])
    soma.DataFrame.create(uri, schema=asch, tiledb_timestamp=1).close()

    pydict = {}
    pydict["soma_joinid"] = [0, 1, 2]
    pydict["foo"] = pd.Series(['a', 'b', 'a'], dtype='category')
    rb = pa.Table.from_pydict(pydict)

    with soma.DataFrame.open(uri, "w", tiledb_timestamp=2) as sdf:
        sdf.write(rb)

    pydict = {}
    pydict["soma_joinid"] = [3, 4]
    pydict["foo"] = pd.Series(['b', 'b'], dtype='category')
    rb = pa.Table.from_pydict(pydict)

    with soma.DataFrame.open(uri, "w", tiledb_timestamp=3) as sdf:
        sdf.write(rb)

    with soma.DataFrame.open(uri) as sdf:
        table = sdf.read().concat()
        print(table)

    with soma.DataFrame.open(uri, tiledb_timestamp=1) as sdf:
        table = sdf.read().concat()
        print(table)

    with soma.DataFrame.open(uri, tiledb_timestamp=2) as sdf:
        table = sdf.read().concat()
        print(table)

    with soma.DataFrame.open(uri, tiledb_timestamp=3) as sdf:
        table = sdf.read().concat()
        print(table)
```

**Before:**

```
pyarrow.Table
soma_joinid: int64
foo: dictionary<values=string, indices=int8, ordered=0>
----
soma_joinid: [[0,1,2,3,4]]
foo: [  -- dictionary:
["a","b"]  -- indices:
[0,1,0,0,0]]
pyarrow.Table
soma_joinid: int64
foo: dictionary<values=string, indices=int8, ordered=0>
----
soma_joinid: [[]]
foo: [  -- dictionary:
[]  -- indices:
[]]
pyarrow.Table
soma_joinid: int64
[...]

PR #2921, opened 2024-08-21
Revert #2895

ryan-williams:
**Issue and/or context:** #2920

[cellxgene_census_builder//test_builder.py](https://github.com/chanzuckerberg/cellxgene-census/blob/main/tools/cellxgene_census_builder/tests/test_builder.py) started failing at #2895: 
- [GHA failure](https://github.com/ryan-williams/tiledb-scratch/actions/runs/10482441869/job/29033553412) at [7241fef7](https://github.com/single-cell-data/TileDB-SOMA/commit/7241fef7)
- [GHA success](https://github.com/ryan-williams/tiledb-scratch/actions/runs/10482609733/job/29034000518) at previous commit [ab0a5b59](https://github.com/single-cell-data/TileDB-SOMA/commit/ab0a5b59)

Originally reported by @ivirshup.

**Changes:**

Revert #2895 (as well as some bits of #2897 and #2909 that were merged since). `test_builder.py` [passes again](https://github.com/ryan-williams/tiledb-scratch/actions/runs/10489797355/job/29055198008) with this change.

**Notes for Reviewer:**

I'm not clear on the significance of including (and now removing) the `TimestampRange` in `soma_array.cc`; hoping someone can confirm they understand why it was causing an issue, and that it's fine to (temporarily?) revert.


johnkerl:
#2895 was crucial for 1.13.1 -- so we need to deep-dive the consolidate/vacuum failure CZI has been seeing and unrevert this PR before TileDB-SOMA 1.14.0. (I.e. restore 2895, 2897, 2909 and the documented/test-cased bugs they fixed -- and also fix the consolidate/vacuum issue.)

nguyenv:
> I'm not clear on the significance of including (and now removing) the TimestampRange in soma_array.cc; hoping someone can confirm they understand why it was causing an issue, and that it's fine to (temporarily?) revert.

For arrays with enumeration columns, when we first create the array, we set the schema so that the values are empty. On the first write, we read the dictionary values from the arrow array and then call extend enumeration to write the values. We were encountering an issue where when writing an enumeration to a specified timestamp, the readback values were not maching. The solution is to perform the array schema evolution for extending the enumation at that specified timestamp, rather than at the default current timestamp. However, this seems to have created unintentional buggy side-effects. I'll put this on my list to look at today.

nguyenv (approved):

Issue #2906, opened 2024-08-16
[Bug] R CI failing


ryan-williams:
[`r-ci.yml`] has [been failing][r-ci.yml main] on `main` [since mid-June][r-ci.yml main pg3].

- The first 4 failures ([#5418], [#5419], [#5422], [#5423]; Jun 17-18) began with [#2732]
- [#5426] and [#5427] then passed (Jun 19-20)
- Every run since has failed, beginning with [#5442] on Jun 28 (42 runs at time of writing).

The last output is typically `70 | ScalarMap`, before a segfault:

```
✔ |         70 | ScalarMap                                                      

 *** caught segfault ***
address 0x1019, cause 'memory not mapped'
An irrecoverable exception occurred. R is aborting now ...
/home/runner/work/_temp/c0e3d832-8dfe-4[52](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/9713375174/job/26810031898#step:14:53)3-ac62-7a65da1f7dad.sh: line 1:  9642 Segmentation fault      (core dumped) Rscript testthat.R
Error: Process completed with exit code 139.
```

The segfault messages vary; `memory not mapped`, `bad_function_call`, `invalid permissions`, `address (nil), cause 'unknown'`. [ryan-williams/tdbs-r-ci-dbg] contains some analysis of the logs of failing runs, including frequencies of various error messages.

## Occasional successful runs
A few runs have passed on PR branches (e.g. [1][success 1], [2][success 2]). However, I tried re-running the first one, and [it failed][re-run fail], so it seems like it's just non-deterministic, and usually fails. The `dark/images` branch did [pass 3x in a row][dark/images], 8/8-11.

## Plan for debugging / fixing?
Several folks on the team have made some progress debugging it, but a fix remains elusive.

From discussions with @johnkerl, ongoing work to remove our dependency on [TileDB-R] may resolve it.

## Change workflow to 24hr cron?
It's been causing "spurious" ❌'s on PRs (and `main`) for 2 months, which can mask other failures.

Disabling it altogether feels too aggressive; it's good to have some visibility into the state of the R tests.

Setting it to run against `main` every 24hrs might be a good compromise.

[`r-ci.yml`]: https://github.com/single-cell-data/TileDB-SOMA/blob/main/.github/workflows/r-ci.yml
[r-ci.yml main]: https://github.com/single-cell-data/TileDB-SOMA/actions/workflows/r-ci.yml?query=branch%3Amain
[...]

PR #2922, opened 2024-08-21
[R] Move failing R CI workflow to daily cron

ryan-williams:
**Issue and/or context:** #2906

**Changes:** Move `r-ci.yml` to a daily cron, to avoid ❌'s on PRs / main / releases.

### DIFF
--- a/.github/workflows/r-ci.yml
+++ b/.github/workflows/r-ci.yml
@@ -1,20 +1,10 @@
 name: TileDB-SOMA R CI
 
 on:
-  pull_request:
-    paths:
-      - '**'
-      - '!**.md'
-      - '!.github/**'
-      - '.github/workflows/r-ci.yml'
-      - '!.pre-commit-config.yaml'
-      - '!apis/python/**'
-      - '!docs/**'
-      - '!LICENSE'
-  push:
-    branches:
-      - main
-      - 'release-*'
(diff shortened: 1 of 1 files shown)


eddelbuettel (changes requested):
I do not think we are improving things by not running any R checks on commits affecting it.

We can achieve 'default green' differently.  I can of course withdraw this objection "on advice of management" and see it overruled but it is not the correct engineering approach in my view.

eddelbuettel:
Yes, doing both is not a bad idea.  To reduce the fail can you please try the following

```diff
modified   .github/workflows/r-ci.yml
@@ -150,7 +150,7 @@ jobs:
 
       - name: Test
         if: ${{ matrix.covr == 'no' }}
-        run: cd apis/r/tests && Rscript testthat.R
+        run: cd apis/r/tests && CI="" Rscript testthat.R
       
       - name: Coverage
```

Happy to do it too, of course.

ryan-williams:
Per https://github.com/single-cell-data/TileDB-SOMA/issues/2906#issuecomment-2318012637, CI is mostly passing since #2926 (though still segfaulting ≈10-20% of the time).

One of the changes @eddelbuettel suggested above might still make sense, but switching the whole CI to cron does not, so I'm going to close this.

PR #2925, opened 2024-08-22
[r] Expose timestamp ranges for libtiledbsoma

mojaveazure:
Improved handling of timestamp and timestamp ranges for libtiledbsoma. When users provide a timestamp to `SOMA*Open()`, it is expected to be a scalar, defined as "read up to this point"; as libtiledbsoma expects a vector of timestamps in the form of `<t0, t1>`, add a new `tiledb_timestamp_range` active binding to automagically convert the scalar `tiledb_timestamp` to a vector, or `NULL` if no timestamp was specified

New SOMA properites:
 - `TileDBObject$tiledb_timestamp_range`: new active binding to return the timestamp as a range `<t0, t1>` for libtiledbsoma; exposed as an active binding for use outside of the R6 objects

Modified SOMA methods:
 - `TileDBObject$initialize(tiledb_timestamp = )`: now requires `tiledb_timestamp` to be either `NULL` or a scalar, finite `POSIXct` object
 - `TileDBOBject$reopen(tiledb_timestamp = )`: same as `$initialize()`


mojaveazure:
Replaced by #2929

PR #2929, opened 2024-08-26
[r] Expose timestamp ranges for libtiledbsoma

mojaveazure:
Improved handling of timestamp and timestamp ranges for libtiledbsoma. When users provide a timestamp to `SOMA*Open()`, it is expected to be a scalar, defined as "read up to this point"; as libtiledbsoma expects a vector of timestamps in the form of `<t0, t1>`, add a new `.tiledb_timestamp_range` active binding to automagically convert the scalar `tiledb_timestamp` to a vector, or `NULL` if no timestamp was specified

New SOMA properites:
 - `TileDBObject$.tiledb_timestamp_range`: new active binding to return the timestamp as a range `<t0, t1>` for libtiledbsoma; exposed as an active binding for use outside of the R6 objects

Modified SOMA methods:
 - `TileDBObject$initialize(tiledb_timestamp = )`: now requires `tiledb_timestamp` to be either `NULL` or a scalar, finite `POSIXct` object
 - `TileDBOBject$reopen(tiledb_timestamp = )`: same as `$initialize()`

Replaces #2925

Follow-up to #2926

[SC-53795](https://app.shortcut.com/tiledb-inc/story/53795)

### DIFF
--- a/apis/r/R/TileDBObject.R
+++ b/apis/r/R/TileDBObject.R
@@ -148,9 +148,15 @@ TileDBObject <- R6::R6Class(
       }
       return(private$.tiledb_timestamp)
     },
-    #' @field tiledb_timestamp_range Time range for libtiledbsoma
+    #' @field uri
+    #' The URI of the TileDB object.
+    uri = function(value) {
+      if (missing(value)) return(private$tiledb_uri$uri)
+      stop(sprintf("'%s' is a read-only field.", "uri"), call. = FALSE)
+    },
+    #' @field .tiledb_timestamp_range Time range for libtiledbsoma
     #'
-    tiledb_timestamp_range = function(value) {
+    .tiledb_timestamp_range = function(value) {
       if (!missing(value)) {
         private$.read_only_error("tiledb_timestamp_range")
       }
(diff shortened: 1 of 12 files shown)


apis/r/R/TileDBObject.R
```diff
@@ -149,6 +153,20 @@ TileDBObject <- R6::R6Class(
     uri = function(value) {
       if (missing(value)) return(private$tiledb_uri$uri)
       stop(sprintf("'%s' is a read-only field.", "uri"), call. = FALSE)
+    },
+    #' @field .tiledb_timestamp_range Time range for libtiledbsoma
+    #'
+    .tiledb_timestamp_range = function(value) {
```
eddelbuettel, line 159: I thought it was going to be 'not dots for active bindings' but 'dots for private elements' ?  

I am fine with either way.

eddelbuettel (approved):
Looks good!

apis/r/R/TileDBObject.R
mojaveazure: @johnkerl requested the leading dot to indicate that while this is a public binding, it's not designed for use by the average user (they should be using `$tiledb_timestamp` instead)

eddelbuettel: Sure but can't win'em all.  Now you have element in the .Rd file that starts with a dot  :man_shrugging:

PR #2930, opened 2024-08-26
[r] Use cached timestamp in `$create()` and `$write()`

mojaveazure:
Use cached timestamps in `$write()` and `$create()` methods instead of exposing a timestamp/timestamp range parameter. If a user wishes to write a at a new timestamp, they can reopen at a new time with `$reopen()`

Modified SOMA methods:
 - `SOMADataFrame$write()`: remove `tsrange` parameter in favor of using `self$.tiledb_timestamp_range`; writing at a new timestamp now requires re-opening at a new time with `$reopen(tiledb_timestamp = )`
 - `TileDBArray$set_metadata()`: remove `timestamps` parameter in favor of using `self$.tiledb_timestamp_range`
 - `SOMANDArrayBase$create()` and `SOMADataFrame$create()`: remove `timestamps` parameter in favor of using `self$.tiledb_timestamp_range`
 - `SOMAArrayBase$private$write_object_type_metadata()`: remove `timestamps` parameter in favor of using `self$.tiledb_timestamp_range`

Modified SOMA functions:
 - `SOMADataFrameCreate()`, `SOMADenseNDArrayCreate()`, and `SOMASparseNDArrayCreate()`: do not pass `timestamps` to `array$create()`

Follow-up to #2926

[SC-53798](https://app.shortcut.com/tiledb-inc/story/53798)

### DIFF
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -6,7 +6,7 @@ Description: Interface for working with 'TileDB'-based Stack of Matrices,
     like those commonly used for single cell data analysis. It is documented at
     <https://github.com/single-cell-data>; a formal specification available is at
     <https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md>.
-Version: 1.13.99.4
+Version: 1.13.99.5
 Authors@R: c(
     person(given = "Aaron", family = "Wolen",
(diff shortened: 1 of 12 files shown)


eddelbuettel (approved):
Looks good to me too.

PR #2935, opened 2024-08-27
Remove `_soma_geometry` field

jp-dark:
Remove `_soma_geometry` attribute from Visium spot point cloud. Add in attribute with spot diameter and metadata with spot radius.


aaronwolen (approved):
Thanks!

PR #2936, opened 2024-08-28
[c++] Pre-neatens before making unit-test helpers for variant-indexed dataframes

johnkerl:
**Issue and/or context:** This is a line-count-reducing under-diff followingon #2919 for issue #2407 / [[sc-51048]](https://app.shortcut.com/tiledb-inc/story/51048).

The intended Python and R API changes are all agreed on and finalized as described in #2407.

**Changes:**

Just some prep for my next step to unit-test #2916 and #2917, namely, we need libtiledbsoma-level support for variant-indexed dataframes (those whose dim is something other than a single int64).

**Notes for Reviewer:**

#2785 is unreviewable as-is: besides being a WIP, it's an all-in-one experimental area. Smaller PRs such as this one are being offered for manageable review.

PR #2939, opened 2024-08-28
[c++] Parameterize schema-creation unit-test helpers [WIP]

johnkerl:
**Issue and/or context:** This is a line-count-reducing under-diff following on #2936 for issue #2407 / [[sc-51048]](https://app.shortcut.com/tiledb-inc/story/51048).

**Changes:**

Some prep for having variant-indexed dataframe objects in libtiledbsoma test so that #2916 and #2917 can be unit-tested.

**Notes for Reviewer:**

This is a WIP. It needs a little more thought for how to handle non-int64 parameters to `DimInfo` (more use of templates -- ??) :| 

#2785 is unreviewable as-is: besides being a WIP, it's an all-in-one experimental area. Smaller PRs such as this one are being offered for manageable review.


johnkerl:
Leaving this in draft until I address unit-test failures

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -417,29 +417,122 @@ Dimension ArrowAdapter::_create_dim(
...
+            // but a _varying_ relative path from all the places that
+            // #include arrow_adapter.h. Hence the code duplication in
+            // logging statements. :(
+            uint64_t* b = (uint64_t*)buff;
+            LOG_DEBUG(fmt::format(
+                "_create_dim name={} b={} b1={} b2={}",
+                name,
+                b[0],
+                b[1],
+                b[2]));
```
nguyenv, line 433: https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/src/soma/soma_array.h#L302-L307

In cases where we use string formatting in the header files, we avoid using `fmt::format` and make a TODO note to replace these lines with `std::Format` once we bump to C++20. I'm not saying to remove these lines in here; it is fine as is but just an fyi.

libtiledbsoma/test/common.cc
```diff
@@ -117,71 +132,139 @@ create_arrow_schema_and_index_columns(
...
+        dim_array->n_children = 0;  // leaf node
+        index_cols_info_array->children[i] = dim_array;
+
+        // The full user-level SOMA API supports many more index types.
+        // Here we support enough types to verify we've got variant-indexed
+        // SOMADataFrame objects baseline-tested in C++, then defer exhaustive
+        // loop-over-all-datatypes handling to Python and R.
+        if (info.tiledb_datatype == TILEDB_INT64) {
+            size_t nbytes = n * sizeof(int64_t);
+            dim_array->buffers[1] = malloc(nbytes);
```
nguyenv, line 216: This comment has nothing to do with your changes and does not (and should not) be addressed in this PR. But in these test arrays we release the `ArrowArray` via `ArrowAdapter::release_array` which relies on `ManagedQuery` to `delete` the `ArrayBuffers`, so we will need a follow up PR to address this.

nguyenv (approved):
lgtm. Comments are just informational.

libtiledbsoma/test/common.cc
nguyenv: I had only meant to highlight `dim_array->buffers[1] = malloc(nbytes);`.

johnkerl: Awesome, thanks @nguyenv !! :)

PR #2948, opened 2024-09-01
[c++] Implement `upgrade_shape` for `SparseNDArray` and `DenseNDArray`

johnkerl:

**Changes:**

Implements the `upgrade_shape` method at the C++ level.

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/soma/soma_array.h
+++ b/libtiledbsoma/src/soma/soma_array.h
@@ -627,7 +627,7 @@ class SOMAArray : public SOMAObject {
      */
     void upgrade_shape(const std::vector<int64_t>& newshape);
 
-   /**
+    /**
      * @brief Increases the tiledbsoma shape up to at most the maxshape,
      * resizing the soma_joinid dimension if it is a dimension.
      *
--- a/libtiledbsoma/src/soma/soma_array.h
+++ b/libtiledbsoma/src/soma/soma_array.h
@@ -620,6 +620,14 @@ class SOMAArray : public SOMAObject {
     void resize(const std::vector<int64_t>& newshape);
 
     /**
+     * @brief Given an old-style array without current domain, sets its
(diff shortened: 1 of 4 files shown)


libtiledbsoma/src/soma/soma_array.h
```diff
@@ -619,6 +619,14 @@ class SOMAArray : public SOMAObject {
      */
     void resize(const std::vector<int64_t>& newshape);
 
+    /**
+     * @brief Given an old-style array without current domain, sets its
+     * current domain. This is applicable only to arrays having all dims
+     * of int64 type. Namely, all SparseNDArray/DenseNDArray, and
+     * default-indexed DataFrame.
+     */
+    void upgrade_shape(const std::vector<int64_t>& newshape);
```
nguyenv, line 628: `set_shape`? `upgrade_with_current_domain`? `upgrade_shape` sounds like what `resize` does where there's already a shape set, but you're updating with a new shape but maybe that's just me.

nguyenv (approved):

libtiledbsoma/src/soma/soma_array.h
johnkerl: Sorry but `upgrade_shape` was agree on by committee -- see #2407 which contains a copy/paste of our Google doc

This many months in I don't want to relitigate design ..... just wanting to move forward with what the group decision was.

johnkerl: > ... upgrade_shape sounds like what resize does where there's already a shape set, but you're updating with a new shape but maybe that's just me.

Indeed it's almost identical code -- `upgrade_shape` throws if there is already a current domain; `resize` throws if there isn't; and then everything else is the same.

I thought about leaving it just `resize` at the C++ level, but decided to keep the names separate at the C++ level to mirror the committee-decided name-separation at the user API level ...

PR #2950, opened 2024-09-03
[python/r] Implement `resize` and `tiledbsoma_upgrade_shape`

johnkerl:
**Issue and/or context:** As tracked on issue #2407 / [[sc-51048]](https://app.shortcut.com/tiledb-inc/story/51048). Here we implement `resize` and `upgrade_shape` for ND arrays. The dataframe mutators are deferred to a separate PR.

**Changes:**

Implements and tests the mutators.

**Notes for Reviewer:**

For a separate PR: needs unit-test cases, including some saved-off arrays (in a tar file, say) lacking the new-shape feature which we can apply upgrade-shape to in unit-test cases.


apis/r/R/SOMASparseNDArray.R
```diff
@@ -189,6 +189,30 @@ SOMASparseNDArray <- R6::R6Class(
...
+      # https://github.com/single-cell-data/TileDB-SOMA/issues/2955
+      newshape <- unlist(newshape, use.names = FALSE)
+      resize(self$uri, newshape, config=as.character(tiledb::config(self$tiledbsoma_ctx$context())))
+    },
+
+    #' @description Allows the array to have a resizeable shape as described in the
+    #' TileDB-SOMA 1.15 release notes.  Raises an error if the new shape exceeds maxshape in any
+    #' dimension. Raises an error if the array already has a shape.
+    #' @return No return value
+    tiledbsoma_upgrade_shape = function(newshape) {
```
mojaveazure, line 211: What is the workflow for going from non-resizeable to resizeable? I would assume that the order would be:
- open old array
- request an upgrade to resizeable array
- resize

But it seems like we can do
- open an old array
- request an upgrade to resizeable _and_ resize at the same time

Was this second workflow explictly agreed to? Also, do we want to raise an error if the array is already resizeable? Is this error classed so that we can catch it and move on à la
```r
array <- SOMASparseNDArrayOpen(uri)
tryCatch(array$tiledbsoma_upgrade_shape(), alreadyResizeableError = \(...) NULL)
array$resize(new_shape)
```

apis/r/tests/testthat/test-shape.r
```diff
@@ -31,9 +31,9 @@ test_that("SOMADataFrame shape", {
     sdf <- SOMADataFrameOpen(uri)
 
     if (.new_shape_feature_flag_is_enabled()) {
-      expect_true(sdf$has_upgraded_domain())
+      expect_true(sdf$tiledbsoma_has_upgraded_domain())
```
mojaveazure, line 34: Hate to nitpick, but can we be consistent with file endings? This is the one file in the package using `.r` instead of `.R`

mojaveazure (changes requested):

apis/r/R/Init.R
```diff
@@ -64,6 +64,27 @@
...
+    .pkgenv[["use_current_domain_transitional_internal_only"]] <- FALSE
+}
+
+# This is temporary only. Please see:
+# * https://github.com/single-cell-data/TileDB-SOMA/issues/2407
+# * https://github.com/single-cell-data/TileDB-SOMA/pull/2950
+.new_shape_feature_flag_is_enabled <- function() {
+    .pkgenv[["use_current_domain_transitional_internal_only"]]
+}
+
```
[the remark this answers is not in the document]
johnkerl: @mojaveazure because I lack your expertise! I can look into using `options`, no problem.

apis/r/tests/testthat/test-shape.r
johnkerl: > Hate to nitpick

Not at all, I appreciate your eyes! :) 

> Hate to nitpick, but can we be consistent with file endings? This is the one file in the package using .r instead of .R

I simply didn't notice -- thanks for pointing it out!

apis/r/R/SOMASparseNDArray.R
johnkerl: > Was this second workflow explictly agreed to? 

@mojaveazure if you mean what I think you mean, it's not only not agreed to, it's not implemented

* `tiledbsoma_upgrade_shape` throws if the array has a shape
* `resize` throws if it does not

mojaveazure: My question was do we want `tiledbsoma_upgrade_shape()` to make arrays resizeable _and_ perform a resize at the same time, or do we want these to be separate steps?

apis/r/R/Init.R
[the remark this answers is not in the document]
mojaveazure: > This needn't be pretty or optimized for R UX -- it simply needs to be operable from CI for a few more PRs now

Then ignore this part of the review; I was more curious than critical here and if this isn't permanent then it's good enough (TM)

apis/r/R/SOMASparseNDArray.R
johnkerl: The spec discussion was we wanted the latter -- separate steps. And I don't want to re-litigate spec discussions.

And the former is not possible as I've got it coded.

So I'm with the spec. I'll make this clearer in the comments.

apis/r/R/Init.R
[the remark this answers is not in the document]
johnkerl: Thanks @mojaveazure !

apis/r/R/SOMASparseNDArray.R
johnkerl: @mojaveazure ah, I think our terminology might be not quite aligned.

* The core domain (soma max shape) is immutable. If it's set to 100 x 200 then it can never be set higher.
  * Every TileDB array already has a core domain -- you can't have an array without one
* A TileDB array can have a core current domain (soma shape) -- old arrays won't have one; new arrays will
* If the array has no core current domain (soma shape), a current domain (soma shape)can be set on it. This is what `tiledbsoma_upgrade_shape` does.
  * The core current domain (soma shape) being set must be `<=` the core domain (soma maxshape) in every slot.
  * So if the core domain is 100x200 then the current domain can be set to 20 x 30, or 100 x 200, but it can't be set to 1000 x 2000.
* If the array has has a core current domain (soma shape), it can be changed:
  * The new core current domain (new soma shape) must be >= the old current domain (old soma shape)
  * The new core current domain (new soma shape) must be <= the core domain (soma maxshape)

So to "make an array resizable":

* If the existing maxshape is 100 x 200 then that is immutable and there is no facility to change that, other than re-creating the array
* If the array doesn't have soma shape, one can be added
* If the array does have one, it can be extended, up to the maxshape

johnkerl: Phrased even more compactly:

* What we're now calling the `maxshape` (the core domain) was and is immutable
* The new `shape` is a new thing which can be _added_, which, if present, can be smaller than the `maxshape`, and this _new_ thing can be resized up to the `maxshape`

apis/r/R/SOMASparseNDArray.R
```diff
@@ -189,6 +189,30 @@ SOMASparseNDArray <- R6::R6Class(
...
+    #' @description Increases the shape of the array as specfied. Raises an error
+    #' if the new shape is less than the current shape in any dimension. Raises
+    #' an error if the new shape exceeds maxshape in any dimension. Raises an
+    #' error if the array doesn't already have a shape: in that case please call
+    #' tiledbsoma_upgrade_shape.
+    #' @return No return value
+    resize = function(newshape) {
+      # TODO: move this to SOMANDArrayBase.R once core offers current-domain support for dense arrays.
+      # https://github.com/single-cell-data/TileDB-SOMA/issues/2955
+      newshape <- unlist(newshape, use.names = FALSE)
```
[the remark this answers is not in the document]
johnkerl: @mojaveazure -- mod pushed! :)

johnkerl:
@mojaveazure ready for round two!

apis/r/R/Init.R
```diff
@@ -50,6 +50,27 @@
...
+    .pkgenv[["use_current_domain_transitional_internal_only"]] <- TRUE
+}
+
+# This is temporary only. Please see:
+# * https://github.com/single-cell-data/TileDB-SOMA/issues/2407
+# * https://github.com/single-cell-data/TileDB-SOMA/pull/2950
+.new_shape_feature_flag_disable <- function() {
+    .pkgenv[["use_current_domain_transitional_internal_only"]] <- FALSE
+}
+
```
mojaveazure, line 66: Same feedback as https://github.com/single-cell-data/TileDB-SOMA/pull/2962#discussion_r1755112169

apis/r/R/SOMASparseNDArray.R
mojaveazure: Okay, IIUC `$tiledbsoma_upgrade_shape()` adds a new SOMA shape to the array, and `$resize()` expands the SOMA shape to `soma_shape` <= `new_shape` <= `array_domain`? In that case, the name `new_shape` for `$tiledbsoma_upgrade_shape()` is confusing as it implies (to me and other biologists) that this method is doing "add SOMA shape" _and_ resize at the same time as its the same parameter name as `$resize()`. I would call it something like `shape` or `soma_shape` to differentiate it from `$resize()`

apis/r/R/SOMASparseNDArray.R
```diff
@@ -189,6 +189,40 @@ SOMASparseNDArray <- R6::R6Class(
...
       nnz(self$uri, config=as.character(tiledb::config(self$tiledbsoma_ctx$context())))
+    },
+
+    #' @description Increases the shape of the array as specfied. Raises an error
+    #' if the new shape is less than the current shape in any dimension. Raises
+    #' an error if the new shape exceeds maxshape in any dimension. Raises an
+    #' error if the array doesn't already have a shape: in that case please call
+    #' tiledbsoma_upgrade_shape.
+    #' @return No return value
+    resize = function(new_shape) {
```
mojaveazure, line 200: We need documentation for `new_shape`
```suggestion
    #' @description Increases the shape of the array as specfied. Raises an error
    #' if the new shape is less than the current shape in any dimension. Raises
    #' an error if the new shape exceeds maxshape in any dimension. Raises an
    #' error if the array doesn't already have a shape: in that case please call
    #' tiledbsoma_upgrade_shape.
    #' @param new_shape New shape to expand array to
    #' @return No return value
    resize = function(new_shape) {
```
With the new docs, `devtools::document()` needs to be run to update the Rd files in `man/`

mojaveazure (changes requested):

johnkerl:
@mojaveazure ready for the next round! :)

mojaveazure (approved):
LGTM, thank you! Please update the `NEWS.md`, bump the develop version in `DESCRIPTION`, and 🚢

apis/python/src/tiledbsoma/soma_sparse_ndarray.cc
```diff
@@ -115,6 +115,28 @@ void load_soma_sparse_ndarray(py::module& m) {
...
+        .def(
+            "tiledbsoma_upgrade_shape",
+            [](SOMAArray& array, const std::vector<int64_t>& newshape) {
+                try {
+                    array.upgrade_shape(newshape);
+                } catch (const std::exception& e) {
+                    throw TileDBSOMAError(e.what());
+                }
+            },
+            "newshape"_a);
```
nguyenv, line 140: I think we should do the rethrow in libtiledbsoma rather than pybind11. It'll simplify the code in here to be a one-liner again and the C++ API and R API would also consistently see `TileDBSOMAError` when erroring out from these methods.

nguyenv (approved):
lgtm but minor suggestion

apis/python/src/tiledbsoma/soma_sparse_ndarray.cc
johnkerl: I wish that worked reliably! We found on #2980 and #2963 that, unfortunately, it does not. This has to be a catch-and-rethrow right here at the pybind layer, unfortunately. :(

Issue #1630, opened 2023-08-23
[r] Port append-mode logic


johnkerl:
Port #1278 from Python to R.

johnkerl:
This will also need to include #2455

jp-dark:
Closing issue as stale.

PR #2533, opened 2024-05-08
[python] Raise `NotImplementedError` for `DataFrame.shape`

nguyenv:
**Issue and/or context:**

As per the SOMA spec, `DataFrame`s do not have `shape` implemented.

This parallels the R API's https://github.com/single-cell-data/TileDB-SOMA/pull/2531.

**Changes:**

Raise `NotImplementedError` for `DataFrame.shape`.

PR #2953, opened 2024-09-04
[python/r] Expose shape-related accessors to Python/R bindings

johnkerl:

**Changes:**

* These accessors are already implemented and unit-tested in C++
* On this PR, I expose only the language bindings as-is with minimal unit-testing -- more unit-testing to follow (including feature-flag logic) as tracked on #2407 
* The `_maybe_sjid_shape` and `_maybe_sjid_maxshape` accessors will be crucial for experiment-level resize logic as on append mode, which does not yet exist in TileDB-SOMA-R (see #1630)

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -317,6 +317,40 @@ class DataFrame(SOMAArray, somacore.DataFrame):
         # if is it in read open mode, then it is a DataFrameWrapper
         return cast(DataFrameWrapper, self._handle).count
 
+    @property
+    def _maybe_soma_joinid_shape(self) -> Optional[int]:
+        """An internal helper method that returns the shape
+        value along the ``soma_joinid`` index column, if the ``DataFrame
+        has one, else ``None``.
+
+
+        Lifecycle:
+            Experimental.
+        """
+        return self._handle.maybe_soma_joinid_shape
+
+    @property
+    def _maybe_soma_joinid_maxshape(self) -> Optional[int]:
+        """An internal helper method that returns the maxshape
+        value along the ``soma_joinid`` index column, if the ``DataFrame
+        has one, else ``None``.
+
+        Lifecycle:
+            Experimental.
+        """
+        return self._handle.maybe_soma_joinid_maxshape
+
+    @property
+    def has_upgraded_domain(self) -> bool:
+        """Returns true if the array has the upgraded resizeable domain feature
+        from TileDB-SOMA 1.14: the array was created with this support, or it has
+        had ``.upgrade_domain`` applied to it.
+
+        Lifecycle:
+            Maturing.
+        """
+        return self._handle.has_upgraded_domain
+
     def __len__(self) -> int:
         """Returns the number of rows in the dataframe. Same as ``df.count``."""
         return self.count
--- a/apis/python/src/tiledbsoma/_tdb_handles.py
+++ b/apis/python/src/tiledbsoma/_tdb_handles.py
@@ -406,7 +406,33 @@ class SOMAArrayWrapper(Wrapper[_ArrType]):
 
     @property
     def shape(self) -> Tuple[int, ...]:
-        return tuple(self._handle.shape)
+        """Not implemented for DataFrame."""
+        return cast(Tuple[int, ...], tuple(self._handle.shape))
+
+    @property
+    def maxshape(self) -> Tuple[int, ...]:
+        """Not implemented for DataFrame."""
+        return cast(Tuple[int, ...], tuple(self._handle.maxshape))
+
+    @property
+    def maybe_soma_joinid_shape(self) -> Optional[int]:
+        """Only implemented for DataFrame."""
+        raise NotImplementedError
+
+    @property
+    def maybe_soma_joinid_maxshape(self) -> Optional[int]:
+        """Only implemented for DataFrame."""
+        raise NotImplementedError
+
+    @property
+    def has_upgraded_shape(self) -> bool:
+        """Not implemented for DataFrame."""
+        raise NotImplementedError
+
+    @property
+    def has_upgraded_domain(self) -> bool:
+        """Only implemented for DataFrame."""
+        raise NotImplementedError
 
 
 class DataFrameWrapper(SOMAArrayWrapper[clib.SOMADataFrame]):
@@ -422,15 +448,54 @@ class DataFrameWrapper(SOMAArrayWrapper[clib.SOMADataFrame]):
         self._handle.write(values)
 
     @property
-    def shape(self) -> Tuple[int, ...]:
-        # Shape is not implemented for DataFrames
-        raise NotImplementedError
+    def maybe_soma_joinid_shape(self) -> Optional[int]:
+        """Return the shape slot for the soma_joinid dim, if the array has one.
+        This is an important test-point and dev-internal access-point,
+        in particular, for the tiledbsoma-io experiment-level resizer.
+
+        Lifecycle:
+            Maturing.
+        """
+        return cast(Optional[int], self._handle.maybe_soma_joinid_shape)
+
+    @property
+    def maybe_soma_joinid_maxshape(self) -> Optional[int]:
+        """Return the maxshape slot for the soma_joinid dim, if the array has one.
+        This is an important test-point and dev-internal access-point,
+        in particular, for the tiledbsoma-io experiment-level resizer.
+
+        Lifecycle:
+            Maturing.
+        """
+        return cast(Optional[int], self._handle.maybe_soma_joinid_maxshape)
+
+    @property
+    def has_upgraded_domain(self) -> bool:
+        """Returns true if the array has the upgraded resizeable domain feature
+        from TileDB-SOMA 1.14: the array was created with this support, or it has
+        had ``.upgrade_domain`` applied to it.
+
+        Lifecycle:
+            Maturing.
+        """
+        return cast(bool, self._handle.has_upgraded_domain)
 
 
 class DenseNDArrayWrapper(SOMAArrayWrapper[clib.SOMADenseNDArray]):
     """Wrapper around a Pybind11 DenseNDArrayWrapper handle."""
 
     _ARRAY_WRAPPED_TYPE = clib.SOMADenseNDArray
+
+    @property
+    def has_upgraded_shape(self) -> bool:
+        """Returns true if the array has the upgraded resizeable shape feature
+        from TileDB-SOMA 1.14: the array was created with this support, or it has
+        had ``.upgrade_shape`` applied to it.
+
+        Lifecycle:
+            Maturing.
+        """
+        return cast(bool, self._handle.has_upgraded_shape)
 
 
 class SparseNDArrayWrapper(SOMAArrayWrapper[clib.SOMASparseNDArray]):
@@ -441,6 +506,17 @@ class SparseNDArrayWrapper(SOMAArrayWrapper[clib.SOMASparseNDArray]):
     @property
     def nnz(self) -> int:
         return int(self._handle.nnz())
+
+    @property
+    def has_upgraded_shape(self) -> bool:
+        """Returns true if the array has the upgraded resizeable shape feature
+        from TileDB-SOMA 1.14: the array was created with this support, or it has
+        had ``.upgrade_shape`` applied to it.
+
+        Lifecycle:
+            Maturing.
+        """
+        return cast(bool, self._handle.has_upgraded_shape)
 
 
 class _DictMod(enum.Enum):
--- a/apis/python/tests/test_shape.py
+++ b/apis/python/tests/test_shape.py
@@ -1,0 +1,211 @@
+from __future__ import annotations
+
+import pyarrow as pa
+import pytest
+
+import tiledbsoma
+
+from tests._util import maybe_raises
+
+
+@pytest.mark.parametrize(
+    "element_dtype",
+    [
+        pa.float64(),
+        pa.float32(),
+        pa.int64(),
+        pa.uint16(),
+    ],
+)
+@pytest.mark.parametrize(
+    "shape_exc",
+    [
+        # Note: non-None exceptions are coming on https://github.com/single-cell-data/TileDB-SOMA/issues/2407
+        [(100,), None],
+        [(100, 200), None],
+        [(100, 200, 300), None],
+    ],
+)
+def test_sparse_nd_array_basics(
+    tmp_path,
+    element_dtype,
+    shape_exc,
+):
+    uri = tmp_path.as_posix()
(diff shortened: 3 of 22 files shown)


johnkerl:
Leaving this in draft status until I add more unit-test cases extracted from #2785

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -317,6 +317,29 @@ def count(self) -> int:
         # if is it in read open mode, then it is a DataFrameWrapper
         return cast(DataFrameWrapper, self._handle).count
 
+    @property
+    def _maybe_sjid_shape(self) -> Optional[int]:
+        """An internal helper method that returns the shape
+        value along the ``soma_joinid`` index column, if the ``DataFrame
```
nguyenv, line 323:
```suggestion
        value along the ``soma_joinid`` (sjid) index column, if the ``DataFrame
```

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -317,6 +317,29 @@ def count(self) -> int:
...
+
+        Lifecycle:
+            Experimental.
+        """
+        return self._handle.maybe_sjid_shape
+
+    @property
+    def _maybe_sjid_maxshape(self) -> Optional[int]:
+        """An internal helper method that returns the maxshape
+        value along the ``soma_joinid`` index column, if the ``DataFrame
```
nguyenv, line 335:
```suggestion
        value along the ``soma_joinid`` (sjid) index column, if the ``DataFrame
```

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -422,9 +438,22 @@ def write(self, values: pa.RecordBatch) -> None:
...
+        return cast(Optional[int], self._handle.maybe_sjid_shape)
+
+    @property
+    def maybe_sjid_maxshape(self) -> Optional[int]:
+        """WRITE ME PLZ THX
+
+        Lifecycle:
+            Maturing.
+        """
+        return cast(Optional[int], self._handle.maybe_sjid_maxshape)
```
nguyenv, line 456: TODO to write docs here.

apis/python/tests/test_shape.py
```diff
@@ -0,0 +1,205 @@
...
+        #for e in snda.maxshape:
+        #    assert e > 2_000_000_000
+        assert snda.non_empty_domain() == coords
+
+    # Test reads out of bounds
+    with tiledbsoma.SparseNDArray.open(uri) as snda:
+        # TODO: make sure this doesn't come through as RuntimeError
+        # https://github.com/single-cell-data/TileDB-SOMA/issues/2407
+        with pytest.raises((RuntimeError, ValueError)):
+            coords = tuple([arg_shape[i] + 10 for i in range(ndim)])
```
nguyenv, line 100: nit: don't need an intermediary list for tuple generators

```suggestion
            coords = tuple(arg_shape[i] + 10 for i in range(ndim))
```

apis/python/tests/test_shape.py
```diff
@@ -0,0 +1,205 @@
...
+            snda.read(coords).tables().concat()
+
+    # Test writes out of bounds
+    with tiledbsoma.SparseNDArray.open(uri, "w") as snda:
+        with pytest.raises(tiledbsoma.SOMAError):
+            dikt = {"soma_data": [30]}
+            for i in range(ndim):
+                dikt[dim_names[i]] = [arg_shape[i] + 20]
```
nguyenv, line 108:
```suggestion
            dikt = {name: [shape + 20] for name, shape in zip(dim_names, arg_shape)}
```

apis/python/tests/test_sparse_nd_array.py
```diff
@@ -51,6 +51,14 @@ def test_sparse_nd_array_create_ok(
     assert a.uri == tmp_path.as_posix()
     assert a.ndim == len(shape)
     assert a.shape == tuple(shape)
+
+    # TODO: more testing with current-domain feature integrated
+    # https://github.com/single-cell-data/TileDB-SOMA/issues/2407
+    assert isinstance(a.maxshape, tuple)
+    assert len(a.maxshape) == len(a.shape)
+    for i in range(len(a.shape)):
+        assert a.maxshape[i] >= a.shape[i]
```
nguyenv, line 60:
```suggestion
    for maxshape, shape in zip(a.maxshape, a.shape):
        assert maxshape >= shape
```

nguyenv (approved):
Looks good for the Python side.

johnkerl:
@mojaveazure can you please take a look at the R side? Thank you!

apis/r/R/SOMADataFrame.R
```diff
@@ -327,9 +327,65 @@ SOMADataFrame <- R6::R6Class(
     #' @return None, instead a \code{\link{.NotYetImplemented}()} error is raised
     #'
     shape = function() stop(errorCondition(
-      "'SOMADataFrame$shape()' is not implemented yet",
+      "'SOMADataFrame$shape()' is not implemented",
```
mojaveazure, line 330: I had purposefully added the "yet" to adhere to R's existing [`.NotYetImplemented()`](https://stat.ethz.ch/R-manual/R-devel/library/base/html/notyet.html)
```r
> .NotYetImplemented()
Error: '.NotYetImplemented' is not implemented yet
```

apis/r/R/SOMANDArrayBase.R
```diff
@@ -84,7 +84,20 @@ SOMANDArrayBase <- R6::R6Class(
...
+    #' or it has had ``upgrade_domain`` applied to it.
+    #' (lifecycle: maturing)
+    #' @return Logical
+    has_upgraded_shape = function() {
+      has_current_domain(
+        self$uri,
+        config=as.character(tiledb::config(self$tiledbsoma_ctx$context()))
+      )
     }
+
```
mojaveazure, line 100: Is there a reason we're using a different name for NDArrays vs data frames, despite both running the same code? Can we agree on a single name for all array-like objects and push this down to a lower-level class (eg. `SOMAArrayBase` or `TileDBArray`) to avoid code duplication?

apis/r/R/SOMADataFrame.R
```diff
@@ -327,9 +327,65 @@ SOMADataFrame <- R6::R6Class(
...
+      retval <- as.integer64(maybe_soma_joinid_maxshape(
+        self$uri,
+        config=as.character(tiledb::config(self$tiledbsoma_ctx$context()))
+      ))
+      if (retval == as.integer64(-1)) {
+        NA
+      } else {
+        retval
+      }
+    },
```
mojaveazure, line 376: These are only used for tests, and don't make use of internal object structures, so do we need these attached to `SOMADataFrame`? In the test files, we can (and probably should) just use

```R
sjid <- maybe_soma_joinid_shape(sdf$uri, config = as.character(tiledb::config(sdf$tiledbsoma_ctx$context()))
```

Tacking these on as public methods to the `SOMADataFrame` class (and only this class?) seems like overkill, especially for an API we may not want exposed or are using anywhere else

Also, I'm not a fan of the `maybe_` style function names. These aren't used anywhere else in tiledbsoma-R (or amongst R packages as a whole); instead the paradigm settled on for this package for accessors that may-or-may-not work is either
 - just return `NA` (like `SOMADataFrame$shape()` prior to #2531)
 - throw an error (like bounding boxes with `soma_array$used_shape()`)

mojaveazure (changes requested):

apis/r/R/SOMADataFrame.R
johnkerl: OK, I'll put it back

(Sounds like a spec discussion ...)

I would like it if `shape` could return what `.maybe_soma_joinid_shape` does -- remembering that a `SOMADataFrame` per the spec must have a dim or attr named `soma_joinid` but (a) it doesn't have to be the only dim, and (b) it doesn't even have to be dim at all. (This is context.)

apis/r/R/SOMANDArrayBase.R
johnkerl: Yes, please see the committee-agreed namings on #2407.

To summarize:

* `SOMASparseNDArray` and `SOMADenseNDArray` must have dims of type `int64` -- only
* `SOMADataFrame` must have a dim or attr named `soma_joinid` and it must be of type `int64`
  * People can have `soma_joinid` as dim by itself (this is the default behavior and by far the most common)
  * People can have `soma_joinid` as a dim along with one or more other dims -- and those can have arbitrary type
  * People can have one or more dims of arbitrary type while `soma_joinid is an attr
* This is why throughout the existing code we have `shape` in terms of array of `int64`, and `domain` is specified for `SOMADataFrame`. Also note that in #2407 we have `shape` and `maxshape` accessors for ND arrays but `domain` and `maxdomain` for dataframes. Likewise `upgrade_shape` vs `upgrade_domain`.

Given that the committee decision was hard-won and hammered out over many meetings and much discussion, I'm adhering to what was agreed on in this implementation.

apis/r/R/SOMADataFrame.R
johnkerl: (Also, Python just raised `NotImplementedError`, not a `NotImplementedYetError` -- which is a difference between the two, albeit minor.)

apis/r/R/SOMADataFrame.R
johnkerl: They're not used _yet_ but they will be deeper in-stack. See discussion on #2407.

They will be absolutely crucial in Python for experiment-level resize, and note that we do have append mode existing in Python so this is very pressing.

For R we do not have append mode (this is backlogged as #1630) and I debated whether to include these accessors in R or not. Reason to exclude: because not used yet. Reason to include: I don't want to add to the pile of parity work needing to be done by whoever implements #1630 someday -- hoping to raise the parity a bit here.

The prefix `maybe_` indicates that the return value may be `None` (in Python) or `NA` (in R), and it's a `std::optional<int64_t>` in C++. I'm happy to leave off the prefix for the R side if that sounds more R-cultural to you.

johnkerl: > and don't make use of internal object structures

Nope -- these are wired all the way down to C++. All on this PR.

apis/r/R/SOMADataFrame.R
johnkerl: Done on this commit: https://github.com/single-cell-data/TileDB-SOMA/pull/2953/commits/840ed2dbd355a6759cc2322ea1e4e2ac37c0c5e6

apis/r/R/SOMANDArrayBase.R
johnkerl: > * `SOMADataFrame` must have a dim or attr named `soma_joinid` and it must be of type `int64`
>   * People can have `soma_joinid` as dim by itself (this is the default behavior and by far the most common)
>   * People can have `soma_joinid` as a dim along with one or more other dims -- and those can have arbitrary type
>   * People can have one or more dims of arbitrary type while `soma_joinid` is an attr

BTW this is what `index_column_names` is all about

mojaveazure: > Yes, please see the committee-agreed namings

Good enough for me. I was not aware that these namings had been decided upon, nor was I apart of any committee making these decisions, but if that's already been hammered out then 👍 

> Re "despite running the same code" -- it's actually not. Given C++'s strong typing and use of classes [...]

Yeah, but from the R API side, that's not important. All I see is these two methods are
1. calling the same function (`has_current_domain()` with the same arguments (`self$uri` and `config = ...`)
2. have different names

What the function is doing (passing data to Rcpp, which then passes on to libtiledbsoma, which then dispatches based on the C++ class) is not important; all I see is code duplication (again, I was not aware of any committee to decide on these names). However, if a committee decided that these differing names are important, then 👍

apis/r/R/SOMADataFrame.R
johnkerl: > sjid <- maybe_soma_joinid_shape(sdf$uri, config = as.character(tiledb::config(sdf$tiledbsoma_ctx$context()))

I'm fine with this as a compromise between these two:

> For R we do not have append mode (this is backlogged as https://github.com/single-cell-data/TileDB-SOMA/issues/1630) and I debated whether to include these accessors in R or not. Reason to exclude: because not used yet. Reason to include: I don't want to add to the pile of parity work needing to be done by whoever implements https://github.com/single-cell-data/TileDB-SOMA/issues/1630 someday -- hoping to raise the parity a bit here.

mojaveazure: > They're not used yet but they will be deeper in-stack. See discussion on #2407.

I don't see where in #2407 these names are specified (maybe I'm missing something) or their behavior defined. I also think figuring out _how_ resize and append-mode will work in R (what their user-facing APIs are) would be important into deciding where these two accessors belong (public, private, public-but-dotted, external)

> The prefix maybe_ indicates that the return value may be [...]

I get that, but the R way would be no prefix and describing the "alternate" behavior in the docs (eg. returning `NA`/`NULL`/zero-length vector or raising an error)

johnkerl: > I don't see where in https://github.com/single-cell-data/TileDB-SOMA/issues/2407 these names are specified (maybe I'm missing something) or their behavior defined

Sorry, I meant the PRs listed there.

apis/r/R/SOMADataFrame.R
mojaveazure: > I personally would love it if `shape` could return what `.maybe_soma_joinid_shape` [...]

That would require re-hashing the discussions that lead to #2533 and #2531; I'm happy to have those conversations and merge these behaviors, but I agree that this is a "task for later"

> (Also, Python just raises `NotImplementedError`, not a `NotImplementedYetError` [...]

I'm aware, this is R's equivalent. I get that these are different, but whenever possible I'd like to adhere to existing R standards

> Done on this commit: 840ed2d

TY!

apis/r/R/SOMADataFrame.R
mojaveazure: I'm still going to ask that the `maybe_` prefix be removed; either just dotted-methods or removed entirely with implementation to be deciced later (merge with `$shape()`/`$maxshape()`, move to private, or expose publicly-but-dotted)

Also, I noticed that these are methods in R, but properties in Python. Do we want this discrepancy between the two APIs or should we settle on both having these be methods or properties?

johnkerl: @mojaveazure https://github.com/single-cell-data/TileDB-SOMA/pull/2953/commits/5312d49cfe08282110645c5faafcf6051951855e

mojaveazure (approved):

johnkerl:
Thanks @nguyenv and @mojaveazure !

PR #2956, opened 2024-09-05
[c++] Also check for leading slash when checking for 'relative' URLs

eddelbuettel:
**Issue and/or context:**

As discussed over DMs, the check for 'absolute vs relative' URLs was not checking for leading slashes as on local filesystems.

**Changes:**

The check has been expanded.  

**Notes for Reviewer:**

[SC 54529](https://app.shortcut.com/tiledb-inc/story/54529/c-generalize-relative-vs-absolute-check-for-group-url)

### DIFF
--- a/libtiledbsoma/src/soma/soma_group.cc
+++ b/libtiledbsoma/src/soma/soma_group.cc
@@ -5,7 +5,7 @@
  *
  * The MIT License
  *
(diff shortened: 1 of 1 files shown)


nguyenv (approved):

nguyenv:
Thank you!

PR #2962, opened 2024-09-06
[python/r] Array-creation mods for new shape

johnkerl:

**Changes:**

* Array-creation mods including the crucial arrow-table link are already bullet-proof tested in C++ on #2911 
* More unit-test cases for Python and R are upcoming on #2407
* The status-quo code and the opt-in to-be-released-later code are both tested in CI
* There are further upcoming PRs as tracked on #2407

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -14,6 +14,8 @@ import somacore
 from somacore import options
 from typing_extensions import Self
 
+from tiledbsoma import _new_shape_feature_flag_enabled
+
 from . import _arrow_types, _util
 from . import pytiledbsoma as clib
 from ._constants import SOMA_JOINID
@@ -215,10 +217,33 @@ class DataFrame(SOMAArray, somacore.DataFrame):
         """
         context = _validate_soma_tiledb_context(context)
         schema = _canonicalize_schema(schema, index_column_names)
-        if domain is None:
-            domain = tuple(None for _ in index_column_names)
+
+        # SOMA-to-core mappings:
+        #
+        # Before the current-domain feature was enabled (possible after core 2.25):
+        #
+        # * SOMA domain <-> core domain, AKA "max domain" which is a name we'll use for clarity
+        # * core current domain did not exist
+        #
+        # After the current-domain feature was enabled:
+        #
+        # * SOMA max_domain <-> core domain
+        # * SOMA domain <-> core current domain
+        #
+        # As far as the user is concerned, the SOMA-level domain is the only
+        # thing they see and care about. Before 2.25 support, it was immutable
+        # (since it was implemented by core domain). After 2.25 support, it is
+        # mutable/up-resizeable (since it is implemented by core current domain).
+
+        # At this point shift from API terminology "domain" to specifying a soma_ or core_
+        # prefix for these variables. This is crucial to avoid developer confusion.
+        soma_domain = domain
+        domain = None
+
+        if soma_domain is None:
+            soma_domain = tuple(None for _ in index_column_names)
         else:
-            ndom = len(domain)
+            ndom = len(soma_domain)
             nidx = len(index_column_names)
             if ndom != nidx:
                 raise ValueError(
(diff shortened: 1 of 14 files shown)


johnkerl:
Back in draft while I fix CI

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -860,3 +922,20 @@ def _find_extent_for_domain(
...
     return extent
+
+
+def _revise_domain_for_extent(
+    domain: Tuple[Any, Any], extent: Any, saturated_range: bool
+) -> Tuple[Any, Any]:
+    """
+    We need to do this to avoid this error at array-creation time:
+
+    Error: [TileDB::Dimension] Error: Tile extent check failed; domain max
```
nguyenv, line 933: This is just the error description right? The exception type should be `SOMAError` and not `TileDBError`.

apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -122,18 +124,72 @@ def create(
...
+            # [1] core max domain hi
+            # [2] core extent parameter
+            # If present, these next two signal to use the current-domain feature:
+            # [3] core current domain lo
+            # [4] core current domain hi
+
+            if _new_shape_feature_flag_enabled():
+                dim_capacity, dim_extent = cls._dim_capacity_and_extent(
+                    dim_name,
+                    None,  # XXX COMMENT
```
nguyenv, line 169: Missing doc

johnkerl:
@nguyenv ready for re-review

nguyenv (approved):
Looks good on the Python side.

johnkerl:
@mojaveazure ping 🙏

apis/r/R/SOMANDArrayBase.R
```diff
@@ -135,6 +135,21 @@ SOMANDArrayBase <- R6::R6Class(
...
+
+      ## ensure newshape is a vector of int64
+      stopifnot("'newshape' must be a vector of integers" = is.vector(newshape));
+                # XXX "'newshape' must be a vector of integers" = all(vapply_lgl(newshape, is.integer)),
+
+      ## convert integer to integer64 to match dimension type
+      newshape <- lapply(newshape, function(x) if (inherits(x, "integer")) bit64::as.integer64(x) else x)
+
+      as.vector(newshape)
+    },
```
mojaveazure, line 151: This function can be strengthened and simplified by:
 - having stronger checking for integerish values
 - making use of vectorized `bit64::as.integer64()`
 - making use of method dispatch of `bit64::as.integer64()` to handle cases where `new_shape` is integer, numeric, or integer64

Also uses `new_shape` to be consistent with style
```suggestion
    .convert_shape_argument = function(new_shape) {
      # ensure new_shape is an integerish vector
      stopifnot(
        "'new_shape' must be an integerish vector" = rlang::is_integerish(new_shape, n = self$ndim(), finite = TRUE) ||
          (bit64::is.integer64(new_shape) && length(new_shape) == self$ndim() && is.finite(new_shape))
      )
      
      # convert integer to integer64 to match dimension type
      return(bit64::as.integer64(new_shape))
    },
```

However, I don't see any place where this is used. Will this be used? If so, why not just call `bit64::as.integer64()` where needed? I don't see a case where this is actually needed

apis/r/R/Init.R
```diff
@@ -17,6 +17,39 @@
...
+    # It will be removed once 2407 is complete.
+    if (Sys.getenv("SOMA_R_NEW_SHAPE") != "") {
+      .pkgenv[["use_current_domain_transitional_internal_only"]] <- TRUE
+      cdmsg <- " SOMA_R_NEW_SHAPE enabled"
+    } else {
+      .pkgenv[["use_current_domain_transitional_internal_only"]] <- FALSE
+      cdmsg <- ""
+    }
+     msg <- sprintf("TileDB Core version %s used by TileDB-R package, but TileDB-SOMA uses %s [%s]",
+                    sQuote(rpkg_lib_version), sQuote(soma_lib_version), sQuote(cdmsg))
```
mojaveazure, line 31: What is `msg` here for? It seems like it's being created and never used

apis/r/R/Init.R
```diff
@@ -17,6 +17,39 @@
...
+# * https://github.com/single-cell-data/TileDB-SOMA/pull/2950
+.new_shape_feature_flag_disable <- function() {
+    .pkgenv[["use_current_domain_transitional_internal_only"]] <- FALSE
+}
+
+# This is temporary only. Please see:
+# * https://github.com/single-cell-data/TileDB-SOMA/issues/2407
+# * https://github.com/single-cell-data/TileDB-SOMA/pull/2950
+.new_shape_feature_flag_is_enabled <- function() {
+    .pkgenv[["use_current_domain_transitional_internal_only"]]
```
mojaveazure, line 52: These functions are being defined, but never used (`.new_shape_feature_flag_enabled()` is used in tests, but that's it). Instead, `.pkenv` is being modified (in `.onLoad()`) and accessed (in `get_domain_and_extent_dataframe()` and `get_domain_and_extent_array()`) directly

mojaveazure (changes requested):

apis/r/R/SOMANDArrayBase.R
johnkerl: Thanks @mojaveazure -- code push coming up soon! Multi-tasking this morning. :)

apis/r/R/Init.R
johnkerl: It was used in the startup message; I'll double-check I still am

apis/r/R/Init.R
johnkerl: I thought I'd use them for unit-test purposes

It looks like the env-var-driven approach will suffice

I'll take these back out

johnkerl: I got rid of the first two; `new_shape_feature_flag_is_enabled` is used in unit tests

apis/r/R/Init.R
johnkerl: Fixed (on next push)

johnkerl:
@mojaveazure I think I've got all your points addressed! :)

mojaveazure (approved):

johnkerl:
Thanks @nguyenv and @mojaveazure !

PR #2965, opened 2024-09-06
[r] Fix bug in blockwise iterator

mojaveazure:
When using the blockwise iterator, I get the following error on the last iteration's read:
```
Error: Subarray: Cannot add range to dimension 'soma_dim_0'; Range [2638, 2638] is out of domain bounds [0, 2637]
```
It turns out we were not handling 0-based coordinates correctly. When creating a new blockwise read, we would start at 0L, but then go up to `array$shape()[dimension]`, which is a 1-indexed value. This needed to be corrected to `array$shape()[dimension] - 1L`

However, this messed up calculating a dimension for the returning matrix; to get around this, if `coords$start == 0L`, we increase the length of the coords by 1 (eg. if the coordinates run from `[0, 499]`, then the length should be 500, not 499)

Modified SOMA methods:
 - `SOMASparseNDArray$new()`: decrease the end of the coordinate strider by 1
 - `CoordsStrider$length()`: if `coords$start` is 0, increase the length by 1

[SC-54350](https://app.shortcut.com/tiledb-inc/story/54350/)

### DIFF
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -6,7 +6,7 @@ Description: Interface for working with 'TileDB'-based Stack of Matrices,
     like those commonly used for single cell data analysis. It is documented at
     <https://github.com/single-cell-data>; a formal specification available is at
     <https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md>.
-Version: 1.13.99.5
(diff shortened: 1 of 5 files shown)


eddelbuettel (approved):
:+1:

PR #2971, opened 2024-09-09
[r/ci] Fix bug when installing tiledb-r

mojaveazure:
The `install-tiledb-r.R` script isn't correctly handling the case where all repos contain a valid version of tiledb-r. Currently, we're using `-idx[!valid]` which returns `integer(0)` when `valid` is all `TRUE`; to fix, adjust this dropping to run only when `!all(valid)`

Also contains the following adjustments:
 - switch upstream dependency solving to an `apply()` in case of multiple valid versions with differing upstream dependencies
 - fix the `bspm::disable()` check to actually use bspm where available

[SC-54635](https://app.shortcut.com/tiledb-inc/story/54635/)

### DIFF
--- a/apis/r/tools/install-tiledb-r.R
+++ b/apis/r/tools/install-tiledb-r.R
@@ -31,10 +31,20 @@ valid <- vapply(
 if (!any(valid)) {
   stop("No valid versions of tiledb-r found")
 }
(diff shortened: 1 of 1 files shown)


johnkerl (approved):
🔥

PR #2972, opened 2024-09-09
[c++] Support option to set log level from environment

johnkerl:
I find this very useful, particularly on #2407 / [[sc-51048]](https://app.shortcut.com/tiledb-inc/story/51048). If useful to other folks, great; if not, hopefully not too disruptive.

### DIFF
--- a/libtiledbsoma/src/utils/logger.cc
+++ b/libtiledbsoma/src/utils/logger.cc
@@ -33,6 +33,7 @@
 
 #include "logger.h"
 
+#include <spdlog/cfg/env.h>
 #include <spdlog/fmt/fmt.h>
 #include <spdlog/fmt/ostr.h>
 #include <spdlog/sinks/basic_file_sink.h>
@@ -71,12 +72,11 @@ Logger::Logger() {
             spdlog::level::critical, console_sink->red_bold);
 #endif
     }
-    char* env_level = std::getenv("LIBTILEDBSOMA_LOGGING_LEVEL");
-    if (env_level != nullptr && strcmp(env_level, "") != 0) {
-        set_level(env_level);
-    } else {
-        set_level("INFO");
-    }
+    set_level("INFO");
+    // Examples:
+    // SPDLOG_LEVEL=trace name-of-program
+    // SPDLOG_LEVEL=tiledbsoma=trace name-of-program
+    spdlog::cfg::load_env_levels();
 }
 
 Logger::~Logger() {
--- a/libtiledbsoma/src/utils/logger.cc
+++ b/libtiledbsoma/src/utils/logger.cc
@@ -71,7 +71,12 @@ Logger::Logger() {
             spdlog::level::critical, console_sink->red_bold);
 #endif
     }
-    set_level("INFO");
+    char* env_level = std::getenv("LIBTILEDBSOMA_LOGGING_LEVEL");
+    if (env_level != nullptr && strcmp(env_level, "") != 0) {
+        set_level(env_level);
+    } else {
+        set_level("INFO");
+    }
 }
 
 Logger::~Logger() {


libtiledbsoma/src/utils/logger.cc
```diff
@@ -71,7 +71,12 @@ Logger::Logger() {
             spdlog::level::critical, console_sink->red_bold);
 #endif
     }
-    set_level("INFO");
+    char* env_level = std::getenv("LIBTILEDBSOMA_LOGGING_LEVEL");
```
eddelbuettel, line 74: There is a built-in activator for spdlog we could use too.  I have been meaning to expose it in the R package as well.  See https://github.com/eddelbuettel/spdlog#load-log-levels-from-env-variable-or-from-argv

I also find it convenient to turn on R _and_ libtiledbsoma logging:  
https://github.com/single-cell-data/TileDB-SOMA/blob/7cacee1ec3824040879e55fc4bdd7d5df4a56e7b/apis/r/src/rinterface.cpp#L189-L192

johnkerl: > I also find it convenient to turn on R and libtiledbsoma logging:

Thanks @eddelbuettel . There's also `tiledbsoma.pytiledbsoma.config_logging("debug")` which is directly callable from Python. And `spdl::set_log_level("DEBUG")` which is directly callable from R. Knowing both of those things, I also do want the option to enable logging _without_ editing source code -- in whichever language.

Thanks for the tip at https://github.com/eddelbuettel/spdlog#load-log-levels-from-env-variable-or-from-argv re `spdlog::cfg::load_env_levels`! TIL :). I do however want the option to set application-specific levels from the environment.

eddelbuettel: Also keep in mind that there are different instances of spdlog involved which is why the C++-side-of-tiledbsoma-R has two enablers, one for C++ (in libtiledbsoma) and one for R + C++ in the R package.  

> I do however want the option to set application-specific levels from the environment.

That is exactly what it does. (Don't get distracted that the reference accidentally goes to my fork.  It is a spdlog upstream feature.)

I used that very feature in the example code I posted on slack when I proposed header-only use of spdlog and fmt as potential fix to occassional build woes.  Anyway ...

nguyenv:
Is this something we should add to [CMakeLists.txt](https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/CMakeLists.txt) too where we default it to OFF? Maybe for debug builds it could be INFO.

eddelbuettel:
It's is strictly run-time. `spdlog` takes care of it.  Using the (zero-other-dependency) example program I used before (included below). We should be able to tuck this away in the non-public part of the logger.

```sh
$ g++ -std=c++17 spdlog_ex_headeronly.cpp -o spdlog_ex_headeronly
$ ./spdlog_ex_headeronly    # no setting
[14:26:53.559466] [my_demo] [I] [thread 2308290] Welcome to spdlog!
[14:26:53.559523] [my_demo] [E] [thread 2308290] Some error message with arg: 1
[14:26:53.559539] [my_demo] [I] [thread 2308290] Elapsed time: 6.8161e-05
[14:26:53.559548] [my_demo] [W] [thread 2308290] Easy padding in numbers like 00000012
[14:26:53.559584] [my_demo] [C] [thread 2308290] Support for int: 42;  hex: 2a;  oct: 52; bin: 101010
[14:26:53.559600] [my_demo] [I] [thread 2308290] Support for floats 1.23
[14:26:53.559612] [my_demo] [I] [thread 2308290] Positional args are supported too..
[14:26:53.559622] [my_demo] [I] [thread 2308290] left aligned                  
[14:26:53.559631] [my_demo] [I] [thread 2308290] Elapsed time: 0.000164105
$
$ SPDLOG_LEVEL=critical ./spdlog_ex_headeronly    # critical only
[14:27:04.225658] [my_demo] [C] [thread 2308818] Support for int: 42;  hex: 2a;  oct: 52; bin: 101010
$ 
$ SPDLOG_LEVEL=error ./spdlog_ex_headeronly    # error or higher
[14:27:15.571203] [my_demo] [E] [thread 2309374] Some error message with arg: 1
[14:27:15.571263] [my_demo] [C] [thread 2309374] Support for int: 42;  hex: 2a;  oct: 52; bin: 101010
$ 
```

It is activated by calling one line as demonstrated eg [here](https://github.com/gabime/spdlog#load-log-levels-from-the-env-variable-or-argv) and shown in the example I compiled and use here.

<details>

```c++
#define FMT_HEADER_ONLY
#define SPDLOG_HEADER_ONLY
#include <spdlog/spdlog.h>
#include <spdlog/stopwatch.h>
#include <spdlog/sinks/stdout_color_sinks.h>
#include <spdlog/cfg/env.h>

int main(void) {

    // this paragraph is entirely optional but related to something we do in RcppSpdlog
    std::string logname = "my_demo"; 						// fix a name for this logger
    auto sp = spdlog::get(logname);       					// retrieve existing one
    if (sp == nullptr) sp = spdlog::stderr_color_mt(logname);// or create new one if needed
    spdlog::set_default_logger(sp);                         // and set as default

    // change log pattern (changed from [%H:%M:%S %z] [%n] [%^---%L---%$] )
    spdlog::set_pattern("[%H:%M:%S.%f] [%n] [%^%L%$] [thread %t] %v");

    // eg ./example SPDLOG_LEVEL=my_demo=warn
    spdlog::cfg::load_env_levels();

    // optional too but very cute in the tic-toc sense
    spdlog::stopwatch sw;       							// instantiate a stop watch

    spdlog::info("Welcome to spdlog!");
    spdlog::error("Some error message with arg: {}", 1);
    spdlog::info("Elapsed time: {}", sw);

    spdlog::warn("Easy padding in numbers like {:08d}", 12);
    spdlog::critical("Support for int: {0:d};  hex: {0:x};  oct: {0:o}; bin: {0:b}", 42);
    spdlog::info("Support for floats {:03.2f}", 1.23456);
    spdlog::info("Positional args are {1} {0}..", "too", "supported");
    spdlog::info("{:<30}", "left aligned");
    spdlog::info("Elapsed time: {}", sw);
}
```

</details>

johnkerl:
> Is this something we should add to [CMakeLists.txt](https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/CMakeLists.txt) too where we default it to OFF? Maybe for debug builds it could be INFO.

@nguyenv do you mind if I take this as a follow-on PR?

eddelbuettel:
I looked into this and there is a potentially simpler two-line solution because it is indeed covered upstream. This is should do:

```diff
modified   libtiledbsoma/src/utils/logger.cc
@@ -37,6 +37,7 @@
 #include <spdlog/fmt/ostr.h>
 #include <spdlog/sinks/basic_file_sink.h>
 #include <spdlog/sinks/stdout_color_sinks.h>
+#include <spdlog/cfg/env.h>
 
 namespace tiledbsoma {
 
@@ -72,6 +73,7 @@ Logger::Logger() {
 #endif
     }
     set_level("INFO");
+    spdlog::cfg::load_env_levels();   // eg ./example SPDLOG_LEVEL=my_demo=warn
 }
 
 Logger::~Logger() {
```

Demo, from R for ease of use, but not that this tickles use in R (I also looked into changing the R wrapper package, this is working via the same two-line trick but widely shipped) and then libtiledbsoma in the last two lines:

```sh
$ SPDLOG_LEVEL="debug" Rscript -e 'library(tiledbsoma); sdf <- SOMADataFrameOpen("/tmp/tiledb/somadf_ts")'
[2024-09-09 17:35:28.930] [default] [Process: 3125929] [debug] [SOMADataFrameOpen] uri /tmp/tiledb/somadf_ts ts (now)
[2024-09-09 17:35:28.969] [default] [Process: 3125929] [debug] [TileDBObject] initialize SOMADataFrame with '/tmp/tiledb/somadf_ts' at (now)
[2024-09-09 17:35:28.969] [default] [Process: 3125929] [debug] [TileDBArray$open] Opening SOMADataFrame '/tmp/tiledb/somadf_ts' in READ mode
[2024-09-09 17:35:28.970] [default] [Process: 3125929] [debug] [tiledb_array] query is READ
[2024-09-09 17:35:28.975] [default] [Process: 3125929] [debug] [TileDBArray$update_metadata_cache] updating metadata cache for SOMADataFrame '/tmp/tiledb/somadf_ts' in READ
[2024-09-09 17:35:28.977] [tiledbsoma] [Process: 3125929] [Thread: 3125929] [debug] [SOMAArray] static method 'ctx' opening array '/tmp/tiledb/somadf_ts'
[2024-09-09 17:35:28.977] [tiledbsoma] [Process: 3125929] [Thread: 3125929] [debug] [SOMAArray] opening array '/tmp/tiledb/somadf_ts'
$ 
```

Without the env var nothing is shown or seen:

```sh
$ Rscript -e 'library(tiledbsoma); sdf <- SOMADataFrameOpen("/tmp/tiledb/somadf_ts")'
$ 
```

nguyenv (approved):

johnkerl:
@eddelbuettel kerl/libtiledbsoma-env-logging-level -- thank you! :)

PR #2976, opened 2024-09-11
Use core 2.26.0

johnkerl:
[[sc-53002]](https://app.shortcut.com/tiledb-inc/story/53002/update-projects-to-tiledb-2-26)

Uses:

* https://pypi.org/project/tiledb/0.32.0/
* https://github.com/TileDB-Inc/tiledb-inc.r-universe.dev/pull/10

The core-arifact URLs were updated as always following our procedure via the one-line, error-proof
```
./scripts/update-tiledb-version.py 2.26.0
```


apis/python/setup.py
```diff
@@ -343,7 +343,9 @@ def run(self):
         "scipy",
         # Note: the somacore version is in .pre-commit-config.yaml too
         "somacore==1.0.14",
-        "tiledb~=0.31.0",
+        # TEMP WHILE WE AWAIT WHEELS
```
johnkerl, line 346: This is temporary while @dudoslav resolves a disk-space issue at PyPI

johnkerl:
@nguyenv @eddelbuettel this has a temporary lack-of-pin on tiledb-py -- which is currently ok since we only use typedb-py for unit-testing now -- and there are no storage-format changes in core 2.26.

Let's go ahead & move forward with this for this morning, and I'll update the tiledb-py pin once https://github.com/pypi/support/issues/4744 after which point @dudoslav will be able to publish all PyPI artifacts for tiledb-py 0.32.0.

nguyenv (approved):

eddelbuettel (approved):
If it has to be two-step because of some pypi woes so be it

PR #2978, opened 2024-09-11
[c++] Refine concurrency configuration (`soma.compute_concurrency_level`)

bkmartinjr:
**Issue and/or context:**

Introduces direct control over the native thread pool concurrency level in SOMA. Previously, it was indirectly controlled by the TileDB Core `sm.compute_currency_control` configuration, and would create a thread pool that was 1/2 the size of that value.

**Changes:**

This PR introduces a new context configuration setting:  `soma.compute_currency_level`, which defaults to the host CPU count. It can be set with an numeric value,eg., in Python:

```python
context = tiledbsoma.SOMATileDBContext(tiledb_config={"soma.compute_currency_level": "20"})
```

**Notes for Reviewer:**

sc-54891

the PR also adds:
* improved error handling for malformed config values
* bound the value between 1 and 1024

### DIFF
--- a/apis/python/tests/test_context.py
+++ b/apis/python/tests/test_context.py
@@ -139,7 +139,7 @@ def test_malformed_concurrency_config_value():
 
     import numpy as np
 
-    with pytest.raises((tiledbsoma.SOMAError, RuntimeError)):
+    with pytest.raises(tiledbsoma.SOMAError):
         ctx = tiledbsoma.SOMATileDBContext(
             tiledb_config={"soma.compute_concurrency_level": "not-a-number"}
         )
@@ -149,5 +149,5 @@ def test_malformed_concurrency_config_value():
                 np.arange(100, dtype=np.int64), context=ctx
             ).get_indexer(np.array([0, 1]))
         except Exception as e:
-            print(type(e), e, file=sys.stderr)
+            print("soma.compute_concurrency_level ERROR TYPE", type(e), e, file=sys.stderr)
             raise
--- a/apis/python/tests/test_context.py
+++ b/apis/python/tests/test_context.py
@@ -135,12 +135,19 @@ def test_replace_config_after_construction():
 
 
 def test_malformed_concurrency_config_value():
+    import sys
+
     import numpy as np
 
     with pytest.raises((tiledbsoma.SOMAError, RuntimeError)):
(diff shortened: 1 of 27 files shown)


apis/python/tests/test_context.py
```diff
@@ -137,7 +137,7 @@ def test_replace_config_after_construction():
 def test_malformed_concurrency_config_value():
     import numpy as np
 
-    with pytest.raises(tiledbsoma.SOMAError):
+    with pytest.raises((tiledbsoma.SOMAError, RuntimeError)):
```
johnkerl, line 140: I'll help you fix this -- already done on https://github.com/single-cell-data/TileDB-SOMA/pull/2963

* https://github.com/single-cell-data/TileDB-SOMA/blob/c08b5cf544040a9601fa8c19496ff4af6c7f2c8a/apis/python/src/tiledbsoma/pytiledbsoma.cc#L36-L45
* https://github.com/single-cell-data/TileDB-SOMA/blob/c08b5cf544040a9601fa8c19496ff4af6c7f2c8a/apis/python/src/tiledbsoma/soma_array.cc#L348-L352 (one example callsite)

johnkerl: Also
https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html

bkmartinjr: Broader issue - filed as sc-54978

johnkerl (approved):
LGTM but I do defer to @nguyenv

apis/python/tests/test_context.py
```diff
@@ -131,3 +132,16 @@ def test_replace_config_after_construction():
             new_tdb_ctx = new_soma_ctx.tiledb_ctx
         mock_ctx.assert_called_once()
         assert new_tdb_ctx.config()["vfs.s3.region"] == "us-west-2"
+
+
+def test_malformed_concurrency_config_value():
+    import numpy as np
```
nguyenv, line 138: Why import in the test instead of up top?

nguyenv (approved):

apis/python/tests/test_context.py
bkmartinjr: Just trying to isolate test dependencies.  But not a major issue either way....

PR #2975, opened 2024-09-11
[c++] Fix `ManagedQuery` usage in `SOMAArray` write path

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2928

**Changes:**

**Notes for Reviewer:**

Will split this into smaller PRs

PR #2979, opened 2024-09-11
[c++] Remove `ArrayBuffers` from `SOMAArray`

nguyenv:
**Issue and/or context:**

[[sc-54744]](https://app.shortcut.com/tiledb-inc/story/54744/python-soma-read-does-an-open-on-an-already-open-array)

https://github.com/single-cell-data/TileDB-SOMA/pull/2975

This is the first of multiple PRs that will be split out from above

**Changes:**

The `ArrayBuffers` are now fully managed by `ManagedQuery` and not `SOMAArray`

### DIFF
--- a/libtiledbsoma/src/soma/managed_query.cc
+++ b/libtiledbsoma/src/soma/managed_query.cc
@@ -215,6 +215,30 @@ void ManagedQuery::set_column_data(
     }
 }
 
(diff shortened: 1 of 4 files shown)


eddelbuettel (approved):

johnkerl (approved):
✨

Issue #2928, opened 2024-08-26
[c++/python] Memory leak in Python sparse-array write


johnkerl:
## Problem

There is a memory leak when writing SOMA sparse arrays (maybe other types).  Perhaps the Arrow table being written is leaked.  Test code below.

Note:

* Issue only occurs in 1.13 or later
* Workaround is to pin to <1.13

The leak is large enough that it prevents copying any large-ish array.  For example, trying to read and rewrite a Census X layer (with a new schema), and that OOMs quickly on a 128GiB host.

Reported by @bkmartinjr.

[[sc-53264]](https://app.shortcut.com/tiledb-inc/story/53264/memory-leak-in-soma-python-sparse-array)

## Repro

Test code (will create a `test_array` in the current working directory):

```python
from __future__ import annotations

import gc
import sys

import tiledbsoma as soma


def copy(from_uri, to_uri, context):
    with soma.SparseNDArray.open(from_uri, mode="r", context=context) as X_from:
        with soma.SparseNDArray.open(to_uri, mode="w", context=context) as X_to:
            for i, tbl in enumerate(X_from.read().tables()):
                print(f"Read {len(tbl)}")
                X_to.write(tbl)

                del tbl
                gc.collect()

                if i == 10:  # OOMs w/o this break
                    break


def create(src_uri, dst_uri, context):
    with soma.open(src_uri, mode="r", context=context) as X:
        n_obs, n_var = X.shape
        type = X.schema.field("soma_data").type

    a = soma.SparseNDArray.create(
        dst_uri,
        type=type,
        shape=(n_obs, n_var),
        platform_config={
            "tiledb": {
                "create": {
                    "capacity": 2**16,
                    "dims": {
                        "soma_dim_0": {
                            "tile": 64,
                            "filters": [{"_type": "ZstdFilter", "level": 9}],
                        },
                        "soma_dim_1": {
                            "tile": 2048,
                            "filters": [
                                "ByteShuffleFilter",
                                {"_type": "ZstdFilter", "level": 9},
                            ],
                        },
                    },
                    "attrs": {
                        "soma_data": {
                            "filters": [
                                "ByteShuffleFilter",
                                {"_type": "ZstdFilter", "level": 9},
                            ]
                        }
                    },
                    "cell_order": "row-major",
                    "tile_order": "row-major",
                    "allows_duplicates": True,
                },
            }
        },
        context=context,
    )
    a.close()
    print(f"Array created at {dst_uri}")


[...]

PR #2989, opened 2024-09-12
[c++] Fix `ManagedQuery` usage in `SOMAArray` write path

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2928

**Changes:**

Commits should be viewed individually. Although much of this code already existed prior to this PR, there were enough  modifications that the changes were too noisy to read clearly.

1. First commit labels all the old code as legacy
2. `ManagedQuery::setup_write_column` takes in a column name, number of elements, and pointers to buffers that contain the passed in data. This use to be three separate calls to `ManagedQuery::setup_column_data`, `ColumnBuffer::set_data`, and `ManagedQuery::set_column_data` but now simplified
3. The old `SOMAArray::set_array_data` made a copy of the passed in `ArrowTable` and then passed the casted buffers to `ColumnBuffer::set_data`. We now iterate through each column calling `SOMAArray::_cast_column`, which casts the values into temporary vectors and the passes the `std::vector::data()` pointer into `ManagedQuery::setup_write_column`
4. `SOMAArray::_cast_column` handles all permutations of attributes with or without enumerations and columns with or without dictionaries. 
   1. In the case where the attribute has a enumeration but the passed in column does not have a dictionary, error out. 
   2. In the case where the attribute is not enumerated but the passed column does have a dictionary, call `SOMAArray::_promote_indexes_to_values` which maps the dictionary's indexes to the associated dictionary values and passes that dictionary value mapping vector into `ManagedQuery::setup_write_column`
   3. In all other cases, call `SOMAArray::_cast_column_aux` for further processing
5. `SOMAArray::_promote_indexes_to_values` calls `SOMAArray::_cast_dictionary_values`. It retrieves the dictionary's indexes via `SOMAArray::_get_index_vector` and then iterates through the index vector and maps and casts to the correct dictionary value
6. `SOMAArray::_cast_column_aux` calls `SOMAArray::_set_column`.
   1. If the attribute is enumerated and the column has a dictionary, we need to write out the dictionary indexes and potentially extend the enumeration if there are any new values
   2. In all other cases, place the values into a a temporary vector, iterate and cast these values to the type on disk, and pass the casted vector into `ManagedQuery::setup_write_buffer`
7. `SOMAArray::_extend_enumeration` calls `SOMAArray::_extend_and_evolve_schema` which check if any new enumerations have been introduced
   1. If so, then we need to call `SOMAArray::_remap_indexes` which calls `SOMAArray::_remap_indexes_aux` to iterate through the original indexes and shift them to match the new extended enumeration on-disk. `SOMAArray::_cast_shifted_indexes` casts the shifted index values to the type on disk and passes this into `ManagedQuery::setup_write_buffer` 
   2. If not, then pass the dictionary's index as-is into `ManagedQuery::setup_write_column`
8. Return a `vector<uint8_t>` of Boolean mappings instead of casting inplace
9. Correct unit test to pass in offset buffer for variable length string columns
10. Last commit removes all dead code


libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -966,6 +966,235 @@ bool SOMAArray::_cast_column_aux<bool>(
...
+        _remap_indexes(
+            column_name,
+            extended_enmr,
+            enums_in_write,
+            index_schema,
+            index_array);
+
+        return true;
+    } else {
+        // We do not need to extend the enumeration; write as-is
```
johnkerl, line 906: Looks like you're losing this bugfix #2897 -- we do need to remap indices even if the enum isn't being extended:

* First write has red,yellow,green at 0,1,2
* Second write has yellow,green at 0,1 needing remap to 1,2

I created a unit-test case for this on #2897 ... 👀

johnkerl: Aha, see line 1192 -- this is an old omission on my part, not a new omission on your part -- my #2897 didn't fix both the string and non-string variants of `_extend_and_evolve_schema` but it should have -- and on 2897 I only unit-tested the enum-of-string case ...

I can put up a separate/follow-on PR, to implement my incomplete bugfix, if you like -- or you can do it on this one; I'm fine either way

libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -1195,6 +1195,31 @@ bool SOMAArray::_extend_and_evolve_schema<std::string>(
     return false;
 }
 
+std::vector<uint8_t> SOMAArray::_cast_bit_to_uint8(
```
johnkerl, line 1000: Does this need to be a method of `SOMAArray`? It only uses its args `schema` and `array`. It could be a utils function if I read correctly ...

libtiledbsoma/test/unit_soma_dataframe.cc
```diff
@@ -172,11 +172,12 @@ struct VariouslyIndexedDataFrameFixture {
         auto i64_data = std::vector<int64_t>({sjid_base + 1, sjid_base + 2});
         auto u32_data = std::vector<uint32_t>({1234, 5678});
         auto str_data = std::vector<std::string>({"apple", "bat"});
+        auto str_offsets = std::vector<uint64_t>({0, 5, 8});
```
johnkerl, line 175: Thanks for fixing this! My old bad. :|

johnkerl (approved):
This is a well-organized PR, easy to review and easy for everyone to understand later. Thank you!!

There are a couple items here, both of which can be follow-on PRs if you like.

libtiledbsoma/src/soma/soma_array.cc
nguyenv: Yeah it would probably be better in `util/util.h`

libtiledbsoma/src/soma/soma_array.cc
nguyenv: I will take care of it in a follow-up PR.

nguyenv:
I will move the bit to uin8_t conversion method and do the extend enumeration fix in two separate follow up PRs.

PR #2990, opened 2024-09-13
[c++] Fix bug in `nnz` of variant-indexed dataframes

johnkerl:
**Issue and/or context:** Pre-existing bug found while working on issue #2407 / [[sc-51048]](https://app.shortcut.com/tiledb-inc/story/51048).

I ran `sdf.count()` on a variant-indexed dataframe which has dim0 being a string (which [the spec](https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md) allows even though most users don't take advantage of this). I got the error

```
[TileDB::FragmentInfo] Error: Cannot get non-empty domain; Dimension is variable-sized
```

There is a repro now built into the unit-test cases on this PR.

**Changes:**

Only use the code that assumes dim0 is `int64` `soma_joinid` when that is in fact the case.

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/soma/soma_array.cc
+++ b/libtiledbsoma/src/soma/soma_array.cc
@@ -1205,13 +1205,32 @@ uint64_t SOMAArray::nnz() {
     // compute total_cell_num while going through the loop
     uint64_t total_cell_num = 0;
     std::vector<std::array<uint64_t, 2>> non_empty_domains(fragment_count);
+
+    // The loop after this only works if dim 0 is int64 soma_joinid.
+    // That's the case for _almost_ all SOMADataFrame objects, but
+    // not the "variant-indexed" ones: the SOMA spec only requires
+    // that soma_joinid be present as a dim or an attr.
+    auto dim = tiledb_schema()->domain().dimension(0);
+    auto dim_name = dim.name();
+    auto type_code = dim.type();
+    if (dim_name != "soma_joinid" || type_code != TILEDB_INT64) {
+        LOG_DEBUG(fmt::format(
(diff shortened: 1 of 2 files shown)


johnkerl:
Leaving in draft while I address a unit-test failure

libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -1205,13 +1205,32 @@ uint64_t SOMAArray::nnz() {
...
     std::vector<std::array<uint64_t, 2>> non_empty_domains(fragment_count);
+
+    // The loop after this only works if dim 0 is int64 soma_joinid.
+    // That's the case for _almost_ all SOMADataFrame objects, but
+    // not the "variant-indexed" ones: the SOMA spec only requires
+    // that soma_joinid be present as a dim or an attr.
+    auto dim = tiledb_schema()->domain().dimension(0);
+    auto dim_name = dim.name();
+    auto type_code = dim.type();
+    if (dim_name != "soma_joinid" || type_code != TILEDB_INT64) {
```
jp-dark, line 1216: Shouldn't the "fast" version work for any int64 dimension whether or not it's the `soma_joinid`?

johnkerl: if they have a dim which is of type `int64` but it isn't `soma_joinid` -- which is custom -- then we should be doing cell-counting i think

jp-dark: In theory this check is redundant then since `soma_joinid` is always `int64`, but it doesn't hurt to have the extra check.

PR #2993, opened 2024-09-13
[c++] Remap indexes even if no new enums were added

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/pull/2989#discussion_r1759104645

**Changes:**

Always call remap indexes even if no new enumerations were added. These were done correctly in the specialized templates for string and Boolean types but needed to be added for the general case as well. Also add unit tests in Python to check for string and integer category value types.


johnkerl (approved):
🚢 
Thank you thank you!!! :)

PR #2994, opened 2024-09-13
[c++] Centralize some `nanoarrow` helpers

johnkerl:

**Changes:**

Extracting readback of core domain, current domain, and non-empty domain are all currently:

* Done slotwise within `libtiledbsoma`, templatized;
* For Python: there is a loop over dim names in the `pybind11` layer lke this https://github.com/single-cell-data/TileDB-SOMA/blob/b3feada0d1282442d1190f56a43670a32dad97ae/apis/python/src/tiledbsoma/soma_array.cc#L671-L830;
* For R: currently these are using TileDB-R -- which is a dependency we're currently removing -- and I don't want to imitate the above in R

The best option for Python/R interop is:

* Push down loop-over-dim-names to `libtiledbsoma`
* Make more use of `nanoarrow` to interface with Python and R

The current PR simply moves some code currently used in `libtiledbsoma`'s unit-test logic to `ArrowAdapter`, where it can -- and will -- be used for readback of core domain, current domain, and non-empty domain for #2407.

**Notes for Reviewer:**

This is just a code-move. If all existing unit tests pass, that is success. No new unit tests are introduced here, since no new unit tests are needed.

### DIFF
--- a/libtiledbsoma/src/utils/arrow_adapter.cc
+++ b/libtiledbsoma/src/utils/arrow_adapter.cc
@@ -1302,6 +1302,9 @@ std::unique_ptr<ArrowArray> ArrowAdapter::make_arrow_array_parent(
     arrow_array->n_children = num_columns;
     arrow_array->release = &ArrowAdapter::release_array;
     arrow_array->children = new ArrowArray*[num_columns];
+    for (int i = 0; i < num_columns; i++) {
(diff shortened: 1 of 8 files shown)


nguyenv (approved):

johnkerl:
Thanks @nguyenv ! :)

PR #2996, opened 2024-09-14
[r] Documentation refresher

eddelbuettel:
**Issue and/or context:**

`R CMD check` was nagging about some missing fields as well as one undefined global

**Changes:**

Injected missing `is_arrow` documentation, removed one stale/unused assignment with a missing global and re-ran `roxygenize()` to refresh Rd files.

`R CMD check` is now down to complaining about `tiledb` being too frequently in Depends, Imports, Suggests or Enhances (which get taken care of) and a pair of missing Rd cross-references, which is in sum a lot quieter.

**Notes for Reviewer:**

[SC 65150](https://app.shortcut.com/tiledb-inc/story/55150/r-package-documentation-refresher_

### DIFF
--- a/apis/r/R/Init.R
+++ b/apis/r/R/Init.R
@@ -25,10 +25,6 @@
     } else {
       .pkgenv[["use_current_domain_transitional_internal_only"]] <- FALSE
     }
-    if (rpkg_lib_version != soma_lib_version) {
-      msg <- sprintf("TileDB Core version %s used by TileDB-R package, but TileDB-SOMA uses %s [%s]",
-                     sQuote(rpkg_lib_version), sQuote(soma_lib_version), sQuote(cdmsg))
-    }
 }
 
 # This is temporary only. Please see:
--- a/apis/r/R/RcppExports.R
+++ b/apis/r/R/RcppExports.R
@@ -60,6 +60,7 @@ c_group_put_metadata <- function(xp, key, obj) {
 
 #' Get nnumber of metadata items
 #' @param uri The array URI
+#' @param is_array A boolean to indicate array or group
 #' @param ctxxp An external pointer to the SOMAContext wrapper
 #' @export
 get_metadata_num <- function(uri, is_array, ctxxp) {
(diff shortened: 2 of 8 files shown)


apis/r/R/Init.R
```diff
@@ -25,10 +25,6 @@
     } else {
       .pkgenv[["use_current_domain_transitional_internal_only"]] <- FALSE
     }
-    if (rpkg_lib_version != soma_lib_version) {
```
johnkerl, line 28: Appears to be unrelated to this PR -- should be a separate PR if I understand correctly.

johnkerl (changes requested):
Thanks for  the audit! :)

apis/r/R/Init.R
eddelbuettel: How come?  `git blame` pins it at five months ago

```diff
5 months ago    Dirk Eddelbuet  0a6ca6b68│ 15 │     if (rpkg_lib_version != soma_lib_version) {
                                         │ 16 │         msg <- sprintf("TileDB Core version %s used by TileDB-R package, but TileDB-SOMA uses %s",
                                         │ 17 │                        sQuote(rpkg_lib_version), sQuote(soma_lib_version))
                                         │ 18 │         stop(msg, call. = FALSE)
                                         │ 19 │     }
```

eddelbuettel: If you expand the line you see that

- this is copy-paste from a few lines above
- it uses a variable that is not defined (which `R CMD check` flagged and how I got there)
- assigns to a variable that is not used

So can we maybe turn it around and let this PR remove the non-working code and re-inject in a future PR what you intended, fully fleshed=out?

eddelbuettel: For reference:

https://github.com/single-cell-data/TileDB-SOMA/blob/b3feada0d1282442d1190f56a43670a32dad97ae/apis/r/R/Init.R#L28-L31

where `cdmsg` is a free variable never assigned, and `msg` is a temporary not used.

johnkerl: Thanks for the clarify @eddelbuettel !

johnkerl (approved):

PR #2997, opened 2024-09-16
[r,c++] Use SOMA Context object throughout

eddelbuettel:
**Issue and/or context:**

SOMA Context objects are used throughout the C++ library part of SOMA.  Where we previously serialized context and configuration to vector, maps or lists of key-value pairs, we now hold on to an allocated object for longer.

**Changes:**

This PR changes how we provide such a context from the R package by allocating one as a shared pointer and passing it around as an external pointer throuch which the SOMA context object can be accessed.  Where needed, the SOMA context also contains a (core) TileDB context object.  

We currently cache the most-recently created object at the package level and provide it if no new one is demanded.  This scheme is exactly what tiledb-r does, and acts reliably and robust via a single access point `soma_context()`.  One unit test show how to create a different configuration without caching, using it and then reverting to the default.

SOMA classes can obtain their initial default object via the instantiating factory methods.  Access is then provided via a member function, allowing both for simple re-use as well variations (object by object) as needed.  How to split the core object values off the `platform_config` object is left for a subsequent PR that will refine the interface.

All tests pass.

**Notes for Reviewer:**

[SC 55146](https://app.shortcut.com/tiledb-inc/story/55146/r-c-use-somacontext-throughout)

### DIFF
--- a/apis/r/R/SOMADataFrame.R
+++ b/apis/r/R/SOMADataFrame.R
@@ -192,7 +192,7 @@ SOMADataFrame <- R6::R6Class(
       spdl::debug("[SOMADataFrame$read] calling sr_setup for {} at ({},{})", self$uri,
                   private$tiledb_timestamp[1], private$tiledb_timestamp[2])
       cfg <- as.character(tiledb::config(self$tiledbsoma_ctx$context()))
-      rl <- sr_setup(uri = self$uri,
+      sr <- sr_setup(uri = self$uri,
                      config = cfg, # needed ?
                      private$.soma_context,
                      colnames = column_names,
@@ -200,8 +200,7 @@ SOMADataFrame <- R6::R6Class(
                      dim_points = coords,
                      timestamprange = self$.tiledb_timestamp_range,  # NULL or two-elem vector
                      loglevel = log_level)
-      private$ctx_ptr <- rl$ctx # is this still needed?
-      TableReadIter$new(rl$sr)
+      TableReadIter$new(sr)
     },
 
     #' @description Update (lifecycle: maturing)
@@ -400,9 +399,6 @@ SOMADataFrame <- R6::R6Class(
       }
 
       schema
-    },
-
-    # Internal variable to hold onto context returned by sr_setup
-    ctx_ptr = NULL
+    }
   )
 )
--- a/apis/r/R/SOMADataFrame.R
+++ b/apis/r/R/SOMADataFrame.R
@@ -356,10 +356,7 @@ SOMADataFrame <- R6::R6Class(
     #' (lifecycle: maturing)
     #' @return Logical
     tiledbsoma_has_upgraded_domain = function() {
-      has_current_domain(
-        self$uri,
-        config=as.character(tiledb::config(self$tiledbsoma_ctx$context()))
-      )
+      has_current_domain(self$uri, private$.soma_context)
     }
 
   ),
--- a/apis/r/R/SOMADataFrame.R
+++ b/apis/r/R/SOMADataFrame.R
@@ -191,9 +191,7 @@ SOMADataFrame <- R6::R6Class(
       }
       spdl::debug("[SOMADataFrame$read] calling sr_setup for {} at ({},{})", self$uri,
                   private$tiledb_timestamp[1], private$tiledb_timestamp[2])
-      cfg <- as.character(tiledb::config(self$tiledbsoma_ctx$context()))
       sr <- sr_setup(uri = self$uri,
-                     config = cfg, # needed ?
                      private$.soma_context,
                      colnames = column_names,
                      qc = value_filter,
--- a/apis/r/R/SOMADataFrame.R
+++ b/apis/r/R/SOMADataFrame.R
@@ -27,6 +27,7 @@ SOMADataFrame <- R6::R6Class(
       schema,
       index_column_names = c("soma_joinid"),
       platform_config = NULL,
+      soma_context = NULL,
       internal_use_only = NULL
     ) {
       if (is.null(internal_use_only) || internal_use_only != "allowed_use") {
@@ -42,6 +43,16 @@ SOMADataFrame <- R6::R6Class(
       # Parse the tiledb/create/ subkeys of the platform_config into a handy,
       # typed, queryable data structure.
       tiledb_create_options <- TileDBCreateOptions$new(platform_config)
+      #print(str(tiledb_create_options$to_list()))
+
+      ## TODO: the private$.soma_context object contains an xptr to a struct containtaining
+      ## a shared point SOMAContext which contains a TileDB Context. The cached variable
+      ## is created in TileDBObject$initialize() so the argument on this signature is not
+      ## strictly needed
+      if (!is.null(soma_context)) {
+          spdl::debug("[SOMADataFrame$create] New soma context")
+          private$.soma_context <- soma_context
+      }
 
       ## we (currently pass domain and extent values in an arrow table (i.e. data.frame alike)
       ## where each dimension is one column (of the same type as in the schema) followed by three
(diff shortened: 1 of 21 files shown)


apis/r/R/SOMADataFrame.R
```diff
@@ -42,6 +43,16 @@ SOMADataFrame <- R6::R6Class(
       # Parse the tiledb/create/ subkeys of the platform_config into a handy,
       # typed, queryable data structure.
       tiledb_create_options <- TileDBCreateOptions$new(platform_config)
+      #print(str(tiledb_create_options$to_list()))
```
johnkerl, line 46: remove please

apis/r/R/SOMADataFrame.R
```diff
@@ -42,6 +43,16 @@ SOMADataFrame <- R6::R6Class(
...
+
+      ## TODO: the private$.soma_context object contains an xptr to a struct containtaining
```
johnkerl, line 48: please file a GitHub issue to track, or Shortcut story, and include its link here

apis/r/R/SOMASparseNDArray.R
```diff
@@ -46,15 +46,14 @@ SOMASparseNDArray <- R6::R6Class(
         coords <- private$.convert_coords(coords)
       }
 
-      cfg <- as.character(tiledb::config(self$tiledbsoma_ctx$context()))
-      rl <- sr_setup(uri = self$uri,
-                     config = cfg,
+      ## not needed anymore  cfg <- as.character(tiledb::config(self$tiledbsoma_ctx$context()))
```
johnkerl, line 49: please remove

apis/r/R/ephemeral.R
```diff
@@ -330,8 +330,8 @@ EphemeralCollectionBase <- R6::R6Class(
       names(members) <- names(private$.data)
       for (i in seq_along(members)) {
         members[[i]] <- list(
-          type = tiledb::tiledb_object_type(private$.data[[i]]$uri),
-          uri = private$.data[[i]]$uri,
+          type = get_tiledb_object_type(private$.data[[i]]$uri, private$.soma_context),
+            uri = private$.data[[i]]$uri,
```
johnkerl, line 334: nit: indentation

johnkerl (commented):
This is awesome! Just a few minor notes at this point.

apis/r/R/SOMADataFrame.R
eddelbuettel: Will do in a few hours once I synced with @mojaveazure.  The comments provide context.

apis/r/R/SOMADenseNDArray.R
```diff
@@ -51,7 +51,7 @@ SOMADenseNDArray <- R6::R6Class(
       }
       coords <- private$.convert_coords(coords)
 
-      cfg <- as.character(tiledb::config(self$tiledbsoma_ctx$context()))
+      soma_context <- soma_context()  # package-level cached instance
```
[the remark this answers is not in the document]
eddelbuettel: When I look at the function I do not think we have an outer object here that would provide it.  Did I miss something?

apis/r/R/SOMADataFrame.R
eddelbuettel: Actually I will look into removing it from `create()` and nix the comment with it.   This was from a first pass, it is now clearer that no `create()` method should be needed.  

Done in 091e1094 

Nice catch!

apis/r/R/SOMASparseNDArray.R
eddelbuettel: Done in 091e1094

apis/r/src/riterator.cpp
```diff
@@ -104,8 +104,12 @@ Rcpp::List sr_setup(const std::string& uri,
 
     auto tdb_result_order = get_tdb_result_order(result_order);
 
-    auto ptr = new tdbs::SOMAArray(OpenMode::read, uri, name, platform_config,
-                                   column_names, batch_size,
+    // old interface
```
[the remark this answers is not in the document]
eddelbuettel: Done in 091e1094

apis/r/R/SOMADenseNDArray.R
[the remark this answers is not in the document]
johnkerl: The paradigm on the Python side is:

* Every class takes an optional soma context
* If omitted, use something like `soma_context()`
* If provided, use that
* Then when a higher-level object instantiates a child, it passes along the soma context it has

So:

* Say the top-level thing being opened by the user is a `SOMAExperiment`
* If they open that experiment with their own soma context, then that same soma context will be used when they access `exp$obs`, etc.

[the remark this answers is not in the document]
johnkerl: Oops, I replied, then just now saw your reply! 😎

johnkerl (approved):

apis/r/R/SOMANDArrayBase.R
```diff
@@ -37,6 +37,8 @@ SOMANDArrayBase <- R6::R6Class(
       ## .is_sparse field is being set by dense and sparse private initialisers, respectively
       private$.type <- type                 # Arrow schema type of data
 
+      private$.soma_context <- soma_context() # package-level cache access
+
```
mojaveazure, line 41: We should be picking up the context from `$new()`

apis/r/R/TileDBObject.R
```diff
@@ -15,7 +15,8 @@ TileDBObject <- R6::R6Class(
     #' @param internal_use_only Character value to signal this is a 'permitted' call,
     #' as `new()` is considered internal and should not be called directly.
     initialize = function(uri, platform_config = NULL, tiledbsoma_ctx = NULL,
-                          tiledb_timestamp = NULL, internal_use_only = NULL) {
+                          tiledb_timestamp = NULL, internal_use_only = NULL,
+                          soma_context = NULL) {
```
mojaveazure, line 19: I thought we had agreed on not exposing `soma_context()` via `$new()` at this point in time https://github.com/single-cell-data/TileDB-SOMA/pull/2988#discussion_r1757629865

apis/r/R/utils.R
```diff
@@ -165,13 +165,13 @@ uns_hint <- function(type = c('1d', '2d')) {
   }
   x$reopen("READ", tiledb_timestamp = x$tiledb_timestamp)
   dimname <- x$dimnames()[axis + 1L]
-  rl <- sr_setup(
+  sr <- sr_setup(
     uri = x$uri,
-    config = as.character(tiledb::config(x$tiledbsoma_ctx$context())),
+    soma_context(),
```
mojaveazure, line 170: Note for future work: this should be replaced with fetching the context from `x`; currently available with
```suggestion
    x$.__enclos_env__$private$.soma_context
```

mojaveazure (changes requested):

apis/r/R/SOMANDArrayBase.R
eddelbuettel: Good catch, I think you are right that this is redundant too (similar to the one I removed in SOMADataFrame).

Fixed in a632da13

apis/r/R/TileDBObject.R
eddelbuettel: Yes.  This gives the object to also set one here but it is not a requirement.  Having the entry point is good for testing, methinks.  See the comment a few lines futher down as well as the need to rework / let go of `tiledbsoma_ctx`.  I think this is all best addressed in a next PR.

https://github.com/single-cell-data/TileDB-SOMA/blob/091e1094380998766995eb1f2bac9ebdff33b35b/apis/r/R/TileDBObject.R#L34-L51

apis/r/R/utils.R
eddelbuettel: Fair point but not urgent and on the fence as whether we should use opaque internals.  "Future work", as you say.

mojaveazure (approved):

PR #3002, opened 2024-09-17
[c++] Use core 2.26.1

johnkerl:

As always, core artifacts were updated using the error-proof and automated `./scripts/update-tiledb-version.py 2.26.1`.

### DIFF
--- a/scripts/bld
+++ b/scripts/bld
@@ -11,7 +11,7 @@ arg() { echo "$1" | sed "s/^${2-[^=]*=}//" | sed "s/:/;/g"; }
 build="Release"
 prefix=""
 tiledb=""
-cmake_verbose="true"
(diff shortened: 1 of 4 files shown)


scripts/bld
```diff
@@ -11,7 +11,7 @@ arg() { echo "$1" | sed "s/^${2-[^=]*=}//" | sed "s/:/;/g"; }
 build="Release"
 prefix=""
 tiledb=""
-cmake_verbose="false"
+cmake_verbose="true"
```
eddelbuettel, line 14: Related or accident?

eddelbuettel (approved):

scripts/bld
johnkerl: Accident -- thank you!

PR #3001, opened 2024-09-16
[python] Python 3.12 support

johnkerl:
For issue #1849 -- which has a long history.

See also #2999 and [[sc-53002]](https://app.shortcut.com/tiledb-inc/story/53002/update-projects-to-tiledb-2-26).

This needs https://github.com/single-cell-data/SOMA/pull/222

PR #3005, opened 2024-09-17
[python] Move temporary feature flag into a module

johnkerl:
**Issue and/or context:** Follow-up from here: https://github.com/single-cell-data/TileDB-SOMA/pull/3001#issuecomment-2355914110

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/_flags.py
+++ b/apis/python/src/tiledbsoma/_flags.py
@@ -5,14 +5,9 @@
 
 import os
 
-# Temporary for https://github.com/single-cell-data/TileDB-SOMA/issues/2407
-_new_shape_feature_flag = os.getenv("SOMA_PY_NEW_SHAPE") is not None
+# This is temporary for
+# https://github.com/single-cell-data/TileDB-SOMA/issues/2407.  It will be
+# removed once https://github.com/single-cell-data/TileDB-SOMA/issues/2407 is
+# complete.
 
-
-def _new_shape_feature_flag_enabled() -> bool:
-    """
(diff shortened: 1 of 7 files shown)


apis/python/src/tiledbsoma/_flags.py
```diff
@@ -0,0 +1,18 @@
...
+_new_shape_feature_flag = os.getenv("SOMA_PY_NEW_SHAPE") is not None
+
+
+def _new_shape_feature_flag_enabled() -> bool:
+    """
+    This is temporary only and will be removed once
+    https://github.com/single-cell-data/TileDB-SOMA/issues/2407
+    is complete.
+    """
+    return _new_shape_feature_flag
```
thetorpedodog, line 18: What's the motivation behind having this live in a function? I would make `NEW_SHAPE_FEATURE_FLAG_ENABLED` be a public constant and use that.

thetorpedodog (approved):
the move is fine; I do have one suggestion for the implementation while you’re moving things that you can take or leave.

apis/python/src/tiledbsoma/_flags.py
johnkerl: @thetorpedodog sure, whatever you like.

johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/3005/commits/a8b5909bdc0d6c037454eb389a7afbe318c92a3d

johnkerl:
Thanks @thetorpedodog ! :)

PR #3006, opened 2024-09-17
[r] Remove core version check

johnkerl:
This was very much put in place for a good reason. However, as of recently we no longer use the tiledb-r path to read or write arrays.

### DIFF
--- a/apis/r/R/Init.R
+++ b/apis/r/R/Init.R
@@ -8,16 +8,6 @@
     ## create a slot for somactx in per-package enviroment, do no fill it yet to allow 'lazy load'
     .pkgenv[["somactx"]] <- NULL
 
-    rpkg_lib <- tiledb::tiledb_version(compact = FALSE)
-    # Check major and minor but not micro: sc-50464
-    rpkg_lib_version <- paste(rpkg_lib[["major"]], rpkg_lib[["minor"]], sep = ".")
-    soma_lib_version <- libtiledbsoma_version(compact = TRUE, major_minor_only = TRUE)
-    if (rpkg_lib_version != soma_lib_version) {
-        msg <- sprintf("TileDB Core version %s used by TileDB-R package, but TileDB-SOMA uses %s",
-                       sQuote(rpkg_lib_version), sQuote(soma_lib_version))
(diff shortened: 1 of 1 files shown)


apis/r/R/Init.R
```diff
@@ -8,16 +8,6 @@
...
-    rpkg_lib <- tiledb::tiledb_version(compact = FALSE)
-    # Check major and minor but not micro: sc-50464
-    rpkg_lib_version <- paste(rpkg_lib[["major"]], rpkg_lib[["minor"]], sep = ".")
-    soma_lib_version <- libtiledbsoma_version(compact = TRUE, major_minor_only = TRUE)
-    if (rpkg_lib_version != soma_lib_version) {
-        msg <- sprintf("TileDB Core version %s used by TileDB-R package, but TileDB-SOMA uses %s",
-                       sQuote(rpkg_lib_version), sQuote(soma_lib_version))
-        stop(msg, call. = FALSE)
-    }
-
```
eddelbuettel, line 20: I think we could consider keeping it as an informational message.  Something like

```r
`packageStartupMessage(msg)        
```

eddelbuettel (dismissed):
Please also remove the `<=` part in DESCRIPTION just keeping the more innocent `>=` for the relation to tiledb-r

apis/r/R/Init.R
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/3006/commits/0af41059d340356536c01e7394fd265c7ef25c1b

eddelbuettel (approved):
:rocket:

mojaveazure (approved):
We can also get rid of the "Install tiledb-r" step of CI as that was necessary only for the doublepin. Now that that's gone, we can fall back on the regular install dependencies step

eddelbuettel:
@mojaveazure Maybe in a follow-up PR?

mojaveazure:
Yeah, that should be a follow-up, I was just making a note of what else can be done thanks to the relaxing of tiledb-r versions. The current CI is working, and that's good enough for now

PR #3022, opened 2024-09-20
[python] Update names/methods to match somacore

jp-dark:
* Update names/methods to match somacore
 * Move functions from `_coordinates.py` to `_spatial_util.py`


nguyenv (approved):
lgtm but maybe we can add a util function that takes in a sequence of axes names to create the `CoordinateSpace`

PR #3012, opened 2024-09-17
[c++] Implementation and unit testing for domainish accessors

johnkerl:
**Issue and/or context:** As tracked on issue #2407 / [[sc-51048]](https://app.shortcut.com/tiledb-inc/story/51048).

This is atop #3011; these will replace #2995.

Note that the intended Python and R API changes are all agreed on and finalized as described in #2407.

**Changes:**

**Notes for Reviewer:**

Python/R PRs will be separate.

PR #3027, opened 2024-09-21
[python/r] Use pushdown domainish accessors at the Python/R UX level

johnkerl:

See also [[sc-55659]](https://app.shortcut.com/tiledb-inc/story/55659/r-current-domain-tasks).

**Changes:**

Following #3012 we can now connect tiledbsoma-py to use the full domainish accessors from `libtiledbsoma`.

**Notes for Reviewer:**

R mods to come on this same PR.

This PR is a work in progress. It is not ready for review.

### DIFF
--- a/apis/python/src/tiledbsoma/_tdb_handles.py
+++ b/apis/python/src/tiledbsoma/_tdb_handles.py
@@ -12,10 +12,10 @@ import abc
 import enum
 from typing import (
     Any,
-    Callable,
     Dict,
     Generic,
     Iterator,
+    List,
     Mapping,
     MutableMapping,
     Optional,
@@ -30,7 +30,6 @@ from typing import (
 import attrs
 import numpy as np
 import pyarrow as pa
-from numpy.typing import DTypeLike
 from somacore import options
 from typing_extensions import Literal, Self
 
@@ -363,41 +362,36 @@ class SOMAArrayWrapper(Wrapper[_ArrType]):
     def ndim(self) -> int:
         return len(self._handle.dimension_names)
 
-    def _get_and_cast_domain(
-        self, domain_slot_getter: Callable[[str, DTypeLike], Tuple[object, object]]
+    def _cast_domainish(
+        self, domainish: List[Any]
     ) -> Tuple[Tuple[object, object], ...]:
         result = []
-        for name in self._handle.dimension_names:
-            dtype = self._handle.schema.field(name).type
-            if pa.types.is_timestamp(dtype):
-                np_dtype = np.dtype(dtype.to_pandas_dtype())
-                dom = domain_slot_getter(name, np_dtype)
+        for i, slot in enumerate(domainish):
+
+            arrow_type = slot[0].type
+            if pa.types.is_timestamp(arrow_type):
+                pandas_type = np.dtype(arrow_type.to_pandas_dtype())
                 result.append(
-                    (
-                        np_dtype.type(dom[0], dtype.unit),
-                        np_dtype.type(dom[1], dtype.unit),
+                    tuple(
+                        pandas_type.type(e.cast(pa.int64()).as_py(), arrow_type.unit)
+                        for e in slot
                     )
                 )
             else:
-                if pa.types.is_large_string(dtype) or pa.types.is_string(dtype):
-                    dtype = np.dtype("U")
-                elif pa.types.is_large_binary(dtype) or pa.types.is_binary(dtype):
-                    dtype = np.dtype("S")
-                else:
-                    dtype = np.dtype(dtype.to_pandas_dtype())
-                result.append(domain_slot_getter(name, dtype))
+                result.append(tuple(e.as_py() for e in slot))
+
         return tuple(result)
 
     @property
     def domain(self) -> Tuple[Tuple[object, object], ...]:
-        return self._get_and_cast_domain(self._handle.soma_domain_slot)
+        return self._cast_domainish(self._handle.domain())
 
     @property
     def maxdomain(self) -> Tuple[Tuple[object, object], ...]:
-        return self._get_and_cast_domain(self._handle.soma_maxdomain_slot)
+        return self._cast_domainish(self._handle.maxdomain())
 
     def non_empty_domain(self) -> Tuple[Tuple[object, object], ...]:
-        return self._get_and_cast_domain(self._handle.non_empty_domain_slot) or ()
+        return self._cast_domainish(self._handle.non_empty_domain())
 
     @property
     def attr_names(self) -> Tuple[str, ...]:
--- a/apis/r/R/SOMADataFrame.R
+++ b/apis/r/R/SOMADataFrame.R
@@ -339,6 +339,30 @@ SOMADataFrame <- R6::R6Class(
       class = 'notYetImplementedError'
     )),
 
+    #' @description Returns a named list of minimum/maximum pairs, one per index
+    #' column, currently storable on each index column of the dataframe. These
+    #' can be resized up to `maxdomain`.
+    #' (lifecycle: maturing)
+    #' @return Named list of minimum/maximum values.
+    domain = function() {
+      as.list(
+        arrow::as_record_batch(
+          arrow::as_arrow_table(
+            domain(self$uri, private$.soma_context))))
+    },
+
+    #' @description Returns a named list of minimum/maximum pairs, one per index
+    #' column, which are the limits up to which the dataframe can have its
+    #' domain resized.
+    #' (lifecycle: maturing)
+    #' @return Named list of minimum/maximum values.
+    maxdomain = function() {
+      as.list(
+        arrow::as_record_batch(
+          arrow::as_arrow_table(
+            maxdomain(self$uri, private$.soma_context))))
+    },
+
     #' @description Returns TRUE if the array has the upgraded resizeable domain
     #' feature from TileDB-SOMA 1.15: the array was created with this support,
     #' or it has had ``upgrade_domain`` applied to it.
--- a/apis/r/R/utils-arrow.R
+++ b/apis/r/R/utils-arrow.R
@@ -378,10 +378,10 @@ get_domain_and_extent_dataframe <- function(tbl_schema, ind_col_names,
         # expansion.
         if (ind_col_name == "soma_joinid") {
           # Must be non-negative
-          ind_max_dom <- arrow_type_unsigned_range(ind_col_type) - c(0,ind_ext)
+          ind_max_dom <- arrow_type_unsigned_range(ind_col_type) - c(0, ind_ext)
         } else {
           # Others can be negative
-          ind_max_dom <- arrow_type_range(ind_col_type) - c(0,ind_ext)
+          ind_max_dom <- arrow_type_range(ind_col_type) - c(0, ind_ext)
         }
 
         ind_cur_dom <- ind_max_dom
--- a/apis/r/R/utils-arrow.R
+++ b/apis/r/R/utils-arrow.R
@@ -370,13 +370,19 @@ get_domain_and_extent_dataframe <- function(tbl_schema, ind_col_names,
             ind_ext <- 64L
         }
 
-        # We need to do this because if we don't:
+        # We need to subtract off extent from the max because if we don't:
         #
         # Error: [TileDB::Dimension] Error: Tile extent check failed; domain max
         # expanded to multiple of tile extent exceeds max value representable by
         # domain type. Reduce domain max by 1 tile extent to allow for
         # expansion.
-        ind_max_dom <- arrow_type_unsigned_range(ind_col_type) - c(0,ind_ext)
+        if (ind_col_name == "soma_joinid") {
+          # Must be non-negative
+          ind_max_dom <- arrow_type_unsigned_range(ind_col_type) - c(0,ind_ext)
+        } else {
+          # Others can be negative
+          ind_max_dom <- arrow_type_range(ind_col_type) - c(0,ind_ext)
+        }
 
         ind_cur_dom <- ind_max_dom
         if (ind_col_type_name %in% c("string", "large_utf8", "utf8")) ind_ext <- NA
--- a/apis/r/tests/testthat/test-shape.R
+++ b/apis/r/tests/testthat/test-shape.R
@@ -85,16 +85,16 @@ test_that("SOMADataFrame shape", {
       expect_true(sjid_mxd[[2]] > bit64::as.integer64(10000000000))
     }
 
-    if ("foo" %in% index_column_names) {
-      int_dom <- dom[["foo"]]
-      int_mxd <- mxd[["foo"]]
+    if ("int_column" %in% index_column_names) {
+      int_dom <- dom[["int_column"]]
+      int_mxd <- mxd[["int_column"]]
       expect_true(int_dom[[1]] < -2000000000)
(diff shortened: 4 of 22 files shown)


apis/r/R/SOMADataFrame.R
```diff
@@ -339,6 +339,30 @@ SOMADataFrame <- R6::R6Class(
...
+    #' @description Returns a named list of minimum/maximum pairs, one per index
+    #' column, currently storable on each index column of the dataframe. These
+    #' can be resized up to `maxdomain`.
+    #' (lifecycle: maturing)
+    #' @return Named list of minimum/maximum values.
+    domain = function() {
+      as.list(
+        arrow::as_record_batch(
+          arrow::as_arrow_table(
+            domain(self$uri, private$.soma_context))))
```
mojaveazure, line 351: What's the R-object return type of `domain()` (and `maxdomain()` and `non_empty_domain()`)? Why do we take the results and go to an Arrow Table, then an Arrow RecordBatch, then a list? Do we need these conversions in the first place? I know that you can call `as.list()` on an Arrow Table directly, but do we need to go through a Table in the first place?

```R
> (tbl <- arrow::as_arrow_table(mtcars))
Table
32 rows x 11 columns
$mpg <double>
$cyl <double>
$disp <double>
$hp <double>
$drat <double>
$wt <double>
$qsec <double>
$vs <double>
$am <double>
$gear <double>
$carb <double>

See $metadata for additional Schema metadata
> identical(as.list(tbl), as.list(arrow::as_record_batch(tbl)))
[1] TRUE
```

apis/r/R/utils-arrow.R
```diff
@@ -370,13 +370,19 @@ get_domain_and_extent_dataframe <- function(tbl_schema, ind_col_names,
...
+        # We need to subtract off extent from the max because if we don't:
         #
         # Error: [TileDB::Dimension] Error: Tile extent check failed; domain max
         # expanded to multiple of tile extent exceeds max value representable by
         # domain type. Reduce domain max by 1 tile extent to allow for
         # expansion.
-        ind_max_dom <- arrow_type_unsigned_range(ind_col_type) - c(0,ind_ext)
+        if (ind_col_name == "soma_joinid") {
+          # Must be non-negative
+          ind_max_dom <- arrow_type_unsigned_range(ind_col_type) - c(0,ind_ext)
```
mojaveazure, line 381: Style fix
```suggestion
          ind_max_dom <- arrow_type_unsigned_range(ind_col_type) - c(0, ind_ext)
```

apis/r/R/utils-arrow.R
```diff
@@ -370,13 +370,19 @@ get_domain_and_extent_dataframe <- function(tbl_schema, ind_col_names,
...
+        } else {
+          # Others can be negative
+          ind_max_dom <- arrow_type_range(ind_col_type) - c(0,ind_ext)
```
mojaveazure, line 384: Style fix
```suggestion
          ind_max_dom <- arrow_type_range(ind_col_type) - c(0, ind_ext)
```

apis/r/tests/testthat/test-shape.R
```diff
@@ -4,17 +4,20 @@ test_that("SOMADataFrame shape", {
...
 
   index_column_name_choices = list(
-    "soma_joinid",
-    c("soma_joinid", "foo"),
-    c("soma_joinid", "baz"),
-    c("baz", "foo")
+    one = "soma_joinid",
+    two = c("soma_joinid", "foo"),
+    three = c("soma_joinid", "baz"),
+    four = c("baz", "foo")
```
mojaveazure, line 10: These test modifications can be simplified:
 - the names of the list aren't being used for anything essential
```suggestion
    "soma_joinid",
    c("soma_joinid", "foo"),
    c("soma_joinid", "baz"),
    c("baz", "foo")
```

apis/r/tests/testthat/test-shape.R
```diff
@@ -4,17 +4,20 @@ test_that("SOMADataFrame shape", {
...
   )
 
-  for (index_column_names in index_column_name_choices) {
+  for (case_name in names(index_column_name_choices)) {
+    index_column_names = index_column_name_choices[[case_name]]
```
mojaveazure, line 14: These test modifications can be simplified:
 - we're iterating through a list and performing the same sets of checks, we can use `seq_along()` to iterate along the list for us
```suggestion
  for (i in seq_along(index_column_name_choices)) {
    index_column_names <- index_column_name_choices[[i]]
```

apis/r/tests/testthat/test-shape.R
```diff
@@ -56,6 +59,60 @@ test_that("SOMADataFrame shape", {
       expect_true(is.na(sjid_maxshape))
     }
 
+    dom = sdf$domain()
+    mxd = sdf$maxdomain()
```
mojaveazure, line 63: Style fixes
```suggestion
    dom <- sdf$domain()
    mxd <- sdf$maxdomain()
```

apis/r/tests/testthat/test-shape.R
```diff
@@ -56,6 +59,60 @@ test_that("SOMADataFrame shape", {
...
+      expect_equal(names(mxd), c("soma_joinid", "foo"))
+
+    } else if (case_name == "three") {
+      expect_equal(names(dom), c("soma_joinid", "baz"))
+      expect_equal(names(mxd), c("soma_joinid", "baz"))
+
+    } else if (case_name == "four") {
+      expect_equal(names(dom), c("baz", "foo"))
+      expect_equal(names(mxd), c("baz", "foo"))
+    }
```
mojaveazure, line 81: These test modifications can be simplified:
 - we're checking that the names of `dom` and `mxd` are the same as `index_column_names` for each case; rather than hard-coding the cases, we can use the already-defined variable to perform the check for us. This also allows us to add new cases should we desire
```suggestion
    # First check names
    expect_equal(names(dom), index_column_names)
    expect_equal(names(mxd), index_column_names)
```

apis/r/tests/testthat/test-shape.R
```diff
@@ -56,6 +59,60 @@ test_that("SOMADataFrame shape", {
...
+
+    # Then check all slots are pairs
+    for (name in names(dom)) {
+      expect_equal(length(dom[[name]]), 2)
+    }
+    for (name in names(mxd)) {
+      expect_equal(length(mxd[[name]]), 2)
+    }
```
mojaveazure, line 89: These test modifications can be simplified:
 - we've already determined that the names of `dom` and `mxd` are identical; we can iterate through them both simultaneously
 - we can use `expec_length()` instead of `expect_equal()` for more informative test failure messages
```suggestion
    for (name in names(dom)) {
      expect_length(dom[[name]], 2L)
      expect_length(mxd[[name]], 2L)
    }
```

mojaveazure (changes requested):

apis/r/R/SOMADataFrame.R
johnkerl: I found that these steps were all necessary. Errors otherwise. I can share details if you like.

johnkerl: The R object returned is a named list of two-element vectors.

mojaveazure: Please do share details; it doesn't make sense for a list-return to throw errors or for `as.list(tbl)` to fail while `as.list(arrow::as_record_batch(tbl))` works

johnkerl: @mojaveazure 

```
# This is a local copy of pbmc3k
> sdf <- SOMAOpen('/var/p/obs')

> sdf$schema()
Schema
soma_joinid: int64 not null
obs_id: string not null
n_genes: int64 not null
percent_mito: float not null
n_counts: float not null
louvain: dictionary<values=string, indices=int8> not null

> ctx <- tiledbsoma:::soma_context()

> a <- domain(sdf$uri, ctx)

> a
<nanoarrow_array struct[0]>
 $ length    : int 0
 $ null_count: int 0
 $ offset    : int 0
 $ buffers   :List of 1
  ..$ :<nanoarrow_buffer validity<bool>[0][0 b]> ``
 $ children  :List of 1
  ..$ soma_joinid:<nanoarrow_array int64[2]>
  .. ..$ length    : int 2
  .. ..$ null_count: int 0
  .. ..$ offset    : int 0
  .. ..$ buffers   :List of 2
  .. .. ..$ :<nanoarrow_buffer validity<bool>[0][0 b]> ``
  .. .. ..$ :<nanoarrow_buffer data<int64>[2][16 b]> `0.00e+00 2.15e+09`
  .. ..$ dictionary: NULL
  .. ..$ children  : list()
 $ dictionary: NULL

> as.list(a)
Error in as.vector.nanoarrow_array(x, "list") :
  identical(mode, "any") is not TRUE

> as.list(arrow::as_record_batch(a))
Error in `[<-`:
! Assigned data `map(.subset(x, unname), vectbl_set_names, NULL)` must be compatible with existing data.
✖ Existing data has 0 rows.
✖ Assigned data has 2 rows.
ℹ Only vectors of size 1 are recycled.
Caused by error in `vectbl_recycle_rhs_rows()`:
! Can't recycle input of size 2 to size 0.
Run `rlang::last_trace()` to see where the error occurred.

> as.list(arrow::as_arrow_table(a))
Error in `[<-`:
! Assigned data `map(.subset(x, unname), vectbl_set_names, NULL)` must be compatible with existing data.
✖ Existing data has 0 rows.
✖ Assigned data has 2 rows.
ℹ Only vectors of size 1 are recycled.
Caused by error in `vectbl_recycle_rhs_rows()`:
! Can't recycle input of size 2 to size 0.
Run `rlang::last_trace()` to see where the error occurred.

> as.list(arrow::as_record_batch(arrow::as_arrow_table(a)))
$soma_joinid
[1]          0 2147483646
```

mojaveazure: Ah, okay. It seems like `domain()` et al don't actually read in the domain but leave the domain information in a nanoarrow buffer. I don't know if going from nanoarrow array w/ buffer -> Arrow Table -> Arrow RecordBatch -> list is the best way forward, but I can't find a way around it in R (especially since it seems like the RecordBatch is needed to get the data out of the buffer)

mojaveazure: My question now is do we want these returns to be a list or a vector? Another way of putting that is will these functions always return two values (min, max) or potentially multiple sets of min/max pairs?

johnkerl: @mojaveazure that's partly a UX question

* Each slot's _value_ will always be a 2-element lo/hi pair -- one per dim -- no changing that
* What's a UX questiion is the _keys_ -- what's better, a named list (which I personally like), or, an unnamed list, or, an outer vector of length N where N is the number of dims

I personally like the above three choices in the order I listed them, b/c (a) there's more info there; (b) people can remove info by `unname(dom)` or `as.vector(x)` but it's harder to put the info back in if we just give them a length-N vector of length-2 vectors

But that's just my general feel as a polyglottal language user; I defer to you

I'd also point out that on the Python side it's a length-n tuple of length-2 tuples ... so for strict partity we might go that way ...

johnkerl: Maybe have an optional `as.something` arg to the domainish accessors where the default is the compact way but people can get the named-list info if they ask for it? Just a thought. (And it doesn't add much complexity.)

mojaveazure: Vectors are strictly flat and cannot be nested. Because it seems like the base unit of return is the dim and there's effectively one return (a min/max pair) per dim, a list is the appropriate return. I agree that named is better than unnamed as it's easer to unname than to re-name

mojaveazure: (I apologize for the back-and-forth on this; I was not part of the discussions for these new methods, so I'm learning them for the first time here)

mojaveazure (approved):
LGTM on the R-side; will defer to other for Python. Before merging, please bump the develop version and update the changelog

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -363,41 +362,36 @@ def meta(self) -> "MetadataWrapper":
     def ndim(self) -> int:
         return len(self._handle.dimension_names)
 
-    def _get_and_cast_domain(
-        self, domain_slot_getter: Callable[[str, DTypeLike], Tuple[object, object]]
+    def _cast_domainish(
+        self, domainish: List[Any]
```
nguyenv, line 366: Is the `domainish` type not a `Tuple[Tuple[object, object], ...]`?

nguyenv, line 366: Would the type for `domainish` be something like `List[np.array[object, object], ...]`?

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -363,41 +362,36 @@ def meta(self) -> "MetadataWrapper":
...
     ) -> Tuple[Tuple[object, object], ...]:
         result = []
-        for name in self._handle.dimension_names:
-            dtype = self._handle.schema.field(name).type
-            if pa.types.is_timestamp(dtype):
-                np_dtype = np.dtype(dtype.to_pandas_dtype())
-                dom = domain_slot_getter(name, np_dtype)
+        for i, slot in enumerate(domainish):
```
nguyenv, line 369: I might have missed it, but I don't think you actually use `i` in this loop.

```suggestion
        for slot in domainish:
```

nguyenv (approved):

apis/python/src/tiledbsoma/_tdb_handles.py
johnkerl:
```
    index_column_names <- index_column_name_choices[[i]]
```

johnkerl: and on #3032, more there

apis/python/src/tiledbsoma/_tdb_handles.py
johnkerl: yeah something like that

it gave me fits & I did `Any` to unblock & forgot to come back & fine-tune ... let me try that ...

thank you! :)

johnkerl: @nguyenv I merged without doing this (I resolved the wrong comment) -- will address in follow-up

johnkerl: https://github.com/single-cell-data/TileDB-SOMA/issues/3081

johnkerl: @nguyenv I may leave it as `Any` but with a comment. Here's what it looks like in one example:

```
>>> sdf._handle._handle.domain()
[<pyarrow.lib.LargeStringArray object at 0x100d8c6a0>
[
  "apple",
  "zebra"
], <pyarrow.lib.Int32Array object at 0x1612e5e40>
[
  -1000000,
  1000000
]]
>>> sdf._handle._handle.domain()[0]
<pyarrow.lib.LargeStringArray object at 0x1612e5ea0>
[
  "apple",
  "zebra"
]
>>> sdf._handle._handle.domain()[1]
<pyarrow.lib.Int32Array object at 0x1612e5e40>
[
  -1000000,
  1000000
]
```

So it's a list of `pyarrow` arrays ...

PR #3030, opened 2024-09-22
[c++] Fix memory-management issues in new domainish helpers

johnkerl:
**Issue and/or context:** Found while debugging memory-management issues on #3027 for #2407 / [[sc-51048]](https://app.shortcut.com/tiledb-inc/story/51048).

**Changes:**

The new `libtiledbsoma` helpers used by #3027 were incorrectly pairing `new` and `free`, as immediately revealed by my running `valgrind build/libtiledbsoma/test/unit_soma` while debugging some issues on #3027. There are code comments saying to use `new` to match the releaser callbacks, but these are incorrect -- the releaser callbacks are using `free`.

**Notes for Reviewer:**


nguyenv (approved):
Thank you for make all these consistently `malloc`/`free`.

PR #3034, opened 2024-09-23
[python] add `--maxfail=50` to GHA `pytest` invocations

ryan-williams:
Mitigate excessive logspam on PRs where many tests fail.

Multiple GHAs recently have failed with a huge amount of logspam, such that my browser tab never finishes loading the page, or any of the logs.

[Here's](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/10910221889/job/30280188276) an example, from #3001; downloading its logs:
```bash
j=30280188276
gh run view -j $j --log > $j.txt
```

Shows 472k lines / 77MiB. It looks like 482 test failures, some of which output ≈700 lines each.

The PRs where I've seen this recently have generally involved changing Python or Typeguard versions; Typeguard issues can cause large numbers of test cases to fail very verbosely.

[This gist](https://gist.github.com/ryan-williams/2206c5b05050ac9d786ad5b20bd5ddf8) has the first 10,000 lines from the example above:
```bash
gh run view -j $j --log | head -n 10000 | gist -opf $j.txt
```

### DIFF
--- a/.github/workflows/python-ci-single.yml
+++ b/.github/workflows/python-ci-single.yml
@@ -134,14 +134,14 @@ jobs:
       # Setting PYTHONPATH ensures the tests load the in-tree source code under apis/python/src
       # instead of the copy we `pip install`ed to site-packages above. That's needed for the code
       # coverage analysis to work.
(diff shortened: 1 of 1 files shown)


johnkerl (approved):
🔥

PR #3063, opened 2024-09-24
Fix URI creation in spatial ingest

jp-dark:
Previously, the code was using hard-coded paths that assumed a posix-like structure. Update to use helper function for joining paths.


aaronwolen (approved):
Thanks!

PR #3032, opened 2024-09-22
[r] Implement missing `domain` argument to `DataFrame` `create`

johnkerl:
**Issue and/or context:** As tracked on issue #2407 / [[sc-51048]](https://app.shortcut.com/tiledb-inc/story/51048). Fixes #2967 / [[sc-55657]](https://app.shortcut.com/tiledb-inc/story/55657/use-domain-in-somadataframecreate).

**Changes:**

As a legacy R/Python-parity issue, the `domain` argument to `SOMADataFrame`'s `create` -- from the [SOMA spec](https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md) -- was never implemented in R. Here we remedy that.

**Notes for Reviewer:**

This PR is a work in progress. It is not ready for review.

Unit-test cases are not yet implemented.

PR #3069, opened 2024-09-25
[c++] Expose custom `DataFrame` domain for `libtiledbsoma` unit-test cases

johnkerl:

**Changes:**

While working on R-level #3032 I realized to my unpleasant surprise that I'd neither tested nor implemented custom domain-setting for `SOMADataFrame` at the C++ level. This PR fixes that oversight.

Along the way I split out a necessary helper function in `ArrowAdapter`, and changed some method names to make them clearer. I'm happy to do this as split-out commits or split-out PRs if that makes it easier to review.

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/utils/arrow_adapter.h
+++ b/libtiledbsoma/src/utils/arrow_adapter.h
@@ -439,8 +439,8 @@ class ArrowAdapter {
 
         if (std::is_same_v<T, std::string>) {
             throw std::runtime_error(
-                "SOMAArray::_core_domain_slot: template-specialization "
-                "failure.");
+                "SOMAArray::get_table_non_string_column_by_index: "
+                "template-specialization failure.");
         }
 
         ArrowArray* child_array = _get_and_check_column(
@@ -542,7 +542,7 @@ class ArrowAdapter {
         const ArrowArray* arrow_array, const ArrowSchema* arrow_schema) {
         if (arrow_array->n_children != 0 || arrow_schema->n_children != 0) {
             throw std::runtime_error(
-                "ArrowAdapter::get_array_non_string_column: expected leaf "
+                "ArrowAdapter::get_array_string_column: expected leaf "
                 "node");
         }
         if (arrow_array->n_buffers != 3) {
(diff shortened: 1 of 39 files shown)


libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -426,84 +426,186 @@ class ArrowAdapter {
...
-                "SOMAArray::_core_domain_slot: template-specialization "
+                "SOMAArray::get_table_non_string_column_by_index: "
+                "template-specialization failure.");
+        }
+
+        ArrowArray* child_array = _get_and_check_column(
+            arrow_table, column_index, 2);
+
+        return get_array_non_string_column<T>(child_array);
+    }
```
nguyenv, line 450: Could overload `get_table_non_string_column` but also fine as is.

```
template <typename T>
static std::vector<T> get_table_non_string_column(
    const ArrowTable& arrow_table, const std::string& column_name);

template <typename T>
static std::vector<T> get_table_non_string_column(
    const ArrowTable& arrow_table, int64_t column_index);
```

nguyenv (approved):

libtiledbsoma/src/utils/arrow_adapter.h
johnkerl: > Could overload get_table_non_string_column but also fine as is.

I should leave a comment

Originally I used template specialization
```
template <typename T>
foo(T a, T b) 

template <>
foo(std::string a, std::string b)
```

and I kept getting spots where the non-string version was getting called even in string contexts

I tried to be clever & I persisted, but, ultimately I decided it wasn't worth it. I've been doing C/C++ a long time, I'm no slouch, and if it was fragile for me, it's not worth leaving in a fragile state.

So I went with the `_string` suffix, and the
```
if (std::is_same_v<T, std::string>) {
    throw std::runtime_error(...)
}
```

johnkerl: @nguyenv https://github.com/single-cell-data/TileDB-SOMA/pull/3069/commits/e05b1ea60fbacd6c6860c340cd927f71eaea7b19

PR #3073, opened 2024-09-26
[python] Remove more Python 3.8 references

ryan-williams:
**Issue and/or context:** https://github.com/single-cell-data/TileDB-SOMA/issues/3072

Daily "TileDB-SOMA python sdist & wheels" workflow was still attempting to use Python 3.8, and failed.


johnkerl (approved):
💥

Issue #3055, opened 2024-09-24
[r] Use `libtiledbsoma` for all `tiledb::schema` callsites


johnkerl:
Currently done using the tiledb-r path which we're removing as a dependency.

#2406 

[[sc-55682]](https://app.shortcut.com/tiledb-inc/story/55682/r-use-libtiledbsoma-for-all-tiledb-schema-callsites)

johnkerl:
Per this https://github.com/single-cell-data/TileDB-SOMA/pull/3079#issuecomment-2384690485 we need schema evolution also ported to libtiledbsoma

johnkerl:
#3057 is merged

Issue #2406, opened 2024-04-08
[r] Use new C++ bindings in R API (parent task)


johnkerl:
* This is a sibling of #1632
* It is assigned to no one and it will remain assigned to no one
* Its sole job is to link-track, at the GitHub Issues level, smaller issues which will be split out as individual GitHub Issues

PR #3079, opened 2024-09-27
[r] Push `schema` accessor down to `libtiledbsoma`

johnkerl:
**Issue and/or context:** #2406 #3055 [[sc-55682]](https://app.shortcut.com/tiledb-inc/story/55682/r-use-libtiledbsoma-for-all-tiledb-schema-callsites)

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/r/R/SOMADataFrame.R
+++ b/apis/r/R/SOMADataFrame.R
@@ -115,6 +115,8 @@ SOMADataFrame <- R6::R6Class(
     write = function(values) {
       private$check_open_for_write()
 
+      ### 3079:04
+
       # Prevent downcasting of int64 to int32 when materializing a column
       op <- options(arrow.int64_downcast = FALSE)
       on.exit(options(op), add = TRUE, after = FALSE)
@@ -295,6 +297,7 @@ SOMADataFrame <- R6::R6Class(
       tiledb_create_options <- TileDBCreateOptions$new(self$platform_config)
 
       # Check compatibility of new/old data types in common columns
+      ### debug 3079:02
       check_arrow_schema_data_types(
         old_schema[common_cols],
         new_schema[common_cols]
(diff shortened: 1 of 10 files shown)


johnkerl:
Update as of 2024-09-30:

* Fails unit test [here](https://github.com/single-cell-data/TileDB-SOMA/blob/40e21784e953d808dbf1b78b2f9ede5ad185be0e/apis/r/tests/testthat/test-SOMADataFrame.R#L641-L642)
  * Old/new schema comparison fails in `check_arrow_schema_data_types` in `R/utils-arrow.R`: value-type binary vs value-type string
* I've traced this down to this [one line](https://github.com/single-cell-data/TileDB-SOMA/blob/40e21784e953d808dbf1b78b2f9ede5ad185be0e/apis/r/R/SOMADataFrame.R#L339)
  * TileDB-R's schema evolution (not yet ported to libtiledbsoma) writes value-type binary
  * Later on in the unit-test spot mentioned above that fails to check against value-type readback using the schema getter on this PR (which has been ported to libtiledbsoma)

apis/r/R/SOMADataFrame.R
```diff
@@ -245,6 +245,10 @@ SOMADataFrame <- R6::R6Class(
           is.data.frame(values) || is_arrow_table(values) || is_arrow_record_batch(values)
       )
 
+      # Leave state unmodified
+      # TODO: this issue will automatically go away on https://github.com/single-cell-data/TileDB-SOMA/issues/3059
+      on.exit(self$reopen(mode = "WRITE"))
```
mojaveazure, line 250: If `mode` ever expands beyond `"READ"` and `"WRITE"`, or changes to `"r"`/`"w"` to align with Python, change to capture the array mode and reopen in said mode
```suggestion
      omode <- self$mode()
      on.exit(self$reopen(mode = omode))
```

mojaveazure (approved):
LGTM; left one minor comment to future-proof the code. Please remember to bump the develop version, update the changelog, and 🚢

apis/r/R/SOMADataFrame.R
johnkerl: Awesome, thanks @mojaveazure !! :)

Issue #3051, opened 2024-09-24
[r] Port the query-condition logic from TileDB-R to TileDB-SOMA-R


johnkerl:
Currently tiledbsoma-r uses query-condition logic from tiledb-r. We need to duplicate this within tiledbsoma-r. We've already done so for tiledbsoma-py and tiledb-py.

#2406 

[[sc-55672]](https://app.shortcut.com/tiledb-inc/story/55672/r-duplicate-query-condition-logic)

Do not close until this is complete: https://github.com/single-cell-data/TileDB-SOMA/pull/3162#discussion_r1801606040

PR #3083, opened 2024-09-27
[r] Port `QueryCondition` from `tiledb-r` to `tiledbsoma-r` [WIP]

johnkerl:
**Issue and/or context:** #2406 #3051 [[sc-55682]](https://app.shortcut.com/tiledb-inc/story/55682/r-use-libtiledbsoma-for-all-tiledb-schema-callsites)

**Changes:**

* A necessary copy-over from tiledb-r
  * We are removing that dependency
  * We can't expect pushdown of R-parsing to the core C++ package
* Namespacing changes: e.g. explicit `Rcpp::as` rather than `as`

**Notes for Reviewer:**

This is a WIP.


johnkerl:
Replaced by #3162

Issue #3045, opened 2024-09-23
[spatial] Add spatial format version and parser


johnkerl:
[[sc-54178]](https://app.shortcut.com/tiledb-inc/story/54178/add-spatial-format-version-and-parser)

PR #3086, opened 2024-09-27
[python/r/c++] Update allowed soma types and the metadata encoding version

jp-dark:
**Changes:** 
* Update the SOMA metadata encoding to 1.1.0. 
* Update `soma_type` checks to describe spatial types as "not implemented" instead of "invalid".

**Issue:** #3045


jp-dark:
[sc-54178]

PR #3088, opened 2024-09-28
[python] Use same default max domain between Python and R

johnkerl:

**Changes:**

In Python we were trying to use default max shapes there were 31-bit friendly for R compatibility. However, in R we're already setting 63-bit maxdomain:
https://github.com/single-cell-data/TileDB-SOMA/blob/1.14.1/apis/r/R/utils-arrow.R#L127

**Notes for Reviewer:**


mojaveazure (approved):

johnkerl:
Thanks @nguyenv and @mojaveazure ! :)

PR #3090, opened 2024-09-28
[c++] Simplify internal API for dataframe-resizer

johnkerl:

**Changes:**

While the full-generality `upgrade_domain` is tasked out on #2407, the foremost use-case for append mode, and `tests/testthat/test-write-soma-resume.R` for PR #3089, will need a simpller keystroke-saver which is already implemented in C++.

What this existing method does in C++ is resize the `soma_joinid` dim, if it's a dim, and leave the other dims as-is; if `soma_joinid` is not a dim, leave it as-is. This exists in C++, it's unit-tested in C++, and it's ready (and it's time) to be plumbed up to Python and R as well.

I realize now that having C++ `maybe_resize_joinid` take a length-1 vector, and assert that it's length-1, is silly -- it should simply take a single `int64_t`. This PR does just that. It also removes the confusing prefix `maybe_`.

**Notes for Reviewer:**

### DIFF
--- a/.github/workflows/python-ci-packaging.yml
+++ b/.github/workflows/python-ci-packaging.yml
@@ -76,7 +76,7 @@ jobs:
         run: |
           mkdir -p external
           # Please do not edit manually -- let scripts/update-tiledb-version.py update this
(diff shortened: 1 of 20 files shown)


johnkerl:
Thanks @nguyenv !

PR #3095, opened 2024-09-30
[c++] Add can-resize helpers in prep for experiment-level resize

johnkerl:

**Changes:**

For experiment-level resize (e.g. append mode) it will be crucial that we implement a mechanism to detect whether all arrays in the experiment are resizable up to the new `(nobs, nvar)` before attempting any resizes. In particular, it would be bad to succeed in resizing some member arrays and fail on one or more.

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/soma/soma_array.h
+++ b/libtiledbsoma/src/soma/soma_array.h
@@ -291,6 +291,13 @@ class SOMAArray : public SOMAObject {
     std::vector<std::string> dimension_names() const;
 
     /**
+     * @brief Sees if the array has a dimension of the given name.
+     *
+     * @return bool
+     */
+    bool has_dimension_name(const std::string& name) const;
+
+    /**
      * @brief Set the dimension slice using one point
      *
      * @note Partitioning is not supported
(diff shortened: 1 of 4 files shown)


libtiledbsoma/src/soma/soma_array.h
```diff
@@ -1031,6 +1038,55 @@ class SOMAArray : public SOMAObject {
...
 
+    /**
+     * This wires up to Python/R to tell a user if they can call resize() on an
+     * array without error. For single arrays, they could just call resize() and
+     * take their chances -- but for experiment-level resize (e.g. append mode)
+     * it's crucial that we provide a can-we-do-them-all pass through all arrays
+     * in the experiment before attempting any of them.
+     *
+     * On failure, returns false and an error string suitable for showing
+     * to the user; on success, returns true and the empty string.
```
nguyenv, line 1049: Just mulling over this. I feel like returning an error code is better (maybe using enums) rather than returning a pair,  if we don't want to throw exceptions.

johnkerl: These are error strings directly to the user without needing extra decode separately in Python/R.

Issue #3057, opened 2024-09-24
[r] Use libtiledbsoma for schema evolution


johnkerl:
Currently done using the tiledb-r path which we're removing as a dependency.

#2406 

[[sc-55679]](https://app.shortcut.com/tiledb-inc/story/55679/r-use-libtiledbsoma-for-schema-evolution)

johnkerl:
It looks like I can move this to `libtiledbsoma` pretty much verbatim
https://github.com/single-cell-data/TileDB-SOMA/blob/1.14.2/apis/python/src/tiledbsoma/pytiledbsoma.cc#L139-L204
and then I can re-use that for R.

CC @nguyenv in case I'm missing something

PR #3100, opened 2024-10-01
[c++/python/r] Use `libtiledbsoma` for R schema evolution

johnkerl:
**Issue and/or context:** This is an item on #2406: in particular #3057 / [[sc-55679]](https://app.shortcut.com/tiledb-inc/story/55679/r-use-libtiledbsoma-for-schema-evolution)/

**Changes:**

* Move the `pybind11` `update_dataframe` entirely to `libtiledbsoma` (it's ready to go, no mods needed)
* Call that from R, removing another TileDB-R callsite

**Notes for Reviewer:**

See also #3115.


mojaveazure (approved):
Left some suggestions where we can be a bit more performant; I'd strongly recommend using them, but if you don't want then ignore and 🚢

PR #3111, opened 2024-10-01
[c++/python] Check `arrow_is_string_type` in `_update_dataframe`

nguyenv:
- Rename `ArrowAdapter::_isvar` to `ArrowAdapter::arrow_is_string_type` and expose in public API
- In pybind11 `SOMAArray._update_dataframe`, check if the Arrow column's type is a string type and set the `Attr`'s cell value to `TILEDB_VAR_NUM` if so


johnkerl (approved):
🚢 
Thank you @nguyenv ! :)

johnkerl:
Note: found on R-side debug on #3057

PR #3113, opened 2024-10-02
[c++/python] Push `update_dataframe_schema` down to `libtiledbsoma`

johnkerl:
**Issue and/or context:** For #3057 [[sc-55679]](https://app.shortcut.com/tiledb-inc/story/55679/r-use-libtiledbsoma-for-schema-evolution)

**Changes:**

On the R side we'll need to use `libtiledbsoma` rather than TileDB-R for schema evolution. Meanwhile the logic currently in our `pybind11` layer can go directly into `libtiledbsoma` where our `Rcpp` layer can call it.

**Notes for Reviewer:**

No functional change: only code motion. No new unit tests: all existing tests should pass.

### DIFF
--- a/apis/python/src/tiledbsoma/pytiledbsoma.cc
+++ b/apis/python/src/tiledbsoma/pytiledbsoma.cc
@@ -138,75 +138,14 @@ PYBIND11_MODULE(pytiledbsoma, m) {
             "consolidate_and_vacuum", &PlatformConfig::consolidate_and_vacuum);
 
     m.def(
-        "_update_dataframe",
+        "_update_dataframe_schema",
         [](std::string uri,
            std::shared_ptr<SOMAContext> ctx,
            std::vector<std::string> drop_attrs,
            std::map<std::string, std::string> add_attrs,
            std::map<std::string, std::pair<std::string, bool>> add_enmrs) {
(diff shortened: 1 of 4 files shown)


apis/python/src/tiledbsoma/pytiledbsoma.cc
```diff
@@ -138,75 +138,14 @@ PYBIND11_MODULE(pytiledbsoma, m) {
...
-                        ordered));
-                    AttributeExperimental::set_enumeration_name(
-                        *ctx->tiledb_ctx(), attr, attr_name);
-                }
-
-                se.add_attribute(attr);
-            }
-            se.array_evolve(uri);
+            SOMADataFrame::update_dataframe_schema(
+                uri, ctx, drop_attrs, add_attrs, add_enmrs);
```
nguyenv, line 148: Are there issue binding directly using `m.def("_update_dataframe_schema", &SOMADataFrame::update_dataframe_schema)`?

nguyenv (approved):

apis/python/src/tiledbsoma/pytiledbsoma.cc
johnkerl: Good call! :) 🤔 👀

PR #3126, opened 2024-10-03
[python] `tiledbsoma.io.to_anndata` performance work

bkmartinjr:
**Issue and/or context:**

Performance of `soma.io.to_anndata` is important for Experiment ETL to AnnData/H5AD.

**Changes:**

* Add parallelism in `soma.io.to_anndata` to take advantage of concurrent I/O and compute operations.
* Release GIL in latent `open` IO operations
* Minor code refactoring in expectation of a future integration of a new COO->CSR converter

**Notes for Reviewer:**

There are no changes to the logic. This PR simply parallelizes IO operations, and increases concurrency by releasing the GIL in costly C++ 'open' operations (which pre-fetch metadata, etc). Changes will slightly increase peak memory use in some cases.


johnkerl (approved):

apis/python/src/tiledbsoma/soma_object.cc
```diff
@@ -58,8 +58,11 @@ void load_soma_object(py::module& m) {
...
                std::optional<std::string> clib_type) -> py::object {
                 try {
-                    auto soma_obj = SOMAObject::open(
-                        uri, mode, context, timestamp, clib_type);
+                    auto soma_obj = ([&]() {
+                        py::gil_scoped_release release;
+                        return SOMAObject::open(
+                            uri, mode, context, timestamp, clib_type);
+                    })();
                     auto soma_obj_type = soma_obj->type();
```
nguyenv, line 66: Was wondering if this is the same as doing

```
py::gil_scoped_release release;
auto soma_obj = SOMAObject::open(
                        uri, mode, context, timestamp, clib_type);
py::gil_scoped_acquire acquire;
```

apis/python/src/tiledbsoma/io/outgest.py
```diff
@@ -472,6 +522,10 @@ def _extract_uns(
     for key, value in collection.metadata.items():
         if level == 0 and uns_keys is not None and key not in uns_keys:
             continue
+        if key in extracted:
+            # This should not be possible, as it implies we wrote the same key to the collection and metadata.
+            # Just skip if it happens, as a precaution.
+            continue
```
nguyenv, line 528: Should this error out or give some sort of warning?

apis/python/src/tiledbsoma/io/outgest.py
```diff
@@ -300,44 +311,73 @@ def to_anndata(
...
+    obsm = _resolve_futures(obsm)
+    varm = _resolve_futures(varm)
+    obsp = _resolve_futures(obsp)
+    varp = _resolve_futures(varp)
+    anndata_X = anndata_X_future.result() if anndata_X_future is not None else None
+    uns: UnsDict = (
+        _resolve_futures(uns_future.result(), deep=True)
+        if uns_future is not None
+        else {}
+    )
```
nguyenv, line 379:
```suggestion
    uns: UnsDict = _resolve_futures(uns_future.result(), deep=True) or {}
```

nguyenv: But it looks like `_resolve_futures` only returns a `Dict` and not `Optional[Dict]`, so could be further simplified to `uns: UnsDict = _resolve_futures(uns_future.result(), deep=True)`?

apis/python/src/tiledbsoma/soma_object.cc
bkmartinjr: I'm not a pybind expert, but I believe it is the same (with a bit more overhead due to the need to destruct both objects).

I did the lambda because I wanted a protected scope, to ensure we didn't accidentally add any code that messed with python objects (I believe this was the bug in the previous PR).

So, TL;DR - it is mostly style, if I understand things correctly.

apis/python/src/tiledbsoma/io/outgest.py
bkmartinjr: metaphysical question as it implies it is a bug - I'm not sure it is as we don't have clear semantics defined for uns  when a user modifies it to contain duplicate keys (e.g., if a user ingests an H5AD, the manually modifies the `uns` data using the SOMA API, is that a bug)?

Previously the code treated the contents of `metadata` as having precedence over the Collection contents.  This code simply reverses that precedence (which in retrospect, is probably a backward-compat breaking change).

@johnkerl - do you have an opinion which should have precedence in the case that both `uns.metadata["key"]` and `uns["key"]` exist?  I think the options are:
* generate error 
* metadata wins
* collection wins
Previous behavior was "metadata wins".

it can definitely happen with post-ingest modification by a user.

bkmartinjr: BTW, I'm going to push a commit that reverts to the historical behavior (i.e., doesn't break backward compat).  But the open question remains - what we _want_ to happen.

johnkerl: @bkmartinjr I think 'keep same' is best -- whatever we may or may not have documented, I think:

* switching the order will break no one and no one will notice; or
* for someone for whom the order matters in ways they might not even realize in their data pipelines, something weird may start happening

bkmartinjr: Concur - just pushed the mod.

PR #3128, opened 2024-10-04
[python] Fix `MultiscaleImage.read_spatial_region` full region read

jp-dark:
The `read_spatial_region` method should be able to accept `None` for a region. When this is provided, the whole array is returned.


johnkerl (approved):

aaronwolen (approved):
Looks good!

PR #2964, opened 2024-09-06
[python] New-shape testing for `tiledbsoma.io` [WIP]

johnkerl:
**Issue and/or context:** As tracked on issue #2407 / [[sc-51048]](https://app.shortcut.com/tiledb-inc/story/51048). Atop #2950.

Note that the intended Python and R API changes are all agreed on and finalized as described in #2407.

**Changes:**

Tracks the things needing changed for `tiledbsoma.io` wherein explicit resize will play a key role.

**Notes for Reviewer:**

This PR is a work in progress. It is not ready for review.

PR #3130, opened 2024-10-04
[c++] Propagate Python/R function names to C++ for upgrade/resize methods

johnkerl:

**Changes:**

* This is underpinning for #2964
* As @nguyenv pointed out, it might be nice for the C++ layer to just return enums only ...
* ... but as I pointed out
  * informative messages really require specific context such as the user-provided shape and the limit it ran into
  * I don't want to write those all twice in Python and R
  * So we are making user-presentable strings in the C++ code
* That said, we should not be hard-coding the function names from the caller -- given that C++ is creating user-facing error strings _on behalf of_ Python and R, the C++ code should enable and empower the Python/R code to pass in their function names (whatever those are)

**Notes for Reviewer:**

Despite the line-count this is a trivial PR. I'm splitting out this PR (as well as recents #3125 #3127 #3130) in order to clear the picture for #2964 which is the contentful goal of my recent work this week.

### DIFF
--- a/apis/python/src/tiledbsoma/soma_dataframe.cc
+++ b/apis/python/src/tiledbsoma/soma_dataframe.cc
@@ -161,7 +161,8 @@ void load_soma_dataframe(py::module& m) {
             "resize_soma_joinid_shape",
             [](SOMADataFrame& sdf, int64_t newshape) {
                 try {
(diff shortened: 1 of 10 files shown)


nguyenv (approved):

johnkerl:
Thanks @nguyenv !

PR #3136, opened 2024-10-04
[c++,python] Fix default reads on unwritten dense arrays

nguyenv:
**Issue and/or context:**

Found issue while working on spatial's `MultiscaleImage` and erroring out when applying `read_spatial_region` to a level that had not been written to yet.

**Changes:**

* Add new `non_empty_domain_slot_opt` alongside `non_empty_domain_slot` that returns `std::nullopt` if it is empty rather than (0, 0)
* Fix logic in C++ `ManagedQuery::setup_read` to check if the array has been written to yet. If it hasn't, set the subarray to use soma_dim_0's full domain. Otherwise, use the non-empty domain as before
* Also fix logic in Python `DenseNDArray.read` to use the new `non_empty_domain_slot_opt` and correctly set the `data_shape` to the full domain if the array has not been written to yet
* Add unit test in pytest

### DIFF
--- a/libtiledbsoma/src/soma/soma_array.h
+++ b/libtiledbsoma/src/soma/soma_array.h
@@ -722,12 +722,48 @@ class SOMAArray : public SOMAObject {
 
     /**
      * Retrieves the non-empty domain from the array. This is the union of the
-     * non-empty domains of the array fragments.
+     * non-empty domains of the array fragments. Returns (0, 0) for empty
+     * domains.
      */
     template <typename T>
     std::pair<T, T> non_empty_domain_slot(const std::string& name) const {
         try {
             return arr_->non_empty_domain<T>(name);
+        } catch (const std::exception& e) {
+            throw TileDBSOMAError(e.what());
+        }
+    }
+
+    /**
+     * Retrieves the non-empty domain from the array. This is the union of the
+     * non-empty domains of the array fragments. Return std::nullopt for empty
+     * domains.
+     */
+    template <typename T>
+    std::optional<std::pair<T, T>> non_empty_domain_slot_opt(
+        const std::string& name) const {
+        try {
+            int32_t is_empty;
+            T ned[2];
+
+            // TODO currently we need to use the TileDB C API in order to check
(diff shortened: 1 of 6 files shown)


libtiledbsoma/src/soma/soma_array.h
```diff
@@ -733,6 +734,41 @@ class SOMAArray : public SOMAObject {
...
+     * domains.
+     */
+    template <typename T>
+    std::optional<std::pair<T, T>> non_empty_domain_slot_opt(
+        const std::string& name) const {
+        try {
+            int32_t is_empty;
+            T ned[2];
+
+            // TODO currently we need to use the TileDB C API in order to check
```
johnkerl, line 749: Do we have a tracking task with @teo-tsirpanis ? If not please open one! :)

libtiledbsoma/src/soma/soma_array.h
```diff
@@ -733,6 +734,41 @@ class SOMAArray : public SOMAObject {
         }
     }
 
+    /**
+     * Retrieves the non-empty domain from the array. This is the union of the
+     * non-empty domains of the array fragments. Return std::nullopt for empty
+     * domains.
+     */
+    template <typename T>
+    std::optional<std::pair<T, T>> non_empty_domain_slot_opt(
```
johnkerl, line 743: This is fine (and thanks for working on this!!) -- but -- I'd love for us to make `non_empty_domain` return `None` when no data have been written. I can do that as a follow-on on #2407 -- or you can ...

johnkerl (approved):
🔥

libtiledbsoma/src/soma/soma_array.h
nguyenv: Yes that would be a good follow-up. I was running into a lot of trouble with converting between the Arrow C structs and pybind11 with `std::optional` / `py::none` so opened this for now to make sure it's working for spatial.

johnkerl: @nguyenv that is awesome! :)

PR #3142, opened 2024-10-07
[python] Avoid RCs of dependent packages for our RC

johnkerl:
We found on `pip install --pre tiledbsoma==0.15.0rc1` that we were getting an undesired `anndata==0.11.0rc2` version of `anndata`.

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -326,8 +326,10 @@ setuptools.setup(
     zip_safe=False,
     setup_requires=["pybind11"],
     install_requires=[
-        # Tracked in https://github.com/single-cell-data/TileDB-SOMA/issues/1785
-        "anndata != 0.10.0",
+        # Temporary for 1.15.0rc series -- avoid RC versions of these packages
+        "anndata==0.10.9",
+        "networkx==3.2.1",
+        "pyparsing==3.1.4",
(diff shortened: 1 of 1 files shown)


apis/python/setup.py
```diff
@@ -326,8 +326,10 @@ def run(self):
     zip_safe=False,
     setup_requires=["pybind11"],
     install_requires=[
-        # Tracked in https://github.com/single-cell-data/TileDB-SOMA/issues/1785
-        "anndata != 0.10.0",
+        # Temporary for 1.15.0rc series -- avoid RC versions of these packages
+        "anndata==0.10.9",
```
ryan-williams, line 330:
```suggestion
        "anndata>=0.10.1,<0.11",
```

I think this will leave downstream users more flexibility

apis/python/setup.py
```diff
@@ -326,8 +326,10 @@ def run(self):
...
+        "networkx==3.2.1",
+        "pyparsing==3.1.4",
```
ryan-williams, line 332: similarly, can these be relaxed a bit? `~=` instead of `==`?

nguyenv (approved):

ryan-williams (approved):

PR #3145, opened 2024-10-08
[r] Handle numeric `coords` better in `SOMASparseNDArray$read()`

mojaveazure:
Fix bug in handling of numeric coordinates when reading a sparse array. Now, always cast any integerish value to `bit64::integer64` (casting a `bit64::integer64` to a `bit64::integer64` is a no-op). Also strengthen value checking of `coords` to
 - require `coords` to be a list of `NULL` or positive, finite, integerish values
 - require `coords` to have a length
 - require `coords` to be a list of length `array$dimnames()` or be named with `array$dimnames()`
 - filter `NULLs` from `coords`; if `coords` is all `NULL`, convert to `NULL`

Modified SOMA methods:
 - `SOMANDArrayBase$private$.convert_coords()`: cast all integerish values to `bit64::integer64`; strengthen value checking of `coords`

[SC-57096](https://app.shortcut.com/tiledb-inc/story/57096)


johnkerl (approved):
:ship:

Issue #1632, opened 2023-08-24
[python] Use new C++ API bindings in Python API (parent task)


nguyenv:
Split-off from https://github.com/single-cell-data/TileDB-SOMA/issues/1240
- [x] #1714
- Pybind11 bindings
    - [x] #1715
    - [x] #2077
    - [x] #2078
    - [x] #2674
- [x] #2228 
- [x] #2240 
- [x] #2241

johnkerl:
I think the @eddelbuettel assign was spurious -- unassigning -- @eddelbuettel if you're truly doing Python work, please let me know.

johnkerl:
All that remains on the Python side is the final deprecation (with phase-out in TileDB-SOMA 1.14) wherein users will need to pass tiledb config via hashmap only, no longer via `tiledb.Ctx`.

So this is blocked until TileDB-SOMA 1.14.

PR #3146, opened 2024-10-08
[c++] Add `SOMAArray::config_options_from_schema ` getter

nguyenv:
As part of the work for #1632 to move tiledb-py (and tiledb-r) out of unit tests, this PR adds accessor `SOMAArray::config_options_from_schema` for retrieving properties such as cell order and filter lists. The method `ArrowAdapter::platform_config_from_tiledb_schema` accepts a `tiledb::ArraySchema` and serializes additional metadata not captured by `SOMAArray::arrow_schema` into a `PlatformConfig` object.

### DIFF
--- a/libtiledbsoma/src/soma/soma_array.h
+++ b/libtiledbsoma/src/soma/soma_array.h
@@ -605,7 +605,7 @@ class SOMAArray : public SOMAObject {
      *
      * @return PlatformConfig
      */
-    PlatformConfig config_options() const {
+    PlatformConfig config_options_from_schema() const {
         return ArrowAdapter::platform_config_from_tiledb_schema(*mq_->schema());
     }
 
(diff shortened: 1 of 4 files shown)


libtiledbsoma/src/soma/soma_array.h
```diff
@@ -598,6 +598,17 @@ class SOMAArray : public SOMAObject {
...
     }
 
+    /**
+     * @brief Get members of the schema (capacity, allows_duplicates,
+     * tile_order, cell_order, offsets_filters, validity_filters, attr filters,
+     * and dim filters) in the form of a PlatformConfig
+     *
+     * @return PlatformConfig
+     */
+    PlatformConfig config_options() const {
```
johnkerl, line 608: This sounds passive & it's not clear from the name alone who's doing what ... maybe `read_back_config_options` or `get_config_options` or `config_options_from_schema` (I like the last one best) ...

johnkerl (changes requested):
This is awesome @nguyenv !!! :)

johnkerl (approved):
🚢

PR #3148, opened 2024-10-08
[python] Append mode with resizing

johnkerl:

**Changes:**

Support experipment-level `upgrade_shape` and `resize`.

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -1206,6 +1206,8 @@ def _write_dataframe_impl(
         arrow_table = _extract_new_values_for_append(df_uri, arrow_table, context)
 
     try:
+        # Note: tiledbsoma.io creates dataframes with soma_joinid being the one
+        # and only index column.
         domain = None
         if NEW_SHAPE_FEATURE_FLAG_ENABLED:
             domain = ((0, shape - 1),)
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -64,6 +64,7 @@ from .._exception import (
     NotCreateableError,
     SOMAError,
 )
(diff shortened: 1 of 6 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1203,9 +1206,13 @@ def _write_dataframe_impl(
         arrow_table = _extract_new_values_for_append(df_uri, arrow_table, context)
 
     try:
+        domain = None
+        if NEW_SHAPE_FEATURE_FLAG_ENABLED:
+            domain = ((0, shape - 1),)
```
jp-dark, line 1211: It looks like this is now assuming the dataframe uses the `soma_joinid` as the one and only index column. If this is the correct assumption (and I think it probably is for the anndata ingestor), can you add a comment about that in the docstring?

jp-dark (approved):
I think this looks good, but I'm not super familiar with the ingestor code so you may want to want to wait for @nguyenv to review as well.

One question: Is that resize code required before ingesting with the new shape feature? If so, what error does the user get if they fail to resize first?

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: I'll do so

Yes `tiledbsoma.io` does choose `soma_joinid` as the one and only index column.

johnkerl:
> One question: Is that resize code required before ingesting with the new shape feature? If so, what error does the user get if they fail to resize first?

@jp-dark they get a message like this:

```
tiledbsoma._exception.SOMAError: TileDB internal: ['UnorderedWriter::dowork]  (FragmentMetadata: Cells are written outside of the defined current domain.)
```

which we can perhaps catch and rethrow with a friendlier error message ...

nguyenv (approved):

PR #3149, opened 2024-10-08
WKB geometry utilities (WIP)

XanthosXanthopoulos:
**Issue and/or context:**

**Changes:**

**Notes for Reviewer:**


libtiledbsoma/src/utils/geometry/geometry.h
```diff
@@ -0,0 +1,42 @@
...
+    MULTIPOINT = 4,
+    MULTILINESTRING = 5,
+    MULTIPOLYGON = 6,
+    GEOMETRYCOLLECTION = 7
+};
+
+using BinaryBuffer = std::vector<uint8_t>;
+
+struct GeometryCollection;
+using GenericGeometry = std::variant<
```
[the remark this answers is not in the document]
XanthosXanthopoulos: My intention is to keep it internal to `libtiledbsoma`. I am open to discuss a different architecture.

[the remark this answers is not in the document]
nguyenv: OK that's fine then. Thanks for clarifying.

libtiledbsoma/src/utils/geometry/operators/io/write.cc
```diff
@@ -0,0 +1,178 @@
...
+size_t WKBSizeOperator::operator()(const LineString& linestring) {
+    if (linestring.points.size() == 0)
+        return 9;
+
+    return 9 +
+           linestring.points.size() * binary_size(linestring.points.front());
+}
+
+size_t WKBSizeOperator::operator()(const Polygon& polygon) {
+    size_t size = 1 + 4 + 4;
```
nguyenv, line 27: Would you consider moving these values as top-level consts to improve readability? For example:

`const size_t WKB_BYTE_ORDER_SIZE = 1;`
`const size_t WKB_GEOEMTRY_TYPE_SIZE = 4;`
`const size_t WKB_SRID_SIZE = 4;`
`const size_t WKB_GEOMETRY_HEADER_SIZE = WKB_BYTE_ORDER_SIZE + WKB_GEOEMTRY_TYPE_SIZE + WKB_SRID_SIZE;`

This would make the code clearer, especially for people (like me) who may not be as familiar with the structure of WKB files.

libtiledbsoma/src/utils/geometry/operators/io/read.h
```diff
@@ -0,0 +1,81 @@
...
+
+#include <cassert>
+#include <cstring>
+#include <stdexcept>
+#include <string>
+#include <variant>
+#include <vector>
+
+#include "../../geometry.h"
+#include "../../utils.h"
```
jp-dark, line 12:
```suggestion
```
This isn't currently being used, I would like to save utils.h for a future PR.

libtiledbsoma/src/utils/geometry/operators/io/write.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include <cassert>
+#include <cstring>
+#include <stdexcept>
+#include <string>
+#include <variant>
+#include <vector>
+
+#include "../../geometry.h"
+#include "../../utils.h"
```
jp-dark, line 12:
```suggestion
```
This isn't currently being used, I would like to save utils.h for a future PR.

libtiledbsoma/src/utils/geometry/utils.h
```diff
@@ -0,0 +1,30 @@
+#ifndef TILEDBSOMA_GEOMETRY_UTILS_H
```
jp-dark, line 1: It doesn't look like `utils.h` is being used in this PR. Let's hold off on adding this in until it's in use.

jp-dark (changes requested):
This PR is large, but very clean. I added a couple of inline comments to the review. In addition:

* For the tests, we should (1) make it easy to add in additional tests for each type, (2) add in additional tests for different edge cases for the geometries and (3) add tests for reading and erroring out on improperly formatted and/or unsupported WKB types. None of this needs to be added to this PR, but we should at least add stories to address this in the future.

* Optional: I think we should move `libtiledbsoma/src/utils/geometry` to just `libtiledbsoma/src/geometry`.

libtiledbsoma/src/utils/geometry/utils.h
```diff
@@ -0,0 +1,30 @@
...
+template <typename... T>
+Operator(T...) -> Operator<T...>;
+
+template <typename G>
+struct Recurse {
+    template <typename... X>
+    decltype(auto) operator()(X&&... x) const& {
+        return g(*this, std::forward<X>(x)...);
+    }
+    G g;
```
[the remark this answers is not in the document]
XanthosXanthopoulos: This is to be removed. Was only useful with lambdas but now with the overloaded `operator()` is unused

jp-dark (approved):

PR #3156, opened 2024-10-09
[python] Canned tests for old/new arrays without/with new shapes

johnkerl:

**Changes:**

Here we keep around an old-style experiment to regress on our ability to upgrade its shapes.

For this PR we only test accessors on the data. On #3148 these unit test will be extended to check their upgrading of the old-style experiment.

**Notes for Reviewer:**


nguyenv (approved):

johnkerl:
Thanks @nguyenv ! :)

PR #3158, opened 2024-10-09
[python/c++] Remove tiledb-py from `test_factory.py`

nguyenv:
**Issue and/or context:**

As part of https://github.com/single-cell-data/TileDB-SOMA/pull/2883 to remove tiledb-py from unit tests

**Changes:**

* Use the SOMA API to create objects and then modify the reserved metadata values; incorrect encoding versions and SOMA object types should still error out as before
* Add `force` flag to `delete_metadata` to match `set_metadata`

### DIFF
--- a/apis/python/src/tiledbsoma/common.cc
+++ b/apis/python/src/tiledbsoma/common.cc
@@ -236,7 +236,7 @@ void set_metadata(
         value_num,
         value_num > 0 ? value.data() : nullptr,
         force);  // The force flag is intended to only be toggled for testing in
-                 // test_factory.py
+                 // fault-injection scenarios
 }
 
 }  // namespace tiledbsoma
--- a/apis/python/src/tiledbsoma/common.cc
+++ b/apis/python/src/tiledbsoma/common.cc
@@ -209,7 +209,10 @@ py::dict meta(std::map<std::string, MetadataValue> metadata_mapping) {
 }
 
 void set_metadata(
-    SOMAObject& soma_object, const std::string& key, py::array value) {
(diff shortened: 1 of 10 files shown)


apis/python/src/tiledbsoma/common.cc
```diff
@@ -228,7 +231,12 @@ void set_metadata(
 
     auto value_num = is_tdb_str(value_type) ? value.nbytes() : value.size();
     soma_object.set_metadata(
-        key, value_type, value_num, value_num > 0 ? value.data() : nullptr);
+        key,
+        value_type,
+        value_num,
+        value_num > 0 ? value.data() : nullptr,
+        force);  // The force flag is intended to only be toggled for testing in
+                 // test_factory.py
```
johnkerl, line 239:
```suggestion
        force);  // The force flag is intended to only be toggled for testing in
                 // fault-injection scenarios
```

apis/python/tests/test_factory.py
```diff
@@ -182,11 +136,18 @@ def test_factory_unknown_files():
 
 
 def _setmetadata(open_tdb_object, metadata_typename, encoding_version):
-    """set only those values which are not None"""
-    changes = {}
+    """force modify the metadata values"""
+    set_metadata = open_tdb_object._handle._handle.set_metadata
```
johnkerl, line 140: i really like this b/c we get access to the `force` flag at the `object._handle._handle` level but it does not work its way up to the the API level (which is good)

johnkerl (approved):
💥

Issue #3087, opened 2024-09-28
[Bug] Can't install tiledbsoma for Python on Fedora 40


rcannood:
**Describe the bug**

Hi all!

Unfortunately, I don't manage to install the latest version of tiledbsoma on my Fedora 40 system. I managed to preproduce it with a Fedora 40 container. I didn't try on other platforms.

Thanks in advance!

**To Reproduce**

```bash
docker run --rm -it fedora:40

dnf install -y python3-devel python3-pip git g++ make

pip install tiledbsoma
```

[Full log](https://github.com/user-attachments/files/17172139/pip-install-tiledbsoma.txt)

      ...
          [ 89%] Linking CXX static library libCatch2.a
          [ 89%] Built target Catch2
          [ 89%] Building CXX object _deps/catch2-build/src/CMakeFiles/Catch2WithMain.dir/catch2/internal/catch_main.cpp.o
          [ 90%] Linking CXX static library libCatch2Main.a
          [ 90%] Built target Catch2WithMain
          cc1plus: all warnings being treated as errors
          gmake[5]: *** [src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/build.make:286: src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/utils/logger.cc.o] Error 1
          gmake[5]: *** Waiting for unfinished jobs....
          cc1plus: all warnings being treated as errors
          gmake[5]: *** [src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/build.make:328: src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/utils/version.cc.o] Error 1
          cc1plus: all warnings being treated as errors
          gmake[5]: *** [src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/build.make:188: src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_context.cc.o] Error 1
          cc1plus: all warnings being treated as errors
          gmake[5]: *** [src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/build.make:244: src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/array_buffers.cc.o] Error 1
          cc1plus: all warnings being treated as errors
[...]

PR #3159, opened 2024-10-09
[c++] Set warnings-as-errors to be off by default

jp-dark:
**Issue and/or context:**
Warnings as errors are useful for developers/CI, but cause unnecessary friction for users. Turn them off by default. See issue #3087.

**Changes:**
* Set `TILEDBSOMA_ENABLE_WERROR` to be `OFF` by default.
* Allow user to set `TILEDBSOMA_ENABLE_WERROR` as an environment variable. This can be overridden by a user provided value.
* Fix CMake code checking if `TILEDBSOMA_ENABLE_WERROR` is `OFF`.
* Update the build script to take `werror` as an argument.
* Explicitly set `werror=true` for CI.


libtiledbsoma/CMakeLists.txt
```diff
@@ -199,8 +199,8 @@ else()
 
   set(TILEDBSOMA_COMPILE_OPTIONS -Wall -Wextra)
 
-  if(TILEDBSOMA_ENABLE_WERROR)
-    set(TILEDBSOMA_WERROR_OPTION -Werror)
+  if(${TILEDBSOMA_ENABLE_WERROR})
+    set(TILEDBSOMA_WERROR_OPTION "") #-Werror)
```
johnkerl, line 203: typo/ to-do?

johnkerl (changes requested):
Unless I'm missing it we need some way to be sure this flag is always set for devleopers. Maybe some env-var check and we can be sure to all set it in our shell `.rc` files ...

PR #3161, opened 2024-10-10
[r] Add Arrow to R type mapper

mojaveazure:
Add new `r_type_from_arrow_type()` function to get R types from Arrow types. This function takes an Arrow schema, field, or datatype and returns the corresponding R type (as indicated by `typeof()`) (or types for a schema)

I chose this route over modifiying the schema directly as modifying the schema turned out to be way harder than I anticipated (thanks to Arrow calling out to C/C++ for nearly all access)

[SC-57387](https://app.shortcut.com/tiledb-inc/story/57387)


apis/r/R/TileDBArray.R
```diff
@@ -132,8 +132,7 @@ TileDBArray <- R6::R6Class(
     #' @description Retrieve the array schema as an Arrow schema (lifecycle: maturing)
     #' @return A [`arrow::schema`] object
     schema = function() {
-      arrow::as_schema(
-        c_schema(self$uri, private$.soma_context));
+      return(arrow::as_schema(c_schema(self$uri, private$.soma_context)))
```
johnkerl, line 135: 🔥

johnkerl (approved):
🚢

PR #3165, opened 2024-10-11
[python/c++] Use `config_options_from_schema` in pytest unit tests

nguyenv:
**Issue and/or context:**

This is the first of multiple changes pulled out from from https://github.com/single-cell-data/TileDB-SOMA/pull/2883 to remove tiledb-py from unit tests

**Changes:**

- Addition of `config_options_from_schema` binding to retrieve `ArraySchema` info not encapsulated by the `ArrowSchema`
- Corrected JSON `dims` and `attrs` by returning `tile` for `dims` and using `filters` keys for `FilterList` values
- Replace `tiledb.FilterList` with `PlatformConfig`'s JSON formatted strings


johnkerl (approved):
🔥 

@nguyenv this is awesome!!!! I'm looking forward to imitating this for R on #2406 :)

PR #3166, opened 2024-10-11
[python/c++] Add `SOMAGroup::is_relative` and use in pytest unit tests

nguyenv:
**Issue and/or context:**

This is the second of multiple changes pulled out from from https://github.com/single-cell-data/TileDB-SOMA/pull/2883 to remove tiledb-py from unit tests

**Changes:**

- Add `SOMAGroup::is_relative` in C++ and associated binding in Pybind11
- Replace tiledb-py `is_relative` test with new binding


johnkerl (approved):
✨

PR #3167, opened 2024-10-11
[python] Consolidate `test_util_tiledb.py` into `test_stats.py`

nguyenv:
**Issue and/or context:**

This is the third of multiple changes pulled out from from https://github.com/single-cell-data/TileDB-SOMA/pull/2883 to remove tiledb-py from unit tests

**Changes:**

Move `test_stats` from `test_util_tiledb.py` into `test_stats.py`


johnkerl (approved):
💥

PR #3168, opened 2024-10-11
[python] Use `pathlib.Path.iterdir` to count fragments

nguyenv:
**Issue and/or context:**

This is the fourth of multiple changes pulled out from from https://github.com/single-cell-data/TileDB-SOMA/pull/2883 to remove tiledb-py from unit tests

**Changes:**

Replace `tiledb.VFS` and `tiledb.FragmentInfoList` with `pathlib.Path.iterdir` to count fragments without relying on tiledb-py


johnkerl (approved):
🚢

PR #3170, opened 2024-10-11
[python] Various tiledb-py removals from pytest

nguyenv:
**Issue and/or context:**

This is the sixth of multiple changes pulled out from from https://github.com/single-cell-data/TileDB-SOMA/pull/2883 to remove tiledb-py from unit tests. These are the more minor remaining changes that could be outlined in a single PR.

**Changes:**

* `test_basic_anndata_io.py`
  - `test_null_obs`: Use arrow schema to check nullability
  - `test_export_obsm_with_holes`: Use `Experiment.metadata` directly
* `test_dataframe.py::test_enum_schema_report`: Remove redundant type checking against tiledb-py. Checks are already encompassed by Arrow schema check below
* `test_dataframe_index_columns.py::test_types_read_errors`: Remove extra parens
* `test_dense_nd_array::pytest_read_to_unwritten_array`: Create expected data output by create a numpy array with the fill value rather than pulling from tiledb-py read query
* `test_experiment_query.py::test_empty_categorical_query`: Replace `tiledb.libtiledb.version()` with parsing using `re` module

**Notes for Reviewer:**

There is one more part that needs to pulled out from https://github.com/single-cell-data/TileDB-SOMA/pull/2883 which involves modifying the `TileDBSOMAContext` tests to reflect `tiledb.Ctx` -> `clib.SOMAContext` changes. That will be tacked onto a PR that fully deprecates `tiledb.Ctx`.

### DIFF
--- a/apis/python/tests/test_basic_anndata_io.py
+++ b/apis/python/tests/test_basic_anndata_io.py
@@ -18,7 +18,6 @@ import tiledbsoma.io
 from tiledbsoma import Experiment, _constants, _factory
 from tiledbsoma._soma_object import SOMAObject
 from tiledbsoma.io._common import _TILEDBSOMA_TYPE, UnsDict, UnsMapping
-import tiledb
 
 from ._util import TESTDATA, assert_adata_equal, make_pd_df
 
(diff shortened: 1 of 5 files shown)


johnkerl (approved):
🔥

PR #3171, opened 2024-10-14
[c++] Geometry envelope computation

XanthosXanthopoulos:
This PR adds the ability to compute an `Envelope` for all supported geometry types.


jp-dark (approved):
Some Doxygen style comments in envelop.h wouldn't hurt, but all in all this looks good to me.

PR #3172, opened 2024-10-15
[python] Adds Visium ingestion option to set image channel order

jp-dark:
Add an input parameter to the experimental `from_visium` ingestor that specifies if images should be ingested with the channel first or last.


nguyenv (approved):
Will we be adding `from_visium` unit tests in pytest in a future PR or will it just be in the Jupyter notebooks?

jp-dark:
> Will we be adding `from_visium` unit tests in pytest in a future PR or will it just be in the Jupyter notebooks?

We will eventually add tests.

PR #3173, opened 2024-10-15
[python] Add `_open_subcollection` helper method to `Scene`

jp-dark:
Add a helper function `_open_subcollection` to the `Scene` that reduces the amount of duplicate code needed in the `Scene` class. This update also adds support for setting/getting transforms on data that is more than one level deep in a scene (e.g. anything stored in `varl`).

### DIFF
--- a/apis/python/src/tiledbsoma/_scene.py
+++ b/apis/python/src/tiledbsoma/_scene.py
@@ -6,7 +6,7 @@
 Implementation of a SOMA Scene
 """
 
-from typing import Any, Optional, Sequence, Union
+from typing import Any, List, Optional, Sequence, Union
 
 import somacore
 from somacore import Axis, CoordinateSpace, CoordinateTransform, IdentityTransform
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/_scene.py
```diff
@@ -61,6 +61,27 @@ def __init__(
...
+        self, subcollection: Union[str, Sequence[str]]
+    ) -> CollectionBase[AnySOMAObject]:
+        if len(subcollection) == 0:
+            raise ValueError("Invalid subcollection: value cannot be empty.")
+        if isinstance(subcollection, str):
+            subcollection = (subcollection,)
+        else:
+            subcollection = tuple(subcollection)
+        coll: CollectionBase[AnySOMAObject] = self
+        parent_name: List[str] = []
```
nguyenv, line 74:
```suggestion
        # Keep track of collection hierarchy for informative error reporting
        parent_name: List[str] = []
```

nguyenv (approved):

PR #3175, opened 2024-10-16
[python] Fix `pytest` instructions for first-time users

johnkerl:
**Issue and/or context:** Found by @kounelisagis . Issue is `test/soco` isn't created by default. It should be.

**Changes:**

* Update the instructions to show the need for `make data`
* On `make data`, extract from the existing tar file rather than re-running the ingestor. ([Our CI already does this](https://github.com/single-cell-data/TileDB-SOMA/blob/1.15.0rc2/.github/workflows/python-ci-single.yml#L119-L125).)
* Surface a better error message in case the user hasn't run `make data`

**Notes for Reviewer:**

This doesn't happen to those of us doing python+soma dev regularly as we have this test data already extracted. And it doesn't happen in CI since CI auto-extracts. This is only a bad-UX issue for first-time/infrequent developers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,13 +46,7 @@ ctest_update:
 
 .PHONY: data
 data:
-	rm -rvf test/soco
-	./apis/python/devtools/ingestor \
-		--soco \
-		-o test/soco \
-		-n \
-		data/pbmc3k_processed.h5ad \
-		data/10x-pbmc-multiome-v1.0/subset_100_100.h5ad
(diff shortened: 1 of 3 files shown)


kounelisagis:
- Cloned repo
- Installed tiledbsoma from source (`pip install -v -e TileDB-SOMA/apis/python`)
- Tried to run tests but got: "RuntimeError: Please run `make data` in the repo base directory"
- Executed `make data`
- Tests run successfully!

Thanks @johnkerl!

kounelisagis (approved):

PR #3182, opened 2024-10-16
[ci] Move CI workers from `macos-12` to `macos-latest`

johnkerl:
**Issue and/or context:** https://github.com/actions/runner-images/issues/10721

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/.github/workflows/python-ci-full.yml
+++ b/.github/workflows/python-ci-full.yml
@@ -21,8 +21,8 @@ jobs:
       fail-fast: false
       matrix:
         # TODO: decide on Windows CI coverage
-        os: [ubuntu-22.04, macos-12]
-        # os: [ubuntu-22.04, macos-12, windows-2019]
+        os: [ubuntu-22.04, macos-latest]
+        # os: [ubuntu-22.04, macos-latest, windows-2019]
         # TODO: add 3.12
         # https://github.com/single-cell-data/TileDB-SOMA/issues/1849
         python-version: ['3.9', '3.10', '3.11']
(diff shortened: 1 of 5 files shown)


.github/workflows/python-ci-full.yml
```diff
@@ -21,16 +21,16 @@ jobs:
       fail-fast: false
       matrix:
         # TODO: decide on Windows CI coverage
-        os: [ubuntu-22.04, macos-12]
-        # os: [ubuntu-22.04, macos-12, windows-2019]
+        os: [ubuntu-22.04, macos-latest]
+        # os: [ubuntu-22.04, macos-latest, windows-2019]
```
johnkerl, line 25: re next line: https://github.com/single-cell-data/TileDB-SOMA/pull/3184

.github/workflows/python-packaging.yml
```diff
@@ -57,7 +57,7 @@ jobs:
             os: ubuntu-20.04
             wheel-name: manylinux2014
           - cibw_build: macosx_x86_64
-            os: macos-12
+            os: macos-latest
```
bkmartinjr, line 60: is there a reason we don't also move to `ubuntu-latest`?

bkmartinjr (approved):

.github/workflows/python-packaging.yml
johnkerl: Great idea @bkmartinjr -- happy to try this as a follow-on PR! :)

johnkerl: #3186

PR #3183, opened 2024-10-16
Fix `is_named_list` bug #3181

### LINKED ISSUE #3181: [r] `set_metadata` method allows unnamed list
## Describe the bug

`set_metadata` method allows  unnamed list.

https://github.com/single-cell-data/TileDB-SOMA/blob/566d88926589241b866c93401f1d167864653c40/apis/r/R/TileDBArray.R#L103-L105

The root cause:

https://github.com/single-cell-data/TileDB-SOMA/blob/566d88926589241b866c93401f1d167864653c40/apis/r/R/utils-assertions.R#L19-L20

Here's the reprex:
```r
is_named_list(list(a = "foo", "hi"))
[1] TRUE
```

Setting  `allow_empty` to `FALSE` for `is_named` assertion, then we're good:
```R
is_named_list <- function(x) {
  is.list(x) && is_named(x, allow_empty = FALSE)
}

# with `allow_empty = FALSE`
is_named_list(list(a = "foo", "hi"))
[1] FALSE

is_named_list(list(a = "foo", b = "hi"))
[1] TRUE
```

johnkerl:
**Issue and/or context:** Fixes #3181 

**Changes:** as on #3181, with unit-test coverage.

### DIFF
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -6,7 +6,7 @@ Description: Interface for working with 'TileDB'-based Stack of Matrices,
     like those commonly used for single cell data analysis. It is documented at
     <https://github.com/single-cell-data>; a formal specification available is at
     <https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md>.
-Version: 1.15.99.6
+Version: 1.15.99.7
 Authors@R: c(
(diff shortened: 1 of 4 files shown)


mojaveazure (approved):
LGTM; please update the changelog, bump the develop version, and 🚢

PR #3201, opened 2024-10-18
[c++] Geometry Dataframe [draft]

XanthosXanthopoulos:
This PR adds a C++ implementation for `SOMAGeometryDataFrame`. The implementation differs from other `SOMAArray` derived objects in a few areas:

- `SOMAGeometryDataFrame` requires two schemas to create itself, one schema similar to all other objects and a schema for the spatial axes. The second schema is necessary to create some additional internal dimension to allow indexing for geometry data.
- `SOMAGeometryDataFrame` can transform geometry data from a list of vertices to the on-disk WKB format automatically. Specifically if the array of data contains a list of vertices in the geometry column (opposed to a binary buffer), the WKB buffer as well as any additional index column are generated automatically. This feature allows both R and Python APIs to behave the same and does not require from them additional dependencies to construct the binary buffers. Moreover in case the indexing for `SOMAGeometryDataFrame` changes in the future minor changes should be necessary to the R and Python APIs.

### DIFF
--- a/libtiledbsoma/src/soma/soma_geometry_dataframe.h
+++ b/libtiledbsoma/src/soma/soma_geometry_dataframe.h
@@ -1,0 +1,190 @@
+/**
+ * @file   soma_geometry_dataframe.h
+ *
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2024 TileDB, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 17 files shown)


libtiledbsoma/src/soma/soma_geometry_dataframe.h
```diff
@@ -0,0 +1,190 @@
...
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
```
jp-dark, line 26: This can be replaced with the smaller C++ license header (see other files).

libtiledbsoma/test/unit_soma_geometry_dataframe.cc
```diff
@@ -0,0 +1,283 @@
...
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
```
jp-dark, line 26: This can be replaced with the smaller copyright header.

PR #3203, opened 2024-10-18
[python] Min-sizing for dataframes/arrays with new shape feature

johnkerl:

**Changes:**

In the design at #2407 we decided that `None` no longer means "as big a domain as possible" but now rather "as small as possible" -- as only upward resizes, not downward resizes, are possible. This is a breaking change we have already agreed on.

**Notes for Reviewer:**


jp-dark (approved):

jp-dark (commented):
Let me verify the spatial Jupyter notebook still runs with the changes to the point cloud dataframe.

jp-dark:
> Let me verify the spatial Jupyter notebook still runs with the changes to the point cloud dataframe.

Success

PR #3204, opened 2024-10-18
[r] Expose block/random writer for `SOMASparseNDArray`

mojaveazure:
Expose a COO writer to enable block/random writes for `SOMASparseNDArrays`; allows downstream consumers to perform block or random writes to sparse arrays

New SOMA methods:
 - `SOMASparseNDArray$.write_coordinates()`: write COO tables to a sparse array; replaces `SOMASparseNDArray$private$.write_coo_dataframe()`

[SC-57804](https://app.shortcut.com/tiledb-inc/story/57804)

**Notes for Reviewer:**
If approved, I will need this released on R-universe as part of the 1.14.x cycle


apis/r/R/SOMASparseNDArray.R
```diff
@@ -181,15 +189,112 @@ SOMASparseNDArray <- R6::R6Class(
         self$tiledb_timestamp %||% "now"
       )
 
-      private$.write_coo_dataframe(coo)
+      self$.write_coordinates(coo)
+      # private$.write_coo_dataframe(coo)
```
johnkerl, line 193: please remove

apis/r/R/SOMASparseNDArray.R
```diff
@@ -181,15 +189,112 @@ SOMASparseNDArray <- R6::R6Class(
...
+
+      stopifnot(
+        "'values' must be a data frame or Arrow Table" = is.data.frame(values) ||
+          inherits(values, what = 'Table'),
+        "'values' must have one column for each dimension and the data" = ncol(values) == length(dnames) + 1L,
+        "'values' must be named with the dimension and attribute labels" = is.null(names(values)) ||
+          identical(names(values), c(dnames, attrn))
+      )
+
+      # Arrow Tables cannot have NULL names, so this only applies to data frames
```
johnkerl, line 225:
```suggestion
      # Arrow Tables cannot have NULL names, so this only applies to dataframes
```

apis/r/R/SOMASparseNDArray.R
```diff
@@ -181,15 +189,112 @@ SOMASparseNDArray <- R6::R6Class(
...
+      # Check dimensions
+      spdl::debug("[SOMASparseNDArray$.write_coordinates] checking dimension values")
+      for (i in seq_along(dnames)) {
+        dn <- dnames[i]
+        integerish <- if (is.data.frame(values)) {
+          rlang::is_integerish(values[[dn]], finite = TRUE) ||
+            (bit64::is.integer64(values[[dn]]) && !any(is.na(values[[dn]])))
+        } else {
+          grepl('^int[[:digit:]]{1,2}$', x = values[[dn]]$type$name)
+        }
```
johnkerl, line 240: this would make a nice helper function, rather than being inline here

apis/r/R/SOMASparseNDArray.R
```diff
@@ -181,15 +189,112 @@ SOMASparseNDArray <- R6::R6Class(
...
+        offending <- sprintf("(offending column: '%s')", dn)
+        if (!integerish) {
+          stop("All dimension columns must be integerish ", offending)
+        }
+        if (as.logical(min(values[[dn]]) < 0L)) {
+          stop("Dimension columns cannot contain negative values ", offending)
```
johnkerl, line 246: perf implications of doing these full data scans, tradeoff against error-checking?

apis/r/R/SOMASparseNDArray.R
```diff
@@ -222,7 +222,7 @@ SOMASparseNDArray <- R6::R6Class(
           identical(names(values), c(dnames, attrn))
       )
 
-      # Arrow Tables cannot have NULL names, so this only applies to data frames
+      # Arrow Tables cannot have NULL names, so this only applies to dataframes
```
johnkerl, line 225: Actually you have 'data frame' throughout this PR, and if that's what people say in R land, that's fine ... I thought it was dataframe all one word & if I'm right you should do a search-and-replace on this PR ...

apis/r/R/SOMASparseNDArray.R
mojaveazure: Addressed in dd2512c

apis/r/R/SOMASparseNDArray.R
mojaveazure: Added in 2537fa0

apis/r/R/SOMASparseNDArray.R
mojaveazure: My thought process is
 - only checking one value (`min()`/`max()`), pulling this value is fast and a single check is fast
 - most users will go through `$write()`, which is limited to 2^32 - 1 non-zero values due to limitations of captial-m-Matrix
 - `$.write_coordinates()` could support more, but with the introduction of my DelayedArray front-end, most usage of this from outside of tiledbsoma-r will be through chunks from DelayedArray (which must fit in memory)

apis/r/R/SOMASparseNDArray.R
mojaveazure: Base-R uses "data frame" as seen in [`?data.frame`](https://stat.ethz.ch/R-manual/R-devel/library/base/html/data.frame.html)

johnkerl (approved):
LGTM!

PR #3206, opened 2024-10-18
[python] Add support for transforms in the `Scene` `add_new_*` methods

jp-dark:
Add support for setting the transform when a `PointCloud` or `MultiscaleImage` is first added to a `Scene`.

### DIFF
--- a/apis/python/src/tiledbsoma/_scene.py
+++ b/apis/python/src/tiledbsoma/_scene.py
@@ -221,6 +221,8 @@ class Scene(  # type: ignore[misc]   # __eq__ false positive
         *,
         transform: Optional[CoordinateTransform],
         uri: Optional[str] = None,
+        axis_names: Sequence[str] = ("c", "y", "x"),
+        axis_types: Sequence[str] = ("channel", "height", "width"),
         **kwargs: Any,
     ) -> MultiscaleImage:
         """Adds a ``MultiscaleImage`` to the scene and sets a coordinate transform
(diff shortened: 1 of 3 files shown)


apis/python/src/tiledbsoma/_scene.py
```diff
@@ -245,21 +247,66 @@ def add_new_multiscale_image(
...
+                axis_names=axis_names,
+                axis_types=axis_types,
                 **kwargs,
             ),
             uri,
         )
 
+        # Store the metadata for the transform.
+        if transform is not None:
+            coll.metadata[f"soma_scene_registry_{key}"] = transform_to_json(transform)
```
nguyenv, line 305: Not in this PR, but one thing I'd like to do in the future is make it so that the user can't modify these metadata values directly.

In the C++ API, we [throw an error](https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/src/soma/soma_array.cc#L1090-L1094) if there is an attempt to modify these reserved metadata keys. BUT there is a way to [bypass](https://github.com/single-cell-data/TileDB-SOMA/blob/main/libtiledbsoma/src/soma/soma_array.cc#L1089) this by passing the `force` flag. Internally, we could use `coll._handle.add_metadata(f"soma_scene_registry_{key}", transform_to_json(transform), force=True)`. This way we can still modify the metadata values as needed but the user cannot.

nguyenv (approved):

PR #3208, opened 2024-10-18
[r] Min-sizing for dataframes/arrays with new shape feature

johnkerl:

**Changes:**

In the design at #2407 we decided that `None` no longer means "as big a domain as possible" but now rather "as small as possible" -- as only upward resizes, not downward resizes, are possible. This is a breaking change we have already agreed on.

This is the R counterpart to the Python PRs #3190 #3191 #3192 #3193 #3194 #3203. I didn't split this one up into multiple PRs as I did those. However, the mods are the same:

* Have unit-test cases always provide `domain` at `create`
* Apply the policy that `domain` being `NULL` overall or slotwise makes smallest-possible rather than largest-possible.

**Notes for Reviewer:**


mojaveazure (approved):
LGTM; please update the changelog, bump the develop version, and 🚢

PR #3209, opened 2024-10-20
[python] Remove the no-longer-needed wrapper layers [WIP]

johnkerl:
These were useful during the cppification project -- they allowed us to migrate from tiledb-py-backed classes to libtiledbsoma-backed classes.

At this point, however, they simply add to the keystroking on all Python dev work. It's time to get rid of them, to save us all from needless copypasta in our development work moving forward.

Notes:

* This PR is not ready for review.
* This is biggish, and not done yet. I am almost always strongly attached to small/split-up PRs, for the benefit of the author as well as the reviewer. This one ... I'm not sure yet how to split it up.


johnkerl:
This would need a massive rebase, as old as it is -- better to start from scratch if needed.

PR #3211, opened 2024-10-21
[c++] `can_upgrade_domain`

johnkerl:

**Changes:**

* This is `can_upgrade_domain` for `SOMADataFrame` at the C++ level

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/soma/soma_array.cc
+++ b/libtiledbsoma/src/soma/soma_array.cc
@@ -1448,7 +1448,7 @@ std::vector<int64_t> SOMAArray::maxshape() {
 
 // This is a helper for can_upgrade_shape and can_resize, which have
 // much overlap.
-std::pair<bool, std::string> SOMAArray::_can_set_shape_helper(
+StatusAndReason SOMAArray::_can_set_shape_helper(
     const std::vector<int64_t>& newshape,
     bool is_resize,
     std::string function_name_for_messages) {
@@ -1525,7 +1525,7 @@ std::pair<bool, std::string> SOMAArray::_can_set_shape_helper(
 // This is a helper for _can_set_shape_helper: it's used for comparing
 // the user's requested shape against the core current domain or core (max)
 // domain.
-std::pair<bool, std::string> SOMAArray::_can_set_shape_domainish_subhelper(
+StatusAndReason SOMAArray::_can_set_shape_domainish_subhelper(
     const std::vector<int64_t>& newshape,
     bool check_current_domain,
     std::string function_name_for_messages) {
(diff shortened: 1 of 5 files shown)


libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -1882,6 +1882,189 @@ void SOMAArray::_set_soma_joinid_shape_helper(
     schema_evolution.array_evolve(uri_);
 }
 
+std::pair<bool, std::string> SOMAArray::can_upgrade_domain(
```
nguyenv, line 1885: Did we add a `using StatusAndReason = std::pair<bool, std::string>` alias in a previous PR? I thought we did but now can't find it.

libtiledbsoma/src/soma/soma_array.h
```diff
@@ -1157,6 +1157,12 @@ class SOMAArray : public SOMAObject {
             newshape, false, function_name_for_messages);
     }
 
+    /**
+     * This is for SOMADataFrame.
+     */
+    std::pair<bool, std::string> can_upgrade_domain(
+        const ArrowTable& newdomain, std::string function_name_for_messages);
```
nguyenv, line 1164: Can this just reside in the `SOMADataFrame` class or are there issues with doing that, so we need to place it in `SOMAArray`?

nguyenv (approved):

libtiledbsoma/src/soma/soma_array.cc
johnkerl: oh yes -- indeed, let me check 👀

libtiledbsoma/src/soma/soma_array.h
johnkerl: There are issues with doing that:
https://github.com/single-cell-data/TileDB-SOMA/blob/62dd129515c8e6928d910a2d20f6e0ac431c9fe7/libtiledbsoma/src/soma/soma_array.h#L1230-L1234

I can propagate that comment material

nguyenv: Ah ic- thanks. No need to copy it over if it's already in the header.

johnkerl: Adding as top-of-file comment :)

libtiledbsoma/src/soma/soma_array.cc
johnkerl: I was puzzled too! Turns out the reason neither of us could find this is it was done in Python not C++ :D 
https://github.com/single-cell-data/TileDB-SOMA/blob/62dd129515c8e6928d910a2d20f6e0ac431c9fe7/apis/python/src/tiledbsoma/_types.py#L82-L84

It's worth adapting here too though.

johnkerl: (added)

PR #3212, opened 2024-10-21
[c++] Geometry Dataframe

XanthosXanthopoulos:
This PR adds the new `SOMAGeometryDataFrame` object. This new object requires spatial axes to be determined and uses them internally to index geometry objects stored as a WKB encoded attribute. To differentiate between binary attributes and WKB binary attributes the Arrow schema requires a `dtype` metadata key with the value `WKB`.

### DIFF
--- a/libtiledbsoma/src/soma/soma_geometry_dataframe.cc
+++ b/libtiledbsoma/src/soma/soma_geometry_dataframe.cc
@@ -1,0 +1,125 @@
+/**
+ * @file   soma_geometry_dataframe.cc
+ *
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2024 TileDB, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @section DESCRIPTION
+ *
+ *   This file defines the SOMAGeometryDataFrame class.
+ */
+
+#include "soma_geometry_dataframe.h"
+#include "../utils/util.h"
+
+#include <regex>
+
+namespace tiledbsoma {
+using namespace tiledb;
+
+//===================================================================
+//= public static
+//===================================================================
+
+void SOMAGeometryDataFrame::create(
+    std::string_view uri,
+    std::unique_ptr<ArrowSchema> schema,
+    ArrowTable index_columns,
+    ArrowTable spatial_columns,
+    std::shared_ptr<SOMAContext> ctx,
+    PlatformConfig platform_config,
+    std::optional<TimestampRange> timestamp) {
+    std::vector<std::string> spatial_axes;
+    auto tiledb_schema = ArrowAdapter::tiledb_schema_from_arrow_schema(
+        ctx->tiledb_ctx(),
+        std::move(schema),
+        ArrowTable(
+            std::move(index_columns.first), std::move(index_columns.second)),
+        "SOMAGeometryDataFrame",
+        true,
+        platform_config,
+        ArrowTable(
+            std::move(spatial_columns.first),
+            std::move(spatial_columns.second)));
+    auto array = SOMAArray::create(
+        ctx, uri, tiledb_schema, "SOMAGeometryDataFrame", timestamp);
+}
+
+std::unique_ptr<SOMAGeometryDataFrame> SOMAGeometryDataFrame::open(
+    std::string_view uri,
+    OpenMode mode,
+    std::shared_ptr<SOMAContext> ctx,
+    std::vector<std::string> column_names,
+    ResultOrder result_order,
+    std::optional<TimestampRange> timestamp) {
+    return std::make_unique<SOMAGeometryDataFrame>(
+        mode, uri, ctx, column_names, result_order, timestamp);
+}
+
+bool SOMAGeometryDataFrame::exists(
+    std::string_view uri, std::shared_ptr<SOMAContext> ctx) {
+    try {
+        auto obj = SOMAObject::open(uri, OpenMode::read, ctx);
+        return "SOMAGeometryDataFrame" == obj->type();
+    } catch (TileDBSOMAError& e) {
+        return false;
+    }
+}
+
+//===================================================================
+//= public non-static
+//===================================================================
+
+std::unique_ptr<ArrowSchema> SOMAGeometryDataFrame::schema() const {
+    return this->arrow_schema();
+}
+
+const std::vector<std::string> SOMAGeometryDataFrame::index_column_names()
+    const {
+    return this->dimension_names();
+}
+
+const std::vector<std::string> SOMAGeometryDataFrame::spatial_column_names()
+    const {
+    std::vector<std::string> names;
+    std::unordered_set<std::string> unique_names;
+    std::regex rgx("tiledb__internal__(\\S+)__");
+    std::smatch matches;
+    for (auto dimension : this->dimension_names()) {
+        if (std::regex_search(dimension, matches, rgx)) {
+            if (unique_names.count(matches[1].str()) == 0) {
+                unique_names.insert(matches[1].str());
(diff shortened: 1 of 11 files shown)


jp-dark (commented):
That parts of this I understand look good to me, but I'm not familiar enough with the arrow adapter code to have a good sense of what impact the change might have there.

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -1150,6 +1190,104 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
...
+    std::shared_ptr<Context> ctx,
+    ArrowSchema* arrow_schema,
+    std::string_view type_metadata,
+    PlatformConfig platform_config) {
+    auto type = ArrowAdapter::to_tiledb_format(arrow_schema->format);
+    if (strcmp(arrow_schema->name, "soma_geometry") == 0) {
+        if (type_metadata.compare("WKB") == 0) {
+            type = TILEDB_GEOM_WKB;
+        } else {
+            throw std::runtime_error(
```
johnkerl, line 1204: please use `fmt::format` with `{}` as elsewhere in this file

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -1150,6 +1190,104 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
...
+            "[ArrowAdapter] dictionary for {} as {} {}",
+            std::string(arrow_schema->name),
+            tiledb::impl::type_to_str(enmr_type),
+            std::string(enmr_format)));
+    }
+
+    return {attr, enmr};
+}
+
+std::pair<Dimension, bool> ArrowAdapter::tiledb_dimension_from_arrow_schema(
```
johnkerl, line 1250: i'd prefer you move the body of this dimension function above the body of the corresponding attribute funtion

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -257,6 +257,37 @@ class ArrowAdapter {
...
+        std::string_view type_metadata,
+        PlatformConfig platform_config = PlatformConfig());
+
+    /**
+     * @brief Get a TileDB dimension from an Arrow schema.
+     *
+     * @return std::pair<Dimension, bool> The TileDB dimension with a boolean
+     * flag indicating whether or not the dimension uses `current domain`.
+     */
+    static std::pair<Dimension, bool> tiledb_dimension_from_arrow_schema(
```
johnkerl, line 283: i'd prefer you move the declaration of this dimension function above the corresponding attribute function

libtiledbsoma/src/soma/soma_geometry_dataframe.h
```diff
@@ -0,0 +1,179 @@
...
+     * @param spatial_columns The spatial column names with associated domains
+     * and tile extents per dimension
+     * @param ctx SOMAContext
+     * @param platform_config Optional config parameter dictionary
+     * @param timestamp Optional the timestamp range to write SOMA metadata info
+     */
+    static void create(
+        std::string_view uri,
+        std::unique_ptr<ArrowSchema> schema,
+        ArrowTable index_columns,
```
johnkerl, line 68: i see no reason not to use const references here

libtiledbsoma/src/soma/soma_geometry_dataframe.h
```diff
@@ -0,0 +1,179 @@
...
+        std::optional<TimestampRange> timestamp = std::nullopt);
+
+    /**
+     * @brief Open and return a SOMAGeometryDataFrame object at the given URI.
+     *
+     * @param uri URI to create the SOMAGeometryDataFrame
+     * @param mode read or write
+     * @param ctx SOMAContext
+     * @param column_names A list of column names to use as user-defined index
+     * columns (e.g., ``['cell_type', 'tissue_type']``). All named columns must
```
johnkerl, line 81: are these example column names legit for geometry dataframes?

libtiledbsoma/src/soma/soma_geometry_dataframe.cc
```diff
@@ -0,0 +1,125 @@
...
+    ArrowTable spatial_columns,
+    std::shared_ptr<SOMAContext> ctx,
+    PlatformConfig platform_config,
+    std::optional<TimestampRange> timestamp) {
+    std::vector<std::string> spatial_axes;
+    auto tiledb_schema = ArrowAdapter::tiledb_schema_from_arrow_schema(
+        ctx->tiledb_ctx(),
+        std::move(schema),
+        ArrowTable(
+            std::move(index_columns.first), std::move(index_columns.second)),
```
johnkerl, line 58: i believe you can avoid this by having this constructor take the `ArrowTable` parameters as const references

libtiledbsoma/test/unit_soma_geometry_dataframe.cc
```diff
@@ -0,0 +1,149 @@
...
+                  .tiledb_datatype = TILEDB_FLOAT64,
+                  .dim_max = 100,
+                  .string_lo = "N/A",
+                  .string_hi = "N/A",
+                  .use_current_domain = use_current_domain})});
+
+        std::vector<helper::AttrInfo> attr_infos({helper::AttrInfo(
+            {.name = "quality", .tiledb_datatype = TILEDB_FLOAT64})});
+
+        // Check the point cloud doesn't exist yet.
```
johnkerl, line 87: copy/paste error?

libtiledbsoma/test/unit_soma_geometry_dataframe.cc
```diff
@@ -0,0 +1,149 @@
...
+        REQUIRE(!SOMAGeometryDataFrame::exists(uri, ctx));
+
+        // Create the point cloud.
```
johnkerl, line 90: copy/paste error?

libtiledbsoma/test/unit_soma_geometry_dataframe.cc
```diff
@@ -0,0 +1,149 @@
...
+        auto [schema, index_columns] =
+            helper::create_arrow_schema_and_index_columns(
+                dim_infos, attr_infos);
+        auto spatial_columns = helper::create_column_index_info(
+            spatial_dim_infos);
+
+        SOMAGeometryDataFrame::create(
+            uri,
+            std::move(schema),
+            ArrowTable(
```
johnkerl, line 100: please use const references to avoid the moves

libtiledbsoma/test/unit_soma_geometry_dataframe.cc
```diff
@@ -0,0 +1,149 @@
...
+            expected_spatial_column_names);
+        REQUIRE(soma_geometry->nnz() == 0);
+        soma_geometry->close();
+
+        auto soma_object = SOMAObject::open(uri, OpenMode::read, ctx);
+        REQUIRE(soma_object->uri() == uri);
+        REQUIRE(soma_object->type() == "SOMAGeometryDataFrame");
+        soma_object->close();
+    }
+}
```
johnkerl, line 149: seems like we could use some tests which write & read data -- is this planned for a future PR?

johnkerl (changes requested):

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -978,85 +976,106 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
...
+                            type_metadata,
+                            "tiledb__internal__",
+                            "__max",
+                            platform_config);
+
+                        dims.insert({min_dim.first.name(), min_dim.first});
+                        dims.insert({max_dim.first.name(), max_dim.first});
+
+                        use_current_domain &= min_dim.second;
+                        use_current_domain &= max_dim.second;
```
nguyenv, line 1018: For readability, I think this can be pulled out into a separate `_set_spatial_dimension` (or better name) helper function.

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -978,85 +976,106 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
...
-            auto achild = index_column_array->children[i];
-            auto schild = index_column_schema->children[i];
-            auto col_name = schild->name;
-            if (strcmp(child->name, col_name) == 0) {
-                if (ArrowAdapter::arrow_is_var_length_type(child->format)) {
-                    type = TILEDB_STRING_ASCII;
-                }
+            if (strcmp(child->name, index_column_schema->children[i]->name) ==
+                0) {
+                if (strcmp(child->name, "soma_geometry") == 0 &&
```
nguyenv, line 981: Maybe we should `const std::string` some of these in `utils/common.h` like "soma_geometry", "tiledb__internal__", "__min", and "__max"

libtiledbsoma/test/unit_soma_geometry_dataframe.cc
XanthosXanthopoulos: Yes, there is #3222 which adds tests for read/write.

libtiledbsoma/src/soma/soma_geometry_dataframe.h
XanthosXanthopoulos: Agrre on const reference here but will require refactoring all other classes which `tiledb_schema_from_arrow_schema`. Should be done to a follow-up PR maybe?

johnkerl: it doesn't need to be done everywhere -- it can be done in places you add

it can also be done all at once in a follow-on

nguyenv (approved):
I think this look good! Maybe wait for one more approval since it's a larger PR though.

libtiledbsoma/src/utils/common.h
```diff
@@ -42,6 +42,9 @@ namespace tiledbsoma {
 const std::string SOMA_OBJECT_TYPE_KEY = "soma_object_type";
 const std::string ENCODING_VERSION_KEY = "soma_encoding_version";
 const std::string ENCODING_VERSION_VAL = "1.1.0";
+const std::string SOMA_GEOMETRY_COLUMN_NAME = "soma_geometry";
+const std::string SOMA_GEOMETRY_DIMENSION_PREFIX = "tiledb__internal__";
```
johnkerl, line 46:
```suggestion
const std::string SOMA_GEOMETRY_DIMENSION_PREFIX = "tiledb_internal__";
```

PR #3229, opened 2024-10-23
[c++] Fix fast path for `SparseNDArray` `nnz`

johnkerl:
**Issue and/or context:** [[sc-58277]](https://app.shortcut.com/tiledb-inc/story/58277/python-sparse-array-nnz-performance-regression)

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/soma/soma_array.cc
+++ b/libtiledbsoma/src/soma/soma_array.cc
@@ -1351,17 +1351,19 @@ uint64_t SOMAArray::nnz() {
     uint64_t total_cell_num = 0;
     std::vector<std::array<uint64_t, 2>> non_empty_domains(fragment_count);
 
-    // The loop after this only works if dim 0 is int64 soma_joinid.
-    // That's the case for _almost_ all SOMADataFrame objects, but
+    // The loop after this only works if dim 0 is int64 soma_joinid or
+    // soma_dim_0. That's the case for _almost_ all SOMADataFrame objects, but
     // not the "variant-indexed" ones: the SOMA spec only requires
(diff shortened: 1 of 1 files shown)


bkmartinjr (changes requested):
I understand why this change "fixes" the problem, but I'm uncomfortable with the design (in both the original PR and in this fix).  It relies on a heuristic (names), and disallows lots of cases where the fast path would work fine (e.g., other non-variable-sized dimensions)

Possible solutions to consider:
1. directly test the dimension to determine if it variable-sized. Only then fall back (ie.., don't test name or hard-wire to only int64)
2. Or, given that the most common case, simply wrap the nonempty-domain fetch in a try/catch, and call the slow path only when you can't get it.

Either of these seem far less likely to bite us in the future as they are not heuristics based

johnkerl:
@bkmartinjr  https://github.com/single-cell-data/TileDB-SOMA/issues/3231 if you're willing to take this as a follow-on

> and disallows lots of cases where the fast path would work fine

You're correct

* The code before #2990 was broken and wrong (in ways that I didn't document well enough at the time)
* The change I applied on #2990 was too strict -- it disallowed the fast path in certain cases, including -- importantly -- the one fixed by this PR
* But we can go further and allow more cases to take the fast path

I do want to do that as separate PRs -- this one and the to-be-created tracked on #3231 -- precisely because this PR "gets it right" for the most important cases which are (a) default-indexed dataframes; (b) all ND arrays always (which _must_ have int64 dims). So the follow-on PR will enable us to take the fast path for non-default-indexed dataframes, and that will be nice.

bkmartinjr (approved):

PR #3240, opened 2024-10-25
Add spatial-io optional dependencies

jp-dark:
Add optional spatial ingest/outgest requirements to `setup.py`.


johnkerl (approved):
LGTM but I defer to @ryan-williams in such things :)

ryan-williams (approved):

Issue #3246, opened 2024-10-28
[python/ci] Investigate recent CI failure


johnkerl:
Recent CI jobs are failing with typeguard-related errors. Examples:

* https://github.com/single-cell-data/TileDB-SOMA/pull/3244
* https://github.com/TileDB-Inc/centralized-tiledb-nightlies/issues/25

As an experiment I took a recent merge to `main` and simply re-ran it with zero code mods:

* https://github.com/single-cell-data/TileDB-SOMA/actions/runs/11522767244/job/32079325840
* https://github.com/single-cell-data/TileDB-SOMA/actions/runs/11522767244/job/32157898699

The only diffs I could find by log-diffing were `pyarrow` 17 vs 18, which I've ruled out via #3245; also `typeguard` 4.4.0 recently came out and I don't believe that's the problem either but I'll verify.

[[sc-58398]](https://app.shortcut.com/tiledb-inc/story/58398/investigate-recent-ci-failure)

ryan-williams:
`main` [is green](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/11556516471) again as of #3247.

Pinning to a specific typeguard SHA should be reasonably stable, though not ideal "long-term." I'll look into the actual issue a bit more, it may require a small change in Typeguard. 

https://github.com/single-cell-data/TileDB-SOMA/issues/3216 may be the best place for follow-up, and close this since CI is OK again?

PR #3247, opened 2024-10-28
Pin typeguard@afad2c7

ryan-williams:
Context: https://github.com/single-cell-data/TileDB-SOMA/issues/3246

https://github.com/agronholm/typeguard/pull/496 caused CI to begin failing on Python 3.9 ([example](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/11522767244/job/32157898699)).

This pins Typeguard to https://github.com/agronholm/typeguard/commit/afad2c7b6be830900776922bb39f9346c2e77f6f (the last commit that works with our build), while we try to root-cause or work around more durably.

### DIFF
--- a/apis/python/requirements_dev.txt
+++ b/apis/python/requirements_dev.txt
@@ -3,8 +3,9 @@ ruff
 pytest
 pytest-cov
 sparse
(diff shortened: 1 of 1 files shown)


johnkerl (approved):
:ship:

PR #3249, opened 2024-10-28
Set domain to be appropriately sized in Visium ingestion

jp-dark:
This sets the domain for all Visium assets to be large enough to hold the current data.

A fix was added to allow setting a string domain to be the full domain instead of the smallest possible domain.


johnkerl (approved):
🚢

PR #3259, opened 2024-10-29
Embrace new-shape feature flag [pending core 2.27]

johnkerl:
**Issue and/or context:** #3273

**Changes:**

Remove the if-before-times logic and keep the is-new-era logic.

**Notes for Reviewer:**

* This PR is a work in progress. It is not ready for review.
* This PR will fail until we depend on core 2.27, before which current-domain support is unavailable for dense arrays.
* Note: https://github.com/single-cell-data/TileDB-SOMA/pull/3259/files?diff=split&w=1 will hide whitespace-only diffs, of which there are a great deal in this PR.


johnkerl:
I'll submit a new PR, once 2.27.0rc0 is out, rather than rebasing this one.

PR #3263, opened 2024-10-30
[c++] Apply subarrays for dense reads and writes

johnkerl:

**Changes:**

As found on #3244:

* The main context of #3244 is:
  * For dense arrays only
  * If any dim's subarray range isn't specified, core uses the core domain
  * With new-shape feature-flag enabled, and with core 2.27 where dense arrays _can_ have current domain (they couldn't in 2.26), we are indeed using data-sized core current domain and huge core domain
  * Put those together and we get these problems
  * Note there are a few other issues on this PR, and a few _not_ yet included in this PR, other than that -- and those should be split out into separate PRs
* On this PR I've ben finding callsites, for write and for read, for Python and R, and touching up the ranges which end up being core subarrays on each dim slot
  * This is tedious and error-prone
* What we should do instead is:
  * In one place -- in `libtiledbsoma`, in `ManagedQuery`, before `submit_query`
  * If any dim's range has not been set up to this point, and if the array has a current domain, simply set the ranges on that dim to be the current domain's `(lo, hi)`
  * We know (for a fact) that core will go ahead and use the too-big domain anyway so this is safe

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/soma/managed_query.cc
+++ b/libtiledbsoma/src/soma/managed_query.cc
@@ -30,10 +30,10 @@
  * This file defines the performing TileDB queries.
  */
 
-#include "managed_query.h"
 #include <tiledb/array_experimental.h>
 #include <tiledb/attribute_experimental.h>
 #include "../utils/logger.h"
+#include "managed_query.h"
 #include "utils/common.h"
 namespace tiledbsoma {
 
@@ -62,7 +62,7 @@ void ManagedQuery::reset() {
     query_ = std::make_unique<Query>(*ctx_, *array_);
     subarray_ = std::make_unique<Subarray>(*ctx_, *array_);
 
-    subarray_range_set_ = false;
+    subarray_range_set_ = {};
     subarray_range_empty_ = {};
     columns_.clear();
     results_complete_ = true;
@@ -109,7 +109,8 @@ void ManagedQuery::setup_read() {
         // domain on dimension 0. In the case that the non-empty domain does not
         // exist (when the array has not been written to yet), use dimension 0's
         // full domain
-        if (schema.array_type() == TILEDB_DENSE && !subarray_range_set_) {
+        if (schema.array_type() == TILEDB_DENSE &&
+            !_has_any_subarray_range_set()) {
             // Check if the array has been written to by using the C API as
             // there is no way to to check for an empty domain using the current
             // CPP API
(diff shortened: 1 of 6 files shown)


libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -191,6 +167,130 @@ void ManagedQuery::submit_read() {
...
+    auto schema = array_->schema();
+    for (const auto& dim : schema.domain().dimensions()) {
+        std::string dim_name = dim.name();
+        if (subarray_range_set_[dim_name]) {
+            LOG_TRACE(fmt::format(
+                "[ManagedQuery] _fill_in_subarrays continue {}", dim_name));
+            continue;
+        }
+
+        // Dense arrays are (as of this writing) all DenseNDArray.
```
nguyenv, line 268:
```suggestion
        // Dense arrays are (as of this writing in 1.15.0) all DenseNDArray.
```

libtiledbsoma/src/soma/managed_query.h
```diff
@@ -414,6 +407,53 @@ class ManagedQuery {
...
+     * of 2.27. Also, without current-domain support, we had small domains; with
+     * it, we have huge core domains (2^63-ish) since they are immutable, and
+     * small current domains as they are upward-mutable. (The soma domain and
+     * maxdomain, respectively, are core current domain and domain.) Thirdly,
+     * for a long time now, if a query doesn't have a subarray set on any
+     * particular dim, core will use the core domain on that dim. That was fine
+     * when core domains were small; not fine now that they are huge. In this
+     * routine, if the array is dense, for each dim without a subarray set,
+     * we set it to match the soma domain. This guarantees correct behavior.
+     */
```
nguyenv, line 451:
```suggestion
    /**
     * This handles a few internals.
     *
     * One is that a dense array must have _at least one_
     * dim's subarray set for a read query. Without that, reads fail immediately
     * with the unambiguous
     *
     *   DenseReader: Cannot initialize reader; Dense reads must have a subarray
     *   set
     *
     * The other is a combination of several things. Firstly, is current-domain
     * support which we have for sparse arrays as of core 2.26, and for dense as
     * of 2.27. Secondly, without current-domain support, we had small domains; with
     * it, we have huge core domains (2^63-ish) which are immutable, and
     * small current domains which are upward-mutable. (The soma domain and
     * maxdomain, respectively, are core current domain and domain.) Thirdly,
     * if a query doesn't have a subarray set on any
     * particular dim, core will use the core domain on that dim. That was fine
     * when core domains were small; not fine now that they are huge. In this
     * routine, if the array is dense, for each dim without a subarray set,
     * we set it to match the soma domain. This guarantees correct behavior.
     */
```

nguyenv (approved):

johnkerl:
Thanks @nguyenv !!! :)

PR #3266, opened 2024-10-31
[c++] Hide internal dimensions

XanthosXanthopoulos:
Geometry dataframe creates some extra internal dimensions to support spatial indexing. When returning back the index column names and arrow schema these dimensions (being implementation-specific) should be hidden and replaces with `soma_geometry` as index column.

Additionally setting a range for spatial axes should only need the axis name and not the internal index names.

### DIFF
--- a/libtiledbsoma/src/utils/arrow_adapter.h
+++ b/libtiledbsoma/src/utils/arrow_adapter.h
@@ -230,7 +230,7 @@ class ArrowAdapter {
         std::shared_ptr<Context> ctx, std::shared_ptr<Array> tiledb_array);
 
     /**
-     * @brief Create a an ArrowSchema from TileDB Dimension
+     * @brief Create an ArrowSchema from TileDB Dimension
      *
      * @return ArrowSchema
      */
--- a/libtiledbsoma/src/utils/arrow_adapter.h
+++ b/libtiledbsoma/src/utils/arrow_adapter.h
@@ -243,9 +243,7 @@ class ArrowAdapter {
      * @return ArrowSchema
      */
     static std::unique_ptr<ArrowSchema> arrow_schema_from_tiledb_attribute(
-        Attribute& attribute,
-        std::shared_ptr<Context> ctx,
-        std::shared_ptr<Array> tiledb_array);
+        Attribute& attribute, const Context& ctx, const Array& tiledb_array);
 
     /**
      * @brief Get members of the TileDB Schema in the form of a PlatformConfig
--- a/libtiledbsoma/src/utils/arrow_adapter.h
+++ b/libtiledbsoma/src/utils/arrow_adapter.h
@@ -238,7 +238,7 @@ class ArrowAdapter {
         const Dimension& dimension);
 
     /**
-     * @brief Create a an ArrowSchema from TileDB Attribute
+     * @brief Create an ArrowSchema from TileDB Attribute
      *
      * @return ArrowSchema
      */
--- a/libtiledbsoma/src/utils/arrow_adapter.h
+++ b/libtiledbsoma/src/utils/arrow_adapter.h
@@ -230,6 +230,24 @@ class ArrowAdapter {
         std::shared_ptr<Context> ctx, std::shared_ptr<Array> tiledb_array);
 
     /**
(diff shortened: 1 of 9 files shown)


libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -229,6 +229,24 @@ class ArrowAdapter {
...
+
+    /**
+     * @brief Create a an ArrowSchema from TileDB Attribute
+     *
+     * @return ArrowSchema
+     */
+    static std::unique_ptr<ArrowSchema> arrow_schema_from_tiledb_attribute(
+        Attribute& attribute,
+        std::shared_ptr<Context> ctx,
+        std::shared_ptr<Array> tiledb_array);
```
jp-dark, line 248: Looks like you can change the implementation so these are passed as const references instead of pointers without too much pain.

jp-dark (approved):

jp-dark (approved):

johnkerl (changes requested):
Please include in this PR's description some copy/paste of:

* Example use of how the user might keystroke out the `index_column_names` and `domain` at `create`
* What `domain` might look like when they read it back

Also:

* Please file a follow-on PR to update https://github.com/single-cell-data/TileDB-SOMA/tree/main/apis/python/notebooks

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -229,6 +229,22 @@ class ArrowAdapter {
     static std::unique_ptr<ArrowSchema> arrow_schema_from_tiledb_array(
         std::shared_ptr<Context> ctx, std::shared_ptr<Array> tiledb_array);
 
+    /**
+     * @brief Create a an ArrowSchema from TileDB Dimension
```
johnkerl, line 233:
```suggestion
     * @brief Create an ArrowSchema from TileDB Dimension
```

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -229,6 +229,22 @@ class ArrowAdapter {
...
+     *
+     * @return ArrowSchema
+     */
+    static std::unique_ptr<ArrowSchema> arrow_schema_from_tiledb_dimension(
+        const Dimension& dimension);
+
+    /**
+     * @brief Create a an ArrowSchema from TileDB Attribute
```
johnkerl, line 241:
```suggestion
     * @brief Create an ArrowSchema from TileDB Attribute
```

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -229,6 +229,22 @@ class ArrowAdapter {
     static std::unique_ptr<ArrowSchema> arrow_schema_from_tiledb_array(
         std::shared_ptr<Context> ctx, std::shared_ptr<Array> tiledb_array);
 
+    /**
+     * @brief Create an ArrowSchema from TileDB Dimension
+     *
+     * @return ArrowSchema
+     */
+    static std::unique_ptr<ArrowSchema> arrow_schema_from_tiledb_dimension(
```
johnkerl, line 237: I wish you had made these helpers as a separate PR. As an ongoing ask I am continuing to ask for split-out PRs.

This PR is only 300 lines, which isn't a lot, but, all the codecov warnings are making this hard to read -- as if it _had_ been a high-line-count PR.

johnkerl (changes requested):

libtiledbsoma/src/utils/arrow_adapter.h
johnkerl: I ask you to split this not due to high line count but because you've got two different things going on. Pulling out helper functions is distinct from the goal of the PR (hiding internal dimensions) and is worth a separate PR. Then you can stack the dimension-hiding on top of it as a second PR.

johnkerl: Specifically, please split out a PR with only the `arrow_adapter.h` and `arrow_adapter.cc` mods. That, I will understand quickly, and will be able to review quickly.

XanthosXanthopoulos:
Replaced by #3427

PR #3262, opened 2024-10-30
[python] Update the `MultiscaleImage` to match `somacore` API changes 

jp-dark:
This is the companion TileDB-SOMA changes for single-cell-data/SOMA#238. It consists of major changes to the `MultiscaleImage` implementation.

Do NOT merge before single-cell-data/SOMA#238.

PR #3274, opened 2024-11-01
[python] Update spatial API to match somacore 1.0.23

jp-dark:
Update the TileDB-SOMA implementation to support the API changes in somacore 1.0.23.

This is includes:
* `MultiscaleImage` changes already approved in #3262 
* Update the `subcollection` parameters in the `Scene` class


johnkerl (approved):
🚢

Issue #3053, opened 2024-09-24
[python/c++] The `read` method does an `open` on an already open array


johnkerl:
We need to factor `ManagedQuery` out of the `SOMAArrray.

[[sc-54744]](https://app.shortcut.com/tiledb-inc/story/54744/python-soma-read-does-an-open-on-an-already-open-array)

PR #3275, opened 2024-11-01
[c++] Move casting from `SOMAArray` to `ManagedQuery`

nguyenv:
**Issue and/or context:**

**Changes:**

The large line count is from moving the Arrow column casting from `SOMAArray` to `ManagedQuery`. The enumeration methods have also been removed from the public `SOMAArray` API as they are completely unnecessary and currently only used as helpers internally. Otherwise, the actual contents of the code remain untouched.


johnkerl (approved):
Easy accept -- thank you @nguyenv ! :)

PR #3276, opened 2024-11-01
Expose `shape` accessor for `DataFrame` [RFC]

johnkerl:

**Changes:**

**Notes for Reviewer:**

This is an RFC.


johnkerl:
I'm abandoning this for now.

What I hoped: users accustomed to saying `adata.obs.shape` should be able to say `exp.obs.shape` and it's weird if they can't.

A balancing problem: if not all `soma_joinid` values in the `domain` are written to, then when we produce a Pandas DataFrame out of the SOMA DataFrame, its `pdf.shape != exp.obs.shape`.

I made this proposal as a naïve, well-intentioned, and very non-comp-bio engineer -- but we need to let users make this call, not me.

PR #3277, opened 2024-11-01
[c++] Use schema from `tiledb::Array` instead of `ManagedQuery`

nguyenv:
**Issue and/or context:**

**Changes:**

This introduces a `schema_` field to the `SOMAArray`, replacing any `mq->schema()` calls with `schema_` directly.

### DIFF
--- a/libtiledbsoma/src/soma/soma_array.h
+++ b/libtiledbsoma/src/soma/soma_array.h
@@ -1567,7 +1567,7 @@ class SOMAArray : public SOMAObject {
     // Array associated with mq_
     std::shared_ptr<Array> arr_;
 
-    // The TileDB ArraySchema. The schema is inaccessible when the TileDB Array 
+    // The TileDB ArraySchema. The schema is inaccessible when the TileDB Array
     // is closed or opened in write mode which means we cannot use arr->schema()
     // directly in those cases. Here, we store a copy of the schema so that it
     // can be accessed in any mode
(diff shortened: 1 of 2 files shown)


libtiledbsoma/src/soma/soma_array.h
```diff
@@ -1587,6 +1567,9 @@ class SOMAArray : public SOMAObject {
     // Array associated with mq_
     std::shared_ptr<Array> arr_;
 
+    // Array schema
```
johnkerl, line 1570: Does this open up a possibility for `arr_` and `schema_` to be out of sync?

Sorry & maybe I missed something but when would we ever want `schema_` to be anything other than `arr_->schema()`? And if the answer is "never", then can we simply refer to `arr_->schema()` rather than `schema_`?

nguyenv: I forgot to comment about this. If the array is not open or opened in write mode, then we can't access `arr_->schema()` directly.

nguyenv: You're right though that this could open up a possibility for the schema in `arr_` and `schema_` to be out of sync since we utilize schema evolutions. I could remove this `schema_` member and then modify the `schema()` accessor to use `arr_->schema()` if already open for writes and then open the array to grab the schema as necessary otherwise?

johnkerl: Ooh nice! ✨

johnkerl: But @nguyenv  if you're saying the answer to my question is 'we need a copy of the schema that was obtained when the array was opened for read, even when the array is currently opened for write and so the schema is otherwise inaccessible" -- then that's a great answer, and we can keep separate `arr_` and `schema_` as you have now. Just please add this information as a code comment in `managed_query.h` saying this.

johnkerl (approved):
💫

PR #3278, opened 2024-11-01
[c++] Add `ManagedQuery::set_layout`

nguyenv:
**Issue and/or context:**

**Changes:**

This is some minor clean up that replaces the switch-case setting in `SOMAArray` into a separate `ManagedQuery::set_layout`.

### DIFF
--- a/libtiledbsoma/src/soma/managed_query.cc
+++ b/libtiledbsoma/src/soma/managed_query.cc
@@ -73,6 +73,27 @@ void ManagedQuery::reset() {
     query_submitted_ = false;
 }
 
+void ManagedQuery::set_layout(ResultOrder layout) {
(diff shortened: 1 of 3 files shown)


libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -73,6 +73,27 @@ void ManagedQuery::reset() {
     query_submitted_ = false;
 }
 
+void ManagedQuery::set_layout(ResultOrder result_order) {
```
johnkerl, line 76: LGTM! For bonus you could have this method return `this` to allow for some fluent-notation at callsites (which may not be useful, just a thought)

johnkerl (approved):

PR #3281, opened 2024-11-02
[python] Add `ManagedQuery` bindings

nguyenv:
**Issue and/or context:**

**Changes:**

Adds `ManagedQuery` bindings to Pybind11 code.


nguyenv:
Yes when I replace `SOMAArray.write` with `ManagedQuery`, I will remove those from `soma_array.cc`. That's something planned for the upcoming GOW refactor.

PR #3292, opened 2024-11-04
[python] Consolidate all `_set_coords` methods into single `_util` function

nguyenv:
**Issue and/or context:**

**Changes:**

This is clean-up work.

This takes the separate implementations of `_set_coords` from `_dataframe`, `_sparse_nd_array`,  `_spatial_dataframe`, and `_soma_array` into a single `_util._set_coords`. `_cast_domainish` was also moved out of `SOMAArrayWrapper` and into `_util` to support casting of domains and NEDs in `_set_coord`.

**Note for reviewers:**

I am unsure if `_util.py` is the appropriate place for this.

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -14,7 +14,6 @@ from typing import (
     Optional,
     Sequence,
     Tuple,
-    Type,
     Union,
     cast,
 )
@@ -39,9 +38,7 @@ from ._types import (
     NPFloating,
     NPInteger,
     OpenTimestamp,
-    Slice,
     StatusAndReason,
-    is_slice_of,
 )
 from .options import SOMATileDBContext
 from .options._soma_tiledb_context import _validate_soma_tiledb_context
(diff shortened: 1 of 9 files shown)


apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -696,7 +693,7 @@ def read(
         if value_filter is not None:
             sr.set_condition(QueryCondition(value_filter), handle.schema)
 
-        self._set_coords(sr, coords)
+        _util._set_coords(sr, coords)
```
johnkerl, line 696: Maybe not on this PR (or maybe on this PR!) can we rename `sr` (which I think used to stand for _soma reader_) into something more reflective of what it currently is? Maybe as simple as `sarr` ......

johnkerl (approved):
This is beautiful ... thank you!!

johnkerl:
> I am unsure if `_util.py` is the appropriate place for this.

@nguyenv I think this is fine :)

apis/python/src/tiledbsoma/_dataframe.py
nguyenv: Yeah. I'll do that on this PR.

Issue #1175, opened 2023-03-27
[c++] Placeholder for `batch_size` and `partitions`


johnkerl:
Needing implementation in C++; then Python and R.

johnkerl:
Along with Behnam -- ?

jp-dark:
Closing issue as stale.

PR #3293, opened 2024-11-04
[python] Remove double open For `SOMAArray` reads

nguyenv:
**Issue and/or context:**

This finishes up https://github.com/single-cell-data/TileDB-SOMA/issues/3053 [[sc-54744]](https://app.shortcut.com/tiledb-inc/story/54744/python-soma-read-does-an-open-on-an-already-open-array)

**Changes:**

- Remove `clib.SOMAArray.open` in all `SOMAArray` derived class `read` calls
- Pass query arguments coords, column names, result order, value filter, and platform config that were formerly passed to `clib.SOMAArray.open` or `clib.SOMAArray.reset` instead to `TableReadIter` and `SparseNDArrayRead`
- `_set_coords` on the `ManagedQuery` rather than through `SOMAArray`
- `_arrow_table_reader` and `SparseTensorReadIterBase` initialize a `ManagedQuery` object and set the query arguments listed above; set up the read; and submit reads and return results until the query is complete

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -28,7 +28,6 @@ from . import _arrow_types, _util
 from . import pytiledbsoma as clib
 from ._constants import SOMA_JOINID
 from ._exception import SOMAError, map_exception_for_create
-from ._query_condition import QueryCondition
 from ._read_iters import TableReadIter
 from ._soma_array import SOMAArray
 from ._tdb_handles import DataFrameWrapper
@@ -722,30 +721,15 @@ class DataFrame(SOMAArray, somacore.DataFrame):
         _util.check_unpartitioned(partitions)
         self._check_open_read()
 
-        handle = self._handle._handle
-
-        context = handle.context()
-        if platform_config is not None:
-            config = context.tiledb_config.copy()
-            config.update(platform_config)
-            context = clib.SOMAContext(config)
-
-        sr = clib.SOMADataFrame.open(
-            uri=handle.uri,
-            mode=clib.OpenMode.read,
-            context=context,
-            column_names=column_names or [],
-            result_order=_util.to_clib_result_order(result_order),
-            timestamp=handle.timestamp and (0, handle.timestamp),
-        )
-
-        if value_filter is not None:
-            sr.set_condition(QueryCondition(value_filter), handle.schema)
-
-        _util._set_coords(sr, coords)
-
         # TODO: batch_size
-        return TableReadIter(sr)
+        return TableReadIter(
+            array=self,
+            coords=coords,
+            column_names=column_names,
+            result_order=_util.to_clib_result_order(result_order),
+            value_filter=value_filter,
+            platform_config=platform_config,
+        )
 
     def write(
         self, values: pa.Table, platform_config: Optional[options.PlatformConfig] = None
--- a/apis/python/src/tiledbsoma/_dense_nd_array.py
+++ b/apis/python/src/tiledbsoma/_dense_nd_array.py
@@ -289,7 +289,7 @@ class DenseNDArray(NDArray, somacore.DenseNDArray):
         """
         _util.check_type("values", values, (pa.Tensor,))
 
-        sr = self._handle._handle
+        clib_handle = self._handle._handle
 
         # Compute the coordinates for the dense array.
         new_coords: List[Union[int, Slice[int], None]] = []
@@ -311,15 +311,15 @@ class DenseNDArray(NDArray, somacore.DenseNDArray):
                 input = np.ascontiguousarray(input)
             order = clib.ResultOrder.rowmajor
 
-        mq = clib.ManagedQuery(sr, sr.context())
+        mq = clib.ManagedQuery(clib_handle, clib_handle.context())
         mq.set_layout(order)
-        _util._set_coords(mq, sr, new_coords)
+        _util._set_coords(mq, clib_handle, new_coords)
         mq.set_soma_data(input)
         mq.submit_write()
 
         tiledb_write_options = TileDBWriteOptions.from_platform_config(platform_config)
         if tiledb_write_options.consolidate_and_vacuum:
-            sr.consolidate_and_vacuum()
+            clib_handle.consolidate_and_vacuum()
         return self
 
     @classmethod
--- a/apis/python/src/tiledbsoma/_dense_nd_array.py
+++ b/apis/python/src/tiledbsoma/_dense_nd_array.py
@@ -20,6 +20,7 @@ from . import pytiledbsoma as clib
 from ._arrow_types import pyarrow_to_carrow_type
 from ._common_nd_array import NDArray
 from ._exception import SOMAError, map_exception_for_create
+from ._read_iters import TableReadIter
 from ._tdb_handles import DenseNDArrayWrapper
 from ._types import OpenTimestamp, Slice
 from ._util import dense_indices_to_shape
@@ -232,42 +233,20 @@ class DenseNDArray(NDArray, somacore.DenseNDArray):
         data_shape = tuple(handle.shape if use_shape else ned)
         target_shape = dense_indices_to_shape(coords, data_shape, result_order)
 
-        context = handle.context()
-        if platform_config is not None:
-            config = context.tiledb_config.copy()
-            config.update(platform_config)
-            context = clib.SOMAContext(config)
-
-        sr = clib.SOMADenseNDArray.open(
-            uri=handle.uri,
-            mode=clib.OpenMode.read,
-            context=context,
+        arrow_table = TableReadIter(
+            array=self,
+            coords=coords,
             column_names=[],
             result_order=_util.to_clib_result_order(result_order),
-            timestamp=handle.timestamp and (0, handle.timestamp),
-        )
+            value_filter=None,
+            platform_config=platform_config,
+        ).concat()
 
-        _util._set_coords(sr, coords)
-
-        arrow_tables = []
-        while True:
-            arrow_table_piece = sr.read_next()
-            if not arrow_table_piece:
-                break
-            arrow_tables.append(arrow_table_piece)
-
-        # For dense arrays there is no zero-output case: attempting to make a test case
-        # to do that, say by indexing a 10x20 array by positions 888 and 999, results
-        # in read-time errors of the form
-        #
-        # [TileDB::Subarray] Error: Cannot add range to dimension 'soma_dim_0'; Range [888, 888] is
-        # out of domain bounds [0, 9]
-        if not arrow_tables:
+        if arrow_table is None:
             raise SOMAError(
                 "internal error: at least one table-piece should have been returned"
             )
 
-        arrow_table = pa.concat_tables(arrow_tables)
         npval = arrow_table.column("soma_data").to_numpy()
         # TODO: as currently coded we're looking at the non-empty domain upper
         # bound but not its lower bound. That works fine if data are written at
@@ -310,7 +289,7 @@ class DenseNDArray(NDArray, somacore.DenseNDArray):
         """
         _util.check_type("values", values, (pa.Tensor,))
 
-        clib_dense_array = self._handle._handle
+        sr = self._handle._handle
 
         # Compute the coordinates for the dense array.
         new_coords: List[Union[int, Slice[int], None]] = []
@@ -331,13 +310,16 @@ class DenseNDArray(NDArray, somacore.DenseNDArray):
             if not input.flags.contiguous:
                 input = np.ascontiguousarray(input)
             order = clib.ResultOrder.rowmajor
-        clib_dense_array.reset(result_order=order)
-        _util._set_coords(clib_dense_array, new_coords)
-        clib_dense_array.write(input)
+
+        mq = clib.ManagedQuery(sr, sr.context())
+        mq.set_layout(order)
+        _util._set_coords(mq, sr, new_coords)
+        mq.set_soma_data(input)
+        mq.submit_write()
 
         tiledb_write_options = TileDBWriteOptions.from_platform_config(platform_config)
         if tiledb_write_options.consolidate_and_vacuum:
-            clib_dense_array.consolidate_and_vacuum()
+            sr.consolidate_and_vacuum()
         return self
 
     @classmethod
--- a/apis/python/src/tiledbsoma/_point_cloud_dataframe.py
+++ b/apis/python/src/tiledbsoma/_point_cloud_dataframe.py
@@ -28,7 +28,6 @@ from ._dataframe import (
     _revise_domain_for_extent,
 )
 from ._exception import SOMAError, map_exception_for_create
-from ._query_condition import QueryCondition
 from ._read_iters import TableReadIter
 from ._spatial_dataframe import SpatialDataFrame
 from ._spatial_util import (
@@ -332,30 +331,15 @@ class PointCloudDataFrame(SpatialDataFrame, somacore.PointCloudDataFrame):
         _util.check_unpartitioned(partitions)
         self._check_open_read()
 
-        handle = self._handle._handle
-
-        context = handle.context()
-        if platform_config is not None:
-            config = context.tiledb_config.copy()
-            config.update(platform_config)
-            context = clib.SOMAContext(config)
-
-        sr = clib.SOMAPointCloudDataFrame.open(
-            uri=handle.uri,
-            mode=clib.OpenMode.read,
-            context=context,
-            column_names=column_names or [],
+        # TODO: batch_size
+        return TableReadIter(
+            array=self,
+            coords=coords,
+            column_names=column_names,
             result_order=_util.to_clib_result_order(result_order),
-            timestamp=handle.timestamp and (0, handle.timestamp),
+            value_filter=value_filter,
+            platform_config=platform_config,
         )
-
-        if value_filter is not None:
-            sr.set_condition(QueryCondition(value_filter), handle.schema)
-
-        _util._set_coords(sr, coords)
-
-        # # TODO: batch_size
-        return TableReadIter(sr)
 
     def read_spatial_region(
         self,
--- a/apis/python/src/tiledbsoma/_read_iters.py
+++ b/apis/python/src/tiledbsoma/_read_iters.py
@@ -368,10 +368,10 @@ class BlockwiseScipyReadIter(BlockwiseReadIterBase[BlockwiseScipyReadIterResult]
 
         if self.compress and self.major_axis in self.reindex_disable_on_axis:
             raise SOMAError(
-                "Unable to disable reindexing of coordinates on CSC/Cclib_handle major axis"
+                "Unable to disable reindexing of coordinates on CSC/CSR major axis"
             )
 
-        # Sanity check: if CSC/Cclib_handle, we _must_ be reindexing
+        # Sanity check: if CSC/CSR, we _must_ be reindexing
         assert not self.compress or self.axes_to_reindex
 
     @property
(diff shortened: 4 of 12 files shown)


apis/python/src/tiledbsoma/_dense_nd_array.py
```diff
@@ -322,7 +301,7 @@ def write(
         """
         _util.check_type("values", values, (pa.Tensor,))
 
-        clib_dense_array = self._handle._handle
+        sr = self._handle._handle
```
johnkerl, line 304:
```suggestion
        clib_handle = self._handle._handle
```

apis/python/src/tiledbsoma/_dense_nd_array.py
```diff
@@ -343,13 +322,16 @@ def write(
...
+        mq = clib.ManagedQuery(sr, sr.context())
+        mq.set_layout(order)
+        _util._set_coords(mq, sr, new_coords)
+        mq.set_soma_data(input)
+        mq.submit_write()
 
         tiledb_write_options = TileDBWriteOptions.from_platform_config(platform_config)
         if tiledb_write_options.consolidate_and_vacuum:
-            clib_dense_array.consolidate_and_vacuum()
+            sr.consolidate_and_vacuum()
```
johnkerl, line 334:
```suggestion
            clib_handle.consolidate_and_vacuum()
```

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -63,8 +64,18 @@
 class TableReadIter(somacore.ReadIter[pa.Table]):
     """Iterator over `Arrow Table <https://arrow.apache.org/docs/python/generated/pyarrow.Table.html>`_ elements"""
 
-    def __init__(self, sr: clib.SOMAArray):
-        self._reader = _arrow_table_reader(sr)
+    def __init__(
+        self,
+        array: SOMAArray,
+        coords: options.SparseDFCoords,
```
johnkerl, line 70: It's a little weird that on the one hand this takes a fully generic `SOMAArray`, but on the other hand, the coords are `options.Sparse_DFCoords` ... 🤔

johnkerl: Maybe this makes total sense ... the types that can be passed to a `DenseNDArray` or `SparseNDArray` read are a strict subset of the types that can be passed to a `DataFrame` read ... maybe just add a comment saying that: this covers the union of all types that are legitimate for `coords` ...

johnkerl (changes requested):
This is really really nice --a well-thought-out refactor, well executed. Just some minor mods requested. :)

apis/python/src/tiledbsoma/_read_iters.py
nguyenv: Good catch, ty. I've corrected this to `Union[options.SparseDFCoords, options.DenseNDCoords]`.

johnkerl (approved):
🚢

apis/python/src/tiledbsoma/_read_iters.py
johnkerl: Union also with `SparseNDCoords`?

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -418,28 +465,62 @@ def _cs_reader(
...
+        self.array = array
+        self.coords = coords
         self.shape = shape
+        self.result_order = result_order
+        self.platform_config = platform_config
+
+        sr = array._handle._handle
+
+        if platform_config is not None:
+            ctx = clib.SOMAContext(platform_config)
```
bkmartinjr, line 485: If I am understanding this correctly, if I have already opened the array with a context, and I specify a read with a `platform_config`, it creates a new context and ignores the one previously specified in the `open`.

Are these semantics correct?  Vs. (say) modifying the existing context config?

I'm actually not sure what is correct, as I am ignorant on the use case for passing a `platform_config` into a read (I always set up my contexts in the `open` call).

bkmartinjr: I looked at the old code in _dataframe.py, and it was taking  pains to keep the config from the context specified in open:
```
        if platform_config is not None:
            config = context.tiledb_config.copy()
            config.update(platform_config)
            context = clib.SOMAContext(config)
```

So we definitely are changing the behavior, for at least Dataframe

nguyenv: Thanks for catching this. I'll take the `sr.context().config()` and then `update` that with the user's passed in config tomrrow.

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -457,12 +538,44 @@ def _from_table(self, arrow_table: pa.Table) -> pa.SparseCOOTensor:
...
+) -> Iterator[pa.Table]:
     """Private. Simple Table iterator on any Array"""
-    tbl = sr.read_next()
-    while tbl is not None:
-        yield tbl
-        tbl = sr.read_next()
+
+    sr = array._handle._handle
+
+    if platform_config is not None:
```
bkmartinjr, line 555: see above comment re: context handling in the presence of a platform_config

apis/python/src/tiledbsoma/_read_iters.py
bkmartinjr: Suggestion - seems like our test coverage missed this, so may want to add at least a quick/simple test to make sure that the correct behavior is codified in a unit test.

bkmartinjr (changes requested):
This looks great.  Two small items:
1. platform_config handling - noted inline
2. ManagedQuery GIL - the read path moves some logic up from C++ to Python -- we need to make sure the pybind entry points are all releasing GIL where appropriate.  My quick skim is that it looks OK, but I suggest we benchmark after landing this to confirm.

bkmartinjr (approved):

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -465,12 +549,88 @@ def _from_table(self, arrow_table: pa.Table) -> pa.SparseCOOTensor:
...
+        self.mq.setup_read()
+
+    def __iter__(self) -> Self:
+        return self
+
+    def __next__(self) -> pa.Table:
+        return self.mq.next()
+
+
+# def _arrow_table_reader(
```
bkmartinjr, line 594: needs cleanup.

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -465,12 +549,46 @@ def _from_table(self, arrow_table: pa.Table) -> pa.SparseCOOTensor:
         return pa.SparseCOOTensor.from_numpy(coo_data, coo_coords, shape=self.shape)
 
 
-def _arrow_table_reader(sr: clib.SOMAArray) -> Iterator[pa.Table]:
-    """Private. Simple Table iterator on any Array"""
-    tbl = sr.read_next()
-    while tbl is not None:
-        yield tbl
-        tbl = sr.read_next()
+class ArrowTableRead:
```
bkmartinjr, line 552: this class should sub-class the Iterable ABC, e.g., define it as:

> class ArrowTableRead(Iterator[pa.Table]):

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -465,12 +549,46 @@ def _from_table(self, arrow_table: pa.Table) -> pa.SparseCOOTensor:
...
+            self.mq.select_columns(list(column_names))
+
+        if value_filter is not None:
+            self.mq.set_condition(QueryCondition(value_filter), sr.schema)
+
+        _util._set_coords(self.mq, sr, coords)
+
+        self.mq.setup_read()
+
+    def __iter__(self) -> Self:
```
bkmartinjr, line 587: Iterators don't need an __iter__ method AFAIK.  Iterables do, but iterators do not.  So you should be able to remove this.

bkmartinjr (approved):
Couple of minor pythonic / lint items, but otherwise LGTM.

nguyenv:
Thanks to @bkmartinjr for helping identify the source of the segfault. The issue stemmed from the new iterator class calling `setup_read` only once during initialization to allocate buffers. As a result, data from previous reads was overwritten during each call to `next`. The fix ensures `setup_read` is called within `next`, allocating new buffers for each read to prevent data corruption.

nguyenv:
Sorry - hit to rereview too quickly. I need to make some additional cleanup...

nguyenv:
I believe my PR is back to a working state. It is failing these unit tests on 3.12, osx, which is already a known issue and unrelated to changes in this branch.
```
FAILED apis/python/tests/test_basic_xarray_io.py::TestDenseNDDataArray1D::test_getitem_with_steps[key0] - Failed: DID NOT RAISE <class 'NotImplementedError'>
FAILED apis/python/tests/test_basic_xarray_io.py::TestDenseNDDataArray3D::test_getitem_with_steps[key0] - Failed: DID NOT RAISE <class 'NotImplementedError'>
```

As stated internally, I would like to get this PR reviewed and merged as-is. I would like to do a follow-up PR that adds a public `ManagedQuery::read_next` method and moves the `setup_read`, `submit_read`, and `result` methods as private.

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -722,31 +721,16 @@ def read(
...
-            config = context.tiledb_config.copy()
-            config.update(platform_config)
-            context = clib.SOMAContext(config)
-
-        sr = clib.SOMADataFrame.open(
-            uri=handle.uri,
-            mode=clib.OpenMode.read,
-            context=context,
-            column_names=column_names or [],
+        # TODO: batch_size
```
johnkerl, line 724: Please file a GitHub issue for tracking this, and include its URL here as a comment

apis/python/src/tiledbsoma/_point_cloud_dataframe.py
```diff
@@ -332,31 +331,16 @@ def read(
...
-            config = context.tiledb_config.copy()
-            config.update(platform_config)
-            context = clib.SOMAContext(config)
-
-        sr = clib.SOMAPointCloudDataFrame.open(
-            uri=handle.uri,
-            mode=clib.OpenMode.read,
-            context=context,
-            column_names=column_names or [],
+        # TODO: batch_size
```
johnkerl, line 334: Ditto re above tracking link

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -66,8 +67,47 @@
...
+            ]):
+                for each index dimension, which rows to read.
+                ``()`` means no constraint -- all IDs.
+
+            column_names (Optional[Sequence[str]]):
+                The named columns to read and return.
+                ``None`` means no constraint -- all column names.
+
+            result_order (clib.ResultOrder):
+                Order of read results.
```
johnkerl, line 98: I was going to ask you to list out the valid options here -- unless ` clib.ResultOrder` is clickable in readthedocs, in which case never mind

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -230,20 +269,26 @@ def _maybe_eager_iterator(
...
-            _util._set_coords(self.sr, step_coords)
 
             joinids = list(self.joinids)
             joinids[self.major_axis] = pa.array(coord_chunk)
-            yield pa.concat_tables(_arrow_table_reader(self.sr)), tuple(joinids)
+            yield pa.concat_tables(
+                ArrowTableRead(
+                    array=self.array,
+                    coords=step_coords,
+                    column_names=[],
```
johnkerl, line 286: I'm not understanding why this is `[]`

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -426,28 +473,58 @@ def _cs_reader(
...
+        result_order: clib.ResultOrder,
+        platform_config: Optional[options.PlatformConfig],
+    ):
+        self.array = array
+        self.coords = coords
         self.shape = shape
+        self.result_order = result_order
+        self.platform_config = platform_config
+
+        sr = array._handle._handle
```
johnkerl, line 490:
```suggestion
        clib_handle = array._handle._handle
```
(the 'old name' `sr` is now confusing)

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -426,28 +473,58 @@ def _cs_reader(
...
 
         If some data has already been retrieved using ``next``, this will return
         the rest of the data after that is already returned.
         """
-        arrow_tables = pa.concat_tables(TableReadIter(self.sr))
+        arrow_tables = pa.concat_tables(
+            TableReadIter(
+                array=self.array,
+                coords=self.coords,
+                column_names=[],
```
johnkerl, line 522: Ditto re above: I don't understand why `[]` is correct here

apis/python/src/tiledbsoma/_sparse_nd_array.py
```diff
@@ -266,27 +266,45 @@ def read(
...
+        return SparseNDArrayRead(
+            array=self,
+            coords=coords,
             result_order=_util.to_clib_result_order(result_order),
-            timestamp=handle.timestamp and (0, handle.timestamp),
+            platform_config=platform_config,
         )
 
-        return SparseNDArrayRead(sr, self, coords)
+    def resize(
```
johnkerl, line 279: Weird that these are showing up new

... I know what happened though:

* We had support for current domain for sparse arrays in core 2.26
* We had support for current domain for dense arrays in core 2.27
* So I had these two methods in `apis/python/src/tiledbsoma/_sparse_nd_array.py` and `apis/python/src/tiledbsoma/_dense_nd_array.py`
* Now that we're on 2.27.0rc3 these don't need to be duplicated anymore
* On #3399 I pulled these down to `apis/python/src/tiledbsoma/_common_nd_array.py` (probably that was confusing of me to have done so on that PR, given that PR's title)

johnkerl: Anyway please remove these, and/or double-check your rebase -- #3399 is how things should be moving forward

apis/python/tests/test_dataframe.py
```diff
@@ -1899,3 +1899,47 @@ def test_bounds_on_somajoinid_domain(tmp_path):
...
+def test_pass_configs(tmp_path, arrow_schema):
+    uri = tmp_path.as_posix()
+
+    with soma.DataFrame.create(uri, schema=arrow_schema()) as sdf:
+        pydict = {}
+        pydict["soma_joinid"] = [0, 1, 2, 3, 4]
+        pydict["foo"] = [10, 20, 30, 40, 50]
+        pydict["bar"] = [4.1, 5.2, 6.3, 7.4, 8.5]
+        pydict["baz"] = ["apple", "ball", "cat", "dog", "egg"]
+        pydict["quux"] = [True, False, False, True, False]
```
johnkerl, line 1913:
```suggestion
        pydict["myint"] = [10, 20, 30, 40, 50]
        pydict["myfloat"] = [4.1, 5.2, 6.3, 7.4, 8.5]
        pydict["mystring"] = ["apple", "ball", "cat", "dog", "egg"]
        pydict["mybool"] = [True, False, False, True, False]
```

Feel free to disregard -- this is nitty on my part, & I can do the renames on a separate PR, and you've put a lot of blood, sweat, and tears into this PR already -- I don't want to make things harder for you by making this nitty request

apis/python/tests/test_dataframe.py
```diff
@@ -1899,3 +1899,47 @@ def test_bounds_on_somajoinid_domain(tmp_path):
...
+    # Pass a custom config to open
+    with soma.DataFrame.open(
+        uri,
+        "r",
+        context=soma.SOMATileDBContext(
+            {"sm.mem.total_budget": "0", "sm.io_concurrency_level": "0"}
+        ),
+    ) as sdf:
+
+        # This error out as 0 are not valid values to set the total memory
```
johnkerl, line 1927:
```suggestion
        # This errors out as 0 is not a valid value to set the total memory
```

apis/python/tests/test_dense_nd_array.py
```diff
@@ -498,3 +498,45 @@ def test_read_to_unwritten_array(tmp_path, shape):
...
+    # Pass a custom config to open
+    with soma.DenseNDArray.open(
+        uri,
+        "r",
+        context=soma.SOMATileDBContext(
+            {"sm.mem.total_budget": "0", "sm.io_concurrency_level": "0"}
+        ),
+    ) as sdf:
+
+        # This error out as 0 are not valid values to set the total memory
```
johnkerl, line 526:
```suggestion
        # This errors out as 0 is not a valid value to set the total memory
```

apis/python/tests/test_sparse_nd_array.py
```diff
@@ -1883,3 +1883,47 @@ def test_global_writes(tmp_path):
...
+    # Pass a custom config to open
+    with soma.SparseNDArray.open(
+        uri,
+        "r",
+        context=soma.SOMATileDBContext(
+            {"sm.mem.total_budget": "0", "sm.io_concurrency_level": "0"}
+        ),
+    ) as sdf:
+
+        # This error out as 0 are not valid values to set the total memory
```
johnkerl, line 1911:
```suggestion
        # This errors out as 0 is not a valid value to set the total memory
```

johnkerl (changes requested):

apis/python/src/tiledbsoma/_dataframe.py
nguyenv: https://github.com/single-cell-data/TileDB-SOMA/issues/1175

apis/python/src/tiledbsoma/_read_iters.py
nguyenv: This means we select all columns for reads. I will write an inline comment for clarification.

johnkerl (approved):
🔥 

Thank you for keeping at this @nguyenv ! :)

apis/python/src/tiledbsoma/managed_query.cc
```diff
@@ -133,6 +136,36 @@ void load_managed_query(py::module& m) {
...
 
+        .def(
+            "next",
+            [](ManagedQuery& mq) -> std::optional<py::object> {
+                // Release python GIL before reading data
+                py::gil_scoped_release release;
+
+                mq.setup_read();
+
+                if (mq.is_empty_query() && mq.is_first_read()) {
```
bkmartinjr, line 147: I think the code would be easier to understand if this code block was removed and empty queries went through the same code as all other queries.

All this would require is that `submit_read` do a check for an empty query, and return immediately if it is empty (i.e., move this check into submit_read). 

As part of that, you could remove the `query_submitted_ = true;` from the results method, which strikes me as an odd place to set this flag.

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -381,6 +381,7 @@ void ManagedQuery::_fill_in_subarrays_if_dense_with_new_shape(
 
 std::shared_ptr<ArrayBuffers> ManagedQuery::results() {
     if (is_empty_query()) {
+        query_submitted_ = true;
```
bkmartinjr, line 384: See my comment in `next()` -- this is an odd place to set this flag.  Makes the code hard to follow.

bkmartinjr (approved):
A small cleanup request:
1. the code in `ManagedQuery.next()` would be cleaner if submit_read just handles empty queries, and `next()` has a single code path for all queries.
2. Once that happens, you don't need to set the `query_submitted` flag in `results()` which is somewhat awkward.

Otherwise this looks good and I'm preemptively approving.  Thank you!

johnkerl (approved):
🚢 

Thank you @nguyenv !!!

bkmartinjr (approved):

PR #3298, opened 2024-11-06
[python] Add export for `PointCloudDataFrame` to `SpatialData` shapes

jp-dark:
Adds a new function that converts a `PointCloudDataFrame` to a `geopandas.GeoDataFrame` that is compatible with the SpatialData "shapes" model.


ivirshup:
Could you go in a little more detail on this one? I think the point clouds should export to spatial data points?

https://spatialdata.scverse.org/en/stable/generated/spatialdata._io.format.PointsFormatV01.html

jp-dark:
> Could you go in a little more detail on this one? I think the point clouds should export to spatial data points?
> 
> https://spatialdata.scverse.org/en/stable/generated/spatialdata._io.format.PointsFormatV01.html

I am planning on providing either option. My thinking is to default to using shapes if the `soma_geometry` metadata exists and points otherwise. The example you provided used the shapes for Census data, so I implemented this one first.

PR #3302, opened 2024-11-06
[r] Vignette for new-shape feature

johnkerl:
**Issue and/or context:** As tracked on issue #2407 / [[sc-51048]](https://app.shortcut.com/tiledb-inc/story/51048).

Note that the intended Python and R API changes are all agreed on and finalized as described in #2407.

**Changes:**

This is the R counterpart to #3294.

**Notes for Reviewer:**

There will be Academy material as well, in a separate PR.

PR #3303, opened 2024-11-07
[r] Data refresh for new shapes

johnkerl:

**Changes:**

* Regenerate data in the `inst/extdata` directory to have new-style shapes -- primarily for the benefit of vignettes on #3302.
* Add a missing script for regenerating one of the two datasets
* Add `README.md` for the benefit of future maintainers

**Notes for Reviewer:**


apis/r/tests/testthat/test-example-datasets.R
```diff
@@ -1,7 +1,7 @@
 test_that("example dataset access", {
   expect_length(
     list_datasets(),
-    length(dir(example_data_dir()))
+    length(dir(example_data_dir())) - 1 # There is a README.md
```
mojaveazure, line 4: I don't see a place in `README.md` that explains why this `- 1` exists

mojaveazure (changes requested):
So it seems the original `pbmc3k-processed-obs` was from Scanpy, not Seurat (based on the column name changes in `test-SOMAArrayReader-Arrow.R` and `test-SOMADataFrame.R`; I've got one comment about a missing `README` entry

apis/r/tests/testthat/test-example-datasets.R
johnkerl: The `README.md` file is an item within `example_data_dir`.

johnkerl: I can say that in the `README.md` file.

johnkerl: @mojaveazure https://github.com/single-cell-data/TileDB-SOMA/pull/3303/commits/b6c59b4c72142a6fd01f4110e08cee0ccf9a45b8

mojaveazure (approved):

PR #3304, opened 2024-11-07
[python/c++] COO to CSX conversion optimization

bkmartinjr:
**Issue and/or context:**

Using scipy.sparse.spmatrix for conversion from SOMA Arrow tables to SciPy csr_matrix is single-threaded and requires contiguous COO input arrays. This PR:
* does the conversion from COO -> CSC/CSR without requiring the extra copy to concat chunked arrays (roughly halving the memory used for this operation).
* parallelizes the conversion - reduces wall-clock time (at the cost of extra total CPU time)

All API/modules are for internal use. There is no external API surface change.

**Changes:**

* C++ package containing low-level operations. Includes vendoring "span" and "parallel_for" from TileDB repo
* Python module and pybind-ings containing "friendlier" API for internal python code
* Wired into `soma.io.to_anndata` and the `blockwise()` reader
* C++ compile flags for Python bindings: added `-O3 -mavx2` to setup.py
* Modify R/Python GHA workflow to match latest build practices and remove legacy steps

Later work will wire into additional code paths.


.github/workflows/r-python-interop-testing.yml
```diff
@@ -27,54 +27,19 @@ jobs:
...
-      # apis/R/DESCRIPTION file and (b) appearance of binaries.
-      #
-      # Please see https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases which
-      # is crucial for anyone doing releases of TileDB-SOMA.
-      #
-      # Please edit both files in the same way:
-      # * r-ci.yml
-      # * r-python-interop-testing.yml
-      #
-      # Do not remove these comments until such time as we have eliminated our dependency on
```
johnkerl, line 58: Please don't remove this comment block until we've removed our dependency on the TileDB-R package (which is till WIP).

johnkerl (commented):
I'm partway through but need to be AFK. Poking the submit button on where I am now, as a checkpoint.

apis/python/setup.py
```diff
@@ -323,6 +328,7 @@ def run(self):
             extra_compile_args=["-std=c++17" if os.name != "nt" else "/std:c++17"]
             + CXX_FLAGS,
             language="c++",
+            # undef_macros=["NDEBUG"],  # enable assertions by undefing NDEBUG
```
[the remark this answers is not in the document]
bkmartinjr: will do!

Somewhat related:  I wish we built a version with asserts enabled, and ran unit tests against it.  There are a lot of good asserts in our C++ code, and we can't enable them in production builds, but they are useful in testing...

apis/python/src/tiledbsoma/_fastercsx.py
```diff
@@ -0,0 +1,269 @@
...
+        self.context = context
+
+    @staticmethod
+    def from_ijd(
+        i: NDArrayIndex | Sequence[NDArrayIndex],
+        j: NDArrayIndex | Sequence[NDArrayIndex],
+        d: NDArrayNumber | Sequence[NDArrayNumber],
+        shape: Tuple[int, int],
+        format: Literal["csc", "csr"],
+        sorted: bool,
```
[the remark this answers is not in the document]
bkmartinjr: I was following the convention you see with SciPy (e.g., the `format` argument, not the `make_format` argument).

I have no real opinion, as this is internal code -- not public API.  And I will make the change. But I note that your suggestion mixes styles in the same function signature.

apis/python/src/tiledbsoma/_fastercsx.py
```diff
@@ -0,0 +1,269 @@
...
+        return int(self.indptr.nbytes + self.indices.nbytes + self.data.nbytes)
+
+    @property
+    def dtype(self) -> npt.DTypeLike:
+        return self.data.dtype
+
+    def to_scipy(
+        self, index: slice | None = None
+    ) -> scipy.sparse.csr_matrix | scipy.sparse.csc_matrix:
+        """Extract a slice on the compressed dimension and return as a dense
```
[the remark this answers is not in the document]
bkmartinjr: given that these are private implementation classes, the comments should never make it into RTD, right?

.github/workflows/r-python-interop-testing.yml
bkmartinjr: added back

apis/python/src/tiledbsoma/_fastercsx.py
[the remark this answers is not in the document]
johnkerl: Thanks @bkmartinjr !

apis/python/src/tiledbsoma/fastercsx.cc
```diff
@@ -0,0 +1,621 @@
...
+    }
+    return vec;
+}
+
+/**
+ * @brief Various helpers which convert from py:array to spans.
+ */
+
+template <typename T>
+tcb::span<T const> make_span_(py::array arr) {
```
johnkerl, line 78: Can this be `const py::array& arr`? (Likewise with functions after this)

apis/python/src/tiledbsoma/fastercsx.cc
```diff
@@ -0,0 +1,621 @@
...
+        if (!Ai[0].dtype().is(Aj[0].dtype()))
+            throw pybind11::type_error(
+                "COO index arrays must have same dtype.");
+    }
+}
+
+/**
+ * @brief python binding for coo_compress, which converts COO chunked array (I,
+ * J, D) into CSX arrays (P, J, D). The output arrays are NOT sorted on the
+ * minor dimension - see `sort_indices` if that is required.
```
johnkerl, line 301:
```suggestion
 * minor dimension -- see `sort_indices` if that is required.
```

apis/python/src/tiledbsoma/fastercsx.cc
```diff
@@ -0,0 +1,621 @@
...
+                Bd_view);
+        },
+        value_type,
+        csx_major_index_type,
+        csx_minor_index_type,
+        coo_index_type);
+}
+
+/**
+ * @brief python binding for sort_indices, which sorts minor dimension of a
```
johnkerl, line 385:
```suggestion
 * @brief Python binding for sort_indices, which sorts minor dimension of a
```

apis/python/src/tiledbsoma/fastercsx.cc
```diff
@@ -0,0 +1,621 @@
...
+    std::shared_ptr<tiledbsoma::SOMAContext> ctx,
+    py::array Bp,
+    py::array Bj,
+    py::array Bd) {
+    // Error checks first
+    //
+    if (Bp.ndim() != 1 || Bj.ndim() != 1 || Bd.ndim() != 1)
+        throw std::length_error("All arrays must be 1D");
+    if (!Bp.writeable() || !Bj.writeable() || !Bd.writeable())
+        throw std::invalid_argument("Output arrays must be writable.");
```
johnkerl, line 398:
```suggestion
        throw std::invalid_argument("Output arrays must be writeable.");
```

apis/python/src/tiledbsoma/fastercsx.cc
```diff
@@ -0,0 +1,621 @@
...
+        },
+        value_type,
+        csx_major_index_type,
+        csx_minor_index_type);
+};
+
+/**
+ * @brief Python binding for parallel slice+to_dense operation.
+ */
+void copy_to_dense(
```
johnkerl, line 446:
```suggestion
void copy_csx_to_dense(
```

apis/python/src/tiledbsoma/fastercsx.cc
```diff
@@ -0,0 +1,621 @@
...
+    auto [n_row, n_col] = shape;
+    if (n_row < 0 || n_col < 0)
+        throw std::length_error("n_row and n_col must be >= 0");
+    auto n_major = (format == "csr") ? n_row : n_col;
+
+    if (major_idx_start < 0 || major_idx_end > n_major)
+        throw std::range_error(
+            "row_start must be >= 0 and row_end < array shape");
+    if (n_major != Bp.size() - 1)
+        throw std::length_error("n_rows does not match Bp.size()");
```
johnkerl, line 470:
```suggestion
        throw std::length_error(fmt::format("n_rows {} does not match Bp.size() {}", n_major, BP.size()-1));
```

(just a thought)

apis/python/src/tiledbsoma/fastercsx.cc
```diff
@@ -0,0 +1,621 @@
...
+    if (!out.writeable())
+        throw std::invalid_argument("out must be writable");
```
johnkerl, line 472:
```suggestion
        throw std::invalid_argument("out must be writeable");
```

apis/python/src/tiledbsoma/fastercsx.cc
```diff
@@ -0,0 +1,621 @@
...
+        throw pybind11::type_error(
+            std::string("All COO data must be tuple of ndarray (") + e.what() +
+            ")");
+    }
+
+    if (Bp.ndim() != 1)
+        throw std::range_error("All arrays must be of dimension rank 1.");
+
+    if (Bp.size() != (n_row + 1))
+        throw std::length_error("Pointer array size does not match n_rows.");
```
johnkerl, line 538: Another `fmt::format` opportunity to give the caller more information about what they did wrong

apis/python/src/tiledbsoma/fastercsx.cc
```diff
@@ -0,0 +1,621 @@
...
+
+    if (!Bp.writeable())
+        throw std::invalid_argument("Output arrays must be writable.");
```
johnkerl, line 541:
```suggestion
        throw std::invalid_argument("Output arrays must be writeable.");
```

apis/python/src/tiledbsoma/io/conversions.py
```diff
@@ -78,10 +80,11 @@ def to_tiledb_supported_array_type(name: str, x: _MT) -> _MT:
     return x
 
 
-def csr_from_coo_table(tbl: pa.Table, num_rows: int, num_cols: int) -> sp.csr_matrix:
+def csr_from_coo_table(
+    tbl: pa.Table, num_rows: int, num_cols: int, context: SOMATileDBContext
+) -> sp.csr_matrix:
     """Given an Arrow Table containing COO data, return a ``scipy.sparse.csr_matrx``."""
```
johnkerl, line 86: `scipy.sparse.csr_matrix` (I know this typo isn't new on this PR but may as well fix it here/now)

apis/python/tests/test_fastercsx.py
```diff
@@ -0,0 +1,370 @@
...
+    ]:
+        with pytest.raises(IndexError):
+            # this will also log an error - which is expected behavior for parallel_for/thread_pool
+            fastercsx.CompressedMatrix.from_ijd(
+                sp.row, sp.col, sp.data, shp, "csr", True, context
+            )
+
+
+"""
+other tests to consider
```
johnkerl, line 367: Implement them, or remove this comment, and/or file a tracking issue for follow-up work?

apis/python/tests/test_libfastercsx.py
```diff
@@ -0,0 +1,313 @@
...
+    context: clib.SOMAContext,
+) -> None:
+
+    rng = np.random.default_rng()
+    sp = sparse.random(
+        shape[0], shape[1], density=0.01, dtype=value_dtype, random_state=rng
+    )
+
+    if sp.nnz >= np.iinfo(csr_major_index_dtype).max:
+        # only occur if we mess up the test params
```
johnkerl, line 71:
```suggestion
        # only occurs if we mess up the test params
```

apis/python/tests/test_libfastercsx.py
```diff
@@ -0,0 +1,313 @@
...
+    [
+        ((10000, 100), 0.1),
+        ((100_001, 83), 0.1),
+    ],
+)
+@pytest.mark.parametrize("nchunks", [1, 2, 3, 7, 10, 13])
+def test_multichunk(
+    nchunks: int, shape: tuple[int, int], density: float, context: clib.SOMAContext
+) -> None:
+    """check that multi-chunk COO input functions correctly"""
```
johnkerl, line 167:
```suggestion
    """Check that multi-chunk COO input functions correctly"""
```

libtiledbsoma/src/utils/fastercsx.h
```diff
@@ -0,0 +1,543 @@
...
+
+/**
+ * @brief Simple partitioning scheme - given a vector of spans, and a minimum
+ * partition size, create up to max_partitions.
+ *
+ * Used to split work across workers in a reasonably even manner. Currently a
+ * greedy n-way linear partition, which works well enough for the common case
+ * where most inputs are of similar sizes (which is typically the case).
+ *
+ * TODO: If we sort (descending) first, it will be LPT which would be better in
```
johnkerl, line 91: Please spell out what LPT means to you

libtiledbsoma/src/utils/fastercsx.h
```diff
@@ -0,0 +1,543 @@
...
+            it = partitions.erase(it);
+        else
+            ++it;
+    }
+
+    return partitions;
+}
+
+/**
+ * @brief Count occurrences of all values present in Ai - used as the basis
```
johnkerl, line 143:
```suggestion
 * @brief Count occurrences of all values present in Ai -- used as the basis
```

libtiledbsoma/src/utils/fastercsx.h
```diff
@@ -0,0 +1,543 @@
...
+    auto [n_row, n_col] = shape;
+    assert(Ai.size() == Aj.size() && Aj.size() == Ad.size());
+    assert(sum_over_size_(Ai) == nnz);
+    assert(sum_over_size_(Aj) == nnz);
+    assert(sum_over_size_(Ad) == nnz);
+    assert(Bp.size() == n_row + 1);
+    assert(Bj.size() == nnz);
+    assert(Bd.size() == nnz);
+
+    // get major axis counts. Important side effect: range checks the Ai
```
johnkerl, line 329:
```suggestion
    // Get major axis counts. Important side effect: range checks the Ai
```

libtiledbsoma/src/utils/fastercsx.h
```diff
@@ -0,0 +1,543 @@
...
+
+template <typename CSX_MINOR_IDX, typename VALUE>
+bool index_lt_(
+    const std::pair<CSX_MINOR_IDX, VALUE>& a,
+    const std::pair<CSX_MINOR_IDX, VALUE>& b) {
+    return a.first < b.first;
+}
+
+/**
+ * @brief Inplace sort of minor axis, used to canonicalize the CSx ordering.
```
johnkerl, line 415:
```suggestion
 * @brief In-place sort of minor axis, used to canonicalize the CSx ordering.
```

libtiledbsoma/src/utils/fastercsx.h
```diff
@@ -0,0 +1,543 @@
...
+                            "Out array is too small for provided "
+                            "coordinate "
+                            "range.");
+                    out[out_idx] = Bd[j];
+                }
+                return Status::Ok();
+            });
+        assert(status.ok());
+    } else {
+        // parallelizing this is less beneficial than the CSR case as it
```
johnkerl, line 505:
```suggestion
        // Parallelizing this is less beneficial than the CSR case as it
```

libtiledbsoma/src/utils/fastercsx.h
```diff
@@ -0,0 +1,543 @@
...
+        // partitions by columns, but writes in C order. This does not align
+        // with CPU cache lines, and is therefore quite a bit slower than
+        // the speedup achieved by the (above) CSR loop.
+        //
+        // We could write a Fortran-ordered output array, but then it would
```
johnkerl, line 510: I know (and you know) C order == row-major and Fortran order == col-major but it mightn't hurt to simply say row-major and col-major here in alignment with core conventions. Just a nit; just a thought.

libtiledbsoma/src/utils/parallel_functions.h
```diff
@@ -0,0 +1,406 @@
+/**
```
johnkerl, line 1: I believe this is simply a copy from core of known-good code -- if so I won't even peek at it

apis/python/src/tiledbsoma/_fastercsx.py
```diff
@@ -0,0 +1,269 @@
...
+
+    @staticmethod
+    def from_soma(
+        tables: pa.Table | Sequence[pa.Table],
+        shape: Tuple[int, int],
+        format: Literal["csc", "csr"],
+        make_sorted: bool,
+        context: SOMATileDBContext,
+    ) -> CompressedMatrix:
+        """Factory method accepting a sequence of Arrow tables containing SOMA sparse matrix data.
```
johnkerl, line 116: So, `from_arrow` maybe? Just a thought.

apis/python/src/tiledbsoma/fastercsx.cc
bkmartinjr: To what end? 

I believe it is simply a typed pointer, and all of the PyBind docs/example code use the above pattern.  I'd rather stick with the recommended pattern unless there is a benefit I'm unaware of...  Thoughts?

apis/python/src/tiledbsoma/fastercsx.cc
bkmartinjr: this is also an inconsistency if done to this one signature (why not also `sort_csx_indices`, etc?)

I will make the change, but don't love the introduction of inconsistent patterns of naming.

apis/python/src/tiledbsoma/fastercsx.cc
johnkerl: Thanks for the feedback that it's a typed pointer -- I did not grok that from my read -- thanks for the info!

apis/python/src/tiledbsoma/fastercsx.cc
johnkerl: As you wish -- this is code review and I'm offering suggestions from my point of view, for your consideration, which you're happy to take or leave.

My personal preference re

> why not also sort_csx_indices, etc?)

is to, yes, do so. (The function internally asserts that the input is CSR or CSC so, IMHO, putting that in function names seems like a sensible thing to do. And doing that in other similarly affected places IMO also makes sense.)

apis/python/src/tiledbsoma/fastercsx.cc
bkmartinjr: According to Miriam's both are correct spelling.  I even see `writable` elsewhere in our code base

johnkerl: That was my thought! :) This is, of course, a nit.

libtiledbsoma/src/utils/fastercsx.h
bkmartinjr: As you mean them, row-major and col-major are Tile concepts.  The memory data structure term of art is being used (see numpy docs for example).

However, I'll add both to hit all readers

apis/python/src/tiledbsoma/_fastercsx.py
bkmartinjr: No, I think that is not correct.  It is a SOMA data structure (due to column names, assumption about column types, etc), encoded as an Arrow table.  You can't hand random Arrow data to it and get anything useful.

I think it has the right name.

johnkerl: Cool, thanks!

apis/python/src/tiledbsoma/fastercsx.cc
bkmartinjr: Yeah, I don't like these types of redundant names when the context is entirely clear to the caller from the C++ namespace or Python package.  But I'll make the change.

apis/python/src/tiledbsoma/fastercsx.cc
johnkerl: (Of course this isn't the only opportunity to do this -- I offer the thought as a general pattern, for your consideration)

apis/python/src/tiledbsoma/_fastercsx.py
bkmartinjr: I guess we could go super verbose, e,g, from_soma_coo_data_as_arrow_table (or a more elegant variant on that), but I think I'd rather just make sure the docstring is clear :-)

apis/python/src/tiledbsoma/fastercsx.cc
bkmartinjr: this is a good idea.  I'll fold some in time permitting, but may not get to all of them on our current schedule.

apis/python/src/tiledbsoma/_fastercsx.py
johnkerl: Works for me!

.github/workflows/r-python-interop-testing.yml
```diff
@@ -27,19 +27,16 @@ jobs:
...
-          fetch-depth: 0  # ensure we get all tags to inform package version determination
+          fetch-depth: 0 # ensure we get all tags to inform package version determination
 
       - name: Bootstrap
         run: cd apis/r && tools/r-ci.sh bootstrap
 
       - name: Dependencies
         run: cd apis/r && Rscript -e "remotes::install_deps(dependencies = TRUE, upgrade = FALSE)"
 
-      - name: CMake
```
johnkerl, line 38: Unclear to me why mods in this file are on this PR

bkmartinjr: Because without these mods, the PR won't build.

johnkerl: Sorry I clicked on the wront lines. I'll re-comment.

.github/workflows/r-python-interop-testing.yml
```diff
@@ -66,15 +63,8 @@ jobs:
       #  if: ${{ matrix.os != 'macOS-latest' }}
       #  run: cd apis/r && Rscript -e "options(bspm.version.check=TRUE); install.packages('tiledb', repos = c('https://eddelbuettel.r-universe.dev/bin/linux/jammy/4.3/', 'https://cloud.r-project.org'))"
 
-      - name: Build and install libtiledbsoma
-        run: sudo scripts/bld --prefix=/usr/local --no-tiledb-deprecated=true --werror=true && sudo ldconfig
-
-      - name: Install R-tiledbsoma
```
johnkerl, line 72: It's unclear to me why changing the R build is necessary for this PR.

I was suspecting you might have been pulling in something from another branch, I don't know what.

If you assert that these mods (lines 73-77 as of this writing) on R-related YAML are necessary for this Python/C++ PR to build, I will take that on your word. But I'd still be grateful for some information from you on what the build failure is: if we do need to modify R YAML for a Python/C++ PR, then I have something to learn, and I'd like to learn it from you.

.github/workflows/r-python-interop-testing.yml
```diff
@@ -66,15 +63,8 @@ jobs:
...
-        run: |
-          cd apis/r
-          R CMD build --no-build-vignettes --no-manual .
-          FILE=$(ls -1t *.tar.gz | head -n 1)
-          R CMD INSTALL $FILE
+      - name: Install Package
+        run: cd apis/r && R CMD INSTALL $(ls -1tr *.tar.gz | tail -1)
```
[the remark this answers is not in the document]
bkmartinjr: I was making this build process match the build process used in `r-ci.yml`.  The R build process seems to have diverged between this file and `r-ci.yml` at some point, and (IMHO) we should use the same build process everywhere.

That was done solely as part of the other cleanup in this file.  Happy to revert it if there is a reason to use two different procedures for the R build.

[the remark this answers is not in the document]
johnkerl: All good, thank you for explaining the change which I had not understood until now! That's all I wanted.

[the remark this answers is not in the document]
johnkerl: As a generate rule I prefer unrelated changes to be done as separate PRs but I don't want to 'make work' for you at this point.

johnkerl (approved):
Please see https://github.com/single-cell-data/TileDB-SOMA/pull/3328#pullrequestreview-2441051729 for narrative.

Thanks @bkmartinjr !

And as on #3328 -- cc @nguyenv for more :eyes.

libtiledbsoma/src/utils/fastercsx.h
```diff
@@ -0,0 +1,545 @@
...
+                                        n_row / tp->concurrency_level())))) +
+                                1;
+    const auto n_partitions = (n_row + (1 << partition_bits) - 1) >>
+                              partition_bits;
+    assert((n_partitions > 0) || (n_row == 0));
+    auto status = parallel_for(
+        tp,
+        0ul,
+        2 * n_partitions,
+        [&n_partitions,
```
johnkerl, line 360: Building locally I see a warning that `n_parititions` isn't used with the lambda body (which is true). Everything builds & runs fine without this line.

There's another warning I don't know how to get rid of

```
warning: captured structured bindings are a C++20 extension
```

referring to `n_col`:

```
  /Users/kerl/git/single-cell-data/TileDB-SOMA/dist/include/tiledbsoma/utils/fastercsx.h:368:11: warning: captured structured bindings are a C++20 extension [-Wc++20-extensions]
    368 |          &n_col](const uint64_t partition) {
        |           ^
  /Users/kerl/git/single-cell-data/TileDB-SOMA/dist/include/tiledbsoma/utils/fastercsx.h:360:9: note: while substituting into a lambda expression here
    360 |         [&partition_bits,
        |         ^
  src/tiledbsoma/fastercsx.cc:367:31: note: in instantiation of function template specialization 'tiledbsoma::fastercsx::compress_coo<unsigned long long, long long, unsigned int, unsigned int>' requested here
    367 |             return fastercsx::compress_coo(
        |                               ^
  /Users/kerl/git/single-cell-data/TileDB-SOMA/dist/include/tiledbsoma/utils/fastercsx.h:321:18: note: 'n_col' declared here
    321 |     auto [n_row, n_col] = shape;
```

We can maybe re-look at this later after @XanthosXanthopoulos has got us on C++ 20.

libtiledbsoma/src/utils/fastercsx.h
```diff
@@ -0,0 +1,545 @@
...
+         &partition_bits,
```
johnkerl, line 361:
```suggestion
        [&partition_bits,
```

Building locally on MacOS I get a warning that `n_partitions` is unused within the lambda body (which is correct).

bkmartinjr: oh, easy fix - I just missed this one

libtiledbsoma/src/utils/fastercsx.h
```diff
@@ -0,0 +1,545 @@
...
+    ThreadPool* const tp,
+    const Shape& shape,
+    uint64_t nnz,
+    const std::vector<tcb::span<const COO_IDX>>& Ai,
+    const std::vector<tcb::span<const COO_IDX>>& Aj,
+    const std::vector<tcb::span<const VALUE>>& Ad,
+    tcb::span<CSX_MAJOR_IDX> Bp,
+    tcb::span<CSX_MINOR_IDX> Bj,
+    tcb::span<VALUE> Bd) {
+    auto [n_row, n_col] = shape;
```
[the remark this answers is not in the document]
bkmartinjr: weird - the specs on the web say this is a C++ 17 feature and safe to use.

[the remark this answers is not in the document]
bkmartinjr: Oh, NM - the issue is that you can't capture them in a lambda until C++ 20.  will fix.

libtiledbsoma/src/utils/fastercsx.h
```diff
@@ -0,0 +1,545 @@
...
+ * always zero).
+ */
+template <typename CSX_MAJOR_IDX>
+void sum_rows_to_pointers_(tcb::span<CSX_MAJOR_IDX>& Bp) {
+    CSX_MAJOR_IDX cumsum = 0;
+    for (uint64_t i = 0; i < Bp.size(); i++) {
+        auto temp = Bp[i];
+        Bp[i] = cumsum;
+        cumsum += temp;
+    }
```
nguyenv, line 225:
```suggestion
    std::partial_sum(Bp.begin(), Bp.end(), Bp.begin());
```

libtiledbsoma/src/utils/fastercsx.h
```diff
@@ -0,0 +1,545 @@
...
+        }
+    }
+
+    // Erase any partitions that are zero length
+    for (auto it = partitions.begin(); it != partitions.end();) {
+        if (it->size == 0)
+            it = partitions.erase(it);
+        else
+            ++it;
+    }
```
nguyenv, line 138:
```suggestion
    partitions.erase(
        std::remove_if(partitions.begin(), partitions.end(),
                       [](const Partition& p) { return p.size == 0; }),
        partitions.end());
```

apis/python/src/tiledbsoma/_fastercsx.py
```diff
@@ -0,0 +1,270 @@
...
+        shape = (
+            (n_major, self.shape[1])
+            if self.format == "csr"
+            else (self.shape[0], n_major)
+        )
+        return CompressedMatrix._to_scipy(
+            indptr, indices, data, shape, self.format, self.sorted
+        )
+
+    def to_numpy(self, index: slice | None = None) -> NDArrayNumber:
```
nguyenv, line 190: I don't think is a necessary change, but we have been more consistently using `Optional` from the `typing` module

```suggestion
    def to_numpy(self, index: Optional[slice] = None) -> NDArrayNumber:
```

apis/python/src/tiledbsoma/_read_iters.py
```diff
@@ -406,12 +409,17 @@ def _cs_reader(
...
+            assert self.context is not None
+            sp = CompressedMatrix.from_ijd(
+                i,
+                j,
+                d,
+                shape,
+                "csr" if self.major_axis == 0 else "csc",
+                True,
+                self.context,
+            ).to_scipy()
```
nguyenv, line 422: To make it easier to read

```suggestion
            sp = CompressedMatrix.from_ijd(
                i,
                j,
                d,
                shape=shape,
                format="csr" if self.major_axis == 0 else "csc",
                make_sorted=True,
                context=self.context,
            ).to_scipy()
```

nguyenv (approved):
LGTM. Some minor suggestions

libtiledbsoma/src/utils/fastercsx.h
bkmartinjr: As I read the spec for std::partial_sum, this won't work for two reasons:
* The suggestion is not equivalent to the original code.   Given [1,2,3], the original produces [0, 1, 3, 6].  Partial sum produces [1, 3, 6].
* We also can't do `std::partial_sum(Bp.begin(), Bp.end()-1, Bp.begin()+1)`, as the output iterator overlaps the input - e.g., it would produce undefined behavior given the std::partial_sum spec..

apis/python/src/tiledbsoma/_fastercsx.py
bkmartinjr: Since Python 3.10, and in our style guide (Google Python style), the `type | None` is recommend, and `Optional` is not recommended.  We have a bunch of the legacy stuff from the Python 3.8 days, but need to clean it up.  Python 3.9 supports this syntax as long as you `from __futures__ import annotations`.

libtiledbsoma/src/utils/fastercsx.h
bkmartinjr: ah, good catch!  If only we could use `erase_if` :-)

apis/python/src/tiledbsoma/_fastercsx.py
johnkerl: IDK about policy. In our codebase we've been using `Optional` for the longest time, and i've been using `Optional` myself, certainly.

Thanks for the information that "Since Python 3.10, and in our style guide (Google Python style), the type | None is recommend, and Optional is not recommended" -- this comes as news to me. (And embarrassingly so -- the shortcoming is my own for not having been on top of that datapoint -- ideally I'd have made myself aware of this current recommendation in the Google Python style.) Re "clean it up", thank you for sharing your perspective that our status quo is unclean. (Personally I dislike "clean" in this context; I prefer to use terminology indicating that we're updating to match more recent recommendations, as such phrasing is more inclusive and value-neutral.)

Let's keep this PR as-is in this regard.

And I'll file a follow-up / lower-pri task to make a blanket pass through our codebase to convert the `Optional` typing to match more recent recommendations.

johnkerl: https://github.com/single-cell-data/TileDB-SOMA/issues/3334

bkmartinjr: I've stopped using Optional for quite some time - just out of sync with the rest of the team I guess.  Definitely advocate for migrating to a more modern syntax, although I don't see any particular urgency for the chore...

johnkerl: @bkmartinjr totally agreed re non-urgency. Thank you for leading the charge here -- it's the rest of the team that's out of sync with _you_. ;)

libtiledbsoma/src/utils/fastercsx.h
nguyenv: Yeah you're right, it is not equivalent. Will mark this as resolved.

apis/python/src/tiledbsoma/_fastercsx.py
nguyenv: OIC. Thanks for the heads up. I'll start moving to using `| None` then instead of `Optional`.

PR #3307, opened 2024-11-07
[python] Ingest `somacore` classes

ryan-williams:
Accompanying SOMA "core" changes: https://github.com/single-cell-data/SOMA/pull/244

It can be awkward having some implementations in that repo:
- I'm experimenting with a `dask` kwarg to `ExperimentAxisQuery.to_anndata`; EAQ impl has lived in the core repo, to date, but arguably should not
- @bkmartinjr has been working on improved "fast CSR" algorithms, which were previously implemented and used in the SOMA core repo, but arguably belong in an implementation repo like this one.

**Changes:**
Move files and classes from single-cell-data/SOMA to this repo:
- `_query.py`:
  - `Axis`
  - `AxisIndexer`
  - `AxisQueryResult`
  - `ExperimentAxisQuery`
  - `MatrixAxisQuery`
  - `JoinIDCache`
- `_fast_csr.py`
- `_eager_iter.py`

### DIFF
--- a/apis/python/src/tiledbsoma/_query.py
+++ b/apis/python/src/tiledbsoma/_query.py
@@ -147,7 +147,7 @@ class AxisIndexer(query.AxisIndexer):
 def _to_numpy(it: Numpyable) -> npt.NDArray[np.int64]:
     if isinstance(it, np.ndarray):
         return it
-    return it.to_numpy()  # type: ignore[no-any-return]
+    return it.to_numpy()
 
 
 @attrs.define(frozen=True)
--- a/apis/python/src/tiledbsoma/_query.py
+++ b/apis/python/src/tiledbsoma/_query.py
@@ -1,0 +1,850 @@
+# Copyright (c) 2021-2023 The Chan Zuckerberg Initiative Foundation
+# Copyright (c) 2021-2023 TileDB, Inc.
+#
+# Licensed under the MIT License.
+
+"""Implementation of a SOMA Experiment.
+"""
+import enum
+from concurrent.futures import ThreadPoolExecutor
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Literal,
+    Mapping,
+    Optional,
+    Protocol,
+    Sequence,
+    Tuple,
+    TypeVar,
+    cast,
+    overload,
+)
+
+import attrs
+import numpy as np
+import numpy.typing as npt
+import pandas as pd
+import pyarrow as pa
+import pyarrow.compute as pacomp
+import scipy.sparse as sp
+from anndata import AnnData
+from somacore import (
+    AxisQuery,
+    DataFrame,
+    ReadIter,
+    SparseRead,
+    query,
+)
+from somacore.data import _RO_AUTO
+from somacore.options import (
+    BatchSize,
+    PlatformConfig,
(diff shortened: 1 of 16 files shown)


apis/python/src/tiledbsoma/_query.py
```diff
@@ -0,0 +1,850 @@
...
+        self.experiment = experiment
+        self.measurement_name = measurement_name
+
+        self._matrix_axis_query = MatrixAxisQuery(obs=obs_query, var=var_query)
+        self._joinids = JoinIDCache(self)
+        self._indexer = AxisIndexer(
+            self,
+            index_factory=index_factory,
+        )
+        self._index_factory = index_factory
```
bkmartinjr, line 245: We could also remove all of this factory support for indexers.  It was added to break the circular dependency when the C++ indexer was added to `tiledbsoma`.  At this point, there is no reason the query implementation can't just use tiledbsoma.IntIndexer directly.

bkmartinjr (approved):
This PR looks good to me, with one optional item I noted in comments (clean-up opportunity in query.py).  I think we should do that cleanup, but I'm fine if it arrives in a later PR.

johnkerl (approved):

apis/python/src/tiledbsoma/_query.py
```diff
@@ -0,0 +1,850 @@
...
+        self._threadpool_: Optional[ThreadPoolExecutor] = None
```
bkmartinjr, line 246: We should also remove this class-specific threadpool, and use Experiment.context.threadpool (the common threadpool used elsewhere).  The implementation in somacore predated the existence of the SOMA context threadpool and at this point is obsolete/wasteful.

bkmartinjr: As an aside, I can do the threadpool cleanup when I add the partitioned reader -- if that is helpful.

ryan-williams: I've just added https://github.com/single-cell-data/TileDB-SOMA/pull/3307/commits/3017ad8913f8a5624131f1a132e8f4108446fd25 to this PR, which was my attempt to address this.

[AFAICT](https://github.com/single-cell-data/TileDB-SOMA/blob/8b6ab64992c71a5d84ce7b690f7afc379791337b/apis/python/src/tiledbsoma/options/_soma_tiledb_context.py#L129), it is now safe to assume `Experiment.context.threadpool is not None`. Existing code+test implied otherwise, but I'm guessing that's out of date / no longer relevant. Can someone confirm?

apis/python/src/tiledbsoma/_query.py
ryan-williams: I looked into this a bit but couldn't immediately figure it out; suggest we leave it for follow-on work.

apis/python/src/tiledbsoma/_query.py
bkmartinjr: Yes, I agree that the tests are out of date.  There is code in the repo that assumes threadpool must not be None.

We should update the tests!

apis/python/src/tiledbsoma/_query.py
bkmartinjr: works for me.  I need to do a bunch of work in this code, so I can take it on

PR #3317, opened 2024-11-08
[python] Config type aliases

ryan-williams:
We use `dict[str, str | float]` in a few places, to mean "TileDB config"; this factors+names that.

#3740 was the impetus for resurrecting this; I'm referencing this type in other files there.

### DIFF
--- a/apis/python/src/tiledbsoma/options/__init__.py
+++ b/apis/python/src/tiledbsoma/options/__init__.py
@@ -2,11 +2,12 @@
 #
 # Licensed under the MIT License.
 
-from ._soma_tiledb_context import SOMATileDBContext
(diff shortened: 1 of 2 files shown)


ryan-williams:
Found myself declaring a `tiledb_config: dict[str, str | float]` parameter in a new method (in Dask integration code), and thought I'd revisit this, partly for discussion @johnkerl @nguyenv. 

It would be nice to canonicalize what types are allowed in configs, otoh it's a bit awkward here distinguishing between `dict` and `Mapping` versions, and when `None` is an allowed value (indicating a key should be removed). `dict` is invariant while `Mapping` is covariant, but the latter doesn't support `.update`.

This was the cleanest version I've found so far, interested in your thoughts!

johnkerl (approved):

PR #3322, opened 2024-11-13
[python] Add `MultiscaleImage` methods to access level URI

jp-dark:
This adds two new methods to access the URIs of resolution levels in the MultiscaleImage.

1. `levels` - a method similar to the `Collection.members` method. Returns shape instead of element types as all levels have type `DenseNDArray`.
2. `level_uri` - returns just the URI for a level specified by either index or name.

This addition is needed for exporting a `MultiscaleImage` level to SpatialData.

### DIFF
--- a/apis/python/src/tiledbsoma/_multiscale_image.py
+++ b/apis/python/src/tiledbsoma/_multiscale_image.py
@@ -9,7 +9,7 @@ Implementation of a SOMA MultiscaleImage.
 
 import json
 import warnings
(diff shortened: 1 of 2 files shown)


johnkerl (approved):

jp-dark:
[sc-59611](https://app.shortcut.com/tiledb-inc/story/59611/add-methods-to-multiscaleimage-to-access-level-uri)

PR #3323, opened 2024-11-13
[python] Return identity for `MultiscaleImage` transform to/from level 0

jp-dark:
The to/from level 0 transform is always the identity transform. Return the `IdentityTransform` class instead of the more generic `ScaleTransform` with ones for the scale factors.


jp-dark:
[sc-59612](https://app.shortcut.com/tiledb-inc/story/59612/return-identity-transform-for-level-0-in-the-multiscaleimage-class)

PR #3325, opened 2024-11-14
[python] Rewrite Xarray support to use dask arrays

jp-dark:
Update the Xarray support to create `xarray.DataArray` objects with data stored in dask arrays with chunksize equal to the `DenseNDArray` tile extents. This change is required to match the SpatialData `Image2DModel` and `Image3DModel` requirements.

### DIFF
--- a/apis/python/tests/test_basic_xarray_io.py
+++ b/apis/python/tests/test_basic_xarray_io.py
@@ -10,62 +10,42 @@ soma_xarray = pytest.importorskip("tiledbsoma.experimental._xarray_backend")
 xr = pytest.importorskip("xarray")
 
 
-@pytest.fixture(scope="module")
-def sample_1d_dense_array(tmp_path_factory):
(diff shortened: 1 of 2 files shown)


apis/python/tests/test_basic_xarray_io.py
```diff
@@ -130,68 +131,32 @@ def test_getitem_all(self, xr_soma_variable, xr_numpy_variable):
...
-                slice(
-                    None,
-                ),
-            ),
+            (2, 1, slice(None)),
             (0, slice(None, None), 1),
             ([1, 3], 1, 1),
             ([1, 3], 1, [0, 2]),
             (-1, -1, -1),
-            (0, 1, slice(0, 4, 2)),
```
ivirshup, line 144: Was this meant to be removed?

ivirshup (approved):

PR #3328, opened 2024-11-15
[python] Partition sparse matrix reads in `tiledbsoma.io.to_anndata`

bkmartinjr:
Partition reads to sparse matrices in the `soma.io.to_anndata` code path. This change is solely for performance, and has no functional impact.


johnkerl (approved):
Thanks @bkmartinjr -- this is a tour de force.

Looks good to me 👀 -wise; also, I ran `to_anndata` on local disk, S3, and cloud-hosted, comparing `main` to this branch, on EC2 Ubuntu `m5.4xlarge`, as well as on my M1 Mac. (I have a general feeling, more from the R side than the Python side admittedly, that 'threads are a little different' on MacOS so this was worth some hands-on/eyes-on experimenting.) I saw a speedup at a solid 2X at 100K cells, with all storage backends. I'll try 300K runs later; I won't report back here unless I see something surprising (which I don't expect.)

cc @nguyenv for another pair of eyes.

PR #3329, opened 2024-11-18
[c++] `SOMAColumn` part 1

XanthosXanthopoulos:
`SOMAColumn` provides an abstraction over TileDB attributes and dimensions and exposes a common interface for all columns regardless of type. Subclasses of `SOMAColumn` can implement complex indexing mechanism through additional dimensions and encapsulate all that logic in one place and make it modular. 

Subsequent PRs will add implementation for dimension and attributes.

Throughout this PR there is extensive use of `std::any` to enable polymorphism with the different `SOMAColumn` types while maintaining a templated interface at the abstract `SOMAColumn`.

Currently there are 3 types of `SOMAColumn`:
 - `SOMAAttribute` wraps a TileDB Attribute and an optional enumeration
 - `SOMADimension` wraps a TileDB Dimension and provides methods to access and manipulate the current domain and set the extend for queries
 - `SOMAGeometryColumn` wraps a set of TileDB Dimensions as well as an Attribute and provides methods to access and manipulate the current domain and set the extend for queries


libtiledbsoma/src/soma/soma_column.h
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    /**
+     * @brief Set the dimension slice using multiple ranges
+     *
+     * @note Partitioning is not supported
+     *
+     * @tparam T
+     * @param dim
+     * @param ranges
```
jp-dark, line 235: These don't match the function template parameter/parameter names.

jp-dark (approved):

libtiledbsoma/src/soma/soma_column.h
```diff
@@ -0,0 +1,343 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @section DESCRIPTION
+ *
+ *   This file defines the SOMAColumn class.
```
johnkerl, line 30: Please say what this class is for, what problem it solves. This comment, as is, doesn't tell anything that `class SOMAColuimn { ... }` doesn't.

In particular, up until there's been a 1-1 mapping of one soma column -> one tiledb dim or one tiledb attr. Here it appears (I'm still reading but it appears so far) that you are allowing for one-to-many mappings of one soma column -> more than one tiledb dim and/or more than one tiledb attr. A top-of-file comment here would need to explain this in at least several sentences' worth of detail.

libtiledbsoma/src/soma/soma_column.h
```diff
@@ -0,0 +1,343 @@
...
+
+    /**
+     * Get the datatype of the TileDB Attributes if any. All attributes must
+     * have the same type.
+     */
+    virtual std::optional<tiledb_datatype_t> data_type() const = 0;
+
+    /**
+     * @brief Select columns names to query (dim and attr). If the
+     * `if_not_empty` parameter is `true`, the column will be selected iff the
```
johnkerl, line 116: "iff" is specialist mathematical jargon that non-mathematicians are unlikely to recognize. Please don't use it.

libtiledbsoma/src/soma/soma_column.h
```diff
@@ -0,0 +1,343 @@
...
+     * list of selected columns is empty. This prevents a `select_columns` call
+     * from changing an empty list (all columns) to a subset of columns.
+     *
+     * @param query the ManagedQuery object to modify
+     * @param if_not_empty Prevent changing an "empty" selection of all columns
```
johnkerl, line 121:
```suggestion
     * @param if_not_empty Prevent changing an empty selection of all columns
```

johnkerl (changes requested):
This looks like interesting work but it's not clear why it's being done.

You need a GitHub issue, or Shortcut link, in the description field.

From your GitHub branch name I can discover a Shortcut story number -- since this is an open-source project, please include within a GitHub issue the context that is necessary for anyone who is not a TileDB employee to understand what changes you're making here, and why.

This PR introduces new code without any unit tests, and no callsite from existing code that would exercise existing unit tests. It could be stamped now with guarantee of unit-test cases to follow, but, I'd be happier if you could offer at least a minimal set of tests.

libtiledbsoma/src/soma/soma_column.h
```diff
@@ -0,0 +1,343 @@
...
+     */
+    virtual void select_columns(
+        const std::unique_ptr<ManagedQuery>& query,
+        bool if_not_empty = false) const = 0;
+
+    /**
+     * @brief Returns true if the SOMACOlumn has a non-empty current domain,
+     * else false.  Note that this only should be used if the SOMACOlumn objects
```
nguyenv, line 129:
```suggestion
     * @brief Returns true if the SOMAColumn has a non-empty current domain,
     * else false.  Note that this only should be used if the SOMAColumn objects
```

libtiledbsoma/src/soma/soma_column.h
```diff
@@ -0,0 +1,343 @@
...
+    //= public non-static
+    //===================================================================
+
+    /**
+     * Construct a new SOMAColumn object.
+     *
+     * @param ctx TileDB context
+     */
+    SOMAColumn(const Context& ctx)
+        : ctx(ctx){};
```
nguyenv, line 63: Should we take in a `tiledb::Context` or `SOMAContext`?

XanthosXanthopoulos:
This PR has been split into #3424 , #3425 , #3426 and #3427.

PR #3331, opened 2024-11-18
[c++/python/r] Migrate to C++20

XanthosXanthopoulos:
This PR migrates minimum C++ version to C++20 and replaces `fmt::format` with the now available `std::format`

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   backport:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     name: Backport
     steps:
       - name: Backport
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -18,7 +18,7 @@ name: pkgdown
 
 jobs:
   pkgdown:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     # This is GitHub Actions magic -- there are no secrets we as package owners need to set up or manage
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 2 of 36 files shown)


johnkerl:
One note and cc @mojaveazure -- given that there are compile errors if `RcppSpdlog` is on a user's system and they have not updated -- I'd love a lower bound in `apis/r/DESCRIPTION` like `RcppSpdlog (>= 0.0.19)` -- however -- this is only a `LinkingTo` not an `Imports` https://github.com/single-cell-data/TileDB-SOMA/blob/75112e5e4646b400bb19f132d70114fc5414b12b/apis/r/DESCRIPTION#L52 ... is there a good way to get some automation around upgrading `RcppSpdlog` ?

mojaveazure:
>  this is only a `LinkingTo` not an `Imports`

I'm not sure what that has to do with anything. In R, the "hard" dependencies are `Depends`, `Imports`, and `LinkingTo`; these must be present and at the specified minimum version before the package can be installed. If we bump RcppSpdlog to `>= 0.0.19`, any user who does `install.packages("tiledbsoma")` or `update.packages()` will be required to install the newer version of RcppSpdlog before tiledbsoma 1.15.0 can be installed

Moreover, R and CRAN allow all types of dependencies (`Depends`, `Imports`, `LinkingTo`, `Suggests`, and `Enhances`) to have lower bounds (though only the first three have a hard check, manual checks must be carried out for `Suggests` and `Enhances`). Case in point, [Seurat](https://cran.r-project.org/package=Seurat) links to `Rcpp >= 0.11.0`

johnkerl:
OK thank you @mojaveazure !

I did get an error on my laptop until I upgraded `RcppSpdlog`.

So since @mojaveazure asserts that `LinkingTo` can have a lower bound, I think let's please put in a lower bound at 0.0.19 for `RcppSpglog`.

PR #3335, opened 2024-11-18
[python] Re-introduce tiledb_ctx for SOMATileDBContext

nguyenv:
**Issue and/or context:**

[sc-58711]

**Changes:**

- Re-introduce the tiledb_ctx argument and attribute
- Correct documentation
- Update typing

**Notes for Reviewer:**

We are not reintroducing tiledb-py as a dependency. The `tiledb_ctx` is only usable when tiledb-py is installed.


nguyenv:
[sc-58711]

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -66,25 +70,30 @@ class SOMATileDBContext(ContextBase):
 
     def __init__(
         self,
-        tiledb_config: Optional[Dict[str, Union[str, float]]] = None,
-        timestamp: Optional[OpenTimestamp] = None,
-        threadpool: Optional[ThreadPoolExecutor] = None,
+        tiledb_ctx: "tiledb.Ctx" | None = None,
+        tiledb_config: Dict[str, Union[str, float]] | None = None,
```
johnkerl, line 74: Please move `tiledb_ctx` after `tiledb_config`, here & in the docstring. The config-dict needs to be clearly "plan A".

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -66,25 +70,30 @@ class SOMATileDBContext(ContextBase):
 
     def __init__(
         self,
-        tiledb_config: Optional[Dict[str, Union[str, float]]] = None,
-        timestamp: Optional[OpenTimestamp] = None,
-        threadpool: Optional[ThreadPoolExecutor] = None,
+        tiledb_config: Dict[str, Union[str, float]] | None = None,
```
bkmartinjr, line 73: the `Union[foo,bar]` and the `foo | bar` syntax are equivalent.  Recommend picking one, e.g., either:

> Dict[str, str|float] | None

or

> Union[Dict[str, Union[str, float]], None]

The first one is the more modern syntax and (I think) is more readable.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -119,28 +128,44 @@ def __init__(
         """
         self._lock = threading.Lock()
         """A lock to ensure single initialization of ``_tiledb_ctx``."""
-        self._initial_config: Optional[Dict[str, Union[str, float]]] = (
-            None if tiledb_config is None else _default_config(tiledb_config)
-        )
+
+        self._initial_config: Dict[str, Union[str, float]] | None = None
```
bkmartinjr, line 132: > Dict[str, str | float] | None

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -7,10 +7,11 @@
 
 import datetime
 import functools
+import importlib.util
```
bkmartinjr, line 10: see comments below - directly using importlib is not required

bkmartinjr (changes requested):
Couple of issues:

1. The PR is missing the `tiledb_ctx` property that was present in the API.  This is the first point mentioned in the bug report.  Previously it was possible to do the following:
```
context = soma.SOMATileDBContext()
A = tiledb.open(uri, ctx=context.tiledb_ctx)
```

We did deprecate this property in 1.14.5, so removing it is fair game.  I'm just unsure what decision @aaronwolen took - the bug report suggests that he wanted to add both back, but I may be misreading it.

Recommend checking with him to confirm removal. 

2. There are some minor style/pattern comments inline.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -119,28 +128,44 @@ def __init__(
...
         """A lock to ensure single initialization of ``_tiledb_ctx``."""
-        self._initial_config: Optional[Dict[str, Union[str, float]]] = (
-            None if tiledb_config is None else _default_config(tiledb_config)
-        )
+
+        self._initial_config: Dict[str, Union[str, float]] | None = None
+        if tiledb_config is not None:
+            self._initial_config = _default_config(tiledb_config)
+        elif tiledb_ctx is not None:
+            if not importlib.util.find_spec("tiledb"):
```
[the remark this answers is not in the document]
nguyenv: What about in a situation where `tiledb` is not installed but `tiledb.cloud` is? You would still be able to `import tiledb` but `tiledb.do_something()` would error out. Is it not necessary to handle that case?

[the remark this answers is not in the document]
nguyenv: I can repro this scenario. Would this be considered a broken user environment though and out of our scope to handle?

```
(soma-3.11) vivian@mangonada:~/TileDB-SOMA$ python
Python 3.11.0 | packaged by conda-forge | (main, Jan 14 2023, 12:27:40) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tiledb
>>> tiledb
<module 'tiledb' (<_frozen_importlib_external.NamespaceLoader object at 0x7fe4bad5de50>)>
>>> tiledb.Ctx()
AttributeError: module 'tiledb' has no attribute 'Ctx'
>>>
```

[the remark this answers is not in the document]
nguyenv:
```
(soma-3.11) vivian@mangonada:~/TileDB-SOMA$ pip install tiledb tiledb-cloud

(soma-3.11) vivian@mangonada:~/TileDB-SOMA$ pip uninstall tiledb tiledb-cloud

(soma-3.11) vivian@mangonada:~/TileDB-SOMA$ python
Python 3.11.0 | packaged by conda-forge | (main, Jan 14 2023, 12:27:40) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tiledb
>>> tiledb.Ctx()
AttributeError: module 'tiledb' has no attribute 'Ctx'
```

This results in an empty `tiledb` directory in `site-packages` where `import tiledb` still works but is unusable.
```
(soma-3.11) vivian@mangonada:~/TileDB-SOMA$ ls /home/vivian/anaconda3/envs/soma-3.11/lib/python3.11/site-packages/tiledb*
/home/vivian/anaconda3/envs/soma-3.11/lib/python3.11/site-packages/tiledb
>>>

```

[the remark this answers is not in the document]
nguyenv: Had I not uninstalled `tiledb-cloud` and only `tiledb`, the `site-packages/tiledb/cloud` would still exist, so `import tiledb` would still work in that case too.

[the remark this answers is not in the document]
bkmartinjr: That is definitely a broken environment, as your manual uninstalls break the package dependencies.

An example:  if I install `tiledbsoma` (which depends on pyarrow, numpy, etc), and then manually uninstall numpy, etc -- you can still import tiledbsoma, but it won't work.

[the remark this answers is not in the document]
nguyenv: OK I will do just do `try`/`except` in the future then. I was doing that initially but made it more complicated due to my own broken env.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -66,25 +70,30 @@ class SOMATileDBContext(ContextBase):
 
     def __init__(
         self,
-        tiledb_config: Optional[Dict[str, Union[str, float]]] = None,
-        timestamp: Optional[OpenTimestamp] = None,
-        threadpool: Optional[ThreadPoolExecutor] = None,
+        tiledb_config: Dict[str, Union[str, float]] | None = None,
+        tiledb_ctx: "tiledb.Ctx" | None = None,
```
teo-tsirpanis, line 74:
```suggestion
        tiledb_config: Dict[str, Union[str, float]] | "tiledb.Ctx" | None = None,
```

I would slightly prefer this. Admittedly passing a context to the `tiledb_config` doesn't sound very good, but it eliminates the error case of passing both a context and a config. You could take a leap and say that a context is part of the "configuration".

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -153,17 +196,53 @@ def native_context(self) -> clib.SOMAContext:
...
                     self._native_context = clib.SOMAContext(
                         {k: str(v) for k, v in cfg.items()}
                     )
-            return self._native_context
+                elif self._tiledb_ctx is not None:
+                    # The user passed in a tiledb_ctx; if tiledb is not installed
+                    # it should be impossible to enter into this block because
+                    # we already check that in the constructor
+                    assert tiledb is not None
+                    cfg = self._tiledb_ctx().config().dict()
```
bkmartinjr, line 211: I don't think `_tiledb_ctx` is a method, so I suspect this should drop the parens, e.g.,
> cfg = self._tiledb_ctx.config().dict()

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -186,6 +265,12 @@ def _internal_tiledb_config(self) -> Dict[str, Union[str, float]]:
         if self._native_context is not None:
             return dict(self._native_context.config())
 
+        # We have TileDB Context. Return its actual config.
+        # TODO This block will be deleted once tiledb_ctx is removed
```
bkmartinjr, line 269: this comment seems obsolete

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
bkmartinjr: ps. assuming my comment is correct, does this imply a missing test case?  (or my comment is wrong!)

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
```diff
@@ -19,16 +19,30 @@
 from .._types import OpenTimestamp
 from .._util import ms_to_datetime, to_timestamp_ms
 
+if TYPE_CHECKING:
```
bkmartinjr, line 22: We should not need this as all the type signatures are quoted.  

Also, given that `tiledb` is no longer a dependency, shouldn't this fail in CI (I wonder why it did not fail?)

bkmartinjr (changes requested):
couple of minor suggestions/issues

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
nguyenv: You are right that there should not be parens.

The missing test case brings up a another issue though. We have removed all tiledb-py usage from the unit tests. But should we temporarily re-introduce it for testing with `tiledb.Ctx()`? We would `xskip` the test if `tiledb` is not installed locally. But then do we need to install `tiledb` for CI runs? (cc @johnkerl)

bkmartinjr: yeah, tough one.  I don't see any great solution - options?

The best I can come up with is:
1. Code the tests to work both with and without the package installed:
    * if installed, test that it works as expected
    * if not installed, test that it generates the right error
2. Have two CI's:  one with it installed, one with it not installed

That at least mimics real-world conditions.

apis/python/src/tiledbsoma/options/_soma_tiledb_context.py
nguyenv: I had assumed if the modules included in `TYPE_CHECKING` are not found, then it treats the quoted types as string literals instead of types.

bkmartinjr: It treats them as string literals for sure, but I don't think it has anything to do with whether the module is found.  TYPE_CHECKING is a constant that mypy sets to true, and is false during actual "use" of the code.

Easiest solution is to probably just do:  `from tiledb import Ctx` and use "Ctx" rather than "tiledb.Ctx".

Sorry this is such a PITA!

nguyenv: This ended up being more complicated than expected, but I think I have something that works now.

bkmartinjr (approved):
Nice, this is much cleaner. TY @nguyenv

PR #3340, opened 2024-11-19
[c++/python] Switch to `std::span`

XanthosXanthopoulos:
Continuing the migration to C++20, this PR replaces `tcb::span` with `std::span`


johnkerl (approved):
@XanthosXanthopoulos I know you still have this as draft, but, since CI is green, and with respect for timezone skew I've gone ahead and approved it so it'll be ready for you to merge in your morning.

Thanks for doing this! :)

Issue #3154, opened 2024-10-09
[c++] Use C++20


johnkerl:
Context:

* Core `libtiledb` is already built using C++20 as of 2023: https://github.com/TileDB-Inc/TileDB/pull/4034
* We would like to also build `libtiledbsoma` using C++20
  * General currency with the community
  * Currency with core
  * Access to newer language features
* PR: #3153
* [[sc-57301]](https://app.shortcut.com/tiledb-inc/story/57301/use-c-20)
* https://github.com/fmtlib/fmt/issues/4179
 
Issues:

* https://github.com/single-cell-data/TileDB-SOMA/pull/3153
* Works on Linux fine
* On MacOS: https://github.com/single-cell-data/TileDB-SOMA/actions/runs/11256492779/job/31298502792#step:10:36
```
using C++ compiler: ‘g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0’
[19](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/11256492779/job/31298502792#step:10:20)using C++20
[20](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/11256492779/job/31298502792#step:10:21)g++ -std=gnu++20 -I"/usr/share/R/include" -DNDEBUG -I. -I../inst/include/ -I/__w/TileDB/TileDB/dist/include -I/usr/local/include -I/usr/local/include/tiledbsoma    -I'/usr/lib/R/site-library/Rcpp/include' -I'/usr/lib/R/site-library/RcppSpdlog/include' -I'/usr/lib/R/site-library/RcppInt64/include' -I'/usr/lib/R/site-library/nanoarrow/include'     -fpic  -g -O2 -ffile-prefix-map=/build/r-base-2y82rL/r-base-4.4.1=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2   -c RcppExports.cpp -o RcppExports.o
[21]In file included from ./rutilities.h:6,
[22]                 from ../inst/include/tiledbsoma_types.h:25,
[23]                 from RcppExports.cpp:4:
[...]

PR #3343, opened 2024-11-19
[c++] Use `std::format` in `spdlog`

XanthosXanthopoulos:
When compiling with `gcc` >13 several error related to `fmt::format` occur when building spdlog. Forcing the use of `std::format` resolves them.

Context: #3154


bkmartinjr (approved):

johnkerl (approved):
🚢

PR #3346, opened 2024-11-20
[python] Update README.md to remove tiledb-py

nguyenv:
This removes the tiledb-py requirement from the `README.md`. All other steps remain the same.


johnkerl (approved):
🚢

PR #3361, opened 2024-11-21
[c++/python] Update minimum macOS version to 13.3

XanthosXanthopoulos:
**Issue and/or context:**
#3154 

**Notes for Reviewer:**
CI broke on a draft PR the above changes fixed it.

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -253,7 +253,7 @@ if os.name != "nt":
     CXX_FLAGS.append(f'-Wl,-rpath,{str(tiledb_dir / "lib")}')
 
 if sys.platform == "darwin":
-    CXX_FLAGS.append("-mmacosx-version-min=11.0")
+    CXX_FLAGS.append("-mmacosx-version-min=13.3")
 
 if os.name == "posix" and sys.platform != "darwin":
     LIB_DIRS.append(str(tiledbsoma_dir / "lib" / "x86_64-linux-gnu"))
--- a/apis/r/src/Makevars.in
+++ b/apis/r/src/Makevars.in
@@ -1,6 +1,6 @@
 CXX_STD = CXX20
 
-## We need the TileDB Headers, and for macOS aka Darwin need to set minimum version 10.14 for macOS
+## We need the TileDB Headers, and for macOS aka Darwin need to set minimum version 13.3 for macOS
 PKG_CPPFLAGS = -I. -I../inst/include/ @tiledb_include@ @cxx20_macos@ -D SPDLOG_USE_STD_FORMAT
 
 ## We also need the TileDB library
--- a/libtiledbsoma/CMakeLists.txt
+++ b/libtiledbsoma/CMakeLists.txt
@@ -112,6 +112,7 @@ endif()
 # Use @rpath on macOS for building shared libraries.
 if(APPLE)
   set(CMAKE_MACOSX_RPATH ON)
+  set(CMAKE_OSX_DEPLOYMENT_TARGET 13.3)
 
   # Don't allow macOS .frameworks to be used for dependencies.
   set(CMAKE_FIND_FRAMEWORK NEVER)


johnkerl:
@XanthosXanthopoulos there are subtleties with MacOS with "10" and "11" which aren't as simple as what one might think.

Looping in @jdblischak for extra eyes.

Re

> CI broke on a draft PR the above changes fixed it.

please share here the link to which draft PR, and which failed CI. (I'm curious to know why that draft PR failed but our `main` is green with C++ 20 for several days now ...)

XanthosXanthopoulos:
This failed #3329 with error `'to_chars' has been explicitly marked unavailable here` caused by `std::format`

jdblischak (commented):
My only concern is from a conda-forge feedstock maintenance perspective:

* We can require a [newer macOS SDK](https://conda-forge.org/docs/maintainer/knowledge_base/#requiring-newer-macos-sdks) in the feedstock builds, but it should be done with caution:
> While the default version 10.13 can be overridden using the following changes to the recipe, it should be done as a last resort.
* I checked, and there are already [50 conda-forge feedstocks](https://github.com/search?q=org%3Aconda-forge+13.3+MACOSX_SDK_VERSION+language%3AYAML&type=code&l=YAML) that set `MACOSX_SDK_VERSION` to 13.3
* The nightly feedstock builds can currently successfully build the Python clients using C++20 and a macOS SDK version of 11 (https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/230#issuecomment-2491993245). And the code in `main` already uses `std::format`, so what is different about the use of `std::format` in #3329?
* If we decide to bump to 13.3, we need to update `MACOSX_SDK_VERSION` (and possibly `c_stdlib_version`, though we should first try without bumping it) in [`conda_build_config.yaml`](https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/fd059f0746315675c5ab5cb93dd24cd81df715ad/recipe/conda_build_config.yaml) and rerendering

johnkerl:
Excellent feedback @jdblischak -- @XanthosXanthopoulos I think we should hold off merging this PR until we do a health-check PR on `tiledbsoma-feedstock` using a `recipe/meta.yml` that points at this PR's commit on TileDB-SOMA. (You can do that if you know how; otherwise I'm happy to.)

Context is we're close to a release, and all releases go through Conda constraints, and there are tricky shoals to navigate on the Conda seas .....

XanthosXanthopoulos:
I don't know how to do the conda feedstock.
Regarding the different behavior on the PR vs. the nightly, I haven't been able to find anything. Even on the same PR some macos workflows built successfully which is even more baffling.

johnkerl:
@XanthosXanthopoulos https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/232

johnkerl:
@XanthosXanthopoulos https://github.com/single-cell-data/TileDB-SOMA/pull/3329 is now passing CI -- ?

XanthosXanthopoulos:
> @XanthosXanthopoulos #3329 is now passing CI -- ?

With the changes of this PR yes. I will try to remove the changes and test again

XanthosXanthopoulos:
The failure is #3329 was triggered by including these 4 lines in libtiledbsoma
```
#include "soma/soma_column.h"
#include "soma/soma_attribute.h"
#include "soma/soma_dimension.h"
#include "soma/soma_geometry_column.h"
```

jdblischak:
> My only concern is from a conda-forge feedstock maintenance perspective:

Quick update. We discovered that we will already have to bump the macOS SDK in the conda feedstock builds to 13.3 in order to accommodate the migration to C++20 (https://github.com/single-cell-data/TileDB-SOMA/issues/3154)

xref: https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/230#issuecomment-2494616252

In other words, we are going to have to migrate the feedstock to 13.3 regardless of the status of the PR https://github.com/single-cell-data/TileDB-SOMA/pull/3329, so I'm fine with this PR.

XanthosXanthopoulos:
Replaced by #3534

Issue #3360, opened 2024-11-21
[python] Inconsistency in `to_anndata` density/sparsity of `obsp`/`varp`


johnkerl:
While `tiledbsoma.io.to_anndata()` exports `obsp`/`varp` as a `scipy.sparse` matrix, `ExperimentAxisQuery.to_anndata()` exports it as a dense ndarray.

The [AnnData API suggests](https://anndata.readthedocs.io/en/stable/generated/anndata.AnnData.obsp.html) the dense ndarray.

Update on #3362: @ivirshup points out that the AnnData docs are wrong and these should be _sparse_ in all cases.

[[sc-59632]](https://app.shortcut.com/tiledb-inc/story/59632)

ivirshup:
Oh let me clarify, it's not that it should be sparse in all cases. It's just that in the majority of cases it should be sparse. If the user really needs a dense array (maybe for a correlation matrix), they can do that.

Ideally you support both. I think the user would decide which they want by what format they stored it in.

PR #3362, opened 2024-11-21
[python] Outgest `obsp`/`varp` as dense

johnkerl:
**Issue and/or context:** #3360

**Changes:**

Outgest these as dense.

**Notes for Reviewer:**

Closed on feedback from @ivirshup (see below). I'll repurpose the tracking task #3360 to have us outgest sparse in all cases.

### DIFF
--- a/apis/python/src/tiledbsoma/io/outgest.py
+++ b/apis/python/src/tiledbsoma/io/outgest.py
@@ -26,7 +26,6 @@ import anndata as ad
 import numpy as np
 import pandas as pd
 import pyarrow as pa
-import scipy.sparse as sp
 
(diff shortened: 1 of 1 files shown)


ivirshup:
That is a problem in the anndata docs, these should definitely be sparse.

These are typically things like nearest neighbor networks. So number of cells X number of cells with like ~15 entries per row. Very sparse!

johnkerl:
cc @aaronwolen and @bkmartinjr -- I'll cancel this.

bkmartinjr:
This just means that the ExperimentAxisQuery is wrong then - one does sparse, one does dense

PR #3375, opened 2024-11-22
[python] Fix incorrect type handling in `ExperimentAxisQuery` sparse reader

bkmartinjr:
PR #3359 introduced a bug in the reading of sparse arrays. In some situations, it would allow a buffer of type insufficient to store the results (int32 instead of int64). This was due to a sign inversion in a comparison.

There was also a second edge case where the dim0 and dim1 buffers would be of different type, which is unsupported by the `fastercsx` package. 

This change chooses the buffer type correctly, based on the max needed by both dimensions.


johnkerl (approved):
Excellent, thanks @bkmartinjr !

PR #3392, opened 2024-11-27
[python] Add support for exporting to SpatialData without transform

jp-dark:
Add support for exporting a `PointCloudDataFrame` or `GeometryDataFrame` to SpatialData even if we don't have a mapping from the object to the `Scene` it is stored in.

### DIFF
--- a/apis/python/src/tiledbsoma/io/spatial/outgest.py
+++ b/apis/python/src/tiledbsoma/io/spatial/outgest.py
@@ -106,6 +106,8 @@ def to_spatial_data_points(
 
     Args:
         points: The point cloud data frame to convert to SpatialData shapes.
+        key: Key for the item in the SpatialData object. Used to set a transformation
+            to the item itself if no scene transform is provided.
         scene_id: The ID of the scene this point cloud dataframe is from.
         scene_dim_map: A mapping from the axis names of the scene to the corresponding
             SpatialData dimension names.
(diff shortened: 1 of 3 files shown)


apis/python/src/tiledbsoma/io/spatial/outgest.py
```diff
@@ -135,9 +139,10 @@ def to_spatial_data_points(
 def to_spatial_data_shapes(
     points: PointCloudDataFrame,
     *,
+    key: str,
```
ktsitsi, line 142: Can you add docstrings for this one. Also could you explain it's use?

ktsitsi (changes requested):

apis/python/src/tiledbsoma/io/spatial/outgest.py
jp-dark: SpatialData requires that you always define a transformation on an object, but we don't. I hit this issue when attempting to transform an entire Scene. This change makes it so I can set a transformation from the object to itself. I just needed to add in a parameter that specifies the name of the object inside SpatialData.

For additional context, these are helper functions. I plan to move them outside of `outgest.py` to a spatial data util module in a future PR. They should only be accessed inside functions that export a full `SpatialData` class.

ktsitsi (approved):

PR #3405, opened 2024-12-06
[c++] Integrate `SOMAColumn`: Arrow adapter methods, part 1

XanthosXanthopoulos:
This PR replaces the Arrow schema to TileDB schema transformation to use the `SOMAColumn` create methods. 
Also there are a set of new data converters from arrow arrays to `std::array` for simplification.

This migration also enforces a current domain restriction for string dimensions to libtiledbsoma in addition to the restriction being present only on the R and Python APIs.


nguyenv (commented):
Can you explain what the `Skip` and `Take` are for and document it? It looks like `Take` is the index of the column to retrieve and `Skip` is relevant only for geometry columns (where it's always 2)?

Also is there a way to use `std::variant` or a templated type instead of `std::any` or would that make things too complicated?

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -724,11 +726,247 @@ class ArrowAdapter {
...
+        }
+
+        return result;
+    }
+
+    /**
+     * Return a copy of the data in a specified column of an arrow table.
+     * Complex column types are supported. The for each sub column are an
+     * std::array<T, 2> casted as an std::any object.
+     */
```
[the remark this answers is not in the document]
XanthosXanthopoulos: `Skip` and `Take` are used in 2 places with 2 specific sets on values (either `Skip=3` and `Take=2` or `Skip=0` and `Take=2`) and are independent of the geometry column. Their usage is to extract specific subranges of ArrowArray data and they come in handy during `ArrowSchema -> TileDBSchema` where the arrow array provided has 5 values per dimension and we only need the last 2 to set the current domain.

As to using `std::variant`, adding more `SOMAColumn` types would require changing multiple variants. The use of `std::any` here is to enable runtime polymorphism and indirectly introduces a runtime type check (via `any_cast`, `make_any`) between the templated function and the actual dimension type. `std::variant` can provide all the above it is just a different style I am open to discuss further.

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -841,11 +843,250 @@ class ArrowAdapter {
         }
     }
 
-    static void set_current_domain_slot(
-        tiledb_datatype_t type,
-        const void* buff,
-        NDRectangle& ndrect,
-        std::string name);
+    template <typename T, size_t Take, size_t Skip = 0>
+    static std::vector<std::array<T, Take>> get_table_column_by_name(
```
jp-dark, line 847: This doesn't appear to be used anywhere. Can you pull it out into a separate PR and/or add it to the branch where it is used?

jp-dark (commented):
I didn't review the `get_table_column_by_name` function, but everything else looks good to me. However, someone with more familiarity with the arrow adapter code should look over this as well.

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -976,8 +980,8 @@ void ArrowAdapter::_set_current_domain_slot(
             LOG_DEBUG(std::format(
                 "[ArrowAdapter] {} current_domain float {} to {}",
                 name,
-                std::to_string(lo),
```
johnkerl, line 979: I do seem to recall these were about crash avoidance on _some_ platform (I don't recall which). I'd rather leave these as-is please.

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -1091,68 +1095,98 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
...
-                if (dim.first.substr(std::max(0ul, dim.first.size() - 5)) ==
-                    std::string("__min")) {
-                    domain.add_dimension(dim.second);
-                }
+        auto column = std::find_if(
+            columns.begin(), columns.end(), [&](auto col) {
+                return strcmp(
+                           col->name().c_str(),
+                           index_column_schema->children[i]->name) == 0;
+            });
```
johnkerl, line 1139: This could be pulled out into a utility function

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -1091,68 +1095,98 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
...
+
+        if (column == columns.end()) {
+            throw TileDBSOMAError(std::format(
+                "[ArrowAdapter][tiledb_schema_from_arrow_schema] Index column "
+                "{} missing",
```
johnkerl, line 1144:
```suggestion
                "'{}' missing",
```

Strings in error messages should always be quoted/bracketed. Even if you think it's impossible for the string to ever be empty. Heaven forbid someday there is some bug somewhere somehow ... and an empty string gets in here ... that needs to be clear to everyone that sees the error message.

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -1091,68 +1095,98 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
...
+                index_column_schema->children[i]->name));
+        }
+
+        if ((*column)->tiledb_dimensions().has_value()) {
```
johnkerl, line 1148: This is a bit awkward.

* `column` needs to be `column_it` or some such -- it is _not_ a column, it is an iterator
* Then, const auto column = *column_it` after you check that `column _it != columns.end()`
* Then the rest of these `(*column)->foo` become `column->foo` as they should be

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -1169,84 +1203,28 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
...
+        if (!column->isIndexColumn()) {
+            continue;
+        }
 
-        for (int64_t i = 0; i < index_column_schema->n_children; ++i) {
-            auto col_name = index_column_schema->children[i]->name;
-            if (strcmp(child->name, col_name) != 0) {
-                continue;
+        if (column->name() == SOMA_GEOMETRY_COLUMN_NAME) {
+            std::vector<std::any> dom;
```
johnkerl, line 1212: As on previous PRs: we cannot simply say `dom` or `domain`, ever.

There are four things it can mean:

* core domain (which is soma maxdomain)
* soma domain (which is core current domain)

The names are confusing (and too late to change), and confusion is too easy, and developer confusion is high-risk

johnkerl: Perhaps rename `dom` to `cdslot`

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -841,11 +843,250 @@ class ArrowAdapter {
         }
     }
 
-    static void set_current_domain_slot(
-        tiledb_datatype_t type,
-        const void* buff,
-        NDRectangle& ndrect,
-        std::string name);
+    template <typename T, size_t Take, size_t Skip = 0>
```
johnkerl, line 846: Please state as a compact summary, right here as a code comment, what `Take` and `Skip` are for, what they do, and an example usage.

johnkerl (changes requested):
Thanks for working on this @XanthosXanthopoulos !

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -841,11 +843,307 @@ class ArrowAdapter {
...
     }
 
-    static void set_current_domain_slot(
-        tiledb_datatype_t type,
-        const void* buff,
-        NDRectangle& ndrect,
-        std::string name);
+    /**
+     * Return a copy of the data in a specified column of an arrow table.
+     * Complex column types are supported. The type for each sub column is an
```
johnkerl, line 848:
```suggestion
     * Complex column types are supported. The type for each subcolumn is an
```

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -841,11 +843,307 @@ class ArrowAdapter {
...
+        } else {
+            result.push_back(get_table_any_column<Take, Skip>(
+                selected_array, selected_schema));
+        }
+
+        return result;
+    }
+
+    /**
+     * Read a part of am Arrow array to an std::array and cast in to and
```
johnkerl, line 904:
```suggestion
     * Read a part of an Arrow array to an std::array and cast in to an
```

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -841,11 +843,307 @@ class ArrowAdapter {
...
+     * std::any object.
+     *
+     * @example get_table_any_column<3, 2>(array, schema) will return an
+     * std::array<T, 3> where T is the appropriate type based on the Arrow array
+     * format casted as an std::any object. The std::array will skip 2 element
```
johnkerl, line 909:
```suggestion
     * format casted as an std::any object. The std::array will skip 2 elements
```

libtiledbsoma/test/unit_soma_column.cc
```diff
@@ -317,6 +317,170 @@ TEST_CASE_METHOD(
...
+        // Can we write empty strings in this range?
+        REQUIRE(str_range[0] <= "");
+        REQUIRE(str_external.first <= "");
+        REQUIRE(str_range[1] >= "");
+        REQUIRE(str_external.second >= "");
+        // Can we write ASCII values in this range?
+        REQUIRE(str_range[0] < " ");
+        REQUIRE(str_external.first <= " ");
+        REQUIRE(str_range[1] > "~");
+        // REQUIRE(str_external.second >= "~");
```
johnkerl, line 346: Please remove

libtiledbsoma/src/utils/util.cc
```diff
@@ -98,4 +98,22 @@ std::vector<uint8_t> cast_bit_to_uint8(ArrowSchema* schema, ArrowArray* array) {
     return casted;
 }
 
+std::shared_ptr<SOMAColumn> find_column_by_name(
```
johnkerl, line 101: Nice 😎

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -841,11 +843,307 @@ class ArrowAdapter {
...
+     * and copy the next 3 from the Arrow array.
+     *
+     * @tparam Take The number of elements to read.
+     * @tparam Skip The number of elements to skip.
```
johnkerl, line 913: I was wondering why `Take` and `Skip` are template parameters and not simply function arguments. (The latter seems simpler.)

After having read through your code here, the answer seems to be: because you are constructing `std::array` objects (which are fixed-size) in various places, rather than `std::vector` (which are not fixed-size).

It's a good reason to have `Take` and `Skip` be implemented as template parameters, but I found it non-obvious.

Please come out and state this reasoning, right here, as a comment.

johnkerl (changes requested):

libtiledbsoma/src/utils/arrow_adapter.h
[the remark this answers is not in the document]
johnkerl: @XanthosXanthopoulos to be clear, I'm not asking you to  use a different solution. I'm asking to be clear to the reader the (very good!) reasons you chose the solution you're using.

Please take both paragraphs you just wrote, and put them in the code as comments as well. They are _highly_ informative. :)

jp-dark (approved):
LGTM once the remaining change requests are addressed.

libtiledbsoma/src/utils/arrow_adapter.h
XanthosXanthopoulos: `Skip` wasn't required to be a template parameter, so I replaced it with just a function argument

libtiledbsoma/src/soma/soma_column.cc
```diff
@@ -32,8 +32,98 @@
...
+                                       ArrayExperimental::get_enumeration(
+                                           ctx, array, attribute.name())) :
+                                   std::nullopt;
+
+            columns.push_back(
+                std::make_shared<SOMAAttribute>(attribute, enumeration));
+        }
+    } else {
+        // All arrays before the introduction of SOMAColumn do not have
+        // composite columns, thus the metadata are trivially contructuble
```
johnkerl, line 103:
```suggestion
        // composite columns, thus the metadata are trivially constructible
```

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -1091,68 +955,87 @@ ArraySchema ArrowAdapter::tiledb_schema_from_arrow_schema(
...
+                ctx, child, type_metadata, platform_config));
             LOG_DEBUG(
                 std::format("[ArrowAdapter] adding attribute {}", child->name));
-
-            schema.add_attribute(attr.first);
         }
     }
 
+    // Unit tests expect dimension order should match the index column schema
+    // and NOT the arrow schema
```
johnkerl, line 991:
```suggestion
    // and NOT the Arrow schema
```

libtiledbsoma/src/utils/common.h
```diff
@@ -50,6 +50,14 @@ const std::string SOMA_GEOMETRY_COLUMN_NAME = "soma_geometry";
 const std::string SOMA_GEOMETRY_DIMENSION_PREFIX = "tiledb__internal__";
 const std::string ARROW_DATATYPE_METADATA_KEY = "dtype";
 
+// SOMAColumn metadata keys
+const std::string TDB_SOMA_SCHEMA_KEY = "tiledb_soma_schema";
+const std::string TDB_SOMA_SCHEMA_VERSION = "0.0.1";
+const std::string TDB_SOMA_SCHEMA_COL_KEY = "tiledb_columns";
+const std::string TDB_SOMA_SCHEMA_COL_TYPE_KEY = "tiledb_column_type";
+const std::string TDB_SOMA_SCHEMA_COL_DIM_KEY = "tiledb_dimensions";
+const std::string TDB_SOMA_SCHEMA_COL_ATTR_KEY = "tiledb_attributes";
```
johnkerl, line 59:
```suggestion
const std::string TILEDB_SOMA_SCHEMA_KEY = "tiledb_soma_schema";
const std::string TILEDB_SOMA_SCHEMA_VERSION = "0.0.1";
const std::string TILEDB_SOMA_SCHEMA_COL_KEY = "tiledb_columns";
const std::string TILEDB_SOMA_SCHEMA_COL_TYPE_KEY = "tiledb_column_type";
const std::string TILEDB_SOMA_SCHEMA_COL_DIM_KEY = "tiledb_dimensions";
const std::string TILEDB_SOMA_SCHEMA_COL_ATTR_KEY = "tiledb_attributes";
```

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -841,11 +846,327 @@ class ArrowAdapter {
         }
     }
 
-    static void set_current_domain_slot(
-        tiledb_datatype_t type,
-        const void* buff,
-        NDRectangle& ndrect,
-        std::string name);
+    /**
+     * Return a copy of the data in a specified column of an arrow table.
```
johnkerl, line 850:
```suggestion
     * Return a copy of the data in a specified column of an Arrow table.
```

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -841,11 +846,327 @@ class ArrowAdapter {
...
+            throw std::runtime_error(
+                "ArrowAdapter::get_table_any_column_by_index: expected "
+                "non-leaf "
+                "node");
+        }
+
+        if (arrow_schema->n_children <= column_index) {
+            throw std::runtime_error(
+                "ArrowAdapter::get_table_any_column_by_index: Column index out "
+                "fo bounds.");
```
johnkerl, line 880:
```suggestion
                "ArrowAdapter::get_table_any_column_by_index: column index out "
                "of bounds.");
```

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -841,11 +846,327 @@ class ArrowAdapter {
...
+        } else {
+            result.push_back(get_table_any_column<S>(
+                selected_array, selected_schema, offset));
+        }
+
+        return result;
+    }
+
+    /**
+     * Read a part of am Arrow array to an std::array and cast in to an
```
johnkerl, line 906:
```suggestion
     * Read a part of an Arrow array to an std::array and cast into an
```

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -841,11 +846,327 @@ class ArrowAdapter {
...
+     * std::any object.
+     *
+     * @example get_table_any_column<3>(array, schema, offset) will return an
+     * std::array<T, 3> where T is the appropriate type based on the Arrow array
+     * format casted as an std::any object. The std::array will skip as many
+     * elements as dictated by `offset` and copy the next 3 from the Arrow
```
johnkerl, line 912:
```suggestion
     * elements as specified by `offset` and copy the next 3 from the Arrow
```

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -841,11 +846,327 @@ class ArrowAdapter {
...
+     * array.
+     *
+     * @tparam S The number of elements to read.
+     *
+     * @param array The Arrow array to read the data from. The array should be
+     * an leaf node with no subarrays.
```
johnkerl, line 918:
```suggestion
     * a leaf node with no subarrays.
```

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -841,11 +846,327 @@ class ArrowAdapter {
...
+     * @param schema The Arrow schema of the given array.
+     * @param offset The number of elements to skip from the beginning of the
+     * Arrow array
+     *
+     * @remarks This method's usage is to extract specific subranges of
+     * ArrowArray data and they come in handy during ArrowSchema -> TileDBSchema
+     * where the arrow array provided has 5 values per dimension and we only
```
johnkerl, line 925:
```suggestion
     * where the Arrow array provided has 5 values per dimension and we only
```

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -841,11 +846,327 @@ class ArrowAdapter {
...
+     * need the last 2 to set the current domain.
+     *
+     * `S` is required to be a template parameter to specify the compile time
```
johnkerl, line 928:
```suggestion
     * `S` is required to be a template parameter to specify the compile-time
```

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -841,11 +846,327 @@ class ArrowAdapter {
...
+     * size of the underlying std::array that will hold the extracted data.
+     *
+     * As to using std::variant, adding more SOMAColumn types would require
+     * changing multiple variants. The use of std::any here is to enable runtime
+     * polymorphism and indirectly introduces a runtime type check (via
+     * any_cast, make_any) between the templated function and the actual
+     * dimension type. std::variant can provide all the above it is just a
+     * different style.
```
johnkerl, line 936:
```suggestion
     * dimension type. std::variant can provide all the above; this is a stylistic choice.
```

libtiledbsoma/src/soma/soma_geometry_column.cc
```diff
@@ -33,6 +33,52 @@
...
+    if (dimension_names.size() % 2 != 0) {
+        throw TileDBSOMAError(std::format(
+            "[SOMAGeometryColumn][deserialize] Invalid number of dimensions. "
+            "Epected number divisible by 2, got {}",
+            dimension_names.size()));
+    }
+    if (attribute_names.size() != 1) {
+        throw TileDBSOMAError(std::format(
+            "[SOMAGeometryColumn][deserialize] Invalid number of attributes. "
+            "Epected 1, got {}",
```
johnkerl, line 66:
```suggestion
            "[SOMAGeometryColumn][deserialize] Invalid number of attributes: "
            "expected 1, got {}",
```

libtiledbsoma/src/soma/soma_geometry_column.cc
```diff
@@ -33,6 +33,52 @@
...
+                              .template get<std::vector<std::string>>();
+
+    std::vector<std::string>
+        attribute_names = soma_schema[TDB_SOMA_SCHEMA_COL_ATTR_KEY]
+                              .template get<std::vector<std::string>>();
+
+    if (dimension_names.size() % 2 != 0) {
+        throw TileDBSOMAError(std::format(
+            "[SOMAGeometryColumn][deserialize] Invalid number of dimensions. "
+            "Epected number divisible by 2, got {}",
```
johnkerl, line 60:
```suggestion
            "[SOMAGeometryColumn][deserialize] Invalid number of dimensions: "
            "expected number divisible by 2, got {}",
```

libtiledbsoma/src/soma/soma_dimension.cc
```diff
@@ -35,6 +35,30 @@
...
+    }
+
+    std::vector<std::string>
+        dimension_names = soma_schema[TDB_SOMA_SCHEMA_COL_DIM_KEY]
+                              .template get<std::vector<std::string>>();
+
+    if (dimension_names.size() != 1) {
+        throw TileDBSOMAError(std::format(
+            "[SOMADimension][deserialize] Invalid number of dimensions. "
+            "Epected 1, got {}",
```
johnkerl, line 53:
```suggestion
            "[SOMADimension][deserialize] Invalid number of dimensions: "
            "expected 1, got {}",
```

libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -207,6 +219,51 @@ void SOMAArray::fill_metadata_cache(std::optional<TimestampRange> timestamp) {
...
+        }
+
+        columns_ = SOMAColumn::deserialize(
+            soma_schema_extension.value(
+                TDB_SOMA_SCHEMA_COL_KEY, nlohmann::json::array()),
+            *ctx_->tiledb_ctx(),
+            *arr_);
+
+    } else {
+        // Non geometry dataframes have trivially constructible columns and do
```
johnkerl, line 260:
```suggestion
        // Non-geometry dataframes have trivially constructible columns and do
```

johnkerl (changes requested):
Thanks @XanthosXanthopoulos !! :) 

Mainly copy-editing mods here

johnkerl (approved):
Thanks @XanthosXanthopoulos !! :)

PR #3407, opened 2024-12-08
[c++] Integrate `SOMAColumn`: Update domain accessors inside `SOMAArray`, part 2

XanthosXanthopoulos:
During the `SOMAArray` object construction and open the `SOMAColumn` objects are initialized and used througout the lifetime of the `SOMAArray` object to access the different domain types for the object. The complexity of transforming the data to an ArrowTable has been transferred  to each `SOMAColumn` and that makes `SOMAArray` implementation agnostic of any column specific restrictions.


jp-dark (approved):
LGTM. Thanks!

PR #3408, opened 2024-12-09
[python] Fix output type when exporting `PointCloudDataFrame` to SpatialData points

jp-dark:
The previous implementation of `to_spatial_data_points` was exporting data to a pandas `DataFrame`, but SpatialData expected a dask `DataFrame`. This type error was not caught in the tests using `spatialdata.models.PointsModel.validate`. This fixes the tests to check the type along with additional validataion by using `spatialdata.models.get_model` instead of the `validate` methods, and updates the exporter to return the correct type.


johnkerl (approved):
🚢

Issue #3353, opened 2024-11-20
[python] Large enumerations in `tiledbsoma.io.from_anndata` (performance issue)


johnkerl:
Scenario:

* User has an AnnData `obs` with a string-valued column with a large number of distinct values (i.e. there is no reason to have this be categorical)
* They do AnnData `adata.write_h5ad(...)(` and later `adata = anndata.read_h5ad(...)`
* AnnData has force-converted this to a categorical within the read-back `adata.obs` Pandas DataFrame -- before we ever see it
* Then, at `tiledbsoma.io.from_anndata`, we are too trusting: given a categorical column in AnnData `adata.obs`, we make a categorical column in TileDB-SOMA
* For cases like 4M distinct values in a categorical column, this faithful/trusting approach is a performance negative

Solution:

* The `tiledbsoma.io` ingest code should inspect each categorical column, and if the number of categories exceeds some threshold (2**12 and 2**15-1 have both been proposed), store it as a simple primitive typed column, i.e, with no enumerations.
* One negative is that outgesting back to AnnData would produce a string column, while the input had categorical. On discussion we believe:
  * Cases of huge enumerations are more likely to be mistakes than intentional
  * If we were to keep a metadata 'outgest hint' for tiledbsoma dataframe string columns, saying, convert back to categorical on outgest -- this would be prohibitive if the input were ordered. And if we can't do it for ordered, we shouldn't do it for non-ordred (that would be confusing).

[[sc-59407]](https://app.shortcut.com/tiledb-inc/story/59407)

PR #3415, opened 2024-12-09
[python] Extend #3354 to categoricals of arbitrary value type

johnkerl:
Follow-up frrom https://github.com/single-cell-data/TileDB-SOMA/pull/3354#discussion_r1876884945 for #3353 [[sc-59407]](https://app.shortcut.com/tiledb-inc/story/59407)

### DIFF
--- a/apis/python/src/tiledbsoma/io/conversions.py
+++ b/apis/python/src/tiledbsoma/io/conversions.py
@@ -74,11 +74,8 @@ def to_tiledb_supported_array_type(name: str, x: _MT) -> _MT:
     # issues in subsequent processing.
     if isinstance(x, pd.Series) and isinstance(x.dtype, pd.CategoricalDtype):
         # Heuristic number
-        if (
-            pandas.api.types.is_string_dtype(x)
-            and len(x.cat.categories) > STRING_DECAT_THRESHOLD
-        ):
-            return x.astype(str)
+        if len(x.cat.categories) > STRING_DECAT_THRESHOLD:
+            return x.astype(x.cat.categories.dtype)
 
     return x
 


bkmartinjr (changes requested):
We need to add a couple of unit tests to make sure it handles the most likely cases of non-string categoricals.  I have occasionally seen in real-world data:
* categorical of float/int
* categorical of bool (which obviously should not trigger the conversion :-)

johnkerl:
@bkmartinjr ready for round two!

bkmartinjr (changes requested):
I believe you can save a copy of each column in the cat->no_cat conversion (and likely the converse as well).  left a note inline to that effect.

Also, is it ever possible that this new conversion code will recieve a table with more than one chunk?  If so, it may result in columns with varied chunk counts, which will then fail when you turn it back into a table. While the ingest code path (i.e., from an AnnData.obs Pandas dataframe) will only have a single chunk, it would be good to either:
* handle multiple chunks explicitly
* or, outlaw them and assert that they won't happen ever
Currently if they do happen, it will result in an error when you try to recompose the modified table, as you will have columns with varied chunkiness.

johnkerl:
@bkmartinjr ready for the next round of review!
https://github.com/single-cell-data/TileDB-SOMA/pull/3415/commits/079ed30f182ebf5bd7c6e29051fbb031ddcc50bb

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1170,24 +1158,43 @@ def is_str_cat_col(field: pa.Field) -> bool:
...
+                # for it, but each chunk does have a .dictionary_decode().
+
+                if isinstance(column, pa.ChunkedArray):
+                    column = pa.chunked_array(
+                        [chunk.dictionary_decode() for chunk in column.chunks]
+                    )
+                elif isinstance(column, pa.DictionaryArray):
+                    column = column.dictionary_decode()
+                else:
+                    column = column.to_pylist()
```
bkmartinjr, line 1175: I'm still confused why this branch needs to round-trip through a Python list.  As I understand the code, you should be able to just leave it as-is, i.e., as an arrow Array.   `Tables.from_pydict` will happily consume that....

bkmartinjr: Actually, thinking about this more -- under what conditions will this branch ever get used?   The column is a dict type, and the first two branches cover chunked and not chunked.  Can this ever execute?

If not, it should be an assert/failure, right, as it indicates that something unexpected made it past line 1163

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1170,24 +1158,43 @@ def is_str_cat_col(field: pa.Field) -> bool:
...
-                )
+                if isinstance(column, pa.ChunkedArray):
+                    column = pa.chunked_array(
+                        [chunk.dictionary_encode() for chunk in column.chunks]
+                    )
+                elif isinstance(column, pa.DictionaryArray):
+                    column = column.dictionary_encode()
+                else:
+                    column = pa.array(
+                        column.to_pylist(),
```
bkmartinjr, line 1190: see comments above - under what cases does this get used?  (aka why does it need to be round-tripped through Python list?)

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1170,24 +1158,43 @@ def is_str_cat_col(field: pa.Field) -> bool:
...
+                # Convert from categorical to non-categorical.  Note that if
+                # this is a pa.ChunkedArray, there is no .dictionary_decode()
+                # for it, but each chunk does have a .dictionary_decode().
+
+                if isinstance(column, pa.ChunkedArray):
+                    column = pa.chunked_array(
+                        [chunk.dictionary_decode() for chunk in column.chunks]
+                    )
+                elif isinstance(column, pa.DictionaryArray):
+                    column = column.dictionary_decode()
```
bkmartinjr, line 1173: coverage indicates missing test for this branch, and a few others

apis/python/src/tiledbsoma/io/conversions.py
```diff
@@ -26,7 +26,7 @@
 _MT = TypeVar("_MT", NPNDArray, sp.spmatrix, PDSeries)
 _str_to_type = {"boolean": bool, "string": str, "bytes": bytes}
 
-STRING_DECAT_THRESHOLD = 4096
+COLUMN_DECAT_THRESHOLD = 4096
```
bkmartinjr, line 29: how was this threshold arrived at?  Might be good to document the thinking?

My thoughts:  there are two issues, the size of the encoded value (category enum) and the "overhead" of loading enums in Core.  There is no extra enum "cost" until you exceed max(signed int16). And the enum overhead only gets punitive in very large numbers (e.g. the motivating test case had 4M enums).

TL;DR - I'd be inclined to make this value `max(int16)` (32767) unless you have evidence that 4096 is warranted due to the second concern (enum overhead in Core).

bkmartinjr (changes requested):
sorry!  A couple of more questions/comments in the latest.  Happy to chat IRT if helpful.

apis/python/src/tiledbsoma/io/conversions.py
johnkerl: > how was this threshold arrived at? Might be good to document the thinking?

It was your original suggestion. Happy to go with newer suggestions from you as you develop new suggestions, as well as to include the reasoning you provide. Thanks @bkmartinjr !

bkmartinjr: oops -- well, short of some benchmarking, I''m not sure we know the perfect value.  But I doubt that there is much loading overhead (ie.., core overhead) difference between 4096 and 32767, so you might as well start with max(int16) to fully utilize the two bytes.  Apologies for my waffling!

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: @bkmartinjr this is operating on Arrow data as a phase within `tiledbsoma.io.from_anndata` / `tiledbsoma.io.from_h5ad`. I don't know how to tweak AnnData objects such that when they're converted to Arrow, their enum columns are fiddle-forced to be either `pa.DictionaryArray` or `pa.ChunkedArray`.

I'll create a separate whitebox test case to test only this routine, with manually manipulated/force-fiddled Arrow columns.

bkmartinjr: got it...the only way I can think to do it is to unit test this function directly (for these cases).   (wasn't sure if that was the same as your white box test?)

johnkerl: @bkmartinjr I'm working on it & hope to have it ready for your review soon -- thanks for waiting!

johnkerl: And yes, "whitebox test" for me means "directly test internals that aren't user-facing"

bkmartinjr (approved):
super clean - thanks for enduring the iterations on this!

johnkerl:
Thanks @bkmartinjr !

PR #3416, opened 2024-12-10
[c++] Fix upgrade-soma-joinid-shape for dataframes having non-standard dimensions

johnkerl:
**Issue and/or context:** As tracked on issue #2407 / [[sc-51048]](https://app.shortcut.com/tiledb-inc/story/51048). Also #3416.

**Changes:**

There should be a closing curly brace here:
https://github.com/single-cell-data/TileDB-SOMA/blob/cbfa704ccc5d326e38a58a8245c8bc328b7b10e5/libtiledbsoma/src/soma/soma_array.cc#L1042
and not here:
https://github.com/single-cell-data/TileDB-SOMA/blob/cbfa704ccc5d326e38a58a8245c8bc328b7b10e5/libtiledbsoma/src/soma/soma_array.cc#L1144

As this correctly describes
https://github.com/single-cell-data/TileDB-SOMA/blob/cbfa704ccc5d326e38a58a8245c8bc328b7b10e5/libtiledbsoma/src/soma/soma_array.cc#L1024-L1025
we need to be copying from core domain (soma maxdomain) to core current domain (soma domain) for all the other dims.

One problem is that the only remaining code which tests this (now that we aren't creating new arrays without current domain anymore) is in
https://github.com/single-cell-data/TileDB-SOMA/blob/cbfa704ccc5d326e38a58a8245c8bc328b7b10e5/apis/python/tests/test_shape.py#L434
-- and the "canned" (saved-off) pre-1.15 xexperiments only have dataframes that are 'standard' i.e. soma_joinid is their only dim.

i'll add some pre-1.15 data to test this.

**Notes for Reviewer:**

When this PR is ready for review, please review it with
https://github.com/single-cell-data/TileDB-SOMA/pull/3416/files?diff=split&w=1 (edited) 
as that will not flag whitespace as red/green, and this PR is (a) just moving a curly brace, with concomitant indentation change, and (b) additional unit-test coverage.


libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -1039,107 +1039,107 @@ void SOMAArray::_set_soma_joinid_shape_helper(
                 }
                 ndrect.set_range<int64_t>(dim_name, 0, newshape - 1);
                 continue;
+            }
```
jp-dark, line 1042: Optional: I think switching this to an if/else instead of the continue would help with clarity and preventing similar bugs in the future.

jp-dark (approved):

PR #3424, opened 2024-12-12
[c++] Add helper methods to convert to ArrowSchema

XanthosXanthopoulos:
Add helpers methods to convert from TileDB dimension and attribute to arrow schema and set current domain.

Issues: [sc-59427](https://app.shortcut.com/tiledb-inc/story/59427/create-an-abstraction-layer-for-soma-columns)

**Notes for Reviewer:**
These methods will be used and tested in the following PR chained of this.

### DIFF
--- a/libtiledbsoma/src/utils/arrow_adapter.cc
+++ b/libtiledbsoma/src/utils/arrow_adapter.cc
@@ -1612,6 +1612,7 @@ void ArrowAdapter::set_current_domain_slot(
     NDRectangle& ndrect,
     std::string name) {
     switch (type) {
+        case TILEDB_STRING_UTF8:
         case TILEDB_STRING_ASCII:
             // Core domain must not be set for string dims.
             // Core current_domain can't _not_ be set for string dims.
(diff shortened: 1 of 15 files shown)


libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -435,6 +435,87 @@ std::unique_ptr<ArrowSchema> ArrowAdapter::arrow_schema_from_tiledb_array(
...
+    arrow_schema->name = strdup(attribute.name().c_str());
+    arrow_schema->metadata = nullptr;
+    arrow_schema->flags = 0;
+    if (attribute.nullable() &&
+        strcmp(attribute.name().c_str(), SOMA_GEOMETRY_COLUMN_NAME.c_str()) !=
+            0) {
+        arrow_schema->flags |= ARROW_FLAG_NULLABLE;
+    } else {
+        arrow_schema->flags &= ~ARROW_FLAG_NULLABLE;
+    }
```
johnkerl, line 474:
```suggestion
    if (attribute.nullable() &&
        attribute.name() != SOMA_GEOMETRY_COLUMN_NAME) {
        arrow_schema->flags |= ARROW_FLAG_NULLABLE;
    } else {
        arrow_schema->flags &= ~ARROW_FLAG_NULLABLE;
    }
```
No need to convert both to C strings and `strcmp` when you can compare as C++ strings

johnkerl (changes requested):
As I've requested before, please always include either a GitHub issue link or a Shortcut story link in every PR's description field

jp-dark (approved):

PR #3425, opened 2024-12-12
[c++] Column abstraction: `SOMADimension`, part 1

XanthosXanthopoulos:
`SOMAColumn` provides an abstraction over TileDB attributes and dimensions and exposes a common interface for all columns regardless of type. Subclasses of `SOMAColumn` can implement complex indexing mechanism through additional dimensions and encapsulate all that logic in one place and make it modular.

Subsequent PRs will add implementation for dimension and attributes.

Throughout this PR there is extensive use of `std::any` to enable polymorphism with the different `SOMAColumn` types while maintaining a templated interface at the abstract `SOMAColumn`.

**Notes for Reviewer:**
This PR introduces the abstract `SOMAColumn` class and the `SOMADimension` concrete class with basic unit tests.


libtiledbsoma/src/soma/soma_column.cc
```diff
@@ -0,0 +1,62 @@
...
+template <>
+std::pair<std::string, std::string>
+SOMAColumn::core_current_domain_slot<std::string>(
+    const SOMAContext& ctx, Array& array) const {
+    try {
+        std::pair<std::string, std::string>
+            current_domain = std::any_cast<std::pair<std::string, std::string>>(
+                _core_current_domain_slot(ctx, array));
+
+        if (current_domain.first == "" && (current_domain.second == "\x7f" ||
```
johnkerl, line 52: There was a comment here about why both `\x7f` and `\xff` -- it needs to be retained. This is very confusing without that comment. (Which is why I put the comment in.)

https://github.com/single-cell-data/TileDB-SOMA/blob/a8349a5cdd295bdb04b965e3e91c3a8959e044e9/libtiledbsoma/src/soma/soma_array.h#L901-L920

johnkerl (commented):
I'm only partway through this PR. I'm still struggling with this (now) 1,875-line PR -- it's doing a lot of things. I'll need to pause and work on some other things for a while, and come back to this. (The other 3 PRs you split out that were smaller were nicely self-contained and self-descriptive and I was able to understand and review them earlier today.)

libtiledbsoma/src/tiledbsoma/tiledbsoma
```diff
@@ -49,6 +49,8 @@
 #include "soma/column_buffer.h"
 #include "soma/soma_array.h"
 #include "soma/soma_collection.h"
+#include "soma/soma_column.h"
+#include "soma/soma_dimension.h"
```
jp-dark, line 53: Looks like you remove these in a future PR. Might as well just not add them in to begin with.

jp-dark (changes requested):
This looks good for the most part, but it's so big it's hard to follow everything. Let's do a code walk through together next time we are both working.

I also added a couple small comments, and John's comment about the missing code comments still needs to be addressed.

libtiledbsoma/src/soma/soma_dimension.h
```diff
@@ -0,0 +1,147 @@
...
+        ArrowArray* array,
+        const std::string& soma_type,
+        std::string_view type_metadata,
+        PlatformConfig platform_config);
+
+    SOMADimension(Dimension dimension)
+        : dimension(dimension) {
+    }
+
+    virtual inline std::string name() const override {
```
jp-dark, line 63: You should drop the `virtual` from these methods. See [ISO C++ Core Guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final) for an explanation.

jp-dark (approved):

PR #3433, opened 2024-12-13
[c++] Fix bug with sliced Arrow-table writes, with string columns

johnkerl:
**Issue and/or context:** still typing

**Changes:**

**Notes for Reviewer:**


libtiledbsoma/src/soma/managed_query.h
```diff
@@ -614,11 +614,17 @@ class ManagedQuery {
...
 
         UserType* buf;
+        // The array->offset is non-zero when we are passed sliced
+        // Arrow tables like arrow_table[:m] or arrow_table[m:].
         if (array->n_buffers == 3) {
             buf = (UserType*)array->buffers[2] + array->offset;
         } else {
             buf = (UserType*)array->buffers[1] + array->offset;
         }
+        uint8_t* validity = (uint8_t*)array->buffers[0];
```
jp-dark, line 624: Nit: This would be a good opportunity to move away from the C-style cast to a named cast (`static_cast` in this case).

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -966,35 +966,45 @@ bool ManagedQuery::_cast_column_aux<std::string>(
...
-        data = array->buffers[2];
-        offset = array->buffers[1];
-        validity = array->buffers[0];
-    } else {
-        data = array->buffers[1];
-        offset = nullptr;
-        validity = array->buffers[0];
+    // A few things in play here:
+    // * Whether the column (array) has 3 buffers (validity, offset, data)
+    //   or 2 (validity, data).
```
jp-dark, line 971: I found this comment confusing since we throw if the column (array) doesn't have 3 buffers.

jp-dark (approved):
Nit: There is a lot of C-style casting in this code which isn't a big deal, but is a little harder to read and little more error prone that more specific cast.

johnkerl:
> Nit: There is a lot of C-style casting in this code which isn't a big deal, but is a little harder to read and little more error prone that more specific cast.

Thanks @jp-dark ! I'll file a follow-up

PR #3438, opened 2024-12-13
[c++] Fix bad merge of #3436

johnkerl:
**Issue and/or context:** From #3436

**Changes:**

**Notes for Reviewer:**


johnkerl:
Dup of #3442

PR #3441, opened 2024-12-14
[python,c++] Improve error handling in `IntIndexer`

bkmartinjr:
**Issue and/or context:**

IntIndexer would do strange things when given malformed inputs.  See for example sc-60689 and sc-60690

IntIndexer also incorrectly handled any arrow array with an offset, and would return incorrect outputs.  See sc-60795

**Changes:**

1. More error checking in pybind-ings, plus tests.  The desired parts of the implicit casting are moved up into the Python code where they are more easily managed.
2. Minimal fix to correctly use array offsets (note: this code needs refactoring to use nanoarrow interfaces)

### DIFF
--- a/apis/python/src/tiledbsoma/_indexer.py
+++ b/apis/python/src/tiledbsoma/_indexer.py
@@ -66,9 +66,21 @@ class IntIndexer:
             Maturing.
         """
         self._context = context
-        self._reindexer = clib.IntIndexer(
-            None if self._context is None else self._context.native_context
+        self._reindexer = (
+            clib.IntIndexer()
+            if self._context is None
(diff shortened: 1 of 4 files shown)


nguyenv:
What is the purpose of adding `noconvert`?

bkmartinjr:
> What is the purpose of adding `noconvert`?

It tells pybind to not try and cast, which removes all kinds of unsupported things (i.e., likely errors) from being done implicitly.  Pybind will try to cast almost any sequence to an ndarray.  If we don't explicitly need that behavior, it is better to just disable it.

johnkerl (approved):

nguyenv (approved):

Issue #3444, opened 2024-12-14
[python] Evaluate scipy 1.15.0rc1


johnkerl:
https://github.com/scipy/scipy/releases/tag/v1.15.0rc1

Issue found: https://github.com/scverse/anndata/issues/1802

Affects our `apis/python/setup.py` as well as https://github.com/TileDB-Inc/tiledbsoma-feedstock

jdblischak:
Status of scipy on conda-forge (**tl;dr** we don't have to immediately worry about this for our conda binaries):

* The latest scipy version currently available is [1.14.1](https://anaconda.org/conda-forge/scipy-tests/files?version=1.14.1)
* There has been some troubleshooting for 1.15.0rc1 in a draft PR opened a few months ago (https://github.com/conda-forge/scipy-feedstock/pull/284), but it is still failing for linux-aarch64

If necessary, after scipy 1.15.0 is released, I can submit a repodata patch to prevent anndata 0.11 from being co-installed with scipy 1.15.

johnkerl:
That would be fabulous, thank you @jdblischak !

h-vetinari:
> but it is still failing for linux-aarch64

It's not failing, emulation is just really slow (and got slower for reasons unrelated to scipy). We'll be able to release 1.15 as soon as it goes GA (by reducing or skipping the tests on aarch; this is tested upstream anyway).

jdblischak:
> It's not failing, emulation is just really slow (and got slower for reasons unrelated to scipy).

@h-vetinari thanks for the clarification!

> We'll be able to release 1.15 as soon as it goes GA (by reducing or skipping the tests on aarch; this is tested upstream anyway).

Wonderful! Thanks for your efforts to push this forward

johnkerl:
Successfully validated TileDB-SOMA Python unit tests with today's `main` and
```
$ pip list|egrep 'anndata|scipy'
anndata                       0.11.2
scipy                         1.15.0
```

jdblischak:
@johnkerl Just to confirm, this means that we can remove the current upper bound on scipy in `setup.py` (and also in the [feedstock recipe](https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/3aa976800d0b7694385e5100156366285c84dcfd/recipe/meta.yaml#L97)), right?

[...]

PR #3445, opened 2024-12-15
[python] Pre-emptively avoid SciPy 1.15, pending a bugfix

johnkerl:
**Issue and/or context:**

Pre-emptively avoids https://github.com/single-cell-data/TileDB-SOMA/pull/new/kerl/pin-scipy-less-than-1.15. Tracking: https://github.com/single-cell-data/TileDB-SOMA/issues/3444.

Note that scipy 1.15.0 is not released yet, but, its 1.15.0rc1 is out, and our next tiledbsoma release (coincidentally also named 1.15.0) is coming out soon, and we want to pre-emptively protect against any time window in which we could suffer from this bug. Once https://github.com/scverse/anndata/issues/1802 is addressed (be it by a bugfix in SciPy or AnnData), we can remove this pin.

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -2152,17 +2152,6 @@ def _find_mean_nnz(matrix: Matrix, axis: int) -> int:
     for lo in range(0, extent, bsz):
         hi = min(extent, lo + bsz)
         coords[axis] = slice(lo, hi)
-        print()
-        print()
-        print("HEY")
-        foo1 = tuple(coords)
-        print(f"TUPLE COORDS <<{foo1}>>")
-        foo2 = matrix[tuple(coords)]
-        print(f"MATRIX TUPLE COORDS <<{foo2}>>")
-        foo3 = matrix[tuple(coords)].nnz
-        print(f"MATRIX TUPLE COORDS NNZ <<{foo3}>>")
-        print()
-        print()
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -2152,6 +2152,17 @@ def _find_mean_nnz(matrix: Matrix, axis: int) -> int:
     for lo in range(0, extent, bsz):
         hi = min(extent, lo + bsz)
         coords[axis] = slice(lo, hi)
+        print()
```
bkmartinjr, line 2155: cleanup?

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -2387,6 +2398,12 @@ def _coo_to_table(
 
     dim_max_size = matrix.shape[stride_axis]
 
+    print()
```
bkmartinjr, line 2401: cleanup

bkmartinjr (changes requested):
Looks OK modulo some debugging code that needs to be removed

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: Sorry about that!

johnkerl:
@bkmartinjr ready for round two!

bkmartinjr (approved):

PR #3402, opened 2024-12-05
[python, c++] Fix misleading `DoesNotExistErrors` in `open` factory

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/3388

The C++ `SOMAObject::open` throws a `SOMAError` exception in the cases where:
- A file does not exist at the requested URI
- The object is not a TileDB object (`INVALID`)
- The TileDB object is not a `SOMAObject` (missing `soma_object_type` in metadata)
- The TileDB array is not a `SOMAArray` type (`soma_object_type` value is invalid)
- The TileDB group is not a `SOMAGroup` type (`soma_object_type` value is invalid)

**Changes:**

- Modify `is_does_not_exist_error` to also take in `SOMAError`
- Add additional error message string checks to catch errors from `[SOMAObject::open]` and rethrow as `DoesNotExistErrors`
- C++ metadata tests had to be rewritten with corrected timestamps
- tiledbsoma-py's `open` factory now uses dictionary to map SOMA type to concrete `clib.SOMAObject.open` method
- tiledbsoma-r's `non_empty_domain` should not use `SOMADataFrame::open` as other `SOMAArray`s may call `non_empty_domain`

PR #3446, opened 2024-12-15
[python] Fix conditional fall-through in `SOMAObject.open`

nguyenv:
**Issue and/or context:**

As reported internally, the pybind11 implemented `SOMAObject.open` contains a conditional fall-through as of #3402.

**Changes:**

Wrap the C++ call in try-catch block where `SOMAObject::open` should already throw an exception for unknown types. In the Pybind11 code, if the conditional falls through then trigger an `assert` for unreachable code.

### DIFF
--- a/apis/python/src/tiledbsoma/soma_object.cc
+++ b/apis/python/src/tiledbsoma/soma_object.cc
@@ -92,8 +92,9 @@ void load_soma_object(py::module& m) {
                 else if (soma_obj_type == "somamultiscaleimage")
                     return py::cast(
                         dynamic_cast<SOMAMultiscaleImage&>(*soma_obj));
-                else{
-                    TPY_ERROR_LOC(*soma_obj_type + " SOMA object type not found");
+                else {
+                    TPY_ERROR_LOC(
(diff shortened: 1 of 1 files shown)


bkmartinjr (approved):

apis/python/src/tiledbsoma/soma_object.cc
```diff
@@ -63,7 +63,11 @@ void load_soma_object(py::module& m) {
                         uri, mode, context, timestamp, clib_type);
                 })();
 
-                auto soma_obj_type = soma_obj->type();
+                try {
+                    auto soma_obj_type = soma_obj->type();
```
johnkerl, line 67: compile error since this isn't defined outside the try

also, IIUC this is a `std::optional<std:string>` so it would be nice to see explicit handling of the has-no-value case

johnkerl (changes requested):

apis/python/src/tiledbsoma/soma_object.cc
bkmartinjr: I get the same

johnkerl (approved):

PR #3449, opened 2024-12-16
[python] Rename `to_spatial_data` -> `to_spatialdata`

jp-dark:
Change name of methods that contain `to_spatial_data` to `to_spatialdata`.


johnkerl (approved):
🚢 🚢 🚢

PR #3453, opened 2024-12-17
[python] Convert from `Optional[Foo]` to `Foo | None`

### LINKED ISSUE #3334: [python] Convert from `Optional[Foo]` to `Foo|None`
Per feedback by @bkmartinjr on #3304:

* Since Python 3.10, and the Google Python style guide, the type | None is recommend, and Optional is not recommended.
* Python 3.9 supports this syntax as long as we do `from __futures__ import annotations`
* Python 3.8 does not, but as of #2141 we no longer support Python 3.8

Thus, now is a good time for us to adapt to evolving style-guide recommendations in this regard.

Also needs to be done for https://github.com/single-cell-data/SOMA

mojaveazure:
Convert `Optional` to [union types](https://docs.python.org/3/library/stdtypes.html#union-type)

[SC-60887](https://app.shortcut.com/tiledb-inc/story/60887)

resolves #3334

### DIFF
--- a/apis/python/setup.py
+++ b/apis/python/setup.py
@@ -13,6 +13,8 @@
 # Based on ideas from https://github.com/pybind/cmake_example
 # The `bld` script here is reused for pip install, CI, and local builds.
 
+from __future__ import annotations
+
 import ctypes
 import os
 import pathlib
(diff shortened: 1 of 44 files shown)


johnkerl (approved):
Thank you @mojaveazure !! :)

PR #3468, opened 2024-12-18
[c++, python] `fastercsx`: improved duplicate coordinate handling

bkmartinjr:
**Issue and/or context:**

the `fastercsx` module will generate a malformed scipy sparse matrix under a specific set of conditions not usually seen with SOMA data (matrix contains duplicate coordinates). This PR detects that condition and correctly initializes the SciPy sparse matrix.  Tested on scipy==1.15.0r1 for good measure.

**Changes:**
* correct dup handling
* remove a small bit of redundant dead code (`from_pjd` method)
* fix sc-61043 - incorrect handling of a multi-chunk coordinate if first chunk is empty


apis/python/src/tiledbsoma/_fastercsx.py
```diff
@@ -156,12 +157,12 @@ def dtype(self) -> npt.DTypeLike:
     def to_scipy(
         self, index: slice | None = None
     ) -> scipy.sparse.csr_matrix | scipy.sparse.csc_matrix:
-        """Extract a slice on the compressed dimension and return as a dense
```
johnkerl, line 159: good catch 😎

apis/python/src/tiledbsoma/_fastercsx.py
```diff
@@ -86,25 +97,15 @@ def from_ijd(
         compress_coo(
             context.native_context, (n_major, n_minor), i, j, d, indptr, indices, data
         )
+        no_duplicates = None  # aka,unknown
```
johnkerl, line 100:
```suggestion
        no_duplicates = None  # aka, unknown
```

johnkerl (approved):
🚢

PR #3469, opened 2024-12-18
[r] Write group-level string metadata as `TILEDB_STRING_UTF8`

### LINKED ISSUE #2698: [r/python] R and Python create groups with incompatible SOMA metadata

mojaveazure:
String group-level metadata was previously encoded using `TILEDB_CHAR` or `TILEDB_STRING_ASCII`; however, this resulting in the metadata being read in as `bytes` in the Python API instead of as `str`. The Python API already [encodes all strings (`str`) as `TILEDB_STRING_UTF8`](https://github.com/single-cell-data/TileDB-SOMA/blob/884342a1ceb994d677c52c74ba2d789fc4e208d4/apis/python/src/tiledbsoma/common.cc#L211-L223) so this PR brings the R API in-line with the Python API

[SC-61001](https://app.shortcut.com/tiledb-inc/story/61001)

adresses #2698

### DIFF
--- a/apis/r/src/groups.cpp
+++ b/apis/r/src/groups.cpp
@@ -254,7 +254,7 @@ void c_group_put_metadata(
             std::string s(v[0]);
             // We use TILEDB_CHAR interchangeably with TILEDB_STRING_ASCII is
             // this best string type?
-	    // Use TILEDB_STRING_UTF8 for compatibility with Python API
+            // Use TILEDB_STRING_UTF8 for compatibility with Python API
             xp->grpptr->set_metadata(
                 key, TILEDB_STRING_UTF8, s.length(), s.c_str());
             break;
--- a/apis/r/src/groups.cpp
+++ b/apis/r/src/groups.cpp
@@ -254,8 +254,9 @@ void c_group_put_metadata(
             std::string s(v[0]);
             // We use TILEDB_CHAR interchangeably with TILEDB_STRING_ASCII is
             // this best string type?
+	    // Use TILEDB_STRING_UTF8 for compatibility with Python API
             xp->grpptr->set_metadata(
-                key, TILEDB_STRING_ASCII, s.length(), s.c_str());
+                key, TILEDB_STRING_UTF8, s.length(), s.c_str());
             break;
         }
         case LGLSXP: {  // experimental: map R logical (ie TRUE, FALSE, NA) to
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -6,7 +6,7 @@ Description: Interface for working with 'TileDB'-based Stack of Matrices,
     like those commonly used for single cell data analysis. It is documented at
     <https://github.com/single-cell-data>; a formal specification available is at
     <https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md>.
-Version: 1.16.99
+Version: 1.16.99.1
 Authors@R: c(
     person(given = "Aaron", family = "Wolen",
            role = c("cre", "aut"), email = "aaron@tiledb.com",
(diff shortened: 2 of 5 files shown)


apis/r/src/groups.cpp
```diff
@@ -254,8 +254,9 @@ void c_group_put_metadata(
             std::string s(v[0]);
             // We use TILEDB_CHAR interchangeably with TILEDB_STRING_ASCII is
             // this best string type?
+	    // Use TILEDB_STRING_UTF8 for compatibility with Python API
             xp->grpptr->set_metadata(
-                key, TILEDB_STRING_ASCII, s.length(), s.c_str());
+                key, TILEDB_STRING_UTF8, s.length(), s.c_str());
```
johnkerl, line 259: Please fix the indentation

johnkerl (approved):
LGTM

The facts that no unit tests have been added -- and that no existing unit-test cases are now failing -- are reassuring confirmation that we do need the cross-language tests here https://github.com/single-cell-data/TileDB-SOMA/issues/2698 to clinch the typing compatibilities

bkmartinjr (commented):
I'd like to see a test run regularly - is that something for this PR, or follow-up backlog?

johnkerl:
> I'd like to see a test run regularly - is that something for this PR, or follow-up backlog?

@bkmartinjr please see

* https://github.com/single-cell-data/TileDB-SOMA/pull/3469#pullrequestreview-2510634967
* #2698 
* https://github.com/single-cell-data/TileDB-SOMA/issues/2698#issuecomment-2284279154

@mojaveazure the closing criterion for #2698 is that cross-language test

apis/r/src/groups.cpp
johnkerl: For the record -- this is a tab vs spaces thing BTW

johnkerl (approved):
@mojaveazure awesome, thank you!

PR #3476, opened 2024-12-18
[python] Re-enable `tiledbsoma.ExperimentAxisQuery`

johnkerl:
**Issue and/or context:** After 1.15.0 was tagged, we found that users can no longer instantiate `tiledbsoma.ExperimentAxisQuery`.

Repro:

<details>

```
import scanpy as sc
import tiledbsoma
import tiledbsoma.io
import tempfile

pbmc3k = sc.datasets.pbmc3k()
sc.pp.calculate_qc_metrics(pbmc3k, inplace=True)

tmpdir = tempfile.TemporaryDirectory().name
tiledbsoma.io.from_anndata(tmpdir, anndata=pbmc3k, measurement_name="RNA")

exp = tiledbsoma.Experiment.open(tmpdir)

exp.obs.read().concat().to_pandas()
exp.ms["RNA"].var.read().concat().to_pandas()
exp.ms["RNA"].X["data"].read().coos().concat()
exp.obs.read(
    coords=[slice(0, 99)],
    value_filter="total_counts > 4000",
    column_names=["soma_joinid", "obs_id", "total_counts"],
).concat().to_pandas()
exp.ms["RNA"].var.read(
    value_filter="var_id in ['ANXA1', 'IFI44', 'IFI44L', 'OAS1']",
    column_names=["var_id", "gene_ids", "mean_counts"],
).concat().to_pandas()

with tiledbsoma.ExperimentAxisQuery(
    experiment=exp,
    measurement_name="RNA",
    obs_query=tiledbsoma.AxisQuery(
        value_filter="n_genes_by_counts > 1000",
    ),
    var_query=tiledbsoma.AxisQuery(
        value_filter="n_cells_by_counts > 100",
    ),
) as query:
    assert (query.n_obs, query.n_vars) == (532, 4722)
    query.obs().concat().to_pandas()
    query.X(layer_name="data").tables().concat().to_pandas()
    query.to_anndata(X_name="data")
```

</details>

This should produce no output and raise no exceptions. Actual behavior in 1.15.0, which is a regression from 1.14.5:

```
  File "/home/ubuntu/git/single-cell-data/TileDB-SOMA/x.py", line 35, in <module>
    with tiledbsoma.ExperimentAxisQuery(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: ExperimentAxisQuery() takes no arguments
```

**Changes:**

* Since on #3307 we moved the impl from `somacore` to `tiledbsoma._query`, have `tiledbsoma` re-export the impl from where it is now located
* Update our unit-test logic to use the same API as exposed to users, namely, `tiledbsoma.ExperimentAxisQuery`

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/__init__.py
+++ b/apis/python/src/tiledbsoma/__init__.py
@@ -140,10 +140,10 @@ from somacore import (
     CoordinateSpace,
     IdentityTransform,
     ScaleTransform,
-    Axis,
     UniformScaleTransform,
 )
 from ._query import (
+    Axis,
     ExperimentAxisQuery,
 )
 from somacore.options import ResultOrder
--- a/apis/python/src/tiledbsoma/__init__.py
+++ b/apis/python/src/tiledbsoma/__init__.py
@@ -133,16 +133,17 @@ except ImportError:
     except OSError:
         # Otherwise try loading by name only.
         ctypes.CDLL(libtiledbsoma_name)
-
 from somacore import (
-    Axis,
-    CoordinateSpace,
     AffineTransform,
-    ScaleTransform,
-    IdentityTransform,
-    UniformScaleTransform,
     AxisColumnNames,
     AxisQuery,
+    CoordinateSpace,
+    IdentityTransform,
+    ScaleTransform,
+    Axis,
+    UniformScaleTransform,
+)
+from ._query import (
     ExperimentAxisQuery,
 )
 from somacore.options import ResultOrder
--- a/apis/python/src/tiledbsoma/__init__.py
+++ b/apis/python/src/tiledbsoma/__init__.py
@@ -135,6 +135,7 @@ except ImportError:
         ctypes.CDLL(libtiledbsoma_name)
 from somacore import (
     AffineTransform,
+    Axis,
     AxisColumnNames,
     AxisQuery,
     CoordinateSpace,
@@ -143,7 +144,6 @@ from somacore import (
     UniformScaleTransform,
 )
 from ._query import (
-    Axis,
     ExperimentAxisQuery,
 )
 from somacore.options import ResultOrder
--- a/apis/python/tests/test_experiment_query.py
+++ b/apis/python/tests/test_experiment_query.py
@@ -14,10 +14,10 @@ from scipy import sparse
 from somacore import AxisQuery, options
 
 import tiledbsoma as soma
-from tiledbsoma import SOMATileDBContext, _factory, pytiledbsoma
+from tiledbsoma import ExperimentAxisQuery, SOMATileDBContext, _factory, pytiledbsoma
 from tiledbsoma._collection import CollectionBase
 from tiledbsoma._experiment import Experiment
-from tiledbsoma._query import Axis, ExperimentAxisQuery
+from tiledbsoma._query import Axis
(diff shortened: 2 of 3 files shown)


apis/python/tests/test_experiment_query.py
```diff
@@ -14,10 +14,10 @@
 from somacore import AxisQuery, options
 
 import tiledbsoma as soma
-from tiledbsoma import SOMATileDBContext, _factory, pytiledbsoma
+from tiledbsoma import ExperimentAxisQuery, SOMATileDBContext, _factory, pytiledbsoma
 from tiledbsoma._collection import CollectionBase
 from tiledbsoma._experiment import Experiment
-from tiledbsoma._query import Axis, ExperimentAxisQuery
+from tiledbsoma._query import Axis
```
bkmartinjr, line 20: IMHO, all of these tests of "public" API should import in the same manner a user would import them (to catch this bug in the future).

In other words, do:
> from tiledbsoma import Axis

apis/python/src/tiledbsoma/__init__.py
```diff
@@ -133,16 +133,17 @@
...
 from somacore import (
-    Axis,
-    CoordinateSpace,
     AffineTransform,
-    ScaleTransform,
-    IdentityTransform,
-    UniformScaleTransform,
     AxisColumnNames,
     AxisQuery,
+    CoordinateSpace,
```
bkmartinjr, line 140: shouldn't these remain in sorted order?  (lint)

apis/python/tests/test_experiment_query.py
johnkerl: @bkmartinjr I tried that and got an error. But I'm not seeing it now when I try it just now ... 🤔

apis/python/src/tiledbsoma/__init__.py
johnkerl: Weird -- pre-commit ran cleanly ... 👀

apis/python/tests/test_experiment_query.py
bkmartinjr: a failure would indicate some other issue, so I don't recommend the "workaround" ....

the top-level namespace of the package should just "work"

apis/python/src/tiledbsoma/__init__.py
bkmartinjr: that is worth investigating as well (at some point) ...

johnkerl: 💯

apis/python/tests/test_experiment_query.py
johnkerl: Totally agreed!

johnkerl: I think probably I did the import of `Axis` in the test file but didn't do the re-export in the `__init__.py` -- that would explain it

Anyway I have it in both files & all is well now

Thanks @bkmartinjr !

apis/python/src/tiledbsoma/__init__.py
ryan-williams: Good catch, https://github.com/single-cell-data/TileDB-SOMA/pull/3495 should fix this.

PR #3482, opened 2024-12-19
[c++] Use const references and `std::string_view` in `create` function

XanthosXanthopoulos:
Replace unnecessary `std::move` and `std::string` with const references and `std::string_view` where possible in all the static create functions.


johnkerl (approved):
Fantastic -- thank you @XanthosXanthopoulos ! :)

PR #3493, opened 2024-12-20
make EAQ inherit somacore, restore EAQ.close

ryan-williams:
**Issue and/or context:**
- [#3307](https://github.com/single-cell-data/TileDB-SOMA/pull/3307)/[somacore#244](https://github.com/single-cell-data/SOMA/pull/244) moved ExperimentAxisQuery over from somacore, but mistakenly didn't inherit from the somacore.EAQ ABC.
- It also made `EAQ.close` a no-op (per [this thread](https://github.com/single-cell-data/TileDB-SOMA/pull/3307#pullrequestreview-2443384746)).
- [#3359](https://github.com/single-cell-data/TileDB-SOMA/pull/3359) removed `EAQ.close` altogether, which should have triggered a pre-commit error since EAQ was no longer implementing the somacore ABC.
- This corrects both issues. I also verified that adding the inheritance causes the missing `close` to fail pre-commit.

### DIFF
--- a/apis/python/src/tiledbsoma/_query.py
+++ b/apis/python/src/tiledbsoma/_query.py
@@ -188,7 +188,7 @@ class AxisQueryResult:
         )
 
 
-class ExperimentAxisQuery:
+class ExperimentAxisQuery(query.ExperimentAxisQuery):
     """Axis-based query against a SOMA Experiment.
 
     ExperimentAxisQuery allows easy selection and extraction of data from a
@@ -568,6 +568,8 @@ class ExperimentAxisQuery:
         return sdata
 
     # Context management
+    def close(self) -> None:
+        pass
 
     def __enter__(self) -> Self:
         return self


johnkerl (changes requested):
Please restore `test_query_cleanup` from
https://github.com/single-cell-data/TileDB-SOMA/blob/1.14.5/apis/python/tests/test_experiment_query.py#L541-L567
but omitting the threadpool-is-None/not-None checks per
https://github.com/single-cell-data/TileDB-SOMA/pull/3307#discussion_r1848851869

We need unit-test coverage that we have a `.close` method on `ExperimentAxisQuery`

apis/python/src/tiledbsoma/_query.py
```diff
@@ -568,6 +568,8 @@ def to_spatialdata(  # type: ignore[no-untyped-def]
         return sdata
 
     # Context management
```
johnkerl, line 570:
```suggestion
    # Context management.
    # Currently a no-op, but, part of public API so we retain it.
```

PR #3495, opened 2024-12-20
[python] `tiledbsoma.__init__.py`: restore lint checks

ryan-williams:
Flagged in [#3476 (comment)](https://github.com/single-cell-data/TileDB-SOMA/pull/3476/files#r1890926269).

`# ruff: noqa` midway through the file was actually turning off `ruff` for the whole file.

This replaces that with a couple of more specific ruff rule disables, which only apply to relevant sections.


johnkerl (approved):
Thanks @ryan-williams !

PR #3510, opened 2025-01-02
[python] Add support for Visium v1

jp-dark:
**Issue and/or context:** [sc-61280](https://app.shortcut.com/tiledb-inc/story/61280/add-support-for-visium-v1)

**Changes:**
1. Explicitly set header columns for 
2. Throw error if Visium version number is unknown.
3. Make check for Visium filename version specific (this can be over-ridden by the user).

### DIFF
--- a/apis/python/src/tiledbsoma/io/spatial/ingest.py
+++ b/apis/python/src/tiledbsoma/io/spatial/ingest.py
@@ -565,6 +565,7 @@ def from_visium(
                     with _write_visium_spots(
                         loc_uri,
                         input_paths.tissue_positions,
+                        input_paths.major_version,
                         pixels_per_spot_diameter,
                         obs_df,
                         obs_id_name,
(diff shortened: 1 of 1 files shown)


apis/python/src/tiledbsoma/io/spatial/ingest.py
```diff
@@ -189,30 +182,22 @@ def from_spatial_folder(
...
             except (KeyError, ValueError):
-                warnings.warn(
+                raise ValueError(
                     "Unable to determine Space Ranger vesion from gene expression file."
                 )
-        major_version = version[0] if isinstance(version, tuple) else version
 
+        # Find the tissue positions file path if it wasn't supplied.
         if tissue_positions is None:
+            major_version = version[0] if isinstance(version, tuple) else version
```
ktsitsi, line 191: Why not using the `property` instead?

ktsitsi (approved):
LGTM just a misc in the comment.

apis/python/src/tiledbsoma/io/spatial/ingest.py
jp-dark: The class wasn't created yet.

PR #3516, opened 2025-01-06
[python] Link lifetimes of `SOMAArray` and `ManagedQuery`

nguyenv:
**Issue and/or context:**

[[sc-61128](https://app.shortcut.com/tiledb-inc/story/61128/python-1-15-regression-on-table-iterator)]

**Changes:**

- Introduce a high-level `ManagedQuery` class, analogous to the other `SOMAObject` classes, serving as a thin wrapper around the `clib.ManagedQuery` pybind11 object. This class links together the lifetime of `clib.ManagedQuery` to the high-level `SOMAArray` object (not `clib.SOMAArray`)
- Clean up the array setters in `_util.py` to use the `_array` stored in `ManagedQuery`, eliminating the need to pass the array in as a separate argument
- Add a test to confirm the resolution of the previously reported bug and to ensure separate `SOMArray` instances at the same URI iterate using distinct `ManagedQuery` instances

**Notes for Reviewer:**

Another solution for the reported bug is to pass a `shared_ptr<SOMAArray>` instead of `unique_ptr` into the `clib.ManagedQuery` constructor. This is similar to what we already do for `clib.SOMAContext`. We attempted this solution, but it was unfeasible to a bug in the Pybind11 library. A pull request addressing this issue is currently under review: https://github.com/pybind/pybind11/pull/2839. Given the current circumstances, the most effective solution is to implement this high-level thin wrapper class.

### DIFF
--- a/apis/python/src/tiledbsoma/_dense_nd_array.py
+++ b/apis/python/src/tiledbsoma/_dense_nd_array.py
@@ -22,7 +22,7 @@ from . import pytiledbsoma as clib
 from ._arrow_types import pyarrow_to_carrow_type
 from ._common_nd_array import NDArray
 from ._exception import SOMAError, map_exception_for_create
-from ._read_iters import TableReadIter
+from ._read_iters import ManagedQuery, TableReadIter
 from ._tdb_handles import DenseNDArrayWrapper
 from ._types import OpenTimestamp, Slice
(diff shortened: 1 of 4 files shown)


nguyenv:
[sc-61128]

bkmartinjr (approved):
did a visual code review only - it looks great.

PR #3526, opened 2025-01-07
[python] Unpin `scipy`

johnkerl:
Follow-up from https://github.com/single-cell-data/TileDB-SOMA/issues/3444#issuecomment-2575991387


nguyenv (approved):

johnkerl:
Thanks @nguyenv !

PR #3534, opened 2025-01-08
[c++] Update minimum macOS deployment target

XanthosXanthopoulos:
For the C++20 migration to work we need to increase the minimum macOS deployment target to 13.3 to have support for floating-point formatter. 

#3154

[[sc-57301]](https://app.shortcut.com/tiledb-inc/story/57301/use-c-20)

### DIFF
--- a/libtiledbsoma/CMakeLists.txt
+++ b/libtiledbsoma/CMakeLists.txt
@@ -114,7 +114,7 @@ if(APPLE)
   set(CMAKE_MACOSX_RPATH ON)
 
   # Set minimum macOS version to enable certain C++20 features
-  set(CMAKE_OSX_DEPLOYMENT_TARGET 13.3)
+  # set(CMAKE_OSX_DEPLOYMENT_TARGET 13.3)
 
   # Don't allow macOS .frameworks to be used for dependencies.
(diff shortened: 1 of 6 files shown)


libtiledbsoma/CMakeLists.txt
```diff
@@ -113,6 +113,9 @@ endif()
 if(APPLE)
   set(CMAKE_MACOSX_RPATH ON)
 
+  # Set minimum macOS version to enable certain C++20 features
+  # set(CMAKE_OSX_DEPLOYMENT_TARGET 13.3)
+
```
johnkerl, line 118: Does this comment need to be here?

johnkerl (approved):
🚢 once CI is ☘️ -- thank you @XanthosXanthopoulos for sticking with this!!! :)

jdblischak (approved):

libtiledbsoma/CMakeLists.txt
XanthosXanthopoulos: I would like to enable it eventually (that's why I have it - as a reminder) but I do not know what effect it would have on conda build

PR #3544, opened 2025-01-10
[C++] Fix `ManagedQuery` segv: ctor/dtor order fix

bkmartinjr:
**Issue and/or context:**

But reported via chanzuckerberg/cellxgene-census#1327 - root cause was incorrect destructor ordering in the new ManagedQuery class.

The tiledb::Array class _requires_ a tiledb::Context to destruct, but only holds a `std::reference_wrapper<Context>`.  It is holder's responsibility to ensure that the Array's context is alive when the array is destructed.

ManagedQuery just had the context and array shared pointers in the wrong order, causing the context to destruct first in some cases.

[sc-61614]

**Changes:**

Change declaration order of members, thereby correctly ordering the dtor calls.

### DIFF
--- a/apis/python/tests/test_sparse_nd_array.py
+++ b/apis/python/tests/test_sparse_nd_array.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import contextlib
 import datetime
+import gc
 import itertools
 import json
 import operator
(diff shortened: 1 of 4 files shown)


apis/python/tests/test_sparse_nd_array.py
```diff
@@ -1979,3 +1980,27 @@ def test_iter(tmp_path: pathlib.Path):
...
+    ) as write_arr:
+        write_arr.write(arrow_tensor)
+
+    def test(path, tiledb_config):
+        context = soma.SOMATileDBContext().replace(tiledb_config=tiledb_config)
+        X = soma.SparseNDArray.open(path, context=context, mode="r")
+        mq = X.read().tables()
+        return mq
+
+    for _ in range(100):
```
nguyenv, line 1998: Is this loop here because the error is intermittent?

libtiledbsoma/src/soma/managed_query.h
```diff
@@ -499,12 +499,16 @@ class ManagedQuery {
         const CurrentDomain& current_domain, bool is_read);
     void _fill_in_subarrays_if_dense_without_new_shape(bool is_read);
 
-    // TileDB array being queried.
-    std::shared_ptr<Array> array_;
+    // NB: the Array dtor REQUIRES that this context be alive, so member
```
johnkerl, line 502: Thank you for leaving this comment! Very helpful for others in the future to understand this non-obvious failure.

johnkerl (approved):

apis/python/tests/test_sparse_nd_array.py
bkmartinjr: It is not intermittent in my setup. But Python doesn't give any real guarantees about GC behavior, so it is possible that multiple calls might be required to trip over it in some Python implementation.  I was just being cautious - trying to be triple-sure we would eventually trigger GC of the reference holders

nguyenv (approved):

apis/python/tests/test_sparse_nd_array.py
bkmartinjr: done

PR #3548, opened 2025-01-10
[python] Handle string metadata with `NULL` bytes

XanthosXanthopoulos:
Python string representation doesn't include a null byte thus strings like `\x00` are possible. Passing such string down to C++ causes issues when using `char*`. 

Additionally when converting an empty string from python to a byte array, `numpy` adds a `NULL` byte. This is not the case when the string is not empty even when the dtype specified is `zero-terminated bytes`.

Reading an empty string metadata entry from a TileDB Array returns a `NULL` data pointer but the value size is 1 which when converted to a python string results in `\x00` string.

[[sc-61094](https://app.shortcut.com/tiledb-inc/story/61094/python-metadata-keys-with-value-x00-saved-as-empty-string), [sc-61093](https://app.shortcut.com/tiledb-inc/story/61093/python-writing-zero-length-string-into-metadata-value-incorrectly-stores-length-1-string-containing-ascii-0)]

### DIFF
--- a/apis/python/src/tiledbsoma/common.cc
+++ b/apis/python/src/tiledbsoma/common.cc
@@ -79,6 +79,15 @@ std::unordered_map<std::string, tiledb_datatype_t> _np_name_to_tdb_dtype = {
     {"bool", TILEDB_BOOL},
 };
 
+void sanitize_string(const std::string data) {
+    for (auto& byte : data) {
+        if (static_cast<uint8_t>(byte) == 0) {
+            throw TileDBSOMAError(
+                "[sanitize_string] String contains NULL bytes");
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


bkmartinjr (changes requested):
Recommend enhancing the key/value unit tests to also include unicode code points that are outside of ASCII values.

I am also unsure why we are disallowing Unicode strings containing the Unicode NULL codepoint (U+0000). Python and most (all?) Unicode string implementations seem to support this.  Is this a TileDB Core limitation?

XanthosXanthopoulos:
TileDB Core is not limiting metadata values to UTF-8. Enabling support for UTF-16 and UTF-32 can be done in Python and R respectively since it is a byte encoding issue primarily.

XanthosXanthopoulos:
Merged as part of #3558

PR #3558, opened 2025-01-12
[python] Add Unicode metadata support 

XanthosXanthopoulos:
A follow-up PR to #3548 that adds support for Unicode metadata. 

UTF-8 support is not enough to store 1D numpy string arrays because Unicode arrays are by default UTF-32 encoded and can contain NULL bytes. 

Additionally improved robustness of metadata writing by attempting to write all modifications and collecting errors throughout the process and raising an exception cumulative at the end, rather than throwing midway. 

[sc-61687](https://app.shortcut.com/tiledb-inc/story/61687/python-add-unicode-metadata-support)

### DIFF
--- a/apis/python/src/tiledbsoma/_tdb_handles.py
+++ b/apis/python/src/tiledbsoma/_tdb_handles.py
@@ -20,7 +20,6 @@ from typing import (
     List,
     Mapping,
     MutableMapping,
-    MutableSequence,
     Sequence,
     Tuple,
     Type,
@@ -811,7 +810,7 @@ class MetadataWrapper(MutableMapping[str, Any]):
             return
         # Only try to get the writer if there are changes to be made.
 
-        errors: MutableSequence[Exception] = []
+        errors: list[Exception] = []
         for key, mod in self._mods.items():
             try:
                 if mod in (_DictMod.ADDED, _DictMod.UPDATED):
--- a/apis/python/src/tiledbsoma/_tdb_handles.py
+++ b/apis/python/src/tiledbsoma/_tdb_handles.py
@@ -20,6 +20,7 @@ from typing import (
     List,
     Mapping,
     MutableMapping,
+    MutableSequence,
     Sequence,
     Tuple,
     Type,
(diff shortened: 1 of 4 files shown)


apis/python/src/tiledbsoma/common.cc
```diff
@@ -135,7 +135,7 @@ tiledb_datatype_t np_to_tdb_dtype(py::dtype type) {
     if (kind.is(py::str("S")))
         return TILEDB_STRING_ASCII;
     if (kind.is(py::str("U")))
-        return TILEDB_STRING_UTF8;
+        return TILEDB_STRING_UTF32;
```
johnkerl, line 138: This looks weird and it looks wrong -- even though it isn't -- please add a clear comment why we need to do this for multi-package API compliance.

apis/python/src/tiledbsoma/common.cc
```diff
@@ -229,7 +241,49 @@ void set_metadata(
     if (value_buffer.ndim != 1)
         throw py::type_error("Only 1D Numpy arrays can be stored as metadata");
 
-    auto value_num = is_tdb_str(value_type) ? value.nbytes() : value.size();
+    auto value_num = is_tdb_str(value_type) ?
```
bkmartinjr, line 244: is it ever possible for value_type to be `== UTF8`?  UTF8 can have a variable number of bytes per character, so this will probably act strangely....

In theory, it would also be possible for numpy to generate an array with UTF-16 and other types.  Suggest either checking/throwing if you get an unexpected type, or adding tests and explicit handling.

bkmartinjr (changes requested):
Recommend a bit more error checking in the C++ code (just defensive checks for unexpected types), and some additional tests for native numpy types that could easily occur in real-world usage

XanthosXanthopoulos:
A few changes are introduces with the latest commit. All string are standardized to be UTF-8 encoded by passing the encoded byte array directly and not allowing numpy to do the encoding. This however introduced issues when trying to store `bytes` which are now handled differently and encoded as a `void` numpy array.

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -809,21 +810,33 @@ def _write(self) -> None:
             # There were no changes (e.g., it's a read handle).  Do nothing.
             return
         # Only try to get the writer if there are changes to be made.
+
+        errors: MutableSequence[Exception] = []
```
bkmartinjr, line 814: just a style comment:  if it is coded as a list (ie., `= []`), why declare a looser type.  Better just to define the type as `list[Exception]` and take advantage of stronger type checking

apis/python/tests/test_metadata.py
```diff
@@ -188,6 +188,11 @@ def non_soma_metadata(obj) -> Dict[str, Any]:
         1.00000001,
         -3.1415,
         "",
+        "\x00",
```
bkmartinjr, line 191: Need to add params that test the `TILEDB_BLOB` code path as well.

Although looking at the code, I can't see how you would ever get a TILED_BLOB value_type in set_metadata. From what I can see, only the metadata read path will ever see this type (so that would require using TileDB-Py to set some metadata, and then confirm it reads correctly as a python bytes object).

apis/python/src/tiledbsoma/common.cc
```diff
@@ -229,7 +251,43 @@ void set_metadata(
     if (value_buffer.ndim != 1)
         throw py::type_error("Only 1D Numpy arrays can be stored as metadata");
 
-    auto value_num = is_tdb_str(value_type) ? value.nbytes() : value.size();
+    auto value_num = (is_tdb_str(value_type) || value_type == TILEDB_BLOB) ?
```
bkmartinjr, line 254: I may be missing something, but how would you ever get a TILEDB_BLOB at this point?  The `np_to_tdb_dtype()` code doesn't look like it will ever generate that type.

bkmartinjr (commented):
the UTF-8 code looks great.  Couple of open questions about blob/bytes:
* is this used in the set_metadata path?
* If so, can we test it?

apis/python/tests/test_metadata.py
XanthosXanthopoulos: That is correct. Blob type metadata are read only and need to be written in some other way.

bkmartinjr: How would you feel about using `tiledb` to write one for testing?   Probably requires a broader discussion, as it would introduce a new package dependency (for test only), along with the issue of keeping their Core versions in sync, etc.

I don't think it is worth holding up this PR for the test.  But I'm a big fan of testing all the branches :-)

@johnkerl - any views on the relative cost/benefit of reintroducing tiledb-py dependency for test?

bkmartinjr (approved):
Marking as approved, with the caveats:
* one style issue in _tdb_handles
* open question about how we can test the BLOB path in set_metadata

apis/python/src/tiledbsoma/common.cc
XanthosXanthopoulos: Correct, I will remove it.

apis/python/tests/test_metadata.py
johnkerl: @bkmartinjr sorry, I just now saw this comment

Let's discuss internally please

PR #3563, opened 2025-01-15
[python] Set `soma_joinid` to always be an index column in the `PointCloudDataFrame` class

jp-dark:
**Issue and/or context:** [sc-61690](https://app.shortcut.com/tiledb-inc/story/61690/add-soma-joinid-back-to-pointclouddataframe)

**Changes:**
This changes the `soma_joinid` to be an `index_column` in the `PointCloudDataFrame`. The initial implementation allowed `soma_joinid` to be an optional column. It was then dropped. Work on querying between observations/measurements and spatial data has motivated adding it back to improve querying by `obs_id` and `var_id`.


apis/python/src/tiledbsoma/_point_cloud_dataframe.py
```diff
@@ -184,9 +185,12 @@ def create(
         else:
             ndom = len(soma_domain)
             nidx = len(index_column_names)
+            # Allow use to set domain for just spatial dimensions. Use default for
```
XanthosXanthopoulos, line 188: Is this a TODO item?

jp-dark: It's something I added and removed. Looks like I forgot to remove the comment.

PR #3565, opened 2025-01-15
[python] Throw error in `SOMAVFSFilebuf.open` on file-not-found

nguyenv:
**Issue and/or context:**

[[sc-61909](https://app.shortcut.com/tiledb-inc/story/61909/misleading-error-message-in-soma-ingestor-when-input-file-is-not-found)]

**Changes:**

Fix misleading error by throwing an explicit error in `SOMAVFSFilebuf.open` if the file does not exist.

### DIFF
--- a/apis/python/src/tiledbsoma/soma_vfs.cc
+++ b/apis/python/src/tiledbsoma/soma_vfs.cc
@@ -106,7 +106,12 @@ void load_soma_vfs(py::module& m) {
         .def(
             "open",
             [](SOMAVFSFilebuf& buf, const std::string& uri) {
-                return buf.open(uri, std::ios::in);
+                auto fb = buf.open(uri, std::ios::in);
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/soma_vfs.cc
```diff
@@ -106,7 +106,12 @@ void load_soma_vfs(py::module& m) {
         .def(
             "open",
             [](SOMAVFSFilebuf& buf, const std::string& uri) {
-                return buf.open(uri, std::ios::in);
+                auto fb = buf.open(uri, std::ios::in);
+                if (fb == nullptr) {
+                    TPY_ERROR_LOC(
+                        std::format("URI {} is not a valid file", uri));
```
johnkerl, line 112:
```suggestion
                        std::format("URI {} is not a valid URI", uri));
```

Issue #2879, opened 2024-08-12
[c++] Dictionary columns do not read back correctly when passed `timestamp`


nguyenv:
**Describe the bug**
When writing a dictionary column at a given `timestamp`, we cannot readback at the same `timestamp`. This issue is in libtiledbsoma and affects both tiledbsoma-py and tiledbsoma-r.

**Background**
When we initially create our `ArraySchema`, the `Enumeration` values are empty. At the first write, we read what the dictionary values are from the passed-in Arrow column and use `ArraySchemaEvolution` to add the values by extending the enumeration. Currently we always evolve the schema at the current timestamp as the C++ API does not contain a setter for the timestamp. This means reading an array with enumerations at a given timestamp errors out 'unexpectedly' (see code below).

**To Reproduce**
```
import tiledbsoma as soma
import pyarrow as pa
import pandas as pd
import tempfile

with tempfile.TemporaryDirectory() as uri:
    asch = pa.schema([("foo", pa.dictionary(pa.int8(), pa.large_string()))])
    soma.DataFrame.create(uri, schema=asch, tiledb_timestamp=1).close()

    pydict = {}
    pydict["soma_joinid"] = [0, 1, 2]
    pydict["foo"] = pd.Series(['a', 'b', 'a'], dtype='category')
    rb = pa.Table.from_pydict(pydict)

    with soma.DataFrame.open(uri, "w", tiledb_timestamp=2) as sdf:
        sdf.write(rb)

    pydict = {}
    pydict["soma_joinid"] = [3, 4]
    pydict["foo"] = pd.Series(['b', 'b'], dtype='category')
    rb = pa.Table.from_pydict(pydict)
                              
    with soma.DataFrame.open(uri, "w", tiledb_timestamp=3) as sdf:
        sdf.write(rb)

    with soma.DataFrame.open(uri) as sdf:
        table = sdf.read().concat()
        print(table)

    with soma.DataFrame.open(uri, tiledb_timestamp=1) as sdf:
        table = sdf.read().concat()
        print(table)
        
    with soma.DataFrame.open(uri, tiledb_timestamp=2) as sdf:
        table = sdf.read().concat()
        print(table)

    with soma.DataFrame.open(uri, tiledb_timestamp=3) as sdf:
        table = sdf.read().concat()
        print(table)
```

Notice how the values do not get readback correctly at the timestamps when they were written but readback looks fine at current timestamp.
```
pyarrow.Table
soma_joinid: int64
[...]

PR #3566, opened 2025-01-15
[python] Correct Arrow offsets for Enumeration and Boolean columns

nguyenv:
**Issue and/or context:**

[[sc-61239](https://app.shortcut.com/tiledb-inc/story/61239/c-tiledbsoma-soma-write-will-corrupt-data-written-to-array-if-array-is-a-slice-with-offset-0)]

**Changes:**

- Add additional unit tests to handle offset slicing for more datatypes
- Enumerations
  - Re-open the array at the end of the write call to get updated schema after performing schema evolution
  - Add in offset calculations where missing
- Booleans
  - Account for offset in `cast_bit_to_uint8`
  - Account for offset in validity buffer

### DIFF
--- a/libtiledbsoma/src/soma/soma_array.cc
+++ b/libtiledbsoma/src/soma/soma_array.cc
@@ -327,7 +327,8 @@ void SOMAArray::write(bool sort_coords) {
     }
     mq_->submit_write(sort_coords);
 
-    mq_->reset();
+    arr_ = std::make_shared<Array>(*ctx_->tiledb_ctx(), uri_, TILEDB_WRITE);
+    mq_ = std::make_unique<ManagedQuery>(arr_, ctx_->tiledb_ctx(), name_);
 }
 
 void SOMAArray::consolidate_and_vacuum(std::vector<std::string> modes) {
--- a/libtiledbsoma/src/soma/soma_array.cc
+++ b/libtiledbsoma/src/soma/soma_array.cc
@@ -327,6 +327,8 @@ void SOMAArray::write(bool sort_coords) {
     }
     mq_->submit_write(sort_coords);
 
+    // When we evolve the schema, the ArraySchema needs to be updated to the
+    // latest version so re-open the Array
     arr_ = std::make_shared<Array>(*ctx_->tiledb_ctx(), uri_, TILEDB_WRITE);
     mq_ = std::make_unique<ManagedQuery>(arr_, ctx_->tiledb_ctx(), name_);
 }
--- a/libtiledbsoma/src/utils/util.cc
+++ b/libtiledbsoma/src/utils/util.cc
@@ -81,18 +81,17 @@ std::vector<uint8_t> cast_bit_to_uint8(ArrowSchema* schema, ArrowArray* array) {
             schema->format));
     }
 
-    const void* data;
+    uint8_t* data;
     if (array->n_buffers == 3) {
-        data = array->buffers[2];
+        data = (uint8_t*)array->buffers[2] + array->offset;
     } else {
-        data = array->buffers[1];
+        data = (uint8_t*)array->buffers[1] + array->offset;
     }
 
     std::vector<uint8_t> casted;
     for (int64_t i = 0; i * 8 < array->length; ++i) {
-        uint8_t byte = ((uint8_t*)data)[i];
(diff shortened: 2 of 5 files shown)


libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -327,7 +327,10 @@ void SOMAArray::write(bool sort_coords) {
     }
     mq_->submit_write(sort_coords);
 
-    mq_->reset();
+    // When we evolve the schema, the ArraySchema needs to be updated to the
```
bkmartinjr, line 330: what happens if the user had opened the array for writing at a time less than the most recent write?  For example:
1. Create a ts=1, close
2. Open at ts=3, write, close
3. Open at ts=2, write, close

At step three, are things OK with the evolve?

libtiledbsoma/src/utils/util.cc
```diff
@@ -81,21 +81,21 @@ std::vector<uint8_t> cast_bit_to_uint8(ArrowSchema* schema, ArrowArray* array) {
...
+        data = (uint8_t*)array->buffers[2];
     } else {
-        data = array->buffers[1];
+        data = (uint8_t*)array->buffers[1];
     }
 
     std::vector<uint8_t> casted;
-    for (int64_t i = 0; i * 8 < array->length; ++i) {
-        uint8_t byte = ((uint8_t*)data)[i];
+    for (int64_t i = 0; i * 8 < array->offset + array->length; ++i) {
```
bkmartinjr, line 92: This approach is going to be inefficient for cases where you are taking a slice of a very long bool array.  It copies the slice twice (line 98 does a copy) and converts all bools less than array->offset (ie, wasted work and memory).

I suggest just unrolling the copy so it can be done once with minimal work and memory:
* do the initial head if not aligned with a byte boundary
* then the middle full-byte (sets of 8 bools)
* then the tail

that will be fast, avoid the copy, and do no extra work

Sketch (CAVEAT: I did not compile/test and left the tail bits unsketched - so treat this as likely buggy):
```
std::vector<uint8_t> casted(array->length);

// head
auto head_offset = array->offset % sizeof(uint8_t);
if (head_offset > 0) {
  auto head_length = std::min(sizeof(uint8_t)-head_offset, array->length);
  auto head = data[array->offset / sizeof(uint8_t)];
  for (int64_t j = head_offset; j < head_length; j++)
    casted.push_back(head >> j & 0x01);
}

// bulk
auto bulk_start = (array->offset + head_offset) / sizeof(uint8_t);
auto bulk_end = (array->offset + array->length) / sizeof(uint8_t);
for (int64_t i = bulk_start; i < bulk_end; i++) {
  for (j = 0; j < sizeof(uint8_t); j++)
    casted.push_back(data[i] >> j & 0x01)
}

// tail
...do the tail bits
```

libtiledbsoma/src/soma/soma_array.cc
nguyenv: No, this is a known issue related to enumerations at timestamps: https://github.com/single-cell-data/TileDB-SOMA/issues/2879

libtiledbsoma/src/utils/util.cc
```diff
@@ -81,20 +81,42 @@ std::vector<uint8_t> cast_bit_to_uint8(ArrowSchema* schema, ArrowArray* array) {
             schema->format));
     }
 
-    const void* data;
+    uint8_t* data;
     if (array->n_buffers == 3) {
-        data = array->buffers[2];
+        data = (uint8_t*)array->buffers[2];
```
bkmartinjr, line 86: this code doesn't seem to consider the validity buffer when converting the bit vector to the byte vector.  I'm not entirely clear on what are semantics should be, but based on Pythonic conventions, an invalid (nulled) position would commonly be cast to `false`.

I suspect (but wanted to confirm) it is all handled upstream (e.g., the validity buffer is written separately, and takes precedence over any value in `casted` data.

nguyenv: This is done upstream: https://github.com/single-cell-data/TileDB-SOMA/blob/viviannguyen/sc-61239/c-tiledbsoma-soma-write-will-corrupt-data/libtiledbsoma/src/soma/managed_query.cc#L1012-L1030

bkmartinjr (approved):

PR #3567, opened 2025-01-15
[c++] Handle add/drop/re-add of enumerated types

johnkerl:
## Scenario

Users adds an enumeration column, then drops it, then re-adds it.

Error message:

```
RuntimeError: [TileDB::REST] Error: Error in libcurl POST operation: libcurl error message 'CURLE_OK';
HTTP code 500; server response data '{"code":1006,"message":"error evolving schema for array try 
again","request_id":"8e33095a-4ea2-4c89-9335-420238c29f9f"}'.
```

or:

```
error evolving schema for array [...] ArraySchema: Error adding enumeration.
Enumeration with name '[...]' already exists in this ArraySchema
```

## Repro

```
#!/usr/bin/env python

import tiledbsoma
import tiledbsoma.io
import anndata as ad
import os, shutil

soma_uri = './repro-460'
h5ad_uri = '/var/s/a/pbmc3k_processed.h5ad'

if os.path.exists(soma_uri):
    shutil.rmtree(soma_uri)

# Add
tiledbsoma.io.from_h5ad(soma_uri, h5ad_uri, measurement_name = 'RNA')

# Drop
with tiledbsoma.Experiment.open(soma_uri, 'r') as exp:
    obs = exp.obs.read().concat().to_pandas()
obs.drop(columns=['louvain'], inplace=True)

with tiledbsoma.Experiment.open(soma_uri, 'w') as exp:
    tiledbsoma.io.update_obs(exp, obs)

# Re-add
adata = ad.read_h5ad(h5ad_uri)
with tiledbsoma.Experiment.open(soma_uri, 'w') as exp:
    tiledbsoma.io.update_obs(exp, adata.obs)
```

## Problem

* When we drop columns, we don't drop enumerations
* We don't _want_ to drop enumerations entirely:
  * Suppose the dataframe is created Monday with enum column `foo`
  * Suppose its column `foo` is dropped Tuesday
  * Suppose `foo` is added Wednesday
  * We don't want to remove the enumeration on Tuesday, since then people would be unable to do a time-travel read of the data when opened at the Monday timestamp

[[sc-61984]](https://app.shortcut.com/tiledb-inc/story/61984/error-evolving-schema-in-a-certain-context)

### DIFF
--- a/libtiledbsoma/src/soma/soma_dataframe.cc
+++ b/libtiledbsoma/src/soma/soma_dataframe.cc
@@ -220,7 +220,10 @@ void SOMADataFrame::update_dataframe_schema(
                 }
                 array_already_has_it = true;
 
-            } catch (const TileDBError& e) {
+            } catch (const std::exception& e) {
+                // Oddly, catch (tiledb::TileDBError& e) did not enter this
+                // block in unit test.
+                //
                 // Make sure the exception was for the reason we're considering.
                 // If it was for some other reason, fail.
                 const std::string msg = e.what();
--- a/libtiledbsoma/src/soma/soma_dataframe.cc
+++ b/libtiledbsoma/src/soma/soma_dataframe.cc
@@ -169,9 +169,10 @@ void SOMADataFrame::update_dataframe_schema(
         //   o ordered: bool
 
         auto enmr_it = add_enmrs.find(attr_name);
-        bool has_enmr = enmr_it != add_enmrs.end();
+        bool request_has_enmr = enmr_it != add_enmrs.end();
+        bool array_already_has_it = false;
 
-        if (has_enmr) {
+        if (request_has_enmr) {
             auto [enmr_type, ordered] = enmr_it->second;
             LOG_DEBUG(std::format(
                 "[SOMADataFrame::update_dataframe_schema] add col name {} "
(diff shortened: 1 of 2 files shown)


johnkerl:
Leaving in draft until I add a unit-test case

libtiledbsoma/src/soma/soma_dataframe.cc
```diff
@@ -180,17 +182,78 @@ void SOMADataFrame::update_dataframe_schema(
...
+            //
+            // It's quite possible the array had this name as an enumerated
+            // column, and then that column was dropped, and it's now being
+            // re-added.
+            //
+            // * We need to check if this is the case.
+            // * If it is, we need to check if the enumeration is compatible.
+            try {
+                const auto existing_enmr = ArrayExperimental::get_enumeration(
+                    tctx, array, attr_name);
```
nguyenv, line 203: Are we able to use this instead of opening a new `Array`?

```suggestion
                const auto existing_enmr = ArrayExperimental::get_enumeration(
                    tctx, *arr_, attr_name);
```

johnkerl: @nguyenv this is a static method

nguyenv (approved):

PR #3586, opened 2025-01-17
[python] Change `from_visium` to create `obs` presence matrix by default

jp-dark:
**Issue and/or context:** [sc-61587](https://app.shortcut.com/tiledb-inc/story/61587/enable-obs-presence-matrix-by-default-in-visium-ingestor)

**Changes:** 
Change the `from_visium` ingestor to create the `obs` presence matrix by default. A presence matrix is required for exporting to SpatialData from a `ExperimentAxisQuery`.


aaronwolen (approved):
Thanks!

PR #3611, opened 2025-01-22
[c++] Fix offsets for nullable columns

nguyenv:
**Issue and/or context:**

[[sc-62104](https://app.shortcut.com/tiledb-inc/story/62104/python-data-corruption-in-validity-layer-handling-when-arrow-array-has-offset-1)]

We were adding the offset to the `ArrowArray`'s validity buffer directly which is incorrect because the validity is stored as a bit, not as a `uint8_t`.

**Changes:**

- Add `util::bitmap_to_uint8` which takes a pointer to a bitmap, length, and offset, and returns a `vector<uint8_t>` representation; if the bitmap is a nullptr, then return a nullopt
- Add helper functions `ManagedQuery::_cast_bool_data`, formally `util::cast_bit_to_uint8`, and `ManagedQuery::_cast_validity_buffer`, which takes the validity buffer of the passed in `ArrowArray` and returns `vector<uint8_t>` with the properly shifted offset
- Update `ManagedQuery::setup_write_column` and `ColumnBuffer::set_data` to take in an `std::optional<std::vector<uint8_t>>` where the validity buffer has already been offset and casted to uint8
- Add pytest unit testing for nullable columns with offsetting for string, Boolean, enumeration, and the general case (int, float, etc.)

### DIFF
--- a/apis/python/src/tiledbsoma/_dense_nd_array.py
+++ b/apis/python/src/tiledbsoma/_dense_nd_array.py
@@ -290,9 +290,6 @@ class DenseNDArray(NDArray, somacore.DenseNDArray):
         """
         _util.check_type("values", values, (pa.Tensor,))
 
-        if np.isnan(values).any():
-            raise SOMAError("soma_data cannot contain nulls")
-
         clib_handle = self._handle._handle
 
         # Compute the coordinates for the dense array.
--- a/apis/python/src/tiledbsoma/_dense_nd_array.py
+++ b/apis/python/src/tiledbsoma/_dense_nd_array.py
@@ -290,6 +290,9 @@ class DenseNDArray(NDArray, somacore.DenseNDArray):
         """
         _util.check_type("values", values, (pa.Tensor,))
 
+        if np.isnan(values).any():
+            raise SOMAError("soma_data cannot contain nulls")
+
         clib_handle = self._handle._handle
 
         # Compute the coordinates for the dense array.
--- a/apis/python/tests/test_dataframe.py
+++ b/apis/python/tests/test_dataframe.py
@@ -2061,7 +2061,7 @@ def test_arrow_table_validity_with_slicing(tmp_path):
             ("myint", pa.int32()),
             ("mystring", pa.large_string()),
             ("mybool", pa.bool_()),
-            ("myenumstr", pa.dictionary(pa.int64(), pa.large_string())),
+            ("myenum", pa.dictionary(pa.int64(), pa.large_string())),
         ]
     )
 
--- a/apis/python/tests/test_dataframe.py
+++ b/apis/python/tests/test_dataframe.py
@@ -2092,9 +2092,9 @@ def test_arrow_table_validity_with_slicing(tmp_path):
     )
     table = pa.Table.from_pydict(pydict)
 
-    with raises_no_typeguard(soma.SOMAError):
-        # soma_joinid cannot be nullable
-        with soma.DataFrame.open(uri, "w") as A:
+    with soma.DataFrame.open(uri, "w") as A:
+        with raises_no_typeguard(soma.SOMAError):
+            # soma_joinid cannot be nullable
             A.write(table)
 
     pydict["soma_joinid"] = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
--- a/apis/python/tests/test_dataframe.py
+++ b/apis/python/tests/test_dataframe.py
@@ -2073,7 +2073,7 @@ def test_arrow_table_validity_with_slicing(tmp_path):
     pydict["myenum"] = pd.Categorical(
         ["g1", "g2", "g3", None, "g2", "g3", "g1", None, "g3", "g1"]
     )
-    
+
     table = pa.Table.from_pydict(pydict)
     domain = ((0, np.iinfo(np.int64).max - 2050),)
 
--- a/apis/python/tests/test_dataframe.py
+++ b/apis/python/tests/test_dataframe.py
@@ -2050,3 +2050,51 @@ def test_arrow_table_sliced_writer(tmp_path):
         np.testing.assert_array_equal(pdf["myenumint"], pydict["myenumint"])
         np.testing.assert_array_equal(pdf["myenumstr"], pydict["myenumstr"])
         np.testing.assert_array_equal(pdf["myenumbool"], pydict["myenumbool"])
+
+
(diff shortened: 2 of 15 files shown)


bkmartinjr (changes requested):
It would be good to add some additional tests:
1. Call Sparse/Dense NDArray `write` with an array that has a validity buffer, and ensure it acts as expected.
2. For DataFrame, need to verify expected behavior if an index (dimension) column is written with a validity buffer.
3. In test_dataframe, I'd also test the "invalid" timestamp types (NaT) alongside the other nullable types.

apis/python/tests/test_dataframe.py
```diff
@@ -2050,3 +2050,51 @@ def test_arrow_table_sliced_writer(tmp_path):
...
+        assert pdf["mybool"].compare(table["mybool"].to_pandas()).empty
+        assert pdf["myenum"].compare(table["myenum"].to_pandas()).empty
+
+    with soma.DataFrame.open(uri, "w") as A:
+        mid = num_rows // 2
+        A.write(table[:mid])
+        A.write(table[mid:])
+
+    with soma.DataFrame.open(uri) as A:
+        pdf = A.read().concat().to_pandas()
```
bkmartinjr, line 2096: the to/from Pandas code does a lot of work trying to infer validity based on values, so these tests are mixing up that behavior with the SOMA read/write of validity buffers.

Suggestion: do the equality comparisons on Arrow Tables directly, e.g., something like:

```
pdf_tbl = A.read().concat()
for col_name in pdf_tbl.column_names:
    assert pdf_tbl[col_name].equals(table[col_name])
```

bkmartinjr (changes requested):
testing requests only at this point - the C++ code looks OK (I did a quick pass only - would be good to get @XanthosXanthopoulos or @jp-dark to review it in more depth)

libtiledbsoma/src/soma/column_buffer.h
```diff
@@ -126,7 +126,7 @@ class ColumnBuffer {
         uint64_t num_elems,
         const void* data,
         T* offsets,
-        uint8_t* validity = nullptr) {
+        std::optional<std::vector<uint8_t>> validity = std::nullopt) {
```
jp-dark, line 129:
```suggestion
        const std::optional<std::vector<uint8_t>>& validity = std::nullopt) {
```

jp-dark, line 129:
```suggestion
        const std::optional<std::vector<uint8_t>>& validity = std::nullopt) {
```

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -1257,4 +1247,29 @@ bool ManagedQuery::_extend_and_evolve_schema<std::string>(
...
+    if (strcmp(schema->format, "b") != 0) {
+        throw TileDBSOMAError(std::format(
+            "_cast_bit_to_uint8 expected column format to be 'b' but saw "
+            "{}",
+            schema->format));
+    }
+
+    uint8_t* data;
+    if (array->n_buffers == 3) {
+        data = (uint8_t*)array->buffers[2];
```
XanthosXanthopoulos, line 1262: Is it valid for a bool array to have 3 buffers?

nguyenv: https://arrow.apache.org/docs/format/CDataInterface.html#binary-view-arrays

It looks like no. You can have a bool array `z` or binary view array `vz`, but those are different types, so I'll get rid of this check.

apis/python/src/tiledbsoma/_dense_nd_array.py
```diff
@@ -291,6 +291,9 @@ def write(
         """
         _util.check_type("values", values, (pa.Tensor,))
 
+        if np.isnan(values).any():
+            raise SOMAError("soma_data cannot contain nulls")
```
jp-dark, line 295: Do we want to be able to write nans for floating point arrays? Not as a validity buffers, but as a regular data value.

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -1257,4 +1247,23 @@ bool ManagedQuery::_extend_and_evolve_schema<std::string>(
...
+std::vector<uint8_t> ManagedQuery::_cast_bool_data(
+    ArrowSchema* schema, ArrowArray* array) {
+    if (strcmp(schema->format, "b") != 0) {
+        throw TileDBSOMAError(std::format(
+            "_cast_bit_to_uint8 expected column format to be 'b' but saw "
+            "{}",
+            schema->format));
+    }
+
+    uint8_t* data = (uint8_t*)array->buffers[1];
```
jp-dark, line 1260: Nit: Prefer named casts to C-style casting (for clarity + slightly less error prone). Probably `static_cast` here.

jp-dark (approved):
I'm not sure the behavior for dense arrays is correct, but otherwise it looks good to me.

nguyenv:
Updated now with all requested changes except:
> Call Dense NDArray write with an array that has a validity buffer

Because `DenseNDArray` takes in a `pa.Tensor` which does not support nulls.

apis/python/tests/test_sparse_nd_array.py
```diff
@@ -2007,3 +2007,38 @@ def test(path, tiledb_config):
...
+            [None, 0, None, 1, 2, None, None, 3, 4, 5], type=pa.float64()
+        ),
+    }
+    table = pa.Table.from_pydict(pydict)
+
+    soma.SparseNDArray.create(uri, type=pa.int64(), shape=(10,))
+
+    with soma.SparseNDArray.open(uri, "w") as A:
+        with raises_no_typeguard(soma.SOMAError):
+            # soma_joinid cannot be nullable
```
bkmartinjr, line 2027: As I understand, soma_data also can't be nullable - can we add a branch to test for both?

bkmartinjr (approved):
One small (optional) suggestion.  Otherwise LGTM!

apis/python/tests/test_sparse_nd_array.py
nguyenv: https://github.com/single-cell-data/TileDB-SOMA/pull/3611/files#diff-e95b695dd56e1f3dc55f80eb74bc3f4fd55aefbe85ae1ee005e42ee6ae5732bfR2041-R2044

Attributes with null values are casted to fill values.

PR #3623, opened 2025-01-24
[python] 

johnkerl:
**Issue and/or context:** Issue found with doing `tiledbsoma.io.update_obs` on cloud URIs, on a REST server for which local storage is configured to true. (Note that in public cloud that is configured to false.)

This is an old bug -- it was always wrong for REST servers that have local storage enabled, but I'd not tested against such, so this scenario all passed for us. I had done a bugfix on #2434, however, in hindsight, that depended on REST servers not having local storage configure. My bugfix on #2434 was, in hindsight, not particularly defensible.

Scenario: for `update_obs`:

* The `exp.uri` the customer has is well-formed like `tiledb://namespace/s3://path/to/exp`
* But `exp.obs` is tiledb://namespace/uuid
* That dataframe does exist, since they are doing `update_obs`
* In `tiledbsoma.io._write_dataframe_impl`, before this PR, we were doing the try-create-catch-open as described by a comment block on this PR
* This `exp.obs` URI with UUID was getting passed to `DataFrame.create`
* With the public-cloud REST server that's an error and it's caught as in `is_not_createable_error`; for a configuration with local storage enabled, it is not caught, and is fatal
all solved taking that extend_creation_uri logic and imitating it within update_dataframe

**Changes:**

Since the issue involved the semantics of doing the create part of try-create-catch-open in situations where the dataframe is required to already exist, it's safer to simply not attempt the try-create in such situations.

**Notes for Reviewer:**

The code here is validated for local-disk URIs in our unit-test suite. For cloud URIs, we rely on tests which are not currently encoded into this repository. Please see however which is the kind of material which will be grist for an upcoming cloud-URI test suite.

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -514,6 +514,7 @@ def from_anndata(
     # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
     # OBS
     df_uri = _util.uri_joinpath(experiment_uri, "obs")
+    # XXX good
     with _write_dataframe(
         df_uri,
         conversions.obs_or_var_to_tiledb_supported_array_type(anndata.obs),
@@ -565,6 +566,7 @@ def from_anndata(
             # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
             # MS/meas/VAR
             with _write_dataframe(
+                # XXX good
                 _util.uri_joinpath(measurement_uri, "var"),
                 conversions.obs_or_var_to_tiledb_supported_array_type(anndata.var),
                 id_column_name=var_id_name,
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -514,6 +514,7 @@ def from_anndata(
     # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
     # OBS
     df_uri = _util.uri_joinpath(experiment_uri, "obs")
+    # XXX good
```
nguyenv, line 517: Should these be removed?

nguyenv (approved):

apis/python/src/tiledbsoma/io/ingest.py
johnkerl: Gar thanks 😬

PR #3625, opened 2025-01-24
[python] `tiledbsoma.io.add_obs_layer` helper function

maarten-devries:
**Issue and/or context:**

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -1811,6 +1811,52 @@ def update_matrix(
         _util.format_elapsed(s, f"FINISH UPDATING {soma_ndarray.uri}"),
     )
 
+def add_obs_layer(
+    exp: Experiment,
+    obs_data: Union[pd.DataFrame, pa.Table],
+    ingest_mode: IngestMode = "write",
+    use_relative_uri: bool | None = None,
+) -> None:
+    """Add observation (cell) metadata to an existing TileDB-SOMA Experiment.
+    
+    This is useful for adding new cell-level metadata, for example from clustering results
+    or other analyses.
+
+    Use ``ingest_mode="resume"`` to not error out if the schema already exists.
+
+    Args:
+        exp: The TileDB-SOMA Experiment to add obs data to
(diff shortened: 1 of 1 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1811,6 +1811,52 @@ def update_matrix(
         _util.format_elapsed(s, f"FINISH UPDATING {soma_ndarray.uri}"),
     )
 
+def add_obs_layer(
```
johnkerl, line 1814: The `add_X_layer` function can be used to write a new component in the `X` collection -- e.g. if `X["raw"]` exists, you can create `X["norm"]` -- as `X` is a `Collection` of `SparseNDArray`.

Here, there's just one `obs` -- since `obs` is a `DataFrame`, not a `Collection` of `DataFrame`.

This is really a variant of `update_obs`.

What this does that `tiledbsoma.io.update_obs` doesn't:

* It accepts a `pyarrow.Table` (which is awesome!)

What `tiledbsoma.io.update_obs` does that this doesn't:

* Check old and new schema, and evolve the schema if necessary
  * That's fine here as long as it's clear, but it might benefit us to assert that the old and new schema match -- rather than deferring to `DataFrame.write` to raise an exception.
* Check that the old and new row-count are the same -- this allows row-count changes
  * If the row-count is increasing here, that's probably a good thing (although it needs a warning label that this may require a referential-integrity update on `X`, `obsm`, and `obsp` matrices -- caveat emptor)
  * If the row-count is decreasing here -- say from 100 rows to 98 -- then those last two rows from the previous write will still be visible. We'd need to write through the TileDB Core deletion API (which is solid and well-tested, simply not yet wired into tiledbsoma) to mark them as deleted. And in this case, too, a warning label about referential integrity.

johnkerl (commented):
This is very interesting -- thanks for sharing, @maarten-devries !!

Let me think about this a bit 🙏

jp-dark:
I'm closing this PR as stale. We are working on improvements to the IO that will hopefully make these types of operations easier.

PR #3630, opened 2025-01-24
[c++] Push down spatial metadata in the `PointCloudDataFrame` class from Python to C++

jp-dark:
**Issue and/or context:** [sc-62101] 

**Changes:**
* Add protected `SOMAArray::_create` method that passes back the raw TileDB-Array for additional metadata writes in subclasses.
* Write coordinate space and spatial-encoding metadata in C++ instead of Python for the `PointCloudDataFrame` class.


XanthosXanthopoulos (approved):
LGTM

PR #3640, opened 2025-01-29
[python] Re-export coordinate transforms from somacore

jp-dark:
**Issue and/or context:** [sc-59919](https://app.shortcut.com/tiledb-inc/story/59919/add-coordinate-space-transform-to-all)

**Changes:**
Re-export coordinate transforms from somacore.

### DIFF
--- a/apis/python/src/tiledbsoma/__init__.py
+++ b/apis/python/src/tiledbsoma/__init__.py
@@ -198,6 +198,7 @@ __version__ = get_implementation_version()
 
 __all__ = [
     "AlreadyExistsError",
+    "AffineTransform",
     "Axis",
     "AxisColumnNames",
     "AxisQuery",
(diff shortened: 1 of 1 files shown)


johnkerl (changes requested):
This has #3639 as its first commit ...

jp-dark:
> This has #3639 as its first commit ...

opps

johnkerl (approved):
Happy to approve but I don't understand -- for me on `main` without this PR:

```
>>> tiledbsoma.AffineTransform
<class 'somacore.coordinates.AffineTransform'>
```

What can we do with this PR in place that we could't do without it in place?

jp-dark:
> What can we do with this PR in place that we could't do without it in place?

I believe it only technically matters if you are using 
```
from tiledbsoma import *
```
but we re-export other somacore classes we expect users to access directly and so these should be included for consistency.

PR #3652, opened 2025-01-31
[r] Use S3 method dispatch on `integer64`

### LINKED ISSUE #3650: Call to abs.integer64
Here is a direct call to an S3 method:

https://github.com/single-cell-data/TileDB-SOMA/blob/a42d6b5b31a41f305b5ad7876b031b2a490f15b9/apis/r/R/CoordsStrider.R#L50

This will break in the future when `abs.integer64` is un-exported -- prefer calling `abs()` and letting S3 dispatch do its thing.

See https://github.com/r-lib/bit64/issues/76.

mojaveazure:
Replace calls directly to `integer64` methods with calls to generics and let S3 method dispatch handle the rest. Includes new calls to `bit64::integer()` to ensure that S3 method dispatch works correctly, and updates minimum version of bit64 to ensure proper using of S3 methods

Also removes unused imports and update docs

[SC-62802](https://app.shortcut.com/tiledb-inc/story/62802)

resolves #3650


johnkerl (approved):
🚢

PR #3663, opened 2025-02-04
#3560 review

ryan-williams:
**Issue and/or context:** stacked on #3560

I chased a few yellow/red squigglies in my editor, and also ran `mypy` on `apis/python/tests/ht` and skimmed the issues it flagged (just as an audit; `mypy` is normally disabled for `tests/` because it generates too many FPs).

**Changes:**

Added/Tweaked some type annotations, removed a few unused functions, etc.

### DIFF
--- a/apis/python/tests/ht/_array_state_machine.py
+++ b/apis/python/tests/ht/_array_state_machine.py
@@ -97,10 +97,6 @@ class SOMAArrayStateMachine(RuleBasedStateMachine):
         self.closed = True
         self.mode = None
 
-    @abstractmethod
-    def _reopen(self, mode: str) -> None:
-        pass
-
     ##
     ## ---- Open/close state
     ##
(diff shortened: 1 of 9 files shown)


apis/python/tests/ht/_array_state_machine.py
```diff
@@ -8,12 +8,13 @@
 import math
 import re
 from abc import abstractmethod
-from typing import Any, Literal, Protocol, Union
+from typing import Any, Literal, Protocol, Tuple, Union
```
bkmartinjr, line 11: I noticed many other files use `tuple`, but this one uses `Tuple`.  Probably should pick one.

My understanding is that `tuple[...]` works fine in 3.9 onward, so I'd prefer that.

bkmartinjr (approved):
Other than one little nit (Tuple vs tuple), this all looks great.

Can we just merge it into the original PR, and let me continue from there (debugging the recent failure)?

ryan-williams:
> Can we just merge it into the original PR, and let me continue from there (debugging the recent failure)?

Yea, if you merge this, it will update bkm/hypothesis and you can continue working there

PR #3668, opened 2025-02-04
[python] Fix enum regression in PR #3647

nguyenv:
**Issue and/or context:**

[[sc-62887](https://app.shortcut.com/tiledb-inc/story/62887/python-c-enum-regression-in-pr-3647)]

**Changes:**

When evolving the schema, we need to re-open the array so that on subsequent writes it sees the latest version. Previously this was done in `SOMAArray::write`. Now that we are replacing this with `ManagedQuery::write`, the reopening step needs to be moved into this method. There should be no performance degradation as are simply moving the same step from `SOMAArray::write` to `ManagedQuery::write`.

**Notes for Reviewer:**

My original thought was to combine chunks in the Python side to avoid the repeated reopening steps for multiple chunks in the first place, but it sounds like the overhead of creating the copy is worse than reopening.

### DIFF
--- a/apis/python/tests/test_dataframe.py
+++ b/apis/python/tests/test_dataframe.py
@@ -2124,3 +2124,44 @@ def test_arrow_table_validity_with_slicing(tmp_path):
         assert_array_equal(pdf["mybool"], table["mybool"])
         assert_array_equal(pdf["mydatetime"], table["mydatetime"])
         assert_array_equal(pdf["myenum"], table["myenum"])
+
+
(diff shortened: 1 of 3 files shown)


bkmartinjr:
sorry for piling on...I think combining the arrays (ie., not being zero copy) will definitely be noticed.  IIRC, we added the chunk/batch iteration originally to avoid this overhead when doing large writes, e.g., for the Census builder.

bkmartinjr (approved):

Issue #3545, opened 2025-01-10
[Bug] Cannot build wheels on linux x86


loguille:
Hello,
I'm encounting an issue regarding the build of tiledbsoma from source. Indeed I would like to install tiledbsoma on my server but I encounted the error : 
```sh
>>> import tiledbsoma
Illegal instruction (core dumped)
```

As mentionned in issue [2580.](https://github.com/single-cell-data/TileDB-SOMA/issues/2580) I'm then trying to build tiledbsoma from source. 
```sh
/home/guillel/TileDB-SOMA/libtiledbsoma/src/reindexer/reindexer.h:37:10: fatal error: format: No such file or directory
         37 | #include <format>
            |          ^~~~~~~~
      compilation terminated.
      gmake[5]: *** [src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/build.make:79: src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/reindexer/reindexer.cc.o] Error 1
      gmake[4]: *** [CMakeFiles/Makefile2:253: src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/all] Error 2
      gmake[4]: *** Waiting for unfinished jobs....
      [  4%] Building CXX object _deps/catch2-build/src/CMakeFiles/Catch2.dir/catch2/benchmark/detail/catch_benchmark_function.cpp.o
      [  4%] Building CXX object _deps/catch2-build/src/CMakeFiles/Catch2.dir/catch2/benchmark/detail/catch_run_for_at_least.cpp.o
      [  5%] Building CXX object _deps/catch2-build/src/CMakeFiles/Catch2.dir/catch2/benchmark/detail/catch_stats.cpp.o
      [  5%] Building CXX object _deps/catch2-build/src/CMakeFiles/Catch2.dir/catch2/generators/catch_generator_exception.cpp.o
      [  6%] Building CXX object _deps/catch2-build/src/CMakeFiles/Catch2.dir/catch2/generators/catch_generators.cpp.o
      [  6%] Built target TILEDBSOMA_NANOARROW_OBJECT
      [  7%] Building CXX object _deps/catch2-build/src/CMakeFiles/Catch2.dir/catch2/generators/catch_generators_random.cpp.o
      [  8%] Building CXX object src/CMakeFiles/TILEDB_SOMA_GEOMETRY_OBJECTS.dir/geometry/multipolygon.cc.o
[...]

PR #3673, opened 2025-02-05
[c++] Robustify detection of insufficient compiler version

johnkerl:
**Issue and/or context:** For #3154 / [[sc-57301]](https://app.shortcut.com/tiledb-inc/story/57301/use-c-20).

Issue is:

* On our `main` branch we now use C++20 (see #3154).
* GCC11 "supports" C++20 in the sense that `g++ -std=c++20` is recognized -- however, builds fail with failure to find `#include <format>`. That is, GCC 11's support of C++20 is inadequate.
  * Aside: Ubuntu 22.04 LTS has GCC 11; Ubutu 24.04 LTS has GCC 13.
* Multiple users have been tripped up by this cryptic error (e.g. #3545). It's important that we surface the problem, and its recommended resolution, more clearly and more actionably.

**Changes:**

* Run a compile-time assert
* It's a little tricky because `clang` (MacOS) does indeed have `__GNUC__` defined but has it set to 4.

Test program:

<details>

```
#include <iostream>
// g++ -E -dM gvers.cpp

int main() {

#ifdef GCC_VERSION
  std::cout << "GCC_VERSION " << GCC_VERSION << "\n";
#else
  std::cout << "GCC_VERSION is not defined\n";
#endif

#ifdef __GNUC__
  std::cout << "__GNUC__ " << __GNUC__ << "\n";
#else
  std::cout << "__GNUC__ is not defined\n";
#endif

#ifdef __GNUC_MINOR__
  std::cout << "__GNUC_MINOR__ " << __GNUC_MINOR__ << "\n";
#else
  std::cout << "__GNUC_MINOR__ is not defined\n";
#endif

#ifdef __GNUC_PATCHLEVEL__
  std::cout << "__GNUC_PATCHLEVEL__ " << __GNUC_PATCHLEVEL__ << "\n";
#else
  std::cout << "__GNUC_PATCHLEVEL__ is not defined\n";
#endif

  return 0;
}

// MacOS
// GCC_VERSION is not defined
// __GNUC__ 4
// __GNUC_MINOR__ 2
// __GNUC_PATCHLEVEL__ 1

// Ubuntu 22.04
// GCC_VERSION is not defined
// __GNUC__ 11
// __GNUC_MINOR__ 4
// __GNUC_PATCHLEVEL__ 0

// GCC_VERSION is not defined
// __GNUC__ 13
// __GNUC_MINOR__ 3
// __GNUC_PATCHLEVEL__ 0
```

</details>

**Notes for Reviewer:**

Note that `#error` doesn't stop the build entirely; we coud use `-Wfatal-errors` (and suitable replacements on Windows), or, use an intentionally bad `#include`.

### DIFF
--- a/libtiledbsoma/CMakeLists.txt
+++ b/libtiledbsoma/CMakeLists.txt
@@ -173,6 +173,19 @@ message(STATUS "Starting TileDB-SOMA regular build.")
 set(EP_SOURCE_DIR "${EP_BASE}/src")
 set(EP_INSTALL_PREFIX "${EP_BASE}/install")
 
+
+# ###########################################################
+# Check compiler version supports full C++20 standard.
+# ###########################################################
+
+set(GCC_MIN 13.0)
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS GCC_MIN)
+        message(FATAL_ERROR "GNU GCC must be at least version ${GCC_MIN}. Found version ${CMAKE_CXX_COMPILER_VERSION}.")
+    endif()
+endif()
+
+
 # ###########################################################
 # Compile options/definitions for all targets
 # ###########################################################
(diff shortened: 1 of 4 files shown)


jp-dark (commented):
Would it make more sense to check the compiler in CMake?

johnkerl:
> Would it make more sense to check the compiler in CMake?
Sure, that's a great idea! Can you help out?

johnkerl:
Closed in favor of #3676 -- thanks @jp-dark !

PR #3676, opened 2025-02-05
[c++] Add CMake check for GCC minimum version 13

jp-dark:
Add GCC minimum version check to CMake.

Context: #3154 / [[sc-57301]](https://app.shortcut.com/tiledb-inc/story/57301/use-c-20). See also https://github.com/single-cell-data/TileDB-SOMA/pull/3673#pullrequestreview-2595929065.

### DIFF
--- a/libtiledbsoma/CMakeLists.txt
+++ b/libtiledbsoma/CMakeLists.txt
@@ -181,7 +181,7 @@ set(EP_INSTALL_PREFIX "${EP_BASE}/install")
 set(GCC_MIN 13.0)
 if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS GCC_MIN)
-        message(FATAL_ERROR "GNU GCC must be at least version ${GCC_MIN}. Using version ${CMAKE_CXX_COMPILER_VERSION}.")
+        message(FATAL_ERROR "GNU GCC must be at least version ${GCC_MIN}. Found version ${CMAKE_CXX_COMPILER_VERSION}.")
     endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


johnkerl:
Thanks @jp-dark ! We don't have repo CI on `ubuntu-22.04` anymore (recall from #3154 that `ubuntu-22.04` has GCC 11 and `ubuntu-22.04` has GCC 13) but I'll check out this PR and run it on an Ubuntu 22.04 system, as a check.

libtiledbsoma/CMakeLists.txt
```diff
@@ -173,6 +173,19 @@ message(STATUS "Starting TileDB-SOMA regular build.")
...
 
+
+# ###########################################################
+# Check compiler version supports full C++20 standard.
+# ###########################################################
+
+set(GCC_MIN 13.0)
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS GCC_MIN)
+        message(FATAL_ERROR "GNU GCC must be at least version ${GCC_MIN}. Using version ${CMAKE_CXX_COMPILER_VERSION}.")
```
johnkerl, line 184:
```suggestion
        message(FATAL_ERROR "GNU GCC must be at least version ${GCC_MIN}. Found version ${CMAKE_CXX_COMPILER_VERSION}.")
```

"Using" sounds a little like "hey, we're switching over to GCC 13 instead!"

Which is obviously not the case, I grant, since this is a build error. Nonetheless I think "Found" is more clear.

johnkerl (approved):
Approved with one requested change (which I trust you to make).

Some experiments (yay!)

* Ubuntu 22.04 https://gist.github.com/johnkerl/252f7600ce6ad5bf8b5a66ad408dcd01
* Ubuntu 24.04 (built fine)

PR #3677, opened 2025-02-05
[c++] Add CMake check for GCC minimum version 13

jp-dark:
Add GCC minimum version check to CMake.

Context: #3154 / [[sc-57301]](https://app.shortcut.com/tiledb-inc/story/57301/use-c-20). See also https://github.com/single-cell-data/TileDB-SOMA/pull/3673#pullrequestreview-2595929065.


johnkerl (approved):
Thanks @jp-dark !! :)

PR #3678, opened 2025-02-05
[r] Replace `SOMAArray` read and write calls with `ManagedQuery`

nguyenv:
**Issue and/or context:**

Part of work for https://github.com/single-cell-data/TileDB-SOMA/issues/2054

**Changes:**

- Change C++ `ManagedQuery` constructor to take in `SOMAArray` instead of `std::unique_ptr<SOMAArray>`. Pybind11 by default holds a `unique_ptr` (`shared_ptr` can be used, but there are known issues which prevents us from using it) whereas Rcpp holds a `shared_ptr` - to easily support both, pass by value rather than smart pointer
- Refactor tiledbsoma-r to use `ManagedQuery` for read and write queries
- Rename `sr_{setup,next,etc}` to `mq_{setup,next,etc}` where `mq_setup` now returns a `ManagedQuery` (rather than `SOMAArray`) that can be iterated through with `sr_next` 
- Update `apply_dim_{points,ranges}` utility functions to take in a `ManagedQuery`

**Notes for Reviewer:**

Planned next steps:
1. Remove `read`, `write`, etc. from `SOMAArray` in C++ API.
2. Currently `ManagedQuery::submit_write` always calls `submit` and `finalize`. Rename `ManagedQuery::submit_read` to `ManagedQuery::submit` which both read and write path can use. Create a separate `ManagedQuery::finalize`.
3. Refactor the tiledbsoma-py write paths so that we only finalize once at the end, even with multiple submits.

### DIFF
--- a/apis/python/src/tiledbsoma/managed_query.cc
+++ b/apis/python/src/tiledbsoma/managed_query.cc
@@ -33,10 +33,7 @@ void load_managed_query(py::module& m) {
             py::init([](SOMAArray array,
                         std::shared_ptr<SOMAContext> ctx,
                         std::string_view name) {
-                return ManagedQuery(
-                    std::make_unique<SOMAArray>(array),
-                    ctx->tiledb_ctx(),
-                    name);
+                return ManagedQuery(array, ctx->tiledb_ctx(), name);
             }),
             py::arg("array"),
             py::arg("ctx"),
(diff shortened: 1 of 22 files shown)


mojaveazure (approved):
LGTM; can you bump the develop version (as of now, should bump to `1.16.99.3`) https://github.com/single-cell-data/TileDB-SOMA/blob/viviannguyen/r-use-managedquery/apis/r/DESCRIPTION#L9 and add an item to the changelog https://github.com/single-cell-data/TileDB-SOMA/blob/viviannguyen/r-use-managedquery/apis/r/NEWS.md?

Issue #3685, opened 2025-02-10
[python] Issues involving entirely null-filled categorical-of-string columns


johnkerl:
## Goal

To wend null-filled categorical-of-string in a Pandas dataframe (`adata.obs`) to a TileDB-SOMA experiment's `obs` dataframe

## Problems

* (Pandas) How to really see the dtype of a Pandas categorical series
* (Pandas) How to make a Pandas series which is a null-filled categorical of string
* (TileDB) How to see the index and value types of a TileDB-SOMA dataframe
* (Pandas/TileDB) How to correctly get an initial Pandas null-filled categorical-of-string into a TileDB-SOMA dataframe
* (TileDB) How to drop and re-add

## Demo script

This compares several different ways to create a categorical-of-string column in a Pandas dataframe. The last three are precisely indicative of the goal stated above. The first four are for compare-and-contrast.

```
#!/usr/bin/env python

import tiledbsoma
import pandas as pd
import numpy as np

pandas_df = pd.DataFrame(data={
    "all-non-null":     pd.Series(["a", "b", "c", "d"], dtype=pd.CategoricalDtype()),
    "one-nan":          pd.Series(["a", "b", np.nan, "d"], dtype=pd.CategoricalDtype()),
    "one-none":         pd.Series(["a", "b", None, "d"], dtype=pd.CategoricalDtype()),
    "all-nan":          pd.Series([np.nan] * 4, dtype=pd.CategoricalDtype()),
    "all-none":         pd.Series([None] * 4, dtype=pd.CategoricalDtype()),
    "workaround":       pd.Series(["X"] * 4, dtype=pd.CategoricalDtype()).cat.remove_categories(["X"]),
})

column_names = list(pandas_df.keys())

print("Pandas dataframe")
print(pandas_df)
print()

print("pandas_df.dtypes:")
print(pandas_df.dtypes)
print()

for column_name in column_names:
    print(f"pandas_df[{column_name}].dtype: { pandas_df[column_name].dtype}")
print()

for column_name in column_names:
    print(f"pandas_df[{column_name}].cat.categories: { pandas_df[column_name].cat.categories}")
print()

arrow_table = tiledbsoma._arrow_types.df_to_arrow(pandas_df.copy())

print("Arrow schema and table")
print(arrow_table)
print()
```

Output of the demo script:

Note in particular:

* When the input data contains some nulls, but not all, everything works.
* The challenge is how to get a categorical-of-string series which is _entirely_ null.

[...]

PR #3688, opened 2025-02-11
[python] Correctly create null-filled enum-of-string columns in `tiledbsoma.io`

johnkerl:
**Issue and/or context:** See #3685.

**Changes:** Jump through the Pandas and Arrow hoops needed to get an entirely null-filled enum-of-strings column.

**Notes for Reviewer:**


johnkerl:
Thanks @nguyenv !

PR #3690, opened 2025-02-12
[c++] Modify enum labels to `{name}_{dtype}`

nguyenv:
**Issue and/or context:**

[[sc-63086](https://app.shortcut.com/tiledb-inc/story/63086/multiple-tiledb-soma-issues-with-schema-evolution-and-categoricals)]

**Changes:**

Previously, we set enumeration labels to be the same as attribute names. However this causes an issue when dropping a column but then readding a column with the same name but different enumeration dtypes. To fix this case, create a new util function `get_enmr_label` to create enumeration labels as `[the name of the attribute][the dictionary value's format type]`.

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -61,7 +61,6 @@ from .. import (
     eta,
     logging,
 )
-from .. import pytiledbsoma as clib
 from .._collection import AnyTileDBCollection, CollectionBase
 from .._common_nd_array import NDArray
 from .._constants import SOMA_JOINID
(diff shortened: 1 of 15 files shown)


libtiledbsoma/src/soma/soma_attribute.cc
```diff
@@ -45,7 +45,7 @@ std::shared_ptr<SOMAColumn> SOMAAttribute::deserialize(
     std::optional<Enumeration>
         enumeration = enumeration_name ?
                           std::make_optional(ArrayExperimental::get_enumeration(
```
johnkerl, line 47: `ArrayExperimental::get_enumeration` returns `std::optional<std::string>`

libtiledbsoma/src/soma/soma_attribute.cc
```diff
@@ -45,7 +45,7 @@ std::shared_ptr<SOMAColumn> SOMAAttribute::deserialize(
...
-                              ctx, array, attribute.name())) :
+                              ctx, array, *enumeration_name)) :
```
johnkerl, line 48: Why do we need the `*` here?

johnkerl: I guess you're doing that in place of `.value()`?

nguyenv: Since it is `optional`, we need to get the value by using `*`.

johnkerl: ✅ , same as `.value()` -- thanks!

johnkerl: And that `std::optional` `enumeration_name` is truthy when it comes to its role in the ternary operator?

I think regardless it'd be more clear to put `enumeration_name.has_value()` before the `?` and `enumeration_name.value()` after ...

nguyenv: Yeah I can modify that to make reading the code more clear.

johnkerl: Thanks!

I'm sure these days the semantics of truthiness and `*` are the same with all compilers, but, I thought it wouldn't hurt to check. Especially since (a) I can't repro the fail on Mac or Ubuntu, and (b) I didn't see anything else weird-looking in the code, e.g. a C-style `char*` needing a `strdup` -- everything here looks like it's `std::string` so I'm not seeing any holes/opportunities for the segfault ... 🤔

libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -169,11 +169,11 @@ SOMAArray::SOMAArray(
     std::optional<TimestampRange> timestamp)
     : uri_(util::rstrip_uri(arr->uri()))
     , ctx_(ctx)
+    , arr_(arr)
```
johnkerl, line 172: Please add a comment here about how the initializer order is important, & controls the order that destructors run.

I know this is "general C++ knowledge" but this is (was!) a _particular_ pain-point and it deserves its own "hazard warning" sign.

johnkerl:
@nguyenv I ran `python -m pre_commit run -a -v` and pushed -- hoping you don't mind ...

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1708,45 +1707,45 @@ def _update_dataframe(
...
+        for add_key in add_keys:
+            # Don't directly use the new dataframe's dtypes. Go through the
+            # to-Arrow-schema logic, and back, as this recapitulates the original
+            # schema-creation logic.
+            atype = arrow_schema.field(add_key).type
+            if pa.types.is_dictionary(arrow_table.schema.field(add_key).type):
+                add_attrs[add_key] = get_arrow_str_format(atype.index_type)
+
+                enmr_format = get_arrow_str_format(atype.value_type)
+                enmr_label = f"{add_key}_{enmr_format}"
```
johnkerl, line 1741: Please put a comment here cross-referencing that if anyone ever wants to change the punctuation here they also need to do it in `get_enmr_label` in `util.cc`

libtiledbsoma/src/utils/util.cc
```diff
@@ -85,4 +85,11 @@ std::shared_ptr<SOMAColumn> find_column_by_name(
     return *column_it;
 }
 
+std::string get_enmr_label(
```
johnkerl, line 88: Please put a comment here cross-referencing that if anyone ever wants to change the punctuation here they also need to do it in `ingest.py`

libtiledbsoma/src/soma/soma_dataframe.cc
```diff
@@ -249,7 +248,14 @@ void SOMADataFrame::update_dataframe_schema(
...
     }
-    se.array_evolve(uri);
+
+    se.array_evolve(uri_);
+
+    // When we evolve the schema, the ArraySchema needs to be updated to the
+    // latest version so re-open the Array
+    auto mode = arr_->query_type();
+    arr_->close();
+    arr_->open(mode);
```
johnkerl, line 258: How does this work if the array was opened at a specific timestamp previously?

nguyenv: TileDB `Array::open` will open it at set timestamp.

johnkerl: Please add a comment to that effect :)

libtiledbsoma/src/utils/util.cc
nguyenv: I reverted the `add_enums` to what it was before so it doesn't take the enumeration name anymore but constructs the label on-the-fly in the method now.

johnkerl (approved):
🚀

PR #3694, opened 2025-02-13
[python] Move `df_to_arrow` to the io module

jp-dark:
**Changes:** 
Move `df_to_arrow` from `_arrow_types` to the `io.conversions` module. 

**Notes for Reviewer:**
This change is pure code movement. It is implemented now to make future changes easier to diff.


johnkerl (approved):
This is right, and long overdue, as `df_to_arrow` is expressly for `tiledbsoma.io`.

Thank you!

PR #3699, opened 2025-02-13
[python] Remove some dead code

johnkerl:
Orphaned code as found by @jp-dark 

I found that the `to_json` and `from_json` are (the only) things in `apis/python/src/tiledbsoma/io/_registration/ambient_label_mappings.py` which are not being unit-tested. I'll add those unit-test cases as a second commit.


jp-dark (approved):
Thanks!

Issue #3271, opened 2024-11-01
[python] Dense subarray-write full corner handling


johnkerl:
Follow-on work from #2407.

With core 2.27 we have current-domain support for dense arrays, and thereby we can make the core domain (soma maxdomain) larger.

PR #3269 fixed a case for core 2.27 (found at https://github.com/TileDB-Inc/centralized-tiledb-nightlies/issues/25) for this kind of write and readback:

```
[#####..................]
 ^^^^^                   data written
 ^^^^^^^^^^^^^^^^^^^^^^^ domain
```

That is a test case that has been existing, and is now passing with core 2.26 as well as core 2.27.

However, there are test cases _which were never written_, and which do not work (with core 2.26, 2.27, or otherwise), like this:

```
[.....#####.............]
      ^^^^^              data written
 ^^^^^^^^^^^^^^^^^^^^^^^ domain
```

```
[..................#####]
                   ^^^^^ data written
 ^^^^^^^^^^^^^^^^^^^^^^^ domain
```

Here's why. Say you have a dense 2-D array with shape (0,99) and you wrote data in (40,43) and you want to do is read it back.  Everything's fine in core and `libtiledbsoma` except this bit:
https://github.com/single-cell-data/TileDB-SOMA/blob/1.15.0rc3/apis/python/src/tiledbsoma/_dense_nd_array.py#L224-L236
This is because the code there (written by yours truly many months ago) only looks at the non-empty domain upper not the non-empty domain lower.  So core gives back an array of length 4 (as it should, with values at `40,41,42,43`) but here we mishandle it:
https://github.com/single-cell-data/TileDB-SOMA/blob/1.15.0rc3/apis/python/src/tiledbsoma/_dense_nd_array.py#L275

* The `arrow_table.column("soma_data")` is fine (length 4)
* The `.to_numpy()` on that is fine (length 4)
* It's only the `.reshape(target_shape)` that's wrong because `target_shape` is 44

See also #3272 for the R side which needs different work done.

johnkerl:
Closed by #3708

PR #3708, opened 2025-02-18
[c++,python] Correct slicing for `DenseNDArray`

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/3271

**Changes:**

Previously, dense arrays returned only the non-empty domain by default. With this update, dense arrays now return the full shape of the array.

### DIFF
--- a/apis/python/tests/test_dense_nd_array.py
+++ b/apis/python/tests/test_dense_nd_array.py
@@ -582,7 +582,7 @@ def test_read_result_order(tmp_path):
             assert np.array_equal(A.read(result_order="auto"), data)
 
 
-def test_slice_with_offset(tmp_path):
+def test_subset_slice(tmp_path):
     uri = tmp_path.as_posix()
 
     soma.DenseNDArray.create(uri, type=pa.int32(), shape=(10, 3))
@@ -596,10 +596,10 @@ def test_slice_with_offset(tmp_path):
 
     with soma.open(uri, mode="r") as A:
         actual = A.read((slice(0, 9), slice(0, 2)))
-        np.array_equal(expected, actual)
+        assert np.array_equal(expected, actual)
 
         actual = A.read()
-        np.array_equal(expected, actual)
+        assert np.array_equal(expected, actual)
 
 
 def test_slice_with_resize(tmp_path):
--- a/apis/python/tests/test_dense_nd_array.py
+++ b/apis/python/tests/test_dense_nd_array.py
@@ -212,19 +212,16 @@ def test_dense_nd_array_requires_shape(tmp_path, shape_is_numeric):
 
 def test_dense_nd_array_ned_write(tmp_path):
     uri = tmp_path.as_posix()
+    input = np.asarray([100, 101, 102, 103])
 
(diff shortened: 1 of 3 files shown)


apis/python/tests/test_dense_nd_array.py
```diff
@@ -583,3 +580,39 @@ def test_read_result_order(tmp_path):
             DeprecationWarning, match="The use of 'result_order=\"auto\"' is deprecated"
         ):
             assert np.array_equal(A.read(result_order="auto"), data)
+
+
+def test_slice_with_offset(tmp_path):
```
bkmartinjr, line 585: Is this test correct?  It doesn't perform any `assert` calls, so I don't think it actually tests anything?

bkmartinjr: Maybe you wanted to use `numpy.testing.assert_array_equal` instead of `numpy.array_equal`?

apis/python/tests/test_dense_nd_array.py
```diff
@@ -583,3 +580,39 @@ def test_read_result_order(tmp_path):
...
+    uri = tmp_path.as_posix()
```
bkmartinjr, line 586: tests should use a fixture for temp paths (otherwise these files don't get cleaned up reliably) - see sc-61088

apis/python/tests/test_dense_nd_array.py
johnkerl: > Is this test correct? It doesn't perform any assert calls, so I don't think it actually tests anything?

It is possible that the test is implicitly checking nothing throws. (If this is the case, the test case should be commented to indicate that that is what it is doing.)

apis/python/tests/test_dense_nd_array.py
bkmartinjr: sorry, this is a bogus comment from me - please ignore.  You are doing the right thing already!

bkmartinjr (approved):
Other than a tiny nit in tests (missing asserts), this LGTM!  Thank you.

apis/python/tests/test_dense_nd_array.py
nguyenv: > It is possible that the test is implicitly checking nothing throws

Yeah this is how I knew the bug was fixed because it was throwing before. But I also did mean to check the array returned correctly!

apis/python/tests/test_dense_nd_array.py
bkmartinjr: Yeah, this was completely _my_ brain segfault.  Your code is great!

johnkerl: @bkmartinjr is `tmp_path` not a fixture for temp paths?

bkmartinjr: It is.  See my two previous comments apologizing for my bogon!  Super sorry

johnkerl: https://docs.pytest.org/en/stable/how-to/tmp_path.html

PR #3711, opened 2025-02-19
[python] Name temp paths for clarity

johnkerl:
**Issue and/or context:** In support of [[sc-61088]](https://app.shortcut.com/tiledb-inc/story/61088).

As detailed there:

* Using the `pytest` fixture `tmp_path`:
  * Are removed only three test runs later, not at process exit -- this is a feature
  * Most of our test cases use these
  * Total usage is tens of thousands of files and directories not reaped until three runs later
  * this is fine
* `tempfile.TemporaryDirectory` files:
  * are deleted at process exit
  * whether or not the test failed
  * whether or not a context-manager syntax was used
  * this is fine
* `tempfile.mkdtemp` files:
  * are not deleted at process exit
  * We should be removing these

**Changes:**

For this PR, simply apply the prefixing feature which makes analysis easier. This is not the final PR on this matter.

**Notes for Reviewer:**


johnkerl:
@nguyenv ping 🙏

Issue #3720, opened 2025-02-21
TileDB-SOMA 1.16.0 [tracking]


johnkerl:
Following our [established procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases)

[[sc-63626]](https://app.shortcut.com/tiledb-inc/story/63626/tiledb-soma-1-16-0)

johnkerl:
Manual CI runs per our [established procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases): 

* https://github.com/single-cell-data/TileDB-SOMA/actions/runs/13466567191
* https://github.com/single-cell-data/TileDB-SOMA/actions/runs/13466569126

johnkerl:
Currently blocked pending analysis of a dataframe-read-performance regression. The regression isn't huge but bears investigation and resolution. Hoping for a release next week.

johnkerl:
Another snag: https://github.com/pypi/support/issues/5820

johnkerl:
> Currently blocked pending analysis of a dataframe-read-performance regression. The regression isn't huge but bears investigation and resolution. Hoping for a release next week.

Analysis complete. We will use core 2.27.2: https://github.com/single-cell-data/TileDB-SOMA/pull/3755

johnkerl:
Hard-blocked by
https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/284

PR #3722, opened 2025-02-21
[c++] Check arrays are open for write at write time

johnkerl:
Found during dev on a forthcoming PR. I believe we lost this check in decoupling `ManagedQuery` from `SOMArray`. And we missed the fact that we lost the check due to missing unit-test coverage. This PR fixes both.

#3720 [[sc-63626]](https://app.shortcut.com/tiledb-inc/story/63626/tiledb-soma-1-16-0)


mojaveazure (approved):
LGTM on the R side

nguyenv (approved):

PR #3723, opened 2025-02-21
[python/r] Remove `used_shape`, which was deprecated in 1.15 with scheduled removal in 1.16

johnkerl:
As noted at https://github.com/single-cell-data/TileDB-SOMA/releases/tag/1.15.0

See also #2407 and  [[sc-51048]](https://app.shortcut.com/tiledb-inc/story/51048).


mojaveazure (approved):
LGTM on the R side, thanks! Please bump the develop version, update the changelog, and 🚢

apis/python/src/tiledbsoma/io/outgest.py
```diff
@@ -487,12 +487,7 @@ def _extract_obsm_or_varm(
     if num_cols is None:
         if soma_nd_array.tiledbsoma_has_upgraded_shape:
             num_cols = soma_nd_array.shape[1]
-        else:
-            try:
-                used_shape = soma_nd_array.used_shape()
-                num_cols = used_shape[1][1] + 1
-            except SOMAError:
-                pass  # We tried; moving on to next option
+        # Else: we tried; moving on to next option
```
jp-dark, line 490: Optional: Combine the if statements with an "and", and either remove the comment or move it above the if statement and say what exactly "we tried".

jp-dark (approved):
The python side looks good to me.

apis/python/src/tiledbsoma/io/outgest.py
johnkerl: Thanks! :)

PR #3724, opened 2025-02-21
Dark/encoding spec/sc 56279

jp-dark:
**Issue and/or context:** [sc-56279](https://app.shortcut.com/tiledb-inc/story/56279/audit-soma-metadata-history)

**Changes:**

Create a document with the current and past encoding specification.

### DIFF
--- a/encoding_specification.md
+++ b/encoding_specification.md
@@ -2,39 +2,53 @@
 title: Encoding Specification
 ---
 
-## SOMA Encoding Version 1.1.0
+## TileDB-SOMA encoding version 1.1.0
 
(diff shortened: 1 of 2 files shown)


encoding_specification.md
teo-tsirpanis, line 1: Should this be in https://github.com/single-cell-data/SOMA/blob/main/abstract_specification.md or a new file in that repository instead?

jp-dark: No, this is specifically about how TileDB encodes SOMA data on disk. The other repo is for the API a generic specification needs to implement.

PR #3727, opened 2025-02-21
[python]  Add Visium v1 dataset

jp-dark:
**Issue and/or context:** [sc-63318](https://app.shortcut.com/tiledb-inc/story/63318/add-visium-version-1-dataset)

**Changes:**
Add Visium v1 dataset.


johnkerl (changes requested):
Changes requested

scripts/prepare-test-data.sh
```diff
@@ -37,3 +37,18 @@ else
     cd ..
     echo "   ... finished preparing dataset 'data/example-visium-v2'."
 fi
+
+
+# Download and extract Visium v2 dataset.
+if [ -d example-visium-v1 ]; then
```
johnkerl, line 43: we're repeating `example-visium-v1` quite a bit

on #3730 i put
```
name="soma-experiment-versions"
```
and re-used `$name` below that

just a thought/suggestion here, but i think it's a good suggestion :)

johnkerl (approved):
LGTM, with a suggestion

Issue #3704, opened 2025-02-18
[python/ci] Support Python 3.13


johnkerl:
From https://github.com/conda-forge/tiledbsoma-feedstock/pull/6#issuecomment-2663491050

[[sc-63457]](https://app.shortcut.com/tiledb-inc/story/63457/try-out-python-3-13-with-tiledb-soma)

johnkerl:
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/13393627032/job/37407086285

```
ImportError while loading conftest '/home/runner/work/TileDB-SOMA/TileDB-SOMA/apis/python/tests/conftest.py'.
apis/python/tests/conftest.py:9: in <module>
    import tiledbsoma.io
/opt/hostedtoolcache/Python/3.[13](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/13393627032/job/37407086285#step:13:14).2/x64/lib/python3.13/site-packages/typeguard/_importhook.py:98: in exec_module
    super().exec_module(module)
apis/python/src/tiledbsoma/io/__init__.py:5: in <module>
    from ._registration import (
/opt/hostedtoolcache/Python/3.13.2/x64/lib/python3.13/site-packages/typeguard/_importhook.py:98: in exec_module
    super().exec_module(module)
apis/python/src/tiledbsoma/io/_registration/__init__.py:106: in <module>
    from .ambient_label_mappings import (
/opt/hostedtoolcache/Python/3.13.2/x64/lib/python3.13/site-packages/typeguard/_importhook.py:98: in exec_module
    super().exec_module(module)
apis/python/src/tiledbsoma/io/_registration/ambient_label_mappings.py:[17](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/13393627032/job/37407086285#step:13:18): in <module>
    from tiledbsoma.io._util import read_h5ad  # Allow us to read over S3 in backed mode
/opt/hostedtoolcache/Python/3.13.2/x64/lib/python3.13/site-packages/typeguard/_importhook.py:98: in exec_module
    super().exec_module(module)
apis/python/src/tiledbsoma/io/_util.py:65: in <module>
    class _FSPathWrapper(pathlib.Path):
apis/python/src/tiledbsoma/io/_util.py:81: in _FSPathWrapper
    _flavour = pathlib.Path().__class__._flavour  # type: ignore[attr-defined]
E   AttributeError: type object 'PosixPath' has no attribute '_flavour'
[...]

PR #3703, opened 2025-02-18
[python/ci] Try Python 3.13 in CI

johnkerl:
**Issue and/or context:** https://github.com/conda-forge/tiledbsoma-feedstock/pull/6, #3704, [[sc-63457]](https://app.shortcut.com/tiledb-inc/story/63457/try-out-python-3-13-with-tiledb-soma)

**Changes:**

**Notes for Reviewer:**

PR #3728, opened 2025-02-23
Fix python 3.13 compatibility

ap--:
**Issue and/or context:**

Closes single-cell-data/TileDB-SOMA#3704
Replaces single-cell-data/TileDB-SOMA#3703
XRef conda-forge/tiledbsoma-feedstock#6

**Changes:**

Fixes the `tiledbsome.io._util._FSPathWrapper` implementation to be compatible with changes in Python `3.13` pathlib.

**Notes for Reviewer:**

Since `3.12` it's a lot easier to subclass `pathlib.Path`, which is why the version check in the `_FSPathWrapper` class is targeting `3.12` now. Your previous implementation worked on `3.12` but used the current working directory as path for all pathlib operations. (It's not used in the 'hack' within anndata, but just so you're aware).

I believe your hack will continue to work with Python `3.14`, if anndata does not start to use more pathlib API features for the provided filename, or refactors the `AnnDataFileManager.filename` property.
In general, it would be a much better solution to contribute support upstream to anndata to allow reading directly from file-like objects. I'll see if I can find some time to do so.

_Also let me know if more changes to the workflow configurations are needed. I just copied from #3703_

Cheers,
Andreas 😃


johnkerl:
My apologies @ap-- -- I did not add myself as reviewer and misplaced this. I just rebased & am preparing to merge now.

Thank you!!!

johnkerl (approved):

PR #3732, opened 2025-02-25
[r] [ci] Migrate R CI to r-lib/actions

mojaveazure:
Use r-lib/actions instead of `tools/r-ci.sh` for R CI; also makes use of RSPM/P3M + pak for binaries and system requirements

[SC-63716](https://app.shortcut.com/tiledb-inc/story/63716)


johnkerl (approved):
Thank you!

PR #3747, opened 2025-02-27
[python] Add `register_visium_dataset` function with `NotImplementedError`

jp-dark:
* Add placeholder function `register_visium_dataset` that throws a `NotImplementedError`.
* Clean-up parameters to the `from_visium` function. Some of the options that are used in append mode don't apply to `from_visium`.
* Clean-up `from_visium` docstring.

### DIFF
--- a/apis/python/src/tiledbsoma/io/spatial/__init__.py
+++ b/apis/python/src/tiledbsoma/io/spatial/__init__.py
@@ -8,7 +8,7 @@ This module is for experimental features. Support for these features
 may be dropped.
 """
 
-from .ingest import VisiumPaths, from_visium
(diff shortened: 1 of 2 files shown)


ktsitsi:
Do we want to expose `register_from_visium` function at the API level? From a UX perspective I think it would be beneficial to a user for us to encapsulate the registration process in the `from_visium` API by resolving internally the destination path.

jp-dark:
> Do we want to expose `register_visium_dataset` function at the API level?

This copies the pattern of the `from_anndata` and `from_h5ad`. It's not needed for ingesting a single dataset.

ktsitsi (approved):

PR #3767, opened 2025-03-11
[r] Replace tiledb-r with SOMA for test helper

mojaveazure:
Replace tiledb-r with SOMA for test helper function to create an empty sparse nd-array

[SC-64301](https://app.shortcut.com/tiledb-inc/story/64301)

continues work for #2406


johnkerl (approved):
🚢

PR #3768, opened 2025-03-12
Run `pip check` in CI

jdblischak:
**Issue and/or context:** Many of the spatial Python packages that SOMA depends on require Python >= 3.10. `pip` still installs them into our Python 3.9 CI env, but reports messages like the [following](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/13766978880/job/38495810318#step:9:210):

```
  Link requires a different Python (3.9.21 not in: '<3.13,>=3.10'): https://files.pythonhosted.org/packages/a5/b5/ea49151c7f846a2f083271c9c0a92c046a0d5b522369e839089d8022e870/spatialdata-0.2.6-py3-none-any.whl (from https://pypi.org/simple/spatialdata/) (requires-python:<3.13,>=3.10)
```

**Changes:** Runs `pip check` to summarize the version incompatibilities

**Notes for Reviewer:**

### DIFF
--- a/.github/workflows/python-ci-single.yml
+++ b/.github/workflows/python-ci-single.yml
@@ -118,6 +118,9 @@ jobs:
     - name: Show XCode version
       run: clang --version
 
+    - name: Run pip check
+      run: pip check
+
     - name: Run libtiledbsoma unit tests
       run: ctest --output-on-failure --test-dir build/libtiledbsoma -C Release --verbose
       env:


jdblischak:
Well that was reassuring. The `pip check` passed for the [py39 build](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/13818040289/job/38656254153?pr=3768#step:12:12).

The messages I had seen from `pip` about `Link requires a different Python` was simply added by the `-v` flag to inform why certain versions were skipped. Using the example I copy-pasted above, spatialdata 0.2.6 was not installed because it requires py>=310, and ultimately spatialdata 0.2.5.post0 was installed.

So the build works here in CI because the older [`fsspec-2023.6.0`](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/13818040289/job/38656254153?pr=3768#step:9:2735) was installed, which is incompatible cloud env requirement of `fsspec>=2024.10.0,<2025`

PR #3772, opened 2025-03-13
[python] Fix `to_anndata` with no `X`

johnkerl:
#3773 

[[sc-63400]](https://app.shortcut.com/tiledb-inc/story/63400/python-tiledbsoma-io-to-anndata-raises-error-when-x-layer-missing-but-not-requested)

### DIFF
--- a/apis/python/src/tiledbsoma/io/outgest.py
+++ b/apis/python/src/tiledbsoma/io/outgest.py
@@ -225,11 +225,18 @@ def _read_dataframe(
 
 
 # ----------------------------------------------------------------
+class Sentinel:
+    pass
+
+
+MISSING = Sentinel()
+
+
 def to_anndata(
     experiment: Experiment,
     measurement_name: str,
     *,
-    X_layer_name: str | None = "data",
+    X_layer_name: str | Sentinel | None = MISSING,
     extra_X_layer_names: Sequence[str] | KeysView[str] | None = None,
     obs_id_name: str | None = None,
     var_id_name: str | None = None,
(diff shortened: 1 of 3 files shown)


apis/python/src/tiledbsoma/io/outgest.py
```diff
@@ -323,9 +323,26 @@ def to_anndata(
             "If X_layer_name is None, extra_X_layer_names must not be provided"
         )
 
+    # Problem: when the specified layer name does not exist.
```
bkmartinjr, line 326: there is a pythonic way to handle this case, and I recommend we use it.

1. You create a module-level sentinal value called `MISSING` (or whatever you like)
2. The default `X_layer` value is this sentinal

For example:
```
class Sentinal:
   pass

MISSING = Sentinal()


def a_function(
    X_layer: str|None|Sentinal = MISSING
):
    if X_layer is MISSING:
        print("no value specified")
    elif X_layer is None:
        print("None was specified")
    else:
        print(f"{X_layer} was specified.")
```

You could also just skip the class declaration and assign `MISSING=object()`.  I did the class simply for clarity

bkmartinjr: To be more explicit, this would let you equate `MISSING` with the `data` layer, and `None` with "do not read X layer" behavior.

johnkerl: Thanks for the code-review feedback, @bkmartinjr !

johnkerl: I'll use the pattern, which TIL (thank you!) but not the misspelling "Sentinal [sic]"

johnkerl: TIL!

apis/python/tests/test_basic_anndata_io.py
```diff
@@ -856,8 +856,27 @@ def test_X_none(h5ad_file_X_none):
         assert exp.ms["RNA"].var.count == 1838
         assert list(exp.ms["RNA"].X.keys()) == []
 
-        tiledbsoma.io.to_anndata(exp, measurement_name="RNA", X_layer_name=None)
-        # TODO: more
+        adata = tiledbsoma.io.to_anndata(exp, measurement_name="RNA", X_layer_name=None)
```
bkmartinjr, line 859: Do we need a test case to differentiate between not specifying an X_layer_name  and  specifying None as X_layer_name, when X.keys() contains `data`?

bkmartinjr (changes requested):
Couple of suggestions about how we could be more explicit about the defaulting.

apis/python/src/tiledbsoma/io/outgest.py
```diff
@@ -225,11 +225,18 @@ def _read_dataframe(
 
 
 # ----------------------------------------------------------------
+class Sentinel:
```
bkmartinjr, line 228: Nit: this (and MISSING) are generally useful, so might consider putting them into a common utils file?

bkmartinjr (approved):

apis/python/src/tiledbsoma/io/outgest.py
johnkerl: Thanks again @bkmartinjr !

PR #3784, opened 2025-03-14
[python/c++] Expose enumeration-values getter

johnkerl:
**Issue and/or context:**

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -390,7 +390,7 @@ class DataFrame(SOMAArray, somacore.DataFrame):
         if not isinstance(field.type, pa.DictionaryType):
             raise ValueError(f"column name '{column_name}' is not of dictionary type")
 
-        return pa.array(["red", "yellow", "green"])
+        return self.handle.column_enumeration_values(column_name)
 
     @property
     def domain(self) -> Tuple[Tuple[Any, Any], ...]:
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -382,6 +382,9 @@ class DataFrame(SOMAArray, somacore.DataFrame):
         schema, or if the specified column is not of Arrow dictionary type."""
         self._check_open_read()
 
+        # These assertions could be done in C++. However, it's easier here
+        # to do the exception-type multiplexing, raising ValueError for one
+        # thing, TileDBSOMAError for another.
         if column_name not in self.schema.names:
             raise ValueError(
                 f"column name '{column_name}' is not present in the schema"
@@ -390,7 +393,7 @@ class DataFrame(SOMAArray, somacore.DataFrame):
         if not isinstance(field.type, pa.DictionaryType):
             raise ValueError(f"column name '{column_name}' is not of dictionary type")
 
-        return self.handle.column_enumeration_values(column_name)
+        return self._handle.column_enumeration_values(column_name)
 
     @property
     def domain(self) -> Tuple[Tuple[Any, Any], ...]:
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -376,6 +376,22 @@ class DataFrame(SOMAArray, somacore.DataFrame):
         """
         return self._tiledb_dim_names()
 
+    def column_enumeration_values(self, column_name: str) -> pa.Array:
+        """Returns an Arrow array of the column's enumeration/dictionary/categorical
+        values. Raises ``ValueError`` if the specified column name is not in the
+        schema, or if the specified column is not of Arrow dictionary type."""
+        self._check_open_read()
+
+        if column_name not in self.schema.names:
+            raise ValueError(
+                f"column name '{column_name}' is not present in the schema"
+            )
+        field = self.schema.field(column_name)
+        if not isinstance(field.type, pa.DictionaryType):
+            raise ValueError(f"column name '{column_name}' is not of dictionary type")
+
+        return pa.array(["red", "yellow", "green"])
+
     @property
     def domain(self) -> Tuple[Tuple[Any, Any], ...]:
         """Returns tuples of minimum and maximum values, one tuple per index column, currently storable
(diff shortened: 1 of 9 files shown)


bkmartinjr:
To close the loop on Pandas - I don't think we need that, as it would duplicate functionality already in Arrow.

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -376,6 +376,31 @@ def index_column_names(self) -> Tuple[str, ...]:
...
         return self._tiledb_dim_names()
 
+    def get_enumeration_values(
+        self: DataFrame, column_names: Sequence[str]
+    ) -> dict[str, pa.Array]:
+        """Returns an Arrow array of the specified columns'
+        enumeration/dictionary/categorical values. Raises ``ValueError`` if any
+        of the the specified column names is not in the schema, or if any is not
+        of Arrow dictionary type."""
+        self._check_open_read()
```
bkmartinjr, line 386: why is this only available for `read`?  it is useful in both modes, the the `tiledb-py` API allows it to be fetched in both cases.

It is essentially schema information, and should be available in all the same modes that schema is available (i.e., read and write)

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -376,6 +376,31 @@ def index_column_names(self) -> Tuple[str, ...]:
...
+
+        # These assertions could be done in C++. However, it's easier here
+        # to do the exception-type multiplexing, raising ValueError for one
+        # thing, TileDBSOMAError for another.
+        for column_name in column_names:
+            if column_name not in self.schema.names:
```
bkmartinjr, line 392: This entire check could be omitted, as it is duplicative of the call to `self.schema.field` on line 396.

Also, it should raise a KeyError not ValueError per python error conventions

bkmartinjr (changes requested):
Primary request: function needs to work when the array is open for read _or_ write, just like any other schema accessor.

One other nit which will simplify the code but is optional

johnkerl:
@bkmartinjr thank you! Ready for round two.

apis/python/tests/test_dataframe.py
```diff
@@ -255,6 +255,127 @@ def test_dataframe_with_enumeration(tmp_path):
...
+    uri = tmp_path.as_posix()
+
+    schema = pa.schema(
+        [
+            pa.field("not_an_enum", pa.large_string()),
+            pa.field("string_ordered", pa.dictionary(pa.int32(), pa.large_string())),
+            pa.field("string_unordered", pa.dictionary(pa.int32(), pa.large_string())),
+            pa.field("int64_ordered", pa.dictionary(pa.int32(), pa.int64())),
+            pa.field("int64_unordered", pa.dictionary(pa.int32(), pa.int64())),
+        ]
```
bkmartinjr, line 269: Suggest also adding a few more weird but in-use enum value types:
> pa.dictionary(pa.uint32(), pa.float32())  # unsigned code, float value
> pa.dictionary(pa.int8(), pa.bool_())  # boolean enum

apis/python/tests/test_dataframe.py
```diff
@@ -255,6 +255,127 @@ def test_dataframe_with_enumeration(tmp_path):
         assert_array_equal(df["myfloat"].chunk(0).dictionary, enums["enmr2"])
 
 
+@pytest.mark.parametrize("mode", ["r", "w"])
```
bkmartinjr, line 258: Missing test:  ensuring timestamps/time travel and enum changes work correctly, i.e., that you can open at a previous time and see the correct values returned by get_enum_values

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -549,6 +554,11 @@ def count(self) -> int:
     def write(self, values: pa.RecordBatch) -> None:
         self._handle.write(values)
 
+    def get_enumeration_values(
+        self, column_names: Sequence[str]
+    ) -> dict[str, pa.Array]:
+        return cast(dict[str, Any], self._handle.get_enumeration_values(column_names))
```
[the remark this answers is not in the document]
johnkerl: It is always necessary at this layer to avoid lint issues. I'll add a comment to that effect.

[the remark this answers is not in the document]
johnkerl: ... sorry, I misread your comment. Straight-up typo on my part. Fix coming in next commit.

libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -359,6 +359,119 @@ std::optional<TimestampRange> SOMAArray::timestamp() {
...
+SOMAArray::get_enumeration_values_for_column(std::string column_name) {
+    // This will throw if column name is not found
+    std::unique_ptr<ArrowSchema> up_column_schema = arrow_schema_for_column(
+        column_name);
+    ArrowSchema* column_schema = up_column_schema.get();
+
+    // Throw if this column is not of dictionary type
+    if (column_schema->dictionary == nullptr) {
+        throw TileDBSOMAError(fmt::format(
+            "[get_get_enumeration_values] column named '{}' is not of "
```
nguyenv, line 404:
```suggestion
            "[get_enumeration_values_for_column] column named '{}' is not of "
```

libtiledbsoma/src/soma/soma_array.h
```diff
@@ -428,6 +441,29 @@ class SOMAArray : public SOMAObject {
      */
     std::optional<TimestampRange> timestamp();
 
+    /**
+     * Retrieves the enumeration values from the array's TileDB schema,
+     * for specified column names. Throws if any of the column names
+     * is not present in the schema, or if any of them is present but
+     * is for a non-enumerated column.
```
nguyenv, line 448:
```suggestion
     * for specified column names. Throws if any of the column names
     * are not present in the schema, or if any of them are present but
     * none are for a non-enumerated column.
```

nguyenv (approved):
lgtm overall

apis/python/tests/test_dataframe.py
johnkerl: Pushed

Issue #3787, opened 2025-03-17
[python] Revert from `std::format` to `fmt::format`


johnkerl:
On https://github.com/single-cell-data/TileDB-SOMA/issues/3154 we moved to C++ 20.

We found in r-universe 
https://github.com/r-universe/tiledb-inc/actions/runs/13768506335/job/38502055087
that we got a build failure for x86 Mac only.

@jeroen generously wrote up https://github.com/single-cell-data/TileDB-SOMA/pull/3778 which shows:

> The [xcode cpp feature table](https://developer.apple.com/xcode/cpp/) shows that `std::format` requires `xcode 15.3`, which is [only available for MacOS-14 and up](https://developer.apple.com/support/xcode/) because the libc++ runtime version that ships with macos-13 is too old. However, macos-13 is [the last version of macos to run on x86 (Intel) architecture](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) on GitHub actions.  

Therefore we need to:

* Keep C++ 20 _except_ use `fmt::format` in place of `std::format`
* I believe we can now revert the GCC 13 requirement as well https://github.com/single-cell-data/TileDB-SOMA/pull/3676
* Follow-up for #3763 

@XanthosXanthopoulos please prepare separate PRs to:

* Use `fmt::format`
* Revert the GCC requirement
* Follow up for #3763 

Please make these three separate PRs.

johnkerl:
Open item: https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/296#issuecomment-2749307091

cc @XanthosXanthopoulos

johnkerl:
@XanthosXanthopoulos is this still open?

XanthosXanthopoulos:
No this is completed

PR #3790, opened 2025-03-17
[c++] Roll back to use `fmt::format` instead of `std::format`

XanthosXanthopoulos:
**Issue and/or context:** #3787 [[sc-64546]](https://app.shortcut.com/tiledb-inc/story/64546/revert-from-std-format-to-fmt-format)

**Changes:**
- Replaces `std::format` with `fmt::format` 
- Updates spdlog to version 1.15.1 to match conda installed version
- Updates Rcppspdlog to 0.0.20 
- Replaces `spdl::*` with `spdlog::*` in R api C++ files. Reason behind this change is that in C++20 fmt::format requires a compile time string and the implementation in `spdl.h` passes a `const char*` doesn't qualify as such.

**Notes for Reviewer:**


johnkerl:
@XanthosXanthopoulos re
https://github.com/single-cell-data/TileDB-SOMA/actions/runs/13908218298/job/38916098550
from #3793 
we also need to update `std::make_format` callsites

johnkerl (approved):
Thank you!!

Issue #3785, opened 2025-03-14
Need API to extend column enum values without doing a write


johnkerl:
There are cases where it is necessary to evolve the schema prior to doing writes, e.g., ingesting into a `DataFrame` with multiple independent writers. This is similar to other schema evolution steps (e.g., set shape).

Request: an API to extend the enum values for a SOMA DataFrame column.

For bulk ingestion, this is a fairly high priority request.

Methods for `DataFrame`:

* `def column_enumeration_values(column_name: str) -> pa.Array:`
* `def extend_column_enumeration_values(values: pa.Array, dedupe: bool) -> None:`

[[sc-63930]](https://app.shortcut.com/tiledb-inc/story/63930/dataframe-need-api-to-extend-column-enum-values-without-doing-a-write#activity-64471)

PR #3812, opened 2025-03-20
[c++] Pass `Enumeration` through `_extend_enumeration`

johnkerl:
**Issue and/or context:** Split out from #3784 for #3785 / [[sc-63930]](https://app.shortcut.com/tiledb-inc/story/63930/dataframe-need-api-to-extend-column-enum-values-without-doing-a-write#activity-64471)

**Changes:**

* On #3784 it will be necessary to have an otherwise-obtained enumeration passed. There's no strict need for `_extend_enumeration` to be instantiating the core `tiledb::Enumeration` object
* As a bonus, we avoid a double instantiate:
  * https://github.com/single-cell-data/TileDB-SOMA/blob/86f08d524efff3d20a9f1779f922252aa07ac3ec/libtiledbsoma/src/soma/managed_query.cc#L953-L954
  * https://github.com/single-cell-data/TileDB-SOMA/blob/86f08d524efff3d20a9f1779f922252aa07ac3ec/libtiledbsoma/src/soma/managed_query.cc#L1038-L1039
* Along the way, I now pass the core `SchemaEvolution` object by reference as there is no need to be copying it.

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/soma/managed_query.cc
+++ b/libtiledbsoma/src/soma/managed_query.cc
@@ -942,16 +942,10 @@ bool ManagedQuery::_extend_enumeration(
     ArrowArray* value_array,
     ArrowSchema* index_schema,
     ArrowArray* index_array,
-    ArraySchemaEvolution se) {
+    Enumeration& enmr,
+    ArraySchemaEvolution& se) {
     // For columns with dictionaries, we need to identify the data type of the
     // enumeration to extend any new enumeration values
-
-    // As of 1.16, and enumeration names are in the format of
-    // {index name}_{value dtype}. Prior to 1.16, enum labels are the same as
-    // the index name. If the new format doesn't work then fall back to the old
-    // format
-    Enumeration enmr = util::get_enumeration(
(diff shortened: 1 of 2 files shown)


libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -156,7 +156,7 @@ void ArrowAdapter::release_array(struct ArrowArray* array) {
         // Dictionary arrays are allocated differently than data arrays
         // We need to free the buffers one at a time, then we can call the
         // release schema to continue the cleanup properly
-        for (size_t i = 0; i < array->dictionary->n_buffers; ++i) {
+        for (int64_t i = 0; i < array->dictionary->n_buffers; ++i) {
```
johnkerl, line 159: This was from a compiler warning I found. Note that `n_children` is of type `int64_t` for both `ArrowSchema` and `ArrowArray` within `nanoarrow.h`.

nguyenv (approved):

johnkerl:
Thanks @nguyenv !

PR #3815, opened 2025-03-20
[c++/python] Expose enumeration-extender

johnkerl:
**Issue and/or context:** Step two for #3785 / [[sc-63930]](https://app.shortcut.com/tiledb-inc/story/63930/dataframe-need-api-to-extend-column-enum-values-without-doing-a-write#activity-64471). This is stacked atop #3784.

**Changes:**

**Notes for Reviewer:**


libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -1090,6 +1090,87 @@ bool ManagedQuery::_extend_and_evolve_schema_and_write(
...
+            return _extend_and_evolve_schema<double>(
+                value_schema, value_array, column_name, enmr, se);
+        default:
+            throw TileDBSOMAError(fmt::format(
+                "ArrowAdapter: Unsupported TileDB dict datatype: {} ",
+                tiledb::impl::type_to_str(value_type)));
+    }
+}
+
+bool ManagedQuery::_extend_and_evolve_schema_string(
```
johnkerl, line 1149: I tried `template <>` but the compiler told me it couldn't resolve the instantiation. I fiddled with it for 20 minutes, and failed. It wasn't worth fighting the compiler over. So, I took out the `template <>` and put a `_string` at the end of the method's name.

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -401,6 +401,30 @@ def get_enumeration_values(
...
+        enumeration/dictionary/categorical values. Raises ``ValueError`` if any
+        of the the specified column names is not in the schema, or if any is not
+        of Arrow dictionary type."""
+        self._check_open_write()
+
+        # These assertions could be done in C++. However, it's easier here
+        # to do the exception-type multiplexing, raising ValueError for one
+        # thing, TileDBSOMAError for another.
+        for column_name in values.keys():
+            if column_name not in self.schema.names:
```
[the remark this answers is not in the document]
johnkerl: Thanks @bkmartinjr 

This PR is not yet ready for review. Please don't review it.

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -401,6 +401,30 @@ def get_enumeration_values(
 
         return self._handle.get_enumeration_values(column_names)
 
+    def extend_enumeration_values(
+        self: DataFrame, values: dict[str, pa.Array], dupes_ok: bool = False
```
bkmartinjr, line 405: suggest that the second param be renamed to `deduplicate`, which is the action it controls.  The current name suggests that it will "allow" the final enum to contain duplicate values, which isn't ever possible.

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -401,6 +401,30 @@ def get_enumeration_values(
...
+
+        # These assertions could be done in C++. However, it's easier here
+        # to do the exception-type multiplexing, raising ValueError for one
+        # thing, TileDBSOMAError for another.
+        for column_name in values.keys():
+            # As with get_enumeration_values: we are trusting pyarrow to raise
+            # KeyError, and raise it with a sufficiently clear error message,
+            # when the column name is not present within the schema.
+            field = self.schema.field(column_name)
+            if not isinstance(field.type, pa.DictionaryType):
```
bkmartinjr, line 423: I believe the suggested way to do this with Arrow is:
> if not pa.types.is_dictionary(field.type):

apis/python/src/tiledbsoma/_soma_object.py
```diff
@@ -333,5 +333,9 @@ def _check_open_read(self) -> None:
         if self.mode != "r":
             raise ValueError(f"{self} is open for writing, not reading")
 
+    def _check_open_write(self) -> None:
+        if self.mode != "w":
+            raise ValueError(f"{self} is open for reading, not writing")
```
bkmartinjr, line 338: it also could be closed, right?  So I suggest the error message be:   `f"{self} - must be open for write."` (or something like that)

apis/python/src/tiledbsoma/soma_dataframe.cc
```diff
@@ -156,6 +156,55 @@ void load_soma_dataframe(py::module& m) {
...
+                        arrow_schema->children[i] = &pa_schemas[i];
+                        arrow_schema->children[i]->name = strdup(
+                            column_name.c_str());
+
+                        i++;
+                    }
+
+                    ArrowTable t(
+                        std::move(arrow_array), std::move(arrow_schema));
+                    sdf.extend_enumeration_values(t, dupes_ok);
```
bkmartinjr, line 198: this can be a very slow operation due to all of the I/O. Can we release the GIL around it?

bkmartinjr: ps. same suggestion around the getter as well

bkmartinjr (changes requested):
Several suggestions:
1. release GIL around the get/set enum calls, as these are slow I/O operations
2. minor naming/wording suggestions

apis/python/src/tiledbsoma/_soma_object.py
johnkerl: @bkmartinjr you are correct

I copy-pasted from `_check_open_read` which shares this defect.

On thisPR, I'll fix them both.

Or, better, I'll put up a separate PR fixing `_check_open_read`, and rebase this on top of that.

johnkerl: > Or, better, I'll put up a separate PR fixing _check_open_read, and rebase this on top of that.

@bkmartinjr that PR is #3847. I've not yet rebased this PR on top of that one, but I will do so today.

I'll put this PR back in draft status until it is once again ready for review (at present, it is not ready for re-review).

johnkerl:
Returning to draft status until such time as this PR is ready for re-review (it is currently not ready for re-review).

apis/python/src/tiledbsoma/_dataframe.py
johnkerl: pushed

apis/python/src/tiledbsoma/soma_dataframe.cc
johnkerl: pushed

johnkerl:
End-of-week status: all that remains is for me to merge in in #3845, which is the single reason unit tests are failing in CI for this PR as of this moment.

johnkerl:
End-of-week status: #3845 is merged in, _mutatis mutandis_. @bkmartinjr this PR is ready for review.

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -401,6 +401,37 @@ def get_enumeration_values(
...
+            field = self.schema.field(column_name)
+            if not pa.types.is_dictionary(field.type):
+                raise KeyError(
+                    f"schema column name '{column_name}' is not of dictionary type"
+                )
+            if pa.types.is_dictionary(values[column_name].type):
+                raise ValueError(
+                    f"value column name '{column_name}' is of dictionary type: pass its dictionary array instead"
+                )
+                # TO DISCUSS IN CODE REVIEW: alternatively:
```
bkmartinjr, line 430: IMHO, you have it right.  The `values` argument should be a simple array of values.  Constraints on that array of values:
* Same primitive type as the dictionary value type, ie,. `self.schema[column_name].type.value_type == values.type`
* no dups unless `deduplicate` is True
As I read the code, both of these are satisfied....

johnkerl: Yes, they are satisfied by the code that isn't commented out.

I didn't think about this as I coded it, but: if someone did pass us a categorical: we could penalize them by throwing, requiring that the do a `.categories` on their input, or, we could take the `.categories` _for them_ and continue. I can really go either way. The non-commented-out code is find; just food for thought.

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -1218,23 +1373,29 @@ ManagedQuery::_extend_and_evolve_schema_with_details(
...
     ArrowArray* value_array,
     std::string column_name,
+    bool deduplicate,
     Enumeration enmr,
     ArraySchemaEvolution& se) {
-    // We need to check if we are writing any new enumeration values. If so,
-    // extend and evolve the schema. If not, just set the write buffers to the
-    // dictionary's indexes as-is
+    if (value_array->n_buffers != 2) {
+        throw std::invalid_argument(fmt::format(
```
bkmartinjr, line 1380: this error is going to confuse most users (I think).  Are you really just prohibiting dictionary arrays?

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -1218,23 +1373,29 @@ ManagedQuery::_extend_and_evolve_schema_with_details(
     ArrowSchema* value_schema,
     ArrowArray* value_array,
     std::string column_name,
+    bool deduplicate,
     Enumeration enmr,
     ArraySchemaEvolution& se) {
-    // We need to check if we are writing any new enumeration values. If so,
-    // extend and evolve the schema. If not, just set the write buffers to the
-    // dictionary's indexes as-is
+    if (value_array->n_buffers != 2) {
```
bkmartinjr, line 1379: this also raises the question of how you handle the nullable buffer - what happens if the user sends a value array with positions nulled?  Is that allowed?

I think the simplest thing is disallow it and perform the test at the very top-level (_extend_enumeration)

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -1218,23 +1373,29 @@ ManagedQuery::_extend_and_evolve_schema_with_details(
...
+        throw std::invalid_argument(fmt::format(
+            "[ManagedQuery] _extend_and_evolve_schema_with_details non-string: "
+            "expected n_buffers == 2; got {}",
+            value_array->n_buffers));
+    }
 
     // Get all the enumeration values in the passed-in column
     std::vector<ValueType> enum_values_in_write;
     uint64_t num_elems = value_array->length;
     if (strcmp(value_schema->format, "b") == 0) {
```
bkmartinjr, line 1389: why is the bool-specific code in the general purpose template? 

related to this: the caller treats bools as if they are the same as uint8, and then catches the problem here.  Seems like a very confusing way to handle the generic method, which is otherwise parameterized by arrow array type.

Recommend that you either have a specialization for the Arrow bool_ type, _or_ perform this explicit cast to uint8 before calling this method

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -1244,22 +1405,78 @@ ManagedQuery::_extend_and_evolve_schema_with_details(
         enum_values_in_write.assign(
             (ValueType*)data, (ValueType*)data + num_elems);
     }
+    std::unordered_set<ValueType> unique_values_in_write;
+    for (auto enum_value_in_write : enum_values_in_write) {
```
bkmartinjr, line 1409: C++ style suggests that you should use the undered_set constructor which accepts an input iterator, rather than writing a manual insertion loop

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -1244,22 +1405,78 @@ ManagedQuery::_extend_and_evolve_schema_with_details(
...
+    //
+    // and one would be correct. However, core uses bitwise comparisons, and
+    // floating-point NaNs have the following properties: (1) NaN != NaN, and
+    // (2) there are multiple floating-point bit patterns which are NaN.  It's
+    // simplest to just use the same logic core does, making std::string_view on
+    // our elements.
+    std::unordered_set<std::string_view> existing_enums_set;
+    for (const auto& enum_value_existing : enum_values_existing) {
+        auto sv = std::string_view(
+            static_cast<char*>((char*)&enum_value_existing),
```
bkmartinjr, line 1441: Why are all ValueTypes being cast to std::string?  (lines 1438-1455)

I understand that you want a "bit value compare" to avoid issues like NaN equality tests, but it seems like casting to a `void*` (a la memcmp) provides clearer intent

apis/python/tests/test_dataframe.py
```diff
@@ -461,6 +468,229 @@ def test_bool_enums(tmp_path, data_and_expected_levels):
         assert actual == expect
 
 
+@pytest.mark.parametrize("extend_not_write", [False, True])
+def test_extend_enumeration_values(tmp_path, extend_not_write):
+    """Compares the older create+write path against create+extend path"""
+    uri = tmp_path.as_posix()
+    domain = [[0, 2]]
+
+    schema = pa.schema(
```
bkmartinjr, line 477: Tests are missing cases of extending an _ordered_ dict.  While the current code base doesn't branch based upon this, we should still test it.

bkmartinjr: In particular, in the dedup case it is super important that the dedup preserve ordering....

apis/python/tests/test_dataframe.py
```diff
@@ -461,6 +468,229 @@ def test_bool_enums(tmp_path, data_and_expected_levels):
...
+    }
+
+    # Do the extend without write, or write with implicit extend
+    with soma.DataFrame.create(
+        uri,
+        schema=schema,
+        domain=domain,
+        index_column_names=["soma_joinid"],
+    ) as sdf:
+        if extend_not_write:
```
bkmartinjr, line 537: this is missing another common case where:
* the dataframe has been written to (e.g., enum values for the column are set to something)
* then the enum is extended

In other words, you current test extension on an empty dataframe, but not on one that already has data

libtiledbsoma/src/soma/managed_query.cc
johnkerl: I hoped my comments were sufficiently clear -- this is explicitly designed to imitate what core is doing.

See also
https://github.com/TileDB-Inc/TileDB/blob/2.27.2/tiledb/sm/array_schema/enumeration.cc#L417-L456

libtiledbsoma/src/soma/managed_query.cc
johnkerl: The check is important. I found this the hard way when, in development, I was passing a Python categorical array rather than the values array.

The check needs to remain here, at this level. This is a fallback, ensuring that before we deference `buffers[1]` or `buffers[2]` that they actually exist. I will more clearly word this to indicate that fact. The context I didn't make sufficiently clear is that it is an internal coding error if this check ever fails.

libtiledbsoma/src/soma/managed_query.cc
johnkerl: For my education, please provide a reference to whichever "C++ style" you're citing. Thank you!

apis/python/tests/test_dataframe.py
johnkerl: @bkmartinjr did you miss seeing

* https://github.com/single-cell-data/TileDB-SOMA/blob/b3965b35265540b61d4bbade6fd247ebc0eb7dc6/apis/python/tests/test_dataframe.py#L472
* https://github.com/single-cell-data/TileDB-SOMA/blob/b3965b35265540b61d4bbade6fd247ebc0eb7dc6/apis/python/tests/test_dataframe.py#L586

I intentionally created more than one unit-test functio; testing an extension on a dataframe that already has data is in the second one

johnkerl: I'll add a comment in the first one that mentions there is another test case after it

apis/python/tests/test_dataframe.py
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/3815/commits/6469953614136b2745cc963afe1c15c7f0991e9c

libtiledbsoma/src/soma/managed_query.cc
johnkerl: Also: https://github.com/single-cell-data/TileDB-SOMA/pull/3815/commits/f215424fcbcbfdfa7c3c50c171a8f5106e43e5d0

libtiledbsoma/src/soma/managed_query.cc
johnkerl: (I haven't found a reference by googling, so I'd love to learn from whatever your preferred style reference is.)

bkmartinjr:
FYI: I have successfully tested the changes in this PR with #3865.

libtiledbsoma/src/soma/managed_query.cc
bkmartinjr: it follows from two principles:  write less code, and RAII

apis/python/tests/test_dataframe.py
bkmartinjr: With the latest changes, I still don't see a test for the sequence I mentioned above:
1. create
2. write (with implicit extend)
3. extend

The second test case (`test-extend_enumeration_values_deduplication`) does not perform any `write()` operations.  And the first test case does a `write()`, but not followed by an `extend`.

I realize this is a very niche edge case -- I'm just trying to make sure that when we use both `write()` and `extend_enumeration()` in combination, the work as expected

libtiledbsoma/src/soma/managed_query.cc
bkmartinjr: Can we ask for some help from a C++ guru?  I don't personally feel these kinds of work-arounds are good for our code base maintainability

apis/python/tests/test_dataframe.py
```diff
@@ -461,6 +468,253 @@ def test_bool_enums(tmp_path, data_and_expected_levels):
         assert actual == expect
 
 
+@pytest.mark.parametrize("extend_not_write", [False, True])
+@pytest.mark.parametrize("ordered", [True, False])
+def test_extend_enumeration_values(tmp_path, extend_not_write, ordered):
```
bkmartinjr, line 473: another missing test case:  extending the schema with an array that contains nulled positions

bkmartinjr: And another:   send `extend_enumeration` a value array that is a slice of another array.   This will catch mistakes where you fail to account for the Arrow array offset.

Example:

> df.extend_enumeration({ 'colname':  pa.array(['a','b','c'])[1:-1])

libtiledbsoma/src/soma/managed_query.cc
johnkerl: cc @ihnorton for triage

libtiledbsoma/src/soma/managed_query.cc
johnkerl: pushed

apis/python/tests/test_dataframe.py
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/3815/commits/979e4c1209121fbc026a2f10ae22645e9d37d65c

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -1113,10 +1209,24 @@ ManagedQuery::_extend_and_evolve_schema_with_details<std::string>(
...
+            "expected n_buffers == 3; got {}",
+            value_array->n_buffers));
+    }
+
+    // Set up input values as a char buffer, and offsets within it.  This is
+    // zero-copy on the data buffers, since the Arrow data buffers are already
+    // contiguous.
+    //
+    // Arrow var-sized cells can have 32-bit or 64-bit offsets.  TileDB only has
+    // 64-bit offsets. Convert from the former to the latter.
```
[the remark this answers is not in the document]
johnkerl: Linking to previous discussion #3860 for reference

apis/python/tests/test_dataframe.py
johnkerl: > another missing test case: extending the schema with an array that contains nulled positions

https://github.com/single-cell-data/TileDB-SOMA/pull/3815/commits/1b24a55eef0e9d5ff7b4696eb7ff8bedc6edfcde

libtiledbsoma/src/soma/managed_query.cc
johnkerl: pushed

libtiledbsoma/src/soma/managed_query.cc
johnkerl: pushed

libtiledbsoma/src/soma/managed_query.cc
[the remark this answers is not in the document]
johnkerl: pushed, and #3860 tracks the refactor

apis/python/tests/test_dataframe.py
johnkerl: pushed

libtiledbsoma/src/soma/managed_query.cc
johnkerl: Crowd-sourcing internally

libtiledbsoma/src/soma/managed_query.cc
johnkerl: I stand with the current factoring. It's just an if-block, and I find extending the profusion of helper methods more bulky than clarifying.

libtiledbsoma/src/soma/managed_query.cc
nguyenv: Did you also define `_extend_and_evolve_schema<std::string>` in the header? (In addition to the generalized  `_extend_and_evolve_schema<>`?)

johnkerl: @nguyenv https://github.com/single-cell-data/TileDB-SOMA/pull/3888

johnkerl: @nguyenv https://github.com/single-cell-data/TileDB-SOMA/pull/3888#discussion_r2023641950

apis/python/src/tiledbsoma/_dataframe.py
bkmartinjr, line 430: once we have discussed sufficiently in code review, would be good to remove this comment

apis/python/tests/test_dataframe.py
```diff
@@ -651,7 +665,31 @@ def test_extend_enumeration_values_deduplication(tmp_path, deduplicate, ordered)
...
+                        ["the", "quick", "brown", "fox"], type=pa.large_string()
+                    ),
+                    "string_enum": pa.DictionaryArray.from_arrays(
+                        [0, 1, 0, 1], values["string_enum"]
+                    ),
+                    "int64_enum": pa.DictionaryArray.from_arrays(
+                        [0, 2, 3, 1], values["int64_enum"]
+                    ),
+                    "float32_enum": pa.DictionaryArray.from_arrays(
+                        [0, 0, 0, 1], values["float32_enum"]
```
bkmartinjr, line 685: This needs to also verify that NaN's are handled correctly, as that is tricky (especially given that the different Nan variants should not de-dup).

I suggest something like [NaN, -NaN , a signalling NaN, etc].  There is code to generate most of these in `test_enum_handling_category_of_nan_62449`

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -1231,11 +1237,13 @@ ManagedQuery::_extend_and_evolve_schema_with_details<std::string>(
     if ((strcmp(value_schema->format, "U") == 0) ||
         (strcmp(value_schema->format, "Z") == 0)) {
         uint64_t* offsets = (uint64_t*)value_array->buffers[1];
-        offsets_v.assign(offsets, offsets + num_elems + 1);
+        offsets_v.assign(
+            offsets + value_array->offset,
+            offsets + value_array->offset + num_elems + 1);
```
bkmartinjr, line 1242: Apologies if I am missing something obvious -- why is there a `+ 1` at the end of the second argument?  The `assign` uses normal iterator ranges [first, last), so I believe it should just be `offsets + value_array->offset + num_elems`

apis/python/tests/test_dataframe.py
johnkerl: @bkmartinjr 
https://github.com/single-cell-data/TileDB-SOMA/blob/kerl/enum-values-2/apis/python/tests/test_dataframe.py#L735-L753

libtiledbsoma/src/soma/managed_query.cc
johnkerl: @bkmartinjr that predates this PR. I'll look into it. Thanks for flagging.

apis/python/src/tiledbsoma/_dataframe.py
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/3815/commits/fc578bd9e804b433f7087cd9928bdb01941cd091

apis/python/tests/test_dataframe.py
johnkerl: (i.e. please scroll down)

bkmartinjr: where does it check that the expected extension occurred?   I.e., read back and assert that the values are as expected?

(really sorry if I'm being blind!)

libtiledbsoma/src/soma/managed_query.cc
johnkerl: @bkmartinjr if the inputs are `["hello", "goodbye"]` then the first offset is 0, the second is 5, and the third 12. This makes it possible to locate the end of the string `"goodbye"` within the char buffer `"hellogoodbye"`.

johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/3815/commits/301a3b503cb8c3010ee86118c3fb362b3292c674

apis/python/tests/test_dataframe.py
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/3815/commits/8cf48c59e279564effa105aff3a0315328c00ae3

libtiledbsoma/src/soma/managed_query.cc
johnkerl: More generally, it takes n+1 offsets to locate the starts and ends of n values in a string column. Please also see https://github.com/single-cell-data/TileDB-SOMA/pull/3815/commits/1ee11a25c0c961e39f7d14cc904934f899e12599.

bkmartinjr (approved):
The bulk of my concerns are resolved. I would like to see one of our C++ experts review as well, but this is coming together nicely.  Thanks for the extra tests!

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -1244,22 +1430,83 @@ ManagedQuery::_extend_and_evolve_schema_with_details(
...
+    // (2) there are multiple floating-point bit patterns which are NaN.  It's
+    // simplest to just use the same logic core does, making std::string_view on
+    // our elements.
+    //
+    // Specifically, please see
+    // https://github.com/TileDB-Inc/TileDB/blob/2.27.2/tiledb/sm/array_schema/enumeration.cc#L417-L456
+    //
+    // It is important that we use core's logic here, so that when we are
+    // able to access its hashmap directly without constructing our own,
+    // that transition will be seamless.
```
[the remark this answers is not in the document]
johnkerl: @nguyenv no. `_find_enum_match` is a sequential search. Before this PR:

* If `std::string`: use a `std::unordered_set` as a performance improvement to check `enum_values_in_write` against the ones in the core enum
* Otherwise: use `find_enum_match` which is O(N^2)

As of this PR:

* Use `std::unordered_set` regardless

libtiledbsoma/src/soma/managed_query.h
```diff
@@ -1086,6 +1104,15 @@ bool ManagedQuery::_extend_and_evolve_schema_and_write<std::string>(
     Enumeration enmr,
     ArraySchemaEvolution& se);
 
+template <>
+bool ManagedQuery::_extend_and_evolve_schema<std::string>(
```
ihnorton, line 1108: nit: not clear to me why this templating is necessary at this level -- I _think_ the only true specialization is in `remap_indexes`, but might have missed something.

johnkerl: @ihnorton this was carefully and thoroughly discussed here as a team effort:
https://github.com/single-cell-data/TileDB-SOMA/pull/3888

apis/python/src/tiledbsoma/soma_dataframe.cc
```diff
@@ -156,6 +159,57 @@ void load_soma_dataframe(py::module& m) {
...
+                try {
+                    sdf.extend_enumeration_values(map_for_cpp, deduplicate);
+                } catch (std::range_error& e) {
+                    throw py::value_error(e.what());
+                } catch (const std::exception& e) {
+                    TPY_ERROR_LOC(e.what());
+                }
+                py::gil_scoped_acquire acquire;
+
+                for (i = 0; i < ncol; i++) {
```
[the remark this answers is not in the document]
johnkerl: @ihnorton thanks, I'll release in the try-catch

libtiledbsoma/src/soma/managed_query.h
johnkerl: What we could perhaps do is to move the `switch` on TileDB type from here
https://github.com/single-cell-data/TileDB-SOMA/blob/2609d8b4545a1d296def5bdc7a4b496429933221/libtiledbsoma/src/soma/managed_query.cc#L1125-L1166
down into a non-templated `ManagedQuery::_extend_and_evolve_schema`

ihnorton: This is fine to discuss in follow-up, the only impact is code gen time / binary size.

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -401,6 +401,35 @@ def get_enumeration_values(
...
         return self._handle.get_enumeration_values(column_names)
 
+    def extend_enumeration_values(
+        self: DataFrame, values: dict[str, pa.Array], deduplicate: bool = False
+    ) -> None:
+        """Extend enumeration values for each column defined in `values`.
+        Raises ``ValueError`` if any of the the specified column names is not in
+        the schema, or if any is not of Arrow dictionary type. All the values
+        must provided must not already exist in the schema (see the output of
+        ``get_enumeration_values``), unless ``deduplicate=True`` is provided."""
```
jp-dark, line 411:
```suggestion
        """Extend enumeration values for each column defined in `values`.
        
        Raises ``ValueError`` if any of the the specified column names is not in
        the schema, or if any is not of Arrow dictionary type. May only contain
        values that already exist in the schema (see the output of
        ``get_enumeration_values``)  unless ``deduplicate=True``."""
```

apis/python/src/tiledbsoma/soma_dataframe.cc
```diff
@@ -156,6 +159,57 @@ void load_soma_dataframe(py::module& m) {
...
+                std::vector<ArrowSchema> arrow_schemas(ncol);
+                std::vector<ArrowArray> arrow_arrays(ncol);
+                size_t i = 0;
+                std::map<std::string, std::pair<ArrowSchema*, ArrowArray*>>
+                    map_for_cpp;
+                for (auto item : values) {
+                    std::string column_name = py::str(item.first);
+                    py::object pa_array_for_column = item.second;
+                    uintptr_t arrow_schema_ptr = (uintptr_t)(&arrow_schemas[i]);
+                    uintptr_t arrow_array_ptr = (uintptr_t)(&arrow_arrays[i]);
```
jp-dark, line 183: Nit: Avoid c-style casting. Named casts provided more context for future developers and have fewer foot-guns.

libtiledbsoma/src/soma/managed_query.h
```diff
@@ -1086,6 +1104,15 @@ bool ManagedQuery::_extend_and_evolve_schema_and_write<std::string>(
...
+    ArrowSchema* value_schema,
+    ArrowArray* value_array,
+    std::string column_name,
```
jp-dark, line 1111: Do we mutate the `column_name` or is it safe to pass it by a const reference? (Similarly for the `column_name` parameter in the other new methods.)

libtiledbsoma/src/soma/managed_query.h
```diff
@@ -460,6 +460,14 @@ class ManagedQuery {
         Enumeration enmr,
         ArraySchemaEvolution& se);
 
+    bool _extend_enumeration(
```
jp-dark, line 463: Optional: Even though this is an internal function, a short doxygen summary can be helpful. Something like "Modifies the the array schema to contain the extended enumeration." Same for other new functions in the file.

jp-dark (approved):
I have a couple small comments, but overall this looks good to me.

libtiledbsoma/src/soma/managed_query.h
johnkerl: Sure. General advice I've seen over the years is const references aren't a worthwhile gain for small strings (which these are). However, it doesn't hurt; I'll make them const references.

apis/python/src/tiledbsoma/soma_dataframe.cc
johnkerl: For this spot:

```
error: static_cast from 'value_type *' (aka 'ArrowSchema *') to 'uintptr_t' (aka 'unsigned long') is not allowed
  182 |                     uintptr_t arrow_schema_ptr = static_cast<uintptr_t>(&arrow_schemas[i]);
      |                                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

:(

libtiledbsoma/src/soma/managed_query.h
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/3815/commits/532958e69f4763fce56b88ab1c7d9159ce78b921

apis/python/src/tiledbsoma/soma_dataframe.cc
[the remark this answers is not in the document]
johnkerl: @ihnorton https://github.com/single-cell-data/TileDB-SOMA/pull/3815/commits/c7221fb03702d218e7761211b3d3a2adf1de7d63 is the try-catch; I'll attempt your (separately communicated) `ScopedExecutor` next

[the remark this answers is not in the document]
johnkerl: @ihnorton several unit-test cases are failing with this -- I put up #3901 separately, atop this PR, so as not to ungreen this current PR's current green-CI status

I'll need to debug the failures tomorrow

Meanwhile, thanks!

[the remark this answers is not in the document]
johnkerl: @ihnorton success!

It just needed this:

```diff
-                ScopedExecutor([&]() {
+                ScopedExecutor cleanup([&]() {
```

-- now that it's named, it runs at the end as intended.

Thanks!

https://github.com/single-cell-data/TileDB-SOMA/pull/3815/commits/c7f8a44336b7e6259d14ed3998634c47141f1aa1

johnkerl:
@ihnorton all CI is green and all suggestions are in to the best of my ability and there are three other accepts.

To unblock @bkmartinjr 's #3865 I'm going to merge this now.

If there are any things we missed (or I missed) I can/will revisit them as immediate follow-ons.

libtiledbsoma/src/soma/managed_query.h
ihnorton: No objection, we can have a separate discussion about where to draw the line between generic and type-specialized code.

apis/python/src/tiledbsoma/soma_dataframe.cc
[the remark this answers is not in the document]
ihnorton: LGTM, thank you.

PR #3834, opened 2025-03-21
[r] Repace `$attributes()` accessor with one that uses libtiledbsoma

mojaveazure:
Replace `TileDBArray$attributes()` with new `SOMAArrayBase$attributes()`; this method uses a libtiledbsoma/Rcpp function rather than `tiledb::attrs()` and returns a named list instead of a tiledb-r wrapper around external pointers

[SC-64814](https://app.shortcut.com/tiledb-inc/story/64814)

continues work for #2406


johnkerl (approved):
🚢 
Thank you @mojaveazure !!! :)

PR #3836, opened 2025-03-21
[c++,python] Pass timestamp when creating `ArraySchema`

nguyenv:
**Issue and/or context:**

https://github.com/single-cell-data/TileDB-SOMA/issues/2879
[[sc-61254](https://app.shortcut.com/tiledb-inc/story/61254/python-array-resize-and-time-travel-not-working-correctly)]

- This is built on top of @johnkerl's https://github.com/single-cell-data/TileDB-SOMA/pull/2940
- This can only pass with https://github.com/TileDB-Inc/TileDB/pull/5484

**Changes:**

- Move `_make_se` to `SOMAArray` and use in relevant `CurrentDomain` methods (note that we still use `ArraySchemaEvolution` with timestamp in `ManagedQuery` but it is only used in a single location so a separate helper method is unnecessary)
- In `ArrowAdapter::tiledb_schema_from_arrow_schema`, use `tiledb_array_schema_alloc{,_at_timestamp}` C API "constructors"
- Add unit tests for extending enumerations and resizing the current domain at timestamps


bkmartinjr (approved):
change looks good

do we need a C++ test as well? (question, not a requirement)

PR #3845, opened 2025-03-25
[c++] Correctly handle NaN category values

nguyenv:
**Issue and/or context:**

[[sc-62449](https://app.shortcut.com/tiledb-inc/story/62449/python-c-enum-evolution-does-not-correctly-handle-category-value-of-nan)]

`std::find` does not work "as expected" for NaN values, since NaN != NaN (which is the correct behavior for floating-point comparisons, but not what is intended in this context).

**Changes:**

Add `ManagedQuery::_find_enum_match` which uses bitwise comparison. Like `std::find`, the method returns an iterator that points to the match value in the container; if not found, returns an iterator to the end of the container.

### DIFF
--- a/apis/python/tests/test_dataframe.py
+++ b/apis/python/tests/test_dataframe.py
@@ -2370,9 +2370,7 @@ def test_enum_handling_category_of_nan_62449(tmp_path):
             "soma_joinid": [4, 5],
             "A": pa.DictionaryArray.from_arrays(
                 indices=pa.array([1, 0], type=pa.int32()),
-                dictionary=pa.array(
-                    [quiet_nan, signaling_nan], type=pa.float32()
-                ),
+                dictionary=pa.array([quiet_nan, signaling_nan], type=pa.float32()),
             ),
         }
     )
--- a/apis/python/tests/test_dataframe.py
+++ b/apis/python/tests/test_dataframe.py
@@ -2365,6 +2365,7 @@ def test_enum_handling_category_of_nan_62449(tmp_path):
         assert len(actual_dict_vals) == 3
         assert nan_check(quiet_nan, actual_dict_vals)
         assert nan_check(negative_nan, actual_dict_vals)
+        assert nan_check(signaling_nan, actual_dict_vals)
         assert_array_equal(expected_data1["A"], actual_data)
 
     # Ensure that the dictionary indexes get shifted correctly when appending
--- a/apis/python/tests/test_dataframe.py
+++ b/apis/python/tests/test_dataframe.py
@@ -2322,6 +2322,7 @@ def test_enum_regression_62887(tmp_path):
 def test_enum_handling_category_of_nan_62449(tmp_path):
     uri = tmp_path.as_posix()
 
+    # Different representations of single-precision NaNs
     quiet_nan = struct.unpack(">f", b"\x7f\xc0\x00\x00")[0]
     negative_nan = struct.unpack(">f", b"\xff\xc0\x00\x00")[0]
     signaling_nan = struct.unpack(">f", b"\x7f\x80\x00\x01")[0]
--- a/apis/python/tests/test_dataframe.py
+++ b/apis/python/tests/test_dataframe.py
@@ -2341,7 +2341,8 @@ def test_enum_handling_category_of_nan_62449(tmp_path):
         ]
     )
 
-    tbl1 = pa.Table.from_pydict(
+    # Ensure that unique NaN values are respected as different dictionary values
+    expected_data1 = pa.Table.from_pydict(
         {
             "soma_joinid": [0, 1, 2, 3],
             "A": pa.DictionaryArray.from_arrays(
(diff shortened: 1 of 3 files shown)


libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -1239,10 +1239,30 @@ ManagedQuery::_extend_and_evolve_schema_with_details(
     // Find any new enumeration values
     std::vector<ValueType> enum_values_to_add;
     for (auto enum_val : enum_values_in_write) {
-        if (std::find(
-                enum_values_existing.begin(),
-                enum_values_existing.end(),
-                enum_val) == enum_values_existing.end()) {
+        bool is_new_value = false;
+
+        if (std::isnan(enum_val)) {
```
bkmartinjr, line 1244: This check has different semantics than TileDB core and pyarrow dictionary - they both use binary (bit) equivalence to determine if the enum is the same.

It matters for NaN, as there are many different bit patterns for NaN, and in Arrow and TileDB core they can all end up as different enum values, but will be treated as equivalent in this test.

Example, in float32, both of these are NaN (little-endian layout):
* `0x7fc00000`
* `0x7fc00100`

Here is a trivial test that shows the effect using pyarrow dictionary_encode - note how the different NaNs are separated encoded:
```
In [63]: np.array([ 1, np.nan, struct.unpack('>f', b'\x7f\xc0\x01\x00')[0]],dtype=np.float32)
Out[63]: array([ 1., nan, nan], dtype=float32)

In [64]: pa.array(np.array([ 1, np.nan, struct.unpack('>f', b'\x7f\xc0\x01\x00')[0]],dtype=np.float32)).dictionary_encode()
Out[64]: 
<pyarrow.lib.DictionaryArray object at 0x7b1d4633e3b0>

-- dictionary:
  [
    1,
    nan,
    nan
  ]
-- indices:
  [
    0,
    1,
    2
  ]
```


I suggest we follow the Arrow/TileDB precedent of equality defined by bit-equality, i.e., if the type is the same, equality is fined by bit-wise equality

If we don't do this, then we are silently de-duplicating all variants of NaN, which is likely to be surprising.

Note regarding testing:  both `math.nan` and `np.nan` will produce the same bit encoding (0xc07f0000 for little-endian float32), so you will need to do something like the above to create a non-standard NaN.

bkmartinjr (changes requested):
Request that we follow the same equality definition used in Core and Arrow (bitwise equality), rather than de-duplicating NaNs which are not bitwise equal.

apis/python/tests/test_dataframe.py
```diff
@@ -2321,6 +2322,17 @@ def test_enum_regression_62887(tmp_path):
 def test_enum_handling_category_of_nan_62449(tmp_path):
     uri = tmp_path.as_posix()
 
+    quiet_nan = struct.unpack(">f", b"\x7f\xc0\x00\x00")[0]
```
johnkerl, line 2325: Just fyi these are single-precision NaNs

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -1239,30 +1239,23 @@ ManagedQuery::_extend_and_evolve_schema_with_details(
...
-                    enum_values_existing.end(),
-                    enum_val) != enum_values_existing.end()) {
-                is_new_value = true;
-            }
-        }
+        // Check if the value already exists using bitwise comparison.
+        // - We cannot use std::find because NaN != NaN
+        // - We cannot use std::isnan because multiple bit representations
+        //   exist for NaN. For example, in float32 (little-endian), both
+        //   0x7fc00000 and 0x7fc00100 represent NaN
```
johnkerl, line 1246: Actually we _totally could_ use `std::isnan`. That's not a problem. The problem is that core uses bit-equality to check against a particular NaN value, while `std::isnan` would check against all possible NaN values. In an ideal world, the latter would (IMO) be better. But as @bkmartinjr pointed out, let's stick with core's bit-equality checking, solely for consistency with core.

XanthosXanthopoulos (changes requested):
`_remap_indexes_aux` is also affected [L834-L836](https://github.com/single-cell-data/TileDB-SOMA/blob/2ed844c779816d37d2fe5e26975bfdebead5e97e/libtiledbsoma/src/soma/managed_query.h#L834-L836)

libtiledbsoma/src/soma/managed_query.h
```diff
@@ -1028,6 +1027,28 @@ class ManagedQuery {
...
+     * different bit patterns as distinct values, we also cannot
+     * rely onstd::isnan and must use to bitwise comparisons.
+     *
+     * @param values the vector of values to search through
+     * @param target the value to search for
+     * @return typename std::vector<T>::const_iterator iterator
+     * to the found element or values.end() if not found
+     */
+    template <typename T>
+    typename std::vector<T>::const_iterator _find_enum_match(
```
bkmartinjr, line 1045: A thought about this -- it is O(N^2) in the worst case.  It will work great for most enum value lists, which are of modest size.

However, we occasionally get large (O(100K)) or even more rarely very large O(1M+) enum value lists.  The O(100K) lists would be ontologies like tissue (UBERON) or cell type (CL).

At some point we may need to consider more efficient algorithms, and potentially pick between them.

@XanthosXanthopoulos - will the upcoming Core changes exposing the underlying enum map address this?  I  believe it is covering this case, but flagging in case it doesn't

bkmartinjr (approved):

PR #3850, opened 2025-03-26
[r] Replace calls to tiledb-r schema checkers with libtiledbsoma

mojaveazure:
Replace calls the following tiledb-r schema accessors with new libtiledbsoma functions:
 - `tiledb::capacity()` replaced with `c_capacity()`
 - `tiledb::tile_order()` replaced with `c_tile_order()`
 - `tiledb::cell_order()` replaced with `c_cell_order()`

[SC-64997](https://app.shortcut.com/tiledb-inc/story/64997)

continues work for #2406

### DIFF
--- a/apis/r/src/rinterface.cpp
+++ b/apis/r/src/rinterface.cpp
@@ -444,25 +444,6 @@ double c_capacity(const std::string& uri, Rcpp::XPtr<somactx_wrap_t> ctxxp) {
     return static_cast<double>(cap);
 }
 
-// Taken from tiledb-r
-// https://github.com/TileDB-Inc/TileDB-R/blob/525bdfc0f34aadb74a312a5d8428bd07819a8f83/src/libtiledb.cpp#L246C1-L261C2
-const char *_tiledb_layout_to_string(tiledb_layout_t layout) {
-    switch (layout) {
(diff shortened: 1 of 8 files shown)


apis/r/src/rinterface.cpp
```diff
@@ -434,6 +434,55 @@ bool c_allows_dups(const std::string& uri, Rcpp::XPtr<somactx_wrap_t> ctxxp) {
...
+double c_capacity(const std::string& uri, Rcpp::XPtr<somactx_wrap_t> ctxxp) {
+    auto sr = tdbs::SOMAArray::open(OpenMode::read, uri, ctxxp->ctxptr);
+    std::shared_ptr<tiledb::ArraySchema> sch = sr->tiledb_schema();
+    sr->close();
+
+    uint64_t cap = sch->capacity();
+    return static_cast<double>(cap);
+}
+
+// Taken from tiledb-r
```
johnkerl, line 447: Maybe move to `apis/r/src/rutilities.cpp`

johnkerl (approved):
🚢 🚢 🚢

apis/r/src/rinterface.cpp
mojaveazure: Moved in 2c35d5a

PR #3851, opened 2025-03-26
[r] Replace calls to `tiledb::filter_list()` on schemas

mojaveazure:
Replace schema filter checking with new libtiledbsoma/Rcpp `c_schema_filters()` function to return a named list of schema fitlers

[SC-64998](https://app.shortcut.com/tiledb-inc/story/64998)

continues work for #2406


apis/r/src/rinterface.cpp
```diff
@@ -434,6 +434,114 @@ bool c_allows_dups(const std::string& uri, Rcpp::XPtr<somactx_wrap_t> ctxxp) {
     return sch->allows_dups();
 }
 
+// Taken from tiledb-r
```
johnkerl, line 437: Just a thought, maybe put this in `apis/r/src/rutilities.cpp`

apis/r/src/rinterface.cpp
```diff
@@ -434,6 +434,114 @@ bool c_allows_dups(const std::string& uri, Rcpp::XPtr<somactx_wrap_t> ctxxp) {
...
+// internal helper function
+// filter options taken from tiledb-r
+// https://github.com/TileDB-Inc/TileDB-R/blob/525bdfc0f34aadb74a312a5d8428bd07819a8f83/src/libtiledb.cpp#L369-L388
+Rcpp::List _get_filter_opts(Rcpp::XPtr<tiledb::Filter> filter) {
+    auto filter_type = _tiledb_filter_to_string(filter->filter_type());
+
+    // catches are intentionally blank to leave the values as NA
+    int compression = R_NaInt, bytewidth = R_NaInt, factor = R_NaInt, offset = R_NaInt;
+    uint32_t bit_width = R_NaInt, positive_delta = R_NaInt;
+    try {filter->get_option(TILEDB_COMPRESSION_LEVEL, &compression);} catch (...) {};
```
johnkerl, line 488: This is a catch-anything-at-all ... which can hide other errors we might not want to ignore ... @nguyenv may have advice here ... if not, we can dig a bit into the core API & see if there's a has-option we can call ...

johnkerl: Looks like `Filter::to_str(filter)`
https://github.com/single-cell-data/TileDB-SOMA/blob/1.16.1/libtiledbsoma/test/unit_soma_dataframe.cc#L343

apis/r/src/rinterface.cpp
mojaveazure: Moved in b7b004e

apis/r/src/rinterface.cpp
mojaveazure: @johnkerl I'm not sure `Filter::to_str()` would work here; I'm trying to get filter options, not their name. libtiledb throws an error if you try to get a non-existent option, hence the try-catch. I can't figure out what class the errors are (I'm not sure they are), and don't know how to parse an error message in C++, so this is why I have the catch-all

johnkerl: ok let's chat a bit ... we do deconstruct core error messages quite a bit in libtiledbsoma so there is copyable precedent

apis/r/src/rutilities.cpp
```diff
@@ -544,3 +544,49 @@ const char *_tiledb_layout_to_string(tiledb_layout_t layout) {
...
 }
+
+// internal helper function for
+// `_get_filter_opts()`
+// <Numeric> should be `int32_t` or `double`
+template <typename Numeric>
+Numeric _get_filter_opt(Rcpp::XPtr<tiledb::Filter> filter, tiledb_filter_option_t option) {
+    try {
+        switch (option) {
+        case TILEDB_BIT_WIDTH_MAX_WINDOW: case TILEDB_POSITIVE_DELTA_MAX_WINDOW:
```
johnkerl, line 555:
```suggestion
        case TILEDB_BIT_WIDTH_MAX_WINDOW:
        case TILEDB_POSITIVE_DELTA_MAX_WINDOW:
```

johnkerl (approved):
Nice! LGTM

Some formatting nits is all :)

apis/r/src/rutilities.cpp
```diff
@@ -544,3 +544,50 @@ const char *_tiledb_layout_to_string(tiledb_layout_t layout) {
...
+        }
+        if (msg.find("Filter does not support options") != std::string::npos) {
+            if (std::is_same<Numeric, double>::value) {
+                return R_NaReal;
+            }
+            return R_NaInt;
+        }
+        throw(e);
+    }
+}
```
jp-dark, line 579: @mojaveazure and I have already discussed this, but I strongly believe try/catch should be avoided when possible except when re-throwing. I think we should push on the C++ API to add the functions we need to do this in a safer way.

johnkerl: @jp-dark I completely agree with you regarding behavior of an ideal API.

We need to call the core API we have, working around its shortcomings, request an improved API in some upcoming core release, and do the best we can (which is this) until then.

mojaveazure: @jp-dark @johnkerl I'm going to land this PR now as:
- core doesn't have a safer API to use
- this isn't going directly to `main` but into a separate branch, so it should be easy to replace once a safer API exists

I've filed [SC-65200](https://app.shortcut.com/tiledb-inc/story/65200) to track adopting a safer API once it exists

apis/r/src/rutilities.cpp
```diff
@@ -544,3 +544,51 @@ const char *_tiledb_layout_to_string(tiledb_layout_t layout) {
...
+// https://github.com/TileDB-Inc/TileDB-R/blob/525bdfc0f34aadb74a312a5d8428bd07819a8f83/src/libtiledb.cpp#L369-L388
+Rcpp::List _get_filter_options(Rcpp::XPtr<tiledb::Filter> filter) {
+    return Rcpp::List::create(
+        Rcpp::Named("filter_type") = tiledb::Filter::to_str(filter->filter_type()),
+        Rcpp::Named("compression_level") = _get_filter_option<int32_t>(filter, TILEDB_COMPRESSION_LEVEL),
+        Rcpp::Named("bit_width") = _get_filter_option<int32_t>(filter, TILEDB_BIT_WIDTH_MAX_WINDOW),
+        Rcpp::Named("positive_delta") = _get_filter_option<int32_t>(filter, TILEDB_POSITIVE_DELTA_MAX_WINDOW),
+        Rcpp::Named("float_bytewidth") = _get_filter_option<double>(filter, TILEDB_SCALE_FLOAT_BYTEWIDTH),
+        Rcpp::Named("float_factor") = _get_filter_option<double>(filter, TILEDB_SCALE_FLOAT_FACTOR),
+        Rcpp::Named("float_offset") = _get_filter_option<double>(filter, TILEDB_SCALE_FLOAT_OFFSET)
```
teo-tsirpanis, line 592: Not all of these options make sense for all filters. It would be better to have a logic of `if (filter is compression filter) { add "compression_level" to the list }` and so on for the other filter types. This would avoid the `try/catch`es above.

PR #3854, opened 2025-03-26
[r] Replace `tiledb::domain()` with Rcpp `c_domain()`

mojaveazure:
Replace calls to `tiledb::domain()` and other domain accessors with libtiledbsom/Rcpp `c_domain()`

[SC-65000](https://app.shortcut.com/tiledb-inc/story/65000)

continues work for #2406

Requires #3851

### DIFF
--- a/apis/r/src/rinterface.cpp
+++ b/apis/r/src/rinterface.cpp
@@ -726,9 +726,8 @@ Rcpp::List c_domain(const std::string& uri, Rcpp::XPtr<somactx_wrap_t> ctxxp) {
             Rcpp::Named("name") = name,
             Rcpp::Named("type") = _tiledb_datatype_to_string(dim->type()),
             Rcpp::Named("ncells") = _get_ncells<Rcpp::XPtr<tiledb::Dimension>>(dim),
-            // TODO: add domain and tile to return value
-            // Rcpp::Named("domain") = "",
-            // Rcpp::Named("tile") = "",
+            Rcpp::Named("domain") = _get_dim_domain(dim),
+            Rcpp::Named("tile") = _get_dim_tile(dim),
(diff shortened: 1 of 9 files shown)


apis/r/src/rinterface.cpp
```diff
@@ -585,6 +585,50 @@ const char *_tiledb_filter_to_string(tiledb_filter_type_t filter) {
...
 
+// internal helper function
+// filter options taken from tiledb-r
+// https://github.com/TileDB-Inc/TileDB-R/blob/525bdfc0f34aadb74a312a5d8428bd07819a8f83/src/libtiledb.cpp#L369-L388
+Rcpp::List _get_filter_opts(Rcpp::XPtr<tiledb::Filter> filter) {
+    auto filter_type = _tiledb_filter_to_string(filter->filter_type());
+
+    // catches are intentionally blank to leave the values as NA
+    int compression = R_NaInt, bytewidth = R_NaInt, factor = R_NaInt, offset = R_NaInt;
+    uint32_t bit_width = R_NaInt, positive_delta = R_NaInt;
```
johnkerl, line 596: Let's check with the core folks and see if there's a more elegant way to do this.

I know you're copying from existing code in TileDB-R, so, obviously leaving this way isn't the worst thing -- however, a catch-anything can mask other unexpected errors. If we can find a more elegant way, I'd love for us to do so. cc @teo-tsirpanis .

mojaveazure: This function is also implemented in #3851, and so whatever resolution we adopt for https://github.com/single-cell-data/TileDB-SOMA/pull/3851#discussion_r2014973214 will end up being used here

mojaveazure: Addresed in #3851

PR #3857, opened 2025-03-26
[c++/python] Unit-test and fix `get_enumeration_values` for boolean types

johnkerl:
**Issue and/or context:** For issue #3785 / [[sc-63930]](https://app.shortcut.com/tiledb-inc/story/63930/dataframe-need-api-to-extend-column-enum-values-without-doing-a-write#activity-64471).

**Changes:** @bkmartinjr found during his testing that the unit-test coverage in #3784 was insufficient to surface a defect. This PR includes unit-test cases that reproduce this defect, along with a fix for that defect.

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/utils/arrow_adapter.h
+++ b/libtiledbsoma/src/utils/arrow_adapter.h
@@ -608,7 +608,7 @@ class ArrowAdapter {
 
         // Two-buffer model for non-string data:
         // * Slot 0 is the Arrow validity buffer which we leave null
-        // * Slot 1 is data, void* but will be derefrenced as T*
+        // * Slot 1 is data, void* but will be dereferenced as T*
         // * There is no offset information
         arrow_array->n_buffers = 2;
         arrow_array->n_children = 0;  // leaf/child node
@@ -663,7 +663,7 @@ class ArrowAdapter {
         // * Slot 0 is the Arrow uint8_t* validity buffer
         // * Slot 1 is the Arrow offsets buffer: uint32_t* for Arrow string
         //   or uint64_t* for Arrow large_string
-        // * Slot 2 is data, void* but will be derefrenced as T*
+        // * Slot 2 is data, void* but will be dereferenced as T*
         arrow_array->n_buffers = 3;
         arrow_array->buffers = (const void**)malloc(3 * sizeof(void*));
         arrow_array->n_children = 0;      // leaf/child node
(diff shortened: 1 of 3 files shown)


libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -702,8 +702,61 @@ class ArrowAdapter {
         return arrow_array;
     }
 
-    // These table-column getters are, as of this writing, intended primarily
-    // for keystroke-reduction in unit-test cases.
+    // Given TileDB 8-bit booleans, packs them to 1-bit Arrow-style booleans.
+    static ArrowArray* make_arrow_array_child_bool(
+        const std::vector<uint8_t>& v) {
+        // Use malloc here, not new, to match ArrowAdapter::release_array
+        auto arrow_array = (ArrowArray*)malloc(sizeof(ArrowArray));
```
bkmartinjr, line 709: possibly naive question:  if we are using nanoarrow, why not use the nanoarrow array allocation routines?

(see: https://arrow.apache.org/nanoarrow/0.2.0/c.html)

libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -702,8 +702,61 @@ class ArrowAdapter {
...
+
+        arrow_array->buffers[0] = nullptr;
+        // Use malloc here, not new, to match ArrowAdapter::release_array.
+        // When computing the number of bits needed to store the n bits, round
+        // up.
+        auto nbytes = (n + 7) / 8;
+        uint8_t* dest = (uint8_t*)malloc(nbytes * sizeof(uint8_t));
+        memset(dest, 0, nbytes);
+
+        for (size_t i = 0; i < n; i++) {
```
bkmartinjr, line 745: I believe this is redundant with the nanoarrow routines which do the same thing [see here](https://arrow.apache.org/nanoarrow/0.2.0/c.html#bitmap-utilities)

bkmartinjr (commented):
functionally looks great. however, suggest that you make use of the nanoarrow allocation and bitset routines, which would result in a lot less code (and use very well tested code).

libtiledbsoma/src/utils/arrow_adapter.h
johnkerl: Apparently we have not had the pattern of using that. TIL, thank you for the feedback.

If you feel it would be worth the dev time to refactor our code to use that API in this library, please file a tracking issue.

Thanks again @bkmartinjr !

bkmartinjr: I'm not going to hold up this PR.  But this is exactly what got us into the Arrow offset bug-fest previously. If we would use the provided API, I believe these issues would be less likely to occur.

So yes, we should be using a framework (I thought we were!).  I see this as your decision as eng lead, so I just leave you with this thought.

bkmartinjr (approved):
approving, with concerns noted (and already acknowledged) elsewhere

libtiledbsoma/src/utils/arrow_adapter.h
johnkerl: (I saw the "Unsafe" in that function name earlier today and quickly icked away from using it. I am writing code which I don't feel deserves worthy of being called "unsafe". I am happy with my work product, and am unit-testing it to my satisfaction. I'm putting up a PR for review here, for the express purpose of seeing if anyone finds it in need of further testing.)

bkmartinjr: unsafe just means you need to do your own bounds checks and allocations.

the code is quite straightforward if you want to have a read (in nanoarrow.h)

johnkerl: See also https://github.com/single-cell-data/TileDB-SOMA/pull/3857#discussion_r2016596989

johnkerl:
Follow-ons here https://github.com/single-cell-data/TileDB-SOMA/issues/3860

PR #3872, opened 2025-03-30
[python] Ingest from S3 performance

bkmartinjr:
**Issue and/or context:**

`soma.io.from_h5ad` will read H5ADs from any path VFS supports. However, performance on S3 is extremely slow.

Analysis shows that this is caused by two issues:
* `h5py` (which underpins AnnData) assumes it is reading from a file system (eg, issues many small reads, does not read the object sequentially, etc)
* The current implementation translates those requests directly to VFS read calls, which is a documented antipattern (see vfs.h header where it recommends issuing fewer, larger reads)

**Changes:**

This PR introduces a Python caching reader, similar in concept to `io.BufferedReader`, to mediate between `h5Py` and VFS. It introduces several changes in behavior:
* the minimum read size is increased significantly (configurable)
* previously read blocks are maintained in an LRU cache (size configurable)
* the file-like object API `readinto` is added, reducing memory copies (h5py will preferentially use this API if it exists)

The PR also adds `SOMAVFS` python bindings required for common usage as a "file-like" object (i.e., those features of `RawIOBase`, necessary to wrap in buffering classes like `io.BufferedReader`), including `readinto`, `closed`, `readable` and other methods.

Benchmarking on a wide variety of H5ADs resident on S3 have resulted in the current configuration of 8MiB blocks, and a total cache size of 64MiB.

Benchmarking on EC2/S3 across a wide range of H5AD sizes show that this PR provides a 30-60X speed up in  total time to perform `from_h5ad` (this time _includes_ writing the Experiment).  I also compared `from_h5ad` to the time required to download and perform the same operation on the local file system - it is now within a factor of 2X of the "make a local copy" time.

**Notes for Reviewer:**

This is one of two PRs relating to ingestion performance.  Related: [sc-43827]

This PR also made a small change to GH CI.  Previously, the installed python packages were logged prior to running tests.  At some point, this was removed and they were only logged prior to running pre-commits.  I added the logging back, as an aid to solving dependency related failures.

### DIFF
--- a/apis/python/src/tiledbsoma/io/_caching_reader.py
+++ b/apis/python/src/tiledbsoma/io/_caching_reader.py
@@ -2,13 +2,20 @@ from __future__ import annotations
 
 import ctypes
 import io
+import sys
 import threading
 from collections import OrderedDict
-from typing import Any, cast
+from typing import Any, Union, cast
 
 import attrs
 import pyarrow as pa
-from typing_extensions import Buffer
+from typing_extensions import Buffer, TypeAlias
+
+# typeguard and cython memoryview incompat in older versions
+if sys.version_info >= (3, 12):
+    WritableBuffer: TypeAlias = Union[memoryview, Buffer]
+else:
+    WritableBuffer: TypeAlias = Any
 
 
 @attrs.define
@@ -132,7 +139,7 @@ class CachingReader:
         self._pos += len(b)
         return cast(bytes, b)
 
-    def readinto(self, buf: Buffer | memoryview) -> int | None:
+    def readinto(self, buf: WritableBuffer) -> int | None:
         if not isinstance(buf, memoryview):
             buf = memoryview(buf)
         if buf.nbytes == 0:
--- a/apis/python/src/tiledbsoma/io/_caching_reader.py
+++ b/apis/python/src/tiledbsoma/io/_caching_reader.py
@@ -13,6 +13,7 @@ from typing_extensions import Buffer
 
 @attrs.define
 class CacheStats:
+    hit: int = 0
     miss: int = 0
 
 
@@ -35,7 +36,7 @@ class CachingReader:
         self,
         file: Any,  # file-like object. Unfortunately, Python lacks a good typing signature for this concept.
         *,
-        memory_budget: int = 128 * 1024**2,
+        memory_budget: int = 64 * 1024**2,
         cache_block_size: int = 1024**2,
     ):
         if not file.readable():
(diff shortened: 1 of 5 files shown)


apis/python/src/tiledbsoma/io/_caching_reader.py
```diff
@@ -0,0 +1,199 @@
...
+    def close(self) -> None:
+        self._reset_cache()
+        self._file.close()
+
+    def tell(self) -> int:
+        return self._pos
+
+    def seek(self, offset: int, whence: int = 0) -> int:
+        if whence not in [io.SEEK_SET, io.SEEK_CUR, io.SEEK_END]:
+            raise ValueError("Invalid whence value")
```
nguyenv, line 176:
```suggestion
        if whence not in [io.SEEK_SET, io.SEEK_CUR, io.SEEK_END]:
            raise ValueError(f"Invalid whence value ({whence})")
```

nguyenv (commented):
I have not ever used the `@lru_cache` built-in Python decorator before and am only aware that it exists. Is that something we could use here, or is it necessary for us to fully implment a custom LRU cache?

bkmartinjr:
> I have not ever used the `@lru_cache` built-in Python decorator before and am only aware that it exists. Is that something we could use here, or is it necessary for us to fully implment a custom LRU cache?

That API is intended for function memoziation.  I can imagine ways to use it - for eample by wrapping a "get IO block" function with it - but that would not be a typical pattern and IMHO a bit confusing.  And given how simple the cache is using an OrderedDict, I think the code is simpler without wrapping in this abstraction.

nguyenv (approved):

apis/python/src/tiledbsoma/io/_caching_reader.py
```diff
@@ -0,0 +1,199 @@
...
+
+        if whence == io.SEEK_END:
+            offset = self._file_length + offset
+        elif whence == io.SEEK_CUR:
+            offset += self._pos
+        elif whence == io.SEEK_SET:
+            offset = offset
+
+        if offset < 0:
+            raise OSError("seek() returned invalid position")
```
jp-dark, line 186: This doesn't match the `IOBase` definition of `seek` for `io.SEEK_END` and `io.SEEK_CUR`:

> Change the stream position to the given byte offset, interpreted relative to the position indicated by whence, and return the new absolute position. Values for whence are:

    [os.SEEK_SET](https://docs.python.org/3/library/os.html#os.SEEK_SET) or 0 – start of the stream (the default); offset should be zero or positive

    [os.SEEK_CUR](https://docs.python.org/3/library/os.html#os.SEEK_CUR) or 1 – current stream position; offset may be negative

    [os.SEEK_END](https://docs.python.org/3/library/os.html#os.SEEK_END) or 2 – end of the stream; offset is usually negative

jp-dark (commented):
I'm still looking this over, but have a couple initial comments:

* Noted inline: the `seek` function doesn't match the `IOBase` API. I'm not sure if that is intentional or not.
* Adding docstrings will help with both reviewing/maintaining.
* Nit: Use normal grammar (capitalization and punctuation) for error messages.

apis/python/src/tiledbsoma/io/_caching_reader.py
bkmartinjr: which part doesn't match?  it looks like they match to my eye, but I'm probably missing something

bkmartinjr:
@jp-dark 

> * Noted inline: the `seek` function doesn't match the `IOBase` API. I'm not sure if that is intentional or not.

Intentional.  I implemented the parts of the file-like object API that are needed for h5py.  I left out stuff like write support, detach, etc.

We _could_ do that extra stuff, but I didn't think we wanted dead code to maintain.  

Thoughts?

apis/python/src/tiledbsoma/io/_caching_reader.py
bkmartinjr: I'll add, whatever mistake I'm making must have been replicated into the test code, as I explicitly tested it against a Python built-in file and got the same results.   Weird....

bkmartinjr: I have slightly refactored the code for clarity - it is exactly the same logic, but (I believe) easier to read.  Let me know if you still see an issue

jp-dark: I was wrong about there being a mismatch. Both versions look fine, I just flipped the offset and position in my head.

bkmartinjr: lol, I've been there.  The good news is that it is now easier to read, so an improvement I think.

apis/python/src/tiledbsoma/io/_caching_reader.py
```diff
@@ -0,0 +1,223 @@
...
+    Primary use case is wrapping a VFS FileBuffer object, to improve performance of read
+    operations by doing fewer, larger reads.
+
+    The implemtation stores cached data as PyArrow arrays, primarily to benefit from the
+    easy, minimal-copy concat and slice operations upon underlying buffers.
+    """
+
+    def __init__(
+        self,
+        file: Any,  # file-like object. Unfortunately, Python lacks a good typing signature for this concept.
```
jp-dark, line 53: Optional: We could use either a protocol or `clib.SOMAVFSFilebuf` for typing here, although I think the comment is probably sufficient.

jp-dark (approved):

apis/python/src/tiledbsoma/io/_caching_reader.py
bkmartinjr: I had several thoughts:
* I wanted to leave the option open for other file-like objects to use this in the future.  Not sure that is very compelling though :-)
* this module doesn't depend on any behavior specific to `clib.SOMAVFSFilebuf`
* typing it as `clib.SOMAVFSFilebuf` is the same as `Any`, as that type is not defined as far as mypy knows (if you do a `reveal_type` on anything from `clib`, they are all `Any` from a static typing perspective)

Issue #3860, opened 2025-03-27
[c++] Refactor `ArrowAdapter` to make more use of `nanoarrow` library routines


johnkerl:
Follow-up from https://github.com/single-cell-data/TileDB-SOMA/pull/3857

johnkerl:
@jp-dark will be leading this effort.

jp-dark:
Closing in favor of an Epic: https://app.shortcut.com/tiledb-inc/epic/65366

PR #3873, opened 2025-03-30
[c++] Fix Arrow memory leaks in unit tests

XanthosXanthopoulos:
**Issue and/or context:** [sc-65116](https://app.shortcut.com/tiledb-inc/story/65116/c-fix-asan-errors-in-unit-tests)

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/soma/managed_query.h
+++ b/libtiledbsoma/src/soma/managed_query.h
@@ -304,9 +304,7 @@ class ManagedQuery {
      * @param arrow_schema
      * @param arrow_array
      */
-    void set_array_data(
-        const std::unique_ptr<ArrowSchema>& arrow_schema,
-        const std::unique_ptr<ArrowArray>& arrow_array);
+    void set_array_data(ArrowSchema* arrow_schema, ArrowArray* arrow_array);
 
     std::optional<std::shared_ptr<ArrayBuffers>> read_next();
 
--- a/libtiledbsoma/src/soma/managed_query.h
+++ b/libtiledbsoma/src/soma/managed_query.h
@@ -305,8 +305,8 @@ class ManagedQuery {
      * @param arrow_array
      */
     void set_array_data(
-        std::unique_ptr<ArrowSchema> arrow_schema,
-        std::unique_ptr<ArrowArray> arrow_array);
+        const std::unique_ptr<ArrowSchema>& arrow_schema,
+        const std::unique_ptr<ArrowArray>& arrow_array);
 
     std::optional<std::shared_ptr<ArrayBuffers>> read_next();
 
--- a/libtiledbsoma/src/soma/soma_collection.cc
+++ b/libtiledbsoma/src/soma/soma_collection.cc
@@ -100,8 +100,8 @@ std::shared_ptr<SOMAExperiment> SOMACollection::add_new_experiment(
     std::string_view uri,
     URIType uri_type,
     std::shared_ptr<SOMAContext> ctx,
-    std::unique_ptr<ArrowSchema> schema,
-    ArrowTable index_columns,
+    const std::unique_ptr<ArrowSchema>& schema,
+    const ArrowTable& index_columns,
     PlatformConfig platform_config,
     std::optional<TimestampRange> timestamp) {
     if (!timestamp) {
@@ -109,13 +109,7 @@ std::shared_ptr<SOMAExperiment> SOMACollection::add_new_experiment(
     }
 
     SOMAExperiment::create(
-        uri,
-        std::move(schema),
-        ArrowTable(
-            std::move(index_columns.first), std::move(index_columns.second)),
-        ctx,
-        platform_config,
-        timestamp);
+        uri, schema, index_columns, ctx, platform_config, timestamp);
 
     // Note that we must return a shared_ptr to the member, instead of a
     // unique_ptr because we place the SOMA object into the `children_` cache
@@ -132,8 +126,8 @@ std::shared_ptr<SOMAMeasurement> SOMACollection::add_new_measurement(
     std::string_view uri,
     URIType uri_type,
     std::shared_ptr<SOMAContext> ctx,
-    std::unique_ptr<ArrowSchema> schema,
-    ArrowTable index_columns,
+    const std::unique_ptr<ArrowSchema>& schema,
+    const ArrowTable& index_columns,
     PlatformConfig platform_config,
     std::optional<TimestampRange> timestamp) {
     if (!timestamp) {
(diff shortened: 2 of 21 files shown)


libtiledbsoma/src/soma/soma_dense_ndarray.cc
```diff
@@ -31,32 +31,38 @@ void SOMADenseNDArray::create(
...
     schema->flags = 0;
     schema->release = &ArrowAdapter::release_schema;
-    schema->children = new ArrowSchema*[schema->n_children];
+    schema->children = (ArrowSchema**)malloc(
+        schema->n_children * sizeof(ArrowSchema*));
 
     std::vector<std::string> index_column_names;
     for (uint64_t dim_idx = 0; dim_idx < index_column_size; ++dim_idx) {
-        ArrowSchema* dim = schema->children[dim_idx] = new ArrowSchema;
+        ArrowSchema* dim = schema->children[dim_idx] = (ArrowSchema*)malloc(
```
bkmartinjr, line 46: style nit: why not call `ArrowMalloc`, and `ArrowSchemaInit`?  (would allow removal of all the explicit field initialization, and prevent errors if we forget to init one of the fields.

libtiledbsoma/src/soma/soma_sparse_ndarray.cc
```diff
@@ -32,32 +32,38 @@ void SOMASparseNDArray::create(
     uint64_t index_column_size = index_column_schema->n_children;
 
     auto schema = std::make_unique<ArrowSchema>();
+    schema->name = nullptr;
```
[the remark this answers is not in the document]
johnkerl: #3860 has been created for this express purpose: to do this review in a planned, audited, thorough manner.

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -124,6 +124,16 @@ void ArrowAdapter::release_array(struct ArrowArray* array) {
         // If the ArrowBuffer.buffer_ shared_ptr is the last reference to the
         // underlying ColumnBuffer, the ColumnBuffer will be deleted.
         delete arrow_buffer;
+    } else {
```
bkmartinjr, line 127: the comment above about Arrow.buffer_ shared_ptr doesn't make sense in this context, as we don't have a shared_ptr in this function (and ArrowBuffer doesn't contain a shared_ptr).  Comment needs clarity

XanthosXanthopoulos: There is a `shared_ptr` in the `private_data` field if the array is holding query results. The `ColumnBuffer` is doing the bookkeeping.

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -124,6 +124,16 @@ void ArrowAdapter::release_array(struct ArrowArray* array) {
...
+        // ArrowArray that allocate buffers directly to ``array->buffers`` will
+        // leak memory. We detect these arrays when the ``array->private_data``
+        // is null
+        for (int64_t i = 0; i < array->n_buffers; ++i) {
```
bkmartinjr, line 131: this is probably a naive question, but why would we ever allocate directly into `buffers` without setting up the `private_data` structure?   (or just using `ArrowArrayInitFromType` to do it for us?)

libtiledbsoma/src/soma/soma_sparse_ndarray.cc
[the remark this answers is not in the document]
bkmartinjr: I would be far more sympathetic to that argument if a) it was a large task, and b) could not be performed incrementally.

I definitely encourage you to prioritize this work.

libtiledbsoma/src/utils/arrow_adapter.cc
bkmartinjr: ah!  TY.

libtiledbsoma/src/soma/soma_collection.cc
```diff
@@ -100,22 +100,16 @@ std::shared_ptr<SOMAExperiment> SOMACollection::add_new_experiment(
     std::string_view uri,
     URIType uri_type,
     std::shared_ptr<SOMAContext> ctx,
-    std::unique_ptr<ArrowSchema> schema,
-    ArrowTable index_columns,
+    const std::unique_ptr<ArrowSchema>& schema,
```
jp-dark, line 103: Similar comment to above. Is there a reason why you needed to change the way the `ArrowSchema` is passed in particular?

jp-dark (commented):
I'm still looking through this PR, but I wanted to flag passing the `unique_ptr` by const reference first as a discrete unit of concern.

libtiledbsoma/src/soma/managed_query.h
```diff
@@ -305,8 +305,8 @@ class ManagedQuery {
      * @param arrow_array
      */
     void set_array_data(
-        std::unique_ptr<ArrowSchema> arrow_schema,
-        std::unique_ptr<ArrowArray> arrow_array);
+        const std::unique_ptr<ArrowSchema>& arrow_schema,
+        const std::unique_ptr<ArrowArray>& arrow_array);
```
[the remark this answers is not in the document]
XanthosXanthopoulos: It indeed doesn't need ownership since it only iterates over the data. 
The owner of the data should call `release()` otherwise we risk to leak data and this function is not expected to change the data it is passed.
Passing a raw pointer is identical in behavior and I can do the change.

libtiledbsoma/src/soma/soma_collection.cc
XanthosXanthopoulos: Similar ownership <-> `release()` reason as in all other places where I have made the change

libtiledbsoma/src/soma/managed_query.h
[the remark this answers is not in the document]
jp-dark: That makes sense. Passing the raw pointer is better, because the reference just adds another unnecessary indirection.

jp-dark (approved):
I'd like to see the references to unique pointers replaced with raw pointers, but overall this is a big improvement. Thanks for going in and cleaning up the leaked data.

XanthosXanthopoulos:
Modifying the function that accept `const std::unique_ptr` requires project wide changes that would change in a follow up PR.
`set_array_data` was modified in this PR. [sc-65281](https://app.shortcut.com/tiledb-inc/story/65281/c-replace-const-std-unique-ptr-with-raw-pointer) was created to track the change from `const std::unique_ptr` to raw pointer.

PR #3877, opened 2025-03-31
[python] Return `datetime` domain/maxdomain as `pa.TimestampScalar`

nguyenv:
**Issue and/or context:**

[[sc-61328](https://app.shortcut.com/tiledb-inc/story/61328/python-domain-maxdomain-fail-to-return-arrow-types-for-datetime-columns)]

**Changes:**

The domain/maxdomain was return as `np.datetime64`. Give that Arrow is used as the default type for all return values, this is now corrected to `pa.TimestampScalar`.


apis/python/src/tiledbsoma/_util.py
```diff
@@ -546,24 +553,18 @@ def _set_coord(
 
     # Note: slice(None, None) matches the is_slice_of part, unless we also check
     # the dim-type part.
-    if is_slice_of(coord, np.datetime64) and pa.types.is_timestamp(dim.type):
+    if (
+        is_slice_of(coord, np.datetime64)
```
johnkerl, line 557: Might be simpler if `is_slice_of` could be modified to take a tuple? So here you could write

```
  if (is_slice_of(coord, (np.datetime64, pa.TimestampScalar, int)):
```

nguyenv: `is_slice_of` is implemented in somacore. I can update the code in somacore to also support tuples and then make a follow-up PR for tiledbsoma-py.

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -1103,29 +1103,8 @@ def _find_extent_for_domain(
...
-
-    if dtype == "datetime64[ns]":
-        ilo = int(lo.astype("int64"))
-        ihi = int(hi.astype("int64"))
-        iextent = min(extent, ihi - ilo + 1)
-        return np.datetime64(iextent, "ns")
+    if dtype in ("datetime64[s]", "datetime64[ms]", "datetime64[us]", "datetime64[ns]"):
+        return min(extent, _util.to_unix_ts(hi) - _util.to_unix_ts(lo) + 1)
 
     return extent
```
bkmartinjr, line 1109: this function should not be typed to return `Any`.  Looking at the code paths, it should be typed  `int | Literal['']`

I realize this pre-dates the PR, but we should be fixing this stuff (using `Any` effectively turns off static type checking)

apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -1103,29 +1103,8 @@ def _find_extent_for_domain(
...
-        ihi = int(hi.astype("int64"))
-        iextent = min(extent, ihi - ilo + 1)
-        return np.datetime64(iextent, "us")
-
-    if dtype == "datetime64[ns]":
-        ilo = int(lo.astype("int64"))
-        ihi = int(hi.astype("int64"))
-        iextent = min(extent, ihi - ilo + 1)
-        return np.datetime64(iextent, "ns")
+    if dtype in ("datetime64[s]", "datetime64[ms]", "datetime64[us]", "datetime64[ns]"):
```
bkmartinjr, line 1106: what happens if dtype is bytes or np.dtype(bytes)?  I would imagine it should return either `b''` or `''`, but it seems like it returns the tile extent.

johnkerl: I would think of that as `pa.binary` / `pa.large_binary`

johnkerl: But tile extent is not meaningful for binary/string types in core array schemas so whatever value we pass around will ultimately be ignored

bkmartinjr: Agree with the intent. I was just noting that this code path has a special case for `str`, but not other variable types.  For str it returns `''`, but all of the others (e.g., pa.binary), it returns the tile dimension.  Seemed inconsistent, or at least undocumented behavior

apis/python/src/tiledbsoma/_util.py
```diff
@@ -138,6 +138,17 @@ def validate_slice(slc: Slice[Any]) -> None:
     if slc.start is None or slc.stop is None:
         # All half-specified slices are valid.
         return
+
```
bkmartinjr, line 141: given the behavior of `to_unix_ts`, couldn't you combine these two checks into one check (merging lines 142-155) - something like:

```
if to_unix_ts(slc.stop) < to_unix_ts(slc.start):
    raise ValueError(...)
```

apis/python/src/tiledbsoma/_dataframe.py
johnkerl: Ah, makes sense!

apis/python/tests/ht/test_ht_dataframe.py
```diff
@@ -241,7 +240,6 @@ def dataframe_domain(
     NB:
```
bkmartinjr, line 240: Given that we now use Arrow timestamps as the "native" domain type, I suggest converting this class to do the same (i.e., `self.domain` should be a tuple of pa.TimestampScalar, not np.datetime64).  That would remove the need to do all the conversions like line 254 (`zero = ...`)

bkmartinjr (changes requested):
some cleanup suggestions. No major concerns.

apis/python/src/tiledbsoma/_dataframe.py
nguyenv: This ended up being complicated to do in this PR so filed this: https://app.shortcut.com/tiledb-inc/story/65378/typing-for-domain-should-be-tuple-int-float-literal-int-float-literal

apis/python/src/tiledbsoma/_util.py
nguyenv: No because the values can be float

bkmartinjr: If I read the code correctly, when they are float (or any type other than np.datetime64 or pa.TimestampScalar), `to_unix_ts` returns the value untouched (ie., the function is a noop).  So it ends up being exactly the same as the other if statement.

I certainly could be missing something - apologies if so.  But I don't think floats (or ints) are a problem here.

bkmartinjr (changes requested):
Couple of small items:
* we should fix the binary/bytes issue noted in _dataframe.py - am OK if you want to file it and treat it as a separate bug
* I'm pretty sure the if statements in _util.py are redundant (even for floats), but I leave it to your judgement/validation - I could be wrong.
No other concerns, and I really appreciate you making these fixes!

apis/python/src/tiledbsoma/_util.py
nguyenv: `to_unix_ts` takes as arguments valid user timestamp inputs `np.datetime64`,`pa.TimestampScalar`, and `int`. The return type is `int` since it should always be a UNIX timestamp. I'd rather keep it like that than change the signature to be `Any` argument type and `Any` return type.

apis/python/src/tiledbsoma/_dataframe.py
nguyenv: The code was actually correct but misleading. The dtype for strings is passed in as `"ascii"` and for binary as `"bytes"` (ie. not `pa.large_string`, `np.dtype("S")`, etc) which means the check `isinstance(dtype, str)` is `True` for both str and binary dimensions because the type of `dtype`s `"ascii"` and `"bytes"`  is infact `str`. I've updated the code to `dtype in ("ascii", "bytes")` instead to make this less confusing.

nguyenv: I also added a test to explicitly check the default extents.

bkmartinjr: oof, that is confusing!  I think it is still confusing, as the param is named `dtype`, which makes me assume it is an `np.dtype` and ot a string.  If it is `numpy.typing.DTypeLike`, we should start the dtype constructor, i.e., the standard pattern:
> dtype = np.dtype(dtype)

This lets the "official" numpy converter translate stuff like `"int64"` into an actual dtype object, and would more accurately detect all the various types that are equivalent to a np.dtype("S").

I'm also not sure where "ascii" comes from - AFAIK neither Pandas or Numpy ever use that.  Looking at the code, it appears to be somethign we made up (i.e., is unique to SOMA).

Suggested code, given all of that:

```
if dtype == 'ascii' or np.dtype(dtype).kind in ('S', 'U'):
    return ''
```

This will handle dtype being the various python tpyes (str, bytes, etc), all of the NumPy/Pandas dtypes, plus our own SOMA-specific "ascii".

bkmartinjr (approved):
One (optional) suggestion for improving type handling clarity in `_find_extent_for_domain`, but otherwise I'm good.

PR #3891, opened 2025-04-01
[python] Pin `build-system.requires`, ARM Linux Docker-builds to `cmake<4`

ryan-williams:
**Issue and/or context:**

CMake 4 (released [Friday, 3/27](https://pypi.org/project/cmake/4.0.0/)) broke our vcpkg-source build:

- #3890
- #3867
- #2482
- #2413

**Changes:**
- Pin `cmake<4` in `pyproject.toml` `build-system.requirements`
- Use a pip ["constraints" file](https://pip.pypa.io/en/stable/user_guide/#constraints-files) in Docker builds, to force setuptools to use `cmake<4`.

**Notes for Reviewer:**
- This makes [python-dockers.yml] work again, but the ARM Linux jobs still take 25-35mins building wheels (vs. 1-2mins on other platforms).
- It also remains a mystery why we're not getting vcpkg binaries, and why our vcpkg source build is getting an [old/broken cmrc][CMakeRC.cmake#L37].

[GitHub Actions screenshot showing job timings; ARM jobs take 25-35mins (vs. 1-2mins for x86 jobs)][GHA]
([GHA])

[GHA]: https://github.com/single-cell-data/TileDB-SOMA/actions/runs/14220200288
[python-dockers.yml]: https://github.com/single-cell-data/TileDB-SOMA/actions/workflows/python-dockers.yml
[CMakeRC.cmake#L37]: https://github.com/vector-of-bool/cmrc/blob/2.0.1/CMakeRC.cmake#L37

PR #3894, opened 2025-04-02
[ci/python] `python-dockers.yml`: improve issue template

ryan-williams:
**Issue and/or context:**
`python-docker.yml` began failing Friday (#3867), then continued failing for several days, generating a new issue each day (#3870, #3871, #3874, #3880, #3892).

**Changes:**
- `update_existing: true`: update any existing open issue with the same title as the one about to be filed
  - Requires removing the date from the title, but I think that makes sense; #3867 turned into a rolling/tracking issue about a build break that lasted multiple days.
- Move assignees from issue-template to `JasonEtco/create-an-issue@v2` invocation (seems a bit nicer)
- Fix an unrelated broken wiki link

**Notes for Reviewer:**
I haven't tested this, seems likely correct / we can troubleshoot if/when the next failure occurs. Alternative is I can manually-dispatch it to create a test issue, and verify that subsequent runs update the existing issue (rather than create a new issue each time).

I've `[skip ci]`'d because Docker-build GHA is failing (until #3891 lands), and it wouldn't file an issue anyway (since it's not a `scheduled` run).

### DIFF
--- a/.github/workflows/daily-python-dockers-issue-template.md
+++ b/.github/workflows/daily-python-dockers-issue-template.md
@@ -1,0 +1,7 @@
+---
+title: Daily {{ tools.context.action }} run failing
+labels: bug
+---
+
+See run for more details:
+https://github.com/{{ env.GITHUB_REPOSITORY }}/actions/runs/{{ env.GITHUB_RUN_ID }}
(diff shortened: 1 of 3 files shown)


johnkerl (approved):
🚢

PR #3906, opened 2025-04-03
[python] More docstring content for `extend_enumeration_values` [WIP]

johnkerl:
**Issue and/or context:**  For issue #3785 / [[sc-63930]](https://app.shortcut.com/tiledb-inc/story/63930/dataframe-need-api-to-extend-column-enum-values-without-doing-a-write#activity-64471). Follow-on to #3815.

**Changes:** Add some more docstring content re why/when users would want to call this.

**Notes for Reviewer:**

I'm still deciding what to say here. This method isn't really a top-level/user-facing method. Rather, #3865 will be. And `extend_enumeration_values` will be called by `tiledbsoma.io` on behalf of the user.

Some options for the docstring:

* Leave it as-is without this PR.
* As in this PR as of the first commit.
* Add a sentence saying it isn't really intended to be called directly, and point the user at the mods on #3865.
* Add a sentence saying it isn't _usually_ intended to be called directly, but add something like the contents of this PR as of the first commit.
* Add a link to Academy documentation, which I'm working on now.

I'm leaving this in draft for now, but I'm leaning toward the last bullet.

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -406,6 +406,16 @@ class DataFrame(SOMAArray, somacore.DataFrame):
     ) -> None:
         """Extend enumeration values for each column defined in `values`.
 
+        You can do ``write`` with columns of Arrow dictionary type, without
+        calling this method, and that works fine for non-parallel data ingestion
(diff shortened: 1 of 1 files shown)


johnkerl:
I will follow up with Academy documentation.

PR #3915, opened 2025-04-05
[python] Fix partition readers' naïve assumption that `nnz` is fast

bkmartinjr:
**Issue and/or context:**

`tiledbsoma.io.to_anndata` and `ExperimentAxisQuery.to_anndata` contain a partitioned reader optimization. This code uses `nnz` to estimate partition sizes, and in doing so naively assumed `nnz` would always be fast (ie., would only use fragment metadata).  However, there are cases where this is not true, and `nnz` performs the very slow operation of reading and counting all data.

In this "slow" case, the end result is that the `to_anndata` methods will read the entire X array once to count `nnz`, and then read it again to construct the AnnData.

**Changes:**

1. Introduced a private, internal-only API to have `nnz` throw an exception if it is unable to use the "fast" path - ie.., compute nnz from fragment metadata.
2. Used this new hook in the partitioned readers to utilize nnz information when it is fast, and fall back on a non-partitioned read when nnz is unavailable.
3. Revised the fallback, when nnz is unavailable, to use a partition size that is reasonable for most ssRNAseq data (previously it fell back on no partitioning, which is not optimal).

**Notes for Reviewer:**

[[sc-65406]](https://app.shortcut.com/tiledb-inc/story/65406)
[[sc-65409]](https://app.shortcut.com/tiledb-inc/story/65409)

### DIFF
--- a/apis/python/requirements_spatial.txt
+++ b/apis/python/requirements_spatial.txt
@@ -2,5 +2,6 @@ geopandas
 tifffile
 pillow
 spatialdata>=0.2.5
+numcodecs<0.16
 xarray
 dask<=2024.11.2
--- a/apis/python/src/tiledbsoma/_query.py
+++ b/apis/python/src/tiledbsoma/_query.py
@@ -856,7 +856,7 @@ def _read_as_csr(
 
     approx_X_shape = tuple(b - a + 1 for a, b in matrix.non_empty_domain())
     # heuristically derived number (benchmarking). Thesis is that this is roughly 80% of a 1 GiB io buffer,
-    # which is the default for SOMA. If we have NNZ, use to partition based upon memory size. If we do not
+    # which is the default for SOMA. If we have fast NNZ, use to partition based upon memory size. If we do not
     # have fast NNZ, pick a default partition size which is large, but not so large as to rule out reasonable
     # parallelism (in this case, calculated based on typical scRNASeq assay density of 3-6K features).
     target_point_count = 96 * 1024**2
--- a/apis/python/src/tiledbsoma/_query.py
+++ b/apis/python/src/tiledbsoma/_query.py
@@ -857,7 +857,7 @@ def _read_as_csr(
     approx_X_shape = tuple(b - a + 1 for a, b in matrix.non_empty_domain())
     # heuristically derived number (benchmarking). Thesis is that this is roughly 80% of a 1 GiB io buffer,
     # which is the default for SOMA. If we have NNZ, use to partition based upon memory size. If we do not
-    # have NNZ, pick a default partition size which is large, but not to large as to rule out reasonable
+    # have fast NNZ, pick a default partition size which is large, but not so large as to rule out reasonable
     # parallelism (in this case, calculated based on typical scRNASeq assay density of 3-6K features).
     target_point_count = 96 * 1024**2
     fallback_row_count = 32768
--- a/apis/python/src/tiledbsoma/_query.py
+++ b/apis/python/src/tiledbsoma/_query.py
@@ -856,13 +856,16 @@ def _read_as_csr(
 
     approx_X_shape = tuple(b - a + 1 for a, b in matrix.non_empty_domain())
     # heuristically derived number (benchmarking). Thesis is that this is roughly 80% of a 1 GiB io buffer,
-    # which is the default for SOMA.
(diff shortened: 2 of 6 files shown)


apis/python/src/tiledbsoma/_query.py
```diff
@@ -852,13 +856,16 @@ def _reindex(batch: pa.RecordBatch) -> pa.RecordBatch:
 
     approx_X_shape = tuple(b - a + 1 for a, b in matrix.non_empty_domain())
     # heuristically derived number (benchmarking). Thesis is that this is roughly 80% of a 1 GiB io buffer,
-    # which is the default for SOMA.
+    # which is the default for SOMA. If we have NNZ, use to partition based upon memory size. If we do not
+    # have NNZ, pick a default partition size which is large, but not to large as to rule out reasonable
```
johnkerl, line 860:
```suggestion
    # have fast NNZ, pick a default partition size which is large, but not so large as to rule out reasonable
```

apis/python/requirements_spatial.txt
```diff
@@ -2,5 +2,6 @@ geopandas
 tifffile
 pillow
 spatialdata>=0.2.5
+numcodecs<0.16
```
johnkerl, line 5: FY #3917 #3916 -- this will go away on rebase/merge of this PR to `main`

apis/python/src/tiledbsoma/_query.py
```diff
@@ -812,7 +813,10 @@ def _read_as_csr(
 
     d0_joinids = d0_joinids_arr.to_numpy()
     d1_joinids = d1_joinids_arr.to_numpy()
-    nnz = matrix.nnz
+    try:
+        nnz = matrix._handle._handle.nnz(raise_if_slow=True)
+    except SOMAError:
+        nnz = -1
```
johnkerl, line 819: `None` would be more Pythonic and less likely to get accidentally used in arithmetic by some future maintainer of this code

apis/python/src/tiledbsoma/io/outgest.py
```diff
@@ -106,11 +106,14 @@ def _read_partitioned_sparse(X: SparseNDArray, d0_size: int) -> pa.Table:
     # density of matrix. Magic number determined empirically, as a tradeoff
     # between concurrency and fixed query overhead.
     tgt_point_count = 96 * 1024**2
-    nnz = X.nnz
+    try:
+        nnz = X._handle._handle.nnz(raise_if_slow=True)
+    except SOMAError:
+        nnz = -1
```
johnkerl, line 112: As above

johnkerl (changes requested):

apis/python/src/tiledbsoma/_query.py
bkmartinjr: I was doing this to also handle the case where nnz == 0.  I'm not sure None is any more pythonic - could you cite a source that applies to this case?

bkmartinjr: to clarify, we need to handle:
* nnz == 0, empty array
* nnz > 0
* no nnz avaialble
making it polymorphic means that tests below go from `nnz > 0` to `nnz is not None and nnz > 0`, which seems cumbersome

johnkerl: OK

bkmartinjr: OK, will do.  I register my disagreement, but I think getting this landed has priority over that.

johnkerl (approved):

apis/python/src/tiledbsoma/_query.py
```diff
@@ -852,13 +856,16 @@ def _reindex(batch: pa.RecordBatch) -> pa.RecordBatch:
 
     approx_X_shape = tuple(b - a + 1 for a, b in matrix.non_empty_domain())
     # heuristically derived number (benchmarking). Thesis is that this is roughly 80% of a 1 GiB io buffer,
-    # which is the default for SOMA.
+    # which is the default for SOMA. If we have NNZ, use to partition based upon memory size. If we do not
```
johnkerl, line 859:
```suggestion
    # which is the default for SOMA. If we have fast NNZ, use to partition based upon memory size. If we do not
```

Sorry I didn't catch both spots earlier! :|

johnkerl (approved):

XanthosXanthopoulos (approved):

PR #3919, opened 2025-04-07
[python] Protect against an AnnData/SciPy incompatibility

johnkerl:
**Issue and/or context:** [[sc-65041]](https://app.shortcut.com/tiledb-inc/story/65041) #3920

**Changes:** As in #3920.

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/io/__init__.py
+++ b/apis/python/src/tiledbsoma/io/__init__.py
@@ -2,6 +2,10 @@
 #
 # Licensed under the MIT License.
 
+import anndata
+import scipy
+from packaging.version import Version
+
 from ._registration import (
     ExperimentAmbientLabelMapping,
 )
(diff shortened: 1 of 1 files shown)


bkmartinjr (commented):
General question: when should we just document these types of upstream incompatibilities, vs. coding them into our package?

I would be just as comfortable with a well documented Issue in GH, closed as "won't fix", but available for people to search & discover.  I'm a bit uncomfortable setting the precedent that we are going to try and perform this function in our code base....seems not scalable.

johnkerl:
@bkmartinjr my hope was to action the bug report at [sc-65041].

You and I are both quite senior on this project, and together we've spent a few hours to get here. All that time could have been reduced to seconds had we had the error message I propose here.

Scalability becomes a concern if we do this many times. But this is not an often-repeating pattern. It is a one-off very sharp edge, involving our dependencies, which at first blush -- to you and me both! -- looked like a tiledbsoma defect.

Let's proactively help our users get to a quicker solution for this particularly sharp edge, which is not a particularly recurring pattern.

bkmartinjr (approved):

Issue #3944, opened 2025-04-09
Docker image build failure for TileDB-SOMA on MacOS with `uv`


jmzhang1911:
Hello Developer,

I'm building a container on macOS (version 15.3.2) with an M1 chip using the following method. However, I encountered the following error during the build.

May I ask if TileDB-SOMA does not support this installation method?
Any guidance would be greatly appreciated. Thank you!

```
FROM python:3.13

RUN pip install --upgrade pip \
  && pip install uv \
  && uv init example

RUN cd example && uv add tiledbsoma
```
```
98.66       [ 20%] Performing configure step for 'ep_tiledb'
98.66       Checking:
98.66       /root/.cache/uv/sdists-v9/pypi/tiledbsoma/1.16.1/aM2XmTE5zFT6ahhcXbgyu/src/dist_links/dist/lib/libtiledbsoma.so
98.66       exists: False
98.66       Checking:
98.66       /root/.cache/uv/sdists-v9/pypi/tiledbsoma/1.16.1/aM2XmTE5zFT6ahhcXbgyu/src/dist_links/dist/lib64/libtiledbsoma.so
98.66       exists: False
98.66       Checking:
98.66       /root/.cache/uv/sdists-v9/pypi/tiledbsoma/1.16.1/aM2XmTE5zFT6ahhcXbgyu/src/dist_links/dist/lib/x86_64-linux-gnu/libtiledbsoma.so
98.66       exists: False
98.66       libtiledbsoma.so: cannot open shared object file: No such file or
98.66       directory
98.66
98.66       [stderr]
98.66       /root/.cache/uv/builds-v0/.tmp4HRpOv/lib/python3.13/site-packages/setuptools/dist.py:759:
98.66       SetuptoolsDeprecationWarning: License classifiers are deprecated.
98.66       !!
98.66
98.66
98.66       ********************************************************************************
98.66               Please consider removing the following classifiers in favor of a
98.66       SPDX license expression:
98.66
98.66               License :: OSI Approved :: MIT License
98.66
98.66               See
98.66       https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license
98.66       for details.
98.66
98.66       ********************************************************************************
98.66
98.66       !!
98.66         self._finalize_license_expression()
98.66       CMake Error at
[...]

PR #3909, opened 2025-04-03
Publish ARM Linux wheels

ryan-williams:
## Issue and/or context
- #2482
- #3867
- #3890

`pip install tiledbsoma` can take [over 40mins]() on ARM Linux, building TileDB-SOMA, TileDB, and vcpkg from source. The former two take most of the time, but the latter broke us from Friday to Weds (#3867), and we've still not root-caused some of what went wrong (#3890).

`pip install` takes <2mins on other platforms, by installing wheels; publishing ARM Linux wheels brings that same performance to e.g. Graviton EC2 instances, and Linux Docker builds on ARM Macs.

## Changes
- Add `manylinux_aarch64` as a 4th set of os/arch wheels (so we now publish {linux,mac} x {x86,arm}).
- Audit/GC CIBW configs:
  - Pin `CIBW_MANYLINUX_{X86_,AARCH}64_IMAGE=quay.io/pypa/manylinux_2_28_*:2025.03.22-2` ([options](https://quay.io/repository/pypa/manylinux_2_28_aarch64?tab=tags))
  - Remove unused `CIBW_BUILD` var
  - Remove unnecessary GCC-version munging
  - Remove unnecessary `CIBW_TEST_SKIP: "*_arm64"`
  - Remove a redundant `bash -x -c 'rm -rf tiledbsoma*/dist_links/dist/lib*'`
- Restore `smoke-test` job for Python 3.9 (related: #2640)
- Fix Test PyPI publish silent failure (e.g. when hitting quota; [example](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/13913129586/job/38931629454))
- Remove unused `matrix.{cc,cxx}` entries

## Notes for Reviewer
- [Latest GHA run](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/14251626772) (in progress)
- Previous test.pypi.org releases and GHA runs:
  - [1.15.0rc0.post386.dev1013776864] ([GHA](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/14245910996), 2h20m)
  - [1.15.0rc0.post387.dev4100197019] ([GHA](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/14248317927), 1h31m)
  - [1.15.0rc0.post388.dev2177748919] ([GHA](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/14250576038), 32m)
  The timing variance is mostly due to jobs waiting in "pending" state, afaict.

### Release-size math
[...]

Issue #3890, opened 2025-04-01
[Bug] `pip install tiledbsoma` fails on ARM Linux with CMake 4


ryan-williams:
### Describe the bug
See https://github.com/single-cell-data/TileDB-SOMA/issues/3867#issuecomment-2769695692:

- On ARM Linux, `pip install tiledbsoma` performs a source build (because we don't publish ARM Linux wheels: [#2482] / [#2413]).
- TileDB-SOMA source build initiates TileDB source build, which fails to find vcpkg binaries for ARM Linux.
- That leads to a vcpkg source build, which for some reason builds cmrc [2.0.1] (Sept '18) instead of [`952ffdd`] (pinned in [portfile.cmake] and [vcpkg.json], which contains a corrected [CMakeRC.cmake#L37]). 

### To Reproduce
See [failed GHA](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/14126265619) from #3867.

<details><summary>Example Dockerfile</summary>

```Dockerfile
# Debian 12 image with GCC 13
ARG gcc=13
FROM gcc:$gcc-bookworm

ENV DEBIAN_FRONTEND=noninteractive VCPKG_FORCE_SYSTEM_BINARIES=1
RUN apt update -y \
 && apt install -y cmake curl g++ git make ninja-build pkg-config tar unzip zip \
 && apt install -y python3 python-is-python3 python3-pip python3-venv \
 && python -m venv .venv
ENV PATH=/.venv/bin:$PATH

# tiledbsoma>=1.16 requires GCC 13
ARG v=''
RUN if [ -z $v ]; then pip install tiledbsoma; else pip install tiledbsoma==$v; fi
ENTRYPOINT [ "python", "-c", "import tiledbsoma; tiledbsoma.show_package_versions()" ]
```
</details> 

### Versions
- TileDB-SOMA version: 1.16.1
- Language and language version: Python 3.12
- OS (e.g. MacOS, Ubuntu Linux): ARM Linux

### Workarounds
- Pinning `cmake<4` in `pyproject.toml` `build-system.requires` will help, but requires a release to unbreak `pip install tiledbsoma` on ARM Linux. I will file a PR for that.
- A pip ["constraints" file](https://pip.pypa.io/en/stable/user_guide/#constraints-files) can be used to force setuptools to use `cmake<4`, in the meantime. I will also file a PR for that, to unbreak [our Docker-build GHA][python-dockers.yml].

### Longer-term fixes
[...]

PR #3948, opened 2025-04-09
[python] Add `python-uv.dockerfile`

ryan-williams:
**Issue and/or context:**
- #3944
- #3890
- #3909

**Changes:**
- Add `python-uv.dockerfile`, verifying using `uv` to `pip install tiledbsoma` in `python:<version>` Docker images (Debian 12 / bookworm-based, but need GCC 13 installed for TileDB-SOMA ≥1.16).
- Add `python:3.{12,13}` x {x86,ARM} builds to `python-dockers.yml` [GHA](https://github.com/single-cell-data/TileDB-SOMA/actions/workflows/python-dockers.yml)

### DIFF
--- a/.github/workflows/python-dockers.yml
+++ b/.github/workflows/python-dockers.yml
@@ -42,6 +42,12 @@ jobs:
             args: gcc=13 v=1.16.0
           - name: Langchain
             file: langchain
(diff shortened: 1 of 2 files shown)


johnkerl (approved):
Gorgeous work! :)

PR #3916, opened 2025-04-07
`setup.py`: remove deprecated `License ::` from `classifiers`

ryan-williams:
**Issue and/or context:**

`pip install tiledbsoma` was issuing a warning like:
```
          ********************************************************************************
          Please consider removing the following classifiers in favor of a SPDX license expression:

          License :: OSI Approved :: MIT License

          See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
          ********************************************************************************
```

([GHA example](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/14276171882/job/40019409594#step:9:52); compare to lack of warning in [GHA for this PR](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/14312394947/job/40110373749#step:9:51))

**Changes:**
Our MIT license is already specified in the `license` field, can be removed from `classifiers`.

Also fixes #3917 by pinning `numcodecs<0.16`.

[sc-65131]

PR #3953, opened 2025-04-10
[python] Update setuptools

teo-tsirpanis:
**Issue and/or context:**

I noticed in build logs that we are getting a warning that `License classifiers are deprecated` (see #3944 for example).

**Changes:**

After some search, I saw that updating `setuptools` will likely fix the warnings (see bokeh/bokeh#14434).

**Notes for Reviewer:**

### DIFF
--- a/apis/python/pyproject.toml
+++ b/apis/python/pyproject.toml
@@ -1,7 +1,7 @@
 [build-system]
 requires = [
     "pybind11[global]>=2.10.0",
-    "setuptools>=70.1",  # `setuptools.command.bdist_wheel`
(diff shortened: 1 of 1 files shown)


teo-tsirpanis:
Remote storage CI fails because the PR is from my fork and has no access to secrets (I don't have write access to this repository).

ryan-williams:
I believe this warning was worked around in a different way by #3916.

e.g. I don't see hits for "license" in [this](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/14661785626/job/41160425449) Ubuntu ARM source install, from this morning.

PR #3956, opened 2025-04-10
[python] Remove `typing.{Dict,List,Optional,Set,Tuple,Type,Union}` references

ryan-williams:
**Issue and/or context:**
- {`Dict`,`List`,`Set`,`Tuple`,`Type`} [are deprecated since Python 3.9](https://docs.python.org/3/library/typing.html), to be replaced with lowercase counterparts (which don't require dedicated `import`s).
- `Union` can be replaced with the pipe `|` operator [since Python 3.10](https://docs.python.org/3/library/typing.html#typing.Union), but in most cases can be replaced in Python 3.9 as long as `from __future__ import annotations` is present. Explicit `Union` is still needed in top-level type-alias declarations, and `cast` and `st.from_type` functions.
- Similarly, `Optional[T]` can usually be replaced with `T | None`, except in e.g. `cast` calls, so I've done more of that here (was mostly done in #3453).

**Changes:**
- Remove all references to {`Dict`,`List`,`Set`,`Tuple`,`Type`}
- Remove most references to {`Optional`,`Union`}

### DIFF
--- a/apis/python/remote_tests/util.py
+++ b/apis/python/remote_tests/util.py
@@ -2,7 +2,6 @@ import datetime
 import os
 import pathlib
 import shutil
-from typing import Tuple
 
 import tiledb.cloud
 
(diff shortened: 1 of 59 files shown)


jp-dark (approved):

johnkerl:
I'll backport this to `release-1.16` -- not because it's a functional change (it isn't) but because it touches a large number of source files -- anything _else_ higher-pri we need to backport from `main` to `release-1.16` _after_ this PR is merged, would likely cause conflict-merge fails

PR #3957, opened 2025-04-10
[python] Ensure schema is explicitly passed to `pa.Table.from*` converters

nguyenv:
**Issue and/or context:**

[[sc-61222](https://app.shortcut.com/tiledb-inc/story/61222/python-somasparsearray-read-returns-arrow-table-not-matching-array-schema-nullability#activity-65401)]

This PR fixes several problems when we do not explicitly pass a schema to `pa.Table.from_{arrays,pydict,pandas}` converters:

1. In the Pybind11 [utility function `_buffer_to_table`](https://github.com/single-cell-data/TileDB-SOMA/compare/viviannguyen/sc-61222/python-somasparsearray-read-returns-arrow?expand=1#diff-d123499d8c03fff7d820bb52ab295e45080a15c7a815123ae43cdd2d8bdf0674R218-R235) which takes an `ArrayBuffer` and converts it to a Python `pa.Table` using `pa.Table.from_arrays`, we were taking in a list of names which meant it was inferring the schema from the data in the arrays. We now generate a list of `pa.Field` objects for each column and generate the `pa.Schema` to explicitly pass in instead of names. This fixes the issue reported in the shortcut story linked.
2. Several unit tests utilize `pa.Table.from_pydict` and `pa.Table.from_pandas` and took in only a `dict` or `pd.DataFrame`. Without explicitly passing in a `pa.Schema` argument, columns without `NULL` values would be inferred to be non-nullable. 

**Would strongly recommend to always pass the `schema` for all Arrow converter methods to preserve nullability.**

### DIFF
--- a/apis/python/tests/test_dataframe.py
+++ b/apis/python/tests/test_dataframe.py
@@ -821,11 +821,11 @@ def simple_data_frame(tmp_path):
     """
     schema = pa.schema(
         [
-            ("index", pa.int64()),
-            ("soma_joinid", pa.int64()),
-            ("A", pa.int64()),
-            ("B", pa.float64()),
-            ("C", pa.large_string()),
+            pa.field("index", pa.int64(), False),
+            pa.field("soma_joinid", pa.int64()),
+            pa.field("A", pa.int64()),
+            pa.field("B", pa.float64()),
+            pa.field("C", pa.large_string()),
         ]
     )
     index_column_names = ["index"]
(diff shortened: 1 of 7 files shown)


apis/python/tests/test_dataframe.py
```diff
@@ -887,50 +892,35 @@ def _check_tbl(tbl, col_names, ids, *, demote):
...
-        sdf.read(ids, column_names=col_names).concat(),
-        col_names,
-        ids,
-        demote=False,
-    )
-    _check_tbl(sdf.read(column_names=col_names).concat(), col_names, None, demote=False)
+    _check_tbl(sdf.read(ids, column_names=col_names).concat(), col_names, ids)
+    _check_tbl(sdf.read(column_names=col_names).concat(), col_names, None)
+
+    # pa.Table.from_pandas infers nullability from the data if a schema is not
```
johnkerl, line 901: 🔥

johnkerl (approved):
This looks solid and well-explained to me -- thanks @nguyenv !

PR #3963, opened 2025-04-11
[python] Ensure `ndarray` byteorder is little-endian on matrix writes

nguyenv:
**Issue and/or context:**

[[sc-63459](https://app.shortcut.com/tiledb-inc/story/63459/python-tiledbsoma-io-from-anndata-ignores-ndarray-byteorder-writes-incorrect-data-if-array-is-not-host-byteorder)]

**Changes:**

In `_write_matrix_to_{dense,sparse}NDArray`, ensure that the byteorder is little-endian.


bkmartinjr (approved):
thank you!

PR #3970, opened 2025-04-13
[python] Fix mishandling of join-ID column

bkmartinjr:
**Issue and/or context:**

The recent changes to ingest registration failed to correctly handle the case where the user requested an obs or var join on something other than the obs/var index.

The docstring for `from_h5ad` clearly states:
```
        obs_id_name/var_id_name: Which AnnData ``obs`` and ``var`` columns, respectively, to use
          for append mode.

          Values of this column will be used to decide which obs/var rows in appended
          inputs are distinct from the ones already stored, for the assignment of ``soma_joinid``.  If
          this column exists in the input data, as a named index or a non-index column name, it will
          be used. If this column doesn't exist in the input data, and if the index is nameless or
          named ``index``, that index will be given this name when written to the SOMA experiment's
          ``obs`` / ``var``.
```

**Changes:**
* added logic to use the correct column for AnnData join
* added a test to verify the column supersedes the dataframe index

**Notes for Reviewer:**

This bug was found in 1.16.2rc0 by visual inspection of the code

### DIFF
--- a/apis/python/src/tiledbsoma/io/_registration/ambient_label_mappings.py
+++ b/apis/python/src/tiledbsoma/io/_registration/ambient_label_mappings.py
@@ -712,7 +712,7 @@ class AnnDataAxisMetadata:
 
 
 def _get_dataframe_joinid_index(df: pd.DataFrame, field_name: str) -> pd.Index:  # type: ignore[type-arg]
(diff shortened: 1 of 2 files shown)


johnkerl (approved):
🚢

PR #3983, opened 2025-04-15
[c++,python] Add `ManagedQuery.submit_batch` for global-order write

nguyenv:
**Issue and/or context:**

Original PR: https://github.com/single-cell-data/TileDB-SOMA/pull/3764
Reverted here: https://github.com/single-cell-data/TileDB-SOMA/pull/3886 due to segfault (https://github.com/single-cell-data/TileDB-SOMA/issues/3879)
But now re-opening with modifications

**Changes:**

On top of the changes already implemented in https://github.com/single-cell-data/TileDB-SOMA/pull/3764, this code introduces `submit_batch` which combines the `ManagedQuery::set_array_data` and `ManagedQuery::submit_write` calls.


apis/python/tests/test_dataframe.py
```diff
@@ -3535,3 +3535,58 @@ def test_no_extent_warning_61509(tmp_path):
             index_column_names=("float_index",),
             domain=((-fmax, fmax),),
         ).close()
+
+
+def test_fragments_in_writes(tmp_path):
```
bkmartinjr, line 3540: Need additional tests. Stuff I can think of:
1. tests that probe the expected error cases the ManagedQuery should catch/check?  OTOMH:
    * batches have out-of-order indices
    * out-of-order submits
    * submitted batches with incompatible schema
    * anything else you can think of
 2. tests on arrays with more dimensions, and dimensions of other types (both the common cases, e.g., spatial dataframes, and user-specified cases such as a soma_joinid+string dimensions)
 3. edge cases, such as empty batches/chunks
 4. tests on sparse & dense array

bkmartinjr (commented):
Do we need C++ tests for the ManagedQuery class changes?

apis/python/src/tiledbsoma/_soma_array.py
```diff
@@ -158,3 +159,47 @@ def _maxdomain(self) -> tuple[tuple[Any, Any], ...]:
...
+            sort_coords:
+                Whether the coordinates need to be sorted (True) or are already
+                sorted in global order (False). In the PlatformConfig, this is
+                is to True by default.
+        """
+        batches = values.to_batches()
+        if not batches:
+            return
+
+        layout = clib.ResultOrder.unordered if sort_coords else clib.ResultOrder.globalorder
```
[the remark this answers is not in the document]
bkmartinjr: AFAIK, `platform_config` is also undocumented.

@aaronwolen - do we need a plan to document this, or are you OK if it lands w/o user-facing help/examples?

apis/python/tests/test_dataframe.py
bkmartinjr: thanks for adding additional tests.  The one important case missing (IMHO) is a DataFrame with user-specified index columns. PointCloud covers some of this (multiple numeric indices), but not the case where the user specifies a string index.

Something like a `DataFrame` with index_column_names containing a string and float column in addition to the normal soma_joinid

I'm just trying to cover the edge cases of variable length or non-integral dimensions, e.g.,
```
df = pd.DataFrame(
    {
        "soma_joinid": pd.Series([4], dtype=np.int64),
        "str_idx": pd.Series(['a'],dtype=str),
        "float_idx": pd.Series([1.1],dtype=np.float32),
        "attr": pd.Series(["hi"], dtype="str"),
    }
)
soma.DataFrame.create(uri, schema=pa.Schema.from_pandas(df), index_column_names=['soma_joinid','str_idx','float_idx'])
```

which creates:
```
ArraySchema(
  domain=Domain(*[
    Dim(name='str_col', domain=('', ''), tile=None, dtype='|S0', var=True, filters=FilterList([ZstdFilter(level=3), ])),
    Dim(name='float_idx', domain=(-3.4028235e+38, 3.4028235e+38), tile=1.0, dtype='float32', filters=FilterList([ZstdFilter(level=3), ])),
    Dim(name='soma_joinid', domain=(0, 9223372036854775805), tile=1, dtype='int64', filters=FilterList([ZstdFilter(level=3), ])),
  ]),
  attrs=[
    Attr(name='attr', dtype='<U0', var=True, nullable=True, enum_label=None, filters=FilterList([ZstdFilter(level=-1), ])),
  ],
  cell_order='row-major',
  tile_order='row-major',
  capacity=100000,
  sparse=True,
  allows_duplicates=False,
)
```

bkmartinjr (changes requested):
Looking good!  I would like to see one additional test (DataFrame with both joinid and a user-specified str index) to make sure we cover the string index corner case.

Also a question (maybe for Aaron?) about docs on this feature

apis/python/HISTORY.md
```diff
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Changed
 
+- \[[#3983](https://github.com/single-cell-data/TileDB-SOMA/pull/3983)\] [python] Multiple writes of pre-sorted data may now be written to a single fragment using TileDB global order writes. Enable this performance optimization by setting the platform_config parameter sort_coords to True in the call to write. Will raise an error if data is not written in global sort order.
```
aaronwolen, line 13: Shouldn't this say

> by setting the platform_config parameter sort_coords to _False_

or am I misunderstanding something?

bkmartinjr (changes requested):
On the tests -- none of the tests confirm that only a single fragment was created.  Given that a primary goal of GOW is minimizing fragments, it would be a good idea if the tests actually confirm it.

Can we add something simple to do so?  (e.g., count elements in the fragments directory before & after)?

bkmartinjr (approved):
Recommend checking fragment count in more tests, but otherwise it LGTM!

PR #3992, opened 2025-04-17
[python] Improve null-index handling in Arrow dictionaries [RFC]

johnkerl:
**Issue and/or context:** Issue to be typed up.

**Changes:**

**Notes for Reviewer:**

No bugfixes yet. At this point I want buy-in from everyone on what we expect the values to be.

### DIFF
--- a/apis/python/tests/test_dataframe.py
+++ b/apis/python/tests/test_dataframe.py
@@ -971,6 +971,122 @@ def test_extend_enumeration_values_offsets(tmp_path, ordered):
         assert actual == expect
 
 
+@pytest.mark.parametrize(
+    "config",
+    [
+        {
+            "inidx1": [0, 1, 2, 3],
+            "inval1": ["red", "yellow", "green", "blue"],
+            "inidx2": [0, 1, 2, 3],
+            "inval2": ["red", "yellow", "green", "blue"],
+            "expidx": [0, 1, 2, 3, 0, 1, 2, 3],
+            "expval": ["red", "yellow", "green", "blue"],
+        },
+        {
+            "inidx1": [0, 1, 2, 3],
+            "inval1": ["red", "yellow", "green", "blue"],
+            "inidx2": [3, 2, 1, 0],
+            "inval2": ["red", "yellow", "green", "blue"],
+            "expidx": [0, 1, 2, 3, 3, 2, 1, 0],
+            "expval": ["red", "yellow", "green", "blue"],
+        },
+        {
+            "inidx1": [0, 1, 2, 3],
+            "inval1": ["red", "yellow", "green", "blue"],
+            "inidx2": [0, 1, 2, 3],
+            "inval2": ["orange", "grey", "green", "blue"],
+            "expidx": [0, 1, 2, 3, 4, 5, 2, 3],
+            "expval": ["red", "yellow", "green", "blue", "orange", "grey"],
+        },
+        {
+            "inidx1": [0, 1, 2, 3],
+            "inval1": ["red", "yellow", "green", "blue"],
+            "inidx2": [0, None, 2, 3],
+            "inval2": ["orange", "grey", "green", "blue"],
+            "expidx": [0, 1, 2, 3, 4, None, 2, 3],
+            "expval": ["red", "yellow", "green", "blue", "orange"],
+        },
+        {
+            "inidx1": [0, 1, 2, 3],
+            "inval1": ["red", "yellow", "green", "blue"],
+            "inidx2": [0, None, 0, 0],
+            "inval2": ["orange"],
+            "expidx": [0, 1, 2, 3, 4, None, 4, 4],
+            "expval": ["red", "yellow", "green", "blue", "orange"],
+        },
+        {
+            "inidx1": [0, 1, 2, 3],
+            "inval1": ["red", "yellow", "green", "blue"],
+            "inidx2": [None, None, None, None],
+            "inval2": ["orange"],
+            "expidx": [0, 1, 2, 3, None, None, None, None],
+            "expval": ["red", "yellow", "green", "blue"],
+        },
+        # This segfaults:
+        # {
+        #     "inidx1": [0, 1, 2, 3],
+        #     "inval1": ["red", "yellow", "green", "blue"],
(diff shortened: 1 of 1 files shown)


libtiledbsoma/src/soma/managed_query.h
```diff
@@ -805,6 +811,139 @@ class ManagedQuery {
...
+
+        if (index_array == nullptr) {
+            // All enum values are considered covered: this is the context
+            // of our having been given a plain Arrow array of enum values
+            // and being asked to extend the array's enumeration to include
+            // them. We should honor all the provided values.
+            return std::nullopt;
+        }
+
+        std::optional<std::vector<uint8_t>>
```
bkmartinjr, line 901: this code converts the validity (bit) vector into a uint8 vector, which can use consume quite a bit of memory across all dict columns in a dataframe. I see helpers in nanoarrow (e..g., `ArrowArrayViewIsNull`) that would make a scan of the bit vector very straight forward.

it seems like there is a time/space trade here, as the index-by-index bit unpacking may be a slower than the optimized use of `ArrowBitsUnpackInt8`.  I'm not sure which is best - probably need to benchmark both to know for sure.

@aaronwolen @jp-dark flagging this for your decision -- i.e., ship as is with memory use increase, or do some benchmarking to determine the impact.

bkmartinjr (approved):
It would be good to have a C++ person also review before landing.

I have one performance-related question, but it is not a show-stopper.  Otherwise, LGTM.

TY John!

libtiledbsoma/src/soma/managed_query.h
johnkerl: Details in Slack.

johnkerl: Results from our discussion in Slack:

I ran this script: https://gist.github.com/johnkerl/bb5eb154ba06f342fbdcb4ff1361b2fa. What this does is add a half-million enum indices to a dataframe, then a half-million more.

I ran it four ways:
* As on this PR, as-is (using the 1-bit-to-8-bit)
* Using the suggestion from https://arrow.apache.org/docs/format/Columnar.html#validity-bitmaps with `/` and `%` in this diff:
  * https://gist.github.com/johnkerl/dbd69b446fa6324fdc771eccea3bce84  
* Tweaking that to use `>>` and `&`
  * see the same diff
* Using @bkmartinjr 's `ArrayViewIsNull` suggestion:
  * https://gist.github.com/johnkerl/ae9aacf6528571f38f7d7340bd393cb6

I did 10 runs each of the four ways. Timing results are here: https://gist.github.com/johnkerl/b8c32a22d8ad7967793e9abd79243bd0

Results:
* With the 1-bit-to-8-bit copy: about 1.5 seconds end-to-end time for the repro script
* With inline bit arithmetic, about 20ms faster than with the 1-bit-to-8-bit copy.
* With `ArrowArrayViewIsNull`, about 20ms slower than with the 1-bit-to-8-bit copy.

We are going with a preference for the third bullet, which is about 1% slower, but uses less memory.

johnkerl:
> I have one performance-related question, but it is not a show-stopper. Otherwise, LGTM.

https://github.com/single-cell-data/TileDB-SOMA/pull/3992#discussion_r2056201974

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -1450,6 +1479,19 @@ ManagedQuery::_extend_and_evolve_schema_with_details(
...
 
+    // The enumeration values are tracked here in as a vector of string-views,
+    // even for non-string-valued data.  We do this so that we can
+    // differentiate, at the bit-by-bit level, between various floating-point
+    // NaN and Inf values.
+    size_t n = enum_values_in_write.size();
+    std::vector<std::string_view> enum_values_in_write_as_sv(n);
+    for (size_t i = 0; i < n; i++) {
+        auto sv = std::string_view(
+            static_cast<char*>((char*)&enum_values_in_write[i]),
```
nguyenv, line 1490:
```suggestion
            reinterpret_cast<const char*>(&enum_values_in_write[i]),
```

libtiledbsoma/src/soma/managed_query.h
```diff
@@ -805,6 +811,139 @@ class ManagedQuery {
...
+            opt_validities = _cast_validity_buffer(index_array);
+        if (!opt_validities.has_value()) {
+            // All enum values are considered covered: this is the context
+            // where the user has given us an Arrow dictionary, and
+            // the index part is all non-null values and the Arrow library
+            // has no validity buffer at all for it.
+            return std::nullopt;
+        }
+        std::vector<uint8_t> validities = opt_validities.value();
```
jp-dark, line 910:
```suggestion
        std::vector<uint8_t>& validities = opt_validities.value();
```
`std::optional::value` returns one of the following:
* `T&`
* `const T&`
* `T&&`
* `const T&&`

I'm pretty sure we want `T&` - what's written here might ultimately do the right thing, but I'm not 100% sure either way.

johnkerl: Definitely I want the reference -- thank you @jp!!

johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/3992/commits/77c5eca679577115bd35a401af384f8aba7d3f3e

libtiledbsoma/src/soma/managed_query.h
```diff
@@ -943,22 +1084,26 @@ class ManagedQuery {
...
-            // index may be a negative integer, so do not index into
-            // enum_values_in_write or it will segfault
-            if (0 > i) {
-                shifted_indexes.push_back(i);
+        std::unordered_map<ValueType, IndexType> enmr_map;
+        IndexType idx = 0;
+        for (const auto& enmr_value : enmr_vec) {
+            enmr_map.insert(std::make_pair(enmr_value, idx));
+            ++idx;
+        }
```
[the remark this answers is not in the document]
johnkerl: https://github.com/single-cell-data/TileDB-SOMA/pull/3992/commits/77c5eca679577115bd35a401af384f8aba7d3f3e

nguyenv (approved):

PR #3999, opened 2025-04-21
[c++] Fix error in raise-if-slow path

johnkerl:
**Issue and/or context:** Found in internal ops-testing

**Changes:** Fix error introduced on #3958.

**Notes for Reviewer:**

Also has a workaround for #4000 which appeared unbidden today, not caused by this PR.


jp-dark (approved):

bkmartinjr:
I am SO glad you found this!  Thanks @johnkerl

PR #4014, opened 2025-04-24
[python] skip redundant metadata writes for improved performance

bkmartinjr:
**Issue and/or context:**

tiledbsoma.io.from_anndata writes metadata on the obs and var DataFrame to indicate the original index name for the ingested dataframe. This is essentially schema information, as it known at create time, and must never change (i.e., even in append mode, the incoming AnnData must share obs/var structure).

The current code sets this metadata, to the same value, for each AnnData ingested.  This will create a new metadata fragment, containing identical data, for each AnnData ingested. 

**Changes:**

Set the metadata value at create time, rather than at data write time. End result is one fragment, rather than one per AnnData.

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -1455,6 +1455,11 @@ def _write_dataframe_impl(
                 platform_config=platform_config,
                 context=context,
             )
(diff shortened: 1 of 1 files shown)


johnkerl (approved):
🚢

PR #4023, opened 2025-04-29
[c++] Use nanoarrow `ArrowSchemaSetTypeDateTime` for datetime values

nguyenv:
**Issue and/or context:**

[[sc-62364](https://app.shortcut.com/tiledb-inc/story/62364/python-dataframe-incorrectly-handles-dictionary-types-with-a-timestamp-value-type)]

**Changes:**

For datetime types, use nanoarrow's `ArrowSchemaSetTypeDateTime` to set the time. Previously, we were using a workaround for the column's schema to set the datetime. For the dictionary value's schema, we were setting all datetime types to int64. Remove datetime types that are not supported.

### DIFF
--- a/libtiledbsoma/src/soma/managed_query.cc
+++ b/libtiledbsoma/src/soma/managed_query.cc
@@ -637,12 +637,12 @@ void ManagedQuery::_promote_indexes_to_values(
             return _cast_dictionary_values<int32_t>(schema, array);
         case TILEDB_UINT32:
             return _cast_dictionary_values<uint32_t>(schema, array);
-        case TILEDB_INT64:
-            return _cast_dictionary_values<int64_t>(schema, array);
         case TILEDB_DATETIME_SEC:
         case TILEDB_DATETIME_MS:
         case TILEDB_DATETIME_US:
         case TILEDB_DATETIME_NS:
+        case TILEDB_INT64:
+            return _cast_dictionary_values<int64_t>(schema, array);
         case TILEDB_UINT64:
             return _cast_dictionary_values<uint64_t>(schema, array);
         case TILEDB_FLOAT32:
--- a/libtiledbsoma/src/soma/managed_query.cc
+++ b/libtiledbsoma/src/soma/managed_query.cc
@@ -589,29 +589,11 @@ bool ManagedQuery::_cast_column(
             return _cast_column_aux<int32_t>(schema, array, se);
         case TILEDB_UINT32:
             return _cast_column_aux<uint32_t>(schema, array, se);
-        case TILEDB_INT64:
-        case TILEDB_DATETIME_YEAR:
-        case TILEDB_DATETIME_MONTH:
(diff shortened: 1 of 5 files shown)


johnkerl (approved):

libtiledbsoma/src/utils/arrow_adapter.cc
```diff
@@ -1490,51 +1479,44 @@ tiledb_datatype_t ArrowAdapter::to_tiledb_format(
...
+        {"tsm:", NANOARROW_TYPE_TIMESTAMP}, {"tsn:", NANOARROW_TYPE_TIMESTAMP},
+        {"tsu:", NANOARROW_TYPE_TIMESTAMP}, {"tdD", NANOARROW_TYPE_TIMESTAMP},
+        {"z", NANOARROW_TYPE_BINARY},       {"Z", NANOARROW_TYPE_LARGE_BINARY},
+    };
+
+    try {
+        return _to_nanoarrow_type_map.at(sv);
+    } catch (const std::out_of_range& e) {
+        throw std::out_of_range(
+            fmt::format("ArrowAdapter: Unsupported nanoarrow format: {} ", sv));
```
bkmartinjr, line 1500: I'm not sure end-users will know what `nanoarrow` is. Suggest reverting to the original us of `Arrow`.

Also, suggest using the term "type", not "format".  E.g.,

> ArrowAdaptor: Unsupported Arrow type: ..."

bkmartinjr: Also, this is going to throw an error using the internal type code (e.g., `ttu`), rather than the user-friendly `Time64('us')`.  Can we improve the error message in some way?  Or alternatively, help the user figure out what the internal type code means?

Best, worst idea I have is to include a pointer to the docs ([here](https://arrow.apache.org/docs/format/CDataInterface.html#data-type-description-format-strings)), but it would be better if the error message just said: 
> ArrowAdaptor: Unsupported Arrow type: Time64('us')

bkmartinjr (changes requested):
No concerns with the functional changes.  One question, and a request for improved error messages (improve comprehension by end user)

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -589,29 +589,11 @@ bool ManagedQuery::_cast_column(
             return _cast_column_aux<int32_t>(schema, array, se);
         case TILEDB_UINT32:
             return _cast_column_aux<uint32_t>(schema, array, se);
-        case TILEDB_INT64:
-        case TILEDB_DATETIME_YEAR:
```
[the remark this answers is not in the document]
nguyenv: Yeah, we never actually supported any of those, and it was mistakenly added here at some point. This cleanup just removes this unused code.

bkmartinjr (approved):
thank you Vivian!

PR #4031, opened 2025-05-05
[python] Sanitize keys when generating URIs

nguyenv:
**Issue and/or context:**

[[sc-63402](https://app.shortcut.com/tiledb-inc/story/63402/python-c-security-tiledbsoma-io-from-anndata-derives-uris-from-collection-element-name-which-does-not-correctly-protect-against)]

**Changes:**

Add `_util.sanitize_key` which encodes unsafe characters when generating URIs.


bkmartinjr (commented):
Needs HISTORY.md entry as this definitely changes API behavior

bkmartinjr:
Is there an equivalent code path we need to fix in the R bindings?

apis/python/src/tiledbsoma/_util.py
```diff
@@ -706,3 +707,24 @@ class Sentinel:
 
 
 MISSING = Sentinel()
+
+
+def sanitize_key(key: str) -> str:
+    decoded_name = urllib.parse.unquote(key)
```
bkmartinjr, line 713: this code is present twice

bkmartinjr (commented):
There are a couple of spots in the spatial ingestion code (`io/spatia/ingest.py`) which are lacking the sanitize_key call.

Example: line 527 uses the measurement_name string without escaping it

nguyenv:
Will do the R changes in a separate PR.

apis/python/HISTORY.md
```diff
@@ -25,6 +25,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Security
 
+- \[[#4031](https://github.com/single-cell-data/TileDB-SOMA/pull/4031)\] [python] Encode URI when generating from key
```
bkmartinjr, line 28: could you expand this so it summarizes the change?  I'm thinking something like:

Storage paths generated from a collection keys will be URL-escaped when they contain characters outside of safe characters (defined as `a-zA-Z0-9-_.()^!@+={}~'`). In addition, the storage paths `..` and `.` are prohibited.

bkmartinjr (approved):
This looks great.  My only final request is a better change log entry - I suggested some text, but please feel free to do whatever you feel is appropriate.  We just need the change log to summarize major changes like this.

Approving on the assumption you will fix the change log prior to landing this.

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -606,7 +606,9 @@ def from_anndata(
 
         # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
         # MS/meas
```
jp-dark, line 608: In line 598 above this we are setting `experiment_ms_uri` using `f"{experiment_uri}/ms"` instead of `_util.uri_joinpath`. Could that cause a problem for TileDB uris?

jp-dark (changes requested):
I have one comment that's not directly related to this PR, but might make sense to address here. 

I also noticed missing sanitizers in the following places:
* io/spatial/ingest.py::575 (generating `X_layer_uri`)
* io/spatial/ingest.py::612 (generating `tissue_uri`)

Assuming a no-op isn't expensive, would it easier to maintain adding the `sanitize_key` directly to the `_util.uri_joinpath` function?

apis/python/src/tiledbsoma/io/ingest.py
nguyenv: Yeah I'll take a look at this again - I did originally initially try `sanitize_path(path)` at the beginning of the function, but it broke a few tests (since `path` is not always just a filename), so I just reverted it. Looking at the code again though, I think after it calls `urllib.parse.urlunparse(parts)`, I can just take the filename and sanitize that at the end.

bkmartinjr:
> Assuming a no-op isn't expensive, would it easier to maintain adding the `sanitize_key` directly to the `_util.uri_joinpath` function?

I recommend not doing this.  They are orthogonal functions, and blending them conflates all uses of URI joining with paths generated from keys.

Just my $0.02 :-)

apis/python/src/tiledbsoma/io/spatial/ingest.py
```diff
@@ -592,7 +594,7 @@ def from_visium(
             _maybe_set(
```
jp-dark, line 594: `X_layer_name` on line 575 above is missing the sanitizer.

apis/python/src/tiledbsoma/io/spatial/ingest.py
```diff
@@ -592,7 +594,7 @@ def from_visium(
...
                 experiment, "spatial", spatial, use_relative_uri=use_relative_uri
             )
-            scene_uri = _util.uri_joinpath(spatial_uri, scene_name)
+            scene_uri = _util.uri_joinpath(spatial_uri, _util.sanitize_key(scene_name))
             with _create_or_open_scene(scene_uri, **ingest_ctx) as scene:
                 _maybe_set(
                     spatial, scene_name, scene, use_relative_uri=use_relative_uri
```
jp-dark, line 600: `image_name` on line 612 is missing the sanitizer.

jp-dark (changes requested):

nguyenv:
Also closes SOMA-153

jp-dark (approved):

PR #4038, opened 2025-05-07
Update TileDB core to version 2.28.0rc0

jp-dark:
Update dependencies for release candidate:
- TileDB core 2.28.0rc0
- TileDB-R 3.31.1.1

### DIFF
--- a/apis/r/DESCRIPTION
+++ b/apis/r/DESCRIPTION
@@ -34,7 +34,7 @@ Imports:
     Matrix,
     stats,
     bit64 (>= 4.6.0.1),
-    tiledb (>= 0.30.0),
+    tiledb (>= 0.31.1),
     arrow (>= 14.0.0),
     utils,
     fs,
--- a/.github/workflows/libtiledbsoma-asan-ci.yml
+++ b/.github/workflows/libtiledbsoma-asan-ci.yml
@@ -25,7 +25,7 @@ jobs:
         run: |
           mkdir -p external
             # Please do not edit manually -- let scripts/update-tiledb-version.py update this
-          wget --quiet https://github.com/TileDB-Inc/TileDB/releases/download/2.27.2/tiledb-linux-x86_64-2.27.2-1757013.tar.gz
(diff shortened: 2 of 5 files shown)


johnkerl:
For transparency, I always prefer to explicitly state in each and every such PR that this is done by running

```
./scripts/update-tiledb-version.py 2.28.0-rc0
```

apis/r/DESCRIPTION
```diff
@@ -34,7 +34,7 @@ Imports:
     Matrix,
     stats,
     bit64 (>= 4.6.0.1),
-    tiledb (>= 0.30.0),
+    tiledb (>= 0.31.1),
```
johnkerl, line 37: @jp-dark I've never ever done this for core-rc dependency PRs. (Please see https://github.com/single-cell-data/TileDB-SOMA/pulls?q=is%3Apr+is%3Aclosed.) And it's not in our [established released procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases) (maybe it should be!).

If you want to do this, and if it works, great! cc @mojaveazure for 👀 though.

mojaveazure: I'm all for updating the minimum R version, though in this case it would be `>= 0.32.0` (or 0.31.1.1 for the development version) as that will be the next version of tiledb-r (currently unreleased, but I can cut a new tiledb-r once core 2.28 is officially out)

I will point out that pinning on a version of tiledb-r that falls back on core 2.28 will require holding off on a CRAN submission of tiledbsoma-r until that new version of tiledb-r is on CRAN (should be fast, no manual reviews needed there)

mojaveazure (approved):
Overall, looks good. See comments in https://github.com/single-cell-data/TileDB-SOMA/pull/4038#pullrequestreview-2822545935 wrt pinning of tiledb-r

I'll also request a bump of the R development version (as of writing, from `1.17.99.2` &rarr; `1.17.99.3`) before merging

mojaveazure:
> Should it be `1.18.99` instead?

What lifecycle is this part of? Will core 2.28 be tied to tiledbsoma 1.17 or 1.18? If 1.17, then we bump the develop component in 1.17.99.xxx; otherwise we'll bump to 1.18.99 if tiledbsoma as a whole is going to 1.18.99

PR #4039, opened 2025-05-07
[python] Add pydocstyle lint & fix

bkmartinjr:
Changes:
* Add Ruff's pydocstyle lint
* Run automated reformatting
* Manually fix unsafe lint errors (e.g., D405)

NB: tiledbsoma.io has many missing docstrings.  The D100 family of errors are temporarily disabled in this code path (see pyproject.toml).

Fixes SOMA-35

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -205,7 +205,6 @@ def register_h5ads(
     can control the concurrency using the ``soma.compute_concurrency_level`` configuration
     parameter in the ``context`` argument.
     """
-
     if isinstance(h5ad_file_names, str):
         h5ad_file_names = [h5ad_file_names]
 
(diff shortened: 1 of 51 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -311,7 +310,7 @@ def from_h5ad(
     uns_keys: Sequence[str] | None = None,
     additional_metadata: AdditionalMetadata = None,
 ) -> str:
-    """Reads an ``.h5ad`` file and writes it to an :class:`Experiment`.
+    r"""Reads an ``.h5ad`` file and writes it to an :class:`Experiment`.
```
jp-dark, line 313:
```suggestion
    """Reads an ``.h5ad`` file and writes it to an :class:`Experiment`.
```

jp-dark (approved):
Thanks for doing this! Looks good other than a stray 'r'.

apis/python/src/tiledbsoma/io/ingest.py
bkmartinjr: it is correct as is (won't pass lint otherwise).

the reason is the slashes in the docstring

PR #4055, opened 2025-05-13
[python] check encoding version in all open code paths

bkmartinjr:
Fixes SOMA-54

The `tiledbsoma.open` code path checks encoding version and generates a nice error if the object is unsupported.  All other class-specific open paths failed to perform this check. This can lead to the situation where the user is able to open  SOMA objects with an unsupported version, which will then fail in surprising ways elsewhere.

Changes:
* added a consistent check to all open code paths - now any `open` will do the same checks and generate the same errors (as needed)
* updated tests to verify


apis/python/src/tiledbsoma/_soma_object.py
```diff
@@ -103,6 +108,10 @@ def open(
             tiledb_timestamp,
             clib_type=cls._wrapper_type._WRAPPED_TYPE.__name__,
         )
+        if _read_soma_type(handle) != cls.soma_type:
+            raise SOMAError(
+                "Unexpected SOMA metadaa encoding - unable to determine object type."
```
jp-dark, line 113: I'm not sure this is the right error. Unless I'm misunderstanding this code, it looks like we were able to determine the type - it just wasn't the type we were expecting.

apis/python/src/tiledbsoma/_soma_object.py
```diff
@@ -335,3 +344,37 @@ def exists(
...
 AnySOMAObject = SOMAObject[_tdb_handles.AnyWrapper]
+
+
+def _read_soma_type(hdl: _tdb_handles.AnyWrapper) -> str:
+    obj_type = hdl.metadata.get(SOMA_OBJECT_TYPE_METADATA_KEY)
+    encoding_version = hdl.metadata.get(SOMA_ENCODING_VERSION_METADATA_KEY)
+
+    if obj_type is None:
+        raise SOMAError(
+            f"stored TileDB object does not have {SOMA_OBJECT_TYPE_METADATA_KEY!r}"
```
jp-dark, line 355:
```suggestion
            f"Cannot access stored TileDB object with TileDB-SOMA. The object is missing "
            f"the required '{SOMA_OBJECT_TYPE_METADATA_KEY!r}' metadata key."
```

apis/python/src/tiledbsoma/_soma_object.py
```diff
@@ -335,3 +344,37 @@ def exists(
...
+        )
+
+    if isinstance(obj_type, bytes):
+        obj_type = str(obj_type, "utf-8")
+
+    if not isinstance(obj_type, str):
+        raise SOMAError(
+            f"stored TileDB object {SOMA_OBJECT_TYPE_METADATA_KEY!r}"
+            f" is {type(obj_type)}"
```
jp-dark, line 364:
```suggestion
            f"Cannot access stored TileDB object with TileDB-SOMA. The metadata key "
            f"'{SOMA_OBJECT_TYPE_METADATA_KEY!r}' has unexpected type {type(obj_type)}."
```

apis/python/src/tiledbsoma/_soma_object.py
```diff
@@ -335,3 +344,37 @@ def exists(
...
+        )
+    if encoding_version is None:
+        raise SOMAError(
+            f"stored TileDB object does not have {SOMA_ENCODING_VERSION_METADATA_KEY}"
```
jp-dark, line 368:
```suggestion
            f"Cannot access stored TileDB object with TileDB-SOMA. The object is missing "
            f" the required '{SOMA_ENCODING_VERSION_METADATA_KEY}' metadata key."
```

apis/python/src/tiledbsoma/_soma_object.py
```diff
@@ -335,3 +344,37 @@ def exists(
...
+        )
+
+    if isinstance(encoding_version, bytes):
+        encoding_version = str(encoding_version, "utf-8")
+
+    if encoding_version not in SUPPORTED_SOMA_ENCODING_VERSIONS:
+        raise ValueError(
+            f"Unsupported SOMA object encoding version {encoding_version}. The TileDB-SOMA "
+            f"client library needs to be updated to a more recent version."
```
jp-dark, line 377:
```suggestion
            f"Unsupported SOMA object encoding version {encoding_version}. TileDB-SOMA "
            f"needs to be updated to a more recent version."
```
This one was mine. In retrospect I think specifying "client library" is more likely to be confusing than helpful.

jp-dark (changes requested):
I have some change requests around the error messages. I'm open to other wording than what I provided here, but I would like to make the errors more helpful if possible.

apis/python/src/tiledbsoma/_soma_object.py
bkmartinjr: You are correct - it is either a sign that we have a logic bug, OR the data was corrupted in some weird way.

I'll update the error message.

bkmartinjr:
@jp-dark - I incorporated your message changes (almost literally, with a couple of minor punctuation changes).  Also tried to clarify the "unexpected" type error.  I debated whether this should be an assertion or error, as it isn't something the user can handle, but most of these "bad metadata" checks are similar, so ended up leaving it as is (with clarified message)

jp-dark (approved):

PR #4071, opened 2025-05-20
[python] treat timestamp of zero as equivalent to None

bkmartinjr:
**Issue and/or context:**

Fixes SOMA-72

A timestamp value of zero is not a legal timestamp in TileDB. TileDB APIs treat `None` and `0` as equivalent to the default time of `now`.  As of TileDB-SOMA 1.15 this package regressed and treated zero as a specific time, which will raise inscrutable errors in various code paths.

**Changes:**

A timestamp value of zero (int) or epoch (datetime.datetime) is equivalent to `now`.


jp-dark (approved):
The change here looks good, but we should also update our docstrings to call out this oddity around `timestamp=0`. If it doesn't go in this PR can you add an issue for that?

PR #4074, opened 2025-05-21
[python] unit tests temp file cleanup

bkmartinjr:
**Issue and/or context:**

A number of python unit tests were leaving temp files after a run. Most simply needed to be converted to use the preferred `tmp_path` pytest fixture.

Fixes SOMA-194

There are no change to test semantics, just how those tests create/cleanup temporary files.


jp-dark (approved):
🚢

PR #4075, opened 2025-05-21
[r] Add examples to docs

mojaveazure:
Add examples to exported functions to aid in CRAN submission

Notes for reviewers:
 - while there are a lot of files being changed, more than half of them are `man/*.Rd` files, which are automatically generated and shouldn't need to be manually reviewed
 - this PR has no code changes; all changes are to the documentation (some style, mostly adding examples)
 - the `#' @examplesIf` construct allows conditional examples. This is used as
   - CRAN requires all docs that write to disk to write to a temporary directory
   - CRAN requires that all temporary directories be cleaned up
   - withr handles the temporary directory management, but I don't want to take a hard dependency on it
   - `#' @examplesIf` allows the examples to run if and only if the conditions are met (eg. withr is installed)

Fixes [SOMA-45](https://linear.app/tiledb/issue/SOMA-45/add-examples-to-docs)


mojaveazure:
Replaced by
 - #4088
 - #4089
 - #4090 
 - #4091
 - #4092
 - #4093
 - #4094
 - #4095

PR #4084, opened 2025-05-23
[python] Clean-up Python opening with handle/wrapper

jp-dark:
Changes:
* Inline function used in one place
* Change handle wrapper classes to return handle wrapper instead of internal handle for reopen.
* Replace clib handle with calls to wrapper classes.

Changelog: Refactor - no external changes.


apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -194,12 +212,19 @@ def _opener(
...
         ts = self.context._open_timestamp_ms(timestamp)
         self.metadata._write()
-        return self._handle.reopen(clib.OpenMode.read if mode == "r" else clib.OpenMode.write, (0, ts))
+        clib_handle = self._handle.reopen(clib.OpenMode.read if mode == "r" else clib.OpenMode.write, (0, ts))
+        return self.__class__.open(
+            uri=clib_handle.uri,
+            mode=clib.mode,
+            context=self.context,
+            timestamp=clib.timestamp,
+        )
```
jp-dark, line 227: We are re-opening twice here. This is not new, but it is more evident now that the code is more isolated.

bkmartinjr (approved):

PR #4085, opened 2025-05-23
Versioning policy

bkmartinjr:
Added a policy document defining a version and compatibility policy

Fixes SOMA-200

### DIFF
--- a/dev_docs/POLICIES.md
+++ b/dev_docs/POLICIES.md
@@ -1,0 +1,61 @@
+# TileDB-SOMA Development Policies
+
+## Version policy
+
+TileDB-SOMA uses a loose variation of [semantic versioning](https://semver.org/) to organize and govern API compatibility, deprecations and version numbers.
+
+TileDB-SOMA also uses lifecycle tags to indicate the maturity of an interface. All public API should have a lifecycle tag in docstrings. These tags are patterned after the RStudio lifecycle stage model. Tags are:
+
+- ``experimental``: Under active development and may undergo significant and breaking changes.
+- ``maturing``: Under active development but the interface and behavior have stabilized and are unlikely to change significantly but breaking changes are still possible.
+- ``stable``: The interface is considered stable and breaking changes will be avoided where possible. Breaking changes that cannot be avoided will be accompanied by a major version bump.
+- ``deprecated``: The API is no longer recommended for use and may be removed in a future release.
+
+If no tag is present, the state is ``experimental``.
+
+> [!NOTE]
+> Prior to version 1.18, TileDB-SOMA had no explicit version policy. Versioning was most commonly `1.MAJOR.MINOR`,
+> where breaking changes would occur in `1.X` releases, and any other change could occur on `1.X.Y` versions.
+> Starting with `1.18.0`, versioning will be aligned with the SemVer `MAJOR.MINOR.PATCH` schema.
+
+A release number is comprised of `MAJOR.MINOR.PATCH`.
+
+For **non-experimental** interfaces and features:
+
+- API breaking changes should occur only in `major` releases. These changes will be documented in the Python HISTORY.md or R NEWS.md change logs, with guidance on what has changed, and how to migrate code to the new version. When possible, advance warning will be provided prior to breaking changes.
+- Deprecations may be introduced in `major` or `minor` releases. A deprecation will maintain existing functionality, but provide a warning about the upcoming API change, and any available guidance on migration. Deprecations must be documented in the appropriate change log.
+- Warnings of future compatibility changes, which are not "deprecate and remove", may be introduced in a `major` or `minor` release. Future compatibility warnings must be included in the appropriate change log.
+- A `patch` release may not introduce deprecations or future incompatibility change warnings.
+
+Experimental features may be modified in a `major` or `minor` release without prior warning. However, where possible, advance warning of changes will be provided.
+
+## Warning period
+
+Breaking API changes will provide notice of at least two `minor` releases or six months, whichever is greater, before effecting the change.
+
+## Implementing Deprecation
+
+### Python deprecations
+
+Upon deprecation of Python functionality:
+
+- a `DeprecationWarning` must be issued by the deprecated functionality and a unit test provided to confirm warning issuance.
+- affected lifecyle tags must be set to "deprecated", typically in the docstring and other user-visible documentation.
+- docstrings must clearly indicate the scope of deprecation (e.g, function, a particular set of parameters, etc) and must include actionable guidance on migration.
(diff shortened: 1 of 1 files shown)


dev_docs/POLICIES.md
```diff
@@ -0,0 +1,61 @@
...
+## Version policy
+
+TileDB-SOMA uses a loose variation of [semantic versioning](https://semver.org/) to organize and govern API compatibility, deprecations and version numbers.
+
+TileDB-SOMA also uses lifecycle tags to indicate the maturity of an interface. All public API should have a lifecycle tag in docstrings. These tags are patterned after the RStudio lifecycle stage model. Tags are:
+
+- ``experimental``: Under active development and may undergo significant and breaking changes.
+- ``maturing``: Under active development but the interface and behavior have stabilized and are unlikely to change significantly but breaking changes are still possible.
+- ``stable``: The interface is considered stable and breaking changes will be avoided where possible. Breaking changes that cannot be avoided will be accompanied by a major version bump.
+- ``deprecated``: The API is no longer recommended for use and may be removed in a future release.
```
jp-dark, line 12:
```suggestion
- ``deprecated``: The API is no longer recommended for use and will be removed in a future release.
```
The numpy documentation had a good point about not deprecating things that aren't going to be removed.

dev_docs/POLICIES.md
```diff
@@ -0,0 +1,61 @@
...
+
+If no tag is present, the state is ``experimental``.
+
+> [!NOTE]
+> Prior to version 1.18, TileDB-SOMA had no explicit version policy. Versioning was most commonly `1.MAJOR.MINOR`,
+> where breaking changes would occur in `1.X` releases, and any other change could occur on `1.X.Y` versions.
+> Starting with `1.18.0`, versioning will be aligned with the SemVer `MAJOR.MINOR.PATCH` schema.
```
jp-dark, line 19: I'm fairly certain will have a breaking change in the next release. Let's go ahead and assuming it will be 2.0 that we start using SemVer.

ihnorton (approved):

dev_docs/POLICIES.md
```diff
@@ -0,0 +1,95 @@
...
+- utilize `typing_extensions.deprecated` decorator to provide both static and runtime warnings.
+- update unit tests to ensure that the warning is issued (e.g., utilize `pytest.deprecated_call()`).
+- ensure that `stacklevel` is set so that the warning comes from the invoking (caller's) code.
+
+### R deprecations
+
+Upon deprecation of R functionality, methods, or functions:
+
+- a deprecation warning generated by `lifecycle::deprecate_warn()` must be issued by the deprecated functionality, and a unit test `lifecycle::expect_deprecated()` provided to confirm warning issuance.
+- if any parameters are being deprecated, they must be given a default value of `lifecycle::deprecated()`
```
aaronwolen, line 64:
```suggestion
- if any parameters are being deprecated, they must be given a default value of `lifecycle::deprecated()`.
```

dev_docs/POLICIES.md
```diff
@@ -0,0 +1,95 @@
...
+- documentation must clearly indicate the scope of deprecation (e.g, function, a particular set of parameters, etc) and must include actionable guidance on migration.
```
aaronwolen, line 65:
```suggestion
- documentation must clearly indicate the scope of deprecation (e.g., function, a particular set of parameters, etc) and must include actionable guidance on migration.
```

dev_docs/POLICIES.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Implementing Deprecation
+
+### Python deprecations
+
+Upon deprecation of Python functionality:
+
+- a `DeprecationWarning` must be issued by the deprecated functionality and a unit test provided to confirm warning issuance.
+- affected lifecyle tags must be set to "deprecated", typically in the docstring and other user-visible documentation.
+- docstrings must clearly indicate the scope of deprecation (e.g, function, a particular set of parameters, etc) and must include actionable guidance on migration.
```
aaronwolen, line 44:
```suggestion
- docstrings must clearly indicate the scope of deprecation (e.g., function, a particular set of parameters, etc) and must include actionable guidance on migration.
```

dev_docs/POLICIES.md
```diff
@@ -0,0 +1,95 @@
...
+### R deprecations
+
+Upon deprecation of R functionality, methods, or functions:
+
+- a deprecation warning generated by `lifecycle::deprecate_warn()` must be issued by the deprecated functionality, and a unit test `lifecycle::expect_deprecated()` provided to confirm warning issuance.
+- if any parameters are being deprecated, they must be given a default value of `lifecycle::deprecated()`
+- documentation must clearly indicate the scope of deprecation (e.g, function, a particular set of parameters, etc) and must include actionable guidance on migration.
+- `NEWS.md` must clearly indicate the deprecation and any guidance on migration.
+
+R functionality additionaly goes through a defunction process begining two `minor` releases after the deprecation
```
aaronwolen, line 68:
```suggestion
R functionality additionally goes through a defunction process beginning two `minor` releases after the deprecation.
```

I think we should spell out what becoming defunct means in practice (will the function throw an error and stop working, or?).

For alignment what about the "or three months, whichever is greater"? If two minor releases happen in less than three months, it would be defunct before the three-month mark.

Maybe something like:

> Deprecated R functionality will become defunct
after two `minor` releases or 3 months, whichever is greater.

dev_docs/POLICIES.md
```diff
@@ -0,0 +1,95 @@
...
+
+- a defunct error generated by `lifecycle::deprecate_soft()` must be issued, and a unit test `lifecycle::expect_defunct()` provided to confirm error issuance.
+- documentation must clearly indicate the scope of defunction (e.g, function, a particular set of parameters, etc) and must include actionable guidance on migration.
+- `NEWS.md` must clearly indicate the defunction and any guidance on migration.
+
+At next major release after the defunction, the affected R functionality must be removed
+
+For other breaking changes which are not "deprecate and remove":
+
+- a must be issued by the deprecated functionality and a unit test provided to confirm warning issuance.
```
aaronwolen, line 78: Missing a word between *a* and *must*

aaronwolen (changes requested):
@mojaveazure content looks good and is consistent with python and overall policies. I made a couple suggestions, mostly re typos/spelling. However, one thing I'm not sure about is the use of "defunction." I think it might be clearer if we stick with "defunct", for example:

- "R functionality additionally enters a defunct stage..." instead of
- "R functionality additionally goes through a defunction process..."

Could you also double check each bullet ends with a period (just for consistency).

aaronwolen (approved):
Thanks, @mojaveazure!

PR #4125, opened 2025-06-13
Add delete mode

jp-dark:
**Issue and/or context:**
* Closes SOMA-168
* Closes SOMA-170
* Closes SOMA-180

**Changes:**

C++ Changes:
* Make `SOMAArray` copy constructor the default copy constructor.
* Add `soma_mode_` as a private variable to `SOMAArray` and `SOMAGroup`.
* Add new `OpenMode::soma_delete` option to `OpenMode` enum (the `soma_` prefix was added since `delete` is a special keyword.
* Add `soma_` prefix to existing `OpenMode` values for consistency.
* 

Python Changes:
* Update to a new version of somacore (adds 'd' mode to `somacore.options.OpenMode`)
* Add `verify_open_for_deleting` to SOMA objects
* Rename `_verify_open_for_reading` to `verify_open_for_reading`
* Add `d` as a valid mode for opening/reopening SOMA objects
* Deprecate deleting elements from a `Collection` in write mode
* Add support for deleting elements from a `Collection` in delete mode    

R Changes:
* Deprecate deleting elements from a `Collection` in write mode
* Add support for deleting elements from a `Collection` in delete mode

Before Merging:
* [x] Update changelogs
* [x] Release and update somacore version

### DIFF
--- a/apis/python/HISTORY.md
+++ b/apis/python/HISTORY.md
@@ -8,9 +8,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Added
 
+* \[[#4125](https://github.com/single-cell-data/TileDB-SOMA/pull/4125)\] Add delete mode specified by `mode='d'`.
+* \[[#4125](https://github.com/single-cell-data/tiledb-soma/pull/4125)\] Add `verify_open_for_reading` and `verify_open_for_deleting` to SOMA objects.
+
 ### Changed
 
 ### Deprecated
+
+* \[[#4125](https://github.com/single-cell-data/tiledb-soma/pull/4125)\] Deprecate removing elements from a collection in write mode. In the future, all new removals will need to be done in delete mode.
 
 ### Removed
 
--- a/apis/python/HISTORY.md
+++ b/apis/python/HISTORY.md
@@ -9,7 +9,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Added
 
 * \[[#4125](https://github.com/single-cell-data/TileDB-SOMA/pull/4125)\] Add delete mode specified by `mode='d'`.
-* \[[#4125](https://github.com/single-cell-data/tiledb-soma/pull/4125)\] Add `verify_open_for_reading` and `verify_open_for_deleting` to SOMA objects.
 
 ### Changed
 
--- a/apis/python/src/tiledbsoma/_tdb_handles.py
+++ b/apis/python/src/tiledbsoma/_tdb_handles.py
@@ -63,6 +63,18 @@ _RawHdl_co = TypeVar("_RawHdl_co", bound=RawHandle, covariant=True)
 _SOMAObjectType = TypeVar("_SOMAObjectType", bound=clib.SOMAObject)
 
 
+def _open_mode_to_clib_mode(mode: options.OpenMode) -> clib.OpenMode:
+    """Convert options.OpenMode to clib.OpenMode."""
+    if mode == "r":
+        return clib.OpenMode.soma_read
+    elif mode == "w":
+        return clib.OpenMode.soma_write
+    elif mode == "d":
+        return clib.OpenMode.soma_delete
+    else:
+        raise ValueError(f"Unexpected mode '{mode}'. Valid modes are 'r', 'w', or 'd'.")
+
+
 def open_handle_wrapper(
     uri: str,
     mode: options.OpenMode,
@@ -71,15 +83,7 @@ def open_handle_wrapper(
     clib_type: str | None = None,
 ) -> "Wrapper[RawHandle]":
     """Determine whether the URI is an array or group, and open it."""
-    if mode == "r":
-        open_mode = clib.OpenMode.soma_read
-    elif mode == "w":
-        open_mode = clib.OpenMode.soma_write
-    elif mode == "d":
-        open_mode = clib.OpenMode.soma_delete
-    else:
-        raise ValueError(f"Unexpected mode '{mode}'. Valid modes are 'r', 'w', or 'd'.")
-
+    open_mode = _open_mode_to_clib_mode(mode)
     timestamp_ms = context._open_timestamp_ms(timestamp)
 
     _type_to_open = {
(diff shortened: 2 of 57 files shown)


apis/python/HISTORY.md
```diff
@@ -8,10 +8,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Added
 
+* \[[#4125](https://github.com/single-cell-data/TileDB-SOMA/pull/4125)\] Add delete mode specified by `mode='d'`.
+* \[[#4125](https://github.com/single-cell-data/tiledb-soma/pull/4125)\] Add `verify_open_for_reading` and `verify_open_for_deleting` to SOMA objects.
```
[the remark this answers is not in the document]
bkmartinjr: although...I also noticed you changed the name from `_verify...` to `verify...`, suggesting that you might have intended these to become public.

Rationale?

If they are supposed to be public, we need docstrings, tests, etc.  But I'm not convinced we need them to be public when the user already has access to `object.mode`, etc.

apis/python/src/tiledbsoma/_tdb_handles.py
```diff
@@ -368,7 +398,14 @@ def _opener(
         context: SOMATileDBContext,
         timestamp: int,
     ) -> clib.SOMAArray:
-        open_mode = clib.OpenMode.read if mode == "r" else clib.OpenMode.write
+        if mode == "r":
```
bkmartinjr, line 401: nit: this is code is duplicated in SOMAGroupWrapper and SOMAArrayWrapper. Suggest pulling into a shared utility

apis/python/src/tiledbsoma/pytiledbsoma.cc
```diff
@@ -62,8 +62,9 @@ PYBIND11_MODULE(pytiledbsoma, m) {
     });
 
     py::enum_<OpenMode>(m, "OpenMode")
-        .value("read", OpenMode::read)
-        .value("write", OpenMode::write);
+        .value("soma_read", OpenMode::soma_read)
```
bkmartinjr, line 65: Nit:   I do not object to the name change, but am wondering about the motivation.  Is the goal to distinguish the "soma_read" mode from some other "read" mode?  Seems redundant to me (at first glance).

apis/python/HISTORY.md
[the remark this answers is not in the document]
jp-dark: My intention was to be consistent with the fact that `verify_open_for_writing` is a fully documented part of the API, but I don't feel strongly about it. Let me know if you prefer I rollback the change or add the public docs.

apis/python/src/tiledbsoma/pytiledbsoma.cc
jp-dark: I did this because both `del` and `delete` are protected keywords and this is similar to how TileDB core writes its enum options (just lower case instead of all caps).

libtiledbsoma/src/soma/soma_array.cc
```diff
@@ -139,12 +139,32 @@ SOMAArray::SOMAArray(
...
     , schema_(std::make_shared<ArraySchema>(arr->schema())) {
-    fill_metadata_cache(timestamp);
+    switch (arr_->query_type()) {
+        case TILEDB_READ:
+            soma_mode_ = OpenMode::soma_read;
+            break;
+        case TILEDB_WRITE:
+            soma_mode_ = OpenMode::soma_write;
+            break;
+        default: {  // Remaining querty types are only supported on TileDB
```
bkmartinjr, line 149: "querty" (spelling)

libtiledbsoma/src/soma/soma_group.cc
```diff
@@ -110,6 +113,31 @@ SOMAGroup::SOMAGroup(
...
+        case TILEDB_MODIFY_EXCLUSIVE:  // Not supported in SOMA
+        {
+            const char* query_type_str = nullptr;
+            tiledb_query_type_to_str(group_->query_type(), &query_type_str);
+            throw std::invalid_argument(fmt::format(
+                "Cannot open a TileDB group with query type '{}' in "
+                "TileDB-SOMA.",
+                query_type_str));
+        }
+        default: {  // Remaining querty types are only supported on TileDB
```
bkmartinjr, line 132: querty?

libtiledbsoma/src/soma/soma_group.cc
```diff
@@ -110,6 +113,31 @@ SOMAGroup::SOMAGroup(
...
     , group_(group)
     , timestamp_(timestamp) {
+    switch (group_->query_type()) {
+        case TILEDB_READ:
+            soma_mode_ = OpenMode::soma_read;
+            break;
+        case TILEDB_WRITE:
+            soma_mode_ = OpenMode::soma_write;
+            break;
+        case TILEDB_MODIFY_EXCLUSIVE:  // Not supported in SOMA
```
bkmartinjr, line 123: the only difference between this case and the default case is the exception type.  Is there a reason to treat this mode differently (e.g., InvalidArgument error vs TileDBSOMAError?)

apis/r/tests/testthat/test-11-reopen.R
```diff
@@ -54,6 +54,15 @@ test_that("`reopen()` works on arrays", {
     )
     expect_true(arr$is_open(), info = is_open)
 
+    # Test reopen("DELETE")
+    expect_invisible(arr$reopen("DELETE"), label=lab)
```
mojaveazure, line 58: Style suggestion: R code tends to include spaces around `=` in function calls (also for assignment, but using `=` for assignment instead of `<-` is frowned upon)
```suggestion
    expect_invisible(arr$reopen("DELETE"), label = lab)
```

apis/python/HISTORY.md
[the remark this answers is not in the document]
bkmartinjr: I was unaware that this method had been added at some point.

Generally I don't like trivial methods like this, unless there is a strong usability argument. I feel it is better to just let the user define a lambda or equivalent micro-function.

But given that we have already accepted the API surface area, I agree finishing it makes sense (it is that or deprecate/delete the write mode checker).

apis/python/src/tiledbsoma/pytiledbsoma.cc
bkmartinjr: ah, that makes sense!

If we are serious about this kind of error, strongly recommend enabling the flake8-builtins (ruff mode `A`) support in our ruff lint settings.

apis/python/HISTORY.md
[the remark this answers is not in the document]
jp-dark: I'll switch them to private here and create an issue to decide if we want to make them public or deprecate the existing public method.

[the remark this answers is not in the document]
jp-dark: See new issue https://linear.app/tiledb/issue/SOMA-314/python-make-all-verify-open-methods-private-or-public

mojaveazure (approved):
LGTM; please remember to bump the R develop version in `DESCRIPTION`

PR #4126, opened 2025-06-17
[python] import time check of expected TileDB version

bkmartinjr:
There have been several examples where users end up with corrupted conda environments, manifesting as strange and difficult to diagnose errors. Root cause in at least one is a mismatch between the build-time and run-time TileDB version. This is an attempt to provide some detection and warning of this condition.

Changes:
1. Add C++ API to report the TileDB version that the Python pybind11 module was built against ("expected" TileDB version)
2. Add package import-time check to validate that our build-time and run-time TileDB versions match. A `RuntimeError` (exception) is raised upon mismatch.
3. Tests

Fixes SOMA-267

### DIFF
--- a/apis/python/HISTORY.md
+++ b/apis/python/HISTORY.md
@@ -10,7 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Changed
 
-- \[[#4126](https://github.com/single-cell-data/TileDB-SOMA/pull/4126)\] [python] at package import time, validate that the expected TileDB version is installed and used. Raises a RuntimeWarning if the condition is not met. This is an attempt to better warn users who have corrupted conda installations.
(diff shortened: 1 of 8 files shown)


apis/python/HISTORY.md
```diff
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Changed
 
+- \[[#4126](https://github.com/single-cell-data/TileDB-SOMA/pull/4126)\] [python] at package import time, validate that the expected TileDB version is installed and used. Raises a RuntimeWarning if the condition is not met. This is an attempt to better warn users who have corrupted conda installations.
```
jp-dark, line 13:
```suggestion
- \[[#4126](https://github.com/single-cell-data/TileDB-SOMA/pull/4126)\] [python] At package import time, validate that the expected TileDB version is installed and used. Raises a RuntimeWarning if the condition is not met. This is an attempt to better warn users who have corrupted conda installations.
```

jp-dark (approved):
Looks good other than one grammar nit. Thanks!

apis/python/HISTORY.md
bkmartinjr: I also corrected the HISTORY.md to reflect the `RuntimeError` exception, rather than a warning

aaronwolen (approved):

PR #4129, opened 2025-06-17
[c++/r/python] Move clang-format from CMake to pre-commit and update

jp-dark:
**Issue and/or context:**
Towards SOMA-176

**Changes:**
1. Move clang-format out of CMake superbuild and downstream Makefile and into pre-commit.
2. Update the clang-format version to 19.1.7.
3. Change line length for clang-format and disable line length for comments (necessary for RCpp docs).
4. Add R C++ files that are not auto-generated to clang-format.

Additional chore: mark auto-generated files as such in .gitattributes for better PR experience

### DIFF
--- a/libtiledbsoma/src/utils/version.cc
+++ b/libtiledbsoma/src/utils/version.cc
@@ -34,8 +34,7 @@ std::tuple<int, int, int> embedded_version_triple() {
  * @return std::tuple<int, int, int>
  */
 std::tuple<int, int, int> expected_version() {
-    return std::make_tuple(
-        TILEDB_VERSION_MAJOR, TILEDB_VERSION_MINOR, TILEDB_VERSION_PATCH);
+    return std::make_tuple(TILEDB_VERSION_MAJOR, TILEDB_VERSION_MINOR, TILEDB_VERSION_PATCH);
 }
 
 };  // namespace tiledbsoma::version


mojaveazure (approved):
One question: how does the `.gitattributes` affect the PR experience? For Rd files, GitHub already marks these as autogenerated eg. on #4119
image

Will this change that experience? I do not want the Rd files hidden from a PR as I find it important to see what changes were made in those files, not just to the roxygen2 comments in the `.R` files

.clang-format
```diff
@@ -76,7 +76,7 @@ PenaltyBreakString: 1000
 PenaltyExcessCharacter: 1000000
 PenaltyReturnTypeOnItsOwnLine: 200
 PointerAlignment: Left
-ReflowComments:  true
+ReflowComments:  false
```
johnkerl, line 79: ❤️

jp-dark:
> One question: how does the `.gitattributes` affect the PR experience? For Rd files, GitHub already marks these as autogenerated eg. on #4119 image
> 
> Will this change that experience? I do not want the Rd files hidden from a PR as I find it important to see what changes were made in those files, not just to the roxygen2 comments in the `.R` files

It should be the same. In theory, the only change it so `RcppExports.cpp` also gets that treatment.

mojaveazure:
> It should be the same. In theory, this only changes it so `RcppExports.cpp` also gets that treatment.


Awesome, thanks!

PR #4142, opened 2025-06-25
[python/c++] Create C++ method for reading SOMA datatype and use in Python open factory

jp-dark:
**Issue and/or context:** Towards SOMA-321

**Changes:**
Add functions that read only the SOMA datatype from a TileDB array or group and use to simplify the opening process in Python.


apis/python/src/tiledbsoma/pytiledbsoma.cc
```diff
@@ -120,6 +120,34 @@ PYBIND11_MODULE(pytiledbsoma, m) {
         },
         "Print TileDB internal statistics. Lifecycle: experimental.");
 
+    m.def(
```
bkmartinjr, line 123: recommend tests for these functions (directly called).  In particular, test them for garbage inputs - do they raise correct errors, etc.

jp-dark: I added basic tests for the PyBind functions, but I only added the tests for non-SOMA objects to the C++ layer since we don't currently install TileDB-Py.

PR #4146, opened 2025-06-25
[c++/python/r] Remove extra ManagedQuery constructors

jp-dark:
**Changes:**
* Remove copy constructor for `ManagedQuery`.
* Remove constructor from `ManagedQuery` that uses private `SOMAArray` members.
* Add method to `SOMAArray` that creates and returns a `ManagedQuery`.
* Add method to `SOMAArray` that returns internal TileDB array.
* Use the above in Python and R.

**Notes for Reviewer:**
Pull off of branch adding deletes.

### DIFF
--- a/apis/python/src/tiledbsoma/managed_query.cc
+++ b/apis/python/src/tiledbsoma/managed_query.cc
@@ -31,7 +31,15 @@ void load_managed_query(py::module& m) {
     py::class_<ManagedQuery>(m, "ManagedQuery")
         .def(
             py::init([](SOMAArray array, std::shared_ptr<SOMAContext> ctx, std::string_view name) {
-                return ManagedQuery(array, ctx->tiledb_ctx(), name);
+                return array.create_managed_query(ctx, name);
+            }),
+            py::arg("array"),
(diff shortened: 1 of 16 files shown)


mojaveazure (approved):

apis/python/src/tiledbsoma/managed_query.cc
```diff
@@ -31,7 +31,15 @@ void load_managed_query(py::module& m) {
...
-                return ManagedQuery(array, ctx->tiledb_ctx(), name);
+                return array.create_managed_query(ctx, name);
+            }),
+            py::arg("array"),
+            py::arg("ctx"),
+            py::arg("name") = "unnamed")
+
+        .def(
+            py::init([](SOMAArray array, std::shared_ptr<SOMAContext> ctx, std::string_view name) {
+                return array.create_managed_query(name);
```
johnkerl, line 42: `ctx` isn't used here?

johnkerl: And these two function signatures look identical ...

PR #4156, opened 2025-06-30
[r/ci] Use r-lib/actions for R CI

mojaveazure:
Use actions from [r-lib/actions](https://github.com/r-lib/actions) to run CI instead of bespoke solutions. `r-ci.yml` is replaced with `R-CMD-check.yaml` and `test-coverage.yaml` generated with [`usethis::use_github_action()`](https://usethis.r-lib.org/reference/use_github_action.html) with slight modifications:

 - Ignoring non-R PRs
 - Run on release branches and workflow dispatch
 - Do not run on Windows
 - Test on r-devel and r-oldrel-1 on Ubuntu
 - Error on `R CMD check` ERRORs only
 - Configure https://tiledb-inc.r-universe.dev as an extra repository
 - Configure [P3M](https://packagemanager.posit.co/client/) as the source for Bioconductor packages
 - Ignore testthat snapshots (not used)

Note: testing and coverage are now two separate files:

 - unit tests now include the full `R CMD check --as-cran` workflow. This includes running examples, checking docs (including `\donttest{}`) and vignettes, and validating `DESCRIPTION` and `NAMESPACE` in addition to unit tests
 - coverage runs only on the unit test suite in `apis/r/tests/`; it doesn't need the overhead of the full `R CMD check` suite

I've added a composite action `.github/actions/setup-r[/action.yaml]` to share setting up R and installing package dependencies between the two workflows. This could eventually be shared by `r-python-interop-testing.yaml` to further reduce CI duplication

I can try to merge the two workflows into a single YAML if desired, but I think it's easiest to keep separate

Fixes [SOMA-341](https://linear.app/tiledb/issue/SOMA-341/create-a-github-action-that-runs-tests-as-cran)

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -65,5 +65,5 @@ jobs:
         with:
           args: 'c("--no-manual", "--as-cran")'
           build_args: '"--no-manual"'
-          error-on: '"warning"'
+          error-on: '"error"'
           working-directory: 'apis/r/'


jp-dark (approved):
Thanks for doing this!

PR #4157, opened 2025-06-30
[python] enable Ruff lint rules SLOT, SIM and TID

bkmartinjr:
Enabled Ruff lint rules:
* SLOT: flake8-slots
* SIM: flake8-simplify
* TDY: flake8-tidy-imports

And fixed lint exposed by these rules.

Part of SOMA-323

### DIFF
--- a/apis/python/tests/test_shape.py
+++ b/apis/python/tests/test_shape.py
@@ -99,12 +99,11 @@ def test_sparse_nd_array_basics(
             snda.read(coords).tables().concat()
 
     # Test writes out of bounds
-    with tiledbsoma.SparseNDArray.open(uri, "w") as snda:
-        with pytest.raises(tiledbsoma.SOMAError):
-            dikt = {name: [shape + 20] for name, shape in zip(dim_names, arg_shape)}
-            dikt["soma_data"] = [30]
-            table = pa.Table.from_pydict(dikt, schema=snda.schema)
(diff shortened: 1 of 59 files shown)


apis/python/tests/test_shape.py
```diff
@@ -679,10 +674,7 @@ def _check_ndarray(ndarray, has_shapes, expected_shape):
     dict_output["ms"]["raw"]["var"]["uri"] = "test"
     dict_output["ms"]["raw"]["X"]["data"]["uri"] = "test"
 
-    if has_shapes:
-        var_max_domain_hi = 9223372036854773968
-    else:
-        var_max_domain_hi = 9223372036854773758
+    var_max_domain_hi = 9223372036854773968 if has_shapes else 9223372036854773758
```
ktsitsi, line 677: Could we assign this magic numbers in a `global` constant?

bkmartinjr: I agree that the tests need cleanup - for example, they don't use type hints. I have had to disable quite a few of the linter rules for the tests directory.

My preference is to just focus on the `src` (i.e. shipped code) in these PRs, and later work on incrementally fixing tests.  Please LMK your thoughts.

The above code is simply an auto-fix - the constants were from the original shaping work done in 1.15

jp-dark (approved):

PR #4175, opened 2025-07-22
[c++] Prepare for adding `delete_cells` to `SOMADataFrame`

jp-dark:
**Issue and/or context:** Towards SOMA-333

**Changes:**
* Create classes that wrap coordinate selection types (std::span of points or std::pair for a slice). Do any type checking in these classes. This switches from std::vector to std::span to make passing data from Python/R without copying easier.
* Move std::variant decay from `SOMASparseNDArray` to `SOMACoordQueryCondition`.

Note:
These changes are to help with implementing `delete_cells` in `SOMADataFrame` where the index columns can take any simple type.

### DIFF
--- a/libtiledbsoma/src/common/soma_column_selection.h
+++ b/libtiledbsoma/src/common/soma_column_selection.h
@@ -38,11 +38,11 @@ struct SOMASliceSelection {
     SOMASliceSelection(T slice_start, T slice_stop)
         : start{slice_start}
         , stop{slice_stop} {
-        // Using sstream because we don't want to include fmt directly in external header.
-        std::stringstream ss;
-        ss << "Invalid slice [ " << start << ", " << stop
-           << "]. The lower bound must be less than or equal to the upper bound.";
         if (stop < start) {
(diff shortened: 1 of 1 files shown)


libtiledbsoma/src/common/soma_column_selection.h
```diff
@@ -0,0 +1,99 @@
...
+     * Creates a closed slice (includes both end points).
+     *
+     * @param slice_start The first value (inclusive) of the slice.
+     * @param slice_stop The last value (inclusive) of the slice.
+     */
+    SOMASliceSelection(T slice_start, T slice_stop)
+        : start{slice_start}
+        , stop{slice_stop} {
+        // Using sstream because we don't want to include fmt directly in external header.
+        std::stringstream ss;
```
bkmartinjr, line 42: suggest moving `ss` create & init inside the if statement scope, so it is only done when there is an actual failure.

bkmartinjr (approved):
one nit, but otherwise looks good

PR #4197, opened 2025-08-08
[python] Fix type check for index columns in dataframe

jp-dark:
**Changes:**
The type check for index columns was skipping checks on dictionary types, but dictionary types get converted to enumerations which are not support only TileDB dimensions.

### DIFF
--- a/apis/python/src/tiledbsoma/_dataframe.py
+++ b/apis/python/src/tiledbsoma/_dataframe.py
@@ -880,7 +880,11 @@ def _canonicalize_schema(
                 f"All index names must be defined in the dataframe schema: '{index_column_name}' not in {schema_names_string}",
             )
         dtype = schema.field(index_column_name).type
-        if not pa.types.is_dictionary(dtype) and dtype not in [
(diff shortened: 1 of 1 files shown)


apis/python/src/tiledbsoma/_dataframe.py
```diff
@@ -880,7 +880,11 @@ def _canonicalize_schema(
                 f"All index names must be defined in the dataframe schema: '{index_column_name}' not in {schema_names_string}",
             )
         dtype = schema.field(index_column_name).type
-        if not pa.types.is_dictionary(dtype) and dtype not in [
+        if pa.types.is_dictionary(dtype):
+            raise TypeError(
+                f"Cannot set index column '{index_column_name}' to an enumeration. Index columns do not support enumerations."
+            )
+        if dtype not in [
```
bkmartinjr, line 887: super tiny nit: the array of types being tested against should be defined in a global constant, as it is a base TileDB characteristic....

bkmartinjr (approved):

PR #4210, opened 2025-08-27
[python] Document supported workflows for registration functions

aaronwolen:
**Issue and/or context:** SOMA-275

**Changes:** Update `register_h5ads()` docstrings to document the two supported append-mode workflows. I also reworded some of the narrative text for clarity.

**Notes for Reviewer:** This also restructures some of the additional narrative text to use indentation rather than underlines to separate sections, which aligns with Google style docstrings.

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -283,9 +283,9 @@ def register_anndatas(
     context: SOMATileDBContext | None = None,
     allow_duplicate_obs_ids: bool = False,
 ) -> ExperimentAmbientLabelMapping:
-    """Extends registration data from the baseline, already-written SOMA
-    experiment to include multiple H5AD input files. See ``from_h5ad`` and
-    ``from_anndata`` on-line help.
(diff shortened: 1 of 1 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -398,9 +419,16 @@ def from_h5ad(
...
-                  exp.obs.metadata.update({"ccc": 123})
+              with soma.open(
+                  uri, "w"
+              ) as exp:
+                  exp.metadata.update({
+                      "aaa": "BBB"
+                  })
+                  exp.obs.metadata.update({
+                      "ccc": 123
+                  })
```
jp-dark, line 430: I think the old formatting was better. Was this an intentional change or something automated? (Similarly with the other formatting-only changes).

aaronwolen: Whoops, I was experimenting with ruff's docstrings formatter and didn't notice it made these changes. Will revert.

jp-dark (approved):

PR #4212, opened 2025-08-28
Adding `.type` property for Sparse/DenseNDArrays

ktsitsi:
**Issue and/or context:**

SOMA SparseNDArray and DenseNDArray arrays are created with a specific data type (e.g., type=pa.float32), but there is no direct way to retrieve this type via the API. Users must currently inspect the underlying Arrow schema and infer the type from there.

**Changes:**
Adding a new `property` inside the NDArray class to be inherited by both Sparse and DenseNDArray classes.

**Notes for Reviewer:**
None


bkmartinjr:
Need to add a line to the changelog

bkmartinjr (approved):
Looks great.  Needs change log entry before merging.

PR #4229, opened 2025-09-09
Fixing shadowing logging.py module

ktsitsi:
**Issue and/or context:**

**Changes:**

**Notes for Reviewer:**


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -844,13 +844,235 @@ def _ingest_obs_var_m_p(
...
+        platform_config=platform_config,
+        context=context,
+        axis_0_mapping=axis_0_mapping,
+        axis_1_mapping=axis_1_mapping,
+        must_exist=True,
+    ):
+        _logging.log_io_same(_util.format_elapsed(s, f"Finish writing X for {X.uri}"))
+    return X.uri
+
+
```
jp-dark, line 1075:
```suggestion
```
Looks like the rebase accidentally added back in these removed functions. They should be re-removed before merging.

jp-dark (changes requested):
There was some removed functions that must have accidentally gotten added back during the rebase, but looks good otherwise.

jp-dark (approved):
Thanks!

apis/python/src/tiledbsoma/__init__.py
```diff
@@ -236,6 +237,9 @@
     "get_implementation_version",
     "get_libtiledbsoma_core_version",
     "get_storage_engine",
+    "logging_level_debug",
```
bkmartinjr, line 240: having these as public functions is really not how the Python logging system is supposed to be configured.  Typical best practices for a package is to document the logger name (typically it is the package or module name), and then let the application (not the package) configure logging with the standard `logging` package (e.g., a [logging config](https://docs.python.org/3/howto/logging.html#configuring-logging)).

In my view, we should deprecate (and eventually remove) these, as they will step on any application-specified logging config.

doc/source/python-tiledbsoma.rst
```diff
@@ -76,3 +76,9 @@ Functions
     tiledbsoma.tiledbsoma_stats_dump
     tiledbsoma.tiledbsoma_stats_as_py
     tiledbsoma.tiledbsoma_stats_json
+
+    tiledbsoma.logging_level_debug
+    tiledbsoma.logging_level_info
+    tiledbsoma.logging_level_warning
+
+    tiledbsoma.tiledbsoma_build_index
```
bkmartinjr, line 84: I thought we removed `tiledbsoma_build_index`.  Accidentally adding it back?

apis/python/src/tiledbsoma/__init__.py
bkmartinjr: More concretely - what I suggest is that we leave the names as they were, deprecate them, and remove in the next release.

ktsitsi: Do you want me to close this PR and instead create a ticket for deprecating the logging as it is now and revisit it in a holistic approach in the near future?

jp-dark: > Do you want me to close this PR and instead create a ticket for deprecating the logging as it is now and revisit it in a holistic approach in the near future?

Yes, please do that and set it to triage. Add a link to the [logging config](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) documentation that @bkmartinjr linked above.

PR #4234, opened 2025-09-15
[ci] Change Python packaging CI in preparation for cibuildwheel changes

jp-dark:
**Issue and/or context:** Towards SOMA-390

**Changes:**
* Remove Python CI packaging tests
* Split sdist and wheel build and testing into separate jobs
* Use brew instead of xcode action to get macos dependencies

### DIFF
--- a/.github/workflows/python-packaging.yml
+++ b/.github/workflows/python-packaging.yml
@@ -51,7 +51,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, ubuntu-20.04-arm, macos-15-intel, macos-15]
+        os: [ubuntu-24.04, ubuntu-24.04-arm, macos-15-intel, macos-15]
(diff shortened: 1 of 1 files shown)


bkmartinjr (approved):

.github/workflows/python-packaging.yml
```diff
@@ -22,186 +20,186 @@ on:
...
-  sdist:
+  build_sdist:
     name: Build source distribution
     runs-on: ubuntu-24.04
     steps:
       - name: Checkout TileDB-SOMA
         uses: actions/checkout@v4
         with:
-          fetch-depth: 0 # ensure we get all tags to inform package version determination
+          fetch-tags: true  # ensure we get all tags for package version number
```
jp-dark, line 30:
```suggestion
          fetch-depth: 0
          fetch-tags: true  # get tags for package version number
```
The fetch-tags doesn't work correctly unless `fetch-depth: 0` is also set.

ktsitsi:
Rebased and commit suggestion

PR #4237, opened 2025-09-16
Adding null support for query condition in SOMA queries

ktsitsi:
**Issue and/or context:**
Support for null test query conditions was added to TileDB core earlier this year and we haven't yet plumbed through support to TileDB-SOMA.

**Changes:**
- Adding `init_null` in tiledbsoma PyQueryCondition
- Adding unit test for `None` value inside a query condition

**Notes for Reviewer:**

This PR does not compare the pandas results against the SOMA results of conditions with None values since the have a different syntax.


apis/python/tests/test_query_condition.py
```diff
@@ -50,6 +50,7 @@ def soma_query(uri, condition):
         'louvain == "NK cells"',  # string
         "percent_mito > 0.02",  # float
         "is_b_cell == True",  # bool
+        "ord == None",  # None values
```
bkmartinjr, line 53: I think it is worth adding all of the other operators that support testing for None/NA, e.g., `ord != None`.  Cheap to test, and may catch edge cases.

apis/python/tests/test_query_condition.py
```diff
@@ -76,15 +77,51 @@ def soma_query(uri, condition):
...
+        pydict["soma_joinid"] = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+        pydict["int"] = [1, 2, 3, 4, 5, 6, None, 8, None, None]
+        pydict["bool"] = [True, True, True, False, True, False, None, False, None, None]
+        pydict["ord"] = pd.Categorical(["g1", "g2", "g3", None, "g2", "g3", "g1", None, "g3", "g1"])
+        data = pa.Table.from_pydict(pydict)
+
+        with soma.DataFrame.create(uri, schema=asch, domain=[[0, 9]]) as sdf:
+            sdf.write(data)
+
+        soma_arrow = soma_query(uri, condition)
```
bkmartinjr, line 113: Pandas does support testing for NA, just with a different syntax.  I think it is worthwhile validating that we get the same result, e.g., table->to_pandas->query works as expected and gives same result as our filter.

bkmartinjr (changes requested):
the changes to the package look fine.  I'd like to see the test improved in a number of ways.

apis/python/tests/test_query_condition.py
ktsitsi: >  just with a different syntax.

Yeap noted that in the description. I ll add it.

bkmartinjr (approved):
approving, but I do think the new test needs more conditions just to be on the safe side

ktsitsi:
@bkmartinjr https://github.com/single-cell-data/TileDB-SOMA/pull/4237/commits/2b87e94dc5e4986987cf5e1ee65dc7c3606cc3f8

PR #4240, opened 2025-09-18
[python] Remove deprecated python functions

jp-dark:
* Remove deprecated function `tiledbsoma_build_index`, `io.append_obs`, `io.append_var`, `io.append_X`, and `io.create_from_matrix`. 
* Remove the deprecated method `config_options_from_schema` from `SOMAArray`.


bkmartinjr (approved):
the code changes look find.  I don't know what to make of the HT failure based upon the logs.  It seems most likely a test issue, or a latent bug unrelated to your changes.

My suggestion is to run this test locally with the `--hypothesis-profile=expensive`, which will test much more of the search space.

Approving, but I do think it would be a good idea to do the above in parallel

aaronwolen (approved):
Thank you!

PR #4245, opened 2025-09-23
[python] Deprecate `NotCreateableError`

jp-dark:
This exception is not being used internally.


bkmartinjr (commented):
I'm not sure that considering `AlreadyExists` and `NotCreatable` as equivalent makes sense.  

And the docstring in most cases was different:
> If the URI is malformed for a particular storage backend.
vs
> If the underlying object already exists at the given URI.

These seem like entirely different error conditions.

I presume `NotCreatable` is no longer used, in favor of some other exception (`ValueError` maybe)?  If that is the case, then:
* I am fine deprecating it
* But the HISTORY.md note is wrong in that `AlreadyExists` is not a synonym

PR #4258, opened 2025-10-01
Adding shape validation for add_matrix_to_collection

ktsitsi:
**Issue and/or context:**
The SOMA specification defines strict requirements for the shape of layers within pre-defined collections:
```
X: (O, V)
obsm: (O, M)
varm: (V, M)
obsp: (O, O)
varp: (V, V)
```
Where` O` is the domain of `obs.soma_joinid`, `V` is the domain of `var.soma_joinid` in the containing `SOMAMeasurement`, and `M` is user-defined.

However, the `soma.io.add_matrix_to_collection()` and `soma.io.add_X_layer()` functions do not validate the shape of incoming matrices.
**Changes:**

- Introducing an internal validation function for testing the shape of the incoming Layer/Matrix to be added in a SOMA measurement.
- Tests for `add_matrix_to_collection` and `add_X_layer`

**Notes for Reviewer:**

**DEFAULT BEHAVIOR: schema_validation = **True** **

### DIFF
--- a/apis/python/tests/test_basic_anndata_io.py
+++ b/apis/python/tests/test_basic_anndata_io.py
@@ -576,7 +576,10 @@ def test_matrix_X_layer_enforcement(conftest_pbmc_small, tmp_path, offset):
     bad_matrix = np.ones(bad_shape)
 
     # Should raise error for invalid shape
-    with _factory.open(output_path, "w") as exp, pytest.raises(tiledbsoma.SOMAError):
+    with (
+        _factory.open(output_path, "w") as exp,
+        pytest.raises(tiledbsoma.SOMAError, match=r"Matrix .* shape .* is incompatible with target shape *."),
+    ):
         tiledbsoma.io.add_X_layer(exp, "RNA", "bad", bad_matrix, schema_validation=True)
 
 


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1681,6 +1682,71 @@ def update_matrix(
...
+    *,
+    context: SOMATileDBContext | None = None,
+) -> None:
+    """Validates the shape of a new matrix against an existing matrix in a collection.
+
+    Raises:
+        ValueError: If the matrix shape is incompatible.
+    """
+    target_shapes = get_experiment_shapes(exp.uri, context=context)
+    target_shape = target_shapes["ms"][measurement_name]["X"]["data"]["shape"]  # (O, V)
```
bkmartinjr, line 1700: I don't think this is the correct way to determine the target shape.  The "shape" is entirely determined by the axis dataframes.

There is no guarantee that an X['data'] will exist - it is simply a named array, like any other.

I believe that the actual constraint we should be checking is that the given shape (aka current domain) is equal to the shape/current domain for the associated axis dataframe.

This allows for cases where deletes have occurred, or other non-contiguous joinids exist.

For example:
* obs_soma_joinid_domain: this is the domain of the soma_joinid dimension for the obs dataframe
* var_soma_joinid_domain: the domain of the soma_joinid dimension for the var dataframe
In which case for an X sparse matrix:
* the soma_dim_0 domain must be equal to the obs_soma_joinid_domain
* the soma_dim_1 domain must be equal to the var_soma_joinid_domain

Also remember that the axis dataframes may have multiple dimensions, so you have to carefully determine the domain that is associated with the soma_joinid.

bkmartinjr (changes requested):
target shape determination is based upon the axis dataframe, not X['data']

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1722,6 +1788,7 @@ def add_matrix_to_collection(
     ingest_mode: IngestMode = "write",
     use_relative_uri: bool | None = None,
     context: SOMATileDBContext | None = None,
+    schema_validation: bool = False,
```
bkmartinjr, line 1791: Missing docstring for this (and all other) parameters.  Seems like a good time to fix?  (CC: @aaronwolen)

bkmartinjr, line 1791: Do we intend for this param to be `False` by default?  It seems like this parameter should be `True` by default, given the nature of the check.

Or if we are looking to migrate users over time, we could:
* Always: issue a warning if the schema fails to validate
* Flip this param to `True` at the next major release (3.0)

@aaronwolen - thoughts?

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1681,6 +1681,70 @@ def update_matrix(
...
+                f"Matrix '{matrix_name}' must match the observation size (O). "
+                f"Found {matrix_data.shape[0]}, expected {target_obs_size}."
+            )
+    elif collection_name == "varm":
+        if matrix_data.shape[0] != target_var_size:
+            raise SOMAError(
+                f"Matrix '{matrix_name}' must match the variable size (V). "
+                f"Found {matrix_data.shape[0]}, expected {target_var_size}."
+            )
+    elif collection_name == "obsp":
```
bkmartinjr, line 1728: `obsp` and `varp` checks are incorrect.  These arrays can have ndim>2.  The check should be that the first two dimensions match the respective axis shape.

I.e., for `obsp`, you want to check that `matrix_data.shape[0:2] == (target_obs_size, target_obs_size)`  (note the slice on the matrix shape).

apis/python/src/tiledbsoma/io/ingest.py
ktsitsi: I thought that since the stage of the function is `maturing` I did not want to go for throttle on that, risking breaking API backwards compatibility. Of course I could set it to `True` if we decide that we want to hard impose it.

apis/python/tests/test_basic_anndata_io.py
```diff
@@ -472,6 +472,85 @@ def test_add_matrix_to_collection(conftest_pbmc_small, tmp_path):
         assert sorted(list(exp_r.ms["RNA"]["newthing"].keys())) == ["X_pcd"]
 
 
+@pytest.mark.parametrize(
```
bkmartinjr, line 475: there do not appear to be any tests with pre-1.15 (old-style shape) Experiments.

Is that explicitly unsupported, or should we add tests?  The implementation in ingest.py seems to support it, so I suggest testing.  There is sample data already available to make this really easy -- see `test_soma_experiment_versions.py` for an example test that uses many old variants for complete coverage

apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -1654,6 +1654,80 @@ def update_matrix(
...
+        measurement_name: Name of the measurement
+        collection_name: Name of the collection
+        matrix_name: Name of the matrix
+        matrix_data: The matrix data to validate
+        meas: The already-opened measurement object
+
+    Raises:
+        ValueError: If the matrix shape is incompatible.
+    """
+    obs_soma_joinid_domain = exp.obs._maybe_soma_joinid_shape
```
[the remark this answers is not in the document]
ktsitsi: Added test case only for `pre-1.15 experiment + validation on + correctly-sized matrix` where the default schema validation is being bypassed.

apis/python/tests/test_basic_anndata_io.py
```diff
@@ -472,6 +473,108 @@ def test_add_matrix_to_collection(conftest_pbmc_small, tmp_path):
...
+        ("obsp", (-1, 0)),
+        ("obsp", (0, 0)),
+        ("varp", (1, 0)),
+        ("varp", (0, 1)),
+        ("varp", (-1, 0)),
+        ("varp", (0, 0)),
+    ],
+)
+# "1.7.3", "1.12.3", "1.14.5, "1.15.0", "1.15.7",
+@pytest.mark.parametrize("version", ["1.14.5"])
```
bkmartinjr: is there a reason only 1.14.5 is tested?

apis/python/tests/test_basic_anndata_io.py
```diff
@@ -472,6 +473,108 @@ def test_add_matrix_to_collection(conftest_pbmc_small, tmp_path):
...
+    # Determine expected shape based on matrix type
+    shape_map = conftest_pbmc_small.n_obs, conftest_pbmc_small.n_vars
+    base_shape = shape_map
+    bad_shape = (max(1, base_shape[0] + offset[0]), max(1, base_shape[1] + offset[1]))
+
+    # Create invalid matrix
+    bad_matrix = np.ones(bad_shape)
+
+    # Should raise error for invalid shape
+    with _factory.open(output_path, "w") as exp, pytest.raises(tiledbsoma.SOMAError):
```
bkmartinjr: nit: I would add a `match` arg to `pytest.raises`, just to confirm you are getting the Error message you expect.

E.g., `pytest.raises(tiledbsoma.SOMAError, match=r"Matrix .* shape")` or something like that....

bkmartinjr (changes requested):
Code in ingest looks good (with one minor suggestion).  However, the new test looks unfinished - see inline comments.

apis/python/tests/test_basic_anndata_io.py
```diff
@@ -472,6 +473,108 @@ def test_add_matrix_to_collection(conftest_pbmc_small, tmp_path):
...
+        "obsp": (n_obs, n_obs),
+        "varp": (n_vars, n_vars),
+    }
+    base_shape = shape_map[matrix_type]
+    bad_shape = (max(1, base_shape[0] + offset[0]), max(1, base_shape[1] + offset[1]))
+
+    # Create invalid matrix
+    bad_matrix = csr_matrix((bad_shape[0], bad_shape[1])) if matrix_type in ["obsp", "varp"] else np.ones(bad_shape)
+
+    # Should raise error for invalid shape
```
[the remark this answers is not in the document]
ktsitsi: True. Just left this comment as a general comment a little bit higher than you should have expected. Fixed it now by moving it below and explain each case with extra comments.

apis/python/tests/test_basic_anndata_io.py
ktsitsi: Added a regex match for the SOMAError.

apis/python/tests/test_basic_anndata_io.py
```diff
@@ -472,6 +474,112 @@ def test_add_matrix_to_collection(conftest_pbmc_small, tmp_path):
...
+        ("obsp", (0, 1)),
+        ("obsp", (-1, 0)),
+        ("obsp", (0, 0)),
+        ("varp", (1, 0)),
+        ("varp", (0, 1)),
+        ("varp", (-1, 0)),
+        ("varp", (0, 0)),
+    ],
+)
+@pytest.mark.parametrize("version", ["1.14.5", "1.7.3", "1.12.3", "1.14.5", "1.15.0", "1.15.7"])
```
bkmartinjr: nit: please add "1.16.1" (we have the dataset, might as well get the coverage)

apis/python/tests/test_basic_anndata_io.py
bkmartinjr: nit: actually, it looks like you did not add it - the code still lacks a `match` param.

bkmartinjr (approved):
Two tiny nits in the tests.  Would be great if you can fix as they are super easy.

Otherwise, it LGTM.

apis/python/tests/test_basic_anndata_io.py
ktsitsi: Oh I missed the second test, I had done it for the other one. Fixed now.

PR #4283, opened 2025-10-14
[python] Carrara/tiledbv3 URI support

bkmartinjr:
Changes:
* preliminary method to distinguish tiledb URI mode: absolute or relative (aka Carrara vs Cloud)
* add relative URI handling for tiledb URIs
* draft test cases - hard-coded against QA backend for now (must be manually executed outside of CI)

This PR also contains the following changes, which have been split out into #4286 and #4287:
* fix mis-use of Core group add API noticed during review
* fix bug in Collection._set_element that was adding every group member as a "Collection" (in the memory-cached collection, i.e., what the `.members()` method return).

Work not complete in this PR:
* port SOMATileDBContext.data_protocol to use the (not yet available) Core API to introspect Ctx state
* align SOMATileDBContext.data_protocol signature with TileDB-Py

### DIFF
--- a/apis/python/tests/test_util.py
+++ b/apis/python/tests/test_util.py
@@ -52,8 +52,6 @@ def test_is_relative_uri() -> None:
     assert not is_relative_uri("gs://foo/A")
     assert not is_relative_uri("tiledb://foo/")
 
-    # TODO: if `tiledb` has a relative syntax (unknown), add a test.
-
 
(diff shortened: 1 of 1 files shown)


jp-dark (approved):
I have a couple nits about comments, but overall this looks good with the understanding we will switch out the tiledb_v3 vs tiledb_v2 test once TileDB core 2.30.0 is released.

PR #4294, opened 2025-10-20
Use vcpkg in place of superbuild

XanthosXanthopoulos:
**Issue and/or context:**
So far TileDB-SOMA depended on a custom superbuild for building the soma-core C++ library and for resolving package dependencies. We substitute the way we package the library by using vcpkg instead.

**Changes:**
- Remove most of the logic under `FindSpdlog_EP.cmake`.
- Add `spdlog` as a dependency for `vcpkg` to resolve.
- Fix the `CMAKE-TOOLCHAIN-PATH` to point to vcpkg location.
- Fix the vcpkg.cmake presets, introducing both a `release` and an `asan` (Linux-only) version
- Substitute all custom builds/installations of the `libtiledbsoma` library with the `vcpkg's` prebuilt library in all CI workflows.
- Changing the R configure script to ingest the prebuilt `libtiledbsoma` from `vcpkg`

**Notes for Reviewer:**

We are gonna move forward in substituting the custom python packaging using `scikit-build-core`. This will require:

- Remove `setup.py` from the apis/python in favor of `project.toml`.
- Modify the custom build backend and introduce `scikit-core-build` for wheel packaging.
- This change will align perfectly with the `cibuildwheel`action completing the build system reform of TileDB-SOMA.

### DIFF
--- a/.github/workflows/build-cpp.yml
+++ b/.github/workflows/build-cpp.yml
@@ -27,6 +27,14 @@ jobs:
         run: |
           ./vcpkg/bootstrap-vcpkg.sh
           echo "CMAKE_TOOLCHAIN_FILE=${{ github.workspace }}/vcpkg/scripts/buildsystems/vcpkg.cmake" >> $GITHUB_ENV
+          
+      - name: Set macOS architecture
+        if: ${{ runner.os == 'macOS' }}
+        run: |
+          # Detect architecture (arm64 for Apple Silicon, x86_64 for Intel)
+          ARCH=$(uname -m)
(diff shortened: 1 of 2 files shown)


jp-dark (commented):
I haven't had a chance to fully review this. I have a couple small initial comments, but still need to test it locally and look at the CI.

jp-dark:
The R universe jobs succeeded (other than the expected failures): https://github.com/r-universe/mojaveazure/actions/runs/19048861521

jp-dark (approved):

PR #4299, opened 2025-10-29
[c++] Implement experimental Memory Pool for `ManagedQuery` read operation

XanthosXanthopoulos:
**Issue and/or context:** 

**Changes:**
This PR introduces a new experimental memory allocation mechanism for read operations. Instead of a per column memory budget, the new mechanism uses a global memory budget which is then split to each column based on the column byte size. 

The new mechanism allows for more predictable memory usage which is independent of the underlying array schemas.

Additionally it changes how arrow arrays are constructed regardless of the memory allocation method. Each arrow table allocates its own buffers and copies data from the `ManagedQuery` read buffers which are reused for all incomplete reads. Additionally dictionary arrays are now allocated similarly to ordinary arrays and can be reused across different columns if the use the same `Enumeration`

**Notes for Reviewer:**
The new mechanism also changes the way memory is allocated (using `resize` instead of `reserve` to avoid UB) and requires lower limits. To enable enable the new mechanism you need to set `soma.read.use_memory_pool` to `true` and `soma.read.memory_budget` (defaults to 256MB)

### DIFF
--- a/apis/python/HISTORY.md
+++ b/apis/python/HISTORY.md
@@ -8,8 +8,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Added
 
+- \[[#4299](https://github.com/single-cell-data/TileDB-SOMA/pull/4299)\] Use a global memory budget for read operations instead of a per column memory budget. The global memory budget allocates splits the budget per column depending on the type and characteristics of each column. Global memory budget is disabled by default under a feature flag and can be enabled by setting `soma.read.use_memory_pool`.
+
 ### Changed
 
+- \[[#4299](https://github.com/single-cell-data/TileDB-SOMA/pull/4299)\] `ManagedQuery` reuses the same buffers for each incomplete read and allocates dedicated buffers when converting to Arrow.
 - \[[#4258](https://github.com/single-cell-data/TileDB-SOMA/pull/4258)\] Add optional schema validation to `tiledbsoma.io.add_X_layer`, which will generate an error if the provided matrix does not match the Experiment shape. Validation is enabled by default, for any dataset created with SOMA 1.15 or later.
 
 ### Deprecated
--- a/apis/r/NEWS.md
+++ b/apis/r/NEWS.md
@@ -2,7 +2,11 @@
 
 ## Added
 
+- Use a global memory budget for read operations instead of a per column memory budget. The global memory budget allocates splits the budget per column depending on the type and characteristics of each column. Global memory budget is disabled by default under a feature flag and can be enabled by setting `soma.read.use_memory_pool`. ([#4299](https://github.com/single-cell-data/TileDB-SOMA/pull/4299))
+
 ## Changed
+
+- `ManagedQuery` reuses the same buffers for each incomplete read and allocates dedicated buffers when converting to Arrow. ([#4299](https://github.com/single-cell-data/TileDB-SOMA/pull/4299))
 
 ## Deprecated
 


libtiledbsoma/src/soma/array_buffers.cc
```diff
@@ -13,11 +13,135 @@
...
+                     if (!is_var && dim.cell_val_num() != 1) {
+                         throw TileDBSOMAError("[ArrayBuffers] Values per cell > 1 is not supported: " + name);
+                     }
+
+                     return (dim.type() == TILEDB_STRING_ASCII || dim.type() == TILEDB_STRING_UTF8) ?
+                                sizeof(uint64_t) :
+                                tiledb::impl::type_size(dim.type());
+                 }
+
+                 throw TileDBSOMAError(fmt::format("[ArrayBuffers] MIssing column name '{}'", name));
```
jp-dark:
```suggestion
                 throw TileDBSOMAError(fmt::format("[ArrayBuffers] Missing column name '{}'", name));
```

libtiledbsoma/src/soma/array_buffers.cc
```diff
@@ -13,11 +13,135 @@
...
+                    "[ArrayBuffers] Error parsing {}: '{}' ({})", CONFIG_KEY_MEMORY_BUDGET, value_str, e.what()));
+        }
+    }
+
+    ArraySchema schema = array->schema();
+    // Split memory budget to each column depending on the byte size of each columns element
+    // Var sized columns will be allocated the same as an 8 byte datatype
+
+    // Ensure minimum buffer size is multiple of 8
+    size_t memory_budget_unit =
```
jp-dark: Do/should enumerations effect the memory budget per unit?

libtiledbsoma/src/soma/managed_query.cc
```diff
@@ -147,11 +147,21 @@ void ManagedQuery::setup_read() {
 
     // Allocate and attach buffers
     LOG_TRACE("[ManagedQuery] allocate new buffers");
-    buffers_ = std::make_shared<ArrayBuffers>();
-    for (auto& name : columns_) {
-        LOG_DEBUG(fmt::format("[ManagedQuery] [{}] Adding buffer for column '{}'", name_, name));
-        buffers_->emplace(name, ColumnBuffer::create(array_, name));
-        buffers_->at(name)->attach(*query_);
+    if (!buffers_) {
```
jp-dark: I haven't fully dug into how the buffers are being shared in the Python/R layers. Will reusing the same buffers be a problem for multi-chunk reads?

jp-dark (commented):
If this works like I think it will, it should be a big improvement, although I still need to dig a little deeper into some of the nuance. It also might be worth flipping it on by default and running the unit tests against it fully enabled.

libtiledbsoma/src/soma/array_buffers.cc
```diff
@@ -13,11 +13,135 @@
...
+        } catch (const std::exception& e) {
+            throw TileDBSOMAError(
+                fmt::format(
+                    "[ArrayBuffers] Error parsing {}: '{}' ({})", CONFIG_KEY_MEMORY_BUDGET, value_str, e.what()));
+        }
+    }
+
+    ArraySchema schema = array->schema();
+    // Split memory budget to each column depending on the byte size of each columns element
+    // Var sized columns will be allocated the same as an 8 byte datatype
```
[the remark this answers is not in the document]
XanthosXanthopoulos: I doubled the default size per element to 16 bytes and added an optional expansion factor to increase that further if needed.

bkmartinjr:
Question: do all of the unit tests pass if `soma.read.use_memory_pool` is set to true?

jp-dark (approved):

PR #4302, opened 2025-11-02
[r/ci] Replace P3M Bioc Mirror

mojaveazure:
P3M's Bioc mirror is down and hasn't been restored; remove the part of `setup-r/action.yaml` that configures the P3M Bioc mirror and use the default mirror instead

Fixes [SOMA-678](https://linear.app/tiledb/issue/SOMA-678/rci-replace-p3m-bioc-mirror)


jp-dark (approved):
Thanks!

PR #4308, opened 2025-11-12
[ci] Remove the ubuntu-src dockerfile

jp-dark:
**Issue and/or context:** Closes SOMA-689

**Changes:**
Remove ubuntu-src dockerfile. This dockerfile is redundant as it install the python API indirectly through pip anyway.


aaronwolen (approved):
Nice catch!

PR #4310, opened 2025-11-13
Integrating scikit-core-build with vcpkg

ktsitsi:
**Issue and/or context:**
- Combining the vcpkg + scikit-core-build + cibuildwheel build system 

**Changes:**
- Build System Migration - `pyproject.toml` and `CMakeLists.txt` updates
- `vcpkg` Integration
- incorporate with cibuildwheel. Fixes to python-packaging.yml, python-remote-storage.yml, and ci-r-python-interop.yml
- Version Management - Simplified setuptools_scm configuration

**Notes for Reviewer:**


apis/python/scripts/setup_build_env.ps1
jp-dark: Since we don't test on windows, I would be inclined to just leave this out for now.

apis/python/CMakeLists.txt
```diff
@@ -0,0 +1,342 @@
...
+        message(WARNING "Could not parse version from _version.py, using default")
+    endif()
+elseif(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/RELEASE-VERSION")
+    # Fallback to RELEASE-VERSION file if it exists (matches version.py convention)
+    file(READ "${CMAKE_CURRENT_SOURCE_DIR}/RELEASE-VERSION" VERSION_STRING)
+    string(STRIP "${VERSION_STRING}" VERSION_STRING)
+    message(STATUS "Package version (from RELEASE-VERSION file): ${VERSION_STRING}")
+else()
+    set(VERSION_STRING "0.0.0.dev0")
+    message(WARNING "No version file found, using default version")
```
jp-dark: Same as above, I think we should fail if the version doesn't parse correctly.

apis/python/scripts/setup_build_env.ps1
ktsitsi: Sure was just there to replace sth `bld.ps1` script we have. But I am more than happy to leave it out.

jp-dark (commented):
Thanks for working on this! I have a couple small initial comments while I continue to experiment locally and dig into all the changes

apis/python/scripts/setup_build_env.sh
```diff
@@ -0,0 +1,116 @@
...
+    fi
+    
+    # Bootstrap vcpkg if needed
+    if [ ! -f "${VCPKG_ROOT}/vcpkg" ] && [ ! -f "${VCPKG_ROOT}/vcpkg.exe" ]; then
+        echo "Bootstrapping vcpkg..."
+        "${VCPKG_ROOT}/bootstrap-vcpkg.sh" || {
+            echo "Error: Failed to bootstrap vcpkg"
+            exit 1
+        }
+    fi
```
teo-tsirpanis: Vcpkg will do this by itself if needed.

apis/python/CMakeLists.txt
```diff
@@ -0,0 +1,332 @@
+cmake_minimum_required(VERSION 3.21)
+cmake_policy(SET CMP0148 NEW)
```
teo-tsirpanis: Why does the policy need to be set to `NEW`? It doesn't change anything if we don't use a deprecated find module, does it?

apis/python/CMakeLists.txt
```diff
@@ -0,0 +1,332 @@
...
+if(APPLE AND NOT DEFINED CMAKE_OSX_ARCHITECTURES)
+    # Detect system architecture
+    execute_process(
+        COMMAND uname -m
+        OUTPUT_VARIABLE DETECTED_ARCH
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+    )
+    set(CMAKE_OSX_ARCHITECTURES "${DETECTED_ARCH}" CACHE STRING "macOS architecture")
+    message(STATUS "Auto-detected macOS architecture: ${CMAKE_OSX_ARCHITECTURES}")
+endif()
```
teo-tsirpanis: This should be done automatically; what warnings does vcpkg emit?

apis/python/CMakeLists.txt
```diff
@@ -0,0 +1,332 @@
...
+            elseif(EXISTS \"${LIBTILEDBSOMA_INSTALL_PREFIX}/lib64/libtiledb.so\")
+                file(INSTALL \"${LIBTILEDBSOMA_INSTALL_PREFIX}/lib64/libtiledb.so\" 
+                     DESTINATION \"\${CMAKE_INSTALL_PREFIX}/tiledbsoma\")
+                message(STATUS \"Installed libtiledb.so from lib64/\")
+            else()
+                message(FATAL_ERROR \"Could not find libtiledb.so in ${LIBTILEDBSOMA_INSTALL_PREFIX}/lib or ${LIBTILEDBSOMA_INSTALL_PREFIX}/lib64\")
+            endif()
+        ")
+    endif()
+endif()
```
teo-tsirpanis: When building wheels, I'd rather statically link libtiledb into libtiledbsoma; it would make things more predictable, and avoid relying on pip rewriting the symbol names. Or even better, statically link libtiledb, libtiledbsoma, and all of their dependencies, into pytiledbsoma.

apis/python/CMakeLists.txt
```diff
@@ -0,0 +1,332 @@
...
+    list(APPEND LIBTILEDBSOMA_CMAKE_ARGS
+        -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+    )
+endif()
+
+if(CMAKE_CXX_COMPILER)
+    list(APPEND LIBTILEDBSOMA_CMAKE_ARGS
+        -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
+    )
+endif()
```
teo-tsirpanis: These variables will always have a value, since the `program()` command will search for compilers. And even if they didn't have a value, you don't need an `if`; `-DCMAKE_C_COMPILER=` is valid.

apis/python/CMakeLists.txt
```diff
@@ -0,0 +1,332 @@
...
+
+if(NINJA_FOUND)
+    list(APPEND LIBTILEDBSOMA_CMAKE_ARGS
+        -DCMAKE_MAKE_PROGRAM=${NINJA_EXECUTABLE}
+    )
+endif()
```
teo-tsirpanis: This can be just a `-G ${CMAKE_GENERATOR}`; no need to search for Ninja.

apis/python/CMakeLists.txt
```diff
@@ -0,0 +1,332 @@
...
+    endif()
+elseif(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/RELEASE-VERSION")
+    # Fallback to RELEASE-VERSION file if it exists (for non-scikit-build builds)
+    file(READ "${CMAKE_CURRENT_SOURCE_DIR}/RELEASE-VERSION" VERSION_STRING)
+    string(STRIP "${VERSION_STRING}" VERSION_STRING)
+    message(STATUS "Package version (from RELEASE-VERSION): ${VERSION_STRING}")
+else()
+    # This should only happen in legacy builds or manual CMake invocations
+    set(VERSION_STRING "0.0.0.dev0")
+    message(WARNING "No version available. If building via pip/cibuildwheel, this indicates a problem.")
```
teo-tsirpanis: Is there an environment variable we can use to fail if we are on pip/cibuildwheel?

apis/python/CMakeLists.txt
```diff
@@ -0,0 +1,332 @@
...
+endif()
+set(LIBTILEDBSOMA_BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/libtiledbsoma-build")
+set(LIBTILEDBSOMA_INSTALL_PREFIX "${CMAKE_CURRENT_BINARY_DIR}/install")
+
+# Include ExternalProject
+include(ExternalProject)
+
+# Prepare CMake configure arguments for libtiledbsoma
+set(LIBTILEDBSOMA_CMAKE_ARGS
+    -DCMAKE_BUILD_TYPE=Release
```
teo-tsirpanis:
```suggestion
    -DCMAKE_BUILD_TYPE=$<CONFIG>
```

apis/python/vcpkg_triplets/arm64-linux-release.cmake
```diff
@@ -0,0 +1,5 @@
+set(VCPKG_TARGET_ARCHITECTURE arm64)
+set(VCPKG_CRT_LINKAGE dynamic)
+set(VCPKG_LIBRARY_LINKAGE dynamic)
```
teo-tsirpanis: Is there a reason for dynamic library linkage? We should better use static, per my other comments.

apis/python/scripts/setup_build_env.sh
```diff
@@ -0,0 +1,116 @@
...
+        echo "Error: vcpkg not found. Please set VCPKG_ROOT or CMAKE_TOOLCHAIN_FILE environment variable."
+        echo "Searched locations:"
+        echo "  - ${REPO_ROOT}/vcpkg"
+        echo "  - ${PYTHON_DIR}/../vcpkg"
+        echo "  - ${HOME}/vcpkg"
+        echo ""
+        echo "To clone vcpkg, run:"
+        echo "  git clone https://github.com/microsoft/vcpkg.git ${REPO_ROOT}/vcpkg"
+        exit 1
+    fi
```
teo-tsirpanis: [What the Core does](https://github.com/TileDB-Inc/TileDB/blob/e96ba69177d945ab318a59e7bf569b1139c6cfcc/cmake/Options/TileDBToolchain.cmake) is check if `CMAKE_TOOLCHAIN_FILE` is set, otherwise check for `ENV{VCPKG_ROOT}`, otherwise clone vcpkg using FetchContent (with an option to skip it and have the system supply all dependencies).

.github/workflows/python-remote-storage.yml
```diff
@@ -71,20 +71,32 @@ jobs:
         with:
           name: cpp-build-release-${{ matrix.os }}
       
+      - name: Bootstrap vcpkg
+        run: |
+          if [ ! -d vcpkg ]; then
+            git clone --depth 1 https://github.com/microsoft/vcpkg.git vcpkg
```
teo-tsirpanis: Better let CMake clone vcpkg, per my other comments.

teo-tsirpanis (changes requested):
We can simplify the vcpkg integration and tiledbsoma's build system by eliminating the superbuild and following the Core's approaches.

PR #4334, opened 2025-12-06
[c++] Provide different modes of handling memory when converting to Arrow objects

XanthosXanthopoulos:
**Issue and/or context:**
#4299 #4311 provide memory optimization to reduce overall memory copies and overallocation of memory but introduced a necessary memory copy in the read path.

**Changes:**
This PR introduces the option to optimize memory management for read operations based on two modes `PERFORMANCE` and `EFFICIENCY`
- `PERFORMANCE`: The allocated buffers used for reading from TileDB are passed to the Arrow arrays and a new set of buffers is allocated for subsequent reads. This mode preserves the current behavior but can be wasteful on memory
- `EFFICIENCY`: A new set of buffers is allocated to be used by the Arrow array. These buffers are appropriately sized to exactly match the data read from TileDB. A memory copy operation is then performed for each buffer. If the read buffers are completely filled then the memory copy is avoided and the buffers are used directly by Arrow and a new buffers is allocated for TileDB reads.

**Notes for Reviewer:**
There are currently two almost equivalent implementations for `ColumnBuffer` to be used during reads. The `std::vector` one uses a custom allocator to avoid initializing the memory during resize. Under Debug the custom allocator is not being optimized by the compiler and the resulting code is orders of magnitude slower (this can be somewhat mitigated by using `-Og` instead of `-O0` optimization). In Release the compiler generated almost identical code with the array based implementation.


libtiledbsoma/src/utils/arrow_adapter.h
```diff
@@ -46,31 +46,32 @@ class SOMACoordinateSpace;
...
+
+// struct PrivateArrowBuffer {
+//     PrivateArrowBuffer(const std::shared_ptr<ArrowBuffer>& buffer)
+//         : buffer_(buffer) {
+//     }
 
-    std::shared_ptr<ArrowBuffer> buffer_;
-};
+//     std::shared_ptr<ArrowBuffer> buffer_;
+// };
```
jp-dark: Is there a reason you are commenting this out instead of deleting it?

PR #4363, opened 2026-01-07
[python] Add new SOMAContext class and deprecate SOMATileDBContext

jp-dark:
**Issue and/or context:** Closes SOMA-789

**Changes:**

Create a new `SOMAContext` that is more closely aligned with the R and libtiledbsoma `SOMAContext` objects.

### DIFF
--- a/apis/python/src/tiledbsoma/_collection.py
+++ b/apis/python/src/tiledbsoma/_collection.py
@@ -28,12 +28,13 @@ from ._dataframe import DataFrame
 from ._dense_nd_array import DenseNDArray
 from ._exception import DoesNotExistError, SOMAError, is_does_not_exist_error, map_exception_for_create
 from ._funcs import typeguard_ignore
+from ._soma_context import SOMAContext
 from ._soma_group import SOMAGroup
 from ._soma_object import SOMAObject
 from ._sparse_nd_array import SparseNDArray
 from ._types import OpenTimestamp
+from ._util import tiledb_timestamp_to_ms
 from .options import SOMATileDBContext
-from .options._soma_tiledb_context import _validate_soma_tiledb_context
 
 # A collection can hold any sub-type of SOMAObject
 CollectionElementType = TypeVar("CollectionElementType", bound=SOMAObject)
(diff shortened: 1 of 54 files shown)


apis/python/src/tiledbsoma/_soma_context.py
```diff
@@ -0,0 +1,239 @@
...
+        cls,
+        config: dict[str, str | float] | None = None,
+        threadpool: ThreadPoolExecutor | None = None,
+        replace: bool = False,
+    ) -> Self:
+        """Initializes and returns a global default context to use in SOMA operations.
+
+        This method should be called once at the beginning of your session before opening any SOMA objects
+        if you want to customize the TileDB context parameters that will apply to all subsequent operations.
+        Otherwise, a default contxt will be created automatically with standard parameters when you first open
```
aaronwolen:
```suggestion
        Otherwise, a default context will be created automatically with standard parameters when you first open
```

apis/python/src/tiledbsoma/_collection.py
```diff
@@ -94,24 +94,25 @@ def create(
...
                 uri=uri,
                 soma_type=cls._handle_type.__name__,
-                ctx=context.native_context,
+                ctx=context._handle,
                 timestamp=(0, timestamp_ms),
             )
         except SOMAError as e:
             raise map_exception_for_create(e, uri) from None
 
+        timestamp_ms = tiledb_timestamp_to_ms(tiledb_timestamp)
```
aaronwolen: I think this is redundant (unless I'm missing something) since we already converted to `ms` before the `try` block.

aaronwolen (changes requested):
I performed a fairly shallow review and found a couple minor issues that I thought I'd report now. As requested, I'll review the new api more thoroughly next.

apis/python/src/tiledbsoma/_soma_context.py
```diff
@@ -0,0 +1,236 @@
...
+    ) -> None:
+        """Internal-only initializer steps.
+
+        This function is internal; users should create a TileDB-SOMA context using `meth`:create:.
+        """
+        if _dont_call_this_use_create_instead != "tiledbsoma-internal-code":
+            name = type(self).__name__
+            raise RuntimeError(
+                f"Directly calling `{name}(...)` is intended for TileDB-SOMA internal use only. Use `name.create(...)` "
+                f"create a {name}."
```
aaronwolen:
```suggestion
                f"Directly calling `{name}(...)` is intended for TileDB-SOMA internal use only. Use `{name}.create(...)` "
                f"to create a {name}."
```

aaronwolen (approved):
Looks great. 

I suggested a fix for one minor formatting issues and had an implementation question, otherwise it's good to go.

PR #4364, opened 2026-01-07
[r] Add functions for setting/getting global context

jp-dark:
Add a method for setting the global context and one for getting the global context. Use the one to get the global context in internal context helper function.


apis/r/R/SOMAContext.R
```diff
@@ -52,3 +52,39 @@ SOMAContext <- R6::R6Class(
...
+
+#' Returns the global default context, creating one if necessary
+#'
+#' @return The context that will be used for TileDB-SOMA API when no context is provided by the user.
+#'
+#' @export
+#'
+get_default_context <- function() {
+  context <- .pkgenv[["somactx"]]
+  if (is.null(context)) {
```
aaronwolen: Same as above comment.

aaronwolen (changes requested):
Thanks!  Great documentation and user messaging.

There seems to be a lot of overlap between the getter/setter's functionality. It looks like both functions:
  - Check if a context exists
  - Create one if needed
  - Store it in `.pkgenv[["somactx"]]`
  - Return the context

 So the only differences are:
  - `set_default_context()` accepts `config` and always creates a new one
  - `get_default_context()` has no args and only creates one if missing

I'd expect `get_default_context()` to error if one wasn't set and suggest users call `set_default_context()`.  That might make the use case for each function clearer.

And then for internal use within `get_soma_context()` maybe we add an internal helper `.get_or_create_default_context()` that matches the current behavior of `get_default_context()` so factory functions automatically handle lazy creation for users.

apis/r/R/SOMAContext.R
```diff
@@ -52,3 +52,40 @@ SOMAContext <- R6::R6Class(
     .handle = NULL
   )
 )
+
+#' Create and return a global default context object
```
aaronwolen:
```suggestion
#' Create a default global context object
```

aaronwolen (approved):
Perfect! As discussed, a took a stab at expanding the docs since this concept may be unfamiliar to some R users. Also suggested updating the `set_default_context()` to invisibly return the context object, which is typical for setters.

PR #4368, opened 2026-01-09
[r] Carrara Collections Support

aaronwolen:
**Issue and/or context:** Adds support for Carrara (TileDB v3) URIs in the collection classes.

**Changes:**

- Block `set()` for Carrara collections (throws `unsupportedOperationError`)
- `add_new_*` methods handle Carrara URIs properly via a new internal `._set_element()` helper
- Member cache now refreshes for Carrara URIs to detect auto-registered children
- Added `SOMAContext$is_tiledbv2()` and `SOMAContext$is_tiledbv3()` methods
- Reduced verbosity of some internal C++ logging
- Added test cases for Carrara providing similar cover to the Python tests

**Notes for Reviewer:**

- The `add_new_collection()`, `add_new_dataframe()`, `add_new_dense_ndarray()`, and `add_new_sparse_ndarray()` methods use a new internal `._set_element()` helper that:
  - For Carrara: validates the member name matches the URI basename, then updates the local cache (server handles registration automatically)
  - For v2: calls `$set()` as before
  - These methods now also check for duplicate keys before creating new objects

- The member cache now:
  - forces cache refresh for Carrara URIs in READ/WRITE modes to detect externally-added members but skips updates in DELETE mode to preserve uncommitted deletions
  - uses `!is.null()` rather than `length()` to check whether it's populated so that deleting the last member doesn't cause a re-fetch from server

- Fixed `remove()` bug where `self$mode` should have been `self$mode()`


apis/r/R/SOMACollectionBase.R
```diff
@@ -349,13 +366,20 @@ SOMACollectionBase <- R6::R6Class(
...
     #'
     add_new_collection = function(object, key) {
       if (!inherits(object, "SOMACollectionBase")) {
         stop("'object' must be a SOMA collection", call. = FALSE)
       }
-      self$set(object, key)
+
+      private$.set_element(object, key)
       return(object)
     },
```
jp-dark: It looks like all the other `add_new_*` methods create new objects (like the Python API), but this method is just a setter. Should we change this to be more consistent?

jp-dark (commented):
I still need to review the tests and I have a question about the `add_new_collection` method, but otherwise the code changes are looking good.

jp-dark (changes requested):
The tests use the default profile instead of the profile specified in the instructions. We should either explicitly set-up the context for the R tests or change the README to specify that they use the `default` TileDB profile. Other than this and the change to the `add_new_collection` API, everything looks good.

jp-dark (approved):

PR #4375, opened 2026-01-15
[c++] Implement buffer resize and resubmission for read queries

XanthosXanthopoulos:
**Issue and/or context:** [SOMA 796](https://linear.app/tiledb/issue/SOMA-796/check-query-incomplete-status-reason-and-cancel-if-not-completing)

**Changes:**
This PR introduces the following changes:
- Enable read buffer resize in case the read query returns zero results due to insufficient buffer size. Adds a retry limit for up to 3 times except for incomplete queries which return data.
- Refactor different memory allocation scheme selection to make `ArrayBuffers` generic.
- Extend the `ColumnBuffer` public API to report maximum capacity for reads

### DIFF
--- a/libtiledbsoma/src/soma/column_buffer.h
+++ b/libtiledbsoma/src/soma/column_buffer.h
@@ -292,7 +292,7 @@ class ColumnBuffer {
     /**
      * @brief Resize the internal buffers to the given size.
      */
-    void resize(size_t size, bool preserve_data = false);
+    void resize(size_t size, size_t num_cells, bool preserve_data = false);
 
    protected:
     size_t num_cells_;
--- a/libtiledbsoma/src/soma/column_buffer.h
+++ b/libtiledbsoma/src/soma/column_buffer.h
@@ -297,14 +297,14 @@ class ColumnBuffer {
     virtual std::unique_ptr<IArrowBufferStorage> export_buffers();
 
     /**
-     * @brief Resize the internal buffers to the given size.
-     * 
-     * @param size Number of bytes to allocate for the data buffer.
-     * @param num_cells Number of cells the buffer should hold. 
-     *  For variable sized and nullable columns this dictates the size of the offsets and validity buffers.
-     * @param preserve_data If true, copy any data written in the old buffers to the new buffers.
+     * @brief Get max size of data buffer in bytes.
      */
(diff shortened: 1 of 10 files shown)


libtiledbsoma/src/soma/column_buffer.h
```diff
@@ -343,6 +345,21 @@ class ColumnBuffer {
         return is_ordered_;
     }
 
+    /**
+     * @brief Resize the internal buffers to the given size.
+     */
+    void resize(uint64_t size, uint64_t num_cells, bool preserve_data = false);
```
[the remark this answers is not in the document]
alancleary: Also add `@param` descriptions to the docstring

libtiledbsoma/src/soma/column_buffer_strategies.h
```diff
@@ -0,0 +1,51 @@
...
+
+inline constexpr std::string_view CONFIG_KEY_MEMORY_POOL_SIZE{"soma.read.memory_budget"};
+inline constexpr std::string_view CONFIG_KEY_VAR_SIZED_FACTOR{"soma.read.var_size_factor"};
+
+class ColumnBufferAllocationStrategy {
+   public:
+    virtual ~ColumnBufferAllocationStrategy() = default;
+
+    virtual std::pair<size_t, size_t> get_buffer_sizes(
+        std::variant<tiledb::Attribute, tiledb::Dimension> column) const = 0;
```
alancleary: `column` should be `const` since none of the implementations seem to modify it

libtiledbsoma/test/unit_soma_array.cc
```diff
@@ -461,6 +461,34 @@ TEST_CASE("SOMAArray: Test buffer size") {
     soma_array->close();
 }
 
+TEST_CASE("SOMAArray: Test resize") {
```
alancleary: Is there a way to constrain the buffer's memory budget? If so, this code should test that resizing beyond the budget throws an error.

alancleary (changes requested):
This is my first SOMA review so it's pretty limited in scope.

Overall everything looks good; just a couple minor change requests and some questions about the buffer's memory budget.

alancleary (approved):
Looks good!

PR #4376, opened 2026-01-15
[r] Add Carrara support to `write_soma` ingestion

aaronwolen:
**Issue and/or context:** [SOMA-758]

**Changes:**

- Updates the internal `.register_soma_object()` function to support both Carrara (TileDB v3) and TileDB v2 URIs.
- Expands tests coverage for Carrara to include Seurat object ingestion/reading.
- Adds test suite for legacy TileDB Cloud workflows to ensure continued compatibility.

**Notes for Reviewer:**

Other than the new tests, the biggest change is the refactor of `.register_soma_object()`. The original implementation was complex due to the need to:

1. Track and restore the mode (READ/WRITE/CLOSED) of both the object and parent collection
2. Use `on.exit()` to ensure original modes were restored
3. Perform a pre-check to see if a key already existed in the parent before attempting registration

The new implementation handles both Carrara and v2:

- For v3 URIS, Carrara automatically registers children with their parent collection when they're created at a nested URI, so no `set()` calls are required.
- For v2 URIs, instead of pre-checking existence (which required the mode tracking), we use `tryCatch()` to attempt registration and silently ignore "already exists" errors (matching Python's `_maybe_set()` helper in `tiledbsoma.io.ingest`) or throw a warning if `options(verbose = TRUE)` (matching the original behavior).

See the `README-cloud.md` for details on running the new cloud tests (recommended before merging).


jp-dark (approved):

apis/r/R/write_soma.R
```diff
@@ -832,37 +852,39 @@ write_soma.TsparseMatrix <- function(
...
+            "in the parent collection"
+          ),
+          class = "existingKeyWarning"
+        ))
+      } else {
+        stop(e)
+      }
+    }
   )
+
```
mojaveazure: What mode is `soma_parent()` in at the end of this function? Is it always in WRITE mode regardless of the input mode?

mojaveazure (commented):
I'm not particularly a fan of grepping through error messages to figure out what type of error we have and how to handle it, which is why I went with the permission model in the original implementation. This implementation feels fragile and prone to breakage. If this particular error was "already exists" error was classed, I would happily switch to a `tryCatch()` to catch that one particular class of error; since it's not, I would strongly prefer the permission model rather than the forgiveness model

jp-dark (approved):

PR #4377, opened 2026-01-20
[c++] Move logging functionality to a separate object library

XanthosXanthopoulos:
**Issue and/or context:** [SOMA-820](https://linear.app/tiledb/issue/SOMA-820/move-logging-functionality-to-separate-object-library)

**Changes:**
This PR includes:
- Minor refactor of logger implementation
- Move logging functionality to separate object library 

**Notes for Reviewer:**
This PR is the first from a series focused on separating non SOMA specific functionality to a separate object library.


libtiledbsoma/src/utils/version.cc
```diff
@@ -13,7 +13,9 @@
 
 #include "version.h"
 #include <tiledb/tiledb>
-#include "logger.h"
+
+#include "common/logging/impl/logger.h"
+
```
jp-dark: Nit: doesn't look like we actually use logging in this file so this can just be removed.

jp-dark (approved):
Two small optional comments, but otherwise looks good.

libtiledbsoma/src/utils/version.cc
XanthosXanthopoulos: This include is used to expose `fmt`

libtiledbsoma/src/common/logging/impl/logger.cc
```diff
@@ -31,19 +39,19 @@ namespace tiledbsoma {
 // [log level]
 // text to log...
 // %$ : end color range
-const std::string LOG_PATTERN = "%^[%Y-%m-%d %H:%M:%S.%e] [%n] [Process: %P] [Thread: %t] [%l] %v%$";
-const std::string CONSOLE_LOGGER = "tiledbsoma";
-const std::string FILE_LOGGER = "tiledbsoma-file";
+inline constexpr std::string_view LOG_PATTERN{"%^[%Y-%m-%d %H:%M:%S.%e] [%n] [Process: %P] [Thread: %t] [%l] %v%$"};
+inline constexpr std::string_view CONSOLE_LOGGER{"tiledbsoma"};
+inline constexpr std::string_view FILE_LOGGER{"tiledbsoma-file"};
```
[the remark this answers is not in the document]
XanthosXanthopoulos: Depending on the string length using a std::string you may have a heap allocation. In this case this is not relevant because LOG_PATTERN is passed by value to spdlog so any optimization is ineffective. i will revert to the old method.

PR #4378, opened 2026-01-20
[r] Test coverage for duplicate key behavior in SOMA collections

aaronwolen:
## Issue and/or context

[SOMA-818](https://linear.app/tiledb/issue/SOMA-818/r-add-test-coverage-for-duplicate-key-behavior-in-soma-collections)

## Changes

Add test coverage for duplicate key handling in R SOMA collections via `write_soma()` and `SOMACollection$set()`.

- Added two new test cases for `SOMACollection$set()` to:
  - verify that attempting to add a member with an existing key within the same session throws an error and the original member is preserved
  - verify the same duplicate key rejection behavior persists after closing and reopening the collection
- Both tests also verify that `add_new_*` methods reject existing keys with "Member 'foo' already exists" error

- Added test case for `write_soma()` to:
  - verify that `write_soma()` emits an `existingKeyWarning` condition when attempting to register an object with a duplicate key (when `verbose = TRUE`)
  - confirm the original data was preserved and not replaced

## Notes for Reviewer

The new tests revealed a bug in R's `SOMACollectionBase$set()` method. The test at `test-06-SOMACollection.R:245` was expected to throw an error but it actually allowed replacing an existing member after reopening the collection (see test failure [here](https://github.com/single-cell-data/TileDB-SOMA/actions/runs/21215626427/job/61036334902?pr=4378#step:7:524).

```r
collection$set(sdf1, name = "foo")
collection$close()
collection <- SOMACollectionOpen(uri, mode = "WRITE")
collection$set(sdf2, name = "foo")  # No error - silently allows replacement
```

This behavior differs from Python's `Collection.__setitem__()`, which correctly raises an error when attempting to add a duplicate key, both within the same session and after reopening the collection.

```python
collection["foo"] = obj1
collection["foo"] = obj2  # Raises SOMAError: "replacing key 'foo' is unsupported"
```

### DIFF
--- a/apis/r/NEWS.md
+++ b/apis/r/NEWS.md
@@ -24,6 +24,7 @@
 
 ## Fixed
 
+- Fixed `SOMACollectionBase$set()` allowing replacement of existing members after reopening the collection. The method now consistently rejects duplicate keys both within the same session and across sessions. ([#4378](https://github.com/single-cell-data/TileDB-SOMA/pull/4378))
 - The SOMA Context is only cached as an environment variable when the function `soma_context` is called directly. ([#4355](https://github.com/single-cell-data/TileDB-SOMA/pull/4355))
 - `SOMATileDBContext` no longer replaces `sm.mem.reader.sparse_global_order.ratio_array_data` when set in the input config. ([#4355](https://github.com/single-cell-data/TileDB-SOMA/pull/4355))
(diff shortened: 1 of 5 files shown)


jp-dark (approved):
There was a stray extra line in one of the comments, but otherwise looks good.

PR #4403, opened 2026-02-03
[python] Merge the `somacore` library directly into the `tiledbsoma` python library

jp-dark:
**Issue and/or context:** Closes SOMA-822

**Changes:**
This PR moves the `somacore` library directly into TileDB-SOMA and removes duplicate code between the two libraries where possible.


XanthosXanthopoulos (approved):

apis/python/tests/test_somacore.py
```diff
@@ -0,0 +1,67 @@
+import pytest
+
+from ._util import create_basic_object
+
+somacore = pytest.importorskip("somacore")
+
+# Note: using `isinstance` instead of `issubclass` for testing since `issubclass` doesn't support Protocols with properties.
+
+
+def test_datafram_protocol(tmp_path):
```
aaronwolen:
```suggestion
def test_dataframe_protocol(tmp_path):
```

aaronwolen (changes requested):
Looks great overall. The biggest issue I noticed is the `test_somacore` tests are missing `assert`. Once that's addressed, it's good to go.

aaronwolen (approved):

PR #4407, opened 2026-02-05
[python] Update Cloud tests to use SOMAContext

jp-dark:
Update the TileDB-SOMA legacy cloud remote tests to use `SOMAContext` instead of the deprecated `TileDBSOMAContext`.


aaronwolen (approved):
🚀

PR #4409, opened 2026-02-06
[c++] Move `ManagedQuery` to separate object library

XanthosXanthopoulos:
**Issue and/or context:** Part of [SOMA-821](https://linear.app/tiledb/issue/SOMA-821/move-managedquery-to-separate-object-library)

**Changes:**
This PR moves `ManagedQuery` to a separate object library and introduces a significant refactor the its implementation. Specifically
- Remove non standard library dependencies from `ManagedQuery` header
- Introduces templated private static methods with to implement templated member functions using forward declared type variables
- Refactor enumeration extension logic. 
  - There is only a single method to handle enumeration extension during writing data and as a standalone procedure.
  - Instead of templated private method to handle each type, a set of templated lambda function are used. 
  - Each enumeration value regardless of each type is treated as a byte collection. This ensures compatibility with TileDB regarding floating point types and simplifies the implementation.
- Refactor write function from Arrow array
  - Explicit type check between the input data and TileDB backend.
  - Removed unnecessary templated functions.

**Notes for Reviewer:**
Outside of the `ManagedQuery` changes the rest of the changes are simple renames and or adding forward declarations.


jp-dark (commented):
Movement looks good so far sans one small nit regarding enum naming conventions. I'll finish reviewing the `ManagedQuery` changes tomorrow.

libtiledbsoma/src/common/query/managed_query.cc
```diff
@@ -0,0 +1,1055 @@
...
+                    std::make_shared<tiledb::Enumeration>(
+                        tiledb::ArrayExperimental::get_enumeration(*ctx_, *array_, enumeration_name.value().data()))));
+
+            return extended_enumeration.has_value();
+        }
+    }
+
+    auto array_type = arrow::to_tiledb_format(schema->format);
+    if (array_type != disk_type) {
+        // Due to type incompatibilities between Arrow and TileDB several type mismaths are allowed. In case of such mismatch we should not throw.
```
jp-dark:
```suggestion
        // Due to type incompatibilities between Arrow and TileDB several type mismatches are allowed. In case of such mismatch we should not throw.
```

jp-dark (approved):

PR #4415, opened 2026-02-13
[c++/python/r] Pushdown DenseNDArray and SparseNDArray creation to C++

XanthosXanthopoulos:
**Issue and/or context:** [SOMA-863](https://linear.app/tiledb/issue/SOMA-863/sparsendarray-and-densendarray-create-pushdown)

**Changes:**

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/soma/soma_dense_ndarray.cc
+++ b/libtiledbsoma/src/soma/soma_dense_ndarray.cc
@@ -12,7 +12,14 @@
  */
 #include "soma_dense_ndarray.h"
 #include "../utils/arrow_adapter.h"
+#include "soma_attribute.h"
 #include "soma_coordinates.h"
+#include "soma_dimension.h"
+
+#include "common/arrow/utils.h"
+#include "common/logging/impl/logger.h"
+
+#include <limits>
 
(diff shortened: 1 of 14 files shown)


libtiledbsoma/src/soma/soma_dense_ndarray.cc
```diff
@@ -70,9 +77,37 @@ void SOMADenseNDArray::create(
     SOMAArray::create(ctx, uri, tiledb_schema, "SOMADenseNDArray", std::nullopt, timestamp);
 }
 
+void SOMADenseNDArray::create(
+    std::string_view uri,
+    std::string_view format,
+    std::span<const std::optional<int64_t>> shape,
```
jp-dark: Allowing ``None`` in the `shape` input is deprecated and will be removed by the next release. Given that, it might make sense to just remove the optional now or leave the handling of this in the Python/R side.

jp-dark (approved):
Looks good. One thing to note is that providing `NULL`/`None` in the shape is deprecated and will be removed in the next PR. We can simplify the code to take this into account in a follow-up when we get around to updating deprecations or you can just make the change now.

PR #4421, opened 2026-02-25
[c++/python/r] Move SOMADataFrame creation logic to C++

XanthosXanthopoulos:
**Issue and/or context:** [SOMA-864](https://linear.app/tiledb/issue/SOMA-864/dataframe-create-pushdown)

**Changes:**
This PR moves the `DataFrame` creation logic down to the common C++ layer and simplifies the R and Python implementations. Specifically
- Schema validation is now common for R and Python and implemented in C++.
- Max domains are now the same between R and Python. Because of R imposed restrictions the max domains for `int32`, `int64` and `uint64` columns are now narrower.
- Both Python and R implementations are now responsible to typecast the domain of column to the correct type before passing it to C++.

**Notes for Reviewer:**

### DIFF
--- a/libtiledbsoma/src/soma/soma_dataframe.cc
+++ b/libtiledbsoma/src/soma/soma_dataframe.cc
@@ -48,7 +48,7 @@ void SOMADataFrame::create(
     std::string_view uri,
     const common::arrow::managed_unique_ptr<ArrowSchema>& schema,
     std::span<const std::string> index_column_names,
-    std::span<const std::any> index_column_domains,
+    std::span<const DomainRange> index_column_domains,
     std::shared_ptr<SOMAContext> ctx,
     PlatformConfig platform_config,
     std::optional<TimestampRange> timestamp) {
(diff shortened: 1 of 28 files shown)


libtiledbsoma/src/soma/soma_dataframe.cc
```diff
@@ -38,6 +44,26 @@ void SOMADataFrame::create(
     SOMAArray::create(ctx, uri, tiledb_schema, "SOMADataFrame", std::nullopt, timestamp);
 }
 
+void SOMADataFrame::create(
+    std::string_view uri,
+    const common::arrow::managed_unique_ptr<ArrowSchema>& schema,
+    std::span<const std::string> index_column_names,
+    std::span<const std::any> index_column_domains,
```
jp-dark: Is there a reason you went with std::any instead of std::variant?

jp-dark (commented):
This PR is a huge improvement, but I do have a couple change requests.

Change requests:
* Unless there is a good reason to use `std::any` over `std::variant`, I think we should switch the domain types.
* A couple nits around error messages.
* TBD: I still need to think through the right way to handle the inconsistencies between R and Python.

jp-dark (approved):

PR #4423, opened 2026-03-02
[c++/python/r] Expose custom `DataType` enum

XanthosXanthopoulos:
**Issue and/or context:** Provide a `DataType` enum class to hide `tiledb_datatype_t` which can't be forward declared to avoid including TileDB headers from SOMA header files.

**Changes:**
This PR adds:
- a `DataType` enum class mirroring `tiledb_datatype_t` (except `TILEDB_ANY`)
- a set of constexpr functions to convert from `DataType` and `tiledb_datatype_t` to `DataType` and `tiledb_datatype_t` and arrow format

**Notes for Reviewer:**

### DIFF
--- a/apis/python/src/tiledbsoma/common.cc
+++ b/apis/python/src/tiledbsoma/common.cc
@@ -64,102 +64,103 @@ DomainRange encode_domain(std::string_view format, py::object domain) {
     }
 }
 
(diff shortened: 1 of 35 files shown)


jp-dark (approved):
Thanks! I have style convention comments, but otherwise this looks good.

PR #4425, opened 2026-03-04
[r] Fix date parsing failure in `.deprecation_stage()`

aaronwolen:
**Issue and/or context:** Addresses [CLOUD-3637](https://linear.app/tiledb/issue/CLOUD-3637).

The package's internal `.deprecation_stage()` function called `as.POSIXlt()` on date strings from `releases.dcf` (format: `YYYY-MM-DD`), which fails on a Conda-install of R 4.3.3  with `"character string is not in a standard unambiguous format"`.

**Changes:**

- Replaced `as.POSIXlt()` with `as.Date()` for calls that parse `YYYY-MM-DD` date strings (from `releases.dcf` and `DESCRIPTION`), which I validated works in the original environment and across R 4.(3|4|5).
- Added tests for both `.deprecate()` and `.deprecation_stage()`, which previously had zero direct test coverage.

**Notes for Reviewer:**  While adding tests I took the opportunity to clean up `.deprecation_stage()` by extracting helper functions and adding some comments, which should make the logic easier to follow. The core behavior is unchanged except for the date parsing fix.

### DIFF
--- a/apis/r/R/utils-deprecations.R
+++ b/apis/r/R/utils-deprecations.R
@@ -177,10 +177,11 @@
   # Defunct if:
   #   1. major version has been bumped, OR
   #   2. same major, minor diff >= 2, AND >= 12 weeks since deprecation release
-  is_defunct <- cur[["major"]] > dep[["major"]] ||
-    (cur[["major"]] == dep[["major"]] &&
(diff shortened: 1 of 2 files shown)


mojaveazure (approved):
Looks good, thanks for taking care of this!

PR #4426, opened 2026-03-05
[python] Conditionally apply `ContextManager` cleanup logic

XanthosXanthopoulos:
**Issue and/or context:** SOMA groups are closing their members when running outside of `ContextManager` when they go out of scope while their member are still being accessed

**Changes:**
This PR changes the cleanup of members of groups:
- When running under a `ContextManager` the behavior is unchanged
- When not running under a context manager when a group goes out of scope its members are not being closed unless there is no reference to them

**Notes for Reviewer:**

### DIFF
--- a/apis/python/tests/test_collection.py
+++ b/apis/python/tests/test_collection.py
@@ -669,3 +669,37 @@ def test_parallel_getitem(tmp_path) -> None:
     with soma.open(uri) as exp:
         results = list(tp.map(get_obsm, repeat(exp), ("X_umap", "X_umap", "X_umap", "X_umap")))
     assert all(r is results[0] for r in results)
+
+
+def test_close_behavior() -> None:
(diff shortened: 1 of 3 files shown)


jp-dark (changes requested):
This needs tests for the intended behavior.

apis/python/tests/test_collection.py
```diff
@@ -669,3 +669,37 @@ def get_obsm(exp, key) -> soma.SparseNDArray:
     with soma.open(uri) as exp:
         results = list(tp.map(get_obsm, repeat(exp), ("X_umap", "X_umap", "X_umap", "X_umap")))
     assert all(r is results[0] for r in results)
+
+
+def test_close_behavior() -> None:
```
jp-dark, line 674: This test would be cleaner using fresh objects instead of the pbmc3k. That dataset should only really be used if you need "real" data for the test.

jp-dark (changes requested):

PR #4436, opened 2026-03-17
[c++/python/r] Pushdown SOMA object management to C++

XanthosXanthopoulos:
**Issue and/or context:** Push SOMA object lifetime management to C++

**Changes:**
This PR introduces the following changes:
- `SOMACollection` caches handles to its member when first accessed. These handles are now used to construct the Python and R wrapper objects to avoid reopening or implementing the caching in the API side.
- Metadata handling is now utilizing a variant to hold the decoded TileDB metadata. This change removes the constraint to hold an open reference a TileDB array/group in order to read the metadata.
- Metadata caching is now part of the common C++ implementation for SOMAGroup and SOMAArray. Both Python and R APIs can now access the metadata in the same way and their metadata caches are now removed.
- The `close` method now accepts an optional argument to define whether members of a group should be closed once a group is closed.
- Reopening a `SOMAObject` now does not recreate the C++ handle but rather it reinitialized the same handle (but can and will make member objects invalid)

**Notes for Reviewer:**
The R devel error is not related to these changes.


jp-dark:
The CI should be fixed by a rebase or merge on main.

libtiledbsoma/src/common/metadata/metadata.h
```diff
@@ -0,0 +1,65 @@
...
+
+namespace tiledb {
+class Array;
+class Group;
+}  // namespace tiledb
+
+#pragma endregion
+
+namespace tiledbsoma::common {
+class MetadataCache {
```
jp-dark, line 38: Add doxygen comments for the class and methods in the file.

libtiledbsoma/src/common/metadata/utils.h
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+#ifndef COMMON_METADATA_UTILS_H
+#define COMMON_METADATA_UTILS_H
+
+#include "../datatype/datatype.h"
+#include "metadata.h"
+
+namespace tiledbsoma::common {
+MetadataValue decode_metadata(DataType type, uint32_t elements, const void* data);
```
jp-dark, line 25: Add doxygen comment.

libtiledbsoma/src/common/metadata/metadata.cc
```diff
@@ -0,0 +1,200 @@
...
+MetadataCache::DictMod MetadataCache::current_state_(const std::string& key) const {
+    if (mods_.contains(key)) {
+        return mods_.at(key);
+    }
+
+    return metadata_.contains(key) ? MetadataCache::DictMod::present : MetadataCache::DictMod::absent;
+}
+
+MetadataCache::DictMod MetadataCache::next_state_(DictMod current_state, const std::string& action) {
+    const std::map<std::pair<MetadataCache::DictMod, std::string>, MetadataCache::DictMod> action_map = {
```
jp-dark:
```suggestion
    static const std::map<std::pair<MetadataCache::DictMod, std::string>, MetadataCache::DictMod> action_map = {
```

libtiledbsoma/src/soma/soma_collection_base.h
```diff
@@ -55,36 +55,55 @@ class SOMACollectionBase : public SOMAGroup {
...
+     * @param mode read or write
+     * @param timestamp Optional pair indicating timestamp start and end
+     */
+    void reopen(OpenMode mode, std::optional<TimestampRange> timestamp = std::nullopt) override;
 
     /**
      * Closes the SOMACollectionBase object.
      */
-    void close();
+    void close([[maybe_unused]] bool recursive = false);
```
jp-dark:
```suggestion
    void close([[maybe_unused]] bool recursive = false) override;
```

libtiledbsoma/src/soma/soma_collection_base.h
```diff
@@ -55,36 +55,55 @@ class SOMACollectionBase : public SOMAGroup {
...
 
     /**
      * Get the SOMAObject associated with the key.
      *
      * @param key of member
      */
-    std::unique_ptr<SOMAObject> get(const std::string& key);
+    std::shared_ptr<SOMAObject> get(const std::string& key);
```
jp-dark, line 91:
```suggestion
    std::shared_ptr<SOMAObject> get(const std::string& key) override;
```

libtiledbsoma/src/soma/soma_collection_base.h
```diff
@@ -55,36 +55,55 @@ class SOMACollectionBase : public SOMAGroup {
...
+
+    void set(
+        const std::string& uri,
+        URIType uri_type,
+        const std::string& name,
+        const std::string& soma_type,
+        std::shared_ptr<SOMAObject> member,
+        bool managed = true);
```
jp-dark, line 99:
```suggestion
        bool managed = true) override;
```

libtiledbsoma/src/common/metadata/utils.cc
```diff
@@ -0,0 +1,64 @@
...
+
+#include <stdexcept>
+
+#include "../logging/impl/logger.h"
+
+namespace tiledbsoma::common {
+MetadataValue decode_metadata(DataType type, uint32_t elements, const void* data) {
+    auto decode = [&]<typename T>() -> MetadataValue {
+        if constexpr (std::is_same_v<T, std::string>) {
+            if (elements == 1 && data == nullptr) {
```
jp-dark, line 11: Is `elements == 0 && data ==  nullptr` possible? If so, we should handle that here and below.

jp-dark (commented):
I haven't finished reviewing this yet, but so far it is looking good. I have a couple small comments. As discussed offline, I'd also like to see some unit tests directly on the new metadata class.

libtiledbsoma/src/common/metadata/utils.cc
XanthosXanthopoulos: The range produced by such combination of input is valid. `std::string` explicitly states that any valid range is allowed. For `vector` there is no mention for validity in the range constructor overload. I have added some test to monitor that but so far from my testing this works as expected.

libtiledbsoma/src/soma/soma_collection_base.h
XanthosXanthopoulos: This is not an override (return types are not co-variant). This is hiding the parent method on purpose.

libtiledbsoma/src/soma/soma_collection_base.h
XanthosXanthopoulos: This is not an override (signature is different). This is hiding the parent method on purpose.

PR #4440, opened 2026-04-14
[python] Set the context in spatial import if not already set

jp-dark:
There was a bug in the `from_visium` ingestion where ingestion would fail if no `SOMAContext` was provided and no default `SOMAContext` was set.


aaronwolen (approved):
Thanks!

PR #4444, opened 2026-04-17
[r] expose `is_relative()` method on collection classes

aaronwolen:
**Issue and/or context:** [SOMA-940] The C++ layer already exposes `is_relative()` on `SOMAGroup` and the Python API uses it extensively in tests. R had no equivalent, making it difficult to write tests that verify correct URI handling or diagnose issues with mixed relative/absolute member URIs.

**Changes:** 

- Add `soma_group_member_is_relative()` Rcpp binding in `src/groups.cpp` that delegates to `SOMAGroup::is_relative()`
- Expose `is_relative(name)` as a public method on `SOMACollectionBase`, available on all collection subclasses
- Add test coverage in `test-06-SOMACollection.R` verifying `TRUE` for relative members and `FALSE` for absolute members

**Notes for Reviewer:** This PR also includes several refactors to the cloud/carrara test helpers that were needed to run those suites. 

- Extract `with_tiledb_profile()` into `helper-remote.R` to DRY out credential setup across cloud and carrara helpers; it sets `TILEDB_PROFILE_NAME` and clears `TILEDB_REST_TOKEN` to prevent interference
- Update `helper-cloud.R` to use `with_tiledb_profile()` and split `cloud_path()` into `cloud_array_path()` / `cloud_group_path()` with separate cleanup functions
- Update `helper-carrara.R` to default to the `"default"` profile and use `with_tiledb_profile()`
- Add `cleanup_array()` helper that deletes arrays via VFS


jp-dark (approved):
Thanks!

PR #4448, opened 2026-04-20
[python] Additional deprecation removals schedules for the 3.0.0 release

jp-dark:
**Changes:**

* Add a defunct section to R NEWS. We have a deprecate, defunct, and remove stage in our deprecation policy. Adding an explicit defunct header to the news helps make it more clear what the current state is.
* Remove resume ingest mode (deprecated in 1.18.0).
* Remove straggling delete-in-write-mode warning.

### DIFF
--- a/apis/python/src/tiledbsoma/io/ingest.py
+++ b/apis/python/src/tiledbsoma/io/ingest.py
@@ -39,7 +39,6 @@ import pandas as pd
 import pyarrow as pa
 import scipy.sparse as sp
 from more_itertools import batched
-from typing_extensions import deprecated
 
 # As of anndata 0.11 we get a warning importing anndata.experimental.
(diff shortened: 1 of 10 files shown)


apis/python/src/tiledbsoma/io/ingest.py
```diff
@@ -2922,7 +2706,7 @@ def _ingest_uns_ndarray(
     context: SOMAContext | None,
     *,
     use_relative_uri: bool | None,
-    ingestion_params: IngestionParams,
+    ingestion_params: IngestionParams,  # noqa: ARG001
```
aaronwolen: Should this just be removed rather than suppressed?

aaronwolen (approved):
Thanks! I had one question about a private method still accepting the `ingest_params` arg but otherwise looks ready to ship.

PR #4450, opened 2026-04-23
[r] Write ragged arrays relatively

mojaveazure:
Calling `write_soma()` on a `Seurat` v5 object with ragged arrays using an absolute file URI (eg. `/home/user/directory`) resulted in the SOMA NDArrays for the ragged `Seurat` matrices having absolute URIs instead of the intuitive relative arrays. This PR changes URI setting to be relative in this case

Modified SOMA methods:
 - `SOMACollectionBase$add_new_sparse_ndarray()`: now takes a parameter `relative`; defaults to `TRUE`
 - `write_soma()` methods for `Seurat`, `SingleCellExperiment`, and `SummarizedExperiment`: disallow passing arguments via `...`
 - `write_soma()` methods for `Assay`, `Assay5`, `DimReduc`, `Graph`, and `SeuratCommand` properly pass `relative` to other write-methods

Fixes [SOMA-906](https://linear.app/tiledb/issue/SOMA-906/r-mixed-uri-types-in-somaexperiments-created-via-write-seurat)

### DIFF
--- a/apis/r/R/SOMACollectionBase.R
+++ b/apis/r/R/SOMACollectionBase.R
@@ -381,6 +381,7 @@ SOMACollectionBase <- R6::R6Class(
     #' type of each element in the array.
     #' @param shape a vector of integers defining the shape of the array.
     #' @template param-platform-config
+    #' @param relative Add the sparse array relative to the collection
     #'
     #' @return Returns the newly-created array stored at \code{key}.
     #'
@@ -388,7 +389,8 @@ SOMACollectionBase <- R6::R6Class(
       key,
       type,
       shape,
-      platform_config = NULL
+      platform_config = NULL,
+      relative = TRUE
     ) {
       if (key %in% self$names()) {
         stop(sprintf("Member '%s' already exists", key), call. = FALSE)
@@ -402,7 +404,7 @@ SOMACollectionBase <- R6::R6Class(
         context = self$context,
         tiledb_timestamp = self$tiledb_timestamp # Cached value from $new()/SOMACollectionOpen
       )
-      private$.set_element(ndarr, key)
+      private$.set_element(ndarr, key, relative = relative)
       return(ndarr)
     },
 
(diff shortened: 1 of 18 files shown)


apis/r/tests/testthat/test-14-SeuratIngest.R
```diff
@@ -588,3 +588,83 @@ test_that("Write Seurat with BPCells layers", {
     expect_identical(th, expected = "Matrix:dgCMatrix", info = lyr)
   }
 })
+
+test_that("Ragged array uploads to S3 (SOMA-906)", {
```
jp-dark: Let's not add in an AWS test. You should be able to just check if the URIs are stored as relative or not. Feel free to to use the TileDB-R API for checking if groups are relative for now (https://github.com/TileDB-Inc/TileDB-R/blob/3760c68aa2341ad034d9d40eb88aec72c3dcd4e3/R/Group.R#L430) and I can switch it to a new libtiledbsoma API for testing in the future.

jp-dark (changes requested):
The test should remove AWS as a dependency.

apis/r/R/SOMACollectionBase.R
```diff
@@ -519,7 +519,7 @@ SOMACollectionBase <- R6::R6Class(
         private$.add_cache_member(name, object)
       } else {
         # v2: register with TileDB group
-        self$set(object, name)
+        self$set(object, name, relative = TRUE)
```
jp-dark: The bug was that we were setting members to `relative=True` instead of `relative=False` when the user provided `relative=False`. This won't fix that.

apis/r/tests/testthat/test-14-SeuratIngest.R
mojaveazure: Addressed in 3bb7806

apis/r/R/SOMACollectionBase.R
mojaveazure: Discussed and resolved offline

